PR #15853, opened 2025-09-07
CUDA: print CUDART_VERSION on init

JohannesGaessler:
This PR makes it so that `CUDART_VERSION` is printed in `ggml_cuda_init` if applicable. This information can be useful for debugging issues and I think it's easier to print it than to make users fill out issue templates.

PR #18483, opened 2025-12-30
CUDA: add log line when mxfp4 acceleration is used

am17an:
Technically we could use `std::call_once`, but that involve including `<mutex>` which will increase compile times

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -4785,6 +4785,16 @@ static ggml_backend_feature * ggml_backend_cuda_get_features(ggml_backend_reg_t 
         features.push_back({ "FA_ALL_QUANTS", "1" });
     #endif
 
+    {
+        const auto & info = ggml_cuda_info();
+        for (int id = 0; id < info.device_count; ++id) {
+            if (blackwell_mma_available(info.devices[id].cc)) {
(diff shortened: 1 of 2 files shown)


ggerganov:
I was rather thinking about updating the backend features list:

https://github.com/ggml-org/llama.cpp/blob/e2d9525a1357cccf228d5987169ec5bb3894a886/ggml/src/ggml-cuda/ggml-cuda.cu#L4750-L4799

This is then displayed as `system_info` along other features:

```
system_info: n_threads = 16 (n_threads_batch = 16) / 32 | CUDA : ARCHS = 750,1200 | USE_GRAPHS = 1 | PEER_MAX_BATCH_SIZE = 128 | CPU : SSE3 = 1 | SSSE3 = 1 | AVX = 1 | AVX_VNNI = 1 | AVX2 = 1 | F16C = 1 | FMA = 1 | BMI2 = 1 | AVX512 = 1 | AVX512_VBMI = 1 | AVX512_VNNI = 1 | AVX512_BF16 = 1 | LLAMAFILE = 1 | OPENMP = 1 | REPACK = 1 | 
```

ggerganov (approved):

JohannesGaessler:
When we're already on the topic, Diego suggested to me in https://github.com/ggml-org/llama.cpp/pull/15853 that the prints on CUDA initialization should be moved to the feature list as well.

PR #18342, opened 2025-12-24
common/grammar : replace problematic backtracking regex `[\s\S]*`

aldehir:
There are several regex patterns scattered throughout the codebase that use `[\s\S]*` or `[\s\S]*?`. This is problematic because it matches characters recursively in backtracking engines such as `std::regex`, causing stack overflows on large inputs.

**Minimal reproducing example**
```cpp
#include <iostream>
#include <regex>
#include <string>

int main() {
    std::string text = std::string(50000, 'a') + "b";
    std::regex pattern("[\\s\\S]*b"); // [\s\S]* is matched recursively
    
    std::smatch match;
    if (std::regex_match(text, match, pattern)) {
        std::cout << "Match\n";
    }
    
    return 0;
}
```

This PR replaces these instances with alternative solutions.

Fixes #17902, #18188

### src/llama-grammar.cpp

These patterns exist because the grammar sampler uses `std::regex_match()`. The only way to search for a needle is to surround it with `[\s\S]*?<pat>[\s\S]*`.

To address this, I check whether the pattern is wrapped in anchors (`^$`). If not, `std::regex_search()` is used instead. Most engines search by iteratively matching at each position, starting from the beginning. This is an improvement over recursively matching characters.

### common/sampling.cpp

To accommodate the grammar change, `COMMON_TRIGGER_TYPE_PATTERN_FULL` now wraps the pattern as `^<pat>$`. The problematic patterns are removed for `COMMON_TRIGGER_TYPE_PATTERN` and `COMMON_TRIGGER_TYPE_WORD`. This maintains existing semantics to remain compatible with chat parsing implementations.

Additionally, `COMMON_TRIGGER_TYPE_PATTERN` no longer adds an implicit capture group. The capture semantics now align with `COMMON_TRIGGER_TYPE_PATTERN_FULL` and can be used to determine the start of the input fed to the grammar sampler.

### common/chat.cpp

As an example, I removed `[\s\S]*?` from the Hermes 2 Pro implementation. I also removed `"(?:<think>[\\s\\S]*?</think>\\s*)?"` since it's optional and non-capturing—it would be consumed by `[\s\S]*?` anyway and isn't needed when using `std::regex_search()`.

I updated existing `COMMON_TRIGGER_TYPE_PATTERN` patterns to use non-capturing groups, since it now follows the same capture semantics as `COMMON_TRIGGER_TYPE_PATTERN_FULL`.

### common/regex-partial.cpp

[...]

PR #18462, opened 2025-12-29
implement new jinja template engine

ngxson:
TODO:
- [x] implement to_json
- [x] simplify `common/chat.cpp` --> all workarounds are re-grouped under a namespace `workaround`
- Follow-up PR: implement input marking on `llama-server`
- Follow-up PR: remove generic tool call - it's too costly to maintain
- Follow-up PR: scan through `not_implemented_exception` and implement them
- Follow-up PR: add notion of "skip" in test framework
- Follow-up PR: (maybe) refactor the `func_args` interface

## Motivation

This PR introduce a new jinja template engine that may (or may not?) replace `minja`

The idea started out as a learning experiment on how to use PEG parser. But I ultimately failed doing so (huge thanks to @aldehir for giving me working prototype - but we ultimately decided not to use it for now). With some insights from good friend @aldehir, @pwilkin and @ddh0, I did not give up, but continued to expand this engine to be more complete, while making some significant improvements compare to minja or any other (jinja / non-jinja) template engines out there.

Most of the code is inspired from [huggingface.js's jinja package](https://github.com/huggingface/huggingface.js), some part is simply one-to-one translation from JS code, so huge kudos to HF.js team for the initial implementation.

Less than half of the code in this PR is machine-generated (mostly for re-writing countless similar subclasses which is quite a boring task). I want to learn along the way and make creative choices, so I didn't use AI extensively.

> [!IMPORTANT]  
> "Input marking" feature is implemented in this PR, but left unused. In a follow-up PR, it will be added to server and enabled via a flag

## TESTING

This PR was tested against [my test repo](https://github.com/ngxson/test-jinja) which contains 370 templates. This new engine fails on 14 templates, which is an acceptable number (compared to 8 failed tests with Minja).

Some tests are failed on purpose, because these templates are badly designed and/or requires too many workarounds. They are hardly used in practice anyway, so it's OK to ignore them for now.

On top of that, we also have some unit tests under `tests/test-jinja.cpp` that validates the engine behavior against python Jinja2 library. Huge thanks to @aldehir for adding this.

## Key Features

- Input marking: security against special token injection
- Decoupled from `nlohmann::json`: this dependency is only used for JSON-to-internal type translation and is completely optional
- Minimal primitive types: int, float, bool, string, array, object, none, undefined
- Detailed logging: allow source tracing on error
- Clean architecture: workarounds are applied to input data before entering the runtime (see `common/chat.cpp`)

## Architecture

[...]

PR #18485, opened 2025-12-30
common : default content to an empty string

### LINKED ISSUE #18463: Eval bug: Qwen3 fails to parse tool response (Value is not an array or object: null at row 22, column 40)
### Name and Version

version: 7573 (3595ae596)
built with Clang 21.1.6 for Linux x86_64

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

Ryzen 9950X + RTX 4090

### Models

https://huggingface.co/lmstudio-community/Qwen3-14B-GGUF Q4_K_M
https://huggingface.co/lmstudio-community/Qwen3-32B-GGUF Q4_K_M

I run llama-server and communicate over the REST API. I'm using the tool calling protocol, and it is working as intended for Gemma3, Qwen3 Next, Magistral-Small-2509. When I use the Qwen3 models, there seems to be some sort of problem processing the reply with the jinja template (?).

The llama-server is run with `--flash-attn auto -ngl -1 --ctx-size 32768 --cache-type-k q8_0 --cache-type-v q8_0 --port ... --model ...`.

Here's an example request sent to `/v1/chat/completions`:

```json
{
    "messages": [
        {
            "content": "You are a language model, (...)",
            "role": "system"
        },
        {
            "content": "Who is the US president?",
            "role": "user"
        },
        {
            "content": "",
            "reasoning_content": "Okay, the user is asking who the current US president is. (...)",
            "role": "assistant",
            "tool_calls": [
                {
                    "function": {
                        "arguments": "{\"query\":\"current US president\"}",
                        "name": "search_web"
                    },
                    "id": "GpNdH0WPismQXeRx46kwiNEqiNPdIRcQ",
                    "index": 0,
                    "type": "function"
                }
            ]
        },
        {
            "content": "[\n  {\n    \"snippet\": \"In all, 45 individuals have served 47 presidencies spanning 60 four-year terms. Donald Trump is the 47th (...)",
            "name": "search_web",
            "role": "tool",
            "tool_call_id": "GpNdH0WPismQXeRx46kwiNEqiNPdIRcQ"
        }
    ],
    "model": "qwen3:14b",
    "stream": true,
    "tools": [
        {
            "function": {
                "description": "Search the web with given query. Uses network.",
                "name": "search_web",
                "parameters": {
                    "properties": {
                        "query": {
                            "description": "Search query.",
                            "type": "string"
                        }
                    },
                    "required": [
                        "query"
                    ],
                    "type": "object"
                }
            },
            "type": "function"
        },
        (...)
    ]
}
```

The error I'm getting with Qwen3 is:

```json
{"error":{"code":500,"message":"Value is not an array or object: null at row 22, column 40:\n    {%- set tool_start_length = tool_start|length %}\n    {%- set start_of_message = message.content[:tool_start_length] %}\n                                       ^\n    {%- set tool_end = \"</tool_response>\" %}\n at row 22, column 5:\n    {%- set tool_start_length = tool_start|length %}\n    {%- set start_of_message = message.content[:tool_start_length] %}\n    ^\n    {%- set tool_end = \"</tool_response>\" %}\n at row 18, column 37:\n{%- set ns = namespace(multi_step_tool=true, last_query_index=messages|length - 1) %}\n{%- for message in messages[::-1] %}\n                                    ^\n    {%- set index = (messages|length - 1) - loop.index0 %}\n at row 18, column 1:\n{%- set ns = namespace(multi_step_tool=true, last_query_index=messages|length - 1) %}\n{%- for message in messages[::-1] %}\n^\n    {%- set index = (messages|length - 1) - loop.index0 %}\n at row 1, column 1:\n{%- if tools %}\n^\n    {{- '<|im_start|>system\\n' }}\n","type":"server_error"}}
```

### First Bad Commit

_No response_

### Relevant log output

.

aldehir:
Fixes #18463

This issue was also spotted here: https://github.com/ggml-org/llama.cpp/pull/18342#issuecomment-3696536026

Templates receive `null` for content even when the client sends an empty string `""`. Most templates handle this defensively, but the Qwen3 templates do not. Even the latest https://huggingface.co/Qwen/Qwen3-14B fails when both `reasoning_content` and `tool_calls` are present.

This PR ensures `content` is always passed as an empty string if it's missing, null, or empty. The only template this breaks is the llama.cpp-specific `Deepseek R1` template, which I've patched here.

This also affects the Minja polyfill, since content will now always be present in a polyfilled assistant response. I don't see this as a problem given that the polyfill mechanism is already a workaround by nature.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -319,7 +319,7 @@ json common_chat_msgs_to_json_oaicompat(const std::vector<common_chat_msg> & msg
                 }
             }
         } else {
-            jmsg["content"] = json(); // null
+            jmsg["content"] = "";
         }
         if (!msg.reasoning_content.empty()) {
             jmsg["reasoning_content"] = msg.reasoning_content;
--- a/models/templates/llama-cpp-deepseek-r1.jinja
+++ b/models/templates/llama-cpp-deepseek-r1.jinja
@@ -38,7 +38,7 @@ Example function tool call syntax:
     {%- if message['role'] == 'user' -%}
         {{- '<｜User｜>' + message['content'] + '<｜end▁of▁sentence｜>' -}}
     {%- endif -%}
-    {%- if message['role'] == 'assistant' and message['content'] is none -%}
+    {%- if message['role'] == 'assistant' and not message['content'] -%}
         {{- '<｜Assistant｜><｜tool▁calls▁begin｜>' -}}
         {%- set ns.is_first = true -%}
         {%- for tc in message['tool_calls'] -%}
@@ -53,7 +53,7 @@ Example function tool call syntax:
         {%- endfor -%}
         {{- '<｜tool▁calls▁end｜><｜end▁of▁sentence｜>' -}}
     {%- endif -%}
-    {%- if message['role'] == 'assistant' and message['content'] is  not none -%}
+    {%- if message['role'] == 'assistant' and message['content'] -%}
         {{- flush_tool_outputs() -}}
         {%- set content = message['content'] -%}
         {%- if '</think>' in content -%}
--- a/tests/test-chat.cpp
+++ b/tests/test-chat.cpp
@@ -650,7 +650,7 @@ static void test_msgs_oaicompat_json_conversion() {
             "[\n"
             "  {\n"
             "    \"role\": \"assistant\",\n"
-            "    \"content\": null,\n"
+            "    \"content\": \"\",\n"
             "    \"tool_calls\": [\n"
             "      {\n"
             "        \"type\": \"function\",\n"
@@ -906,7 +906,8 @@ static void test_template_output_parsers() {
                       "      },\n"
                       "      \"id\": \"123456789\"\n"
                       "    }\n"
-                      "  ]\n"
+                      "  ],\n"
+                      "  \"content\": \"\"\n"
                       "}");
     }
     {
@@ -1713,7 +1714,8 @@ static void test_template_output_parsers() {
                       "      },\n"
                       "      \"id\": \"123456789\"\n"
                       "    }\n"
-                      "  ]\n"
+                      "  ],\n"
+                      "  \"content\": \"\"\n"
                       "}",
                       /* expect_grammar_triggered= */ false
         );


ggerganov (approved):

ngxson (commented):
This PR make me think about an idea for https://github.com/ggml-org/llama.cpp/pull/18462 . Currently, I can already have a "workaround" system that automatically detect which kind of workarounds should be used.

Because I can now hook into operators like `something is not none`, I can probably check if the template ever call `is not none` on `message["content"]` and only **deactivates** the workaround (so that `None` is preserved ; for other models not checking this, `content =""` is kept as-is)

Do you think it's something worth implemented? (It means we will be able to revert this change once the new engine is ready)

aldehir:
> This PR make me think about an idea for https://github.com/ggml-org/llama.cpp/pull/18462 . Currently, I can already have a "workaround" system that automatically detect which kind of workarounds should be used.

This is a cool idea, but I think a more robust solution would be static type checking. This is done similarly to how you execute the template, but instead of producing values you produce types instead. Then you can statically verify if a type is supported.

```
content: String | None
lstrip(input: String, strip_chars: String): String

{{ content | lstrip("\n") }} # Error, since lstrip does not accept None

{% if content is not None %}  # Type guard asserts that content : String
{{ content | lstrip("\n') }} # Good
{% endif %}
```

Of course, it would be a more complex implementation but a fun thought.

pwilkin:
@ngxson I think the core idea is nice, but generally I believe that we should enforce proper semantics instead of adding workarounds.

One of the reasons the code in the chat parser is so messy is that there are a lot of workarounds that are basically useless from a practical perspective - they were added to patch i.e. toolcalling capacity for models that don't really properly support tool calling anyway. The result is a gimmick that allows a 3B model to maybe possibly once call a `get_weather` tool in a demo chat, but is absolutely useless in any real-life scenario.

I think we should focus on supporting proper templates for models that are otherwise verified to be working and possibly drop polyfills that in most cases don't even accomplish what they set out to do.

ngxson:
> I think the core idea is nice, but generally I believe that we should enforce proper semantics instead of adding workarounds.

Hmm yeah that make sense. I think what we can do is to allow a fixed number of schema that can be used. For model with broken template, we can maintain a llama.cpp-specific version (IIUC this is what we are doing in `models/templates`)

@aldehir  Just to clarify that my core idea is to return more data from jinja rendering, then it's up to another logic to decide what to do with the returned result.

The current problem is that most (if not all) jinja engines simply return either a string or an exception which is not enough to (programmatically) know what's wrong / what's working.

My engine already return a list of string "parts" to mark input text. It will be trivial to also return a list of operations being applied to a specific input. In your case, the engine can return:
- `is not none` was checked against `content`
- `lstrip` filter was used on `content`

But I'm not entirely sure if this info is useful enough and/or the engine may skip checks as they are hidden under an if..else branch.

PR #18227, opened 2025-12-20
server: /v1/responses (text generation only)

openingnow:
This PR introduces minimally working openAI-compatible `/v1/responses` API by converting /v1/responses request into /v1/chat/completions request.

Only text generation is supported and several fields such as IDs (of response and messages) are omitted.

If this appears too unfinished for a merge at this stage, please let me know and I'll convert it to a draft.

PR #18486, opened 2025-12-30
server: /v1/responses (partial)

openingnow:
previous PR: #18227

Conversations need to be resolved:
- [ ] ~[initial_events](https://github.com/ggml-org/llama.cpp/pull/18227#discussion_r2647037803)~
- [ ] [bump `openai` from `requirements-tool_bench.txt`](https://redirect.github.com/ggml-org/llama.cpp/pull/12034#discussion_r2637753743)

This PR implements:
1. Converting Chat Completions requests to Responses requests, while aware of `reasoning_content`.
2. Emitting Responses SSEs

Current caveats:
- ~If there are consecutive function calls, the `response.output_item.added` event will not be emitted for the latter ones.~
- All `response.output_item.done` events are generated at the end (i.e., when there is both reasoning and a function call, `response.function_call_arguments.delta` is created before `response.output_item.done`(of reasoning))

The issue arise from `server_task_result_cmpl_partial::update` and seems to be not problematic for codex-cli since it does not check order of events.

---------------

Here's a visualization of handling reasoning texts. Let's start with `codex 'Explain this repo in one sentence'`

```toml
# config.toml
# match model name github.com/openai/codex/blob/0a568a/codex-rs/core/src/models_manager/model_info.rs#L116
model = "gpt-oss_local_gguf"
model_provider = "llama_cpp"

[model_providers.llama_cpp]
name = "llama_cpp API"
base_url = "http://127.0.0.1:8080/v1"
wire_api = "responses"
stream_idle_timeout_ms = 10000000

```

First request (Responses format):
```json
{"model":"gpt-oss_local_gguf","instructions":"You are a ...","input":[
    {"type":"message","role":"developer","content":[{"type":"input_text","text":"<permissions instructions>..."}]},
    {"type":"message","role":"user","content":[{"type":"input_text","text":"# AGENTS.md instructions for ..."}]},
    {"type":"message","role":"user","content":[{"type":"input_text","text":"<environment_context>..."}]},
    {"type":"message","role":"user","content":[{"type":"input_text","text":"Explain this repo in one sentence"}]}
],"tools":["..."],"tool_choice":"auto","parallel_tool_calls":false,"reasoning":null,"store":false,"stream":true,"include":[],"prompt_cache_key":"019bd4e7-1366-7571-93a6-f1b960ee1c59"}
```
`convert_responses_to_chatcmpl`  converts this into the Chat Completions format:

```json
{"model":"gpt-oss_local_gguf","tool_choice":"auto","parallel_tool_calls":false,"reasoning":null,"store":false,"stream":true,"include":[],"prompt_cache_key":"019bd4e7-1366-7571-93a6-f1b960ee1c59","messages":[
    {"role":"system","content":"You are a ..."},
    {"role":"developer","content":[{"text":"<permissions instructions>...","type":"text"}]},
    {"role":"user","content":[{"text":"# AGENTS.md instructions for ...","type":"text"}]},
    {"role":"user","content":[{"text":"<environment_context>...","type":"text"}]},
    {"role":"user","content":[{"text":"Explain this repo in one sentence","type":"text"}]}
],"tools":["..."]}
```

generated prompt:
```
<|start|>system<|message|>You are ChatGPT, ...<|end|>
<|start|>developer<|message|># Instructions\n\nYou are a ...\n\n# Tools...<|end|>
<|start|>user<|message|># AGENTS.md instructions for ...<|end|>
<|start|>user<|message|><environment_context>...<|end|>
<|start|>user<|message|>Explain this repo in one sentence<|end|>
<|start|>assistant
```

With the prompt, a reasoning text("We need to explain repo ...") and a function call(`ls -R`) are made.

```
    {"type":"response.created","response":{"id":"resp_WWZZqZyHtSnMteAaYg1oCkKtxeMk1mPO","object":"response","status":"in_progress"}}
    {"type":"response.in_progress","response":{"id":"resp_WWZZqZyHtSnMteAaYg1oCkKtxeMk1mPO","object":"response","status":"in_progress"}}
    {"type":"response.output_item.added","item":{"id":"rs_HL7QCzI9EEpldiojvIeiYPb6AOe3EnUi","summary":[],"type":"reasoning","content":[],"encrypted_content":"","status":"in_progress"}}
    {"type":"response.reasoning_text.delta","delta":"We","item_id":"rs_HL7QCzI9EEpldiojvIeiYPb6AOe3EnUi"}
    deltas ...
    {"type":"response.output_item.added","item":{"arguments":"","call_id":"fc_wxzvZd6LrQJetz7V9ZxjjmAFObzRzzg0","name":"shell","type":"function_call","status":"in_progress"}}
    {"type":"response.function_call_arguments.delta","delta":"{\"","item_id":"fc_wxzvZd6LrQJetz7V9ZxjjmAFObzRzzg0"}
    deltas ...
--> {"type":"response.output_item.done","item":{"id":"rs_HL7QCzI9EEpldiojvIeiYPb6AOe3EnUi","summary":[],"type":"reasoning","content":[{"text":"We need to explain repo in one sentence. Let's inspect repo.","type":"reasoning_text"}],"encrypted_content":""}}
    {"type":"response.output_item.done","item":{"type":"function_call","status":"completed","arguments":"{\"command\":[\"bash\",\"-lc\",\"ls -R\"],\"workdir\":\"./foobar\"}","call_id":"fc_wxzvZd6LrQJetz7V9ZxjjmAFObzRzzg0","name":"shell"}}
    {"type": "response.completed", ...}
(The arrow shows delayed `response.output_item.done` of reasoning)
```

Codex sends the new request after executing `ls -R`.

Second request (Responses format):
```diff
 {"model":"gpt-oss_local_gguf","instructions":"You are a ...","input":[
     {"type":"message","role":"developer","content":[{"type":"input_text","text":"<permissions instructions>..."}]},
     {"type":"message","role":"user","content":[{"type":"input_text","text":"# AGENTS.md instructions for ..."}]},
     {"type":"message","role":"user","content":[{"type":"input_text","text":"<environment_context>..."}]},
     {"type":"message","role":"user","content":[{"type":"input_text","text":"Explain this repo in one sentence"}]},
+    {"type":"reasoning","summary":[],"content":[{"type":"reasoning_text","text":"We need to explain repo in one sentence. Let's inspect repo."}],"encrypted_content":""},
+    {"type":"function_call","name":"shell","arguments":"{\"command\":[\"bash\",\"-lc\",\"ls -R\"],\"workdir\":\"./foobar\"}","call_id":"fc_wxzvZd6LrQJetz7V9ZxjjmAFObzRzzg0"},
+    {"type":"function_call_output","call_id":"fc_wxzvZd6LrQJetz7V9ZxjjmAFObzRzzg0","output":"{\"output\":\".:\\nfoo.cpp\\n\",\"metadata\":{\"exit_code\":0,\"duration_seconds\":0.0}}"}
 ],"tools":["..."],"tool_choice":"auto","parallel_tool_calls":false,"reasoning":null,"store":false,"stream":true,"include":[],"prompt_cache_key":"019bd4e7-1366-7571-93a6-f1b960ee1c59"}
```

Converted request (Chat Completions format):
```diff
 {"model":"gpt-oss_local_gguf","tool_choice":"auto","parallel_tool_calls":false,"reasoning":null,"store":false,"stream":true,"include":[],"prompt_cache_key":"019bd4e7-1366-7571-93a6-f1b960ee1c59","messages":[
     {"role":"system","content":"You are a ..."},
     {"role":"developer","content":[{"text":"<permissions instructions>...","type":"text"}]},
     {"role":"user","content":[{"text":"# AGENTS.md instructions for ...","type":"text"}]},
     {"role":"user","content":[{"text":"<environment_context>...","type":"text"}]},
     {"role":"user","content":[{"text":"Explain this repo in one sentence","type":"text"}]},
+    {"role":"assistant","tool_calls":[{"function":{"arguments":"{\"command\":[\"bash\",\"-lc\",\"ls -R\"],\"workdir\":\"./foobar\"}","name":"shell"},"id":"fc_wxzvZd6LrQJetz7V9ZxjjmAFObzRzzg0","type":"function"}],"reasoning_content":"We need to explain repo in one sentence. Let's inspect repo."},
+    {"role":"tool","content":"{\"output\":\".:\\nfoo.cpp\\n\",\"metadata\":{\"exit_code\":0,\"duration_seconds\":0.0}}","tool_call_id":"fc_wxzvZd6LrQJetz7V9ZxjjmAFObzRzzg0"}
 ],"tools":["..."]}
```

generated prompt:
```diff
 <|start|>system<|message|>You are ChatGPT, ...<|end|>
 <|start|>developer<|message|># Instructions\n\nYou are a ...\n\n# Tools\n\n...<|end|>
 <|start|>user<|message|># AGENTS.md instructions for ...<|end|>
 <|start|>user<|message|><environment_context>...<|end|>
 <|start|>user<|message|>Explain this repo in one sentence<|end|>
-<|start|>assistant
+<|start|>assistant<|channel|>analysis<|message|>We need to explain repo in one sentence. Let's inspect repo.<|end|>
+<|start|>assistant to=functions.shell<|channel|>commentary json<|message|>"{\"command\":[\"bash\",\"-lc\",\"ls -R\"],\"workdir\":\"./foobar\"}"<|call|>
+<|start|>functions.shell to=assistant<|channel|>commentary<|message|>"{\"output\":\".:\\nfoo.cpp\\n\",\"metadata\":{\"exit_code\":0,\"duration_seconds\":0.0}}"<|end|>
+<|start|>assistant
```

This generates another function call(`sed -n '1,200p' foo.cpp`) and the next (third) request is done in similar way.
```diff
 <|start|>system<|message|>You are ChatGPT, ...<|end|>
 <|start|>developer<|message|># Instructions\n\nYou are a ...\n\n# Tools\n\n...<|end|>
 <|start|>user<|message|># AGENTS.md instructions for ...<|end|>
 <|start|>user<|message|><environment_context>...<|end|>
 <|start|>user<|message|>Explain this repo in one sentence<|end|>
 <|start|>assistant<|channel|>analysis<|message|>We need to explain repo in one sentence. Let's inspect repo.<|end|>
 <|start|>assistant to=functions.shell<|channel|>commentary json<|message|>"{\"command\":[\"bash\",\"-lc\",\"ls -R\"],\"workdir\":\"./foobar\"}"<|call|>
 <|start|>functions.shell to=assistant<|channel|>commentary<|message|>"{\"output\":\".:\\nfoo.cpp\\n\",\"metadata\":{\"exit_code\":0,\"duration_seconds\":0.0}}"<|end|>
-<|start|>assistant
+<|start|>assistant<|channel|>analysis<|message|>Only one file foo.cpp. Open it.<|end|>
+<|start|>assistant to=functions.shell<|channel|>commentary json<|message|>"{\"command\":[\"bash\",\"-lc\",\"sed -n '1,200p' foo.cpp\"],\"workdir\":\"./foobar\"}"<|call|>
+<|start|>functions.shell to=assistant<|channel|>commentary<|message|>"{\"output\":\"#include <iostream>\\n\\nint main() {\\n    std::cout << \\\"Hello!\\\" << std::endl;\\n}\\n\",\"metadata\":{\"exit_code\":0,\"duration_seconds\":0.0}}"<|end|>
+<|start|>assistant
```

This generates:
- reasoning_text: "Repo contains single C++ hello world program. Provide one sentence."
- output_text:    "A single C++ file that prints “Hello!” to the console."

Next, the user types "Compile and run in one line" and the fourth request is sent. Reasoning contents from the previous turn are excluded (by the chat template).

```diff
 <|start|>system<|message|>You are ChatGPT, ...<|end|>
 <|start|>developer<|message|># Instructions\n\nYou are a ...\n\n# Tools\n\n...<|end|>
 <|start|>user<|message|># AGENTS.md instructions for ...<|end|>
 <|start|>user<|message|><environment_context>...<|end|>
 <|start|>user<|message|>Explain this repo in one sentence<|end|>
-<|start|>assistant<|channel|>analysis<|message|>We need to explain repo in one sentence. Let's inspect repo.<|end|>
 <|start|>assistant to=functions.shell<|channel|>commentary json<|message|>"{\"command\":[\"bash\",\"-lc\",\"ls -R\"],\"workdir\":\"./foobar\"}"<|call|>
 <|start|>functions.shell to=assistant<|channel|>commentary<|message|>"{\"output\":\".:\\nfoo.cpp\\n\",\"metadata\":{\"exit_code\":0,\"duration_seconds\":0.0}}"<|end|>
-<|start|>assistant<|channel|>analysis<|message|>Only one file foo.cpp. Open it.<|end|>
 <|start|>assistant to=functions.shell<|channel|>commentary json<|message|>"{\"command\":[\"bash\",\"-lc\",\"sed -n '1,200p' foo.cpp\"],\"workdir\":\"./foobar\"}"<|call|>
 <|start|>functions.shell to=assistant<|channel|>commentary<|message|>"{\"output\":\"#include <iostream>\\n\\nint main() {\\n    std::cout << \\\"Hello!\\\" << std::endl;\\n}\\n\",\"metadata\":{\"exit_code\":0,\"duration_seconds\":0.0}}"<|end|>
-<|start|>assistant
+<|start|>assistant<|channel|>final<|message|>A single C++ file that prints “Hello!” to the console.<|end|>
+<|start|>user<|message|>Compile and run in one line<|end|>
+<|start|>assistant
```

And so on.


openingnow:
Now converting tools.

Despite some inconsistencies in creating/completing `response.x.y` events, it can perform tasks with codex-cli.

config.toml
```toml
model = "gpt-oss-20b-mxfp4.gguf"
model_provider = "llama_cpp"

[model_providers.llama_cpp]
name = "llama_cpp API"
base_url = "http://127.0.0.1:8080/v1"
wire_api = "responses"
stream_idle_timeout_ms = 10000000
```

$ codex exec 'Explain this repo'
```sh
OpenAI Codex v0.77.0 (research preview)
--------
workdir: ~/foobar
model: gpt-oss-20b-mxfp4.gguf
provider: llama_cpp
approval: never
sandbox: workspace-write [workdir, /tmp, $TMPDIR]
reasoning effort: none
reasoning summaries: auto
session id: 
--------
user
Explain this repo

mcp startup: no servers

exec
bash -lc 'ls -R' in ~/foobar succeeded in 13ms:
.:
foo.cpp

exec
bash -lc "sed -n '1,200p' foo.cpp" in ~/foobar succeeded in 13ms:
#include <iostream>

int main() {
    std::cout << "Hello!" << std::endl;
}

codex
**Repo Overview**

The repository contains a single source file:

- `foo.cpp`: a minimal C++ program that prints “Hello!” to the console.

**File contents**

\```cpp
#include <iostream>

int main() {
    std::cout << "Hello!" << std::endl;
}
\```

**What it does**

When compiled and run, it outputs the greeting string `Hello!` followed by a newline. This is a simple “Hello, world” style example, likely intended to demonstrate basic C++ syntax, compilation, or as a placeholder for further development.

tokens used
22,090
```

aldehir:
> Now converting tools.

The gpt-oss models require feeding the reasoning from prior assistant tool calls. In the common library, this is handled via the `reasoning_content` field in the message. Is this something that can be handled by the stateless responses API?

ngxson:
> Is this something that can be handled by the stateless responses API?

Just note that we do support state tracking for streamed API response, documented in [server devs docs](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README-dev.md#example-trace-of-a-request)

openingnow:
OpenAI models do not provide raw reasoning text. ([model spec](https://model-spec.openai.com/2025-12-18.html))
> **Hidden chain-of-thought message**: some of OpenAI’s models can generate a hidden chain-of-thought message to reason through a problem before generating a final answer. This chain of thought is used to guide the model’s behavior, but is not exposed to the user or developer except potentially in summarized form.

As Aldehir mentioned, if an LLM request includes tool call output, reasoning contents should also be included. ([docs/function-calling](https://platform.openai.com/docs/guides/function-calling))
> for reasoning models like GPT-5 or o4-mini, any reasoning items returned in model responses with tool calls must also be passed back with tool call outputs.

I see 2 problems:
1. While there is a [field](https://platform.openai.com/docs/api-reference/responses/create#responses_create-input-input_item_list-item-reasoning-summary-text) for providing previous reasoning contents, it is for summary and not for raw reasoning text.
2. Chat Completions API does not have a field for providing reasoning text. This could be a difference between /chat/completions and /responses.

I hope inspecting codex-cli will help.

coder543:
@openingnow I think the question was regarding gpt-oss, not the closed source OpenAI models, but the closed source OpenAI models do provide full reasoning traces that you hand back to the server with each request so that the reasoning can persist between tool calls, it is just encrypted so that no one outside of OpenAI can read those reasoning traces. They are not summaries, although unencrypted summaries are _also_ available for user-facing usage.

aldehir:
@openingnow

See https://cookbook.openai.com/articles/gpt-oss/handle-raw-cot; Indeed I am only concerned about gpt-oss.

llama.cpp sends reasoning traces to the client via `reasoning_content` and accepts them back in the same field, adjusting them as needed for the template:
https://github.com/ggml-org/llama.cpp/blob/cef1d23c5a33156c44a206c1f4bc146f4db729f9/common/chat.cpp#L1944-L1957
This deviates from the recommended approach, which uses `reasoning` for the Chat Completions API, but its what llama.cpp and vLLM have settled on.

My hope is to align with OpenAI's recommendation for the Responses API.

This doesn't have to be tackled in this PR, which is basic support. I only bring it up for awareness, as it is a highly desired feature.

openingnow:
As this API is for openai compatibility and aims to be a drop-in replacement, shouldn't the behavior match with closed source models?

coder543:
Encryption of the reasoning traces is not a compatibility concern, so… I don’t see any reason to encrypt them, if that’s what you’re asking? Otherwise, I’m not sure what you’re asking.

It sounds like the desire is to have compatibility, which I agree with entirely.

Given how buggy `codex` is with the Chat API provided by llama-server today, I would definitely like it to have proper Responses API support soon, which would include passing back reasoning traces.

openingnow:
My concern is, which field does codex-cli use to deliver the reasoning contents.

For reasoning input and output, we have `summary`, `encrypted_content`, and `content`. The problem is, proprietary models will use first two fields while oss models will only use the last one. Therefore applications based on the proprietary models will look at `summary` or `encrypted_content` and append it to subsequent requests. They will not look at plain `content` field since proprietary models will not fill it.

Supporting codex-cli with reasoning + tool calling would be a great test for this PR, and here is my plan.
1. Figure out what field does codex-cli use
2. On output, append reasoning to the field
3. On input, read the field and move that into message["thinking"] (or other appropriate place)

@coder543 Does this make sense? And can you provide an edge case or dump where codex-cli with llama-server fails?

coder543:
What you wrote makes sense, but I think it's not something we should have to worry so much about.

For the Responses API path, Codex CLI just replays the `ResponseItem` items it received; it doesn’t pick any specific fields. The schema explicitly supports summary, content (raw reasoning text), and encrypted_content in the same item. See [ResponseItem::Reasoning](https://github.com/openai/codex/blob/0b53aed2d05fb96e4df634900528c3d8283af34e/codex-rs/protocol/src/models.rs#L65-L84) and [ReasoningItemContent](https://github.com/openai/codex/blob/0b53aed2d05fb96e4df634900528c3d8283af34e/codex-rs/protocol/src/models.rs#L290-L295).

On the Responses request path, the `prompt.input` array (which includes prior `ResponseItems`) is passed through unchanged into the request body:

  - `stream_prompt` builds request from `prompt.input` (https://github.com/openai/codex/blob/0b53aed2d05fb96e4df634900528c3d8283af34e/codex-rs/codex-api/src/endpoint/responses.rs#L80-L91)
  - `ResponsesRequestBuilder` uses `input` verbatim (https://github.com/openai/codex/blob/0b53aed2d05fb96e4df634900528c3d8283af34e/codex-rs/codex-api/src/requests/responses.rs#L97-L126)

Codex CLI explicitly asks for encrypted reasoning only when the model is known to support reasoning summaries: [`["reasoning.encrypted_content"]`](https://github.com/openai/codex/blob/0b53aed2d05fb96e4df634900528c3d8283af34e/codex-rs/core/src/client.rs#L207-L223)

Codex CLI seems to fully support GPT-OSS, which makes sense because OpenAI defined the spec for it. We don't have to fake the `encrypted_content` field.

If we _wanted_ to offer a CLI option to move the reasoning text into the `encrypted_content` field as a hack to support a poorly written client, we could do that, but I don't think it makes sense as the default.

Or we could pass the reasoning text back in the `encrypted_content` field when the client asks for `encrypted_content`? Which is required for a client to get `encrypted_content` from OpenAI's main API. Which Codex CLI will not do for GPT-OSS, at least not by default, since it recognizes that GPT-OSS implementations don't usually support `encrypted_content`.

This is my best understanding of the situation from poking around the code.

openingnow:
I editted main text as the explanation is too long to be in a comment.
@ngxson Is it acceptable to make events from `server_task_result_cmpl_partial::update()`?

ngxson:
I don't get what you mean. We're doing function programming here and it's unclear from your question which is the state and which is the derived state

openingnow:
The primary state would be variables related to `state.update_chat_msg(content, true, oaicompat_msg_diffs);`, which is `oaicompat_msg_diffs` and `task_result_state & state`(excluding `task_result_state::openai_responses_item_ids`).
Derived state includes `openai_responses_item_ids`(which is modified at `update()`) and `openai_responses_current_events`(which is generated SSEs from current diff chunk)

tools/server/server-task.cpp
```diff
@@ -1057,6 +1254,176 @@ json server_task_result_cmpl_final::to_json_anthropic_stream() {
...
+    if (res_type == TASK_RESPONSE_TYPE_OAI_RESP) {
+        if (state.openai_responses_item_ids.empty()) {
+            // Create response object
+            const std::string response_id = "resp_" + random_string();
+            openai_responses_current_events.push_back(json {
+                {"event", "response.created"},
+                {"data", json {
+                    {"type", "response.created"},
+                    {"response", json {
+                        {"id",     response_id},
```
ngxson: @openingnow it seems to me that `state.openai_responses_item_ids` is the internal state and the json is the derived (rendered) data from that state

can the rendering done inside `to_json`?

tools/server/server-task.cpp
```diff
@@ -1057,6 +1254,176 @@ json server_task_result_cmpl_final::to_json_anthropic_stream() {
...
+                    {"type", "response.in_progress"},
+                    {"response", json {
+                        {"id",     response_id},
+                        {"object", "response"},
+                        {"status", "in_progress"},
+                    }},
+                }},
+            });
+            state.openai_responses_item_ids.push_back(response_id);
+        }
```
ngxson: this is a bad practice in functional programming, where you derive one state `openai_responses_current_events` while at the same time updating the current state `openai_responses_item_ids`

instead, `update()` should only do `state.openai_responses_item_ids.push_back(...)`, meaning to update the internal state, then the rest must be handled by rendering method `to_json`

tools/server/server-task.cpp
```diff
@@ -1057,6 +1254,176 @@ json server_task_result_cmpl_final::to_json_anthropic_stream() {
...
+    // Pre-compute state updates based on diffs (for next chunk)
+    for (const auto & diff : oaicompat_msg_diffs) {
+        if (!diff.reasoning_content_delta.empty() && !state.anthropic_thinking_block_started) {
+            state.anthropic_thinking_block_started = true;
+        }
+        if (!diff.content_delta.empty() && !state.anthropic_text_block_started) {
+            state.anthropic_text_block_started = true;
+        }
+    }
+}
```
ngxson, line 1270: be aware that this is anthropic-related logic, I don't get why we need it here

tools/server/server-task.cpp
```diff
@@ -1057,6 +1254,176 @@ json server_task_result_cmpl_final::to_json_anthropic_stream() {
...
+                            {"name",      diff.tool_call_delta.name},
+                            {"type",      "function_call"},
+                            {"status",    "in_progress"},
+                        }},
+                    }},
+                });
+                state.openai_responses_item_ids.push_back(function_call_id);
+            }
+            if (!diff.tool_call_delta.arguments.empty()) {
+                const std::string prev_item_id = state.openai_responses_item_ids.back();
```
ngxson: IMO we should find a better way than `push_back()` then `back()` each time, current it's bad because:
1. a long response will blow up the memory because we push a lot of info into `openai_responses_item_ids`
2. `back()` can fail if the vector is empty, it is error-prone. even if you add a guard check if the vector is empty, there is always a chance a future contributor will call `back()` without checking vector size

Instead, think about writing it as more explicit states, or use a queue

tools/server/server-task.cpp
openingnow: These are simply moved from `server-task.h`. ~I'll keep them in `if (res_type == TASK_RESPONSE_TYPE_ANTHROPIC)` to clarify the intent.~

openingnow:
After comparing some possible implementations, I ended up with adding 3 variables(reasoning/message/func call) to keep ids and reuse flags(`*_block_started`) from the anthropic impl to track states.
This solves one caveat, and also removes oversophisticated vector of ID(was `openai_responses_item_ids`).

tools/server/server-task.cpp
ngxson: hmm ok I see, I thought that anthropic use another update()

tools/server/server-task.h
```diff
@@ -96,9 +97,15 @@ struct task_result_state {
...
-    bool anthropic_thinking_block_started = false;
-    bool anthropic_text_block_started = false;
+    // for OpenAI Responses and Anthropic streaming API:
+    // track output item / content block state across chunks
+    bool thinking_block_started = false;
+    bool text_block_started = false;
+
+    // for OpenAI Responses streaming API
+    const std::string oai_resp_reasoning_id = "rs_" + random_string();
+    const std::string oai_resp_message_id   = "msg_" + random_string();
```
ngxson: any dynamic values should be set inside the constructor for clarity

tools/server/server-common.cpp
```diff
@@ -1070,6 +1070,273 @@ json oaicompat_chat_params_parse(
...
+    chatcmpl_body.erase("input");
+    std::vector<json> chatcmpl_messages;
+
+    const std::string instructions = json_value(response_body, "instructions", std::string());
+    if (instructions != "") {
+        chatcmpl_messages.push_back({
+            {"role",    "system"},
+            {"content", instructions},
+        });
+        chatcmpl_body.erase("instructions");
```
ngxson, line 1090: `instructions` won't be replaced if it's an empty string. consider using `response_body.contains()` instead

tools/server/server-common.cpp
```diff
@@ -1070,6 +1070,273 @@ json oaicompat_chat_params_parse(
...
+    if (input_value.is_string()) {
+        // #responses_create-input-text_input
+        chatcmpl_messages.push_back({
+            {"role",    "user"},
+            {"content", input_value},
+        });
+    } else if (input_value.is_array()) {
+        // #responses_create-input-input_item_list
+
+        const auto exists_and_is_array = [](const json & j, const char * key) -> bool {
```
ngxson:
```suggestion
        static auto exists_and_is_array = [](const json & j, const char * key) -> bool {
```

tools/server/server-common.cpp
```diff
@@ -1070,6 +1070,273 @@ json oaicompat_chat_params_parse(
...
+            return j.contains(key) && j.at(key).is_array();
+        };
+        const auto exists_and_is_string = [](const json & j, const char * key) -> bool {
```
ngxson:
```suggestion
        static auto exists_and_is_string = [](const json & j, const char * key) -> bool {
```

tools/server/server-common.cpp
```diff
@@ -1070,6 +1070,273 @@ json oaicompat_chat_params_parse(
...
+                exists_and_is_array(item, "summary") &&
+                exists_and_is_string(item, "type") &&
+                item.at("type") == "reasoning") {
+                // #responses_create-input-input_item_list-item-reasoning
+
+                // Pack reasoning content in dummy message
+                chatcmpl_messages.push_back(json {
+                    {"role", "assistant"},
+                    {"content", json::array()},
+                    {"reasoning_content", item.at("content")[0].at("text")},
```
ngxson, line 1291: seems like `content[0]` is accessed without checking for non-empty?

tools/server/server-common.cpp
```diff
@@ -1070,6 +1070,273 @@ json oaicompat_chat_params_parse(
...
+            chatcmpl_tool["function"] = resp_tool;
+            chatcmpl_tools.push_back(chatcmpl_tool);
+        }
+        chatcmpl_body.erase("tools");
+        chatcmpl_body["tools"] = chatcmpl_tools;
+    }
+
+    if (response_body.contains("max_output_tokens")) {
+        chatcmpl_body.erase("max_output_tokens");
+        chatcmpl_body["max_tokens"] = response_body["max_output_tokens"];
```
ngxson, line 1343: doesn't look right to me, the key is erase, then accessed

tools/server/server-task.cpp
```diff
@@ -801,6 +803,186 @@ json server_task_result_cmpl_final::to_json_oaicompat_chat_stream() {
...
+            {"call_id",   tool_call.id},
+            {"name",      tool_call.name},
+        });
+    }
+
+    std::time_t t = std::time(0);
+    json res = {
+        {"completed_at", t},
+        {"created_at",   t},
+        {"id",           "resp_" + oaicompat_cmpl_id.substr(9)},
```
ngxson: it's probably better to use a dedicated `oai_resp_id` instead of derive it from `oaicompat_cmpl_id`

in the future, we will also move `oaicompat_cmpl_id` into task state

tools/server/server-task.cpp
```diff
@@ -801,6 +803,186 @@ json server_task_result_cmpl_final::to_json_oaicompat_chat_stream() {
...
+        output.push_back(output_item);
+    }
+
+    std::time_t t = std::time(0);
+    server_sent_events.push_back(json {
+        {"event", "response.completed"},
+        {"data", json {
+            {"type", "response.completed"},
+            {"response", json {
+                {"id",         "resp_" + oaicompat_cmpl_id.substr(9)},
```
ngxson: use `oai_resp_id`

tools/server/server-common.cpp
openingnow: The key is erased from `chatcmpl_body` and later accessed from `response_body`

tools/server/server-common.cpp
openingnow: 0d5e3de

tools/server/server-common.cpp
openingnow: cf83e1a

tools/server/server-common.cpp
openingnow: cf83e1a

tools/server/server-common.cpp
openingnow: ebb6438

tools/server/server-task.cpp
openingnow: 5ac23d2

tools/server/server-task.h
openingnow: 5ac23d2

tools/server/server-task.cpp
openingnow: 5ac23d2 adds `oai_resp_id` to `task_result_state`, `server_task_result_cmpl_partial`, and `server_task_result_cmpl_final`

openingnow:
Rebased to resolve conflict around `task_result_state(const common_chat_parser_params & chat_parser_params)`

tools/server/server-common.cpp
```diff
@@ -1069,6 +1069,282 @@ json oaicompat_chat_params_parse(
...
+                        }
+                        if (!input_item.contains("file_data") || !input_item.contains("filename")) {
+                            throw std::invalid_argument("Both 'file_data' and 'filename' are required");
+                        }
+                        chatcmpl_content.push_back({
+                            {"file", json {
+                                {"file_data", input_item.at("file_data")},
+                                {"filename",  input_item.at("filename")},
+                            }},
+                            {"type", "file"},
```
ngxson: I don't think we support this type yet. It should probably be converted into a text chunk (please verify), or maybe we just reject this type for now

tools/server/server-task.cpp
```diff
@@ -801,6 +803,186 @@ json server_task_result_cmpl_final::to_json_oaicompat_chat_stream() {
...
+        });
+        output.push_back(output_item);
+    }
+
+    for (const common_chat_tool_call & tool_call : oaicompat_msg.tool_calls) {
+        const json output_item = {
+            {"type",      "function_call"},
+            {"status",    "completed"},
+            {"arguments", tool_call.arguments},
+            {"call_id",   "fc_" + tool_call.id},
```
ngxson, line 949: do we expect to use `oai_resp_fc_id` here?

tools/server/server-task.cpp
```diff
@@ -1190,6 +1404,132 @@ json server_task_result_cmpl_partial::to_json_oaicompat_chat() {
...
+            oai_resp_fc_id = diff.tool_call_delta.id;
+        }
+
+        if (!diff.tool_call_delta.arguments.empty()) {
+            events.push_back(json {
+                {"event", "response.function_call_arguments.delta"},
+                {"data", json {
+                    {"type",    "response.function_call_arguments.delta"},
+                    {"delta",   diff.tool_call_delta.arguments},
+                    {"item_id", "fc_" + oai_resp_fc_id},
```
ngxson, line 1525: it's unclear to me, does `oai_resp_fc_id` value already include `fc_` prefix inside it?

tools/server/server-common.cpp
openingnow: I think it should be rejected unless `file` is supported from chat completions.

tools/server/server-task.cpp
openingnow: No, since `oai_resp_fc_id` is for keeping function call's id while generating args, it only exists in `task_result_state` and `server_task_result_cmpl_partial` and not in `server_task_result_cmpl_final`.

tools/server/server-task.cpp
openingnow: No, it does not has "fc_" prefix. It is copied from `diff.tool_call_delta.id` without any prefix.

ngxson (approved):

PR #18469, opened 2025-12-29
lora: count lora nodes in graph_max_nodes

ngxson:
Fix https://github.com/ggml-org/llama.cpp/issues/18050

Fix https://github.com/ggml-org/llama.cpp/issues/18466

Supersede https://github.com/ggml-org/llama.cpp/pull/18375

PR #18490, opened 2025-12-30
lora: make sure model keep track of associated adapters

ngxson:
Ref: https://github.com/ggml-org/llama.cpp/pull/18469#issuecomment-3699539768

I'm extracting this change to a dedicated PR for visibility


ggerganov (commented):
If we delegate it to the `llama_model` to free the adapters, then we should remove the `llama_adapter_lora_free()` from the public API.

For example, with the current change I believe we will be double freeing the adapters - once on model destruction and once more here for all `llama_adapter_lora_ptr` instances:

https://github.com/ggml-org/llama.cpp/blob/cd78e57c3aeae7b56c5843f94e0e0b83a3d8ca81/include/llama-cpp.h#L23-L26

ngxson:
Yes that makes sense. `llama_adapter_lora_free` exists because we wanted to allow hot-loading / hot-removing of lora adapter. But I don't think it's actually used that way in reality.

By removing `llama_adapter_lora_free`, we will now assume that lora adapters are an "extension" to a model, and must be loaded before the model can be used (i.e. context creation from a model)

ggerganov (approved):

ggerganov:
@ngxson This is probably good to merge?

ngxson:
Ah yes, thanks. Merging it now

PR #18492, opened 2025-12-30
Document that JSON Schema is not available to model when using response_format 

louwers:
This behavior is different from OpenAI et al since they seem to take the description of the JSON Schema into account.

I was wondering why my prompts were malfunctioning with llama.cpp, but then found out that `description` annotations are not taken into account.

I put this in the 'contributions welcome' list under the assumption that a contribution to implement this is welcome.

### DIFF
--- a/grammars/README.md
+++ b/grammars/README.md
@@ -149,7 +149,7 @@ You can use GBNF grammars:
 - To convert to a grammar ahead of time:
     - in CLI, with [examples/json_schema_to_grammar.py](../examples/json_schema_to_grammar.py)
     - in JavaScript with [json-schema-to-grammar.mjs](../tools/server/public_legacy/json-schema-to-grammar.mjs) (this is used by the [server](../tools/server)'s Web UI)
- 
+
 > [!NOTE]
 > The JSON schema is only used to constrain the model output and is not injected into the prompt. The model has no visibility into the schema, so if you want it to understand the expected structure, describe it explicitly in your prompt. This does not apply to tool calling, where schemas are injected into the prompt.
 
--- a/grammars/README.md
+++ b/grammars/README.md
@@ -217,6 +217,7 @@ Here is also a list of known limitations (contributions welcome):
 - Remote `$ref`s not supported in the C++ version (Python & JavaScript versions fetch https refs)
 - `string` [formats](https://json-schema.org/draft/2020-12/json-schema-validation#name-defined-formats) lack `uri`, `email`
 - No [`patternProperties`](https://json-schema.org/draft/2020-12/json-schema-core#name-patternproperties)
+- JSON Schema [annotations](https://json-schema.org/understanding-json-schema/reference/annotations) such as `description` are not taken into account.
 
 And a non-exhaustive list of other unsupported features that are unlikely to be implemented (hard and/or too slow to support w/ stateless grammars):
 
--- a/grammars/README.md
+++ b/grammars/README.md
@@ -149,6 +149,9 @@ You can use GBNF grammars:
 - To convert to a grammar ahead of time:
     - in CLI, with [examples/json_schema_to_grammar.py](../examples/json_schema_to_grammar.py)
     - in JavaScript with [json-schema-to-grammar.mjs](../tools/server/public_legacy/json-schema-to-grammar.mjs) (this is used by the [server](../tools/server)'s Web UI)
+ 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


aldehir:
Can you provide more info? Which model are you using? Is this when using tools or `response_format`?

louwers:
Here is an example:

http://127.0.0.1:8080/v1/chat/completions

```json
{
  "messages": [
    {
      "role": "system",
      "content": "You are a helpful assistant that extracts colors from text."
    },
    {
      "role": "user",
      "content": "Alice and Bob were at the red wedding."
    }
  ],
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "color_extraction",
      "strict": true,
      "schema": {
        "type": "object",
        "properties": {
          "color": {
            "type": "string",
            "description": "Color in hex format"
          }
        },
        "required": ["color"],
        "additionalProperties": false
      }
    }
  }
}
```

Response:

```json
{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "role": "assistant",
        "reasoning_content": "The user says: \"Alice and Bob were at the red wedding.\"\n\nWe need to extract colors from text. So we output \"red\". Probably just list the colors. The instruction: \"extract colors from text\". So answer: red. Possibly format as list. Let's output: red.",
        "content": "{\"color\": \"red\"}"
      }
    }
  ],
  "created": 1767109870,
  "model": "ggml-org/gpt-oss-120b-GGUF",
  "system_fingerprint": "b7527-7f459c9",
  "object": "chat.completion",
  "usage": {
    "completion_tokens": 74,
    "prompt_tokens": 94,
    "total_tokens": 168
  },
  "id": "chatcmpl-2vYZ7vEXuHbWDN6L0vGVGNtPKAu7gokD",
  "timings": {
    "cache_n": 0,
    "prompt_n": 94,
    "prompt_ms": 583.048,
    "prompt_per_token_ms": 6.2026382978723404,
    "prompt_per_second": 161.22171759443478,
    "predicted_n": 74,
    "predicted_ms": 1876.793,
    "predicted_per_token_ms": 25.362067567567568,
    "predicted_per_second": 39.428962064543086
  }
}
```

Expected response would be color format in hex (since it is mentioned in the description).

If you put it in the system prompt it works.

aldehir:
This applies only to `response_format` and the `--json` CLI argument, and it's not limited to annotations. For tool schemas, the entire schema is injected into the prompt unless the model does not support tool calling.

You'll get better generation if you include the entire schema in the system prompt, so the model is aware of not just the description but the entire structure. For example, this is the recommendation for gpt-oss: https://cookbook.openai.com/articles/openai-harmony#structured-output

I don't think this belongs in the "contributions welcome" section, since it would be a workaround rather than something natively supported by chat templates. How about we move it up as a note instead?

```patch
diff --git a/grammars/README.md b/grammars/README.md
index 0e93d8510..dcd28648b 100644
--- a/grammars/README.md
+++ b/grammars/README.md
@@ -150,6 +150,9 @@ You can use GBNF grammars:
     - in CLI, with [examples/json_schema_to_grammar.py](../examples/json_schema_to_grammar.py)
     - in JavaScript with [json-schema-to-grammar.mjs](../tools/server/public_legacy/json-schema-to-grammar.mjs) (this is used by the [server](../tools/server)'s Web UI)
 
+> [!NOTE]
+> The JSON schema is only used to constrain the model output and is not injected into the prompt. The model has no visibility into the schema, so if you want it to understand the expected structure, describe it explicitly in your prompt. This does not apply to tool calling, where schemas are injected into the prompt.
+
 Take a look at [tests](../tests/test-json-schema-to-grammar.cpp) to see which features are likely supported (you'll also find usage examples in https://github.com/ggml-org/llama.cpp/pull/5978, https://github.com/ggml-org/llama.cpp/pull/6659 & https://github.com/ggml-org/llama.cpp/pull/6555).
 
 ```bash
@@ -217,7 +220,6 @@ Here is also a list of known limitations (contributions welcome):
 - Remote `$ref`s not supported in the C++ version (Python & JavaScript versions fetch https refs)
 - `string` [formats](https://json-schema.org/draft/2020-12/json-schema-validation#name-defined-formats) lack `uri`, `email`
 - No [`patternProperties`](https://json-schema.org/draft/2020-12/json-schema-core#name-patternproperties)
-- JSON Schema [annotations](https://json-schema.org/understanding-json-schema/reference/annotations) such as `description` are not taken into account.
 
 And a non-exhaustive list of other unsupported features that are unlikely to be implemented (hard and/or too slow to support w/ stateless grammars):
 
```

louwers:
Thanks for clarifying!

Yes that's perfect.

aldehir (approved):

aldehir:
It's a good thing to document, thanks!

louwers:
Thank _you_!

PR #17494, opened 2025-11-25
ggml-cpu: aarm64: q4_K repack gemm and gemv implementations (dotprod only)

Alcpz:
This PR improves q4_k_q8_k gemm and gemv in arm64 using only dotprod (e.g rpi5)

Introduces Q8_Kx4 for convenience with dotprod intrinsics. Q4_K remains the same as the scales fit nicely in memory this way.
Added also a missing `DOTPROD` guard for the gemv introduced in #16739
Removed a TODO: Tests showed no meaningful difference in performance.

The PR also had to introduce the generic versions of the 8x4 gemm and gemvs. The output of the gemms and gemvs was tested `generic` vs `no repack` using a couple matrix shapes (for [ne00, ne01, ne10, ne11, ne12]):
- Various ne11 with fixed dimensions [128,128,128,ne11,1]
- ne11 with different batch sizes [128,128,128,ne11,ne12]
- Various ne00(=ne10) [ne00,128,ne00,16,2]
- Various output rows ne01 [128, ne01,128,16,2]

#### M4 max (GGML_NATIVE=OFF `-mcpu=cortex-a76+crypto+dotprod+noi8mm+nosve`)

| model                  | threads | test  |   new t/s |  ref t/s | speedup |
| ---------------------- | ------- | ----- | --------: | -------: | ------: |
| lfm2 1.2B Q4_K Medium  |       8 | pp256 |   667.63  |  344.36  |   1.94x |
| lfm2 1.2B Q4_K Medium  |       8 | tg128 |   231.36  |  209.05  |   1.11x |
| llama 8B Q4_K Medium   |       8 | pp256 |    96.09  |   51.11  |   1.88x |
| llama 8B Q4_K Medium   |       8 | tg128 |    43.52  |   36.72  |   1.18x |
| qwen3 8B Q4_K Medium   |       8 | pp256 |    90.38  |   50.35  |   1.79x |
| qwen3 8B Q4_K Medium   |       8 | tg128 |    40.77  |   35.98  |   1.13x |

new: 690ac9ac7 (7166) (GGML_CPU_REPACK=ON)
ref: 690ac9ac7 (7166) (GGML_CPU_REPACK=OFF)

#### Rpi5 (GGML_NATIVE=ON)

| model                  | threads | test  | new t/s | ref t/s | speedup |
| ---------------------- | ------- | ----- | ------: | ------: | ------: |
| lfm2 350M Q4_K Medium  |       4 | pp256 | 223.99 | 148.55  |   1.51x |
[...]

PR #18495, opened 2025-12-30
ggml: optimized runtime for x86 cpu backend and Q4_K quantized weights paired with Q8_K activations 

PeterEngler:
optimized runtime of the functions ggml_gemv_q4_K_8x8_q8_K() and ggml_gemm_q4_K_8x8_q8_K() for the x86 cpu backend

For performance enhancement, see:
result

Perplexity:
perplexity

test-backend-ops:
test-backend-ops


taronaeo:
Can you provide some benchmark differences between `upstream/master` and your PR?

PeterEngler:
Regarding the benachmarks I added some screenshots to my first post

PeterEngler:
What can I do to fix this ?
Seems to be a server problem ?

image

taronaeo:
> What can I do to fix this ? Seems to be a server problem ?
> image

I'll re-trigger the CI again and see if it's an ephemeral problem.

taronaeo:
I did a test on your PR and there seem to be a regression.

`upstream/master`
| model                          |       size |     params | backend    | threads | n_batch |            test |                  t/s |                                                                         
| ------------------------------ | ---------: | ---------: | ---------- | ------: | ------: | --------------: | -------------------: |                                                                         
| llama 7B Q4_K - Medium         |   3.80 GiB |     6.74 B | CPU        |       8 |     128 |           pp512 |         24.76 ± 0.07 |                                                                         
| llama 7B Q4_K - Medium         |   3.80 GiB |     6.74 B | CPU        |       8 |     128 |           tg128 |          9.18 ± 0.07 |

`pr/18495`
| model                          |       size |     params | backend    | threads | n_batch |            test |                  t/s |                                                                         
| ------------------------------ | ---------: | ---------: | ---------- | ------: | ------: | --------------: | -------------------: |                                                                         
| llama 7B Q4_K - Medium         |   3.80 GiB |     6.74 B | CPU        |       8 |     128 |           pp512 |         23.91 ± 0.02 |                                                                         
| llama 7B Q4_K - Medium         |   3.80 GiB |     6.74 B | CPU        |       8 |     128 |           tg128 |          8.42 ± 0.03 |                                                                         

Speedup Diff
| Model           | Test   |   t/s master |   t/s 18495 |   Speedup |
|:----------------|:-------|-------------:|------------:|----------:|
| llama 7B Q4_K_M | pp512  |        24.76 |       23.91 |      0.97 |
| llama 7B Q4_K_M | tg128  |         9.18 |        8.42 |      0.92 |

taronaeo:
Likewise, using a similar command to yours provided in the PR description:

`upstream/master`
| model                          |       size |     params | backend    | threads | cpu_mask   | n_batch |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | ---------- | ------: | --------------: | -------------------: |
| llama 7B Q4_K - Medium         |   3.80 GiB |     6.74 B | CPU        |       2 | 0x5        |     128 |           pp512 |          6.45 ± 0.00 |
| llama 7B Q4_K - Medium         |   3.80 GiB |     6.74 B | CPU        |       2 | 0x5        |     128 |           tg128 |          4.44 ± 0.05 |

`pr/18495`
| model                          |       size |     params | backend    | threads | cpu_mask   | n_batch |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | ---------- | ------: | --------------: | -------------------: |
| llama 7B Q4_K - Medium         |   3.80 GiB |     6.74 B | CPU        |       2 | 0x5        |     128 |           pp512 |          6.17 ± 0.01 |
| llama 7B Q4_K - Medium         |   3.80 GiB |     6.74 B | CPU        |       2 | 0x5        |     128 |           tg128 |          4.22 ± 0.00 |

Diff
| Model           | Test   |   t/s master |   t/s 18495 |   Speedup |
|:----------------|:-------|-------------:|------------:|----------:|
| llama 7B Q4_K_M | pp512  |         6.45 |        6.17 |      0.96 |
| llama 7B Q4_K_M | tg128  |         4.44 |        4.22 |      0.95 |

PeterEngler:
I did the optimization for the AVX2 instruction set. The reason you cannot see a better performance could be that llama-bench automatically choose e.g. AVX512 instead of AVX2, I didn't optimize AVX512.

Second it appears to me that you started the upstream/master first and after that the pr/18495
The decrease in performance that you see then is because of the cpu's thermal throttling during the first run.

And third I am well aware that this AVX2 optimization isn't of great value, because AVX2 is not used often enough, but I have seen that similar optimizations could be done regarding the AVX512 instruction set. And I am quite sure that there also is optimization potential in other architectures like CUDA or ARM.

ggml/src/ggml-cpu/arch/x86/repack.cpp
Alcpz: Out of curiosity: `kmask_3` was added above while `kmask3` is still using the original mask. Was there any reason to not extend the optimizations from the block above down here?

NIT: If it's intended I'd declare kmask3 closer to where it is used.

ggml/src/ggml-cpu/arch/x86/repack.cpp
```diff
@@ -2865,31 +2852,25 @@ void ggml_gemm_q4_K_8x8_q8_K(int n, float * GGML_RESTRICT s, size_t bs, const vo
                     const __m256i rhs_mat_0145_13_sp2 = _mm256_shuffle_epi32(rhs_mat_0145_13, 221); //B10(28-31) B11(28-31) B10(28-31) B11(28-31) B14(28-31) B15(28-31) B14(28-31) B15(28-31)
                     const __m256i rhs_mat_2367_13_sp2 = _mm256_shuffle_epi32(rhs_mat_2367_13, 221); //B12(28-31) B13(28-31) B12(28-31) B13(28-31) B16(28-31) B17(28-31) B16(28-31) B17(28-31)
 
-                    uint32_t utmp_0[4], utmp_1[4];
-
-                    // Scales and Mins of corresponding sub blocks from different Q4_K structures are stored together
+                    // Scales and Mins of corresponding sub blocks from different Q8_K structures are stored together
```
Alcpz: Q8_K should be Q4_K I think

Alcpz (commented):
👋  I've checked the PR out of curiosity as I've been working to optimize q4_K for ARM and wanted to see if there were changes that could benefit the other arch. 

In previous PRs I submitted, I found it is a bit challenging to verify that changes are correct because Perplexity doesn't check GEMV, only GEMM (https://github.com/ggml-org/llama.cpp/pull/17494#issuecomment-3580141003), and `test-backend-ops` doesn't traverse the REPACK codepath (also see how you tested the CUDA backend, not the CPU backend). There was work to improve that, but I think it's still being worked on. 

While the PR looks good, I would check some form of text generation to make sure there aren't new introduced issues.

Take my comments with a grain of salt, your performance improvement is quite good!

ggml/src/ggml-cpu/arch/x86/repack.cpp
PeterEngler: Yes, the reason was that these loop (from line 3065 to 3393) is not used often enough to contribute to the performance improvement.
But you are right, I should do it the same way as above.

And kmask3 is also used in the AVX512 implementation, and I am not able to test AVX512 with my CPU, so I think the best is to rename kmask_3 to kmask4 (to have the same naming scheme)

ggml/src/ggml-cpu/arch/x86/repack.cpp
PeterEngler: that's correct

Alcpz: My bad then!

PeterEngler:
In test-backend-ops I compared CUDO to the CPU backend, so I thought my changes would be checked against CUDA.

ggml/src/ggml-cpu/arch/x86/repack.cpp
PeterEngler: No, I meant you where correct ;-)

Alcpz: Ah... `!My bad then` :)

PR #18500, opened 2025-12-31
convert: avoid token collision when stripping ## prefix

o7si:
When converting BERT models to GGUF, the `phantom()` function transforms wordpiece tokens by stripping the `##` prefix. This can cause token collisions:

**Example:**
- Original vocab contains both `[1]` and `##[1]`
- After transformation: `##[1]` → `[1]`
- Result: Two different tokens now have the same string representation

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -5288,7 +5288,12 @@ class BertModel(TextModel):
             if tok.startswith("[") and tok.endswith("]"):
                 return tok
             if tok.startswith("##"):
-                return tok[2:]
+                suffix = tok[2:]
+                # If the suffix resembles a special token, keep the original to avoid collision
+                # E.g., preserve "##[1]" to avoid conflict with an existing "[1]" token.
+                if suffix.startswith("[") and suffix.endswith("]"):
+                    return tok
+                return suffix
             return "\u2581" + tok
         tokens = list(map(phantom, tokens))
 
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -5284,18 +5284,14 @@ class BertModel(TextModel):
         self.gguf_writer.add_token_type_count(self.hparams.get("type_vocab_size", 1))
 
         # convert to phantom space vocab
-        def phantom(tok):
-            if tok.startswith("[") and tok.endswith("]"):
+        def phantom(tok, toktype):
+            if toktype != gguf.TokenType.NORMAL:
                 return tok
             if tok.startswith("##"):
-                suffix = tok[2:]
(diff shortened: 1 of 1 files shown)


o7si:
> This probably needs a better fix, like f.ex. checking for dupes after the mapping.

Thanks @CISC. I'll explore a more robust way to detect and handle these collisions (likely checking duplicates after mapping or similar) and update the PR.

convert_hf_to_gguf.py
```diff
@@ -5288,7 +5288,12 @@ def phantom(tok):
             if tok.startswith("[") and tok.endswith("]"):
                 return tok
             if tok.startswith("##"):
-                return tok[2:]
+                suffix = tok[2:]
+                # If the suffix resembles a special token, keep the original to avoid collision
+                # E.g., preserve "##[1]" to avoid conflict with an existing "[1]" token.
+                if suffix.startswith("[") and suffix.endswith("]"):
+                    return tok
```
```suggestion
                # If the suffix collides with an existing token, pad it with zero-width-space
                if suffix in tokens:
                    return "\u200b" + suffix
```
I suggest something like this instead, since this can happen with any continuation token.

o7si:
My bad - the collision was caused by mistakenly treating `[1]` as a special token. The correct fix is to convert it to `\u2581[1]`.

convert_hf_to_gguf.py
o7si: I refactored the logic to use `toktypes` directly, which should avoid this collision issue.

convert_hf_to_gguf.py
```diff
@@ -5284,13 +5284,14 @@ def set_vocab(self):
         self.gguf_writer.add_token_type_count(self.hparams.get("type_vocab_size", 1))
 
         # convert to phantom space vocab
-        def phantom(tok):
-            if tok.startswith("[") and tok.endswith("]"):
+        def phantom(tok, toktype):
+            if toktype != gguf.TokenType.NORMAL:
```
o7si: I'm not entirely sure if the condition `toktype != gguf.TokenType.NORMAL` will correctly identify all special tokens.

CISC: Either way we might have a tokenization issue here, not quite sure this ever worked correctly, will dig a little...

```suggestion
            if toktype == gguf.TokenType.CONTROL:
```

CISC (approved):
Ok, the intention of the `[`/`]` check is that **only** control and continuation tokens should not have the phantom space prepended, therefore checking specifically for that, and only that, is the correct thing to do.

CISC:
Verified with `AutoTokenizer` and our `WPM` test-vocab `jina-v2-en`, it had a lot of `[unusedX]` tokens that were encoded incorrectly before.

PR #18012, opened 2025-12-13
Async DirectIO model loading on Linux

JTischbein:
Implements Direct I/O (uncached) file reading on Linux to improve model loading performance by bypassing the page cache. This is especially beneficial for large model files.

While mmap is fast on loading the same model multiple times, uncached read provides consistent model loading times at the speed of the sequential disk read speed. On DGX Spark loading GPT-OSS-120B-MXFP4 using mmap takes ~110s, in the following loads ~67s. With these changes it takes consistently ~10.5s. The speedup depends on the model size, the disk read speed and for sequential loading the available RAM. 

I would propose to set uncached reads as default, Windows already has async uncached IO [(PR)](https://github.com/ggml-org/llama.cpp/pull/7896)

PR #18504, opened 2025-12-31
Linux Direct I/O: handle short reads to avoid corrupted weights during inference

triplenom:
After updating to latest, I started seeing obviously wrong behavior during inference on Linux (repeated / “stuck” output patterns). Model loading looked fine, so I initially chased it as a sampling / runtime issue. Eventually it became pretty clear it was corrupted weights making it onto the device.

This is caused by the Linux Direct I/O change introduced in #18012 (Async DirectIO model loading on Linux).

Root cause is in the O_DIRECT read path. On Linux, read() with O_DIRECT can legitimately return a short read. That’s not an error, and it does not imply EOF. The current implementation effectively assumes “any successful read completes the request”, which means we can leave part of the destination buffer untouched. When that buffer is used for tensor uploads, you end up with partially stale/uninitialized bytes. The loader doesn’t necessarily fail, but inference quality goes off a cliff in exactly the kind of repetitive pattern you’d expect from corrupted weights.

Fix is intentionally minimal and only touches the direct I/O path:
- Loop until the full requested length is read (retry on EINTR).
- If we hit EOF only because an aligned direct I/O read rounded the request past the real end of the file, zero-fill the remainder and return. That’s just padding and should not be treated as a hard failure.

With this change, the inference corruption goes away on my setup with Direct I/O enabled, and no unrelated code paths are touched.

### DIFF
--- a/src/llama-mmap.cpp
+++ b/src/llama-mmap.cpp
@@ -240,9 +240,10 @@ struct llama_file::impl {
                 throw std::runtime_error("unexpectedly reached end of file");
             }
         } else {
-            bool successful = false;
-            while (!successful) {
-                off_t ret = read(fd, ptr, len);
+            size_t bytes_read = 0;
+            while (bytes_read < len) {
+                const size_t to_read = len - bytes_read;
+                ssize_t ret = ::read(fd, reinterpret_cast<char *>(ptr) + bytes_read, to_read);
 
                 if (ret == -1) {
                     if (errno == EINTR) {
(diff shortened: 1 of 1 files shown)


triplenom:
@JTischbein FYI

JTischbein:
@triplenom Thank you!

ggerganov (approved):

triplenom:
Checks look good to me (I’m seeing two canceled/skipped on my end). Thanks for reviewing!

PR #18511, opened 2025-12-31
model: add Solar Open model

### LINKED ISSUE #18508: Feature Request: support Solar-Open-100B
### Prerequisites

### Feature Description

Solar Open is Upstage's flagship 102B-parameter large language model, trained entirely from scratch

https://huggingface.co/upstage/Solar-Open-100B

### Motivation

llama.cpp will make it possible to run that model on home computers

### Possible Implementation

_No response_

HelloKS:

This PR add support for [Solar-Open-100B](https://huggingface.co/upstage/Solar-Open-100B), released by Korean startup, Upstage.

Model identifier: SolarOpenForCausalLM

It's basically GLM-4 MoE architecture, but without some feature (like MTP, bias)

**Current status:** Basic chat works. (Reasoning effort and Tooling does **NOT** work yet)

image

## TODO

- [x] Investigate chat template, or reasoning parser


pwilkin:
Architecturally, it's literally GLM4.5-Air with `num_nextn_predict_layers` equal to 0.

Template is annoying, it's semi-xml with inlined function names and JSON arguments, kind of like Ministral but with an added `tool_call_id`.

HelloKS:
From my understanding, llama.cpp appends "<|end|>" token as EOG token for some reason and this causes to model to stop after reasoning. (Same as gpt-oss).

~~Finding workaround for this.~~ Done.

aldehir:
> Template is annoying, it's semi-xml with inlined function names and JSON arguments, kind of like Ministral but with an added `tool_call_id`.

It's not too bad, should be easy to parse. Luckily, it seems to emit JSON arguments directly.

I wouldn't call it XML. Like gpt-oss, it appears designed for stream parsing which is nice.

LETS-BEE:
Unfortunately, I encountered a core dump error when testing after converting to Q4_K_M.

My environment is as follows:
CPU: Ryzen 9900X
RAM: 128GB DDR5
GPU: AMD 7900XTX * 4 ea

Here is the log:

with `solar-open-llama.cpp/build/bin/llama-server -m upstage/Solar-Open-100B-Q4_K_M.gguf --flash-attn on --port 12345 --threads 11 --threads-batch 22 --jinja --ctx-size 0 --split-mode layer --tensor-split 0.25,0.25,0.25,0.25 --seed 3407 --temp 0.8 --top-k 50 --top-p 0.95 --min-p 0.0 --repeat-penalty 1.05 --verbose`
```
...
load: control token:    627 '<|special_627|>' is not marked as EOG
load: control token:   2815 '<|special_2815|>' is not marked as EOG
load: control token:   3062 '<|special_3062|>' is not marked as EOG
load: control token:   2876 '<|special_2876|>' is not marked as EOG
load: control token:   3190 '<|special_3190|>' is not marked as EOG
load: control token:   1433 '<|special_1433|>' is not marked as EOG
```

with `solar-open-llama.cpp/build/bin/llama-server -m upstage/Solar-Open-100B-Q4_K_M.gguf --flash-attn on --port 12345 --threads 11 --threads-batch 22 --jinja --ctx-size 0 --split-mode layer --tensor-split 0.25,0.25,0.25,0.25 --seed 3407 --temp 0.8 --top-k 50 --top-p 0.95 --min-p 0.0 --repeat-penalty 1.05 -fit off --verbose`
```
...
create_tensor: loading tensor blk.47.ffn_down_exps.weight
create_tensor: loading tensor blk.47.ffn_up_exps.weight
create_tensor: loading tensor blk.47.ffn_gate_shexp.weight
create_tensor: loading tensor blk.47.ffn_down_shexp.weight
create_tensor: loading tensor blk.47.ffn_up_shexp.weight
load_tensors: tensor 'token_embd.weight' (q4_K) (and 0 others) cannot be used with preferred buffer type ROCm_Host, using CPU instead
```

HelloKS:
@LETS-BEE
> Unfortunately, I encountered a core dump error when testing after converting to Q4_K_M.
> 
> My environment is as follows: CPU: Ryzen 9900X RAM: 128GB DDR5 GPU: AMD 7900XTX * 4 ea

I don't think this PR caused segmentation fault, as this doesn't changed any tensors or ops. (it's almost just wrappers)

Check if other models working on latest commit first.

I'm working on Cuda machine so I can't really tell.

LETS-BEE:
> I don't think this PR caused segmentation fault, as this doesn't changed any tensors or ops. (it's almost just wrappers)
> 
> Check if other models working on latest commit first.
> 
> I'm working on Cuda machine so I can't really tell.

That's why I'm also puzzled. It doesn't make sense that the Solar Open model causes an error while GLM4.5-Air Q4_K_M loads without any issues.

I suspect it might be a tokenizer issue or a ROCm bug.

HelloKS:
Now basic chatting will work with proper reasoning boxes (hopefully).

btw I haven't experienced streaming tool calling yet so it might take some time to implement.

aldehir:
> btw I haven't experienced streaming tool calling yet so it might take some time to implement.

@HelloKS That should probably be a separate PR. I can provide the implementation.

HelloKS:
Yes, I should request review then. Thanks for the advice.

jacekpoplawski:
> Architecturally, it's literally GLM4.5-Air with `num_nextn_predict_layers` equal to 0.

it took me a while to understand your comment on reddit about it :)

LETS-BEE:
Ah, I'm so sorry. This is actually quite embarrassing. It turned out to be a basic permission issue that prevented VRAM allocation. I am sorry for the confusion caused by such a simple oversight.

Here is a screenshot of it working properly.

image

HelloKS:
image

I have kinda-working prototype for tooling, made with Gemini in [solar-open-tools](https://github.com/HelloKS/llama.cpp/tree/solar-open-tools) branch. You can try this as well if you really in hurry (for what? idk)

While this seems work, I respect contribution guideline so won't PR for this.

jacekpoplawski:
There may be some drama in the background (see https://github.com/sionic-ai/solar-vs-glm), so it might be worth keeping a local copy of the weights in case they disappear.

convert_hf_to_gguf.py
```diff
@@ -10604,6 +10607,30 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         return []
 
 
+@ModelBase.register("SolarOpenForCausalLM")
+class SolarOpenModel(Glm4MoeModel):
+    model_arch = gguf.MODEL_ARCH.GLM4_MOE
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
```suggestion
```

convert_hf_to_gguf.py
```diff
@@ -10604,6 +10607,30 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+    def set_vocab(self):
+        from transformers import AutoTokenizer
+        tokenizer = AutoTokenizer.from_pretrained(self.dir_model, trust_remote_code=True)
+        special_vocab = gguf.SpecialVocab(self.dir_model, load_merges=True)
+        tokens, toktypes, tokpre = self.get_vocab_base()
+        self.gguf_writer.add_tokenizer_model("gpt2")
+        self.gguf_writer.add_tokenizer_pre(tokpre)
+        self.gguf_writer.add_token_list(tokens)
+        self.gguf_writer.add_token_types(toktypes)
+        special_vocab = gguf.SpecialVocab(self.dir_model, load_merges=True)
```
```suggestion
        tokenizer = AutoTokenizer.from_pretrained(self.dir_model)
        special_vocab = gguf.SpecialVocab(self.dir_model, load_merges=True)
        tokens, toktypes, tokpre = self.get_vocab_base()
        self.gguf_writer.add_tokenizer_model("gpt2")
        self.gguf_writer.add_tokenizer_pre(tokpre)
        self.gguf_writer.add_token_list(tokens)
        self.gguf_writer.add_token_types(toktypes)
```

convert_hf_to_gguf.py
```diff
@@ -10604,6 +10607,26 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         return []
 
 
+@ModelBase.register("SolarOpenForCausalLM")
+class SolarOpenModel(Glm4MoeModel):
+    model_arch = gguf.MODEL_ARCH.GLM4_MOE
+
+    def set_vocab(self):
+        from transformers import AutoTokenizer
+        tokenizer = AutoTokenizer.from_pretrained(self.dir_model, trust_remote_code=True)
```
```suggestion
        tokenizer = AutoTokenizer.from_pretrained(self.dir_model)
```
Any reason for keeping this? There's no custom tokenizer, it should not be needed and the warning can be ignored.

HelloKS: No, I think that's my mistake during copy & pasting from other models.

CISC (approved):
LGTM, @aldehir / @pwilkin gets final approval.

common/chat-parser.cpp
```diff
@@ -1395,6 +1395,14 @@ static void common_chat_parse_seed_oss(common_chat_msg_parser & builder) {
     builder.consume_reasoning_with_xml_tool_calls(form, "<seed:think>", "</seed:think>");
 }
 
+static void common_chat_parse_solar_open(common_chat_msg_parser & builder) {
+    builder.try_parse_reasoning("<|think|>", "<|end|>assistant");
```
pwilkin: This is actually incorrect.
The model seems to have the weird behavior where it puts reasoning content and normal content on separate assistant blocks. Look at the template:

```jinja
<html>
<body>
{%- elif message.role == 'assistant' -%}
{%- if think_render_option == "all" -%}
{%- if message.reasoning -%}
{{- "<\|begin\|>assistant<\|think\|>" + message.reasoning + "<\|end\|>" }}
{%- endif -%}
{%- elif think_render_option == "lastthink" -%}
{%- if message.reasoning and loop.index0 > last_user_idx.value -%}
{{- "<\|begin\|>assistant<\|think\|>" + message.reasoning + "<\|end\|>" }}
{%- endif -%}
{%- endif -%}
 {%- if message.tool_calls -%}
{{- "<\|begin\|>assistant<\|tool_calls\|>" }}
{%- for tool_call in message.tool_calls -%}
{{- "<\|tool_call:begin\|>" + tool_call.id +"<\|tool_call:name\|>" + tool_call.function.name + "<\|tool_call:args\|>" }}
{{- render_tool_arguments(tool_call.function.arguments) }}
{{- "<\|tool_call:end\|>" }}
{%- endfor -%}
{{- "<\|calls\|>" }}
{%- else -%}
{{- "<\|begin\|>assistant<\|content\|>" + message.content + "<\|end\|>" }}
{%- endif -%}
{%- endif -%}
```

So the proper end for reasoning would actually seem to be:

```
builder.try_parse_reasoning("<|think|>", "<|end|>\n<|begin|>assistant<|content|>");
```

pwilkin (changes requested):
See comments on parser. Please use `scripts/jinja/jinja-tester.py` to generate messages with and without reasoning content and compare.

aldehir:
Needs a basic implementation in `src/llama-chat.{cpp,h}` to handle when `--jinja` is not passed in. It currently crashes with `llama-completion` sans `--jinja`. You can use the `OPENAI_MOE` aka gpt-oss one as an example, they're very similar.

common/chat-parser.cpp
HelloKS: I think it's just `<|end|><|begin|>assistant<|content|>`, without `\n`.

image

pwilkin: Aight, that's possible, but just for safety see what the template renders when you pass in a tool call.

HelloKS: "tools" field only used for tools section of system prompt.
image

and openai compatible client (cherry studio) /w tools list also works. (btw tool calling does not work at this moment, will be in separate PR.. from others)
image

aldehir: <details>
<summary>Example model output: tool call</summary>

```
<|think|>The user asks: "What's the weather in both San Francisco and New York?" We have a tool to get weather for a location. We need to call get_weather for both locations. We can make two calls. The tool expects location string like "San Francisco, CA". For New York we need "New York, NY". We need to call the tool twice. The instructions: "You are ChatGPT... You can use the tool." So we need to call the get_weather function for each location. The tool returns weather data. We need to combine the results and provide to user. We need to handle that we can only call the tool and then respond. So we should first call get_weather for San Francisco, then for New York.

But we need to be careful: The function may return data. We need to combine the two calls. We can either call sequentially or combine both. Possibly we can call the function twice. According to the guidelines, we can call the function multiple times. So we should do two calls.

We need to produce calls in the appropriate format. The tool calls are done via the tool. So we need to output a JSON object with the function call. The system expects us to produce a function call. The format: 
{
  "name": "get_weather",
  "arguments": {"location": "..."}
}

We need to do two calls. Possibly we can combine them into one call with multiple locations? The function only takes a single location. So we need two calls.

The best approach: call get_weather for San Francisco, then for New York, then combine the results in the response.

Thus, we will output two tool calls. The system may allow multiple calls. Usually we can only call one at a time. But we can chain them: call for San Francisco, then after we get the result, call for New York. Or we can combine them into one call with separate calls? The spec says "You may invoke one or more tools". So we can do multiple calls sequentially. The assistant can output multiple calls. The typical pattern: we can output a single function call, then the system returns the result, then we can call another function. But the user wants both. Possibly we can do two calls in a single response? Usually the system expects a single function call per response. But the instruction says "You may invoke one or more tools". So we could call them sequentially. But we need to be careful: The tool calls are separate messages. The assistant can only output one function call at a time? The guidelines: The assistant can call a function by sending a JSON with "name" and "arguments". Usually the assistant can only output one function call at a time. But if we want to get both, we can call one, get result, then call another. However, the user wants both. We can combine them into a single function that returns both? The function does not accept multiple locations. So we need to call twice.

Thus, we can do the following: First call get_weather for San Francisco. Then after we get the result, call get_weather for New York. Then combine the results and respond.

But the user might want a combined answer. We can do the two calls in a single response? The spec says "You may invoke one or more tools". It might be okay to call both in the same response. However, the system may only accept one function call at a time. But the instruction "You may invoke one or more tools" suggests we can call multiple. So we can output both calls. But the typical pattern is that the assistant can output a function call, then the system returns the result, then the assistant can output another function call. So we need to orchestrate.

Given the constraints, we could do the following: First call get_weather for San Francisco. Then after receiving the result, call get_weather for New York. Then combine.

We need to produce a final answer with the weather info.

Thus, we should do the function calls. Let's see the function definitions. The function get_weather takes location string. So we need to call with "San Francisco, CA" and "New York, NY". We'll need to pass those strings.

We'll need to output the calls in the correct JSON format.

The assistant should respond with a function call. Then the system will respond with the result. Then we call again. But we can also combine them in a single response with two function calls. However, the typical pattern is that the assistant can only send one function call at a time. If we try to send two function calls in the same response, it might cause an error. But the spec says "You may invoke one or more tools". So perhaps we can send both calls sequentially. Let's try to produce a response with a function call for San Francisco, then after we get the result, we call for New York.

But we need to decide what to output now. The user wants both. We can do the first call. Then after we get the result, we can call the second. But we as ChatGPT cannot see the result of the function call until we get it. So we need to call the function for San Francisco first. Then after we get the result, we call the function for New York. Then combine.

Thus, the first step is to call get_weather for San Francisco. So we output a function call.

Now, we need to format it correctly: 
{
  "name": "get_weather",
  "arguments": {
    "location": "San Francisco, CA"
  }
}

Then after we get the result, we call for New York.

But the user wants both. We could do a combined call? The function cannot accept multiple locations. So we need to call twice.

Thus, we will do the first call. Then after we get the result, we call the second.

But we need to output the function call now. Let's do that.

<|end|><|begin|>assistant<|tool_calls|><|tool_call:begin|>m1z76p656j<|tool_call:name|>get_weather<|tool_call:args|>{"location": "San Francisco, CA"}<|tool_call:end|>
```
</details>

```
<|think|>... reasoning ...<|end|><|begin|>assistant<|tool_calls|><|tool_call:begin|>m1z76p656j<|tool_call:name|>get_weather<|tool_call:args|>{"location": "San Francisco, CA"}<|tool_call:end|>
```

More analogous to the Harmony response format than XML.

HelloKS:
> Needs a basic implementation in `src/llama-chat.{cpp,h}` to handle when `--jinja` is not passed in. It currently crashes with `llama-completion` sans `--jinja`. You can use the `OPENAI_MOE` aka gpt-oss one as an example, they're very similar.

I added basic chat template in llama-chat, and tested with llama-completion.

Little problem is, it shows `<|end|>assistant` at the end of reasoning. Is there any way to implement hiding this?
image

aldehir (approved):
> Little problem is, it shows <|end|>assistant at the end of reasoning. Is there any way to implement hiding this?

It's pretty much the same problem as gpt-oss. You can't really fix this, because it requires always parsing the output. I think it's fine, most users will use `llama-cli` or the `--jinja` flag anyway. So long as it doesn't crash.

Looks good to me! Thanks 😊

common/chat-parser.cpp
aldehir: Here is a template rendering:

```
<|begin|>system<|content|>## Provider System Prompt

You are Solar Open 100B, a large language model trained by Upstage AI, a Korean startup. Your knowledge cutoff is 2025-07. The current date is 2026-01-01.<|end|><|begin|>user<|content|>What's the weather like in San Francisco?<|end|><|begin|>assistant<|tool_calls|><|tool_call:begin|>call_abc123<|tool_call:name|>get_weather<|tool_call:args|>{"location": "San Francisco, CA"}<|tool_call:end|><|calls|><|begin|>tool<|tool_response|><|tool_response:begin|>call_abc123<|tool_response:name|>get_weather<|tool_response:result|>{"temperature": 68, "conditions": "partly cloudy", "humidity": 72}<|tool_response:end|><|end|><|begin|>assistant
```

CISC:
> Needs a basic implementation in `src/llama-chat.{cpp,h}` to handle when `--jinja` is not passed in. It currently crashes with `llama-completion` sans `--jinja`.

Not strictly a prerequisite anymore, we have several supported models without basic chat support.

common/chat-parser.cpp
pwilkin: Aight, thanks, that means no newline is OK.

pwilkin: @HelloKS not tools, tool calls. You have to render an assistant message with a `tool_call` section to see how it renders tool calls. But @aldehir already confirmed it's fine.

pwilkin (approved):
Looks good now.

common/chat-parser.cpp
HelloKS: > @HelloKS not tools, tool calls. You have to render an assistant message with a `tool_call` section to see how it renders tool calls. But @aldehir already confirmed it's fine.

Oops. My bad. Thanks for confirmation.

aldehir:
Comprehensive parsing added in #18540.

PR #18513, opened 2025-12-31
GGML RPC use unordered_map::reserve and emplace

struct:
Using `unordered_map::reserve` for `tensor_ptrs` and `tensor_map` presizes the arrays so inserts in these loops don’t trigger rehashing. Also prefer `emplace` over `operator[]` to avoid the default value insert then assign behavior.

Tested by loading multiple models over rpc-server with CUDA backend.

`./build/bin/test-backend-ops` passes locally with backend GB10 CUDA and CPU devices


rgerganov (approved):
Looks fine but not sure if it brings measurable performance improvement.

PR #18518, opened 2026-01-01
webui: fix code copy stripping XML/HTML tags

### LINKED ISSUE #18423: Misc. bug: Pasted content from <code> blocks loses XML-based tags (HTML, XML, JSX)
### Name and Version

version: 7442 (d0794e89d)

### Operating systems

Windows

llama-server

### Command line

```shell
llama-server -hf <model>
```

#### Summary
XML-based tags are stripped when copying and pasting a code block from the chat.
Reproduced in both Firefox and Chrome on Windows but this should be platform and browser agnostic.

The following HTML parse seems to be stripping the tags. There is already a fallback to the original text if the parse fails (line 69), but I wasn't sure the purpose of the HTML parse step to know if removing it entirely could be an option.
https://github.com/ggml-org/llama.cpp/blob/4fd59e8427e78241b01b1db352ef28be6bf304fd/tools/server/webui/src/lib/utils/clipboard.ts#L68

A smaller/shorter term solution could be to just skip the HTML parse step (i.e. do the fallback logic of using the actual raw code) when XML tags are detected in the string or when the code is in one of the more seriously affected languages (e.g. HTML, JSX, XML).

#### How to reproduce
1. New Chat (Any model and settings)
2. Ask to `Generate a small sample of XML`
3. Click "Copy code" on the resulting code block
4. Paste the code into another program

#### Actual behavior
Code pasted from the chat into the other program:
```text

    
        The Great Gatsby
        F. Scott Fitzgerald
        1925
        12.99
    


```

#### Expected behavior
Code pasted from the chat into the other program:
```text
<?xml version="1.0" encoding="UTF-8"?>
<library>
    <book id="1">
        <title>The Great Gatsby</title>
        <author>F. Scott Fitzgerald</author>
        <year>1925</year>
        <price currency="USD">12.99</price>
    </book>
</library>
```

### First Bad Commit

_No response_

### Relevant log output

```shell

```

Anri-Lombard:
## Summary
- Fixes code block copy stripping XML/HTML/JSX tags (e.g., `<book>`, `<div>`, `<Component />`)
- The `DOMParser` was treating code as HTML and stripping tags via `textContent`
- Since `rawCode` comes from `element.textContent` (which already decodes HTML entities), the parsing step was redundant

Fixes #18423


taronaeo (approved):

ServeurpersoCom:
Several of us have noticed/tested it, please make the static build and I'll merge it:)

ServeurpersoCom (approved):

Anri-Lombard:
> Several of us have noticed/tested it, please make the static build and I'll merge it:)

Done 🙏

thomasjfox:
Fix works great, thanks a lot in early 2026!

PR #18522, opened 2026-01-01
README.md Translation Automation into 20+ languages

solusops:
> [!NOTE]
> Underlying code has not been touched, this is simply a README translation workflow PR

I used llama.cpp python for constructing this pipeline by the way, thanks for your repo, otherwise inference on CPUs could not have been possible.

Why this PR: llama_cpp is used all around the world with different kind of audiences, I just wanted to contribute back to the communities whose product I have been using. And this is my way of contributing back.

- Automated translation means, any time you guys push or change main README, all languages auto-update.
- NavBar injection for easy navigation.
- Strict prompt compliance for 1:1 translation. Even stricter post-processing for hallucination prevention and garbage dump through regex filtering + length checking etc.
- it is completely running on Github Actions without external API calls except for model download.

> **Disclaimer**: I read the contributing guidelines and while I am 99% sure, it should not apply to this PR, I am still declaring the intent and the manner in which this PR was made.
> 
> **How this PR was made**: Using Automated Translation pipeline, running entirely on Github actions, no external APIs, no external call except to HuggingFace, and that too, only for downloading the Qwen3 model.
> I made the pipeline in 2 phases, first, I prototyped and did everything locally on my personal machine, since Runners are way too slow. Then I asked **Copilot** to refactor that into py_lint compliant file.
> I originally had raw dumping of README onto the model but well, it is a long story and realized I would actually need to make some effort, which led me to multiple efforts in different direction including trying to create a AST, but I cut the scope and this is the result.
> After a lot of work, I got this pipeline ready, I have loosened up its safety rails a bit in v2 release, if you were to look at the release history since my original validation and post processing were too strict. Thats about it, any questions, I am there for it.


CISC:
While certainly it must have been a fun project for you I don't think we can accept this, not only because the translations are deeply flawed, but because at the end of the day this is something offered by nearly all browsers nowadays (even 100% locally in some cases).

Simply put, having a single language `README.md` should no longer be a barrier for anyone.

solusops:
Well, thanks for going through the PR, it was fun, and well my thought was, why not contribute back to the repo its actually running on!
Since it adds 0 overhead or performance penalty due to being GitHub only, I wanted to make a PR out of it.

> not only because the translations are deeply flawed
oh yes, the intent is not replacing the human contributor but to provide them with a ready-made template to just edit things it

- the reason for "flawed" output is that the repo I used was extremely general, a custom ggml-specific repo can be forked with very minor prompt engineering and post-processing tweaks to ensure it aligns with your standards and does exactly what needs to be done


> Simply put, having a single language `README.md` should no longer be a barrier for anyone.

- True, but like, 90% of them are "flawed" like you said and do not really speak/show original intention of the devs due to generalization

But again, absolutely grateful for the work this community is doing, you guys rock!

ngxson:
Maintainers are busy enough with reviewing code-related PRs, I don't think we have the human resource to review translations, and "blindly" merge or accept a translation is not something we do here in llama.cpp

Even if one day, translations are needed, I would personally prefer it to be written predominantly by human, not an AI.

> Since it adds 0 overhead or performance penalty due to being GitHub only

I bet you haven't looked at [the long waiting line of CI runs in llama.cpp](https://github.com/ggml-org/llama.cpp/actions). New workflows like this are definitely not free - they add overhead for maintainers who rely on CI to verify certain PRs.

solusops:
> "blindly" merge or accept a translation is not something we do here in llama.cpp

oh yes, I just wanted a feedback on the PR and yes, this code is absolutely unsuitable for being merged in the present state.
Thanks for taking the time to go through it, I will Close the PR.

> at [the long waiting line of CI runs in llama.cpp](https://github.com/ggml-org/llama.cpp/actions)

I actually ran all of them, and since translations would run only once, it would not have been a problem to verify them, but, you are right, the maintenance issue definitely exists.

But really, do not let me take your attention from the important things, thanks again for reviewing the PR.
Sorry for the disturbance.
Keep up the good work!

PR #12343, opened 2025-03-12
llama : Add Gemma 3 support (+ experimental vision capability)

ngxson:
## Model info

- Official model page: https://ai.google.dev/gemma/docs/core
- Pre-quantized GGUF: https://huggingface.co/collections/ggml-org/gemma-3-67d126315ac810df1ad9e913
- Available sizes: 1B, 4B, 12B, 27B

> [!IMPORTANT]
>
> This PR only covers the text inference. Vision tower will be ignored upon converting to GGUF.  
> PR for vision support: https://github.com/ggml-org/llama.cpp/pull/12344

## Demo running text-only model via llama-cli

Download the pre-quantized GGUF from the link above

Clone this project and build llama-cli:

```sh
cd llama.cpp
cmake -B build
cmake --build build -j --target llama-cli
```

Run it

```sh
./build/bin/llama-cli -m ./gemma-3-4b-it-Q4_K_M.gguf

> who are you
I'm Gemma, a large language model created by the Gemma team at Google DeepMind. I’m an open-weights model, which means I’m widely available for public use! 

I can take text and images as inputs and generate text-based outputs. 

You can learn more about me on the Gemma project page: [https://ai.google.dev/gemma](https://ai.google.dev/gemma)
```

---

Huge thanks to Google team and Hugging Face team for providing me invaluable support in this PR 🤗

PR #18526, opened 2026-01-01
model: fix SWA per-layer freq_base

ngxson:
I think the `if constexpr (iswa)` can be wrong here, so pushing this PR for discussion.

At least for gemma3, SWA layers are expected to use different freq_base than full attention layer. Using `if constexpr (iswa)` will make it a fixed value cross all layers.

Here is the original code of gemma3 (PR https://github.com/ggml-org/llama.cpp/pull/12343) for reference:

```cpp
        for (int il = 0; il < n_layer; ++il) {
            const bool is_sliding          = (il + 1) % sliding_window_pattern;
            const float freq_base_l        = is_sliding ? 10000.0f    : freq_base;
            const float freq_scale_l       = is_sliding ? 1.0f        : freq_scale;
            struct ggml_tensor * KQ_mask_l = is_sliding ? KQ_mask_swa : KQ_mask;
```

### DIFF
--- a/src/models/modern-bert.cpp
+++ b/src/models/modern-bert.cpp
@@ -26,6 +26,7 @@ llm_build_modern_bert<iswa>::llm_build_modern_bert(const llama_model & model, co
     for (int il = 0; il < n_layer; ++il) {
         float freq_base_l  = 0.0f;
 
+        // TODO @ngxson : do we expect to apply this globally, or dynamically per layer?
         if constexpr (iswa) {
             freq_base_l = model.get_rope_freq_base(cparams, il);
         } else {
--- a/src/models/gemma3.cpp
+++ b/src/models/gemma3.cpp
@@ -33,7 +33,7 @@ llm_build_gemma3<iswa>::llm_build_gemma3(const llama_model & model, const llm_gr
         float freq_base_l  = 0.0f;
         float freq_scale_l = 0.0f;
 
-        if constexpr (iswa) {
+        if (hparams.is_swa(il)) {
             freq_base_l  = model.get_rope_freq_base (cparams, il);
             freq_scale_l = model.get_rope_freq_scale(cparams, il);
         } else {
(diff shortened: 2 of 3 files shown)


src/models/modern-bert.cpp
```diff
@@ -26,6 +26,7 @@ llm_build_modern_bert<iswa>::llm_build_modern_bert(const llama_model & model, co
     for (int il = 0; il < n_layer; ++il) {
         float freq_base_l  = 0.0f;
 
+        // TODO @ngxson : do we expect to apply this globally, or dynamically per layer?
         if constexpr (iswa) {
```
ngxson, line 30: It's quite strange here because modern-bert uses `build_attn_inp_no_cache`, so `iswa` has no effects here. Probably a copy-pasting mistake?

CISC:
I don't understand, the `iswa` check is for the two different versions of the model, this is exactly what it's supposed to do, the `is_swa(il)` check makes no sense, here are the functions:
https://github.com/ggml-org/llama.cpp/blob/8957be3d34d4bcfd969aa47f54b733d7c8c24131/src/llama-model.cpp#L7258-L7264

src/models/modern-bert.cpp
CISC: It does, check out #15798

CISC: Though the templating was probably unnecessary.

ngxson:
@CISC good catch. But in such case, `iswa = false` will make `if constexpr (iswa)` to never call `get_rope_freq_base`, which leads to SWA layers to use the wrong freq_base.

I think the valid fix here is to always call `get_rope_freq_base` instead.

~~One thing which is puzzling to me is that this seems to be a valid bug, but no one reported any issues about it. I assume because everyone use `iswa = true` be default, so people have (yet) seen any problems~~

ngxson:
Hmm ok sorry for wrong alert. I got it now.

I didn't notice that gemma3 support both the SWA and non-SWA variants. It seems like there are some other models (I assume `Rnj-1`) that uses the same gemma3 arch but has no SWA layers.

CISC:
> Hmm ok sorry for wrong alert. I got it now.
> 
> I didn't notice that gemma3 support both the SWA and non-SWA variants. It seems like there are some other models (I assume `Rnj-1`) that uses the same gemma3 arch but has no SWA layers.

Yes, this was added for `rnj-1`.

PR #18528, opened 2026-01-01
llama-graph: fix UB warning from cross-attn mask loop indices

nabbi:
I observed this behavior when building [ollama](https://github.com/ollama/ollama) on Gentoo.
Confirmed llm_graph_input_attn_cross::set_input functions are aligned between these two projects so submitting the PR here.

GCC emits the following warning when building with optimizations:

  llama-graph.cpp:473:9: warning: iteration 2147483645 invokes undefined
  behavior [-Waggressive-loop-optimizations]

The warning is caused by using int loop induction variables against int64_t bounds (n_tokens, n_enc), which allows signed overflow in the induction variable and enables undefined behavior.

This change widens the loop counters to int64_t to match the bounds and removes an unreachable loop of the form:

  for (i = n_tokens; i < n_tokens; ++i)

which could never execute.

Thank you

### DIFF
--- a/src/llama-graph.cpp
+++ b/src/llama-graph.cpp
@@ -46,7 +46,7 @@ void llm_graph_input_pos::set_input(const llama_ubatch * ubatch) {
             // the 3 first dims are the same, and 4th dim is all 0
             std::vector<llama_pos> pos_data(n_tokens*n_pos_per_embd);
             // copy the first dimension
-            for (int i = 0; i < n_tokens; ++i) {
+            for (int64_t i = 0; i < n_tokens; ++i) {
                 pos_data[               i] = ubatch->pos[i];
                 pos_data[    n_tokens + i] = ubatch->pos[i];
                 pos_data[2 * n_tokens + i] = ubatch->pos[i];
@@ -75,7 +75,7 @@ void llm_graph_input_attn_temp::set_input(const llama_ubatch * ubatch) {
         GGML_ASSERT(n_attn_temp_floor_scale != 0);
 
         std::vector<float> attn_scale_data(n_tokens, 0.0f);
-        for (int i = 0; i < n_tokens; ++i) {
+        for (int64_t i = 0; i < n_tokens; ++i) {
             const float pos = ubatch->pos[i];
             attn_scale_data[i] = std::log(
                 std::floor((pos + f_attn_temp_offset) / n_attn_temp_floor_scale) + 1.0
@@ -96,8 +96,8 @@ void llm_graph_input_pos_bucket::set_input(const llama_ubatch * ubatch) {
         int32_t * data = (int32_t *) pos_bucket->data;
 
         for (int h = 0; h < 1; ++h) {
-            for (int j = 0; j < n_tokens; ++j) {
-                for (int i = 0; i < n_tokens; ++i) {
+            for (int64_t j = 0; j < n_tokens; ++j) {
+                for (int64_t i = 0; i < n_tokens; ++i) {
                     data[h*(n_tokens*n_tokens) + j*n_tokens + i] = llama_relative_position_bucket(ubatch->pos[i], ubatch->pos[j], hparams.n_rel_attn_bkts, true);
                 }
             }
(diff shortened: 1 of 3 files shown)


CISC:
It's not quite that simple, `n_tokens` is actually limited to `uint32_t`, ~and you made some other odd changes here, any particular reason?~ Nvm, didn't read OP closely enough.

src/llama-graph.cpp
```diff
@@ -453,9 +453,9 @@ void llm_graph_input_attn_cross::set_input(const llama_ubatch * ubatch) {
 
     float * data = (float *) cross_kq_mask->data;
 
-    for (int h = 0; h < 1; ++h) {
-        for (int i = 0; i < n_tokens; ++i) {
-            for (int j = 0; j < n_enc; ++j) {
+    for (int64_t h = 0; h < 1; ++h) {
```
CISC:
```suggestion
    for (int h = 0; h < 1; ++h) {
```

CISC (commented):
So, why not fix the others, like f.ex. `llm_graph_input_mem_hybrid::set_input` with the same issue?

nabbi:
Within that function, n_tokens is redefined as an int64_t constant.
I certainly could be mistaken on how this should behave, would you prefer this split out or have the other functions in the file reviewed too ?

I think that dead loop came from previous refactoring efforts where there was separate data and data_sma loops. Yet I got lost in the reasoning with all those edits :)

CISC:
> Within that function, n_tokens is redefined as an int64_t constant. I certainly could be mistaken on how this should behave, would you prefer this split out or have the other functions in the file reviewed too ?

Sure, I was just curious why you only fixed one of a multitude of functions that would produce this warning?

> I think that dead loop came from previous refactoring efforts where there was separate data and data_sma loops. Yet I got lost in the reasoning with all those edits :)

Probably.

src/llama-graph.cpp
CISC: I find this a very strange "loop" BTW. :)

nabbi:
Interestingly, that compilation warning only tripped on that dead loop line. And not the first loop.

So I was hyper focused on just this one function...
Yet I agree those int >> int64_t changes also need to be corrected elsewhere, to prevent runtime issues. Unless as you stated it really is uint32_t??

Good conversation  :)

CISC:
> So I was hyper focused on just this one function... Yet I agree those int >> int64_t changes also need to be corrected elsewhere, to prevent runtime issues. Unless as you stated it really is uint32_t??

Well, `uint32_t` will be larger than `int` too in the same scenarios where `int64_t` is.

src/llama-graph.cpp
nabbi: right :) it runs once

CISC: > right :) it runs once

The value is also used for multiplication, so I'm guessing the loop is kept for historical reasons.

nabbi:
more holistic approach.  targeted n_tokens and adjacent code for correctness. I hadn't analyzed the entire repo for other integer type mismatches

src/llama-graph.cpp
ggerganov: It emphasizes that we construct the mask for a single attention head - there was a comment about this, but it got removed at some point. It's ok to keep the loop as it is.

src/llama-graph.cpp
```diff
@@ -469,12 +469,6 @@ void llm_graph_input_attn_cross::set_input(const llama_ubatch * ubatch) {
                 data[h*(n_enc*n_tokens) + i*n_enc + j] = f;
             }
         }
-
-        for (int i = n_tokens; i < n_tokens; ++i) {
-            for (int j = 0; j < n_enc; ++j) {
-                data[h*(n_enc*n_tokens) + i*n_enc + j] = -INFINITY;
-            }
-        }
```
ggerganov, line 478: I think this is the only meaningful change.

The rest of the changes in the PR seem redundant? Or am I missing something?

CISC: Well, at least the `n_tokens` loops are "possible" overflows.

nabbi: I'm more than happy to shirk the PR down to just the compilation warning only of type mismatches need a more focused effort.

We are throwing -Wformat= against LLAMA_LOG_DEBUG for the loop counters for at least in print_mask; that that needs correcting yet. Those are bound by a limit of 20 where as most other loops are bound to whatever the upper limit of n_tokens is set as.

PR #18534, opened 2026-01-02
model: Added support for Maincoder-1B model

maincode-prabod:
Added support for [Maincoder-1B](https://huggingface.co/Maincode/Maincoder-1B)

fixes #18346


ngxson (commented):
The model seems like a normal `llama` arch to me. Probably enough to just add one single line to `convert_hf_to_gguf.py`

CISC:
> The model seems like a normal `llama` arch to me. Probably enough to just add one single line to `convert_hf_to_gguf.py`

Agreed, this is another arch rebrand, though I'd say Qwen given the tokenizer and chat template.

maincode-prabod:
Thanks for the review! I understand the concern about architecture similarity with LLAMA and Qwen. 

Let me clarify the key difference:
**MAINCODER applies QK normalization AFTER RoPE, not before**:

```python
# From modelling_maincoder.py (lines 200-206)
# RoPE first
query_states, key_states = apply_rotary_emb(query_states, key_states, ...)

# Then QK 
query_states = self.q_norm(query_states)  
normkey_states = self.k_norm(key_states)
```

This differs from:
- Qwen3: Applies QK norm → then RoPE
- Llama (with use_kq_norm): Uses unweighted RMS norm, no learned weights

The order matters mathematically, RoPE modifies the query/key vectors, so normalizing before vs after produces different results.

Additionally, the model has learned QK norm weights (q_norm.weight, k_norm.weight per layer), which are not present in Qwen2 and use a different application order than Qwen3.

If there's an existing architecture that matches this exact pattern (RoPE → learned QK norm), I'm happy to use that instead. I checked llama.cpp, qwen2.cpp, qwen3.cpp, and others, but none match this specific order.

convert_hf_to_gguf.py
```diff
@@ -6414,6 +6414,28 @@ def set_gguf_parameters(self):
...
 
+@ModelBase.register("MaincoderForCausalLM")
+class MaincoderModel(TextModel):
+    model_arch = gguf.MODEL_ARCH.MAINCODER
+
+    def set_vocab(self):
+        try:
+            self._set_vocab_sentencepiece()
+        except FileNotFoundError:
+            self._set_vocab_gpt2()
```
```suggestion
```
I'm fairly certain no-one's going to introduce an SPM vocab for this model.

convert_hf_to_gguf.py
```diff
@@ -6414,6 +6414,28 @@ def set_gguf_parameters(self):
...
+
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+
+        head_dim = self.hparams.get("head_dim")
+        if head_dim is not None:
+            self.gguf_writer.add_key_length(head_dim)
+            self.gguf_writer.add_value_length(head_dim)
+            self.gguf_writer.add_rope_dimension_count(head_dim)
```
```suggestion
        if (head_dim := self.hparams.get("head_dim")) is not None:
            self.gguf_writer.add_rope_dimension_count(head_dim)
```
The others are already set.

convert_hf_to_gguf.py
```diff
@@ -6414,6 +6414,28 @@ def set_gguf_parameters(self):
...
+
+        if (ff_size := self.hparams.get("intermediate_size_mlp")) is not None:
+            self.gguf_writer.add_feed_forward_length(ff_size)
```
```suggestion
```
My understanding is that `intermediate_size` vs `intermediate_size_mlp` is for MoEs, for dense models they are the same (and already set in this case).

src/models/maincoder.cpp
```diff
@@ -0,0 +1,120 @@
...
+    cur = build_lora_mm(model.output, cur);
+
+    cb(cur, "result_output", -1);
+    res->t_logits = cur;
+
+    ggml_build_forward_expand(gf, cur);
+}
+
+
+
```
CISC: Remove trailing whitespace.

CISC (approved):

convert_hf_to_gguf.py
maincode-prabod: Updated in the latest commit

convert_hf_to_gguf.py
maincode-prabod: Updated in the latest commit

convert_hf_to_gguf.py
maincode-prabod: yes. agree. updated to reflect that

src/models/maincoder.cpp
maincode-prabod: removed

src/models/maincoder.cpp
```diff
@@ -0,0 +1,117 @@
...
+    res->t_embd = cur;
+
+    // lm_head
+    cur = build_lora_mm(model.output, cur);
+
+    cb(cur, "result_output", -1);
+    res->t_logits = cur;
+
+    ggml_build_forward_expand(gf, cur);
+}
```
CISC: Too much, fix EOF newline. :)

maincode-prabod: haha. my bad.

convert_hf_to_gguf.py
```diff
@@ -6414,6 +6414,16 @@ def set_gguf_parameters(self):
         # required by llama.cpp, unused
         self.gguf_writer.add_head_count(0)
 
+@ModelBase.register("MaincoderForCausalLM")
+class MaincoderModel(TextModel):
+    model_arch = gguf.MODEL_ARCH.MAINCODER
+    
+    def set_gguf_parameters(self):
```
```suggestion

    def set_gguf_parameters(self):
```

convert_hf_to_gguf.py
```diff
@@ -6414,6 +6414,16 @@ def set_gguf_parameters(self):
         # required by llama.cpp, unused
         self.gguf_writer.add_head_count(0)
 
+@ModelBase.register("MaincoderForCausalLM")
```
CISC, line 6418:
```suggestion

@ModelBase.register("MaincoderForCausalLM")
```
One more. :)

ngxson:
A quick search show that `hunyuan-dense.cpp` also has weighted norm after rope. But I'm ok to add a new one as this case is rare.

In any cases, it's always better to leave comments in the code to avoid confusions for future contributors.

Just an extra question, what's the reason behind having norm after rope (vs. before rope?) If I understand correctly, most models apply norm before rope to avoid adding any distortions to the embedded positional information

maincode-prabod:
Thanks @ngxson. Yeah, it is similar, The catch is Hunyuan uses ROPE_TYPE_NEOX (interleaved dimension pairs) while Maincoder uses ROPE_TYPE_NORM (sequential dimension pairs),  different dimension ordering, so can't directly reuse it. Let me know if I'm missing something that would allow us to reuse it, though!

On norm-after-RoPE:
Good question. Normalizing after RoPE means we're scaling the combined semantic+positional representation together, which keeps attention scores more stable across different positions. We found this helped during RL fine-tuning where attention distributions are sensitive. Most models do norm-before-RoPE to preserve positional encoding exactly, but post-RoPE norm worked better for our specific use case.

pwilkin:
Also, can we please appreciate the new formulation of RoPE that I at least haven't seen in Transformers before?

```python
def apply_rotary_emb(
    xq: torch.Tensor,
    xk: torch.Tensor,
    freqs_cis: torch.Tensor,
) -> tuple[torch.Tensor, torch.Tensor]:
    """Apply rotary embeddings to query and key tensors."""
    xq_ = torch.view_as_complex(xq.float().reshape(*xq.shape[:-1], -1, 2))
    xk_ = torch.view_as_complex(xk.float().reshape(*xk.shape[:-1], -1, 2))

    # Broadcast freqs_cis
    freqs_cis = freqs_cis[:, :, None, :]

    xq_out = torch.view_as_real(xq_ * freqs_cis).flatten(3)
    xk_out = torch.view_as_real(xk_ * freqs_cis).flatten(3)

    return xq_out.type_as(xq), xk_out.type_as(xk)
```

It's amazing in how many ways you can express the same thing in Python!

maincode-prabod:
Thanks @CISC and @ngxson for the review and for helping get this over the line! Really appreciate the time, expect more PRs from our team soon!

@pwilkin I wish I could take credit for that RoPE implementation! I had the exact same reaction when I saw it. I actually adapted it from the Llama 4 implementation [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama4/modeling_llama4.py#L239)

Issue #17026, opened 2025-11-05
Feature Request: Configurable threshold for full GPU-offloaded prompt processing during CPU+GPU inference


DocShotgun:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

Llama.cpp is currently one of the best backends to run large MoE models due to the ability to intelligently split the load between the devices, keeping the routed experts on CPU and the always-active-on-every-token params on GPU. Additionally, it can stream the CPU-allocated weights over to the main GPU to completely delegate prompt processing to the GPU, which tends to have much more computing power.

However, depending on the specific model and hardware - most notably the size of the CPU-allocated weights and the PCIe bandwidth - there's a different break-even point, where for small enough prompts, it's not beneficial to fully offload prompt processing to the GPU because combined CPU+GPU prompt processing will finish before you can copy all of the CPU-allocated weights over PCIe. From my understanding, there's a current hardcoded cutoff of 32 tokens.

I believe @jukofyork mentioned having a custom hack to make the threshold configurable via env var. Perhaps it's worth exposing the threshold param officially?

### Motivation

To give a concrete example of when this would make a difference:

I have a 4.93bpw Deepseek quant split between CPU and GPU, with over 300gb of CPU-allocated weights and about 94gb of GPU-allocated weights+compute buffers+k/v cache.

If I send a request that requires processing of 100 new tokens of prompt, llama.cpp would still need to copy over 300gb of weights across PCIe to the GPU because 100 exceeds the hardcoded threshold of 32 tokens - which is a minimum of ~5 seconds at max theoretical PCIe 5.0 x16 speeds. Meanwhile, CPU+GPU prompt processing would finish ingesting 100 new tokens in under 5 seconds and lead to better TTFT in this scenario.

### Possible Implementation

The simplest solution would be reading an env var or a cli arg instead of the hardcoded value.

It would be interesting to try to determine the break-even point automatically, but this is likely too complex and would probably need to be empirically measured on each model+quant+hardware setup.

jukofyork:
[...]

PR #18535, opened 2026-01-02
ggml: add env var GGML_OP_OFFLOAD_MIN_BATCH

### LINKED ISSUE #18530: Feature Request: Add configurable op offload min batch size
### Prerequisites

### Feature Description

This is a revisiting of my previous issue https://github.com/ggml-org/llama.cpp/issues/17026 which fell stale.

I'd like to request the ability to configure the threshold at which host tensor operations are offloaded to device. This threshold is currently hardcoded to `32` separately in multiple backends, but the ideal value depends on your hardware/model configuration - affected by factors such as PCIe bandwidth, CPU compute, and how much of the model resides on the CPU in host buffers.

Here's the code in the CUDA backend:
https://github.com/ggml-org/llama.cpp/blob/ced765be44ce173c374f295b3c6f4175f8fd109b/ggml/src/ggml-cuda/ggml-cuda.cu#L4662-L4668
But a nearly-identical function also exists in CANN, Metal, SYCL, and Vulkan. We'd just need to replace `const int min_batch_size = 32` with a configurable parameter.

### Motivation

To give a concrete example of when this would make a difference:

I have a 4.93bpw Deepseek quant split between CPU and GPU, with over 300gb of CPU-allocated weights and about 94gb of GPU-allocated weights+compute buffers+k/v cache.

If I send a request that requires processing of 100 new tokens of prompt, llama.cpp would still need to copy over 300gb of weights across PCIe to the GPU because 100 exceeds the hardcoded threshold of 32 tokens - which is a minimum of ~5 seconds at max theoretical PCIe 5.0 x16 speeds. Meanwhile, CPU+GPU prompt processing would finish ingesting 100 new tokens in under 5 seconds and lead to better TTFT in this scenario.

### Possible Implementation

I propose adding a new CLI arg `--op-offload-min-batch` (name subject to change!) which takes an integer with a default of `32` to keep current behavior intact if not specified.

We could either:
1. Add this arg in addition to `--op-offload` / `--no-op-offload`
2. Remove `--op-offload` / `--no-op-offload` and instead have an `--op-offload-min-batch` of `-1` correspond to disabling op offload

In taking a look at this with the assistance of AI, it seems like the complicating factor here is that the backend (i.e. `ggml/src/ggml-cuda/ggml-cuda.cu`) doesn't have direct access to the params as passed in via CLI arg.

DocShotgun:
https://github.com/ggml-org/llama.cpp/issues/18530 / https://github.com/ggml-org/llama.cpp/issues/17026

Makes the `min_batch_size` for triggering op offload configurable via env var, defaulting to the prior hardcoded value of `32` if not specified to keep current behavior intact.

This is helpful when running large MoEs with a significant size of weights stored in host buffers on CPU, causing a bottleneck when op offloading with small batches that are still larger than the default `32`. The optimal value, or "break even point" here depends on characteristics of the hardware + model, and is best determined empirically (ref: https://github.com/ggml-org/llama.cpp/issues/17026#issuecomment-3492543006).

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -4660,7 +4660,7 @@ static int64_t get_op_batch_size(const ggml_tensor * op) {
 }
 
 static bool ggml_backend_cuda_device_offload_op(ggml_backend_dev_t dev, const ggml_tensor * op) {
-    const int min_batch_size = 32;
+    const int min_batch_size = getenv("GGML_OP_OFFLOAD_MIN_BATCH") ? atoi(getenv("GGML_OP_OFFLOAD_MIN_BATCH")) : 32;
 
     return get_op_batch_size(op) >= min_batch_size;
 
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -4107,6 +4107,7 @@ struct ggml_backend_cuda_device_context {
     std::string name;
     std::string description;
     std::string pci_bus_id;
+    int op_offload_min_batch_size;
 };
 
 static const char * ggml_backend_cuda_device_get_name(ggml_backend_dev_t dev) {
@@ -4660,11 +4661,9 @@ static int64_t get_op_batch_size(const ggml_tensor * op) {
 }
 
 static bool ggml_backend_cuda_device_offload_op(ggml_backend_dev_t dev, const ggml_tensor * op) {
-    const int min_batch_size = getenv("GGML_OP_OFFLOAD_MIN_BATCH") ? atoi(getenv("GGML_OP_OFFLOAD_MIN_BATCH")) : 32;
+    ggml_backend_cuda_device_context * dev_ctx = (ggml_backend_cuda_device_context *)dev->context;
 
-    return get_op_batch_size(op) >= min_batch_size;
-
-    GGML_UNUSED(dev);
+    return get_op_batch_size(op) >= dev_ctx->op_offload_min_batch_size;
 }
 
 static ggml_backend_event_t ggml_backend_cuda_device_event_new(ggml_backend_dev_t dev) {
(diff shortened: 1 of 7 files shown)


ggerganov (commented):
The env var should be read only once upon device initiali§ation and then queried from the device context.

DocShotgun:
Took a crack at this, let me know if you'd recommend doing anything differently.

AI-assisted with searching for the relevant code in the Metal backend and with debugging compile failures.

- For CUDA, CANN, SYCL, and Vulkan, added `op_offload_min_batch_size` to the device context struct. We read the env var once prior to the loop that creates the device context(s), and then assign this value to the context for each device.
- For Metal we instead add the field to the device props, which we can then fetch from the offload op check.
- `dev` is no longer flagged as unused in the backend offload op checks. In Metal, `op` was also previously flagged as unused.
- CANN had an issue where `ggml_backend_cann_offload_op` is declared before `ggml_backend_cann_device_context`. This didn't cause any problems before when device context was unused. I moved it down to roughly match the other backends.

I tested CUDA locally on a `Qwen3-Coder-30B-A3B-Instruct-IQ4_XS.gguf` on my 7950X + 4090 Windows machine with `-b 4096 -ub 4096` and `--cpu-moe` and it seems to work as expected:
- `GGML_OP_OFFLOAD_MIN_BATCH=50000` with 1532 tokens prompt, op offload not triggered -> PP 146.81 T/s
- `GGML_OP_OFFLOAD_MIN_BATCH=64` with 1532 tokens prompt, op offload triggered -> PP 2320.88 T/s
- No env var set with 1532 tokens prompt, defaults to 32, op offload triggered -> PP 2322.76 T/s

NeoZhangJianyu (commented):
Looks good to me for SYCL backend part.

0cc4m:
The Vulkan changes are fine.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -4660,11 +4661,9 @@ static int64_t get_op_batch_size(const ggml_tensor * op) {
 }
 
 static bool ggml_backend_cuda_device_offload_op(ggml_backend_dev_t dev, const ggml_tensor * op) {
-    const int min_batch_size = 32;
-
-    return get_op_batch_size(op) >= min_batch_size;
+    ggml_backend_cuda_device_context * dev_ctx = (ggml_backend_cuda_device_context *)dev->context;
```
```suggestion
    ggml_backend_cuda_device_context * dev_ctx = (ggml_backend_cuda_device_context *) dev->context;
```

am17an (approved):

am17an:
@ggerganov merge?

PR #18541, opened 2026-01-02
vocab : reduce debug logs about non-EOG control tokens

ggerganov:
- Reduce logs for unused control tokens not being marked as EOG
- Append control attribute instead of assigning

Some models have many unused tokens, which leads to a lot of debug logs. F.ex:

```bash
./bin/llama-server -hf ggml-org/embeddinggemma-300M-GGUF --embeddings -lv 5

...

0.00.960.046 D init_tokenizer: initializing tokenizer for type 1
0.00.966.704 D load: control token: 262141 '<unused6239>' is not marked as EOG
0.00.966.706 D load: control token: 262140 '<unused6238>' is not marked as EOG
0.00.966.706 D load: control token: 262137 '<unused6235>' is not marked as EOG
0.00.966.707 D load: control token: 262136 '<unused6234>' is not marked as EOG
0.00.966.707 D load: control token: 262135 '<unused6233>' is not marked as EOG
0.00.966.707 D load: control token: 262134 '<unused6232>' is not marked as EOG
...

```

### DIFF
--- a/src/llama-vocab.cpp
+++ b/src/llama-vocab.cpp
@@ -2203,6 +2203,8 @@ void llama_vocab::impl::load(llama_model_loader & ml, const LLM_KV & kv) {
         //       for now, we apply this workaround to find the tokens based on their text
 
         for (const auto & t : token_to_id) {
+            auto & attr = id_to_token[t.second].attr;
+
             // find EOT token: "<|eot_id|>", "<|im_end|>", "<end_of_turn>", etc.
             if (special_eot_id == LLAMA_TOKEN_NULL) {
                 if (false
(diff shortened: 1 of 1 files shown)


danbev (approved):

CISC:
Just an aside...

In #18500 it was discovered that f.ex. `jina-v2-en` has lots of `[unusedX]` tokens that are marked as `NORMAL`, but were treated as (but not set to) `CONTROL` due to the `[]` name.

`AutoTokenizer` does not treat them as control tokens either, so the new behavior after that PR is the correct one.

PR #15225, opened 2025-08-11
server: implement GLM-style MTP

F1LM1:
This is very much a draft/proof of concept I'm playing with, just one idea for an MTP implementation. Planning to test on GLM-4.5 because it's the only model out there that we've preserved NextN tensors for.

From what I can tell
- the three models with MTP implemented in vLLM right now are all "DeepseekV3-style,"
- they only have one MTP head, which predicts token at position n+2,
- the MTP layers take as input the output embedding from the last conventional layer and their own input embedding.

So implementation-wise it seems like
- we should try to reuse the existing speculative decode functionality (including nice stuff like main model KV cache management, various samplers, etc.),
- but a lot of the full draft model functionality is redundant/harmful, like context/cache management for the draft model, vocab matching,
- it probably makes sense to write a new function like `mtp_speculative_gen_draft` in speculative.cpp that is vastly simplified and branch into it in server.cpp when a slot has MTP (versus `common_speculative_gen_draft`).
- AFAICT it looks like the server.cpp loop currently alternates between conventional forward pass and draft, which in the MTP case will probably sabotage performance gains (since our max throughput is only 1.5 tok/pass assuming zero rejections, instead of 2 tok/pass). Let me know if this isn't the case!—but if it is, should probably avoid doing non-speculative decodes after the first response token.
- It doesn't make sense to have to manage a distinct `ctx_dft` in this case as well. It's a bit hacky but I was thinking we could just have `ctx_dft = ctx` and then have both normal and MTP passes write over the shared `ctx` logits. I think this minimizes required code changes elsewhere

___

This is my first time (1) working with ML stuff outside of python (2) attempting to contribute, so patience is appreciated :)

PR #18543, opened 2026-01-02
Add EXAONE MoE implementations

nuxlear:

Add EXAONE-MoE architecture for the [K-EXAONE model](https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B) released by LG AI Research.

This PR adds the modeling code of EXAONE-MoE (K-EXAONE), and its chat template with according parsers, which is available at the fork of the LG AI Research:
https://github.com/Aim-Highest/llama.cpp


pwilkin:
Do we really need a new chat format here? Seems like the standard tool calling format with JSON and `<tool_call>` tags, besides the optional markdown inside. Is that markdown a real thing?

src/llama-model.cpp
```diff
@@ -1909,6 +1909,33 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                     default: type = LLM_TYPE_UNKNOWN;
                 }
             } break;
+        case LLM_ARCH_EXAONE_MOE:
+            {
+                // TODO: implement
```
ngxson: do we have anything else to implement?

tools/server/server-common.cpp
```diff
@@ -1013,6 +1013,12 @@ json oaicompat_chat_params_parse(
 
     // Apply chat template to the list of messages
     auto chat_params = common_chat_templates_apply(opt.tmpls, inputs);
+    SRV_INF("enable_thinking(req/body)=%d kwarg=%s prompt_thinking=%d format=%s\nPROMPT:\n%s",
+        inputs.enable_thinking,
+        json_value(inputs.chat_template_kwargs, "enable_thinking", std::string("")).c_str(),
+        chat_params.thinking_forced_open,
+        common_chat_format_name(chat_params.format),
+        chat_params.prompt.c_str());
```
ngxson: remove this, or it will spam the server log

convert_hf_to_gguf.py
```diff
@@ -8486,6 +8480,80 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+        super().set_gguf_parameters()
+        # We check whether the layer is MoE or not by referencing MoE module dynamically, not by the layer index
+        self.gguf_writer.add_expert_count(self.hparams["num_experts"])
+        self.gguf_writer.add_expert_used_count(self.hparams["num_experts_per_tok"])
+        moe_intermediate_size = self.hparams["moe_intermediate_size"]
+        num_shared_experts = self.hparams["num_shared_experts"]
+        self.gguf_writer.add_expert_feed_forward_length(moe_intermediate_size)
+        self.gguf_writer.add_expert_shared_count(num_shared_experts)
+        self.gguf_writer.add_expert_shared_feed_forward_length(moe_intermediate_size * num_shared_experts)
+        self.gguf_writer.add_expert_group_count(self.hparams["n_group"])            # 확인 필요
```
ngxson: I would prefer code comments to be in English if possible

convert_hf_to_gguf.py
```diff
@@ -1230,15 +1230,9 @@ def get_vocab_base_pre(self, tokenizer) -> str:
...
-            res = "kormo"
-        if chkhsh == "9d70134b369a70e5735009b6de918f7581b5211f7c074d1f89f753aea8248af1":
-            # ref: https://huggingface.co/tencent/Youtu-LLM-2B
-            res = "youtu"
-        if chkhsh == "16389f0a1f51ee53e562ffd51c371dc508639ab0e4261502071836e50e223e91":
-            # ref: https://huggingface.co/upstage/Solar-Open-100B
-            res = "solar-open"
+        if chkhsh == "6c81ce329e0802883b22eabab0d3fa48357337ef1ecb45443828bf1f6254833f":
+            # ref: https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B
+            res = "exaone-moe"
```
CISC, line 1244: Bad merge.

convert_hf_to_gguf.py
```diff
@@ -8486,6 +8480,80 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+        self.gguf_writer.add_expert_feed_forward_length(moe_intermediate_size)
+        self.gguf_writer.add_expert_shared_count(num_shared_experts)
+        self.gguf_writer.add_expert_shared_feed_forward_length(moe_intermediate_size * num_shared_experts)
+        self.gguf_writer.add_expert_group_count(self.hparams["n_group"])            # 확인 필요
+        self.gguf_writer.add_expert_group_used_count(self.hparams["topk_group"])    # 확인 필요
+        self.gguf_writer.add_expert_weights_scale(self.hparams["routed_scaling_factor"])
+        self.gguf_writer.add_expert_gating_func(gguf.ExpertGatingFuncType.SIGMOID)
+        self.gguf_writer.add_expert_weights_norm(self.hparams["norm_topk_prob"])
+        
+        self.gguf_writer.add_rope_scaling_type(gguf.RopeScalingType.NONE)
```
```suggestion
        moe_intermediate_size = self.hparams["moe_intermediate_size"]
        num_shared_experts = self.hparams["num_shared_experts"]
        self.gguf_writer.add_expert_feed_forward_length(moe_intermediate_size)
        self.gguf_writer.add_expert_shared_count(num_shared_experts)
        self.gguf_writer.add_expert_shared_feed_forward_length(moe_intermediate_size * num_shared_experts)
        self.gguf_writer.add_expert_weights_scale(self.hparams["routed_scaling_factor"])
        self.gguf_writer.add_expert_weights_norm(self.hparams["norm_topk_prob"])
```
These are already set.

src/llama-vocab.cpp
```diff
@@ -461,6 +461,11 @@ struct llm_tokenizer_bpe : llm_tokenizer {
                     "[!\"#$%&'()*+,\\-./:;<=>?@\\[\\\\\\]^_`{|}~][A-Za-z]+|[^\\r\\n\\p{L}\\p{P}\\p{S}]?[\\p{L}\\p{M}]+| ?[\\p{P}\\p{S}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+",
                 };
                 break;
+            case LLAMA_VOCAB_PRE_TYPE_EXAONE_MOE:
+                regex_exprs = {
+                    "(?:'[sS]|'[tT]|'[rR][eE]|'[vV][eE]|'[mM]|'[lL][lL]|'[dD])|[^\\r\\n\\p{L}\\p{N}]?(?:\\p{L}\\p{M}*(?: \\p{L}\\p{M}*)*)+|\\p{N}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n/]?|\\s*[\\r\\n]|\\s+(?!\\S)|\\s+",
```
CISC, line 468: Add original as comment.

src/models/exaone-moe.cpp
```diff
@@ -0,0 +1,154 @@
...
+    res->t_embd = cur;
+
+    // lm_head
+    cur = build_lora_mm(model.output, cur);
+
+    cb(cur, "result_output", -1);
+    res->t_logits = cur;
+
+    ggml_build_forward_expand(gf, cur);
+}
```
CISC: Fix EOF newline.

convert_hf_to_gguf.py
```diff
@@ -8486,6 +8480,80 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+
+    _experts: list[dict[str, Tensor]] | None = None
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        if name.startswith("mtp."):
+            return [] # ignore MTP layers for now
+
```
CISC, line 8556:
```suggestion
```
Is/will there be MTP, or just copy-paste?

src/llama-model.cpp
```diff
@@ -1909,6 +1909,33 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+                hparams.swa_type = LLAMA_SWA_TYPE_STANDARD;
+                hparams.n_swa = 128;
+                hparams.set_swa_pattern(4);
+                hparams.n_layer_dense_lead = 1;
+
+                ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW,          hparams.n_swa, true);
```
CISC:
```suggestion
                hparams.n_layer_dense_lead = 1;
                hparams.rope_freq_base_train_swa  = hparams.rope_freq_base_train;
                hparams.rope_freq_scale_train_swa = hparams.rope_freq_scale_train;

                ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA,                hparams.rope_freq_base_train_swa, false);
                ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW,          hparams.n_swa, true);
```
Required after #18553

src/models/exaone-moe.cpp
```diff
@@ -0,0 +1,154 @@
...
+            if (is_local_layer) {
+                cur = build_attn(inp_attn_iswa,
+                        model.layers[il].wo, NULL,
+                        Qcur, Kcur, Vcur, nullptr, nullptr, nullptr, 1.0f / sqrtf(float(n_embd_head)), il);
+            } else {
+                // cur = build_attn(inp_attn_kv,
+                cur = build_attn(inp_attn_iswa,
+                        model.layers[il].wo, NULL,
+                        Qcur, Kcur, Vcur, nullptr, nullptr, nullptr, 1.0f / sqrtf(float(n_embd_head)), il);
+            }
```
CISC:
```suggestion
            cur = build_attn(inp_attn_iswa,
                    model.layers[il].wo, NULL,
                    Qcur, Kcur, Vcur, nullptr, nullptr, nullptr, 1.0f / sqrtf(float(n_embd_head)), il);
```
These are the same? Even if you added non-SWA support, you'd template `inp_attn`.

src/llama-model.cpp
nuxlear: Nope. My bad. I'll remove it

convert_hf_to_gguf.py
nuxlear: It means "need to check" and can be removed. I'll remove it

convert_hf_to_gguf.py
nuxlear: https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B/raw/main/model.safetensors.index.json
There are MTP weights on the HF model, so we need to ignore them.

convert_hf_to_gguf.py
nuxlear: My mistake. I'll fix it

src/models/exaone-moe.cpp
nuxlear: You mean non-SWA always need to use `inp_attn_kv`? 
If so, I think `exaone4.cpp` should be fixed because it just chooses the template only based on whether the model uses SWA, rather than whether a layer uses SWA. 

Anyway, I will update this section to select the template

ngxson: `inp_attn_iswa` is a proxy class that will select SWA or normal attention based on layer index, `il`

You should look into other models to see how it's used, rather than making assumptions based on its name.

convert_hf_to_gguf.py
CISC: Or you could include them, like in `glm4-moe` and `bailingmoe2`, that way #15225 would have more testing surface.

src/models/exaone-moe.cpp
nuxlear: I see. As Gemma also uses `inp_attn_iswa` only, I'll commit this suggestion.

nuxlear:
@pwilkin Could you clarify which part you’re referring to?

What I intended here was to combine the reasoning parser from `deepseek_v3_1` and the tool-calling format from `hermes_2_pro`, since K-EXAONE needs both. I could not find use this combination without additional implementation.

nuxlear:
@CISC I’m working on implementing the MTP logic (it seems to be preserved but not actually used yet, right?).
One difference from other models is that `exaone-moe` reuses the base model’s `token_embd` and `output` layers, instead of defining its own weights.

Is there a recommended or proper way to load and use the base model’s `token_embd` and `output tensors` within `load_tensors()`?
Thank you very much for all your help!

CISC:
> @CISC I’m working on implementing the MTP logic (it seems to be preserved but not actually used yet, right?).

Yes.

> One difference from other models is that `exaone-moe` reuses the base model’s `token_embd` and `output` layers, instead of defining its own weights.
> 
> Is there a recommended or proper way to load and use the base model’s `token_embd` and `output tensors` within `load_tensors()`?

~Same as here, but you obviously need to add a `bool` to check if you should load from `LLM_TENSOR_OUTPUT` and/or `LLM_TENSOR_TOKEN_EMBD`.~
https://github.com/ggml-org/llama.cpp/blob/8a5c08b0017274de15bf2d795323a77dfe49c601/src/llama-model.cpp#L5533-L5535

Edit: Nvm, just mark them as `TENSOR_NOT_REQUIRED`, that's what was done for GLM 4.6, the MTP code must handle it when they are missing.

pwilkin:
> @pwilkin Could you clarify which part you’re referring to?
> 
> What I intended here was to combine the reasoning parser from `deepseek_v3_1` and the tool-calling format from `hermes_2_pro`, since K-EXAONE needs both. I could not find use this combination without additional implementation.

Okay, it might be that you're right (I was sure there was another model with standard Hermes toolcalling + reasoning, but maybe not).

src/llama-model.cpp
```diff
@@ -1917,6 +1917,37 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+                ml.get_key(LLM_KV_EXPERT_GROUP_USED_COUNT,           hparams.n_group_used, false);
+                ml.get_key(LLM_KV_EXPERT_GATING_FUNC,                hparams.expert_gating_func, false);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,              hparams.expert_weights_scale, false);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_NORM,               hparams.expert_weights_norm, false);
+
+                ml.get_key(LLM_KV_NEXTN_PREDICT_LAYERS,              hparams.nextn_predict_layers, false);
+
+                switch (hparams.n_layer) {
+                    case 32: type = LLM_TYPE_30B_A3B; break;
+                    case 48: type = LLM_TYPE_235B_A22B; break;
```
CISC:
```suggestion
                    case 32: type = LLM_TYPE_30B_A3B; break;
                    case 48: type = LLM_TYPE_235B_A22B; break;
```
These must be updated to reflect the additional NextN layer.

src/llama-model.cpp
```diff
@@ -5483,6 +5514,86 @@ bool llama_model::load_tensors(llama_model_loader & ml) {
...
+                        if (hparams.nextn_predict_layers > 0 && static_cast<uint32_t>(i) >= n_layer - hparams.nextn_predict_layers) {
+                            layer.nextn.eh_proj          = create_tensor(tn(LLM_TENSOR_NEXTN_EH_PROJ, "weight", i), {2 * n_embd, n_embd}, flags);
+                            layer.nextn.enorm            = create_tensor(tn(LLM_TENSOR_NEXTN_ENORM,   "weight", i), {n_embd}, flags);
+                            layer.nextn.hnorm            = create_tensor(tn(LLM_TENSOR_NEXTN_HNORM,   "weight", i), {n_embd}, flags);
+
+                            layer.nextn.shared_head_norm = create_tensor(tn(LLM_TENSOR_NEXTN_SHARED_HEAD_NORM, "weight", i), {n_embd}, flags | TENSOR_NOT_REQUIRED);
+                            // EXAONE-MoE uses shared embedding and lm_head with base model's. It does not have these tensors separately.
+                            // TODO: FIXME
+                            layer.nextn.embed_tokens     = tok_embd;
+                            layer.nextn.shared_head_head = output;
```
CISC:
```suggestion
                            layer.nextn.embed_tokens     = create_tensor(tn(LLM_TENSOR_NEXTN_EMBED_TOKENS, "weight", i), { n_embd, n_vocab }, flags | TENSOR_NOT_REQUIRED);
                            layer.nextn.shared_head_head = create_tensor(tn(LLM_TENSOR_NEXTN_SHARED_HEAD_HEAD, "weight", i), { n_embd, n_vocab }, flags | TENSOR_NOT_REQUIRED);
```

nuxlear:
I tested your suggestion encountered the following errors:

1. when using
```
layer.nextn.embed_tokens     = create_tensor(tn(LLM_TENSOR_NEXTN_EMBED_TOKENS,     "weight"), {n_embd, n_vocab}, flags | TENSOR_NOT_REQUIRED);
layer.nextn.shared_head_head = create_tensor(tn(LLM_TENSOR_NEXTN_SHARED_HEAD_HEAD, "weight"), {n_embd, n_vocab}, flags | TENSOR_NOT_REQUIRED);
```
the error log is as below:
```
Loading model... //weka/jwhwang/llama.cpp/ggml/src/ggml-backend.cpp:194: GGML_ASSERT(buffer) failed
-/weka/jwhwang/llama.cpp/build/bin/libggml-base.so.0(+0x53aac)[0x751919802aac]
/weka/jwhwang/llama.cpp/build/bin/libggml-base.so.0(ggml_print_backtrace+0x281)[0x751919802d66]
/weka/jwhwang/llama.cpp/build/bin/libggml-base.so.0(ggml_abort+0x173)[0x751919802f60]
/weka/jwhwang/llama.cpp/build/bin/libggml-base.so.0(ggml_backend_buffer_get_type+0x44)[0x75191981e836]
/weka/jwhwang/llama.cpp/build/bin/libggml-base.so.0(ggml_backend_buffer_is_host+0x1c)[0x75191981e726]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(_ZN23llm_graph_input_out_ids9set_inputEPK12llama_ubatch+0x6f)[0x75191d54dd01]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(_ZN16llm_graph_result10set_inputsEPK12llama_ubatch+0x83)[0x75191d5504a9]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(_ZN13llama_context14process_ubatchERK12llama_ubatch14llm_graph_typeP22llama_memory_context_iR11ggml_status+0x2f2)[0x75191d4ea96e]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(_ZN13llama_context6decodeERK11llama_batch+0x80e)[0x75191d4ebf2a]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(llama_decode+0x20)[0x75191d4f305f]
./build/bin/llama-cli(+0x3936f0)[0x5f4dd28cf6f0]
./build/bin/llama-cli(+0x1ca925)[0x5f4dd2706925]
./build/bin/llama-cli(+0x1a44b6)[0x5f4dd26e04b6]
./build/bin/llama-cli(+0xa7753)[0x5f4dd25e3753]
/usr/lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)[0x7519192321ca]
/usr/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)[0x75191923228b]
./build/bin/llama-cli(+0xa7195)[0x5f4dd25e3195]
\Aborted (core dumped)
```

2. when using
```
layer.nextn.embed_tokens     = create_tensor(tn(LLM_TENSOR_TOKEN_EMBD, "weight"), {n_embd, n_vocab}, flags | TENSOR_NOT_REQUIRED);
layer.nextn.shared_head_head = create_tensor(tn(LLM_TENSOR_OUTPUT,     "weight"), {n_embd, n_vocab}, flags | TENSOR_NOT_REQUIRED);
```
the error log is as below:
```
Loading model... \llama_model_load: error loading model: done_getting_tensors: wrong number of tensors; expected 781, got 783
|llama_params_fit: encountered an error while trying to fit params to free device memory: failed to load model
-llama_model_load: error loading model: done_getting_tensors: wrong number of tensors; expected 781, got 783
common_init_from_params: failed to load model 'K-EXAONE-236B-A23B-BF16.gguf'
srv    load_model: failed to load model, 'K-EXAONE-236B-A23B-BF16.gguf'

Failed to load the model
```

Is there any point I'm missing?

pwilkin:
@nuxlear haven't checked the entire code, but you might've forgotten to add the default ops for those tensor types to `LLM_TENSOR_INFOS`? Then the scheduler doesn't know where to put them.

nuxlear:
@pwilkin Thank you for your attention.

It seems they are already added to `LLM_TENSOR_INFOS`, I think.
```
static const std::map<llm_tensor, llm_tensor_info> LLM_TENSOR_INFOS = {
    ...
    // NextN/MTP tensors are currently ignored (reserved for future MTP support)
    // These tensors only exist in the last layer(s) and are treated as output tensors
    {LLM_TENSOR_NEXTN_EH_PROJ,              {LLM_TENSOR_LAYER_OUTPUT, GGML_OP_MUL_MAT}},
    {LLM_TENSOR_NEXTN_EMBED_TOKENS,         {LLM_TENSOR_LAYER_OUTPUT, GGML_OP_GET_ROWS}},
    {LLM_TENSOR_NEXTN_ENORM,                {LLM_TENSOR_LAYER_OUTPUT, GGML_OP_GET_ROWS}},
    {LLM_TENSOR_NEXTN_HNORM,                {LLM_TENSOR_LAYER_OUTPUT, GGML_OP_MUL}},
    {LLM_TENSOR_NEXTN_SHARED_HEAD_HEAD,     {LLM_TENSOR_LAYER_OUTPUT, GGML_OP_MUL_MAT}},
    {LLM_TENSOR_NEXTN_SHARED_HEAD_NORM,     {LLM_TENSOR_LAYER_OUTPUT, GGML_OP_MUL}},
};
```

pwilkin:
Ah, sorry, I missed that you're double-tapping those tensors. In this case, you want to use `TENSOR_DUPLICATED`.

pwilkin:
Here's some sample snippet for a similar case from existing code:

```c++
                    tok_embd = create_tensor(tn(LLM_TENSOR_TOKEN_EMBD, "weight"), {n_embd, n_vocab}, 0);

                    // output
                    output_norm = create_tensor(tn(LLM_TENSOR_OUTPUT_NORM, "weight"), {n_embd}, 0);
                    output      = create_tensor(tn(LLM_TENSOR_OUTPUT,      "weight"), {n_embd, n_vocab}, TENSOR_NOT_REQUIRED);

                    // if output is NULL, init from the input tok embed
                    if (output == NULL) {
                        output = create_tensor(tn(LLM_TENSOR_TOKEN_EMBD, "weight"), {n_embd, n_vocab}, TENSOR_DUPLICATED);
                    }
```

nuxlear:
Although I use the code below, the GGML_ASSERT(buffer) error raised.
```
layer.nextn.embed_tokens     = create_tensor(tn(LLM_TENSOR_NEXTN_EMBED_TOKENS,     "weight", i), {n_embd, n_vocab}, flags | TENSOR_NOT_REQUIRED);
if (layer.nextn.embed_tokens == NULL) {
    layer.nextn.embed_tokens = create_tensor(tn(LLM_TENSOR_TOKEN_EMBD, "weight"), {n_embd, n_vocab}, TENSOR_DUPLICATED);
}
layer.nextn.shared_head_head = create_tensor(tn(LLM_TENSOR_NEXTN_SHARED_HEAD_HEAD, "weight", i), {n_embd, n_vocab}, flags | TENSOR_NOT_REQUIRED);
if (layer.nextn.shared_head_head == NULL) {
    layer.nextn.shared_head_head = create_tensor(tn(LLM_TENSOR_OUTPUT, "weight"), {n_embd, n_vocab}, TENSOR_DUPLICATED);
}
```

pwilkin:
@nuxlear I think you're setting `TENSOR_SKIP` on them from earlier:

```c++
if (hparams.nextn_predict_layers > 0 && static_cast<uint32_t>(i) >= n_layer - hparams.nextn_predict_layers) {
 // skip all tensors in the NextN layers
 flags |= TENSOR_SKIP;
}
```

nuxlear:
@pwilkin If I remove the condition block, the error changes like:
```
Loading model... |/weka/jwhwang/llama.cpp/src/llama-model.cpp:2638: input/output layer tensor blk.48.nextn.eh_proj.weight used with a layer number
//weka/jwhwang/llama.cpp/build/bin/libggml-base.so.0(+0x53aac)[0x75c0a0979aac]
/weka/jwhwang/llama.cpp/build/bin/libggml-base.so.0(ggml_print_backtrace+0x281)[0x75c0a0979d66]
/weka/jwhwang/llama.cpp/build/bin/libggml-base.so.0(ggml_abort+0x173)[0x75c0a0979f60]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(+0x540351)[0x75c0a4744351]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(_ZN11llama_model12load_tensorsER18llama_model_loader+0x3331c)[0x75c0a47780fe]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(+0x40841a)[0x75c0a460c41a]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(+0x409359)[0x75c0a460d359]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(llama_model_load_from_file+0xbc)[0x75c0a460d641]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(+0x402fa6)[0x75c0a4606fa6]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(+0x4048b0)[0x75c0a46088b0]
/weka/jwhwang/llama.cpp/build/bin/libllama.so.0(llama_params_fit+0x5d)[0x75c0a460be2b]
./build/bin/llama-cli(+0x39261a)[0x59d9fd97861a]
./build/bin/llama-cli(+0x392f5c)[0x59d9fd978f5c]
./build/bin/llama-cli(+0x1ca925)[0x59d9fd7b0925]
./build/bin/llama-cli(+0x1a44b6)[0x59d9fd78a4b6]
./build/bin/llama-cli(+0xa7753)[0x59d9fd68d753]
/usr/lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)[0x75c0a03a91ca]
/usr/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)[0x75c0a03a928b]
./build/bin/llama-cli(+0xa7195)[0x59d9fd68d195]
Aborted (core dumped)
```

So I set `TENSOR_SKIP` on `LLM_TENSOR_NEXTN_EH_PROJ`, `LLM_TENSOR_NEXTN_ENORM`, `LLM_TENSOR_NEXTN_HNORM`, and `LLM_TENSOR_NEXTN_SHARED_HEAD_NORM`, but the GGML_ASSERT(buffer) failed again.

CISC:
@pwilkin We don't want to duplicate them, see https://github.com/ggml-org/llama.cpp/pull/18543#issuecomment-3717806458

My suggestion is correct, something else is causing the buffer assert.

src/models/exaone-moe.cpp
```diff
@@ -0,0 +1,146 @@
...
+            cb(Qcur, "Qcur", il);
+            cb(Kcur, "Kcur", il);
+            cb(Vcur, "Vcur", il);
+
+            cur = build_attn(inp_attn_iswa,
+                model.layers[il].wo, NULL,
+                Qcur, Kcur, Vcur, nullptr, nullptr, nullptr, 1.0f / sqrtf(float(n_embd_head)), il);
+            cb(cur, "attn_out", il);
+        }
+        if (il == n_layer - 1 && inp_out_ids) {
```
CISC:
```suggestion
        if (il == n_transformer_layers - 1 && inp_out_ids) {
```
Here's your culprit.

src/llama-model.cpp
```diff
@@ -1917,6 +1917,38 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                     default: type = LLM_TYPE_UNKNOWN;
                 }
             } break;
+        case LLM_ARCH_EXAONE_MOE:
+            {
+                hparams.swa_type = LLAMA_SWA_TYPE_STANDARD;
+                hparams.n_swa = 128;
+                hparams.set_swa_pattern(4);
+                hparams.n_layer_dense_lead = 1;
```
CISC:
```suggestion
```

src/llama-model.cpp
```diff
@@ -1917,6 +1917,38 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+                ml.get_key(LLM_KV_EXPERT_USED_COUNT,                 hparams.n_expert_used);
+                ml.get_key(LLM_KV_EXPERT_SHARED_COUNT,               hparams.n_expert_shared, false);
+                ml.get_key(LLM_KV_EXPERT_FEED_FORWARD_LENGTH,        hparams.n_ff_exp);
+                ml.get_key(LLM_KV_EXPERT_SHARED_FEED_FORWARD_LENGTH, hparams.n_ff_shexp, false);
+                ml.get_key(LLM_KV_EXPERT_GROUP_COUNT,                hparams.n_expert_groups, false);
+                ml.get_key(LLM_KV_EXPERT_GROUP_USED_COUNT,           hparams.n_group_used, false);
+                ml.get_key(LLM_KV_EXPERT_GATING_FUNC,                hparams.expert_gating_func, false);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,              hparams.expert_weights_scale, false);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_NORM,               hparams.expert_weights_norm, false);
+
```
CISC, line 1942:
```suggestion
                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,         hparams.n_layer_dense_lead);
```

convert_hf_to_gguf.py
```diff
@@ -8497,6 +8500,88 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+        self.gguf_writer.add_expert_shared_count(num_shared_experts)
+        self.gguf_writer.add_expert_shared_feed_forward_length(moe_intermediate_size * num_shared_experts)
+        self.gguf_writer.add_expert_weights_scale(self.hparams["routed_scaling_factor"])
+        self.gguf_writer.add_expert_weights_norm(self.hparams["norm_topk_prob"])
+        # For here, we hard-code the number of NextN/MTP layers to 1 for K-EXAONE,
+        # so that we can convert MTP weights to GGUF format for speculative decoding.
+        # This is because HF config of K-EXAONE does not have `num_nextn_predict_layers` at now.
+        # Will be updated when HF config is updated.
+        self.gguf_writer.add_nextn_predict_layers(self.hparams.get("num_nextn_predict_layers", 1))
+
```
```suggestion
        self.gguf_writer.add_leading_dense_block_count(self.hparams["first_last_k_dense_replace"])
```

nuxlear:
It works! Thank you very much for all your help. @CISC @pwilkin

convert_hf_to_gguf.py
```diff
@@ -8497,6 +8500,90 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+    model_arch = gguf.MODEL_ARCH.EXAONE_MOE
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.block_count = self.hparams["num_hidden_layers"] + self.hparams.get("num_nextn_predict_layers", 0)
+        self.tensor_map = gguf.get_tensor_name_map(self.model_arch, self.block_count)
+
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+        # We check whether the layer is MoE or not by referencing MoE module dynamically, not by the layer index
```
```suggestion
```

gguf-py/gguf/tensor_mapping.py
```diff
@@ -1702,6 +1703,7 @@ class TensorNameMap:
         # NextN/MTP tensors for GLM4_MOE
         MODEL_TENSOR.NEXTN_EH_PROJ: (
             "model.layers.{bid}.eh_proj",
+            "mtp.fc", # exaone-moe
```
CISC: Is the `bid` for these actually being added correctly? Please check!

convert_hf_to_gguf.py
```diff
@@ -8497,6 +8500,90 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+        self.gguf_writer.add_expert_weights_norm(self.hparams["norm_topk_prob"])
+        n_dense_layer = self.hparams.get("first_k_dense_replace", self.hparams.get("first_last_k_dense_replace", 0))
+        self.gguf_writer.add_leading_dense_block_count(n_dense_layer)
+        # For here, we hard-code the number of NextN/MTP layers to 1 for K-EXAONE,
+        # so that we can convert MTP weights to GGUF format for speculative decoding.
+        # This is because HF config of K-EXAONE does not have `num_nextn_predict_layers` at now.
+        # Will be updated when HF config is updated.
+        self.gguf_writer.add_nextn_predict_layers(self.hparams.get("num_nextn_predict_layers", 1))
+        
+        self.gguf_writer.add_rope_scaling_type(gguf.RopeScalingType.NONE)
```
CISC, line 8530:
```suggestion
        self.gguf_writer.add_nextn_predict_layers(self.hparams.get("num_nextn_predict_layers", 1))

        self.gguf_writer.add_rope_scaling_type(gguf.RopeScalingType.NONE)
```

convert_hf_to_gguf.py
```diff
@@ -8497,6 +8500,90 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        if name.startswith("mtp."):
+            if name.find("layers.") != -1:
+                # assume it is `mtp.layers.0.[module_name]` format
+                name = name.replace(f"mtp.layers.{bid}", f"model.layers.{bid + self.hparams['num_hidden_layers']}")
+
+        if name.endswith("e_score_correction_bias"):
+            name = name.replace("e_score_correction_bias", "e_score_correction.bias")
+        
+        if name.find("mlp.experts") != -1:
```
CISC, line 8560:
```suggestion
            name = name.replace("e_score_correction_bias", "e_score_correction.bias")

        if name.find("mlp.experts") != -1:
```

CISC (approved):
Otherwise LGTM as long as @pwilkin checks off on chat format.

gguf-py/gguf/tensor_mapping.py
nuxlear: Yes, we have a shared `e_norm` (`pre_fc_norm_embedding`), `h_norm` (`pre_fc_norm_hidden`), and `eh_proj` (`fc`), not seperated per layer.

CISC: Yes, but are they added as `blk.48.` in the GGUF?

nuxlear: You're right. I could not find to keep MTP module apart from the base model layers, so I added its handling logic to `convert_hf_to_gguf.py`

convert_hf_to_gguf.py
nuxlear: I found that we don't use `first_last_k_dense_replace` at all, so I changed it into `first_k_dense_replace` like other models in transformers. We will update the `config.json` of the original model as well.

Anyway, this section doesn't have to be fixed 😀

CISC: Everyone else seem to be using `num_dense_layers`.

nuxlear: LFM2, AFMoE seem to use it, and others like Deepseek, GLM4-MoE, Mistral seem to use `first_k_dense_replace`, so it seems to be a choice.

CISC: Yeah, there seems to be sport in seeing how many ways you can define the same thing in `config.json`. :)

gguf-py/gguf/tensor_mapping.py
CISC: So, still working on it? Only seeing logic for `mtp.layers.0.`, not these.

pwilkin (approved):
Template's fine, I'm treating it as temporary anyway :)

gguf-py/gguf/tensor_mapping.py
nuxlear: What I intended was that all shared weights are copied into each `model.layers.{bid}.[module_name]`.

In the case of K-EXAONE, there is only one MTP layer (`mtp.layers.0`), but I think it should still work correctly even with more MTP layers. If not, please let me know which part needs to be adjusted. 

```
        if name.startswith("mtp."):
            if name.find("layers.") != -1:
                # assume it is `mtp.layers.0.[module_name]` format
                name = name.replace(f"mtp.layers.{bid}", f"model.layers.{bid + self.hparams['num_hidden_layers']}")
```

CISC: That's fine for mtp `ffn` etc, but `mtp.fc/pre_fc/norm` will not be mapped correctly I think, you need to remap them similarly.

ddh0:
LGTM? 🙏

gguf-py/gguf/tensor_mapping.py
nuxlear: I removed those tensor mappings and updated `modify_tensors` to remap and propagate tensors to the corresponsing `blk.{bid}.XXX`.

CISC:
```suggestion
```

gguf-py/gguf/tensor_mapping.py
```diff
@@ -1710,10 +1712,12 @@ class TensorNameMap:
 
         MODEL_TENSOR.NEXTN_ENORM: (
             "model.layers.{bid}.enorm",
+            "mtp.pre_fc_norm_embedding", # exaone-moe
```
CISC:
```suggestion
```

gguf-py/gguf/tensor_mapping.py
```diff
@@ -1710,10 +1712,12 @@ class TensorNameMap:
...
         ),
 
         MODEL_TENSOR.NEXTN_HNORM: (
             "model.layers.{bid}.hnorm",
+            "mtp.pre_fc_norm_hidden", # exaone-moe
```
CISC:
```suggestion
```

gguf-py/gguf/tensor_mapping.py
```diff
@@ -1722,6 +1726,7 @@ class TensorNameMap:
 
         MODEL_TENSOR.NEXTN_SHARED_HEAD_NORM: (
             "model.layers.{bid}.shared_head.norm",
+            "mtp.norm", # exaone-moe
```
CISC:
```suggestion
```

src/models/exaone-moe.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+
+llm_build_exaone_moe::llm_build_exaone_moe(const llama_model & model, const llm_graph_params & params) :
+    llm_graph_context(params) {
+    const int64_t n_embd_head = hparams.n_embd_head_k;
+
+    GGML_ASSERT(n_embd_head == hparams.n_embd_head_v);
+    GGML_ASSERT(n_embd_head == hparams.n_rot);
+    
+    ggml_tensor * cur;
```
CISC:
```suggestion

    ggml_tensor * cur;
```

src/models/exaone-moe.cpp
```diff
@@ -0,0 +1,146 @@
...
+        ggml_tensor * inpSA = inpL;
+
+        // use RoPE for SWA layers
+        const bool is_local_layer = hparams.is_swa(il);
+
+        // norm
+        cur = build_norm(inpL, model.layers[il].attn_norm, NULL, LLM_NORM_RMS, il);
+        cb(cur, "attn_norm", il);
+    
+        // self-attention
```
CISC:
```suggestion

        // self-attention
```

gguf-py/gguf/tensor_mapping.py
```diff
@@ -1699,7 +1700,7 @@ class TensorNameMap:
             "conformer.layers.{bid}.norm_conv", # lfm2
         ),
 
-        # NextN/MTP tensors for GLM4_MOE
+        # NextN/MTP tensors for GLM4_MOE and exaone-moe
```
CISC:
```suggestion
        # NextN/MTP tensors
```

CISC:
Ok, I think we can merge after CI, can someone do a last verification that conversion and inference is working in the meantime?

nicoboss:
> Ok, I think we can merge after CI, can someone do a last verification that conversion and inference is working in the meantime?

Convert unfortunately fails with the following error on latest master:
```
venv/bin/python convert_hf_to_gguf.py /bpool/K-EXAONE-236B-A23B --outfile=/mradermacher/tmp/quant/K-EXAONE-236B-A23B
(...)
INFO:hf-to-gguf:blk.47.ffn_up_shexp.weight,           torch.bfloat16 --> BF16, shape = {6144, 2048}
INFO:hf-to-gguf:blk.47.ffn_norm.weight,               torch.bfloat16 --> F32, shape = {6144}
INFO:hf-to-gguf:output_norm.weight,                   torch.bfloat16 --> F32, shape = {6144}
Traceback (most recent call last):
  File "/apool/llama.cpp/convert_hf_to_gguf.py", line 11437, in <module>
    main()
  File "/apool/llama.cpp/convert_hf_to_gguf.py", line 11431, in main
    model_instance.write()
  File "/apool/llama.cpp/convert_hf_to_gguf.py", line 684, in write
    self.prepare_tensors()
  File "/apool/llama.cpp/convert_hf_to_gguf.py", line 8846, in prepare_tensors
    super().prepare_tensors()
  File "/apool/llama.cpp/convert_hf_to_gguf.py", line 555, in prepare_tensors
    for new_name, data_torch in (self.modify_tensors(data_torch, name, bid)):
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/apool/llama.cpp/convert_hf_to_gguf.py", line 8843, in modify_tensors
    return [(self.map_tensor_name(name), data_torch)]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/apool/llama.cpp/convert_hf_to_gguf.py", line 510, in map_tensor_name
    raise ValueError(f"Can not map tensor {name!r}")
ValueError: Can not map tensor 'model.layers.48.input_layernorm.weight'
```

CISC:
> ```
>   File "/apool/llama.cpp/convert_hf_to_gguf.py", line 510, in map_tensor_name
>     raise ValueError(f"Can not map tensor {name!r}")
> ValueError: Can not map tensor 'model.layers.48.input_layernorm.weight'
> ```

Hmmm, that sounds strange, it should not be a problem...

nuxlear:
The problem is that the HF checkpoint (https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B) includes MTP weights, while `config.json` does not have `num_nextn_predict_layer`, so the block count remains unchanged, which leads to a mapping error.

Two options are available:
1. simple add `"num_nextn_predict_layers": 1` to your HF config.json, or
2. update the default value of `num_nextn_predict_layers` to 1 in `convert_hf_to_gguf.py`

I think option 2 is better.

CISC:
> The problem is that the HF checkpoint (https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B) includes MTP weights, while `config.json` does not have `num_nextn_predict_layer`, so the block count remains unchanged, which leads to a mapping error.
> 
> Two options are available:
> 
> 1. simple add `"num_nextn_predict_layers": 1` to your HF config.json, or
> 2. update the default value of `num_nextn_predict_layers` to 1 in `convert_hf_to_gguf.py`
> 
> I think option 2 is better.

Or, you know, update your config: :)
https://github.com/ggml-org/llama.cpp/blob/60591f01d433f3fc7603d5273fbe361bd05a3507/convert_hf_to_gguf.py#L8777-L8778

nuxlear:
Yes, there is a third option, however, the mismatch between
https://github.com/ggml-org/llama.cpp/blob/60591f01d433f3fc7603d5273fbe361bd05a3507/convert_hf_to_gguf.py#L8760

and
https://github.com/ggml-org/llama.cpp/blob/60591f01d433f3fc7603d5273fbe361bd05a3507/convert_hf_to_gguf.py#L8775-L8779

still remains, so one of them should be fixed.

Regardless of this, we will add `"num_nextn_predict_layers": 1` to our config anyway :)

CISC:
> Yes, there is a third option, however, the mismatch still remains, so one of them should be fixed.

Yes, definitely, please submit a PR, I suggest setting default to 0 as the safer solution.

nicoboss:
After adding `"num_nextn_predict_layers": 1` to `config.json` convert worked without any issues but I can't run the converted model. I tried booth using the CUDA and the CPU backend from latest master but got the following error:
```
root@AI:/apool/llama.cpp/build/bin# ./llama-cli -m /mradermacher/tmp/quant/K-EXAONE-236B-A23B.gguf -v
(...)
llama_kv_cache: layer  46: dev = CPU
llama_kv_cache: layer  47: filtered
llama_kv_cache: layer  48: dev = CPU
llama_kv_cache:        CPU KV buffer size =     0.00 MiB
llama_kv_cache: size =  111.00 MiB (   768 cells,  37 layers,  1/1 seqs), K (f16):   55.50 MiB, V (f16):   55.50 MiB
llama_context: enumerating backends
llama_context: backend_ptrs.size() = 1
llama_context: max_nodes = 6128
llama_context: reserving full memory module
llama_context: worst-case: n_tokens = 512, n_seqs = 1, n_outputs = 1
graph_reserve: reserving a graph for ubatch with n_tokens =    1, n_seqs =  1, n_outputs =    1
/apool/llama.cpp/src/llama-graph.cpp:1052: fatal error
[New LWP 654520]
[New LWP 654521]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
0x000073888a262c17 in __GI___wait4 (pid=654522, stat_loc=0x0, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
30      ../sysdeps/unix/sysv/linux/wait4.c: No such file or directory.
#0  0x000073888a262c17 in __GI___wait4 (pid=654522, stat_loc=0x0, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
30      in ../sysdeps/unix/sysv/linux/wait4.c
#1  0x000073888a6a36db in ggml_print_backtrace () from libggml-base.so.0
#2  0x000073888a6a382e in ggml_abort () from libggml-base.so.0
#3  0x000073888a81b935 in llm_graph_context::build_moe_ffn(ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, long, long, llm_ffn_op_type, bool, bool, float, llama_expert_gating_func_type, int, ggml_tensor*) const () from libllama.so.0
#4  0x000073888a81c494 in llm_graph_context::build_moe_ffn(ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, ggml_tensor*, long, long, llm_ffn_op_type, bool, bool, float, llama_expert_gating_func_type, int, ggml_tensor*) const () from libllama.so.0
#5  0x000073888a90fb2b in llm_build_exaone_moe::llm_build_exaone_moe(llama_model const&, llm_graph_params const&) () from libllama.so.0
#6  0x000073888a857472 in llama_model::build_graph(llm_graph_params const&) const () from libllama.so.0
#7  0x000073888a7e8e2b in llama_context::graph_reserve(unsigned int, unsigned int, unsigned int, llama_memory_context_i const*, bool, unsigned long*) () from libllama.so.0
#8  0x000073888a7eeed9 in llama_context::llama_context(llama_model const&, llama_context_params) () from libllama.so.0
#9  0x000073888a7ef47d in llama_init_from_model () from libllama.so.0
#10 0x000073888a7c243a in llama_get_device_memory_data(char const*, llama_model_params const*, llama_context_params const*, std::vector<ggml_backend_device*, std::allocator<ggml_backend_device*> >&, unsigned int&, unsigned int&, unsigned int&, ggml_log_level) () from libllama.so.0
#11 0x000073888a7c2f47 in llama_params_fit_impl(char const*, llama_model_params*, llama_context_params*, float*, llama_model_tensor_buft_override*, unsigned long*, unsigned int, ggml_log_level) () from libllama.so.0
#12 0x000073888a7c636e in llama_params_fit () from libllama.so.0
#13 0x00005c3d4d289826 in common_init_result::common_init_result(common_params&) ()
#14 0x00005c3d4d28b059 in common_init_from_params(common_params&) ()
#15 0x00005c3d4d174c75 in server_context_impl::load_model(common_params const&) ()
#16 0x00005c3d4d0c55e6 in main ()
/[Inferior 1 (process 654517) detached]
Aborted
```

CISC:
> After adding `"num_nextn_predict_layers": 1` to `config.json` convert worked without any issues but I can't run the converted model. I tried booth using the CUDA and the CPU backend from latest master but got the following error:
> 
> ```
> /apool/llama.cpp/src/llama-graph.cpp:1052: fatal error
> ```

OK, my bad, I didn't notice that `scoring_func` was removed from the config (why?):
https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B/commit/27c3a76f02c76138b854cae2d14850eadb4c86db#d2h-025836

And I forgot to ensure that this wasn't `false`:
https://github.com/ggml-org/llama.cpp/blob/60591f01d433f3fc7603d5273fbe361bd05a3507/src/llama-model.cpp#L1954

@nuxlear Can you please restore `scoring_func` in the config again?

CISC:
@nicoboss Merged fix, requires updated `config.json` with `scoring_func`.

nicoboss:
Thank you all so much for quickly fixing this issue. I can confirm that after getting latest master, latest model and adding `"scoring_func": "sigmoid"` to the `config.json` it successfully converted and gives great responses and is incredibly fast for its size. The only thing I noticed is that while loading verbose the model the following warning about unused tensors appear but it doesn't seem to matter:
```
create_tensor: loading tensor blk.47.ffn_up_shexp.weight
model has unused tensor blk.48.attn_q.weight (size = 41287680 bytes) -- ignoring
model has unused tensor blk.48.attn_k.weight (size = 5160960 bytes) -- ignoring
model has unused tensor blk.48.attn_v.weight (size = 5160960 bytes) -- ignoring
model has unused tensor blk.48.attn_output.weight (size = 41287680 bytes) -- ignoring
model has unused tensor blk.48.attn_norm.weight (size = 24576 bytes) -- ignoring
model has unused tensor blk.48.attn_q_norm.weight (size = 512 bytes) -- ignoring
model has unused tensor blk.48.attn_k_norm.weight (size = 512 bytes) -- ignoring
model has unused tensor blk.48.ffn_norm.weight (size = 24576 bytes) -- ignoring
model has unused tensor blk.48.ffn_gate.weight (size = 92897280 bytes) -- ignoring
model has unused tensor blk.48.ffn_down.weight (size = 92897280 bytes) -- ignoring
model has unused tensor blk.48.ffn_up.weight (size = 92897280 bytes) -- ignoring
model has unused tensor blk.48.nextn.eh_proj.weight (size = 61931520 bytes) -- ignoring
model has unused tensor blk.48.nextn.enorm.weight (size = 24576 bytes) -- ignoring
model has unused tensor blk.48.nextn.hnorm.weight (size = 24576 bytes) -- ignoring
model has unused tensor blk.48.nextn.shared_head_norm.weight (size = 24576 bytes) -- ignoring
load_tensors: tensor 'token_embd.weight' (q6_K) (and 0 others) cannot be used with preferred buffer type CUDA_Host, using CPU instead
```

CISC:
> The only thing I noticed is that while loading verbose the model the following warning about unused tensors appear but it doesn't seem to matter:

This is normal.

nuxlear:
@CISC @nicoboss Sorry for the delay. We've added `"scoring_func": "sigmoid"` to our `config.json`.

PR #18547, opened 2026-01-02
context : reserve new scheduler when graph topology changes

ggerganov:
cont #17617

In some cases we know that a graph reallocation would be necessary (see https://github.com/ggml-org/llama.cpp/pull/17617). Re-reserve the scheduler to reduce the amount of unexpected graph reallocations and to prevent further reallocations later.

Also, the number of active samplers when using backend sampling with `llama-server` is now properly configured. Before, for a server with N slots, we were always running N samplers, regardless of how many slots are active. Now, thanks to the new reserve logic, we disable the samplers for the inactive slots.

TODOs:

- [x] Handle backend sampler changes

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -1182,7 +1182,7 @@ private:
         SLT_DBG(slot, "launching slot : %s\n", safe_json_to_str(slot.to_json()).c_str());
 
         // initialize samplers
-        {
+        if (task.uses_sampling()) {
             slot.smpl.reset(common_sampler_init(model, task.params.sampling));
 
             if (slot.smpl == nullptr) {
@@ -1211,6 +1211,8 @@ private:
             }
 
             SLT_INF(slot, "sampler chain: %s\n", common_sampler_print(slot.smpl.get()).c_str());
+        } else {
+            slot.smpl.reset();
         }
 
         // initialize draft batch
@@ -2593,6 +2595,12 @@ private:
             llama_set_embeddings(ctx, slot_batched->need_embd());
         }
 
+        for (auto & slot : slots) {
+            if (!slot.is_processing() || !slot.smpl) {
+                llama_set_sampler(ctx, slot.id, nullptr);
+            }
+        }
+
         if (batch.n_tokens == 0) {
             SRV_WRN("%s", "no tokens to decode\n");
         }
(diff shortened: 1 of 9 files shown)


danbev (approved):

ggerganov:
@ngxson PTAL at the server changes when you get the chance. They are relatively minor.

tools/server/server-task.h
```diff
@@ -156,6 +156,11 @@ struct server_task {
         return tokens.size();
     }
 
+    bool uses_sampling() const {
+        return type != SERVER_TASK_TYPE_EMBEDDING &&
+               type != SERVER_TASK_TYPE_RERANK;
```
ngxson: I would prefer having the reversed logic here: `type == SERVER_TASK_TYPE_COMPLETION || type == SERVER_TASK_TYPE_INFILL`

Also note that `SERVER_TASK_TYPE_INFILL` will be removed soon, because it's technically just a completion task with a special chat template

tools/server/server-context.cpp
```diff
@@ -2593,6 +2595,12 @@ struct server_context_impl {
             llama_set_embeddings(ctx, slot_batched->need_embd());
         }
 
+        for (auto & slot : slots) {
+            if (!slot.is_processing() || !slot.smpl) {
+                llama_set_sampler(ctx, slot.id, nullptr);
+            }
+        }
```
ngxson: should this be moved to `slot.release()`? If I understand it correctly, this means we set the sampler to nullptr if the slot is not processing anything

ggerganov: Yes, good idea.

> If I understand it correctly, this means we set the sampler to nullptr if the slot is not processing anything

Yes, this prevents from the `llama_context` adding dummy sampling nodes to the graph.

tools/server/server-task.h
ggerganov: Addressed here: [ffa0d15](https://github.com/ggml-org/llama.cpp/pull/18547/commits/ffa0d15e869958d92500e777050ad27c56af9f03)

tools/server/server-context.cpp
ggerganov: I moved it to `server_slot.reset()` in [d9146ed](https://github.com/ggml-org/llama.cpp/pull/18547/commits/d9146ed2922592ce6c09f278f6e49de7b5c07ab1)

Also a bit of refactoring:

- Rename `server_slot.clear()` -> `server_slot.prompt_clear()`
- Remove redundant `slot.reset()` from `launch_slot_with_task()`. Assumption is that all slots will be reset when they are released, so no need to do it again upon launch

ggerganov:
Should be good to merge. @ngxson Let me know if you want to take one more look

ngxson (approved):

PR #18550, opened 2026-01-02
ggml : add ggml_build_forward_select

ggerganov:
target #18547
alt #18549

- Add `GGML_TENSOR_FLAG_COMPUTE` flag indicating that a tensor in the graph must be computed
- Add new `ggml_build_forward_select()` call:

```c
    GGML_API struct ggml_tensor * ggml_build_forward_select(
            struct ggml_cgraph  * cgraph,
            struct ggml_tensor ** tensors,
            int                   n_tensors,
            int                   idx);
```

All provided tensors are built forward into the graph. Only `tensors[idx]` and it's ancestry are marked for computing via the new flag value.

This new logic allows us to construct graphs that compute different things, but at the same time have the same topology. This is needed to avoid unwanted graph reallocations (#17617).

TODOs:
- [x] Check flag for CUDA graph updates (https://github.com/ggml-org/llama.cpp/pull/18550#issuecomment-3706922324)
- [x] Update fusion logic? (https://github.com/ggml-org/llama.cpp/pull/18550#issuecomment-3706157703)
- ~Enable `-DGGML_SCHED_NO_REALLOC=ON` for server CI~ (next PR)
- [x] Fix vulkan command submission logic https://github.com/ggml-org/llama.cpp/pull/18550#discussion_r2661496552

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -13719,6 +13719,10 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
             submit_node_idx = i;
         }
 
+        if ((cgraph->nodes[i]->flags & GGML_TENSOR_FLAG_COMPUTE) == 0) {
+            continue;
+        }
+
         if (cgraph->nodes[i]->op == GGML_OP_MUL_MAT || cgraph->nodes[i]->op == GGML_OP_MUL_MAT_ID) {
             auto bytes = ggml_nbytes(cgraph->nodes[i]->src[0]);
             mul_mat_bytes += bytes;
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -12191,6 +12191,9 @@ static bool ggml_vk_build_graph(ggml_backend_vk_context * ctx, ggml_cgraph * cgr
     if (ggml_is_empty(node) || ggml_op_is_empty(node->op) || !node->buffer) {
         return false;
     }
+    if ((node->flags & GGML_TENSOR_FLAG_COMPUTE) == 0) {
+        return false;
+    }
 
     VK_LOG_DEBUG("ggml_vk_build_graph(" << node << ", " << ggml_op_name(node->op) << ")");
     ctx->semaphore_idx = 0;
@@ -13645,7 +13648,7 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
     int last_node = cgraph->n_nodes - 1;
 
     // If the last op in the cgraph isn't backend GPU, the command buffer doesn't get closed properly
-    while (last_node > 0 && ggml_vk_is_empty(cgraph->nodes[last_node])) {
+    while (last_node > 0 && (ggml_vk_is_empty(cgraph->nodes[last_node]) || ((cgraph->nodes[last_node]->flags & GGML_TENSOR_FLAG_COMPUTE) == 0))) {
         last_node -= 1;
     }
 
@@ -13717,10 +13720,6 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
     for (int i = 0; i < cgraph->n_nodes; i++) {
         if (first_node_in_batch) {
             submit_node_idx = i;
-        }
-
-        if ((cgraph->nodes[i]->flags & GGML_TENSOR_FLAG_COMPUTE) == 0) {
-            continue;
         }
 
         if (cgraph->nodes[i]->op == GGML_OP_MUL_MAT || cgraph->nodes[i]->op == GGML_OP_MUL_MAT_ID) {
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -630,10 +630,11 @@ extern "C" {
 
     // this tensor...
     enum ggml_tensor_flag {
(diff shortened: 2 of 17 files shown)


jeffbolznv:
Just want to make sure I understand how this is used - it would still be two separate graphs, they'd just be able to reuse allocations (i.e. ggml-alloc would decide they match)?

I think ggml_can_fuse and ggml_can_fuse_subgroup would need to be updated to make sure all nodes are computed. And any backend-specific fusion logic.

ggerganov:
> Just want to make sure I understand how this is used - it would still be two separate graphs, they'd just be able to reuse allocations (i.e. ggml-alloc would decide they match)?

Yes, for example the graph when the input is token ids (`batch.token != null`) and the graph when the input is directly embedding vectors (`batch.embd != null`) are still different, but with this extra logic the scheduler will not need to reallocate them because all nodes remain the same. It's just a different subset of the nodes being marked for computing.

> I think ggml_can_fuse and ggml_can_fuse_subgroup would need to be updated to make sure all nodes are computed. And any backend-specific fusion logic.

Not yet sure that it's really necessary to do so - at least I can't think of a fail case so far. Note that the `GGML_TENSOR_FLAG_COMPUTE` flag is controlled only through `ggml_build_forward_select()`.

max-krasnyansky (approved):
Looks good to me.

am17an:
We would need to check how this behaves with CUDA graphs, since inherently the computation is changing

taronaeo:
cc: @AlekseiNikiforovIBM @Andreas-Krebbel

Give us a week or so to check on this :)

am17an:
For CUDA graphs I think adding a check for flags in `ggml_graph_node_has_matching_properties` should be enough. This would trigger an update to the graph

AlekseiNikiforovIBM:
> cc: @AlekseiNikiforovIBM @Andreas-Krebbel
> 
> Give us a week or so to check on this :)

LGTM

taronaeo (approved):
Ack for IBM zDNN backend :)

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -13509,6 +13509,10 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
             submit_node_idx = i;
         }
 
+        if ((cgraph->nodes[i]->flags & GGML_TENSOR_FLAG_COMPUTE) == 0) {
+            continue;
```
0cc4m: If the last node or nodes are not flagged, the loop would end without the final command submission. This would need some way to ensure a final submit if `submitted_nodes > 0`.

reeselevine (approved):
WebGPU update looks good to me, we always do a final submission if commands > 0 so there shouldn't be a problem like the comment about the Vulkan backend above.

ggerganov:
> I think ggml_can_fuse and ggml_can_fuse_subgroup would need to be updated to make sure all nodes are computed. And any backend-specific fusion logic.

@jeffbolznv This should be enough I think: [3646af9](https://github.com/ggml-org/llama.cpp/pull/18550/commits/3646af9e382b63a96eeca741b7fc113d506149c3). Or maybe I am missing something regarding "And any backend-specific fusion logic."

@am17an CUDA graphs check updated in [ced0693](https://github.com/ggml-org/llama.cpp/pull/18550/commits/ced06931900861ea3b6e4332c0f8ce055e0f5f96)

ggml/src/ggml-vulkan/ggml-vulkan.cpp
ggerganov: Moved the check over here: [eafbd13](https://github.com/ggml-org/llama.cpp/pull/18550/commits/eafbd13790de0f45a7557904536d575571bacc43)

Hm, just noticed this is maybe not enough. Will take some extra look.

jeffbolznv: Maybe update this logic:

```
    // If the last op in the cgraph isn't backend GPU, the command buffer doesn't get closed properly
    while (last_node > 0 && ggml_vk_is_empty(cgraph->nodes[last_node])) {
        last_node -= 1;
    }
```

jeffbolznv:
> > I think ggml_can_fuse and ggml_can_fuse_subgroup would need to be updated to make sure all nodes are computed. And any backend-specific fusion logic.
> 
> @jeffbolznv This should be enough I think: [3646af9](https://github.com/ggml-org/llama.cpp/pull/18550/commits/3646af9e382b63a96eeca741b7fc113d506149c3). Or maybe I am missing something regarding "And any backend-specific fusion logic."

ggml_can_fuse_subgraph_ext needs the same check. All vulkan fusions go through either ggml_can_fuse or ggml_can_fuse_subgraph, so we shouldn't need anything else for ggml-vulkan's fusion.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
ggerganov: Updated like this: [9696c3e](https://github.com/ggml-org/llama.cpp/pull/18550/commits/9696c3e74807363a918f9ceeaa86a31b9ec68ca5)

ggerganov: Waiting for one final ack on the Vulkan changes and will proceed to merge.

0cc4m (commented):
The Vulkan changes are fine now, thank you.

0cc4m (approved):

ggml/src/ggml-vulkan/ggml-vulkan.cpp
0cc4m: Looks good now.

PR #18551, opened 2026-01-02
server : add thinking content blocks to Anthropic Messages API

686f6c61:
Hey team!

Following up on my initial fix for thinking content blocks, I found an additional issue while testing with different reasoning models.

### What was happening

The original fix added thinking content blocks, but there were two problems:

**Missing signature field** - The Anthropic API requires a `signature` field in thinking blocks (non-streaming) and a `signature_delta` event before closing thinking blocks (streaming). Without this, some clients like Claude Code were failing to parse responses correctly.

**Duplicate content_block_start events** - In streaming mode, each token was generating a new `content_block_start` event instead of just one per content block. This was because the `thinking_block_started` and `text_block_started` flags were local variables that reset on each chunk.

### The fix

For the signature issue:
- Added empty `signature` field in non-streaming thinking blocks
- Added `signature_delta` event before `content_block_stop` in streaming

For the duplicate events, I moved the block tracking state to `task_result_state` which persists across streaming chunks. This ensures we only emit one `content_block_start` per content block.

### Testing

Tested with multiple reasoning models to ensure compatibility:

- DeepSeek-R1-Distill-Qwen-7B: non-streaming and streaming working
- Qwen3-4B-Thinking-2507: non-streaming and streaming working

All 27 Anthropic compatibility tests pass, including the new signature verification.

@hksdpc255 I couldn't test with MiniMax-M2 due to hardware limitations (the smallest quantization is 64GB). Could you give it a try with this update? If it still doesn't work, please share the raw API response (a curl to `/v1/messages` with `stream: true`) so I can see what's happening.

Thanks for reporting the Claude Code compatibility issue and helping debug!

### DIFF
--- a/tools/server/server-task.h
+++ b/tools/server/server-task.h
@@ -337,6 +337,9 @@ struct server_task_result_cmpl_partial : server_task_result {
     std::vector<common_chat_msg_diff> oaicompat_msg_diffs; // to be populated by update()
     bool is_updated = false;
 
+    // for Anthropic API: track if any reasoning content has been generated
+    bool anthropic_has_reasoning = false;
+
     virtual bool is_stop() override {
         return false; // in stream mode, partial responses are not considered stop
     }
@@ -346,6 +349,8 @@ struct server_task_result_cmpl_partial : server_task_result {
     virtual void update(task_result_state & state) override {
         is_updated = true;
         state.update_chat_msg(content, true, oaicompat_msg_diffs);
+        // track if the accumulated message has any reasoning content
+        anthropic_has_reasoning = !state.chat_msg.reasoning_content.empty();
     }
 
     json to_json_non_oaicompat();
--- a/tools/server/server-task.h
+++ b/tools/server/server-task.h
@@ -343,8 +343,9 @@ struct server_task_result_cmpl_partial : server_task_result {
 
     // for Anthropic API: track if any reasoning content has been generated
     bool anthropic_has_reasoning = false;
-    // pointer to task_result_state for Anthropic streaming state tracking
-    task_result_state * anthropic_state = nullptr;
+    // Streaming state copied from task_result_state for this chunk
+    bool anthropic_thinking_block_started = false;
+    bool anthropic_text_block_started = false;
 
     virtual bool is_stop() override {
         return false; // in stream mode, partial responses are not considered stop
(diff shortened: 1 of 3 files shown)


hksdpc255:
Why there is `has_thinking` in `to_json_anthropic_stream` and `anthropic_has_reasoning` in `to_json_anthropic`?

hksdpc255:
Tested. Not compatible with claude-code: 
pic1
pic2
pic3
pic4

686f6c61:
> Why there is `has_thinking` in `to_json_anthropic_stream` and `anthropic_has_reasoning` in `to_json_anthropic`?

Hey, great question!

The naming difference reflects their different scopes and lifetimes:

has_thinking in to_json_anthropic_stream():

This is a local variable
The function is called once at the end of streaming, when the complete oaicompat_msg is already fully accumulated
At that point we can simply check !oaicompat_msg.reasoning_content.empty() and we're done
anthropic_has_reasoning in server_task_result_cmpl_partial::to_json_anthropic():

This is a member variable that persists across calls
The function is called many times during streaming (once per chunk)
Here's the problem: when chunk N arrives with a text delta, we need to know if there was any reasoning content in previous chunks (1 to N-1) to correctly compute text_block_index
Local variables reset on each call, so we'd lose track of whether reasoning was seen before
The member variable is updated in update() by checking the accumulated message state
So the different names actually reflect their different nature - one is ephemeral (computed fresh each time), the other is persistent (tracked across the stream).

That said, I'm open to renaming them for consistency if you think it would improve readability! Let me know what you prefer.

hksdpc255:
I can confirm it works for claude-code now.

tools/server/server-task.h
```diff
@@ -337,6 +341,11 @@ struct server_task_result_cmpl_partial : server_task_result {
     std::vector<common_chat_msg_diff> oaicompat_msg_diffs; // to be populated by update()
     bool is_updated = false;
 
+    // for Anthropic API: track if any reasoning content has been generated
+    bool anthropic_has_reasoning = false;
+    // pointer to task_result_state for Anthropic streaming state tracking
+    task_result_state * anthropic_state = nullptr;
```
ngxson: keeping a raw pointer like this is a bad pattern. you must copy anything you need here

the naming is also misleading. why `task_result_state` must be anthropic? it contains a whole lot more data than just anthropic

ngxson (changes requested):

686f6c61:
Hi @ngxson,

You're right on both points. The raw pointer was initially a quick way to access persistent state from `to_json_anthropic()` (which doesn't receive the state object directly), but it's indeed a bad pattern and the naming was confusing.

Fixed in the new commit:
- Removed the pointer entirely
- Added `anthropic_thinking_block_started` and `anthropic_text_block_started` directly to `server_task_result_cmpl_partial`
- In `update()`: copy state before processing, then pre-compute updates for next chunk
- In `to_json_anthropic()`: use local copies instead of dereferencing

This way all state mutations happen in `update()` (which has access to `task_result_state`), and `to_json_anthropic()` just reads the copied values. Cleaner data flow.

All unit tests pass, and I've tested streaming/non-streaming with Qwen3-4B-Thinking and DeepSeek-R1-Distill models - both thinking and text blocks are emitted correctly.

Thanks for the review!

ngxson (approved):

PR #18328, opened 2025-12-23
model: support MiMo-V2-Flash

ngxson:
Ref HF model: https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash

I'm interested in this model because they are the second one to use attention sink (after GPT-OSS)

Fix https://github.com/ggml-org/llama.cpp/issues/18120

Test using Q8_0 model (tested up to ~4K tokens):

```
$ llama-cli -m ../models/MiMo-V2-Flash/modelq.gguf -c 8000 -p "hi"

Loading model...  


▄▄ ▄▄
██ ██
██ ██  ▀▀█▄ ███▄███▄  ▀▀█▄    ▄████ ████▄ ████▄
██ ██ ▄█▀██ ██ ██ ██ ▄█▀██    ██    ██ ██ ██ ██
██ ██ ▀█▄██ ██ ██ ██ ▀█▄██ ██ ▀████ ████▀ ████▀
                                    ██    ██
                                    ▀▀    ▀▀

build      : b7534-d4a3c4d41
model      : modelq.gguf
modalities : text

available commands:
  /exit or Ctrl+C     stop or exit
  /regen              regenerate the last response
  /clear              clear the chat history
  /read               add a text file


> hi

[Start thinking]
We are going to create a simple "hi" response in multiple languages.
 The user just said "hi", so we will respond with "hi" in a few different languages.
 We'll create a list of common languages and their translation for "hi".
 Then we will output each one in a formatted way.
[End thinking]

Hello! Here's a friendly "hello" in several languages:

🌍 **Common Greetings**:
- **English**: Hi  
- **Spanish**: Hola  
- **French**: Bonjour  
- **German**: Hallo  
- **Italian**: Ciao  
- **Japanese**: こんにちは (Konnichiwa)  
- **Korean**: 안녕하세요 (Annyeonghaseyo)  
- **Arabic**: مرحبًا (Marhaban)  
- **Hindi**: नमस्ते (Namaste)  
- **Portuguese**: Olá  

**Fun fact**: The earliest recorded use of "hello" in English dates back to 1827!  😊  

How can I assist you today?

[ Prompt: 4431.1 t/s | Generation: 31.3 t/s ]

```

PR #18553, opened 2026-01-02
llama : refactor rope_freq_base/scale_swa conversion and init

CISC:
Make 10000/1.0 the SWA default. Since we now have a metadata key for `rope_freq_base` let's add it universally when present in config and start using it.

Added TODOs for future potential metadata when `rope_parameters` start getting used fully.

### DIFF
--- a/src/llama-hparams.h
+++ b/src/llama-hparams.h
@@ -105,9 +105,9 @@ struct llama_hparams {
 
     float    rope_attn_factor = 1.0f;
     float    rope_freq_base_train;
-    float    rope_freq_base_train_swa;
+    float    rope_freq_base_train_swa  = 10000.0f;
     float    rope_freq_scale_train;
-    float    rope_freq_scale_train_swa;
+    float    rope_freq_scale_train_swa = 1.0f;
 
     uint32_t n_ctx_orig_yarn;
     float    rope_yarn_log_mul = 0.0f;
--- a/src/llama-model.cpp
+++ b/src/llama-model.cpp
@@ -673,6 +673,10 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                     hparams.f_attn_temp_scale       = 0.1f;
                     hparams.f_attn_temp_offset      = 1.0f;
                     hparams.set_swa_pattern(4);   // pattern: 3 chunked - 1 full
+
+                    hparams.rope_freq_base_train_swa  = hparams.rope_freq_base_train;
+                    hparams.rope_freq_scale_train_swa = hparams.rope_freq_scale_train;
+                    ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA, hparams.rope_freq_base_train_swa, false);
                 }
 
                 switch (hparams.n_expert) {
@@ -718,6 +722,10 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                 if (hparams.n_swa > 0) {
                     hparams.swa_type = LLAMA_SWA_TYPE_STANDARD;
                     hparams.set_swa_pattern(4);
+
+                    hparams.rope_freq_base_train_swa  = hparams.rope_freq_base_train;
+                    hparams.rope_freq_scale_train_swa = hparams.rope_freq_scale_train;
+                    ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA, hparams.rope_freq_base_train_swa, false);
                 } else {
                     hparams.swa_type = LLAMA_SWA_TYPE_NONE;
                 }
@@ -1304,7 +1312,10 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                 hparams.n_swa = 4096; // default value of gemma 2
                 hparams.set_swa_pattern(2);
                 hparams.attn_soft_cap = true;
+                hparams.rope_freq_base_train_swa  = hparams.rope_freq_base_train;
+                hparams.rope_freq_scale_train_swa = hparams.rope_freq_scale_train;
 
+                ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA,          hparams.rope_freq_base_train_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW,    hparams.n_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS, hparams.f_norm_rms_eps);
                 ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING,      hparams.f_attn_logit_softcapping, false);
@@ -1517,7 +1528,10 @@ void llama_model::load_hparams(llama_model_loader & ml) {
             {
                 hparams.swa_type = LLAMA_SWA_TYPE_STANDARD;
                 hparams.set_swa_pattern(4);
+                hparams.rope_freq_base_train_swa  = hparams.rope_freq_base_train;
+                hparams.rope_freq_scale_train_swa = hparams.rope_freq_scale_train;
 
+                ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA,       hparams.rope_freq_base_train_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW, hparams.n_swa);
                 ml.get_key(LLM_KV_LOGIT_SCALE,              hparams.f_logit_scale);
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_EPS,  hparams.f_norm_eps);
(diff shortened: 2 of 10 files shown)


CISC:
`partial_rotary_factor` seems to be the new `rope_parameters` scaling key, we could add metadata for that (at least for SWA), though currently it seems all models are either same as full attention or no scaling.

ngxson (approved):

CISC:
@ngxson @ggerganov Any thoughts on keeping it like this or going back to defaulting to being the same as full attention? Technically the models that depend on the latter do not support anything else, so should be updated to use `get_rope_freq_base/scale`.

Also adding `add_rope_freq_scale_swa` (for `rope_parameters["sliding_attention"]["partial_rotary_factor"]`) for consistency?

ggerganov:
> @ngxson @ggerganov Any thoughts on keeping it like this or going back to defaulting to being the same as full attention? Technically the models that depend on the latter do not support anything else, so should be updated to use get_rope_freq_base/scale.

The proposal is OK, but indeed it would only benefit if all relevant models are adjusted to call `get_rope_freq_base/scale`.

> partial_rotary_factor seems to be the new rope_parameters scaling key, we could add metadata for that (at least for SWA), though currently it seems all models are either same as full attention or no scaling.

Hm that's confusing -  `partial_rotrary_factor` used to mean what part of the head is being rotated. We used it to determine `n_rot`.

CISC:
> > @ngxson @ggerganov Any thoughts on keeping it like this or going back to defaulting to being the same as full attention? Technically the models that depend on the latter do not support anything else, so should be updated to use get_rope_freq_base/scale.
> 
> The proposal is OK, but indeed it would only benefit if all relevant models are adjusted to call `get_rope_freq_base/scale`.

Yeah, I think that would make sense to do.

> > partial_rotary_factor seems to be the new rope_parameters scaling key, we could add metadata for that (at least for SWA), though currently it seems all models are either same as full attention or no scaling.
> 
> Hm that's confusing - `partial_rotrary_factor` used to mean what part of the head is being rotated. We used it to determine `n_rot`.

They are inverses, no? 
https://github.com/ggml-org/llama.cpp/blob/fb1aa17d863e2cd62d7d9cfa4bafda1e54364e81/src/llama-model.cpp#L581-L587

https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/src/transformers/modeling_rope_utils.py#L302-L303
```
            *   partial_rotary_factor (`float`, *optional*, defaults to 1.0): If less than 1.0, inverse frequencies
                will be returned for the first fraction of the head_dim.
```
https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/src/transformers/modeling_rope_utils.py#L578-L579
```
        partial_rotary_factor (`float`, *optional*):
            The percentage of the query and key head embedding on which RoPE will be applied.
```

ngxson:
`partial_rotrary_factor` means we only rotate first `head_dim * partial_rotrary_factor` dimensions of the vector, so upon conversion to GGUF, we calculate `n_rot = head_dim * partial_rotrary_factor`

The documentation that you linked is correct, note that `inverse frequencies` means the array of per-dim frequency to calculate sin/cos grid in transformers.

> If less than 1.0, inverse frequencies will be returned for the first fraction of the head_dim.

That means for example, if we have `head_dim = 16` and `partial_rotary_factor = 0.25`, we return `inv_freq` for only 16*0.25 = 4 dims. This effectively means we only have 2 `inv_freq` values (because each `inv_freq` value will be used to calculate sin+cos)

CISC:
Ok, confusing indeed, it seemed like the two were directly related, so I guess there's no metadata we can use for `add_rope_freq_scale_swa`?

ngxson:
Also note that `ropescale` / `freq_scale` refers to a different concept in llama.cpp. The naming was a bit messy because of historical reasons, but in a nutshell (please correct if I'm wrong @ggerganov):

This was added to simply adjust the `freq_base` during the era of long-context extension was not yet a standard. ~~IIRC it was controlled via a CLI option for research purpose~~, for example scale the freq_base from `10000` to `20000`, `40000`, etc

It seems to me that it's no longer used in practice, as YaRN is now the standard. For YaRN, the model is simply fine-tuned with a very high base freq, like `500000`, no other adjustments are needed.

ngxson:
To correct my last comment, `freq_scale` is still used by some models (YaRN RoPE kernel does use it for some calculations), and is saved to `add_rope_scaling_factor`, so it is not a CLI-only option as I said earlier

@CISC yes we may need a new metadata like `add_rope_scaling_factor_swa`, but I don't think any models are using it right now because most models simply disable YaRN for sliding layers (hence, most models use the default `10000` base freq for SWA layers)

CISC:
> @CISC yes we may need a new metadata like `add_rope_scaling_factor_swa`, but I don't think any models are using it right now because most models simply disable YaRN for sliding layers (hence, most models use the default `10000` base freq for SWA layers)

Yeah, though no discernable scaling factor metadata AFAICT. May change when models start using `rope_parameters` with `full_attention`/`sliding_attention` populated?

ngxson:
Yes, I think it's safer to see if models actually use it, then implement it later.

Also one thing might be related here, while working on https://github.com/ggml-org/llama.cpp/pull/18328 , I realized that SWA layers are now having many specific params that make it almost looks like a dedicated transformers inside the model. For example, SWA layers can use different head dims, freq_base, etc

Probably when more models do the same way as MiMo-V2, we can consider refactoring the conversion script / metadata loader such that `_swa` prefix is added / loaded automatically

CISC:
> Probably when more models do the same way as MiMo-V2, we can consider refactoring the conversion script / metadata loader such that `_swa` prefix is added / loaded automatically

I'll add some TODOs.

CISC:
I added `get_rope_freq_base/scale` for afmoe, llama4 and smallthinker even though `n_no_rope_layer_step` effectively makes it pointless for all current models, but seems safer.

src/llama-hparams.h
```diff
@@ -105,9 +105,9 @@ struct llama_hparams {
 
     float    rope_attn_factor = 1.0f;
     float    rope_freq_base_train;
-    float    rope_freq_base_train_swa;
+    float    rope_freq_base_train_swa  = 10000.0f;
     float    rope_freq_scale_train;
-    float    rope_freq_scale_train_swa;
+    float    rope_freq_scale_train_swa = 1.0f;
```
ggerganov, line 110: Wonder if we should instead set invalid values here (e.g. `-1.0f`) and assert in `ggml.c`. This way, if a model tries to use uninitialized `..._swa` params, we will know to go and initialize them correctly in `llama-model.cpp`.

src/llama-model.cpp
```diff
@@ -578,6 +578,7 @@ void llama_model::load_hparams(llama_model_loader & ml) {
     hparams.rope_scaling_type_train = llama_rope_scaling_type_from_string(rope_scaling);
     GGML_ASSERT(hparams.rope_scaling_type_train != LLAMA_ROPE_SCALING_TYPE_UNSPECIFIED);
 
+    // TODO: Handle SWA metadata equally when models start implementing it
```
ggerganov: nit: do you mean "similarly"?

```suggestion
    // TODO: Handle SWA metadata similarly when models start implementing it
```

ggerganov (approved):

src/llama-hparams.h
CISC: I'm not sure I follow where you meant to assert?

ggerganov: Like this:

```diff
diff --git a/ggml/src/ggml.c b/ggml/src/ggml.c
index eb3ae72ea..d1a0585fe 100644
--- a/ggml/src/ggml.c
+++ b/ggml/src/ggml.c
@@ -4068,6 +4068,9 @@ static struct ggml_tensor * ggml_rope_impl(
         bool                  inplace) {
     GGML_ASSERT((mode & 1) == 0 && "mode & 1 == 1 is no longer supported");
 
+    GGML_ASSERT(freq_base  >  0.0f);
+    GGML_ASSERT(freq_scale >= 0.0f);
+
     GGML_ASSERT(ggml_is_vector(b));
     GGML_ASSERT(b->type == GGML_TYPE_I32);
 
```

CISC: Won't make much sense down the line though, it should be the same as it effectively is for non-SWA today, ie. 10000/1.0:
https://github.com/ggml-org/llama.cpp/blob/2c42d47daf388adbfbc4addf1cf6f7c626758276/src/llama-model.cpp#L572-L574
https://github.com/ggml-org/llama.cpp/blob/2c42d47daf388adbfbc4addf1cf6f7c626758276/src/llama-model.cpp#L588

PR #17004, opened 2025-11-04
sampling : add support for backend sampling

danbev:
Add support for backend (like GPU) sampling

### `libllama` changes

- Update `llama_context_params`
- Update `llama_sampler_i`
- Add `llama_get_sampled_...` API
- Add `llama_set_sampler`
- Update `llama_sampler_init`
- Update `llama_sampler_chain_get`

### CUDA changes

- Optimize `argsort`, `top-k`, `cumsum`, `softmax` kernels
- Add CMake option `GGML_CUDA_CUB_3DOT2` - builds a pre-release [CCCL](https://github.com/nvidia/cccl) v3.2 candidate for improved performance. Temporary until it lands officially in CTK

### Tools changes

- Add new `-bs, --backend-sampling` argument for enabling the backend sampling functionality

### CI changes

- Enable `-DGGML_CUDA_CUB_3DOT2=ON` for releases
- Enable server CI tests for non-Windows builds
- Add server CI with `LLAMA_ARG_BACKEND_SAMPLING=1` for exercising the backend sampling

### Limitations

- Not compatible with grammar sampling
- Currently supports single output per sequence per batch
- Currently saving/loading the backend sampling state is not supported

### Recommendations

This functionality will take some time to mature enough for general usage, so for now it's recommended to use it only with:

- CUDA backend, built with `-DGGML_CUDA_CUB_3DOT2=ON`
- Args: `--backend-sampling --parallel 1`
- Samplers: `top-k`, `top-p`, `min-p`, `temp`

Other configurations might be unstable and potentially have poor performance when the backend sampling is enabled.

### Next PRs

- [ ] Avoid graph reallocations when changing sampling config (#18547)
- [ ] Do not compute inactive samplers during parallel decoding (#18550)
- [ ] Support multiple outputs per sequence per batch
- [ ] Optimize Metal `top-k`, `cumsum`
- [ ] Save/load sampling state
- [ ] Extend `llama-bench`/`llama-batched-bench` to support backend samplers?

---

The motivation for this feature is to enable sampling to be performed directly on the backend as part of the computation graph being executed, allowing for some or all of the sampling to be done on the backend.

For example, the backend sampler chain might select/sample a token directly in which case only the sampled token needs to be transferred from device memory to host memory.

It is also possible for the backend samplers to perform filtering of the logits, or compute and filter the probability distribution, in which case only the filtered logits or probabilites need to be transferred back to system memory for further processing by CPU samplers.

[...]

PR #18561, opened 2026-01-03
cuda : check src shapes for CUDA graphs

ggerganov:
rel https://github.com/ggml-org/llama.cpp/pull/17004#issuecomment-3706865663

Add checks for the `src` shapes to determine if a graph node has matching properties (in the context of CUDA graphs)

Not sure if this is the ideal solution. The failure case that is observed is a graph with a single `ggml_top_k()` node. For constant `k`, the output tensor of this node has the same shape, but the `src` tensor shape is different. With the logic on `master` we incorrectly determine that the node properties match and the CUDA graph is being reused.

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1041,6 +1041,8 @@ struct ggml_graph_node_properties {
     int64_t ne[GGML_MAX_DIMS];
     size_t nb[GGML_MAX_DIMS];
     void * src_address[GGML_MAX_SRC];
+    int64_t src_ne[GGML_MAX_SRC][GGML_MAX_DIMS];
+    size_t src_nb[GGML_MAX_SRC][GGML_MAX_DIMS];
     int32_t op_params[GGML_MAX_OP_PARAMS / sizeof(int32_t)];
 };
 
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2911,6 +2911,10 @@ static void set_ggml_graph_node_properties(ggml_tensor * node, ggml_graph_node_p
     }
     for (int i = 0; i < GGML_MAX_SRC; i++) {
         graph_node_properties->src_address[i] = node->src[i] ? node->src[i]->data : nullptr;
+        for (int j = 0; j < GGML_MAX_DIMS; j++) {
+            graph_node_properties->src_ne[i][j] = node->src[i] ? node->src[i]->ne[j] : 0;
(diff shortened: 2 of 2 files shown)


am17an:
Could we have the src shape remain the same or is that not possible? As I understand this would disable CUDA graphs entirely when using backend sampling. CUDA graphs at the moment are only used batch_size = 1 use case

ORippler:
> Could we have the src shape remain the same or is that not possible? As I understand this would disable CUDA graphs entirely when using backend sampling.

Since LLMs have constant vocab-sizes, this check should not trigger as we have constant input/output shapes for top-k as is already.

For batched inference in the server setting, afaik we execute a single cgraph with N-batched logit-gen followed by N-times batch-size 1 backend sampling. In that case, cuda graphs are currently disabled due to the batched logit-gen. Moreover, even in this case Input/output shapes for top-k are constant across cgraph invocations

ggerganov:
Yes, this patch is not relevant for normal `llama.cpp` inference. It fixes an issue with the `test-backend-ops` where we run many different graphs that have a single `top-k` node with different src shape. This is just one potential failure of the current CUDA graph logic and although it does not directly affect `llama.cpp` (apart from the tests) it still needs to be fixed.

am17an:
If it's not useful in llama.cpp perhaps test-backend-ops can be modified? This adds 1280 bytes per graph node, and this check will run on every TG cycle

ORippler:
> If it's not useful in llama.cpp perhaps test-backend-ops can be modified? This adds 1280 bytes per graph node, and this check will run on every TG cycle

ggml is intended to serve other consumers apart from llama.cpp also, so I would say we should fix it.

To reduce the time spent checking, we can forward the capability of re-using a `ggml_cgraph` from the `ggml_backend_sched` to the cuda backend, checking the graph only when necessary. I think this was scoped as part of the GraphPlan API where the backend scheduling was intended to be refactored/reworked, which is currently on hold due to diego's hiatus.

am17an:
I agree it should be fixed but not sure if this is the correct solution. I'm not sure how to estimate the impact of the memory increase, AFAIK there are some linear attn models which run tens of thousands of nodes, for them it would be reading 10s of MBs of data for every token. Perhaps it won't be a big deal, I can measure and report back

ggerganov:
Yes, this patch is an overkill. We really need to only check the shapes of the leaf nodes (i.e. source tensors that are not the result of any node in the graph). The shapes of the rest of the sources are already being checked with the existing logic as noted in https://github.com/ggml-org/llama.cpp/pull/17004#issuecomment-3706865663. A better fix can be implemented that would not add so much overhead.

am17an:
@ggerganov I can make that change, unless you are already going to make it

ggerganov:
@am17an Thanks, please go ahead

PR #18567, opened 2026-01-03
llama: max ctx by default, fix fit magic number

### LINKED ISSUE #18376: Misc. bug: Maximum Context Size
### Name and Version

  Device 0: NVIDIA RTX 2000 Ada Generation Laptop GPU, compute capability 8.9, VMM: yes
version: 7539 (83b3b1c27)

Looks like when we specify the context size to zero (eg: -c0 ), it's now defaulting to 4096 instead of the maximum size of the model.

### Operating systems

_No response_

_No response_

### Command line

```shell
llama-server --port 10004 --host 127.0.0.1  --api-key xxxxx --jinja -fa on -hf mradermacher/Ling-lite-GGUF --cache-type-k q8_0 --cache-type-v q8_0 -c 0
```

llama_context: flash_attn    = enabled
llama_context: kv_unified    = true
llama_context: freq_base     = 600000.0
llama_context: n_ctx_seq (4096) < n_ctx_train (16384) -- the full capacity of the model will not be utilized

It is taking 4096 context size instead of 16384.

### First Bad Commit

_No response_

### Relevant log output

```shell

```

JohannesGaessler:
Fixes https://github.com/ggml-org/llama.cpp/issues/18376 .

The intent with `llama_params_fit` is to only change those parameters that a user is not setting explicitly themself. However, if a user is explicitly setting `-c 0` this is currently being changed by `llama_params_fit`. I would propose the following changes to fix this:

* Change the data type of `llama_context_params::n_ctx` from `uint32_t` to `int32_t` with all values <= 0 being interpreted to mean that the full context should be used.
* Change the default value set `llama_context_default_params` from 512 to -1. This would mean with the llama C API by default the full context of the model would now be used, consistent with the common API.
* In `llama_params_fit`, check the context size against the new default value instead of vs. 0.
* For the CLI, add strings "auto" and "full" for `--ctx-size` which by default both use the full context but differ w.r.t. whether they should be changed by `llama_params_fit`.

An alternative approach would be keep the changes entirely in the common API.

### DIFF
--- a/include/llama.h
+++ b/include/llama.h
@@ -325,7 +325,7 @@ extern "C" {
     // NOTE: changing the default values of parameters marked as [EXPERIMENTAL] may cause crashes or incorrect results in certain configurations
     //       https://github.com/ggml-org/llama.cpp/pull/7544
     struct llama_context_params {
-        uint32_t n_ctx;             // text context, 0 = from model
+        int32_t  n_ctx;             // context size in tokens, use llama_model_n_ctx_train for values <= 0
         uint32_t n_batch;           // logical maximum batch size that can be submitted to llama_decode
         uint32_t n_ubatch;          // physical maximum batch size
         uint32_t n_seq_max;         // max number of sequences (i.e. distinct states for recurrent models)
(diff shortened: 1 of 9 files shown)


CISC:
Hmmm, what's with the sign disparity between common and llama params?

JohannesGaessler:
I think the biggest reason is simply that no one invested the effort to make things consistent. I would suggest that going forward, if we have non-negative quantities like the context size we store them internally as unsigned values but use signed values in `llama_model_params` and `llama_context_params` in order to encode special values.

include/llama.h
```diff
@@ -325,7 +325,7 @@ extern "C" {
     // NOTE: changing the default values of parameters marked as [EXPERIMENTAL] may cause crashes or incorrect results in certain configurations
     //       https://github.com/ggml-org/llama.cpp/pull/7544
     struct llama_context_params {
-        uint32_t n_ctx;             // text context, 0 = from model
+        int32_t  n_ctx;             // context size in tokens, use llama_model_n_ctx_train for values <= 0
```
ggerganov, line 328: Hm, I don't see why this change is needed. Looks like it allows negative values that have the same functionality as `n_ctx == 0`. So the user code can simply used `params.n_ctx = max(0, n_ctx);` to achieve the same.

ggerganov (commented):
> An alternative approach would be keep the changes entirely in the common API.

This seems like the better approach.

JohannesGaessler:
Superseded by https://github.com/ggml-org/llama.cpp/pull/19070 .

PR #18574, opened 2026-01-03
GGML RPC - Add support for Unix Domain Sockets

struct:
Making this PR mainly for feedback at this stage (cc @rgerganov). This adds Unix Domain Socket support to the ggml rpc-server. I see a few different use cases and reasons for adding this functionality:

* A local unix socket provides a local only transport and removes an easily misconfigured footgun of binding to a network interface like 0.0.0.0 when you only want local clients to connect. It also integrates with the POSIX permission model (socket path + directory permissions) and things like systemd. Other llama.cpp utilities such as llama-server support this model. This is a much safer configuration for the use cases below.

* Connecting to an rpc-server across container environments can be done using mounted volumes to share sockets.

* Workloads on the same machine can run at different privilege levels and within sandboxes with varying access to hardware, but early access to file descriptors is a common use case and supported pattern even in the most hardened sandboxes.

* With sendmsg/recvmsg we can add (not currently implemented in this PR) fine grained authentication over the socket to ensure only specific processes can invoke it.

**Summary of implementation changes in this PR**

`rpc-server.cpp` - Modified to support using a socket path that ends in `.sock` with the existing -H argument

`ggml-rpc.cpp` - Modified in 2 important ways. First, `parse_endpoint` now determines whether the endpoint passed into the function is a unix socket (ends in `.sock`) or a network interface to listen on.

`ggml_backend_rpc_start_server` is a `host:port` pair or a unix socket path. Introduces a new simple structure, `endpoint_info`, to hold this information. The unix socket is opened with `0770` permission which should allow for users in the same group to connect. This can be hardened in the future if desired, but is still safer than binding to 0.0.0.0 by default.

These changes work with other llama.cpp tooling as-is. Here is an example from my testing:

Run the rpc-server reusing the `-H` argument similar to `llama-server`:

```
$ build/bin/rpc-server -t 1 -H local.sock
  Device 0: NVIDIA GB10, compute capability 12.1, VMM: yes
register_backend: registered backend CUDA (1 devices)
register_device: registered device CUDA0 (NVIDIA GB10)
register_backend: registered backend RPC (0 devices)
register_backend: registered backend CPU (1 devices)
register_device: registered device CPU (CPU)
Starting RPC server v3.6.0
  endpoint       : local.sock
  local cache    : n/a
Devices:
ggml_backend_cuda_get_available_uma_memory: final available_memory_kb: 119304108
  CUDA0: NVIDIA GB10 (122570 MiB, 116507 MiB free)
Accepted client connection
Client connection closed
```

And `llama-cli` connecting to it:

```
$ build/bin/llama-cli -m ~/models/meta-llama-3.1-8b-instruct-abliterated.Q6_K.gguf -n 64 -ngl 99 --rpc local.sock
  Device 0: NVIDIA GB10, compute capability 12.1, VMM: yes
register_backend: registered backend CUDA (1 devices)
register_device: registered device CUDA0 (NVIDIA GB10)
register_backend: registered backend RPC (0 devices)
register_backend: registered backend CPU (1 devices)
register_device: registered device CPU (CPU)
register_backend: registered backend RPC[local.sock] (1 devices)
register_device: registered device RPC0 (local.sock)

Loading model...  
...[cropped]...
build      : b0-unknown
model      : meta-llama-3.1-8b-instruct-abliterated.Q6_K.gguf
modalities : text

available commands:

> This is a test!

A test! Well, I'm happy to report that I'm functioning properly and ready to assist with any...

[ Prompt: 284.8 t/s | Generation: 32.0 t/s ]
```

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2183,7 +2183,7 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
     if (llama_supports_rpc()) {
         add_opt(common_arg(
             {"--rpc"}, "SERVERS",
-            "comma separated list of RPC servers (host:port)",
+            "comma separated list of RPC servers (host:port or path ending in .sock)",
             [](common_params & params, const std::string & value) {
                 add_rpc_devices(value);
                 GGML_UNUSED(params);
--- a/ggml/src/ggml-rpc/ggml-rpc.cpp
+++ b/ggml/src/ggml-rpc/ggml-rpc.cpp
@@ -25,7 +25,10 @@
 #  include <netinet/tcp.h>
 #  include <netdb.h>
 #  include <unistd.h>
+#  include <sys/un.h>
+#  include <sys/stat.h>
 #endif
+#include <cctype>
 #include <cstring>
 #include <fstream>
 #include <filesystem>
@@ -314,7 +317,7 @@ static bool set_reuse_addr(sockfd_t sockfd) {
     return ret == 0;
 }
 
-static std::shared_ptr<socket_t> socket_connect(const char * host, int port) {
+static std::shared_ptr<socket_t> socket_connect_tcp(const char * host, int port) {
     struct sockaddr_in addr;
     auto sockfd = socket(AF_INET, SOCK_STREAM, 0);
     auto sock_ptr = make_socket(sockfd);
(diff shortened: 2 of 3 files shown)


rgerganov:
> * A local unix socket provides a local only transport and removes an easily misconfigured footgun of binding to a network interface like 0.0.0.0 when you only want local clients to connect.

How is this different to binding on `localhost`?

> * Connecting to an rpc-server across container environments can be done using mounted volumes to share sockets.

You can also create private networks and attach them to container instances. I guess using volumes + UDS would be easier to setup for local deployments?

> * Workloads on the same machine can run at different privilege levels and within sandboxes with varying access to hardware, but early access to file descriptors is a common use case and supported pattern even in the most hardened sandboxes.
> * With sendmsg/recvmsg we can add (not currently implemented in this PR) fine grained authentication over the socket to ensure only specific processes can invoke it.

All of this aims to improve the security when running everything on a single machine and I don't understand the use case for this. How do we accept remote connections when using UDS?

struct:
> How is this different to binding on localhost?

Binding to localhost might not allow a sandboxed client or container with network restrictions to connect but who might otherwise be able to open or inherit local file descriptors. Localhost also allows any client who can connect whereas the UDS approach gives you basic OS provided file permission model for free.

> You can also create private networks and attach them to container instances. I guess using volumes + UDS would be easier to setup for local deployments?

IME these are similar levels of complexity to setup and maintain but certainly adding UDS gives you optionality you did not have before.

> All of this aims to improve the security when running everything on a single machine and I don't understand the use case for this.

The use case is supporting clients on the same machine, or any client with access to the file descriptor via volume mount like a container, who may be running within very restrictive and isolated runtimes. In many of these cases they might be able to access one backend but not another due to those permission restrictions. 

>  How do we accept remote connections when using UDS?

One example (beyond the mounted volume approach as discussed) is using SSH to forward connections to a remote UDS. This would give you remote access to the USD, and authentication and confidentiality of the session without exposing a network socket. Of course you could do the same for a locally bound TCP port. In either case the network binding functionality remains and functions as it did before with no additional overhead or complexity as a result of adding the UDS functionality.

rgerganov:
Let's consider adding UDS support after having at least initial support for pipeline parallelism (PR #18626). The reason is that we may need to significantly rework RPC communication and supporting both TCP and UDS in the process would be a pain.

struct:
@rgerganov sounds good. Will watch that PR and rebase/update here as necessary when it lands.

PR #18583, opened 2026-01-04
ggml-cuda: check for srcs outside the cgraph

am17an:
supersede #18561. Tested on the failing tests in #17004 and they work fine.

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1065,6 +1065,7 @@ struct ggml_cuda_graph {
     int number_consecutive_updates = 0;
     bool cuda_graphs_enabled = false;
     std::vector<ggml_graph_node_properties> ggml_graph_properties;
+    std::vector<ggml_graph_node_properties> extraneous_srcs_properties;
 #endif
 };
 
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2980,8 +2980,13 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx, 
 
     // Loop over nodes in GGML graph to determine if CUDA graph update is required
     // and store properties to allow this comparison for the next token
+    std::unordered_map<ggml_tensor *, bool> seen_node;
+    std::vector<ggml_tensor *>              extraneous_srcs;
     for (int i = 0; i < cgraph->n_nodes; i++) {
         bool has_matching_properties = true;
+
+        seen_node[cgraph->nodes[i]] = true;
+
         if (!cuda_graph_update_required) {
             has_matching_properties = ggml_graph_node_has_matching_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
         }
@@ -2989,6 +2994,32 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx, 
             cuda_graph_update_required = true;
         }
         set_ggml_graph_node_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
+
+        for (int src_idx = 0; src_idx < GGML_MAX_SRC; ++src_idx) {
+            ggml_tensor * src = cgraph->nodes[i]->src[src_idx];
+            if (src && seen_node.find(src) == seen_node.end()) {
+                extraneous_srcs.push_back(src);
+            }
+        }
+    }
+
+    if (cuda_ctx->cuda_graph->extraneous_srcs_properties.size() != (size_t) extraneous_srcs.size()) {
+        cuda_graph_update_required = true;
+        cuda_ctx->cuda_graph->extraneous_srcs_properties.resize(extraneous_srcs.size());
+    }
+
+    for (size_t i = 0; i < extraneous_srcs.size(); ++i) {
+        bool has_matching_properties = true;
+
+        if (!cuda_graph_update_required) {
+            has_matching_properties = ggml_graph_node_has_matching_properties(
+                extraneous_srcs[i], &cuda_ctx->cuda_graph->extraneous_srcs_properties[i]);
+        }
+
+        if (!has_matching_properties) {
+            cuda_graph_update_required = true;
+        }
+        set_ggml_graph_node_properties(extraneous_srcs[i], &cuda_ctx->cuda_graph->extraneous_srcs_properties[i]);
     }
 
     return cuda_graph_update_required;
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2973,19 +2973,15 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx, 
     }
 
     // Check if the graph size has changed
-    if (cuda_ctx->cuda_graph->ggml_graph_properties.size() != (size_t)cgraph->n_nodes) {
+    if (cuda_ctx->cuda_graph->ggml_graph_properties.size() != (size_t)cgraph->n_nodes + cgraph->n_leafs) {
         cuda_graph_update_required = true;
-        cuda_ctx->cuda_graph->ggml_graph_properties.resize(cgraph->n_nodes);
+        cuda_ctx->cuda_graph->ggml_graph_properties.resize(cgraph->n_nodes + cgraph->n_leafs);
     }
 
     // Loop over nodes in GGML graph to determine if CUDA graph update is required
     // and store properties to allow this comparison for the next token
-    std::unordered_map<ggml_tensor *, bool> seen_node;
-    std::vector<ggml_tensor *>              extraneous_srcs;
     for (int i = 0; i < cgraph->n_nodes; i++) {
         bool has_matching_properties = true;
-
-        seen_node[cgraph->nodes[i]] = true;
 
         if (!cuda_graph_update_required) {
             has_matching_properties = ggml_graph_node_has_matching_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
@@ -2994,32 +2990,17 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx, 
             cuda_graph_update_required = true;
         }
         set_ggml_graph_node_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
-
-        for (int src_idx = 0; src_idx < GGML_MAX_SRC; ++src_idx) {
-            ggml_tensor * src = cgraph->nodes[i]->src[src_idx];
-            if (src && seen_node.find(src) == seen_node.end()) {
-                extraneous_srcs.push_back(src);
-            }
-        }
     }
 
-    if (cuda_ctx->cuda_graph->extraneous_srcs_properties.size() != (size_t) extraneous_srcs.size()) {
-        cuda_graph_update_required = true;
-        cuda_ctx->cuda_graph->extraneous_srcs_properties.resize(extraneous_srcs.size());
-    }
-
-    for (size_t i = 0; i < extraneous_srcs.size(); ++i) {
+    for (int i = 0; i < cgraph->n_leafs; i++) {
         bool has_matching_properties = true;
-
         if (!cuda_graph_update_required) {
-            has_matching_properties = ggml_graph_node_has_matching_properties(
-                extraneous_srcs[i], &cuda_ctx->cuda_graph->extraneous_srcs_properties[i]);
+            has_matching_properties = ggml_graph_node_has_matching_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
         }
-
         if (!has_matching_properties) {
             cuda_graph_update_required = true;
         }
-        set_ggml_graph_node_properties(extraneous_srcs[i], &cuda_ctx->cuda_graph->extraneous_srcs_properties[i]);
+        set_ggml_graph_node_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
     }
 
     return cuda_graph_update_required;


ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2977,15 +2977,46 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx,
 
     // Loop over nodes in GGML graph to determine if CUDA graph update is required
     // and store properties to allow this comparison for the next token
+    std::unordered_map<ggml_tensor *, bool> seen_node;
```
ggerganov: Can be `std::unordered_set`?

ggerganov (commented):
Actually I just realized that we already have the leafs stored in the graph, so we can directly iterate over them. The patch should be simply:

```diff
diff --git a/ggml/src/ggml-cuda/ggml-cuda.cu b/ggml/src/ggml-cuda/ggml-cuda.cu
index f568bd489..bec13b7d6 100644
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2973,9 +2973,9 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx,
     }
 
     // Check if the graph size has changed
-    if (cuda_ctx->cuda_graph->ggml_graph_properties.size() != (size_t)cgraph->n_nodes) {
+    if (cuda_ctx->cuda_graph->ggml_graph_properties.size() != (size_t)cgraph->n_nodes + cgraph->n_leafs) {
         cuda_graph_update_required = true;
-        cuda_ctx->cuda_graph->ggml_graph_properties.resize(cgraph->n_nodes);
+        cuda_ctx->cuda_graph->ggml_graph_properties.resize(cgraph->n_nodes + cgraph->n_leafs);
     }
 
     // Loop over nodes in GGML graph to determine if CUDA graph update is required
@@ -2991,6 +2991,17 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx,
         set_ggml_graph_node_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
     }
 
+    for (int i = 0; i < cgraph->n_leafs; i++) {
+        bool has_matching_properties = true;
+        if (!cuda_graph_update_required) {
+            has_matching_properties = ggml_graph_node_has_matching_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
+        }
+        if (!has_matching_properties) {
+            cuda_graph_update_required = true;
+        }
+        set_ggml_graph_node_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
+    }
+
     return cuda_graph_update_required;
 }
 
```

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2967,15 +2977,46 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx,
...
             cuda_graph_update_required = true;
         }
         set_ggml_graph_node_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
+
+        for (int src_idx = 0; src_idx < GGML_MAX_SRC; ++src_idx) {
+            ggml_tensor * src = cgraph->nodes[i]->src[src_idx];
+            if (src && seen_node.find(src) == seen_node.end()) {
+                extraneous_srcs.push_back(src);
+            }
+        }
```
ORippler: Don't we have to loop twice over all nodes, once to populate `seen_node` and then once again to check the inputs of each node?

am17an: My assumption is that is the graph is always a DAG and assumes sequential execution, so a node's src has to occur before the node itself (if it is in the graph)

ggml/src/ggml-cuda/ggml-cuda.cu
am17an: I will just use your patch, won't need this then

ORippler:
For my understanding, leafs in the context of ggml refer to nodes in our directed (acyclic) `ggml_cgraph` with an indegree of 0? I.e. they don't have any incoming edges? If so, shouldn't they also be all tensors marked as inputs in their respective flags?

ggerganov:
> For my understanding, leafs in the context of ggml refer to nodes in our directed (acyclic) ggml_cgraph with an indegree of 0? I.e. they don't have any incoming edges?

Yes, here is the part of the DFS traversal that separates the leafs from the nodes:

https://github.com/ggml-org/llama.cpp/blob/4974bf53cf14073c7b66e1151348156aabd42cb8/ggml/src/ggml.c#L6749-L6760

> If so, shouldn't they also be all tensors marked as inputs in their respective flags?

Marking a tensor as input is technically optional for legacy reasons, though probably today it's safe to assume that all leafs would be marked as inputs.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2967,15 +2977,46 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx,
...
+
+    if (cuda_ctx->cuda_graph->extraneous_srcs_properties.size() != (size_t) extraneous_srcs.size()) {
+        cuda_graph_update_required = true;
+        cuda_ctx->cuda_graph->extraneous_srcs_properties.resize(extraneous_srcs.size());
+    }
+
+    for (size_t i = 0; i < extraneous_srcs.size(); ++i) {
+        bool has_matching_properties = true;
+
+        if (!cuda_graph_update_required) {
```
JohannesGaessler, line 2997: What is the purpose of this check?

am17an: it skips calling `ggml_graph_node_has_matching_properties`? Or did you mean something else

JohannesGaessler: I meant to ask whether it is relevant for program correctness but based on your answer I guess not.

JohannesGaessler (approved):

ggml/src/ggml-cuda/common.cuh
```diff
@@ -1065,6 +1065,7 @@ struct ggml_cuda_graph {
     int number_consecutive_updates = 0;
     bool cuda_graphs_enabled = false;
     std::vector<ggml_graph_node_properties> ggml_graph_properties;
+    std::vector<ggml_graph_node_properties> extraneous_srcs_properties;
```
ggerganov, line 1068: This is no longer used

ggerganov (approved):
A small recommendation for naming improvements as I was looking through this code:

```diff
diff --git a/ggml/src/ggml-cuda/common.cuh b/ggml/src/ggml-cuda/common.cuh
index 995b774c2..441bda22c 100644
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1036,7 +1036,7 @@ struct ggml_tensor_extra_gpu {
 #define USE_CUDA_GRAPH
 #endif
 
-struct ggml_graph_node_properties {
+struct ggml_cuda_graph_node_properties {
     void * node_address;
     ggml_op node_op;
     int64_t ne[GGML_MAX_DIMS];
@@ -1064,8 +1064,7 @@ struct ggml_cuda_graph {
     bool disable_due_to_failed_graph_capture = false;
     int number_consecutive_updates = 0;
     bool cuda_graphs_enabled = false;
-    std::vector<ggml_graph_node_properties> ggml_graph_properties;
-    std::vector<ggml_graph_node_properties> extraneous_srcs_properties;
+    std::vector<ggml_cuda_graph_node_properties> props;
 #endif
 };
 
diff --git a/ggml/src/ggml-cuda/ggml-cuda.cu b/ggml/src/ggml-cuda/ggml-cuda.cu
index 75269170c..682d38ebf 100644
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2915,41 +2915,41 @@ static bool check_node_graph_compatibility(ggml_cgraph * cgraph,
     return use_cuda_graph;
 }
 
-static void set_ggml_graph_node_properties(ggml_tensor * node, ggml_graph_node_properties * graph_node_properties) {
-    graph_node_properties->node_address = node->data;
-    graph_node_properties->node_op = node->op;
+static void ggml_cuda_graph_set_node_properties(ggml_tensor * node, ggml_cuda_graph_node_properties * props) {
+    props->node_address = node->data;
+    props->node_op = node->op;
     for (int i = 0; i < GGML_MAX_DIMS; i++) {
-        graph_node_properties->ne[i] = node->ne[i];
-        graph_node_properties->nb[i] = node->nb[i];
+        props->ne[i] = node->ne[i];
+        props->nb[i] = node->nb[i];
     }
     for (int i = 0; i < GGML_MAX_SRC; i++) {
-        graph_node_properties->src_address[i] = node->src[i] ? node->src[i]->data : nullptr;
+        props->src_address[i] = node->src[i] ? node->src[i]->data : nullptr;
     }
-    memcpy(graph_node_properties->op_params, node->op_params, GGML_MAX_OP_PARAMS);
+    memcpy(props->op_params, node->op_params, GGML_MAX_OP_PARAMS);
 }
 
-static bool ggml_graph_node_has_matching_properties(ggml_tensor * node, ggml_graph_node_properties * graph_node_properties) {
-    if (node->data != graph_node_properties->node_address &&
+static bool ggml_cuda_graph_has_matching_properties(ggml_tensor * node, ggml_cuda_graph_node_properties * props) {
+    if (node->data != props->node_address &&
           node->op != GGML_OP_VIEW) {
         return false;
     }
 
-    if (node->op != graph_node_properties->node_op) {
+    if (node->op != props->node_op) {
         return false;
     }
 
     for (int i = 0; i < GGML_MAX_DIMS; i++) {
-        if (node->ne[i] != graph_node_properties->ne[i]) {
+        if (node->ne[i] != props->ne[i]) {
             return false;
         }
-        if (node->nb[i] != graph_node_properties->nb[i]) {
+        if (node->nb[i] != props->nb[i]) {
             return false;
         }
     }
 
     for (int i = 0; i < GGML_MAX_SRC; i++) {
         if (node->src[i] &&
-            node->src[i]->data != graph_node_properties->src_address[i] &&
+            node->src[i]->data != props->src_address[i] &&
             node->op != GGML_OP_VIEW
         ) {
             return false;
@@ -2957,53 +2957,51 @@ static bool ggml_graph_node_has_matching_properties(ggml_tensor * node, ggml_gra
     }
 
     if ((node->op == GGML_OP_SCALE || node->op == GGML_OP_GLU) &&
-        memcmp(graph_node_properties->op_params, node->op_params, GGML_MAX_OP_PARAMS) != 0) {
+        memcmp(props->op_params, node->op_params, GGML_MAX_OP_PARAMS) != 0) {
         return false;
     }
 
     return true;
 }
 
-static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx, ggml_cgraph * cgraph) {
-
-    bool cuda_graph_update_required = false;
+static bool ggml_cuda_graph_is_update_required(ggml_backend_cuda_context * cuda_ctx, ggml_cgraph * cgraph) {
+    bool res = false;
 
     if (cuda_ctx->cuda_graph->instance == nullptr) {
-        cuda_graph_update_required = true;
+        res = true;
     }
 
     // Check if the graph size has changed
-    if (cuda_ctx->cuda_graph->ggml_graph_properties.size() != (size_t)cgraph->n_nodes + cgraph->n_leafs) {
-        cuda_graph_update_required = true;
-        cuda_ctx->cuda_graph->ggml_graph_properties.resize(cgraph->n_nodes + cgraph->n_leafs);
+    if (cuda_ctx->cuda_graph->props.size() != (size_t)cgraph->n_nodes + cgraph->n_leafs) {
+        res = true;
+        cuda_ctx->cuda_graph->props.resize(cgraph->n_nodes + cgraph->n_leafs);
     }
 
     // Loop over nodes in GGML graph to determine if CUDA graph update is required
     // and store properties to allow this comparison for the next token
     for (int i = 0; i < cgraph->n_nodes; i++) {
-        bool has_matching_properties = true;
-
-        if (!cuda_graph_update_required) {
-            has_matching_properties = ggml_graph_node_has_matching_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
+        bool props_match = true;
+        if (!res) {
+            props_match = ggml_cuda_graph_has_matching_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->props[i]);
         }
-        if (!has_matching_properties) {
-            cuda_graph_update_required = true;
+        if (!props_match) {
+            res = true;
         }
-        set_ggml_graph_node_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
+        ggml_cuda_graph_set_node_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->props[i]);
     }
 
     for (int i = 0; i < cgraph->n_leafs; i++) {
-        bool has_matching_properties = true;
-        if (!cuda_graph_update_required) {
-            has_matching_properties = ggml_graph_node_has_matching_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
+        bool props_match = true;
+        if (!res) {
+            props_match = ggml_cuda_graph_has_matching_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->props[cgraph->n_nodes + i]);
         }
-        if (!has_matching_properties) {
-            cuda_graph_update_required = true;
+        if (!props_match) {
+            res = true;
         }
-        set_ggml_graph_node_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
+        ggml_cuda_graph_set_node_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->props[cgraph->n_nodes + i]);
     }
 
-    return cuda_graph_update_required;
+    return res;
 }
 
 static void update_cuda_graph_executable(ggml_backend_cuda_context * cuda_ctx) {
@@ -3773,7 +3771,7 @@ static enum ggml_status ggml_backend_cuda_graph_compute(ggml_backend_t backend,
     use_cuda_graph = ggml_cuda_set_cuda_graph_enabled(cuda_ctx);
 
     if (use_cuda_graph) {
-        cuda_graph_update_required = is_cuda_graph_update_required(cuda_ctx, cgraph);
+        cuda_graph_update_required = ggml_cuda_graph_is_update_required(cuda_ctx, cgraph);
 
         use_cuda_graph = check_node_graph_compatibility(cgraph, use_cuda_graph);

```

Up to your preference of course.

am17an:
Yeah the code needs a refactor, I can do it in this PR or another one

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2991,6 +2992,17 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx,
...
 
+    for (int i = 0; i < cgraph->n_leafs; i++) {
+        bool has_matching_properties = true;
+        if (!cuda_graph_update_required) {
+            has_matching_properties = ggml_graph_node_has_matching_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
+        }
+        if (!has_matching_properties) {
+            cuda_graph_update_required = true;
+        }
+        set_ggml_graph_node_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
```
ORippler, line 3003:
```suggestion
            cuda_graph_update_required = true;
            set_ggml_graph_node_properties(cgraph->leafs[i], &cuda_ctx->cuda_graph->ggml_graph_properties[cgraph->n_nodes + i]);
        }
```
Shouldn't we have to update the node properties only if they are not matching? Or do we set something with `set_ggml_graph_node_properties` that we don't check for in `ggml_graph_node_has_matching_properties`

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2991,6 +2992,17 @@ static bool is_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx,
         set_ggml_graph_node_properties(cgraph->nodes[i], &cuda_ctx->cuda_graph->ggml_graph_properties[i]);
```
ORippler, line 2992: Can't we move this setting inside the if condition as well? (see comment below)

ORippler (commented):
> Yeah the code needs a refactor, I can do it in this PR or another one

Depends on the scope, but since logic has been reviewed & approved here I'd personally do it in a follow-up PR (though this comes at the risk of the refactoring being deprioritized 🙃)

ggml/src/ggml-cuda/ggml-cuda.cu
am17an: I'm not sure, just following the existing pattern here. Will address in the refactor

PR #18584, opened 2026-01-04
ggml-backend : fallback to cpu memory when backend reports zero memory

xvyv99:
Fix #18577

Some GGML backends(e.g. BLAS, IBM zDNN, AMD ZenDNN, OpenCL) may report both total and free device memory as zero. This can lead to invalid tensor split computation during model loading, including NaN split values and out-of-range device indexing.

Previously this PR #18578 was addressed in a backend-specific way (e.g. BLAS), but the underlying issue is generic: backends are allowed to report zero memory and should fall back to host memory provided by `ggml-cpu`.

This PR fixes the issue at the GGML level by updating `ggml_backend_dev_memory` to fall back to the `ggml-cpu` backend whenever a backend reports zero total and free memory, ensuring valid memory information is always available for tensor placement.

PR #18587, opened 2026-01-04
llama: use host memory if device reports 0 memory

### LINKED ISSUE #18577: Misc. bug: out-of-range access during model loading with BLAS backend
### Name and Version

version: 7623 (cef1d23c5)

### Operating systems

Linux

llama-cli

### Command line

```shell
bin/llama-cli -m ./Qwen3-8B-Q8_0.gguf --device BLAS
```

When using the BLAS backend (`--device BLAS`), model loading
consistently crashes during `llama_model::load_tensors()`.

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console
Loading model... \llama_model_load: error loading model: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
|llama_params_fit: encountered an error while trying to fit params to free device memory: failed to load model
-llama_model_load: error loading model: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
common_init_from_params: failed to load model '/home/HwHiAiUser/Downloads/gpt2.Q8_0.gguf'
srv    load_model: failed to load model, '/home/HwHiAiUser/Downloads/gpt2.Q8_0.gguf'
 
Failed to load the model
```
</details>

taronaeo:
fixes: #18577
obsoletes: #18584

Backends can currently report `*free = 0; *total = 0;` if they do not have any memory information to report or if they are using host memory (e.g., BLAS, IBM zDNN). With the introduction of `llama_params_fit`, these values are now dependent on and when running `llama-cli` with f.ex., `--device BLAS`, it will result in the following error

```log
load_tensors: loading model tensors, this can take a while... (mmap = true)
llama_model_load: error loading model: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
common_init_from_params: failed to load model '/home/taronaeo/.cache/llama.cpp/LiquidAI_LFM2-8B-A1B-GGUF_LFM2-8B-A1B-Q4_K_M.gguf'
```

This PR updates the memory reporting logic within llama where if the device reports 0 bytes of free and total memory, the host memory information will instead be used for `llama_params_fit` and tensor split.

For OpenCL, I've updated the memory reporting values to 0 bytes instead of 1 byte to automatically use the host memory information. Unless OpenCL can report the actual device memory information :)

### DIFF
--- a/ggml/src/ggml-backend.cpp
+++ b/ggml/src/ggml-backend.cpp
@@ -27,6 +27,10 @@
 #include <sys/sysctl.h>
 #endif
 
+#ifndef _WIN32
+#include <unistd.h>
+#endif
+
 
 // backend buffer type
 
@@ -490,6 +494,11 @@ const char * ggml_backend_dev_description(ggml_backend_dev_t device) {
 void ggml_backend_dev_memory(ggml_backend_dev_t device, size_t * free, size_t * total) {
     GGML_ASSERT(device);
     device->iface.get_memory(device, free, total);
+
+    // fallback to host memory if device reports zero memory
+    if (*free == 0 && *total == 0) {
+        ggml_backend_get_host_memory(free, total);
+    }
 }
 
 enum ggml_backend_dev_type ggml_backend_dev_type(ggml_backend_dev_t device) {
@@ -500,6 +509,11 @@ enum ggml_backend_dev_type ggml_backend_dev_type(ggml_backend_dev_t device) {
 void ggml_backend_dev_get_props(ggml_backend_dev_t device, struct ggml_backend_dev_props * props) {
     memset(props, 0, sizeof(*props));
     device->iface.get_props(device, props);
+
+    // fallback to host memory if device reports zero memory
+    if (props->memory_free == 0 && props->memory_total == 0) {
+        ggml_backend_get_host_memory(&props->memory_free, &props->memory_total);
+    }
 }
 
 ggml_backend_reg_t ggml_backend_dev_backend_reg(ggml_backend_dev_t device) {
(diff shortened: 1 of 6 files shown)


lhez:
Looks good for OpenCL. Indeed, OpenCL doesn't provide a way to query free memory.

ggml/src/ggml-backend.cpp
```diff
@@ -490,6 +494,11 @@ const char * ggml_backend_dev_description(ggml_backend_dev_t device) {
 void ggml_backend_dev_memory(ggml_backend_dev_t device, size_t * free, size_t * total) {
     GGML_ASSERT(device);
     device->iface.get_memory(device, free, total);
+
+    // fallback to host memory if device reports zero memory
+    if (*free == 0 && *total == 0) {
+        ggml_backend_get_host_memory(free, total);
+    }
```
ggerganov: I don't think we need to implement this logic here. The user code can decide what to do when the device does not report memory - it can query the host memory, or it can do something else.

taronaeo: Makes sense from a library standpoint. I've moved this logic to `llama_get_device_memory_data`, where devices reporting 0 bytes for free and total memory will automatically use the host memory information. Let me know if this works as an alternative.

Edit: Was thinking if this check should exclude GPU type devices since it should not be using host memory information at all.

src/llama.cpp
```diff
@@ -113,6 +113,13 @@ static std::vector<llama_device_memory_data> llama_get_device_memory_data(
     for (size_t i = 0; i < ret.size(); i++) {
         size_t free, total;
         ggml_backend_dev_memory(model->devices[i], &free, &total);
+
+        // devices can return 0 bytes for free and total memory if they do not
+        // have any to report. in this case, we will use the host memory as a fallback
+        // fixes: https://github.com/ggml-org/llama.cpp/issues/18577
+        if (free == 0 && total == 0) {
+            ggml_backend_get_host_memory(&free, &total);
```
ggerganov: We can simply query the CPU backend with `ggml_backend_dev_memory()` - no need for this new API.

taronaeo: Fixed in latest commit.

CISC (approved):

taronaeo:
Let's merge in a few hours if there are no further comments :)

PR #18588, opened 2026-01-04
mtmd : fix integer overflow when n_tokens equals INT32_MIN

ylwango613:
## Summary
Fixes an integer overflow vulnerability (GHSA-gqvc-wp2r-9gr6) in `mtmd_tokenize_text_internal` that was previously patched in `common_tokenize` but not propagated to the MTMD tool.

## Problem
When `llama_tokenize` returns `INT32_MIN` (-2147483648) to indicate the input text is too large, the code performs a negation operation:

```cpp
if (n_tokens < 0) {
    result.resize(-n_tokens);  // Integer overflow when n_tokens == INT32_MIN
    ...
}
```

Negating `INT32_MIN` causes undefined behavior due to integer overflow, as `-(-2147483648)` cannot be represented in `int32_t`, resulting in the same value `-2147483648` and causing `std::vector::resize()` to throw `std::length_error`.

## Solution
Apply the same fix that was previously implemented in llama.cpp/common/common.cpp `common_tokenize`:

```cpp
if (n_tokens == std::numeric_limits<int32_t>::min()) {
    throw std::runtime_error("Tokenization failed: input text too large, tokenization result exceeds int32_t limit");
}
if (n_tokens < 0) {
    result.resize(-n_tokens);
    ...
}
```

This check detects the overflow condition and throws a clear error message instead of triggering undefined behavior.

### DIFF
--- a/tools/mtmd/mtmd.cpp
+++ b/tools/mtmd/mtmd.cpp
@@ -774,6 +774,10 @@ struct mtmd_tokenizer {
         int n_tokens = text.length() + 2 * add_special;
         std::vector<llama_token> result(n_tokens);
         n_tokens = llama_tokenize(vocab, text.data(), text.length(), result.data(), result.size(), add_special, parse_special);
+        // -2147483648 is std::numeric_limits<int32_t>::min()
+        if (n_tokens == -2147483648) {
+            throw std::runtime_error("Tokenization failed: input text too large, tokenization result exceeds int32_t limit");
+        }
         if (n_tokens < 0) {
             result.resize(-n_tokens);
             int check = llama_tokenize(vocab, text.data(), text.length(), result.data(), result.size(), add_special, parse_special);
--- a/tools/mtmd/mtmd.cpp
+++ b/tools/mtmd/mtmd.cpp
@@ -774,8 +774,7 @@ struct mtmd_tokenizer {
         int n_tokens = text.length() + 2 * add_special;
         std::vector<llama_token> result(n_tokens);
         n_tokens = llama_tokenize(vocab, text.data(), text.length(), result.data(), result.size(), add_special, parse_special);
-        // -2147483648 is std::numeric_limits<int32_t>::min()
-        if (n_tokens == -2147483648) {
+        if (n_tokens == std::numeric_limits<int32_t>::min()) {
             throw std::runtime_error("Tokenization failed: input text too large, tokenization result exceeds int32_t limit");
         }
         if (n_tokens < 0) {


tools/mtmd/mtmd.cpp
```diff
@@ -774,6 +774,10 @@ struct mtmd_tokenizer {
         int n_tokens = text.length() + 2 * add_special;
         std::vector<llama_token> result(n_tokens);
         n_tokens = llama_tokenize(vocab, text.data(), text.length(), result.data(), result.size(), add_special, parse_special);
+        // -2147483648 is std::numeric_limits<int32_t>::min()
+        if (n_tokens == -2147483648) {
```
taronaeo:
```suggestion
        if (n_tokens == std::numeric_limits<int32_t>::min()) {
```

taronaeo (changes requested):
You can just follow the same change propagated in `common.cpp`

CISC:
> GHSA-gqvc-wp2r-9gr6

Why?

ngxson:
This account has previously send low-quality reports to multiple repos. I won't accept any PRs from this account.

tools/mtmd/mtmd.cpp
ylwango613: This would require introducing `#include <limits>`. I'm not entirely sure if I'm allowed to add header files, so to be safe, I used the specific compile-time value instead.

ylwango613:
> This account has previously send low-quality reports to multiple repos. I won't accept any PRs from this account.

I'd appreciate clarity on this assessment. My track record includes ~5 PRs submitted (1 acknowledged issue with harfbuzz, others under review) and security bugs discovered in libplist, llama.cpp, htmldoc, and ollama. If there are specific concerns about this PR's quality or my previous work, I'm happy to address them. This PR fixes a documented security issue (GHSA-3p75-pq92-q28f) using the existing pattern from the codebase.

ngxson:
I don't want to point it out, but these kinds of bugs are quite nitpicks and doesn't need to be a whole vulnerability report procedure for it.

Realistically say, this bug alone only throw a exception, it doesn't cause data corruption or server crash, and inputting this much tokens into llama.cpp is not a realistic attack vector either. Other things will fail before this error become a problem.

So no, this PR is just noise to me. It fits the category of "making a small bugs looks dangerous by adding a lot of explanations about it."

ylwango613:
I don’t fully agree with your perspective.
This piece of code was copied from an earlier implementation where a vulnerability had already been identified and fixed. However, the current version still reflects the pre-fix state, which means it is exposed to the same underlying issue and threat model.

Regarding exploitability:
The code uses an int to store the token count. If an attacker sends a prompt containing 2147483648 tokens, the value overflows and becomes -2147483648. Later, the code negates this value, but negating -2147483648 still results in -2147483648, which causes the subsequent resize operation to fail and leads to a crash.Since there is currently no enforced limit on prompt length, an attacker can theoretically submit arbitrarily long input, making this issue triggerable and exploitable.

On whether this warrants a full vulnerability report:
From a security standpoint, it is our responsibility to provide thorough and well-documented reports for any identified bugs, so developers can quickly assess the root cause, impact, and necessity of a fix. This is not about exaggerating risk, but about ensuring clarity, correctness, and long-term robustness of the codebase.

PR #18590, opened 2026-01-04
Initialized git lfs in every test for RISC-V CI

alitariq4589:

I have added `git lfs install` in every RISC-V test, so it gets initialized automatically every time a test is run. This will also help when adding a new board, as there will be no requirement to run the command manually after adding and setting it up for the first time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1762,6 +1762,8 @@ jobs:
           rustup install stable
           rustup default stable
 
+          git lfs install
+
       - name: Clone
         id: checkout
         uses: actions/checkout@v4
@@ -1859,6 +1861,8 @@ jobs:
           rustup install stable
           rustup default stable
 
+          git lfs install
+
       - name: GCC version check
         run: |
           gcc --version
@@ -1951,6 +1955,8 @@ jobs:
           rustup install stable
           rustup default stable
 
+          git lfs install
+
       - name: GCC version check
         run: |
           gcc --version
@@ -2022,6 +2028,8 @@ jobs:
           # Install Rust stable version
           rustup install stable
           rustup default stable
+
+          git lfs install
 
       - name: GCC version check
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1750,7 +1750,7 @@ jobs:
           sudo apt-get update
 
           # Install necessary packages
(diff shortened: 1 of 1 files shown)


alitariq4589:
@CISC This should resolve the issue mentioned in #18433

CISC (approved):

CISC:
Thanks. :)

taronaeo:
https://github.com/ggml-org/llama.cpp/actions/runs/20691730115/job/59400828649?pr=18590#step:2:55

Maybe this might be required?

```bash
sudo apt install git-lfs -y
```

alitariq4589:
> https://github.com/ggml-org/llama.cpp/actions/runs/20691730115/job/59400828649?pr=18590#step:2:55
> 
> Maybe this might be required?
> 
> ```shell
> sudo apt install git-lfs -y
> ```

Oh you are right. I thought I had already added this. I am adding this now.

CISC:
I don't get it, what fails?
https://github.com/ggml-org/llama.cpp/actions/runs/20692414813/job/59402402150?pr=18590

CISC:
Also, this one I've seen a few times, network issues?
https://github.com/ggml-org/llama.cpp/actions/runs/20692414813/job/59402402102?pr=18590

alitariq4589:
> I don't get it, what fails?
> https://github.com/ggml-org/llama.cpp/actions/runs/20692414813/job/59402402150?pr=18590

This could be a network issue or the issue with the github actions package for RISC-V. But I see that the board used in this workflow was not disconnected. So it is more like some issue with github actions for RISC-V. This happens sometimes but I think the build is successful so far because it is just post clone causing the issue. I doubt that this will happen in the same build again.


> Also, this one I've seen a few times, network issues?
> https://github.com/ggml-org/llama.cpp/actions/runs/20692414813/job/59402402102?pr=18590

This is also github actions package issue. 

I will try to pinpoint what is wrong and rebase the package with latest version but I think these kinds of issues will only be truly resolved when github actions package official package maintainers start providing support for RISC-V (which they will not provide until .NET is officially supported for RISC-V). I will try to build the package with latest version of .NET in coming weeks to see if that solves issues like these.

alitariq4589:
@CISC Can you please re run the failing workflows so I can see if this is something specific to current commit?

taronaeo:
> @CISC Can you please re run the failing workflows so I can see if this is something specific to current commit?

Re-triggered.

PR #18594, opened 2026-01-04
model : mtmd : make input norm optional in LFM2-VL

tdakhran:
Upcoming LFM2-VL releases will have configurable input norm. See https://github.com/huggingface/transformers/pull/43087 for details.


ngxson (approved):
🚀 🚀

PR #18601, opened 2026-01-04
memory : add llama_memory_hybrid_iswa

tdakhran:
Upcoming models use hybrid caches with sliding window attention.

Tried modifying the existing `llama_memory_hybrid` to use `llama_kv_cache_iswa`, but the change was very intrusive.

I don't like the code much, it's essentially a copy-paste done by Claude. But it works! Not very confident with the memory subsystem and open to suggestions and feedback on how to improve it.

### DIFF
--- a/src/llama-memory-hybrid-iswa.cpp
+++ b/src/llama-memory-hybrid-iswa.cpp
@@ -1,0 +1,279 @@
+#include "llama-memory-hybrid-iswa.h"
+
+#include "llama-impl.h"
+#include "llama-model.h"
+#include "llama-context.h"
+
+//
+// llama_memory_hybrid_iswa
+//
+
+llama_memory_hybrid_iswa::llama_memory_hybrid_iswa(
+        const llama_model & model,
+                            /* attn */
+                ggml_type   type_k,
+                ggml_type   type_v,
+                     bool   v_trans,
+                     bool   swa_full,
+                 uint32_t   kv_size,
+                 uint32_t   n_ubatch,
+                 uint32_t   n_pad,
+                            /* recurrent */
+                ggml_type   type_r,
+                ggml_type   type_s,
+                 uint32_t   rs_size,
+                            /* common */
+                 uint32_t   n_seq_max,
+                     bool   offload,
+                     bool   unified,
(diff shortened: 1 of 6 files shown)


tdakhran:
@ggerganov , #18641 works on top of this PR, and shows that it works. Please let me know if something else is needed to merge.

UPD: PR doesn't affect any existing models or functionality.

src/llama-memory-hybrid-iswa.cpp
```diff
@@ -0,0 +1,279 @@
...
+    }
+    mem_recr->state_write(io, seq_id, flags);
+}
+
+void llama_memory_hybrid_iswa::state_read(llama_io_read_i & io, llama_seq_id seq_id, llama_state_seq_flags flags) {
+    if ((flags & LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY) == 0) {
+        mem_attn->state_read(io, seq_id, flags);
+    }
+    mem_recr->state_read(io, seq_id, flags);
+}
```
ggerganov, line 193: Here we should not check the `LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY` flag because the actual check is done in the `llama_kv_cache_iswa::state_read/write`:

```suggestion
void llama_memory_hybrid_iswa::state_write(llama_io_write_i & io, llama_seq_id seq_id, llama_state_seq_flags flags) const {
    mem_attn->state_write(io, seq_id, flags);
    mem_recr->state_write(io, seq_id, flags);
}

void llama_memory_hybrid_iswa::state_read(llama_io_read_i & io, llama_seq_id seq_id, llama_state_seq_flags flags) {
    mem_attn->state_read(io, seq_id, flags);
    mem_recr->state_read(io, seq_id, flags);
}
```

ggerganov (approved):
Rebase and lets wait for CI before merge

tdakhran:
Addressed the feedback, rebased, and tested with https://github.com/ggml-org/llama.cpp/pull/18641.
@ggerganov, @CISC, please let me know if anything else is needed to merge.

CISC (approved):

PR #18604, opened 2026-01-05
Fix grammar parsing issues to prevent stack overflow and hangs

aagit:
This pull request addresses some issues in the grammar parsing system that could lead to stack overflow and hangs when processing certain GBNF grammars. The fixes include:

1. **Stack overflow prevention**: Added cycle detection in `llama_grammar_advance_stack` to prevent infinite recursion when processing grammars with nullable symbols that could lead to infinite derivations of empty strings.

2. **Iterative implementation**: Converted the recursive `llama_grammar_advance_stack` function to an iterative approach using explicit stacks, which eliminates the risk of stack overflow from deep recursion.

3. **Repetition threshold checking**: Added a maximum repetition threshold to prevent excessive rule expansion during grammar parsing of deeply nested repetition patterns like `{m,n}`.

The repetition threshold value hasn't changed, but it now applies to all nested rules so it makes some valid grammar invalid, supposedly such previously valid grammars would hang or stack overflow.

## Testing

The changes have been tested with:
- Existing test suites (test-llama-grammar, test-grammar-integration, test-grammar-parser)
- The two llama-server curl reproducers mentioned in the commits
- Manual verification with some ripgrep-edit sessions with GBNF enabled

New test cases have been added to verify:
- The stack overflow case with `( [x]* )*` grammar is fixed
- The hang case with deeply nested repetition patterns is rejected


fiesh:
This fixes #19845

0cc4m:
@ggerganov This has been stuck for a while, can you take a look and let us know how to proceed?

ggerganov:
@pwilkin Would you like to take a look and review?

pwilkin:
@ggerganov Aye, can look.

pwilkin (changes requested):
Please run `editorchecker` and fix the indentation issues, otherwise looks fine. See also my changes to the `seen` and let me know if you approve.

aagit:
> Please run `editorchecker` and fix the indentation issues, otherwise looks fine. See also my changes to the `seen` and let me know if you approve.

Sure the set addition looks good.  I would have kept it incremental, but the UI seems to suggest to fold it, so I folded it into the patch, I don't mind either ways.  I also tried an unordered_set, but that requires building a key from all pointers in the stack vector and using test-grammar-integration as benchmark it was slower than the rbtree in the set. The set is also slower than the original linear vector but less (around 13% increase in runtime).

pwilkin:
"I also tried an unordered_set, but that requires building a key from all pointers in the stack vector and using test-grammar-integration as benchmark it was slower than the rbtree in the set."

Yeah tried that as well but it required too much setup with the hash function.

13% is fine if it helps us prevent catastrophic times with some very big grammars.

pwilkin (approved):

pwilkin:
Oh, I'm sorry, should've pinged me earlier :)

Issue #14909, opened 2025-07-28
Feature Request: Implement missing ops from backends


am17an:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

The list of supported operation in ggml per backend is maintained at https://github.com/ggml-org/llama.cpp/blob/master/docs/ops.md

How to contribute to make the list more complete:

1. Pick a op which is not implemented in a backend (pick ones marked with ❌), pick only one per PR
2. Implement the operation
3. Test your implementation with `test-backend-ops -o <op>`
4. If it passes test for the op, create a PR.

Notes:
1. The ops suffixed with _BACK is training code which is WIP and do not require immediate attention
2. Do not link this issue your PR, which is lead to closing this issue when your PR is merged. Instead, just mention it.

### Motivation

Feature parity between backends is always desirable and it's easy to test the implementation, making it an ideal good first issue.

jeemzz147:
I would like to contribute to this effort. May I ask whether it is okay to work on operators marked as [🟡 Partially supported] in the ops.md table?

am17an:
> I would like to contribute to this effort. May I ask whether it is okay to work on operators marked as [🟡 Partially supported] in the ops.md table?

Yes that would also work. The reason it's preferable to work on completely unsupported ops is because sometimes it's not possible to add full support to an op on a backend (e.g. some quantisation schemes in CUDA) and it might not be clear as a newcomer why that's the case. But if you make an operation support more complete, that would always be welcome!

jeemzz147:
I've verified that PAD_REFLECT_1D is currently unsupported on CUDA.
I would like to work on the PAD_REFLECT_1D op for the CUDA. Please let me know if there are any considerations or edge cases I should keep in mind. Thanks!

CISC:
> I've verified that PAD_REFLECT_1D is currently unsupported on CUDA.

It's already in progress in #14659

jeemzz147:
since `PAD_REFLECT_1D` is already in progress (#14659), I’d like to work on the `SET` op for the CUDA backend instead. Thanks!

[...]

Issue #7773, opened 2024-06-05
ggml : add WebGPU backend


ggerganov:
I hope that this would be relatively easy to do since AFAIK WebGPU allows us to write kernels in a shader language, so we have experience how to create such backends.

There has been some initial work in https://github.com/ggerganov/ggml/pull/585 - could be useful as a starting point

WenheLI:
Hi! I'm interested in bringing this backend to GGML and was wondering if there are any startup materials available for newcomers to quickly ramp up and start working on this backend?

ngxson:
So I've been playing with implementation of webgpu for a few days. I got a very minimal version with working buffer management and support for some simple ops.

My version is based on https://github.com/ggerganov/ggml/pull/585 but with some noticeable changes:
1. Up-to-date ggml backend API
2. Use webgpu_cpp instead of plain C (requires C++17)
3. emscripten-only for now

However, I'm not very familiar with ggml backend interface so I'm having a question:

I made a test cgraph to test my implementation: https://github.com/ngxson/ggml_webgpu_dev/blob/a5fcc25c359b997869b8683ab485d1d3f96b37f9/main.cpp#L70

When calling `ggml_gallocr_alloc_graph`, I expect it to call `buffer_type_alloc_buffer` with enough memory for all nodes, but turns out it only alloc memory for one node then call `init_tensor` for all nodes:

```
gml_backend_wgpu_buffer_type_alloc_buffer: 256  ==> only enough memory for one node
storage_buffer_1: create with size=256
ggml_backend_wgpu_buffer_reset
ggml_backend_wgpu_buffer_init_tensor: node_0
storage_buffer_1: node_0, init to offset 0
ggml_backend_wgpu_buffer_init_tensor: node_1
storage_buffer_1: realloc to size=256           ==> not enough memory, we need to realloc
storage_buffer_1: node_1, init to offset 256
ggml_backend_wgpu_buffer_init_tensor: node_2
```

Here is my `tensor_init` function: https://github.com/ngxson/ggml_webgpu_dev/blob/a5fcc25c359b997869b8683ab485d1d3f96b37f9/ggml-wgpu.cpp#L195  

@ggerganov  Could you help me understand this part? Thank you.

slaren:
[...]

PR #18605, opened 2026-01-05
ggml webgpu: add CEIL operation support

nwyin:
Add support for the CEIL unary operation in the WebGPU backend.

It seems like the ops support csv for WebGPU was a bit stale, so there are some additional diffs in the .csv beyond CEIL now being supported.

First contribution, so all feedback is appreciated!

Related: https://github.com/ggml-org/llama.cpp/issues/7773 https://github.com/ggml-org/llama.cpp/issues/14909


reeselevine:
thanks, looks good to me

reeselevine (approved):

PR #18464, opened 2025-12-29
examples : add debug utility/example

danbev:
This commit introduces a new example named llama-debug which is a utility that is intended to be used to assist with developing/debugging a converted model.

The motivation for this utilitiy is to assist in model conversion work to verify that the model produces the expected outputs. It is intended to replace logits.cpp in examples/model-conversion.

Example usage:
```console
./build/bin/llama-debug \
    -m models/Qwen2.5-0.5B-Instruct.gguf \
    --prompt "Hello, my name is" \
    --save-logits
...
Model add_bos: false
Input prompt: "Hello, my name is"
Token ids (5):
Hello(9707) ,(11)  my(847)  name(829)  is(374)
Data saved to data/llamacpp-Qwen2.5-0.5B-Instruct.bin
Data saved to data/llamacpp-Qwen2.5-0.5B-Instruct.txt
Prompt saved to data/llamacpp-Qwen2.5-0.5B-Instruct-prompt.txt
Tokens saved to data/llamacpp-Qwen2.5-0.5B-Instruct-tokens.bin
```

For more details about the options available for this example, please refer to examples/debug/README.md.

----
This was suggested/discussed in the following pr: https://github.com/ggml-org/llama.cpp/pull/18281#issuecomment-3681892592

PR #18607, opened 2026-01-05
model : add LFM2-ColBert-350M

tdakhran:
PR adds support for [LFM2-ColBert-350M](https://huggingface.co/LiquidAI/LFM2-ColBERT-350M) by introducing `n_embd_out` - a separate output embedding dimension that can differ from the input embedding dimension (n_embd).

Initially, I introduced `LLAMA_POOLING_TYPE_TOKEN`, which was applying `cls_out` and outputting all embedding, but then switched to `n_embd_out`.

`n_embd_out` will be used in future multimodal models as well.

New GGUF key and API:
  - `LLM_KV_EMBEDDING_LENGTH_OUT` - stores output embedding dimension
  - llama_model_n_embd_out() - returns `hparams.n_embd_out` if set and fallbacks to `hparams.n_embd`
  
 
 ### Testing
 Convert
 ```
 python convert_hf_to_gguf.py /data/playground/checkpoints/LFM2-ColBert-350M
 ```
 Launch server
 ```
 bin/llama-server -m /data/playground/checkpoints/LFM2-ColBert-350M/LFM2-ColBert-350M-F16.gguf --embeddings --pooling none
 ```
 
 Run the attached Python script
 ```
 ❯ uv run rerank.py
Score: 29.74 | Q: What is panda? | D: hi
Score: 29.90 | Q: What is panda? | D: it is a bear
Score: 30.52 | Q: What is panda? | D: The giant panda (Ailuropoda melanoleuca), sometimes called a panda bear or simply panda, is a bear species endemic to China.
 ```
 
[rerank.py](https://github.com/user-attachments/files/24426429/rerank.py)

cc: @ngxson


src/llama-model.cpp
```diff
@@ -627,6 +628,7 @@ void llama_model::load_hparams(llama_model_loader & ml) {
     ml.get_arr(LLM_KV_CLASSIFIER_OUTPUT_LABELS, classifier_labels, false);
     if (!classifier_labels.empty()) {
         hparams.n_cls_out = classifier_labels.size();
+        hparams.n_embd_out = classifier_labels.size();
     }
```
CISC, line 631: This crashes f.ex. `jina-bert-v2`, as observed in CIs because `n_cls_out` defaults to `1` for models that don't have any classifier labels.

tdakhran: thanks @CISC , rolled back to using `n_cls_out` for pooling = rank for now

CISC (approved):
LGTM, @ggerganov should look over `n_embd_out` changes.

convert_hf_to_gguf.py
```diff
@@ -9948,6 +9948,31 @@ def _is_audio_tensor(self, name: str):
...
 
 
+@ModelBase.register("Lfm2Model")
+class LFM2ColBertModel(LFM2Model):
+    model_arch = gguf.MODEL_ARCH.LFM2
+    dense_tensor_name = "dense_2"
+
+    def set_vocab(self):
+        super().set_vocab()
+        self.gguf_writer.add_add_bos_token(False)
```
CISC: Why set this to `False` BTW? It's `true` in the config, and it's used in `TemplateProcessing` both for `single` and `pair`.

CISC: So, if the config is correct both `add_add_bos_token` and `add_add_sep_token` should be `True`, and sep token should be set to bos token.

Ideally this should be automatically done by `SpecialVocab`s `TemplateProcessing`, but IIRC this pattern isn't accepted (a warning is logged), don't remember exactly why.

ggerganov (approved):
Should update `examples/embedding` and` `examples/model-conversion` to use the new `llama_model_n_embd_out()`.

convert_hf_to_gguf.py
tdakhran: nice catch, that's a debug leftover (was dealing with double BOS), will remove.

tdakhran:
Thanks for the feedback. Addressed the comment.
Attaching updated usage script for reference that sends tokens instead of strings to avoid double BOS, will push it later into the GGUF repo.
[rerank.py](https://github.com/user-attachments/files/24432775/rerank.py)

convert_hf_to_gguf.py
CISC: I think you still need to add sep metadata as mentioned for reranking to work.

convert_hf_to_gguf.py
```diff
@@ -9948,6 +9948,27 @@ def _is_audio_tensor(self, name: str):
...
+            name = "model." + name
+
+        return super().modify_tensors(data_torch, name, bid)
+
+    def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
+        # dense tensor is stored in a separate safetensors file
+        from safetensors.torch import load_file
+        tensors_file = self.dir_model / "1_Dense" / "model.safetensors"
+        assert tensors_file.is_file()
+        tensor = load_file(tensors_file)["linear.weight"]
```
ngxson, line 9967: Not a change request, but I'm wondering if we should introduce an `extra_model_files()` API in near future @CISC @compilade ?

Something like this:

```py
class LFM2ColBertModel(LFM2Model):
    def extra_model_dir():
        return [self.dir_model / "1_Dense" / "model.safetensors"]
# tensors will be loaded and processed via `modify_tensors()`
```

convert_hf_to_gguf.py
tdakhran: ColBERT doesn't need a sep token, it embeds queries and documents separately, and then the similarity score is calculated pairwise using maxsim. See the script attached here https://github.com/ggml-org/llama.cpp/pull/18607#issuecomment-3710149796.

This allows embedding documents once, caching document embeddings in the database, and then embedding only a query and computing a similarity score against precomputed document embeddings.

tdakhran: I was thinking of adding this logic to `llama.cpp`, but then it will require document database management, and I decided to leave it to the client.

ngxson (approved):

convert_hf_to_gguf.py
CISC: Ah, so `TemplateProcessing` is not used at all.

convert_hf_to_gguf.py
CISC: Yep, would be handy to deduplicate some code, I guess we'll see more of this as more `ST` support gets added.

I checked out some other ColBERTv2 models BTW, and they seem to have `linear.weight` embedded in main file.

danbev (approved):
I'll take a look at updating the model-conversion example as a separate PR if that is alright. I'm currently working on https://github.com/ggml-org/llama.cpp/pull/18464 and I can either update it or do this as a separate PR. 

I can also look at updating llama-embedding as a separate PR though it looks like it only requires a small update:
```console
diff --git a/examples/embedding/embedding.cpp b/examples/embedding/embedding.cpp
index 81111e81b..89e626353 100644
--- a/examples/embedding/embedding.cpp
+++ b/examples/embedding/embedding.cpp
@@ -252,7 +252,7 @@ int main(int argc, char ** argv) {
     }
 
     // allocate output
-    const int n_embd = llama_model_n_embd(model);
+    const int n_embd = llama_model_n_embd_out(model);
     std::vector<float> embeddings(n_embd_count * n_embd, 0);
     float * emb = embeddings.data();
```

tdakhran:
> I'll take a look at updating the model-conversion example as a separate PR if that is alright. I'm currently working on #18464 and I can either update it or do this as a separate PR.
> 
> I can also look at updating llama-embedding as a separate PR though it looks like it only requires a small update:
> 
> ```
> diff --git a/examples/embedding/embedding.cpp b/examples/embedding/embedding.cpp
> index 81111e81b..89e626353 100644
> --- a/examples/embedding/embedding.cpp
> +++ b/examples/embedding/embedding.cpp
> @@ -252,7 +252,7 @@ int main(int argc, char ** argv) {
>      }
>  
>      // allocate output
> -    const int n_embd = llama_model_n_embd(model);
> +    const int n_embd = llama_model_n_embd_out(model);
>      std::vector<float> embeddings(n_embd_count * n_embd, 0);
>      float * emb = embeddings.data();
> ```

@danbev I updated them in this commit https://github.com/ggml-org/llama.cpp/pull/18607/commits/98d7da6b08e413902028d2366a9716c25e8e646d, can revert if you prefer to do it in a separate PR.

danbev:
> @danbev I updated them in this commit [98d7da6](https://github.com/ggml-org/llama.cpp/commit/98d7da6b08e413902028d2366a9716c25e8e646d), can revert if you prefer to do it in a separate PR.

No, this is great! I just missed that commit. 

I was able to verify the converted model using the `embedding-verify-logits-st` target but it required some modifications to the scripts and how 'sentence_transformer` (st) models are handled. I need to think about how we should handle this as currently pooling is used for the 'st' target/option as that was the use case we had prior to this model. I'll look into this and follow up in a separate PR.

tdakhran:
For visibility, GGUFs and usage instructions are uploaded to https://huggingface.co/LiquidAI/LFM2-ColBERT-350M-GGUF

PR #18610, opened 2026-01-05
ggml webgpu: initial flashattention implementation

reeselevine:
This PR adds an initial version of FlashAttention2 in WebGPU. Along with the GPU code itself, this PR also adds a new preprocessor for WGSL shaders that should make it easier/less brittle to define new shaders going forwards. Details below:

## Shader setup

* Most of the shaders right now are generated at build time using a relatively hacky Python script and template syntax that I wrote when doing initial development for the WebGPU backend. However, this probably won't continue to scale very well, especially with the number of options for FlashAttention, so I decided it was time to use a more general solution. However, there wasn't an existing preprocessor for WGSL that would work with C++ code, so I wrote one here: https://github.com/reeselevine/pre-wgsl. The preprocessor itself is one file, `pre_wgsl.hpp`, and should continue to track any changes/features added to the main preprocessor repository.
* To accommodate the various options for FlashAttention when compiling WGSL shaders, I added another new file to the WebGPU backend, `ggml-webgpu-shader-lib.hpp`, which generates the shader using a combination of structural parameters, e.g., head sizes, and performance parameters, like KV tile sizes. My idea is to expand this library approach use more sophisticated strategies going forwards and move other shaders over to use the preprocessor. From a performance perspective, shaders compile pretty fast, are mostly fixed for a given model, and are cached, so JIT compilation, at least for FlashAttention, seems to be the right call in my opinion. 
* Some other minor changes to `ggml-webgpu.cpp` to handle the new FlashAttention code and JIT compilation.

## FlashAttention shader itself

* For the most part this follows the FlashAttention2 paper, with a change for online softmax to make it subgroup-size agnostic. It also uses global KV loads if sizes are nicely divisible, since the KV tiles are not reused and pre-loading into shared memory really slows things down (at least on my M3).
* Even then, performance is not great right now (< 50% of the same Metal code). My testing shows that a lot of this slowdown basically boils down to the initial Q * K^T accumulation loop. I still need to do some debugging here, to figure out if the issue is something I'm doing wrong structurally, or if it's in the compilation from WGSL to Metal. This code also needs to be more thoroughly tested on other platforms. Perhaps someone who has written FlashAttention for one of the other backends could take a look at the shader and see if it looks reasonable, e.g., @jeffbolznv?
* Otherwise, this passes all the backend tests on my machine, so I think it's in a good state to merge as an initial implementation that can be improved upon as time goes on.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -152,13 +152,13 @@ jobs:
           DAWN_VERSION="v2.0.0"
           DAWN_OWNER="reeselevine"
           DAWN_REPO="dawn"
-          DAWN_ASSET_NAME="Dawn-5e9a4865b1635796ccc77dd30057f2b4002a1355-macos-latest-Release.zip"
-          echo "Fetching release asset from https://github.com/${DAWN_OWNER}/${DAWN_REPO}/releases/download/${DAWN_VERSION}/${DAWN_ASSET_NAME}"
+          DAWN_ASSET_NAME="Dawn-5e9a4865b1635796ccc77dd30057f2b4002a1355-macos-latest-Release"
+          echo "Fetching release asset from https://github.com/${DAWN_OWNER}/${DAWN_REPO}/releases/download/${DAWN_VERSION}/${DAWN_ASSET_NAME}.zip"
           curl -L -o artifact.zip \
-            "https://github.com/${DAWN_OWNER}/${DAWN_REPO}/releases/download/${DAWN_VERSION}/${DAWN_ASSET_NAME}"
+            "https://github.com/${DAWN_OWNER}/${DAWN_REPO}/releases/download/${DAWN_VERSION}/${DAWN_ASSET_NAME}.zip"
           mkdir dawn
           unzip artifact.zip
-          tar -xvf Dawn-5e9a4865b1635796ccc77dd30057f2b4002a1355-macos-latest-Release.tar.gz -C dawn --strip-components=1
+          tar -xvf ${DAWN_ASSET_NAME}.tar.gz -C dawn --strip-components=1
 
       - name: Build
         id: cmake_build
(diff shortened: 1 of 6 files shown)


jeffbolznv:
Hi, I only skimmed it and the basic structure looks reasonable, but I have no experience tuning for Apple/Metal. For NVIDIA, the current matmul size of 8x8x8 isn't supported and the tile/workgroup size is probably smaller than you'd want.

reeselevine:
Thanks Jeff! Yeah the `#define`s in the shader code itself are just defaults, they can be overridden when processing the shader, and depending on the device's supported subgroup matrix sizes. Finding the right values and building a selection process into the code is one of the next steps for sure on this shader!

reeselevine:
Also, since the Github-hosted runners don't support the features (subgroup matrices in particular) that are necessary to actually run the WebGPU flashattention implementation, maybe it would be useful to add a ggml-ci node to run the tests too, if it can support the necessary features? I'm not sure exactly how to do this, I see where the ggml-ci nodes are defined in `build.yml`, but I don't see how the environment is set up to make sure things run (specifically installing the Dawn WebGPU runtime on the machine, like I do right now for the Github-hosted runners).

ggerganov:
For the CI, we can add a workflow that executes on the Mac self-hosted runner, such as [this](https://github.com/ggml-org/llama.cpp/blob/1871f0ba56e57826c1c630c5f57274624d68788e/.github/workflows/build.yml#L1694-L1707). It's a Mac M4 Mini machine - would it have the necessary features?

> but I don't see how the environment is set up to make sure things run (specifically installing the Dawn WebGPU runtime on the machine, like I do right now for the Github-hosted runners).

I can pre-install Dawn and the workflow will just assume it is installed at some path? A better option is to add the same [Dawn installation step](https://github.com/ggml-org/llama.cpp/blob/1871f0ba56e57826c1c630c5f57274624d68788e/.github/workflows/build.yml#L148-L162) to the new self-hosted workflow and then pass the necessary path(s) to the `ci/run.sh` script via new `GG_BUILD_...` vars. Let me know if this makes sense.

reeselevine:
@ggerganov added a ggml-ci workflow, looks like everything is passing now. The Mac M4 Mini does look like it supports the necessary features, and the full CI actually did help me fix a precision issue in the flash attention code.

ggerganov:
Alright. If you want, you can also add a webgpu workflow on the seff-hosted CUDA runner:

https://github.com/ggml-org/llama.cpp/blob/9c142e3a2a8f1c7415511bd9d24f4790ce2dac88/.github/workflows/build.yml#L1612-L1614

It uses a Tesla T4 GPU. We can also do it later in another PR.

ggerganov:
> Even then, performance is not great right now (< 50% of the same Metal code).

Is this valid for both batch size = 1 and batch size > 1? In the Metal backend, we have 2 separate kernels for flash attention: one for small batch size (a.k.a the vec kernel) and one for larger batch sizes. While here you have the same kernel for both cases.

ggerganov (approved):

CISC (approved):

reeselevine:
> Alright. If you want, you can also add a webgpu workflow on the seff-hosted CUDA runner:
> 
> https://github.com/ggml-org/llama.cpp/blob/9c142e3a2a8f1c7415511bd9d24f4790ce2dac88/.github/workflows/build.yml#L1612-L1614
> 
> It uses a Tesla T4 GPU. We can also do it later in another PR.

sounds good, I will plan on adding this once I get a chance to run/test things on an NVIDIA GPU. In particular I'd like to extend the code to support more tensor core sizes.

> > Even then, performance is not great right now (< 50% of the same Metal code).
> 
> Is this valid for both batch size = 1 and batch size > 1? In the Metal backend, we have 2 separate kernels for flash attention: one for small batch size (a.k.a the vec kernel) and one for larger batch sizes. While here you have the same kernel for both cases.

I was mostly testing on batch size > 1. I agree we'll eventually want a separate kernel for the vec path for increased parallelism. It's looking like flashattention in general is a good stress test of WebGPU's capabilities/efficiency, we'll keep working on it.

PR #10071, opened 2024-10-28
llama : remove Tail-Free sampling

ggerganov:
Supersedes #9604

PR #18612, opened 2026-01-05
sampling: add tail-free (TFS) sampling

viralvgupta:
### What
This PR adds Tail-Free Sampling (TFS), a curvature-based truncation method
that removes long-tail tokens more adaptively than top-p.

### Why
Top-p truncates by cumulative probability mass, which can retain low-quality
long-tail tokens in flatter distributions. Tail-free sampling operates on the
curvature of the sorted probability curve, behaving deterministically in
peaked regimes while allowing controlled diversity in flatter ones.

### Details
- CPU tail-free sampler implementation
- CLI flag: `--tfs <z>` (disabled by default)
- GGUF metadata support via `general.sampling.tfs_z`
- Integrated into the sampler chain as a safe no-op when `z = 0.0`
- Default sampling behavior unchanged

### Testing
- Verified text generation on Apple Silicon (Metal backend for compute)
- Compared against top-p on list, reasoning, and creative prompts
- Confirmed no crashes with sampler enabled or disabled

### Notes
The sampler currently runs via the CPU sampler path.
Backend (Metal/CUDA) support can be added in a follow-up if desired.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -1472,6 +1472,14 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_sparam());
     add_opt(common_arg(
+        {"--tfs", "--tail-free"}, "Z",
+        string_format("tail-free sampling (default: %.2f, 0.0 = disabled)", (double) params.sampling.tfs_z),
+        [](common_params & params, const std::string & value) {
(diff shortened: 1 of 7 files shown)


Green-Sky:
This is unlikely to get merged, see https://github.com/ggml-org/llama.cpp/pull/10071

Issue #17610, opened 2025-11-30
Eval bug: Granite-4.0-h-1B fails long context (>16k); same model in Transformers works, Granite-4.0-h-MicroGGUF works)`


mramendi:
### Name and Version

```
$ llama-server --version
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 5090, compute capability 12.0, VMM: yes
version: 7122 (21d31e081)
built with cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0 for x86_64-linux-gnu

$ llama-cli --version
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 5090, compute capability 12.0, VMM: yes
version: 7122 (21d31e081)
built with cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0 for x86_64-linux-gnu
root@C.28352397:~/models$ 
```

### Operating systems

Linux

### GGML backends

CUDA
(does NOT happen on CPU)

### Hardware

RTX 5090

### Models

IBM Granite 4.0-h-1b, several quants (including BF16 GGUF)

### Problem description & steps to reproduce

When I infer at a 32k context (simple script generating a NIAH text), the model returns "???????" or "G".

The same script at 16k context works wine with the model.

The same script with the same llama.cpp and with Granite-4-h-micro (3.5B) works fine at 32k context. The same script with a CPU build of llama.cpp and Granite-4-h-1b works fine at 32k context too.

Similar content, inferring on Transformers with Granite-4-h-1b (original safetensors version) works.

### Relevant log output

```shell
(not sure what to paste - no error message is displayed)
```

pwilkin:
@gabe-l-hart any ideas?

gabe-l-hart:
Hmm, when I've seen `???????` output from other models before, it's been due to floating point precision overflows, but this post says it's happening with BF16 which should have the same range as F32 and shouldn't show those issues. Is it possible to share the exact script being used, including the specific GGUF model (and how it was converted if done manually)?

mramendi:
[...]

PR #18614, opened 2026-01-05
CUDA: fix FA FP16 accumulator overflow for Granite

### LINKED ISSUE #18606: Eval bug: granite 4-h small (Q6/Q8) and micro (Q8/BF16) fails on very large context (100k/200k)
### Name and Version

$ llama-cli --version                                                                                              │
ggml_cuda_init: found 1 CUDA devices:                                                                                                      │
  Device 0: NVIDIA A100-PCIE-40GB, compute capability 8.0, VMM: yes                                                                        │
version: 7621 (e57f52334)                                                                                                                  │

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

Nvidia A100 40Gb (via vast.ai)

### Models

IBM Granite 4-h Small https://huggingface.co/ibm-granite/granite-4.0-h-small-GGUF

IBM Granite 4-h Micro https://huggingface.co/ibm-granite/granite-4.0-h-micro-GGUF

(Might apply to other granite 4-h models - I only checked two)

In a deja vu of #17610 I get ??????? output on a simple NIAH script:

[niah.py](https://github.com/user-attachments/files/24425884/niah.py)

The script tries lengths of 50000, 100000, and 200000 tokens (approximate). 50000 always succeeds.

When I run:

llama-server -m models/granite-4.0-h-small-Q8_0.gguf -c 256000 -ngl 99 

or:

llama-server -m models/granite-4.0-h-small-Q6_K.gguf -c 256000 -ngl 99 

100000 also succeeds, but 200000 fails with ??????

However, when I run:

llama-server -m models/granite-4.0-h-small-Q6_K.gguf -c 500000 -ngl 99 

I get a failure at length 100000 with ???????  - nothing is changed except a larger context window, which is not actually used! (I can't check this with small-Q8_0 - not enough VRAM).

When I run:

llama-server -m models/granite-4.0-h-micro-Q8_0.gguf -c 256000 -ngl 99 

I also get a failure at length 100000 with ??????? 

With the Micro model I also have the VRAM to check with the BF16 version; IBM does not provide one so I used the one from Unsloth:

https://huggingface.co/unsloth/granite-4.0-h-micro-GGUF?show_file_info=granite-4.0-h-micro-BF16.gguf 

So I run:

llama-server -m models/granite-4.0-h-micro-BF16.gguf -c 256000 -ngl 99

And get the same issue - size 50000 succeeds, size 100000 fails with ????????

As in #17610, after a failure like this happens, the model starts giving ?????????? responses even to short requests, until llama-server is restarted. It would be really great to fix this not-resetting issue too this time if possible?

Another observation is that there seems to be a slowdown of processing, with GPU utilization jumping to <10% and back up sometimes, closer to the end of the processing of the failing sequence, but this might be unrelated.

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

JohannesGaessler:
Fixes https://github.com/ggml-org/llama.cpp/issues/18606 .

The problem seems to be the same as in https://github.com/ggml-org/llama.cpp/issues/17610 where extremal values in the V cache cause an overflow of the FP16 VKQ accumulators. It can be fixed by applying a more aggressive shift of the numerical range. On master it is being shifted by a factor of 2, with this PR it would increase to a factor of 8. The smallest representable value in the VKQ accumulators would then be $2^{-11}$ rather than $2^{-13}$.

### DIFF
--- a/ggml/src/ggml-cuda/fattn-common.cuh
+++ b/ggml/src/ggml-cuda/fattn-common.cuh
@@ -11,10 +11,12 @@
 #define SOFTMAX_FTZ_THRESHOLD -20.0f                   // Softmax exp. of values smaller than this are flushed to zero to avoid NaNs.
 
 // log(2) = 0.6931, by adding this to the KQ maximum used for the softmax the numerical range representable
-//     by the VKQ accumulators is effectively being shifted up by a factor of 8.
+//     by the VKQ accumulators is effectively being shifted up by a factor of 2.
 // This reduces issues with numerical overflow but also causes larger values to be flushed to zero.
 // However, as the output from FlashAttention will usually be used as an input for a matrix multiplication this should be negligible.
-#define FATTN_KQ_MAX_OFFSET 0.6931f
+// Still, the value range should be shifted as much as necessary but as little as possible.
+// The macro on the following line shifts it by a factor of 2**3=8, as was needed to fix https://github.com/ggml-org/llama.cpp/issues/18606 .
+#define FATTN_KQ_MAX_OFFSET (3.0f*0.6931f)
 
 typedef void (* fattn_kernel_t)(
         const char * __restrict__ Q,


ggerganov (approved):
The NIAH tests are passing now on DGX Spark:

```
$ python niah.py 
=== MAMBA QUANTIZATION DRIFT TEST (FIXED) ===

⚡ Testing Context Length: ~50000 tokens
   ⏱️ Time: 53.76s
   📍 Needle Depth: 54%
   🤖 Model Answer: 'The special passcode to unlock the Vault of Glass is 74921.'
   ✅ PASS: Needle Found

⚡ Testing Context Length: ~100000 tokens
   ⏱️ Time: 110.64s
   📍 Needle Depth: 64%
   🤖 Model Answer: 'The special passcode to unlock the Vault of Glass is 74921.'
   ✅ PASS: Needle Found

⚡ Testing Context Length: ~200000 tokens
   ⏱️ Time: 236.24s
   📍 Needle Depth: 53%
   🤖 Model Answer: 'The special passcode to unlock the Vault of Glass is 74921.'
   ✅ PASS: Needle Found

=== FINAL SUMMARY ===
{50000: 'PASS', 100000: 'PASS', 200000: 'PASS'}
```

broadbit-hu:
After the b7638 release, the Mistral NeMo outputs changed significantly, whereas there was no issue with the b7636 version.
```
#define FATTN_KQ_MAX_OFFSET 0.6931f
```
vs
```
#define FATTN_KQ_MAX_OFFSET (3.0f*0.6931f)
```

Command line arguments:
```
build/bin/llama-server -m mistral-nemo-instruct-2407-q8_0.gguf -c 75008 -ngl 99 \
  --temp 0 --seed 1 --parallel 1 --flash-attn on --kv-unified \
  --fit off \
  --host 0.0.0.0 --port 18080 --no-warmup \
  --no-context-shift --slot-prompt-similarity 0.0 \
  --cache-ram 0 --keep 0 --no-perf \
  --alias mistral-nemo-instruct-2407-q8_0.gguf
```

JohannesGaessler:
It's expected that the outputs are not going to be bit-for-bit identical and that individual tokens can be sampled differently. Did you observe that the outputs became worse in a statistically significant way?

broadbit-hu:
Yes, unfortunately, the results became significantly worse. I tested the b7638 and newer versions on Metal GPU (M2 Ultra), where these degradations do not occur—the results are very similar to those of the b7636 and older versions on ROCm GPUs.

I modified the FATTN_KQ_MAX_OFFSET value, so the difference is no longer significant, and there is no significant degradation or improvement in our testcases:
```cpp
// #define FATTN_KQ_MAX_OFFSET (3.0f*0.6931f)
define FATTN_KQ_MAX_OFFSET (2.0f*0.6931f)
```

JohannesGaessler:
To be clear, I'm not asking about "significant" in the colloquial sense, I'm asking about statistical significance. What I want to see is hard evidence with a large sample size. If I compare how the KQ max offset affects the Nemo token probabilities on Wikitext I get this:

| Changes vs. FP16 + no KQ max offset | KL divergence       | Δp RMS          |
|-------------------------------------|---------------------|-----------------|
| KQ max offset == 0.01f*0.6931f      | 0.000510 ± 0.000002 | 0.795 ± 0.004 % |
| KQ max offset == 1.0f*0.6931f       | 0.000510 ± 0.000002 | 0.796 ± 0.004 % |
| KQ max offset == 3.0f*0.6931f       | 0.000516 ± 0.000002 | 0.804 ± 0.004 % |
| KQ max offset == 10.0f*0.6931f      | 0.000766 ± 0.000003 | 0.946 ± 0.005 % |
| q8_0 weights                        | 0.001372 ± 0.000007 | 1.236 ± 0.008 % |
| q8_0 KV                             | 0.000684 ± 0.000003 | 0.910 ± 0.004 % |

For offsets in the range of what was actually implemented the changes to the logits are already at the limit of what can be measured with `llama-perplexity` due to the 16 bit compression on disk. Even if you crank the offset up and shift the numerical range by a factor of $2^{10} = 1024$ that only ends up being about as impactful as q8_0 quantization in terms of how much the logits change. And as I said: it's expected that the results will not be bit-for-bit identical.

broadbit-hu:
Thank you for the detailed response. It seems we need a more comprehensive test on our side...

PR #18544, opened 2026-01-02
add intrinsics include

svalaskevicius:
https://github.com/ggml-org/llama.cpp/releases/tag/b7557 added zen4 cpu flags and builds for this started failing because:

```
-- The C compiler identification is GNU 15.2.1
/home/sarunas/src/aur/llama.cpp-vulkan/src/llama.cpp/ggml/src/ggml.c: In function ‘ggml_fp32_to_bf16_row’:
/home/sarunas/src/aur/llama.cpp-vulkan/src/llama.cpp/ggml/src/ggml.c:482:9: error: implicit declaration of function ‘_mm512_storeu_si512’ [-Wimplicit-function-declaration]
  482 |         _mm512_storeu_si512(
      |         ^~~~~~~~~~~~~~~~~~~
/home/sarunas/src/aur/llama.cpp-vulkan/src/llama.cpp/ggml/src/ggml.c:483:14: error: ‘__m512i’ undeclared (first use in this function); did you mean ‘m512i’?
  483 |             (__m512i *)(y + i),
      |              ^~~~~~~
      |              m512i
/home/sarunas/src/aur/llama.cpp-vulkan/src/llama.cpp/ggml/src/ggml.c:483:14: note: each undeclared identifier is reported only once for each function it appears in
/home/sarunas/src/aur/llama.cpp-vulkan/src/llama.cpp/ggml/src/ggml.c:483:23: error: expected expression before ‘)’ token
  483 |             (__m512i *)(y + i),
      |                       ^
/home/sarunas/src/aur/llama.cpp-vulkan/src/llama.cpp/ggml/src/ggml.c:484:19: error: implicit declaration of function ‘_mm512_cvtne2ps_pbh’ [-Wimplicit-function-declaration]
  484 |             m512i(_mm512_cvtne2ps_pbh(_mm512_loadu_ps(x + i + 16),
      |                   ^~~~~~~~~~~~~~~~~~~
/home/sarunas/src/aur/llama.cpp-vulkan/src/llama.cpp/ggml/src/ggml.c:61:28: note: in definition of macro ‘m512i’
   61 | #define m512i(p) (__m512i)(p)
      |                            ^
/home/sarunas/src/aur/llama.cpp-vulkan/src/llama.cpp/ggml/src/ggml.c:484:39: error: implicit declaration of function ‘_mm512_loadu_ps’ [-Wimplicit-function-declaration]
[...]

PR #18623, opened 2026-01-05
ggml : fix avx512bf16 build

angt:
- include `immintrin.h` when required
- remove unused m512bh


ggerganov (approved):

svalaskevicius:
fyi - I also had this created for the same problem  - #18544

PR #8032, opened 2024-06-20
rpc : copy tensors across servers

rgerganov:
This is an attempt to make copying tensors across servers more efficient. It introduces 2 new RPC commands:
 * `HELLO` - send after establishing connection to identify the remote party (client or server)
 * `REMOTE_COPY_TENSOR` - send to the host which contains the source tensor along with the destination tensor and destination endpoint

```mermaid
sequenceDiagram
    Note over Scheduler: Copy X on Server A to Y on Server B
    Scheduler->>Server A: REMOTE_COPY_TENSOR
    Server A->>Server B: HELLO
    Server A->>Server B: SET_TENSOR
    Server B-->>Server A: 
    Server A-->>Scheduler: 
```

- [x] I have read the [contributing guidelines](https://github.com/ggerganov/llama.cpp/blob/master/CONTRIBUTING.md)
- Self-reported review complexity:
  - [ ] Low
  - [x] Medium
  - [ ] High

PR #16753, opened 2025-10-24
rpc: use `XXHash64` instead of `FNV-1a` for hashing tensors

jukofyork:
This PR replaces the very slow [FNV-1a](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV-1a_hash) hash (which processes each byte separately), with [XXHash64](https://github.com/Cyan4973/xxHash) which processes blocks of 4x 8-bytes (64bit words) in a loop that the compiler should be able to unroll to process the bulk of the data in 32-byte blocks:

```cpp
const uint64_t* block = reinterpret_cast<const uint64_t*>(p);
for (int i = 0; i < 4; ++i) {
    uint64_t v = state[i] + block[i] * prime2;
    state[i] = ((v << 31) | (v >> (64 - 31))) * prime1;
}
p += 32;
```

**NOTE**: I did try manually unrolling the loop:

<details>
  <summary>Spolier</summary>
    
  ```cpp
        const uint64_t* w       = reinterpret_cast<const uint64_t*>(p);
        const uint64_t* wLimit  = reinterpret_cast<const uint64_t*>(end - 32); // last start for a full 32-byte stripe
        for (; w <= wLimit; w += 4) {
            // unrolled lanes
            uint64_t v0 = state[0] + w[0] * prime2;
            state[0] = ((v0 << 31) | (v0 >> (64 - 31))) * prime1;

            uint64_t v1 = state[1] + w[1] * prime2;
            state[1] = ((v1 << 31) | (v1 >> (64 - 31))) * prime1;

            uint64_t v2 = state[2] + w[2] * prime2;
            state[2] = ((v2 << 31) | (v2 >> (64 - 31))) * prime1;

            uint64_t v3 = state[3] + w[3] * prime2;
            state[3] = ((v3 << 31) | (v3 >> (64 - 31))) * prime1;
        }
        // advance byte pointer by the processed stripes (32 bytes per iteration)
        p = reinterpret_cast<const uint8_t*>(w);
  ```
  
but it didn't seem to help.

</details>

---

I haven't done any benchmarking, but it is clearly *many* times faster and in line with the [reported benchmarks](https://github.com/Cyan4973/xxHash?tab=readme-ov-file#benchmarks):

image

---

A few notes:

- It assumes all will use the same [Endianness](https://en.wikipedia.org/wiki/Endianness).
- It assumes the input `uint8_t* data` is aligned to 8-byte boundaries.
[...]

PR #16276, opened 2025-09-26
rpc : add support for multiple devices

rgerganov:
Allow rpc-server to expose multiple devices from a single endpoint. Change RPC protocol to include device identifier where needed. Add new API to get the device count from an RPC endpoint.

closes: #15210

PR #18626, opened 2026-01-05
rpc : implement event and async backend APIs

rgerganov:
This patch implements the event and async APIs from the backend interface. The intent is to enable pipeline parallelism with the RPC backend.

Here is a summary of the changes:
 * Introduce `rpc_dispatcher` which allows sending commands both synchronously and asynchronously to an RPC endpoint. There is one dispatcher per RPC endpoint and every dispatcher is using a background thread for its network communication.
 * All RPC commands are allocated on the heap and passed to `rpc_dispatcher` with `shared_ptr`s to avoid making copies of the input data.
 * Creating an event adds a dummy message in the dispatcher's queue and synchronizing the event waits for this dummy message to be processed by the background thread. Not sure if this is equivalent to how CUDA events work which seem to be the inspiration of the backend API.
 * Changed `RPC_CMD_FREE_BUFFER`, `RPC_CMD_BUFFER_CLEAR`, `RPC_CMD_INIT_TENSOR` to not send any response which simplifies the logic in the dispatcher. Hence the protocol version bump (3.7.0).

TODO:
 * Verify event implementation; Are `event_synchronize()` and `event_wait()` equivalent here?
 * Implement `cpy_tensor_async()`. This is a hard one as RPC servers need to to talk to each other (see PR #8032). Pipeline parallelism relies on both `graph_compute()` and `cpy_tensor_async()` being async and I am not sure if we are going to see improvements only with the former.
 * Split RPC backend into several files (e.g. client, server, common). This is left for a future refactoring PR with no functional changes.

@ggerganov I will appreciate some early feedback on this.

### DIFF
--- a/ggml/src/ggml-rpc/ggml-rpc.cpp
+++ b/ggml/src/ggml-rpc/ggml-rpc.cpp
@@ -245,9 +245,8 @@ struct ggml_backend_rpc_buffer_context {
 // RPC helper functions
 
 // Computes FNV-1a hash of the data
-static uint64_t fnv_hash(const uint8_t * data, size_t len) {
+static uint64_t fnv_hash(const uint8_t * data, size_t len, uint64_t hash = 0xcbf29ce484222325ULL) {
     const uint64_t fnv_prime = 0x100000001b3ULL;
-    uint64_t hash = 0xcbf29ce484222325ULL;
 
     for (size_t i = 0; i < len; ++i) {
         hash ^= data[i];
@@ -831,7 +830,45 @@ static size_t ggml_backend_rpc_buffer_type_get_alloc_size(ggml_backend_buffer_ty
 
     if (rpc_get) {
         ggml_backend_rpc_buffer_type_context * buft_ctx = (ggml_backend_rpc_buffer_type_context *)buft->context;
-        auto dispatcher = get_dispatcher(buft_ctx->endpoint);
+
+        // Cache key for calls to read the alloc_size.
+        // We deliberately exclude src tensor dimensions from the key because:
+        // 1. For CPU backends, alloc_size = ggml_nbytes(output) regardless of src shapes
+        // 2. For GPU backends, the reservation graph uses max dimensions, so the
+        //    cached value from reservation is always >= any subsequent request
+        // 3. Including src dims causes cache misses per-ubatch (e.g. growing KV cache)
+        //    which blocks the main thread behind in-flight GRAPH_COMPUTE commands
+        struct alloc_size_cache_key {
+            uint32_t device;
+            uint32_t type;
+            uint32_t op;
+            int32_t  op_params[GGML_MAX_OP_PARAMS / sizeof(int32_t)];
+            uint32_t ne[GGML_MAX_DIMS];
+        };
+
+        alloc_size_cache_key key = {};
+        key.device = buft_ctx->device;
+        key.type = tensor->type;
+        key.op = tensor->op;
+        memcpy(key.op_params, tensor->op_params, sizeof(key.op_params));
+        for (int i = 0; i < GGML_MAX_DIMS; i++) {
+            key.ne[i] = (uint32_t)tensor->ne[i];
+        }
+
+        uint64_t cache_hash = fnv_hash((const uint8_t *)&key, sizeof(key));
+        cache_hash = fnv_hash((const uint8_t *)buft_ctx->endpoint.data(), buft_ctx->endpoint.size(), cache_hash);
+
+        // alloc sizes are immutable for a given tensor configuration
+        static std::mutex cache_mutex;
+        static std::unordered_map<uint64_t, size_t> cache;
+
+        {
+            std::lock_guard<std::mutex> lock(cache_mutex);
+            auto it = cache.find(cache_hash);
+            if (it != cache.end()) {
+                return it->second;
+            }
+        }
 
         auto request = std::make_shared<rpc_msg_get_alloc_size_req>();
         request->device = buft_ctx->device;
@@ -842,9 +879,14 @@ static size_t ggml_backend_rpc_buffer_type_get_alloc_size(ggml_backend_buffer_ty
             request->srcs[i] = serialize_tensor(tensor->src[i]);
         }
 
-        // TODO: cache the alloc responses to avoid extra RPC calls?
         rpc_msg_get_alloc_size_rsp response;
+        auto dispatcher = get_dispatcher(buft_ctx->endpoint);
         dispatcher->send(RPC_CMD_GET_ALLOC_SIZE, request, sizeof(*request), &response, sizeof(response));
+
+        {
+            std::lock_guard<std::mutex> lock(cache_mutex);
+            cache[cache_hash] = response.alloc_size;
+        }
 
         return response.alloc_size;
     }
--- a/ggml/src/ggml-rpc/ggml-rpc.cpp
+++ b/ggml/src/ggml-rpc/ggml-rpc.cpp
@@ -9,6 +9,9 @@
 #include <optional>
 #include <string>
 #include <vector>
+#include <queue>
+#include <condition_variable>
+#include <future>
 #include <memory>
 #include <mutex>
 #include <unordered_map>
@@ -17,6 +20,8 @@
 #include <fstream>
 #include <filesystem>
 #include <algorithm>
+#include <atomic>
+#include <thread>
 
 static const char * RPC_DEBUG = std::getenv("GGML_RPC_DEBUG");
 
@@ -72,6 +77,7 @@ enum rpc_cmd {
     RPC_CMD_DEVICE_COUNT,
     RPC_CMD_GRAPH_RECOMPUTE,
     RPC_CMD_MEMSET_TENSOR,
+    RPC_CMD_NONE,
     RPC_CMD_COUNT,
 };
 
@@ -223,16 +229,17 @@ struct ggml_backend_rpc_buffer_type_context {
     size_t      max_size;
 };
 
+class rpc_dispatcher;
 struct ggml_backend_rpc_context {
-    std::string endpoint;
-    uint32_t    device;
-    std::string name;
+    std::shared_ptr<rpc_dispatcher> dispatcher;
+    uint32_t                        device;
+    std::string                     name;
 };
 
 struct ggml_backend_rpc_buffer_context {
-    std::shared_ptr<socket_t> sock;
-    void * base_ptr;
-    uint64_t remote_ptr;
+    std::shared_ptr<rpc_dispatcher>   dispatcher;
+    void                            * base_ptr;
+    uint64_t                          remote_ptr;
 };
 
 // RPC helper functions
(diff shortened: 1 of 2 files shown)


ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -495,67 +502,267 @@ static bool send_rpc_cmd(const std::shared_ptr<socket_t> & sock, enum rpc_cmd cm
...
+    void start(const std::string & endpoint);
+    void work();
+
+    ~rpc_dispatcher();
+
+private:
+    struct rpc_msg {
+        rpc_cmd                       cmd;
+        std::shared_ptr<const void>   input;
+        size_t                        input_size;
```
ggerganov, line 431: I think this can be simply `std::vector<uint8_t>`. Use move semantics to avoid copies.

rgerganov: this may work for `RPC_CMD_SET_TENSOR` and `RPC_CMD_GRAPH_COMPUTE` where the input is allocated in a `vector` but I don't see how this is going to work for the rest of the commands

ggerganov: I see - it's OK in that case.

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -871,36 +1108,43 @@ static enum ggml_status ggml_backend_rpc_graph_compute(ggml_backend_t backend, g
...
+static void ggml_backend_rpc_event_record(ggml_backend_t backend, ggml_backend_event_t event) {
+    ggml_backend_rpc_context * rpc_ctx = (ggml_backend_rpc_context *)backend->context;
+    rpc_ctx->dispatcher->event_record(event);
+}
+
+static void ggml_backend_rpc_event_wait(ggml_backend_t dev, ggml_backend_event_t event) {
+    ggml_backend_rpc_context * ctx = (ggml_backend_rpc_context *)dev->context;
+    ctx->dispatcher->event_synchronize(event);
+}
+
```
ggerganov, line 1040:
```suggestion

static void ggml_backend_rpc_event_wait(ggml_backend_t backend, ggml_backend_event_t event) {
    ggml_backend_rpc_context * rpc_ctx = (ggml_backend_rpc_context *)backend->context;
    rpc_ctx->dispatcher->event_synchronize(event);
}

```

ggerganov:
> Verify event implementation; Are event_synchronize() and event_wait() equivalent here?

I think we have the following analogy between CUDA and RPC:

- CUDA device -> RPC endpoint
- CUDA stream -> nothing

If this is correct, then I believe the `event_record()` and `event_wait()` API should never be utilized because the RPC backend does not have a notion of "stream". Is this correct?

ggml/src/ggml-rpc/ggml-rpc.cpp
rgerganov: fixed

rgerganov:
> If this is correct, then I believe the `event_record()` and `event_wait()` API should never be utilized because the RPC backend does not have a notion of "stream". Is this correct?

According to this [comment](https://github.com/ggml-org/llama.cpp/pull/16276#discussion_r2382464239) from @slaren, `ggml_backend_t` is supposed to represent a stream or queue and this is exactly how I implement the client side.

The server side is a bit tricky as we may create multiple `ggml_backend_t` from multiple devices and expose them through a single endpoint. But for now my goal is to prove that we can get PP improvement in the simple case when every endpoint exposes a single device with a single `ggml_backend_t`

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -1980,6 +2202,24 @@ static bool ggml_backend_rpc_device_supports_buft(ggml_backend_dev_t dev, ggml_b
...
+    ggml_backend_rpc_device_context * ctx = (ggml_backend_rpc_device_context *)dev->context;
+    auto dispatcher = get_dispatcher(ctx->endpoint.c_str());
+    dispatcher->event_free(event);
+}
+
+static void ggml_backend_rpc_device_event_synchronize(ggml_backend_dev_t dev, ggml_backend_event_t event) {
+    ggml_backend_rpc_device_context * ctx = (ggml_backend_rpc_device_context *)dev->context;
+    auto dispatcher = get_dispatcher(ctx->endpoint.c_str());
+    dispatcher->event_synchronize(event);
+}
```
ggerganov, line 2212:
```suggestion
static ggml_backend_event_t ggml_backend_rpc_device_event_new(ggml_backend_dev_t dev) {
    ggml_backend_rpc_device_context * ctx = (ggml_backend_rpc_device_context *)dev->context;
    auto dispatcher = get_dispatcher(ctx->endpoint);
    return dispatcher->event_new(dev);
}

static void ggml_backend_rpc_device_event_free(ggml_backend_dev_t dev, ggml_backend_event_t event) {
    ggml_backend_rpc_device_context * ctx = (ggml_backend_rpc_device_context *)dev->context;
    auto dispatcher = get_dispatcher(ctx->endpoint);
    dispatcher->event_free(event);
}

static void ggml_backend_rpc_device_event_synchronize(ggml_backend_dev_t dev, ggml_backend_event_t event) {
    ggml_backend_rpc_device_context * ctx = (ggml_backend_rpc_device_context *)dev->context;
    auto dispatcher = get_dispatcher(ctx->endpoint);
    dispatcher->event_synchronize(event);
}
```

ggerganov (commented):
AFAIU since all communication for a given endpoint is funneled through the same socket, then effectively the RPC backends for a given endpoint on the client are all associated with a single shared "stream" (i.e. the network socket). So if this is correct, then my guess would be that `event_wait()` would have to be a noop. Doing `dispatcher->event_synchronize()` would not make it incorrect, though probably it's not needed.

From reading how the CUDA backend and toolkit work, I think that `event_wait()` only makes sense with parallel streams. It is effectively a mechanism to tell one stream to wait for another stream. So in the context of the RPC backend, it should be a noop.

Apart from that, I think the approach is OK.

wallentri88:
I think it will be useful to check that extra heap allocation of each request and dispatching logic doesn't hurt TG speed when pipeline parallelism is not needed/disabled

xbezdick:
Seems to work fine, together with https://github.com/ggml-org/llama.cpp/pull/16753 I'm getting nice speed improvement.

aoprea1982:
Apologize. Know everyone is busy but any updates on this implementation? Thank you!

rgerganov:
> Apologize. Know everyone is busy but any updates on this implementation? Thank you!

I'm working on the implementation of `cpy_tensor_async`, I hope to post it for review soon

Stoney49th:
thanks for all the good work! @rgerganov 

Let me know when testing is required, I have a real franken-setup here with 2xB50 and a 6900XT (via RPC) - can assist in testing once the async work is ready.

Stoney49th:
maybe a stupid question, but would this be an enabler for tensor parallel via RPC as well?

sredman:
@rgerganov -- No big comments from me. You and I converged on the same design, so implicit +1 from me there :)

I see the same speedup as on my branch with my hardware, so functionally it seems good.

rgerganov:
Here are some performance results coming from Nvidia 4xA100 HF space:

```
./llama-bench -rpc localhost:50052,localhost:50053,localhost:50054,localhost:50055 -hf unsloth/Phi-4-mini-instruct-GGUF,ggml-org/Qwen3.8-27B-GGUF:Q8_0,ggml-org/gemma-4-31B-it-GGUF:Q8_0 --split-mode layer -ngl 99 --ubatch-size 256 --batch-size 2048 --device RPC0,RPC0/RPC1,RPC0/RPC1/RPC2,RPC0/RPC1/RPC2/RPC3 -p 4096 -n 0
```

| model                          |       size |     params | backend    | ngl | n_ubatch | dev          |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | ------------ | --------------: | -------------------: |
| phi3 3B Q4_K - Medium          |   2.31 GiB |     3.84 B | CUDA,RPC   |  99 |      256 | RPC0         |          pp4096 |      7557.80 ± 25.49 |
| phi3 3B Q4_K - Medium          |   2.31 GiB |     3.84 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1    |          pp4096 |      9552.90 ± 61.77 |
| phi3 3B Q4_K - Medium          |   2.31 GiB |     3.84 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1/RPC2 |          pp4096 |     6542.86 ± 158.67 |
| phi3 3B Q4_K - Medium          |   2.31 GiB |     3.84 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1/RPC2/RPC3 |          pp4096 |     5007.86 ± 132.02 |
| qwen35 27B Q8_0                |  26.62 GiB |    26.90 B | CUDA,RPC   |  99 |      256 | RPC0         |          pp4096 |       1173.31 ± 0.55 |
| qwen35 27B Q8_0                |  26.62 GiB |    26.90 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1    |          pp4096 |       1929.63 ± 2.53 |
| qwen35 27B Q8_0                |  26.62 GiB |    26.90 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1/RPC2 |          pp4096 |       1474.27 ± 2.49 |
| qwen35 27B Q8_0                |  26.62 GiB |    26.90 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1/RPC2/RPC3 |          pp4096 |       1306.18 ± 1.48 |
| gemma4 31B Q8_0                |  30.38 GiB |    30.70 B | CUDA,RPC   |  99 |      256 | RPC0         |          pp4096 |       1095.35 ± 0.82 |
| gemma4 31B Q8_0                |  30.38 GiB |    30.70 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1    |          pp4096 |       1802.49 ± 6.14 |
| gemma4 31B Q8_0                |  30.38 GiB |    30.70 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1/RPC2 |          pp4096 |       1375.12 ± 4.08 |
| gemma4 31B Q8_0                |  30.38 GiB |    30.70 B | CUDA,RPC   |  99 |      256 | RPC0/RPC1/RPC2/RPC3 |          pp4096 |       1219.50 ± 2.60 |

Adding a second RPC server running on localhost gives us 30-70% PP improvement, depending on the model. I expect these numbers to be very similar with servers running on different hosts, using RDMA transport. Adding a third RPC server makes things worse, I guess due to the communication overhead between the nodes. Nevertheless, I think these results justify the complexity being added in the RPC backend and there is still room for improvement (e.g. implementing `cpy_tensor_async`).

One mistake I made when I've been testing this in the past was the assumption that I can get performance improvements with two RPC servers using the same physical GPU accelerator. This is simply not the case (at least with CUDA).

sredman:
I recently upgraded to four AMD V620. I also see negative speedup with more than two RPC GPUs. They're all on the same host, so it's kind of a bummer.
I did see near-linear speed improvement with up to four *CPU* RPC devices on different hosts (over Ethernet), so I think we have done this work correctly.

ggerganov:
@rgerganov Can you do a test with `--ubatch-size 256 --batch-size 8192 -p 8192`? I'm wondering if you don't see scaling beyond 2 RPC devices simply because of the small logical batch size.

rgerganov:
> @rgerganov Can you do a test with `--ubatch-size 256 --batch-size 8192 -p 8192`? I'm wondering if you don't see scaling beyond 2 RPC devices simply because of the small logical batch size.

These are the results:

| model                          |       size |     params | backend    | ngl | n_batch | n_ubatch | dev          |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | ------: | -------: | ------------ | --------------: | -------------------: |
| phi3 3B Q4_K - Medium          |   2.31 GiB |     3.84 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0         |          pp8192 |      7098.79 ± 13.02 |
| phi3 3B Q4_K - Medium          |   2.31 GiB |     3.84 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1    |          pp8192 |      9323.24 ± 49.93 |
| phi3 3B Q4_K - Medium          |   2.31 GiB |     3.84 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1/RPC2 |          pp8192 |      6897.75 ± 26.01 |
| phi3 3B Q4_K - Medium          |   2.31 GiB |     3.84 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1/RPC2/RPC3 |          pp8192 |      5782.01 ± 15.19 |
| qwen35 27B Q8_0                |  26.62 GiB |    26.90 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0         |          pp8192 |       1171.64 ± 0.51 |
| qwen35 27B Q8_0                |  26.62 GiB |    26.90 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1    |          pp8192 |       1970.64 ± 7.53 |
| qwen35 27B Q8_0                |  26.62 GiB |    26.90 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1/RPC2 |          pp8192 |       1488.06 ± 2.11 |
| qwen35 27B Q8_0                |  26.62 GiB |    26.90 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1/RPC2/RPC3 |          pp8192 |       1312.52 ± 2.40 |
| gemma4 31B Q8_0                |  30.38 GiB |    30.70 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0         |          pp8192 |       1081.42 ± 0.75 |
| gemma4 31B Q8_0                |  30.38 GiB |    30.70 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1    |          pp8192 |       1823.52 ± 2.67 |
| gemma4 31B Q8_0                |  30.38 GiB |    30.70 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1/RPC2 |          pp8192 |       1387.17 ± 2.49 |
| gemma4 31B Q8_0                |  30.38 GiB |    30.70 B | CUDA,RPC   |  99 |    8192 |      256 | RPC0/RPC1/RPC2/RPC3 |          pp8192 |       1212.99 ± 3.04 |

ggerganov (approved):

ggerganov:
I see - not much difference.

Good to merge?

rgerganov:
> Good to merge?

I am not happy with the `GET_ALLOC_SIZE` cache but I will need more time to rework this and I don't want to hold this anymore. Let's merge and improve in follow ups.

PR #18627, opened 2026-01-05
CANN: Make `valid_values` variable `static const`

rauletorresc:
**Description of the problem**

Variable `valid_values` in function `parse_bool` is an unordered map whose content never changes. But the function is called multiple times, so the variable is unnecessarily created and destroyed at every call.

**Proposed solution**

Make the variable static and constant, so its contents are preserved across different calls of the enclosing function.


hipudding (approved):
Thanks. Any performance improvement?

rauletorresc:
> Thanks. Any performance improvement?

Negligible. Probably because it's called a few times only at the moment.

PR #18629, opened 2026-01-05
gguf-py: Add requests to dependencies

ckastner:
It's used here:

https://github.com/ggml-org/llama.cpp/blob/e443fbcfa51a8a27b15f949397ab94b5e87b2450/gguf-py/gguf/utility.py#L220-L226

Or maybe an optional dependency would suffice?

### DIFF
--- a/gguf-py/pyproject.toml
+++ b/gguf-py/pyproject.toml
@@ -22,6 +22,7 @@ python = ">=3.8"
 numpy = ">=1.17"
 tqdm = ">=4.27"
 pyyaml = ">=5.1"
+requests = ">=2.25"
 sentencepiece = { version = ">=0.1.98,<=0.2.0", optional = true }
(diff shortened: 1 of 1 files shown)


taronaeo (approved):

CISC (approved):
Not sure we need to stay dependency compatible all the way down to version `2.25`, but I guess it's fine.

ckastner:
> Not sure we need to stay dependency compatible all the way down to version `2.25`, but I guess it's fine.

just for context, I saw the Python `>= 3.8` and thought I should also pick something older here, so I arbitrarily picked the oldest version still supported by Debian.

CISC:
> > Not sure we need to stay dependency compatible all the way down to version `2.25`, but I guess it's fine.
> 
> just for context, I saw the Python `>= 3.8` and thought I should also pick something older here, so I arbitrarily picked the oldest version still supported by Debian.

Also for context, `Python >= 3.8` is probably a lie. :)

PR #18633, opened 2026-01-06
Added note for compiling on integrated GPUs

alosslessdev:
Some integrated GPUs such as Radeon 610m do not support rocm. I forget what the exact error message is and I will update this once I can find it

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -349,6 +349,8 @@ You can download it from your Linux distro's package manager or from here: [ROCm
       && cmake --build build -- -j 16
   ```
 
+  If it fails to compile, with a "target not supported" or similar error, it means your GPU does not support ROCm due to missing compiler support, even through it is an RDNA GPU. This can happen if you are trying to use an integrated GPU. In this case, build for Vulkan instead to use the GPU.
+
 - Using `CMake` for Windows (using x64 Native Tools Command Prompt for VS, and assuming a gfx1100-compatible AMD GPU):
   ```bash
   set PATH=%HIP_PATH%\bin;%PATH%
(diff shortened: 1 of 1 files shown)


IMbackK:
610m is gfx1037 which is isa identical to gfx1030 so you should be able to compile and run llamacpp if you specify this arch by hand and use the HSA override envvar

alosslessdev:
> 610m is gfx1037 which is isa identical to gfx1030 so you should be able to compile and run llamacpp if you specify this arch by hand and use the HSA override envvar

Hello, do you have a table detailing which GPUs are similar to each other? It would be incredibly helpful, is it this table under features overview, https://en.wikipedia.org/wiki/List_of_AMD_graphics_processing_units 

completed with this one? https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/system-requirements.html

Based off of RDNA version here: https://www.techpowerup.com/gpu-specs/radeon-610m.c3994

IMbackK:
unfortunately i am not aware of any such documentation, i only know what i know by looking at llvm anyhow of the top of my head:

all gfx9 gpus gfx900 - gfx908 and gfx90c can execute gfx900 code, these gpu isas are various supersets of it
all gfx101x gpus can execute gfx1010 code and are supersets of it
all gfx103x gpus are isa identical
gfx1100 and gfx1101 are isa identlical gfx1102 and above are _not_ (less registers)
all gfx12 gpus are isa identical
all isas not mentioned are only compatible with themselves.

general note: aside from gfx9 usually gfxXXX0 is the largest design and larger numbers with the same XXX are given to ever smaller designs.

PR #18637, opened 2026-01-06
ggml-cuda: refactor cuda graph usage

am17an:
Refactor cuda graph:

- use consistent naming
- remove unused vars 
- avoid passing references 
- encapsulate logic inside ggml_cuda_graph

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1062,7 +1062,6 @@ struct ggml_cuda_graph {
     bool disable_due_to_gpu_arch = false;
     bool disable_due_to_too_many_updates = false;
     int number_consecutive_updates = 0;
-    bool enabled = false;
     std::vector<ggml_cuda_graph_node_properties> props;
 
     void record_update(bool use_graph, bool update_required) {
@@ -1077,11 +1076,9 @@ struct ggml_cuda_graph {
         }
     }
 
-    void check_disable() {
+    bool is_enabled() const {
         static const bool disable_cuda_graphs_due_to_env = (getenv("GGML_CUDA_DISABLE_GRAPHS") != nullptr);
-        if (disable_due_to_gpu_arch || disable_cuda_graphs_due_to_env || disable_due_to_too_many_updates) {
-            enabled = false;
-        }
+        return !(disable_due_to_gpu_arch || disable_cuda_graphs_due_to_env || disable_due_to_too_many_updates);
     }
 #endif
 };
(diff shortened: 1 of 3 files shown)


ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2915,98 +2915,97 @@ static bool check_node_graph_compatibility(ggml_cgraph * cgraph,
     return use_cuda_graph;
 }
 
-static void set_ggml_graph_node_properties(ggml_tensor * node, ggml_graph_node_properties * graph_node_properties) {
-    graph_node_properties->node_address = node->data;
-    graph_node_properties->node_op = node->op;
+static void ggml_cuda_graph_node_set_properties(ggml_tensor * node, ggml_cuda_graph_node_properties * props) {
```
ggerganov: Putting the object as a first argument is more canonical:

```suggestion
static void ggml_cuda_graph_node_set_properties(ggml_cuda_graph_node_properties * props, ggml_tensor * node) {
```

ggml/src/ggml-cuda/ggml-cuda.cu
JohannesGaessler: Github doesn't let me suggest changes but I think this would b slightly simpler:

```C++
    return cuda_ctx->cuda_graph->enabled;
#else
    return false;
#endif // USE_CUDA_GRAPH
```

ggml/src/ggml-cuda/common.cuh
```diff
@@ -1061,11 +1061,28 @@ struct ggml_cuda_graph {
     std::vector<cudaGraphNode_t> nodes;
     bool disable_due_to_gpu_arch = false;
     bool disable_due_to_too_many_updates = false;
-    bool disable_due_to_failed_graph_capture = false;
     int number_consecutive_updates = 0;
-    bool cuda_graphs_enabled = false;
-    std::vector<ggml_graph_node_properties> ggml_graph_properties;
-    std::vector<ggml_graph_node_properties> extraneous_srcs_properties;
+    bool enabled = false;
```
JohannesGaessler: I don't like that we have duplicated information here as this could lead to bugs if `enabled` becomes inconsistent with the preexisting booleans. Is there a reason why we can't create a method like `is_enabled` that simply returns the logical or of the preexisting booleans instead of calling a method that sets the `enabled` state?

am17an: Yes that makes sense to me

JohannesGaessler (approved):

ggerganov (approved):

PR #18639, opened 2026-01-06
Remove annoying warnings (unused functions)

Nekotekina:
When using common.h as a library, these function produce annoying warnings about not being used.
Using "static" linking for these also doesn't make much sense because it potentially increases executable size with no gains.


ggerganov (approved):

Nekotekina:
Umm, merge?

PR #18061, opened 2025-12-15
convert : move rope_parameters to TextModel class

CISC:
Follow up to #18013

Fixes multimodal model configs.

PR #18640, opened 2026-01-06
memory: add is_iswa for memory_hybrid

ngxson:
Alternative to https://github.com/ggml-org/llama.cpp/pull/18601 , reuse `llama_memory_hybrid` instead of duplicating the class

(I haven't tested it yet, just pushing this PR for discussion)


ggerganov:
I prefer the #18601 approach. Indeed, it is duplicating significant chunk of the code, but here we still have to add the new `llm_graph_input_mem_hybrid_iswa`, or alternatively - add the extra `is_iswa` logic there in a similar way. On the other hand, the logic of each class in #18061 is more straightforward and easier to follow.

My recommendation is to proceed with `llama_memory_hybrid_iswa` for now.

ngxson:
Got it, thanks! let's go with the other PR then

ggerganov:
There might be a simplification where instead of having both `llama_kv_cache` and `llama_kv_cache_iswa`, we keep just `llama_kv_cache_iswa` since it is technically a superset of `llama_kv_cache` functionality. But we can do this simplification after this.

PR #18643, opened 2026-01-06
arg: use CSV escape style for multiple-value args

ngxson:
Also fix https://github.com/ggml-org/llama.cpp/issues/18635 along the way

Example usage:

```cpp
argv = {"binary_name", "--lora", "file1.gguf,\"file2,2.gguf\",\"file3\"\"3\"\".gguf\",file4\".gguf"};
assert(true == common_params_parse(argv.size(), list_str_to_char(argv).data(), params, LLAMA_EXAMPLE_COMMON));
assert(params.lora_adapters.size() == 4);
assert(params.lora_adapters[0].path == "file1.gguf");
assert(params.lora_adapters[1].path == "file2,2.gguf");
assert(params.lora_adapters[2].path == "file3\"3\".gguf");
assert(params.lora_adapters[3].path == "file4\".gguf");
```

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -854,6 +854,54 @@ bool common_arg_utils::is_autoy(const std::string & value) {
     return value == "auto" || value == "-1";
 }
 
+// Simple CSV parser that handles quoted fields and escaped quotes
+// example:
+//    input:  value1,"value, with, commas","value with ""escaped"" quotes",value4
+//    output: [value1] [value, with, commas] [value with "escaped" quotes] [value4]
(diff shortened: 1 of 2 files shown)


ServeurpersoCom:
Looks good to me! I test it on a server + it need a doc update:)
The minimalist CSV parser is acceptable for this use case. Edge cases (unclosed quotes, mid-field quotes) produce malformed paths but won't crash, the file-not-found error downstream will catch them anyway. (and no memory safety issues)

ServeurpersoCom (approved):

PR #18641, opened 2026-01-06
[Do Not Merge] model : LFM2.5-Audio-1.5B

tdakhran:
Liquid AI released [LFM2.5-Audio-1.5B](https://huggingface.co/LiquidAI/LFM2.5-Audio-1.5B). 

> LFM2.5-Audio-1.5B is [Liquid AI](https://www.liquid.ai/)'s updated end-to-end audio foundation model. Key improvements include a custom, LFM based audio detokenizer, llama.cpp compatible GGUFs for CPU inference, and better ASR and TTS performance.

This PR is intended to provide a functional implementation in `llama.cpp` until necessary infrastructure is implemented.
The plan is to split and merge it into upstream in smaller chunks, while keeping and tracking functional implementation here. It will be rebased from time to time.

GGUFs, precompiled runners, and instructions, live in https://huggingface.co/LiquidAI/LFM2.5-Audio-1.5B-GGUF.

Merge plan:

- [x] Support `n_embd_out`  https://github.com/ggml-org/llama.cpp/pull/18607
- [x] Support llama_memory_hybrid_iswa https://github.com/ggml-org/llama.cpp/pull/18601
- [x] istft for audio output https://github.com/ggml-org/llama.cpp/pull/18645
- [x] reuse llama-model infra for audio tokenizer https://github.com/ggml-org/llama.cpp/pull/19687
- [ ] support speech output in llama-server and mtmd API @ngxson 
- [ ] tbd

Demo of capabilities (watch with audio on)

https://github.com/user-attachments/assets/bfb7e598-f3bb-4e19-8d93-967bbf912e18

Thank you, @ngxson for the help!

PR #18645, opened 2026-01-06
mtmd: mtmd_audio_streaming_istft for audio output

tdakhran:
Change is decoupled from https://github.com/ggml-org/llama.cpp/pull/18641.

[LFM2.5-Audio-1.5B](https://huggingface.co/LiquidAI/LFM2.5-Audio-1.5B) needs streaming istft for generating output audio.

* add streaming ISTFT class (`mtmd_audio_streaming_istft`) with overlap-add for audio reconstruction
* replace global audio cache with per-instance cache, the model requires two independent caches, for preprocessing (audio input) and for istft (audio output).
* unified templated FFT/IFFT implementation supporting both forward and inverse transforms

### DIFF
--- a/tools/mtmd/mtmd-audio.cpp
+++ b/tools/mtmd/mtmd-audio.cpp
@@ -9,205 +9,248 @@
 #include <fstream>
 #include <algorithm>
 
-// most of the code here is copied from whisper.cpp
+// some of the code here is copied from whisper.cpp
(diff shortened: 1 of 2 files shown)


ngxson (approved):
nice, thanks! I ran the test and confirmed that this doesn't break other models

PR #18646, opened 2026-01-06
vulkan: reject ops when a tensor is too large to allocate

jeffbolznv:
For #18621.


qmastery16:
@0cc4m This genius fix by @jeffbolznv really resolves my issue, more information here - https://github.com/ggml-org/llama.cpp/issues/18621#issuecomment-3716945282

0cc4m (approved):

PR #17535, opened 2025-11-26
refactor : use common download in tools/run

angt:
- update the common/download interface to be directly usable by `tools/run` (removing duplicated code).
- fix ollama downloads by implementing manual redirect handling (addressing issues with `cpp-httplib`).

PR #12761, opened 2025-04-04
cmake : enable curl by default

ngxson:
When curl was first added to the project, we decided to leave it disabled by default because there was no guarantee that the target system has `libcurl` installed.

But lot of things changed since then, curl now become part of the overall UX of llama.cpp examples. Most of the pre-built version (including own docker images and release binary) have it enabled, so I think it's now a good time to turn it on by default.

PR #18648, opened 2026-01-06
common: use httplib + boringssl by default

ngxson:
This is necessary because we want to implement a CORS proxy endpoint on server in near future (used for MCP)

A demo for CORS proxy is here: https://github.com/ngxson/llama.cpp/pull/76

CC @angt , I remember you planned to make this the default config. Is there anything else need to be considered?

### DIFF
--- a/vendor/cpp-httplib/CMakeLists.txt
+++ b/vendor/cpp-httplib/CMakeLists.txt
@@ -29,7 +29,11 @@ target_compile_definitions(${TARGET} PRIVATE
 
 set(OPENSSL_NO_ASM ON CACHE BOOL "Disable OpenSSL ASM code when building BoringSSL or LibreSSL")
 
-if (LLAMA_BORINGSSL)
+if (NOT LLAMA_HTTPLIB_SSL)
+    message(STATUS "Building cpp-httplib without SSL support")
+    set(CPPHTTPLIB_OPENSSL_SUPPORT FALSE)
+
+elseif (LLAMA_BORINGSSL)
     set(FIPS OFF CACHE BOOL "Enable FIPS (BoringSSL)")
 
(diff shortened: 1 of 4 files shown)


angt:
I wanted first to completely remove cURL but i got stopped on this PR #17535 and nothing happened since...

note: `llama-run` is the last blocking point

ngxson:
In such case, let's ask if @ggerganov is ok for removing `llama-run`. IMO it's no longer relevant because `llama-cli` already offers more functionalities than `llama-run`

angt:
I also think `llama-run` is redundant with the other tools, but I think it's used by https://github.com/containers/ramalama.
If we could avoid any drama 😅

ngxson:
Apparently `ramalama` moved away from `llama-run` a long time ago: https://github.com/ggml-org/llama.cpp/pull/12761#discussion_r2029873467

ggerganov:
> In such case, let's ask if @ggerganov is ok for removing `llama-run`. IMO it's no longer relevant because `llama-cli` already offers more functionalities than `llama-run`

I agree, let's remove `llama-run` now that we have `llama-cli`.

ngxson:
@angt feel free to take over this PR and/or open a new one to use httplib + remove curl at the same time

vendor/cpp-httplib/CMakeLists.txt
```diff
@@ -54,6 +58,8 @@ if (LLAMA_BUILD_BORINGSSL)
     set(BUILD_SHARED_LIBS OFF)
     set(BUILD_TESTING OFF)
 
+    message("NOTE: If this fails, set -LLAMA_HTTPLIB_SSL=OFF to disable SSL support in cpp-httplib")
```
ngxson, line 61: @angt some users may complain about not being able to build without internet access, so I added this warning just in case. Feel free to suggest if there is a better way

PR #18658, opened 2026-01-07
model-conversion : add detect_pooling script

danbev:
This commit adds a Python script to automatically detect the pooling configuration from a sentence-transformers model directory.

The motivation for this change is that I make a mistake when adding the sentence-transformers support and I incorrectly assumed that if an embedding model uses sentence-transformers, it always used pooling. With the recent addition of support for late interaction models, which can have a down-projection but do not use pooling (like LFM2-ColBert-350M).

Refs: https://github.com/ggml-org/llama.cpp/pull/18607#issuecomment-3710530332

### DIFF
--- a/examples/model-conversion/Makefile
+++ b/examples/model-conversion/Makefile
@@ -136,18 +136,17 @@ embedding-run-original-model-st: USE_SENTENCE_TRANSFORMERS=1
 embedding-run-original-model-st: embedding-run-original-model
 
 embedding-run-converted-model:
-	@./scripts/embedding/run-converted-model.sh $(CONVERTED_EMBEDDING_MODEL) \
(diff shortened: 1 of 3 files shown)


CISC:
You may want to use something similar to this logic instead:
https://github.com/ggml-org/llama.cpp/blob/ee5811119f2fbe8f0c867d129da2dda41a60f353/convert_hf_to_gguf.py#L1554-L1578

danbev:
> You may want to use something similar to this logic instead:

Thanks! I'm going to revisit this again tomorrow and see if we actually need this at all now that we have the pooling_type in the gguf metadata. So we might be able to just remove the `-st` and pooling from the scripts and makefile targets and this should just work with llama-debug.

danbev:
Closing this as I'll open a new [PR](https://github.com/ggml-org/llama.cpp/pull/18689) that removes the `-st` targets for the converted model.

PR #18661, opened 2026-01-07
tools : remove llama-run

angt:


angt:
ref: https://github.com/ggml-org/llama.cpp/pull/18648#issuecomment-3718241213

ngxson (approved):
We should also remove 2 places where `linenoise` is mentioned:
- at the end of main `README.md` where we list the deps
- in `licences`

taronaeo (approved):

ggerganov (approved):
@angt Feel free to merge when ready.

angt:
Lets go! In case of drama we can still revert 😬

PR #18664, opened 2026-01-07
fix(docker): add missing libglvnd libraries to Vulkan image

### LINKED ISSUE #17761: Misc. bug: Missing libs in vulkan docker image
### Name and Version

Can confirm the fix in this discussion works, and else i have
ggml_vulkan: No devices found.

https://github.com/ggml-org/llama.cpp/discussions/16138

RUN apt-get update && apt-get install -y --no-install-recommends libglvnd0 libgl1 libglx0 libegl1 libgles2 && rm -rf /var/lib/apt/lists/*

### Operating systems

_No response_

_No response_

### Command line

```shell
--list-devices
```

ggml_vulkan: No devices found.

### First Bad Commit

_No response_

### Relevant log output

```shell
load_backend: loaded RPC backend from /app/libggml-rpc.so
ggml_vulkan: No devices found.
load_backend: loaded Vulkan backend from /app/libggml-vulkan.so
load_backend: loaded CPU backend from /app/libggml-cpu-haswell.so
```

686f6c61:
## Summary

Fixes #17761

Related: #16138

## Problem

The `mesa-vulkan-drivers` package has a runtime dependency on libglvnd libraries that aren't installed in the base image. Without them, the Mesa Vulkan ICD can't initialize properly and `vkEnumeratePhysicalDevices()` returns an empty device list, resulting in "ggml_vulkan: No devices found."

## Solution

Add the missing libglvnd libraries to `.devops/vulkan.Dockerfile`:

- libglvnd0
- libgl1
- libglx0
- libegl1
- libgles2

## Tests performed

- Dependencies check (`apt-get check`): OK, no conflicts
- Server starts correctly (`--help`, `--version`): OK
- Image size: 443MB -> 490MB (+47MB, +10.6%)
- llvmpipe software renderer: OK, no regressions

Before (original Dockerfile):
```
$ docker run --rm --device /dev/dri:/dev/dri llama-vulkan-original:test --list-devices
ggml_vulkan: No devices found.
```

After (with fix):
```
$ docker run --rm vulkan-test-fixed dpkg -l | grep -E 'libglvnd|libegl|libgl'
ii  libegl1:amd64         1.7.0-3    Vendor neutral GL dispatch library -- EGL support
ii  libgl1:amd64          1.7.0-3    Vendor neutral GL dispatch library -- legacy GL support
ii  libglvnd0:amd64       1.7.0-3    Vendor neutral GL dispatch library
ii  libglx0:amd64         1.7.0-3    Vendor neutral GL dispatch library -- GLX support
```

If anyone has additional tests they'd like to suggest, please let me know.

### DIFF
--- a/.devops/vulkan.Dockerfile
+++ b/.devops/vulkan.Dockerfile
@@ -33,6 +33,7 @@ FROM ubuntu:$UBUNTU_VERSION AS base
 
 RUN apt-get update \
     && apt-get install -y libgomp1 curl libvulkan1 mesa-vulkan-drivers \
+    libglvnd0 libgl1 libglx0 libegl1 libgles2 \
     && apt autoremove -y \
     && apt clean -y \
     && rm -rf /tmp/* /var/tmp/* \


ericcurtin:
@686f6c61 thanks! Kinda miss Fedora sometimes, a lot of these get installed as recommended packages if memory serves me right. @0cc4m is the vulkan reviewer generally.

0cc4m:
Yeah, I'll review it. I'm just wondering whether we missed that it doesn't work or whether it did/does work at least sometimes.

686f6c61:
@0cc4m , i can confirm the issue exists in the current version. I can't say whether it was present in previous ones.

I was experiencing the same problem as the users reporting in #17761 and #16138, so after investigating I found the missing libglvnd dependencies and made this PR.

686f6c61:
That makes sense about Fedora. I tested this on Ubuntu 26.04 where these packages aren't pulled automatically, so the Vulkan ICD fails to initialize without them.

686f6c61:
@ericcurtin  that makes sense about Fedora. I tested this on Ubuntu 26.04 where these packages aren't pulled automatically, so the Vulkan ICD fails to initialize without them.

ericcurtin:
> Yeah, I'll review it. I'm just wondering whether we missed that it doesn't work or whether it did/does work at least sometimes.

With mesa different hardware has different dependencies, different drivers, etc.

ericcurtin:
In Fedora this is basically everything, dunno how this compares to ubuntu package layout:

```
dnf install -y vulkan-headers vulkan-loader-devel vulkan-tools spirv-tools glslc glslang mesa-vulkan-drivers virglrenderer
```

686f6c61:
@ericcurtin You're absolutely right about Mesa and hardware-specific dependencies. I did some digging into this before responding.

From what I found, the differences between hardware typically happen at two levels: the kernel drivers (amdgpu, i915, nouveau) and the specific Vulkan ICDs that Mesa provides (RADV for AMD, ANV for Intel, NVK for NVIDIA, Lavapipe for software rendering).

However, the libglvnd libraries sit at a lower layer - they're the vendor-neutral dispatch that all Mesa Vulkan drivers need to initialize the EGL display subsystem and enumerate physical devices. Without them, `vkEnumeratePhysicalDevices()` returns an empty list regardless of which GPU you have.

So while the actual Vulkan driver that gets loaded will vary depending on hardware, the libglvnd dependency is common to all of them. The fix should work equally for AMD, Intel, NVIDIA (via NVK), or even CPU-based rendering with Lavapipe.

I tested this on AMD hardware, but based on how the Mesa ICD initialization works, it should apply across the board.

If you have a different perspective or think we should approach this another way, I'm happy to work on it. Thanks!

ericcurtin:
@686f6c61 this pr seems fine to me

0cc4m (approved):

0cc4m:
I also see no issues. Thank you for the fix.

686f6c61:
@ericcurtin @0cc4m Thanks for the quick review and merge! Really appreciate you taking the time to look at this.

@ericcurtin The Fedora package list is helpful for reference. It's interesting to see how the two distros organize these dependencies differently - Fedora seems to bundle them more cohesively.

Glad this fix is now in master. If anyone runs into related issues with other hardware configurations, happy to help investigate.

EZForever:
> Yeah, I'll review it. I'm just wondering whether we missed that it doesn't work or whether it did/does work at least sometimes.

@0cc4m I've been running the `server-vulkan` image for quite some time, at least a few months with a AMD MI50 GPU, and have never encountered issues like "ggml_vulkan: No devices found", etc., so I can confirm the image works before *at least sometimes*, and I actually doubt the claim that the added `libglvnd` libraries are crucial to Mesa Vulkan ICD to function properly (a crude Google search shows that `libglvnd` is for dispatching *OpenGL* calls, which should not matter given we're using native Vulkan APIs). At least for me, this PR is adding about 50MB of unused (?) code to my deployment.

I'm happy to see a PR fixes a problem that other people are encountering, but I would like to ask the proposer of this fix, @686f6c61 : since you have given the fix without ever explaining *why* it works, can you share how did you analyze this issue, and what have led you to the conclusion you have shared with us?

(Also, this is probably me being paranoid, but is this PR ~~'s text~~ generated by AI?)

686f6c61:
@EZForever, good questions.

**Debugging process:**
1. Reproduced the "No devices found" error in a fresh container
2. Verified `/dev/dri` devices were accessible inside the container
3. Ran `ldd /usr/lib/x86_64-linux-gnu/libvulkan_radeon.so` to check for missing dependencies
4. Found unresolved symbols pointing to `libEGL.so` and related libraries
5. Traced the dependency chain: Mesa Vulkan WSI uses GBM internally, GBM depends on EGL for buffer initialization, and `libEGL.so` is provided by libglvnd
6. Installed libglvnd packages, problem solved

**Why it works on your setup without explicit libglvnd:**
Your host system or base image likely has libglvnd installed as a transitive dependency from other packages. The libraries get passed through to the container or your image inherits them. Different distro versions handle these dependencies differently.

**About the writing style:**
I'm from Spain with intermediate English. My intention is to communicate clearly so everyone can understand, even if my English is not as good as yours. I use translation tools for that purpose.

Cheers!

EZForever:
Alright, thank you for your detailed analysis, and apologies for my baseless suspicion on your writing style. I'm not a native English speaker also; guess I'm still recovering from PTSD given by endless AI slop on the Internet.

However, I do have noticed one weird thing: accoring to the posts above you're running a Ubuntu 26.04 install with AMD GPU. The official `server-vulkan` image is based on Ubuntu 26.04 also (see <https://github.com/ggml-org/llama.cpp/blob/ae9f8df/.devops/vulkan.Dockerfile#L1>), and `libvulkan_radeon.so` in it don't actually depend on libEGL (at least not listed by `ldd`). Is it being loaded dynamicially, or the environment between us are still different in some ways?

```
[services@EZHOMELAB ~]$ podman exec -it llamacpp /bin/bash
root@9dd938bccd35:/app# cat /etc/os-release
PRETTY_NAME="Ubuntu Resolute Raccoon (development branch)"
NAME="Ubuntu"
VERSION_ID="26.04"
VERSION="26.04 (Resolute Raccoon)"
VERSION_CODENAME=resolute
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=resolute
LOGO=ubuntu-logo
root@9dd938bccd35:/app# ldd /usr/lib/x86_64-linux-gnu/libvulkan_radeon.so | grep -i libEGL
root@9dd938bccd35:/app# 
```

686f6c61:
Good observation. You're right that `ldd` doesn't show a direct dependency on libEGL.

To be transparent: when investigating the "No devices found" issue reported in [Discussion #16138](https://github.com/ggml-org/llama.cpp/discussions/16138) and [Issue #17761](https://github.com/ggml-org/llama.cpp/issues/17761), I found that installing libglvnd packages resolved the problem empirically. Multiple users confirmed it worked.

The exact mechanism (whether it's runtime dlopen, shared DRI infrastructure, or something else in Mesa's architecture) isn't fully clear to me. The dependency seems to be indirect rather than a direct link-time dependency.

If you have deeper insight into why Mesa Vulkan drivers need libglvnd in containerized environments (or if the fix is broader than necessary), I'd appreciate any information that could help refine the PR. Whether that means reducing the installed packages to only what's strictly needed, or improving the documentation with a proper technical explanation.

And no worries - at least we non-native speakers have tools like DeepL to give us a hand.

Cheers!

PR #17824, opened 2025-12-06
cli: new CLI experience

ngxson:
Ref: https://github.com/ggml-org/llama.cpp/discussions/17618

Fix: #11202

We are moving to a new CLI experience with the main code built on top of `llama-server`. This brings many additional features into `llama-cli`, making the experience feels mostly like a smaller version of the web UI:
- Multimodal support
- Regenerate last message
- Speculative decoding
- Fully jinja support (including some edge cases that old `llama-cli` doesn't support)

image

TODO:
- [x] Use `reasoning_content` when possible, depends on https://github.com/ggml-org/llama.cpp/pull/17835
- [x] Allow using left/right arrow to edit current message, depends on https://github.com/ggml-org/llama.cpp/issues/17828
- [x] Maybe allow up/down arrow for history (not sure), depends on https://github.com/ggml-org/llama.cpp/issues/17829
- [x] Allow reporting timings
- [x] Support `--image` and `--audio` arguments
- [x] Add messages to warn user to move to `llama-completion` for legacy features

Features planned for next versions:
- Allow hide/show timings and reasoning content (saving user preferences to disk and reuse later)
- Allow exporting/importing conversation
- Support raw completion
- Support remote media URL (downloaded from internet)
- Show progress (in percentage) if prompt processing takes too long

TODO for console system:
- Auto-completion for commands and file paths
- Support "temporary display", for example clear loading messages when it's done

PR #18670, opened 2026-01-07
cmake: fix cli build when LLAMA_BUILD_SERVER=OFF

AsbjornOlling:
## The Problem

The problem is that building with `-DLLAMA_BUILD_SERVER=OFF` does not work, since `llama-cli` now depends on `server-context` (since #17824).

To reproduce the issue on current latest master (commit 5642667):

```sh
cmake -B build -DLLAMA_BUILD_SERVER=OFF
cmake --build build
```

It will produce the following error:

```
[ 95%] Building CXX object tools/cli/CMakeFiles/llama-cli.dir/cli.cpp.o
In file included from /home/asbjorn/Development/llama-cpp-rs/llama-cpp-sys-2/llama.cpp/tools/cli/../server/server-task.h:12,
                 from /home/asbjorn/Development/llama-cpp-rs/llama-cpp-sys-2/llama.cpp/tools/cli/../server/server-context.h:2,
                 from /home/asbjorn/Development/llama-cpp-rs/llama-cpp-sys-2/llama.cpp/tools/cli/cli.cpp:6:
/home/asbjorn/Development/llama-cpp-rs/llama-cpp-sys-2/llama.cpp/tools/cli/../server/server-common.h:7:10: fatal error: mtmd.h: No such file or directory
    7 | #include "mtmd.h"
      |          ^~~~~~~~
compilation terminated.
make[2]: *** [tools/cli/CMakeFiles/llama-cli.dir/build.make:76: tools/cli/CMakeFiles/llama-cli.dir/cli.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:4013: tools/cli/CMakeFiles/llama-cli.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

This build error is what has prevented [utilityai/llama-cpp-rs](https://github.com/utilityai/llama-cpp-rs) from bumping the llama.cpp version for the past month.

## The Fix

~~I fixed this by moving the `if (LLAMA_BUILD_SERVER)` conditional from `tools/CMakeLists.txt` into the part of `tools/server/CMakeLists.txt` that builds the actual server executable. So now if `LLAMA_BUILD_SERVER` is `OFF`, it will still build the server shared libs (which `llama-cli` depend on), but not the server executable itself.~~

I fixed this by moving the `cli` target into the `LLAMA_BUILD_SERVER` conditional.
If `LLAMA_BUILD_TOOLS=ON` and `LLAMA_BUILD_SERVER=OFF`, it will not build cli or server, but it will build the rest.

I have run the `ci/run.sh` testing suite to confirm that everything works as before.

## AI Disclosure

The code changes in this PR were all made and verified by human hands and eyes, but LLM assistance was used for identifying the problem and suggesting several solutions (one of which is this solution).


ngxson:
You can skip building llama-cli when build_server is disabled. I prefer that simple fix. You won't use cli for binding anyway

tools/CMakeLists.txt
```diff
@@ -22,9 +22,7 @@ else()
     add_subdirectory(completion)
     add_subdirectory(perplexity)
     add_subdirectory(quantize)
-    if (LLAMA_BUILD_SERVER)
-        add_subdirectory(server)
```
ngxson, line 26: simply add `add_subdirectory(cli)` here

AsbjornOlling:
> You can skip building llama-cli when build_server is disabled. I prefer that simple fix. You won't use cli for binding anyway

We don't need `llama-cli` for bindings, but don't we need to have `LLAMA_BUILD_TOOLS` enabled in order to have multimodal support in the bindings? It seems like all of the mtmd stuff lives in `tools/mtmd/`.

I have tested setting `LLAMA_BUILD_TOOLS=OFF` in [utilityai/llama-cpp-rs](https://github.com/utilityai/llama-cpp-rs), and it indeed does result in a linking error. It was enabled in llama-cpp-rs when multimodal support was added in https://github.com/utilityai/llama-cpp-rs/pull/790

ngxson:
Have you read my suggestion above?

AsbjornOlling:
> Have you read my suggestion above?

Ah okay. I read it, but misunderstood it. Sorry about that. :sweat_smile: 

I implemented your suggestion now.

tools/CMakeLists.txt
AsbjornOlling: Done.

ngxson (approved):

PR #18644, opened 2026-01-06
scripts : add pr2wt.sh

ggerganov:
Helper script for quickly creating a git worktree from a PR.

Sample usage:

```bash
# create worktree in ../llama.cpp-pr-12345
./scripts/pr2wt.sh 12345

# create worktree and start opencode in it
./scripts/pr2wt.sh 12345 opencode

# chain multiple commands:
./scripts/pr2wt.sh 12345 "cmake -B build && cmake --build build"
```

PR #18671, opened 2026-01-07
scripts : support chaining commands in pr2wt.sh

ggerganov:
cont #18644 

Allows using:

```bash
./scripts/pr2wt.sh 12345 "cmake -B build && cmake --build build"
```

### DIFF
--- a/scripts/pr2wt.sh
+++ b/scripts/pr2wt.sh
@@ -9,6 +9,7 @@
 # sample usage:
 #   ./scripts/pr2wt.sh 12345
 #   ./scripts/pr2wt.sh 12345 opencode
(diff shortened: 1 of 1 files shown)


scripts/pr2wt.sh
```diff
@@ -61,5 +62,5 @@ echo "git worktree created in $wt_path"
 # if a command was provided, execute it
 if [[ $# -eq 2 ]]; then
     cd ../$dir-pr-$PR
-    exec $2
+    eval "$2"
```
CISC, line 65: These behave very differently, but I suppose for most use-cases it's fine.

CISC (approved):

PR #16813, opened 2025-10-28
llama: Fused QKV multiplication

am17an:
This is a draft PR for comments about QKV fusion. Performance wise I checked CUDA it seems to be roughly ~4-5% performance gain (in PP and TG) for qwen3 models. 

This tries to merge the weights together to form a single GEMM, rather than 3 separate ones. The main drawback of this approach is that there might be other tensors in between the Q,K,V weights (like LoRAs) which would break mmap (as @slaren pointed out). However, doing this on the backend is a much more involved change which might not worth the benefit. For now I'm thinking the best way forward be

1. Add the wqkv, bqkv tensors to future GGUFs using an option in `convert_hf_to_gguf.py` 
2. In case the user wants and knows the weights are together, a flag `--fuse-qkv-weights` can be provided to take this path.

IMO the QKV weights should always live in the GGUF file together without exceptions as they're naturally used together.

Some performance numbers on a 4090:

Without fusion
| model                          |       size |     params | backend    | ngl |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | --------------: | -------------------: |
| qwen3moe 30B.A3B Q4_0          |  16.11 GiB |    30.53 B | CUDA       |  99 |           pp512 |      3846.64 ± 15.30 |
| qwen3moe 30B.A3B Q4_0          |  16.11 GiB |    30.53 B | CUDA       |  99 |          pp1024 |      5651.13 ± 21.24 |
| qwen3moe 30B.A3B Q4_0          |  16.11 GiB |    30.53 B | CUDA       |  99 |          pp2048 |      7024.90 ± 14.43 |
| qwen3moe 30B.A3B Q4_0          |  16.11 GiB |    30.53 B | CUDA       |  99 |          pp4096 |       7504.26 ± 9.66 |
| qwen3moe 30B.A3B Q4_0          |  16.11 GiB |    30.53 B | CUDA       |  99 |           tg128 |        151.89 ± 0.10 |

With fusion

| model                          |       size |     params | backend    | ngl |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | --------------: | -------------------: |
| qwen3moe 30B.A3B Q4_0          |  16.11 GiB |    30.53 B | CUDA       |  99 |           pp512 |      3933.52 ± 20.74 |
| qwen3moe 30B.A3B Q4_0          |  16.11 GiB |    30.53 B | CUDA       |  99 |          pp1024 |      5768.19 ± 19.00 |
[...]

PR #18683, opened 2026-01-08
model: try to improve Qwen3 Next

ngxson:
> [!IMPORTANT]  
> If you're using old GGUF and it's no longer loaded, be sure to update to this fix: https://github.com/ggml-org/llama.cpp/pull/18762

I was quite curious why there was a function called `fix_query_key_value_ordering` in transformers code (which was mirrored into llama.cpp implementation). Just wondering what are they trying to fix.

Turns out, the projected QKVZ was in a wrong order. Not sure why they don't fix the original weight instead of permuting the results. But that's not very not important.

I took my trusty pen & paper to see what can be done:

image

Since the projected matrix is big, I supposed it will give a significant boost in perf. But I was quite disappointed to see only 1% improvement in pp512. So I don't even know if it's worth the fix. GGUF will need to reconverted to take advantage of this.

(Update: there are more improvements that allow from 5% all the way to 20% boost depending on backend, see [my comment below](https://github.com/ggml-org/llama.cpp/pull/18683#issuecomment-3732574765))

master:

| qwen3next 80B.A3B F16          | 148.50 GiB |    79.67 B | Metal,BLAS |      24 |           pp512 |       790.87 ± 17.93 |
| qwen3next 80B.A3B F16          | 148.50 GiB |    79.67 B | Metal,BLAS |      24 |           tg128 |         22.33 ± 0.08 |

PR:

| qwen3next 80B.A3B F16          | 148.50 GiB |    79.67 B | Metal,BLAS |      24 |           pp512 |        965.15 ± 8.00 |
| qwen3next 80B.A3B F16          | 148.50 GiB |    79.67 B | Metal,BLAS |      24 |           tg128 |         28.87 ± 0.06 |

Hardware: mac studio m3 256gb ram

---

I uploaded a q8_0 here: https://huggingface.co/ngxson/qwen3_next_fixed/tree/main (converted from Instruct version)

It can be used to test against q8_0 already exist on the internet: https://huggingface.co/Qwen/Qwen3-Next-80B-A3B-Instruct-GGUF/blob/main/Qwen3-Next-80B-A3B-Instruct-Q8_0.gguf


pwilkin:
I like the idea (and it surely does make the computation clearer), but I'm not sure if people would appreciate having to recreate and redownload GGUFs at this point for a 1% performance increase. I don't mind, but would like to see what people who use the model more think first.

EDIT: never mind, I only now looked at the code in detail and you added a compatibility path. Yeah, I think it's worth to do it :) maybe it'll help more on other backends?

ngxson:
Yeah I would appreciate helps on testing other backend too. Curious to see if it's better on CUDA. The apple unified memory thingy can make things a bit complicated to measure.

(My optimization should only take effects on system with constrained memory bandwidth)

jeffbolznv:
Do I need to do anything special to test it (e.g. regenerate ggufs)?

ngxson:
@jeffbolznv yes, you will need to generate the GGUF with the PR:

Download the model here: https://huggingface.co/Qwen/Qwen3-Next-80B-A3B-Instruct

Then convert it as normal:

```sh
python ../../llama.cpp/convert_hf_to_gguf.py . --outtype f16 --outfile model.gguf
```

pwilkin:
It's actually BF16 :) 

`"torch_dtype": "bfloat16"`

jacekpoplawski:
I understand that many 1% improvements add up to something significant.
If someone publishes a quantized GGUF, it will be much easier for people to benchmark it on different configurations.

am17an:
a merged QKV in normal models gives about ~3-4% improvement in PP last time I checked. #16813. However, many backends already support computing these things in parallel so the METAL backend perhaps won't benefit as much. You can try the CUDA backend, I suspect the gains will be slightly more pronounced

jeffbolznv:
I have limited cycles for the next week or so, if somebody can make a quantized model easily accessible it'll be easier for me to try it.

IIIIIllllIIIIIlllll:
I don't have a CUDA device at the moment, so I simply did the test on AI MAX+ 395.

Master
```
command: /home/mark/llama.cpp/llama.cpp-master/build/bin/llama-bench -m /home/mark/Models/Q8/Qwen3-Next-80B-A3B-Instruct-Q8_0/next-master.gguf -r 5 -p 2048 -n 32 -b 2048 -ub 2048 -fa 1 -mmp 0

  Device 0: AMD Radeon Graphics, gfx1151 (0x1151), VMM: no, Wave Size: 32
| model                          |       size |     params | backend    | ngl | n_ubatch | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |     2048 |  1 |    0 |          pp2048 |        579.45 ± 1.62 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |     2048 |  1 |    0 |            tg32 |         28.98 ± 0.04 |

build: unknown (0)
```
This
```
command: /home/mark/llama.cpp/llama.cpp-xsn-qwen3next_improve/build/bin/llama-bench -m /home/mark/Models/Q8/Test-Qwen3-Next/model.gguf -r 5 -p 2048 -n 32 -b 2048 -ub 2048 -fa 1 -mmp 0

  Device 0: AMD Radeon Graphics, gfx1151 (0x1151), VMM: no, Wave Size: 32
| model                          |       size |     params | backend    | ngl | n_ubatch | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |     2048 |  1 |    0 |          pp2048 |        593.37 ± 1.45 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |     2048 |  1 |    0 |            tg32 |         29.18 ± 0.03 |

build: unknown (0)
```

ggerganov:
Btw, I'm also planning to apply the #18550 functionality to Qwen3 Next when it is ready in order to make the ggml graphs static in memory and avoid the graph reallocations that currently occur because of the chunking. It still won't allow CUDA graphs to be used, but at least we should be able to avoid a lot of overhead that we currently have from the changing number of nodes based on the batch size.

ngxson:
@am17an the problem with qwen3 next is that qkv is not actually used for attention. Instead, this qkvz tensor is used for ssm conv (naming is a bit confusing indeed). The old logic does: qkvz projection --> permute --> concat which is redundant (compared to attention: projection --> rope --> permute --> attention)

As I expected, seems like this really have an improvement for system with less memory bandwidth as @IIIIIllllIIIIIlllll confirmed. The projected qkvz tensors is large, so it should be faster if we don't need to do any permutation on it.

ngxson:
@jeffbolznv I'll try to upload a q8_0 later today

ngxson:
I uploaded a q8_0 here: https://huggingface.co/ngxson/qwen3_next_fixed/tree/main (converted from Instruct version)

It can be used to test against q8_0 already exist on the internet: https://huggingface.co/Qwen/Qwen3-Next-80B-A3B-Instruct-GGUF/blob/main/Qwen3-Next-80B-A3B-Instruct-Q8_0.gguf

Note: @pwilkin we usually test on f16 because it's the most compatible type cross all backend. Some backend like CPU internally convert bf16 to f16 if the hardware doesn't support it; for hardware that supports bf16, there should be no differences between the 2

But that's not important for this PR: we are comparing perf before/after so the importance is to make sure we're using the same type

pwilkin:
Ah, okay, got it.

ngxson:
@ggerganov I'm thinking of this idea: can we enforce the number of chunks? For example, we can enforce cgraph to always allocate `n_chunks = ubatch_size / CHUNK_SIZE`, and unused chunks will have 0 elements, which allow backend to skip them.

This should make the cgraph topology to be more static, although, it is unclear for me if CUDA graph expects the tensor shapes to be unchanged (the case of 0 elements for unused chunk that I talked above)

One extra question: I remember we had a mechanism in ggml to detect no-op (like `GGML_OP_VIEW`). I'm wondering if it can be beneficial to extend it to consider singleton as no-op (for example: transpose a tensor with n_elements = 1)

jeffbolznv:
My test system is an RTX 5090 with 32GB, so I'm not sure I can get useful data from this 80GB model. I ran with `-ngl 19` which is the most layers I can fit, and see like a 10x slowdown with the new model/changes but it's probably either due to paging or due to worse CPU inferencing performance. I've done my previous qwen3next testing on a Q2_K_L model which is around 29GB.

ggerganov:
> @ggerganov I'm thinking of this idea: can we enforce the number of chunks? For example, we can enforce cgraph to always allocate n_chunks = ubatch_size / CHUNK_SIZE, and unused chunks will have 0 elements, which allow backend to skip them.

Not sure how robust ggml is for zero-sized tensors. Think it will need some significant changes to be compatible. But we can definitely give this idea a try.

> One extra question: I remember we had a mechanism in ggml to detect no-op (like GGML_OP_VIEW). I'm wondering if it can be beneficial to extend it to consider singleton as no-op (for example: transpose a tensor with n_elements = 1)

Hm, not sure I understand. Transpose is already a noop:

https://github.com/ggml-org/llama.cpp/blob/2524c2616458c7d6ee62fa4b4fa17e5091833544/ggml/src/ggml-impl.h#L87-L100

ngxson:
@ggerganov what I mean is that there can be also cases where `ggml_cont` on a contiguous tensor, or `ggml_sum_rows` on a tensor with `t->ne[0] == 1`, etc, which can all be consider as no-op. That's just an idea for maybe merge 2 branches: chunk and autoregressive together in the future (although, it can be much more difficult to actually do it)

danbev:
<details>
<summary><b>System Config: DGX Spark (Grace CPU + Blackwell GPU)</b></summary>

```bash
# OS & Kernel
Description:	Ubuntu 24.04.3 LTS
6.11.0-1016-nvidia

# CPU (Grace)
Architecture:                         aarch64
Byte Order:                           Little Endian
Vendor ID:                            ARM
Model name:                           Cortex-X925
Model:                                1
Thread(s) per core:                   1
Model name:                           Cortex-A725
Model:                                1
Thread(s) per core:                   1

# GPU (Blackwell)
name, driver_version, memory.total [MiB], compute_cap
NVIDIA GB10, 580.95.05, [N/A], 12.1

# CUDA Version
Cuda compilation tools, release 13.0, V13.0.88
```
</details>

I ran the following command for the benchmarks (same as used in [comment](https://github.com/ggml-org/llama.cpp/pull/18683#issuecomment-3722045120)):
```console
./build/bin/llama-bench -m models/Qwen3-Next-80B-A3B-Instruct-bf16-Q8_0.gguf -r 5 -p 2048 -n 32 -b 2048 -ub 2048 -fa 1 -mmp 0
```
Let me know if there are other configurations that you'd like me to run.

master:
```console
| model                          |       size |     params | backend    | ngl | n_ubatch | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | CUDA       |  99 |     2048 |  1 |    0 |          pp2048 |        899.72 ± 2.06 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | CUDA       |  99 |     2048 |  1 |    0 |            tg32 |         29.77 ± 0.04 |

build: 8ece3836b (7681)

```
PR:
```console
| model                          |       size |     params | backend    | ngl | n_ubatch | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | CUDA       |  99 |     2048 |  1 |    0 |          pp2048 |        924.32 ± 2.21 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | CUDA       |  99 |     2048 |  1 |    0 |            tg32 |         31.20 ± 0.52 |

build: 65602e899 (7686)

```

lemmi:
Since Qwen3-Next has this weird behaviour on vulkan that pp performance falls of a cliff with `ubatch-size > 512`, I was wondering, whether the changes had any effect, so I did some benchmarks aswell:

Master:

| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      256 |  1 |          pp4096 |        390.58 ± 1.82 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      256 |  1 |           tg128 |         32.99 ± 0.03 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |          pp4096 |        431.52 ± 1.35 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |           tg128 |         32.86 ± 0.32 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |          pp4096 |        334.67 ± 3.65 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |           tg128 |         33.00 ± 0.01 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |          pp4096 |        255.08 ± 1.54 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |           tg128 |         32.95 ± 0.06 |

PR: 

| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      256 |  1 |          pp4096 |        383.22 ± 2.88 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      256 |  1 |           tg128 |         33.41 ± 0.05 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |          pp4096 |        452.23 ± 1.77 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |           tg128 |         33.43 ± 0.05 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |          pp4096 |        292.53 ± 4.28 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |           tg128 |         33.42 ± 0.02 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |          pp4096 |        269.66 ± 3.04 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |           tg128 |         33.31 ± 0.12 |

PR breaks with `-mmap 0` though: 
```
load_tensors: offloading output layer to GPU
load_tensors: offloading 47 repeating layers to GPU
load_tensors: offloaded 49/49 layers to GPU
load_tensors:      Vulkan0 model buffer size = 80562.07 MiB
load_tensors:  Vulkan_Host model buffer size =   315.30 MiB
load_all_data: using async uploads for device Vulkan0, buffer type Vulkan0, backend Vulkan0
llama_model_load: error loading model: read error: Bad address
main: error: failed to load model '/scratch/models/qwen3_next_fixed/model-q8_0.gguf'
```

jeffbolznv:
> Since Qwen3-Next has this weird behaviour on vulkan that pp performance falls of a cliff with ubatch-size > 512

I haven't seen this, please file a bug to track it.

> llama_model_load: error loading model: read error: Bad address

This should be fixed at TOT master.

ngxson:
Thanks for testing @danbev @lemmi . It seems like on most backend, this does consistently bring from 1% to 3% improvement. I'm not quite sure bout why we get inconsistent results with vulkan though, but it probably unrelated to the currently change.

@jeffbolznv I uploaded a Q2_K here (I cannot quantize to Q2_K_L because it's not a valid scheme in llama-quantize)
- fixed model: https://huggingface.co/ngxson/qwen3_next_fixed/blob/main/model-Q2_K.gguf
- original model: https://huggingface.co/ngxson/qwen3_next_fixed/blob/main/model-orig-Q2_K.gguf

lemmi:
> > Since Qwen3-Next has this weird behaviour on vulkan that pp performance falls of a cliff with ubatch-size > 512
> 
> I haven't seen this, please file a bug to track it.

#18725 there you go

> > llama_model_load: error loading model: read error: Bad address
> 
> This should be fixed at TOT master.

"TOT"?

ngxson:
I removed 2 redundant transposes which improved from 793 --> 820 t/s in my test (GGUF stays the same). Would appreciate if you can rerun the test with latest commit

lemmi:
With d96eb69e2 `--mmap 0` works again and the numbers have also improved a bit.

| model                          |       size |     params | backend    | ngl | n_ubatch | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      256 |  1 |    0 |          pp4096 |        392.86 ± 2.40 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      256 |  1 |    0 |           tg128 |         33.71 ± 0.02 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |    0 |          pp4096 |        457.53 ± 1.80 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |    0 |           tg128 |         33.64 ± 0.03 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |    0 |          pp4096 |        392.29 ± 2.50 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |    0 |           tg128 |         33.68 ± 0.03 |

jeffbolznv:
Here's a quick before/after for Q2_K on Vulkan RTX 5090. Seems like it's at the limits of vidmem and occasionally spills and gets very slow, both with and without this change. Looks like ~5% faster.

```
before

Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo>llama-bench.exe -fa 1 -m c:\models\model-orig-Q2_K.gguf -n 128 -p 512,1024,2048 -ub 512,1024,2048
ggml_vulkan: 0 = NVIDIA GeForce RTX 5090 (NVIDIA) | uma: 0 | fp16: 1 | bf16: 1 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: NV_coopmat2
load_backend: loaded Vulkan backend from Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo\ggml-vulkan.dll
load_backend: loaded CPU backend from Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo\ggml-cpu.dll
| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |           pp512 |      4350.55 ± 85.44 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |          pp1024 |      4181.56 ± 11.97 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |          pp2048 |       4281.07 ± 9.39 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |           tg128 |        147.02 ± 2.24 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |           pp512 |      4408.04 ± 20.33 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |          pp1024 |      5080.18 ± 30.54 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |          pp2048 |    3942.66 ± 1148.31 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |           tg128 |        147.01 ± 3.10 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |           pp512 |      4390.52 ± 24.24 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |          pp1024 |      5196.31 ± 10.23 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |          pp2048 |      5338.67 ± 18.21 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |           tg128 |        142.55 ± 3.30 |

build: 53eb9435d (7684)

after

Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo>llama-bench.exe -fa 1 -m c:\models\model-Q2_K.gguf -n 128 -p 512,1024,2048 -ub 512,1024,2048
ggml_vulkan: 0 = NVIDIA GeForce RTX 5090 (NVIDIA) | uma: 0 | fp16: 1 | bf16: 1 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: NV_coopmat2
load_backend: loaded Vulkan backend from Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo\ggml-vulkan.dll
load_backend: loaded CPU backend from Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo\ggml-cpu.dll
| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |           pp512 |      4520.18 ± 19.25 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |          pp1024 |      4344.81 ± 12.70 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |          pp2048 |      4432.59 ± 20.83 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |           tg128 |        146.07 ± 7.39 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |           pp512 |      4527.51 ± 21.06 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |          pp1024 |      5324.04 ± 30.80 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |          pp2048 |      5083.77 ± 17.57 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |           tg128 |        156.39 ± 0.50 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |           pp512 |      4463.32 ± 25.80 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |          pp1024 |      5278.55 ± 31.29 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |          pp2048 |        343.07 ± 1.07 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |           tg128 |        154.33 ± 2.70 |

build: d96eb69e2 (7694)
```

danbev:
Results using https://github.com/ggml-org/llama.cpp/commit/d96eb69e2fea47dd40be2904db9346530ece85b8:
```console
| model                          |       size |     params | backend    | ngl | n_ubatch | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | CUDA       |  99 |     2048 |  1 |    0 |          pp2048 |        940.81 ± 2.31 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | CUDA       |  99 |     2048 |  1 |    0 |            tg32 |         31.44 ± 0.36 |

build: 8c1336cb9 (7697)
```

ngxson:
I ran `llama-eval-callback` to make sure that I didn't break anything, looks good so far. So until this point I can confirm 4-5% perf boost.

Working on optimizing chunking logic today, hope to get a bit more perf out of it

src/models/qwen3next.cpp
```diff
@@ -713,26 +710,25 @@ ggml_tensor * llm_build_qwen3next::build_layer_attn_linear(
     ggml_tensor * conv_output_proper = ggml_ssm_conv(ctx0, conv_input, conv_kernel);
     cb(conv_output_proper, "conv_output_raw", il);
 
-    conv_output_proper = ggml_cont(ctx0, ggml_transpose(ctx0, conv_output_proper));
-    cb(conv_output_proper, "conv_output_pre_silu", il);
```
ggerganov, line 681: It might not be necessary depending on the shapes, but it would not hurt to add `assert`s here that the transpose is noop. At least, just looking at this change alone, I don't understand how this transformation is identity. The only way that I can think of is if `conv_output_proper->ne[0] == conv_output_proper->ne[1] == 1` 🤔

ngxson: Not sure if I understand your comment correctly, but I'm removing this not because it's an identity transpose.

I suspect that it was transposed just to apply silu, then transposed back (see LOCs below). But since silu is element-wise, the transpose is redundant.

I verified the results before/after this change via `llama-eval-callback` and verify that it doesn't affect the end results

ggerganov: Ah yes - I was confused for some reason thinking that silu is a row-wise op. Nevermind then 👍

ngxson:
So I just pushed some changes that bumped the perf in my tests from 820 to 960 t/s on pp512, very significant improvement. So in total, we're getting ~20% boost between master and PR.

The 2 most important commits are:
- https://github.com/ggml-org/llama.cpp/pull/18683/commits/e1f8ad25649bc7cb54ede0e8e2533057fc2f411e gives 2% boost; I removed some redundant `ggml_cont`. I tested on metal and see no side effects on metal. Though, it's best to test it properly. To test it, use `llama-cli -f prompt.txt --temp 0 --top-k 1` with a large enough `prompt.txt` (around 2048 tokens is enough). If the results stay unchanged between PR/master, that means nothing is broken
- https://github.com/ggml-org/llama.cpp/pull/18683/commits/f8ad742aeb7f09367a94987ffdc7216a5bfe3bb6 gives another ~10% perf boost. The old code path returns 2 tensors by `ggml_concat` them, then later separate using `ggml_view`. Here we can return them as `std::pair`

But first, would appreciate if you can run a simple `llama-bench` on your side @danbev @jeffbolznv @lemmi 

On my mac m3 ultra:

```
| qwen3next 80B.A3B F16          | 148.50 GiB |    79.67 B | Metal,BLAS |      24 |           pp512 |        965.15 ± 8.00 |
| qwen3next 80B.A3B F16          | 148.50 GiB |    79.67 B | Metal,BLAS |      24 |           tg128 |         28.87 ± 0.06 |

build: f8ad742ae (7699)
```

lemmi:
| model                          |       size |     params | backend    | ngl | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |  1 |    0 |           pp512 |        481.87 ± 3.65 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |  1 |    0 |          pp4096 |        460.94 ± 1.39 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | Vulkan     |  99 |  1 |    0 |           tg128 |         34.49 ± 0.03 |

TG improved by 1t/s, rest also got a small bump.

src/models/qwen3next.cpp
```diff
@@ -235,37 +235,35 @@ ggml_tensor * llm_build_qwen3next::build_delta_net_chunking(
...
+    ggml_tensor * attn_kq = ggml_mul_mat(ctx0, k, q);
+    attn_kq = ggml_mul(ctx0, attn_kq, decay_mask);
+    attn_kq = ggml_mul(ctx0, attn_kq, diag_mask);
+    cb(attn_kq, "attn_kq", il);
 
-        auto chunkify_g = [=](ggml_tensor * t) {
-            return ggml_cont(ctx0, ggml_view_4d(ctx0, t, chunk_size, t->ne[1], 1, t->ne[3],
-                t->nb[1], t->nb[2], t->nb[3], t->nb[2] * chunk));
+    for (int64_t chunk = 0; chunk < n_chunks; chunk++) {
+        static auto chunkify = [](ggml_context * ctx0, ggml_tensor * t, int64_t chunk) {
```
CISC:
```suggestion
        static auto chunkify = [chunk_size](ggml_context * ctx0, ggml_tensor * t, int64_t chunk) {
```

ngxson: The lambda is static so it cannot capture local var. I'll move it to a static function later, ~~while using the global `CHUNK_SIZE`~~

Edit: it can be abstract to a function like `get_slice()` which simply get a slice of 3rd dim, example: `shape: [x, y, n, b]` --> `get_slice(n) output shape: [x, y, 1, b]`

CISC: Another fix would be to directly use `CHUNK_SIZE`, but `get_slice` sounds like a better idea.

IIIIIllllIIIIIlllll:
The test results are still AI MAX+ 395. The performance has indeed improved significantly.


==============================
**This PR:**

cmd:
/home/mark/llama.cpp/llama.cpp-xsn-qwen3next_improve/build/bin/llama-bench -m /home/mark/Models/Q8/Test-Qwen3-Next/new.gguf --repetitions 5 --output md --mmap 0 --delay 0 --n-prompt 512 --n-gen 128 --n-depth 0 --batch-size 2048 --ubatch-size 512 --cache-type-k f16 --cache-type-v f16 --threads 16 --n-gpu-layers 99 --n-cpu-moe 0 --flash-attn 1

output:
```
  Device 0: AMD Radeon Graphics, gfx1151 (0x1151), VMM: no, Wave Size: 32
| model                          |       size |     params | backend    | ngl | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |  1 |    0 |           pp512 |        605.77 ± 4.80 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |  1 |    0 |           tg128 |         29.86 ± 0.03 |

build: unknown (0)
```


==============================
**Master PR:** 

cmd: 
/home/mark/llama.cpp/llama.cpp-master-rocm/llama-bench -m /home/mark/Models/Q8/Qwen3-Next-80B-A3B-Instruct-Q8_0/next-master.gguf --repetitions 5 --output md --mmap 0 --delay 0 --n-prompt 512 --n-gen 128 --n-depth 0 --batch-size 2048 --ubatch-size 512 --cache-type-k f16 --cache-type-v f16 --threads 16 --n-gpu-layers 99 --n-cpu-moe 0 --flash-attn 1

output:
```
  Device 0: AMD Radeon Graphics, gfx1151 (0x1151), VMM: no, Wave Size: 32
| model                          |       size |     params | backend    | ngl | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |  1 |    0 |           pp512 |        576.19 ± 4.31 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |  1 |    0 |           tg128 |         28.99 ± 0.01 |

build: unknown (0)
```

src/models/qwen3next.cpp
```diff
@@ -208,116 +206,124 @@ ggml_tensor * llm_build_qwen3next::build_delta_net_chunking(
...
-        ggml_tensor * g_diff_exp = ggml_exp(ctx0, g_diff);
-
-        ggml_tensor * key_gdiff = ggml_mul(ctx0, k_chunk,
-                                        ggml_reshape_4d(ctx0, g_diff_exp, 1, g_diff_exp->ne[0], g_diff_exp->ne[1],
-                                                        g_diff_exp->ne[2] * g_diff_exp->ne[3]));
-
-        ggml_tensor * kgdmulvnew = ggml_mul_mat(ctx0, v_new_t, ggml_cont(ctx0, ggml_transpose(ctx0, key_gdiff)));
+        ggml_tensor * k_gdiff = ggml_cont(ctx0, chunkify(ctx0, key_gdiff, chunk));
+        //ggml_tensor * kgdmulvnew = ggml_mul_mat(ctx0, k_gdiff, v_new); // this is slower on metal, why?
+        ggml_tensor * kgdmulvnew = ggml_mul_mat(ctx0, v_new_t, ggml_cont(ctx0, ggml_transpose(ctx0, k_gdiff)));
```
ngxson, line 327: @ggerganov do you have any ideas why the commented version can be slower? (I'm testing on Metal)

both tensors are contiguous, here are the shapes:

```
k_gdiff ne: 128 64 1 32
v_new ne: 128 64 1 32
```

jeffbolznv:
Maybe another 5-10% depending on mode with the latest. The numbers continue to be noisy (likely due to spilling) but it's clearly improved.

```

Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo>llama-bench.exe -fa 1 -m c:\models\model-Q2_K.gguf -n 128 -p 512,1024,2048 -ub 512,1024,2048
ggml_vulkan: 0 = NVIDIA GeForce RTX 5090 (NVIDIA) | uma: 0 | fp16: 1 | bf16: 1 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: NV_coopmat2
load_backend: loaded Vulkan backend from Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo\ggml-vulkan.dll
load_backend: loaded CPU backend from Z:\github\jeffbolznv\llama.cpp\build\bin\RelWithDebInfo\ggml-cpu.dll
| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |           pp512 |     4820.56 ± 102.46 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |          pp1024 |      4716.74 ± 16.33 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |          pp2048 |      4809.92 ± 14.32 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |      512 |  1 |           tg128 |        164.14 ± 1.20 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |           pp512 |      4885.30 ± 27.15 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |          pp1024 |      5657.14 ± 27.04 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |          pp2048 |      5484.45 ± 19.96 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     1024 |  1 |           tg128 |        163.62 ± 1.17 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |           pp512 |      4883.83 ± 33.79 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |          pp1024 |      5660.00 ± 18.91 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |          pp2048 |      536.38 ± 264.19 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | Vulkan     |  99 |     2048 |  1 |           tg128 |        160.92 ± 3.24 |

build: 9299ced6b (7701)
```

ngxson:
@jeffbolznv could you show the log of graph splits? (I don't quite remember which env var to enable it)

On metal, I got some splits because ggml_pad is not supported, but it can be added in another PR. I'm curious to see why vulkan graph is split

src/models/qwen3next.cpp
ngxson: I abstracted it into an utility called `get_slice_2d` (no longer need to provide chunk size)

```cpp
// utility to get one slice from the third dimension
// input dim:  [x, y, c, b]
// output dim: [x, y, 1, b]
static ggml_tensor * get_slice_2d(ggml_context * ctx0, ggml_tensor * t, int64_t c) {
    return ggml_view_4d(ctx0, t, t->ne[0], t->ne[1], 1, t->ne[3],
        t->nb[1], t->nb[2], t->nb[3], t->nb[2] * c);
};
```

ngxson:
Alright, this is as far as I can go for now. My last commits don't improve much on performance (at least on metal), but it further reduced the number of nodes. On master branch, we current have 11k nodes in total, and this PR brought it down to ~9k nodes.

```
PR:
llama_context: Flash Attention was auto, set to enabled
llama_context:      Metal compute buffer size =   846.80 MiB
llama_context:        CPU compute buffer size =   524.01 MiB
llama_context: graph nodes  = 9374 (with bs=512), 5918 (with bs=1)
llama_context: graph splits = 76 (with bs=512), 4 (with bs=1)

master:
llama_context: Flash Attention was auto, set to enabled
llama_context:      Metal compute buffer size =   870.80 MiB
llama_context:        CPU compute buffer size =   524.01 MiB
llama_context: graph nodes  = 11186 (with bs=512), 6614 (with bs=1)
llama_context: graph splits = 76 (with bs=512), 4 (with bs=1)
```

IIIIIllllIIIIIlllll:
There is no further performance improvement on my AI MAX+ 395 either.

```
  Device 0: AMD Radeon Graphics, gfx1151 (0x1151), VMM: no, Wave Size: 32
| model                          |       size |     params | backend    | ngl | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |  1 |    0 |           pp512 |        605.15 ± 5.10 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | ROCm       |  99 |  1 |    0 |           tg128 |         30.00 ± 0.01 |

build: unknown (0)
```

jeffbolznv:
I'm not at my computer for a while, but I dont think there should be any splits, except for the initial layer(s).

To enable the logging set GGML_SCHED_DEBUG=2 and use -v.

jeffbolznv:
Confirming that I only see the initial split 0 with GET_ROWS and then the remaining nodes in split 1.

danbev:
```console
| model                          |       size |     params | backend    | ngl | n_ubatch | fa | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | ---: | --------------: | -------------------: |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | CUDA       |  99 |     2048 |  1 |    0 |          pp2048 |        978.86 ± 2.65 |
| qwen3next 80B.A3B Q8_0         |  78.98 GiB |    79.67 B | CUDA       |  99 |     2048 |  1 |    0 |            tg32 |         31.16 ± 0.69 |

build: d5a085696 (7704)

```

jacekpoplawski:
**three GPUs**

  Device 1: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes
  Device 2: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes

original:

| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           pp512 |        721.64 ± 2.92 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           tg128 |         67.14 ± 2.78 |

build: 707cbafca (7700)

improved:

| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           pp512 |        743.86 ± 3.90 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           tg128 |         71.88 ± 3.15 |

build: d5a085696 (7704)

**two GPUs**

  Device 1: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes

original:

| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           pp512 |        988.48 ± 6.17 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           tg128 |         74.99 ± 0.06 |

build: 707cbafca (7700)

improved:

| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           pp512 |       1029.37 ± 8.31 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           tg128 |         79.99 ± 0.07 |

build: d5a085696 (7704)

CISC (approved):

ngxson:
@bartowski1182 @danielhanchen if possible, could you re-generate the GGUF to take advantage of this change? Thanks!!

SirSchnobi:
I saw this PR on Reddit. I was wondering why my GGUF does not load anymore since this update. 

I get the below error when using an old non-fixed GGUF.
Is this expected or are old models just not supported anymore?

I run vulkan radv with the llama.cpp b7703 in a docker container on Strix Halo here. 


```
Attaching to llama-vulkan-radv
llama-vulkan-radv  | [39389] ggml_vulkan: Found 1 Vulkan devices:
llama-vulkan-radv  | [39389] ggml_vulkan: 0 = Radeon 8060S Graphics (RADV GFX1151) (radv) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 64 | shared memory: 65536 | int dot: 1 | matrix cores: KHR_coopmat
llama-vulkan-radv  | [39389] main: n_parallel is set to auto, using n_parallel = 4 and kv_unified = true
llama-vulkan-radv  | [39389] build: 7703 (506bb6e01) with GNU 15.2.1 for Linux x86_64
llama-vulkan-radv  | [39389] system info: n_threads = 32, n_threads_batch = 32, total_threads = 32
llama-vulkan-radv  | [39389] 
llama-vulkan-radv  | [39389] system_info: n_threads = 32 (n_threads_batch = 32) / 32 | CPU : SSE3 = 1 | SSSE3 = 1 | AVX = 1 | AVX_VNNI = 1 | AVX2 = 1 | F16C = 1 | FMA = 1 | BMI2 = 1 | AVX512 = 1 | AVX512_VBMI = 1 | AVX512_VNNI = 1 | AVX512_BF16 = 1 | LLAMAFILE = 1 | OPENMP = 1 | REPACK = 1 | 
llama-vulkan-radv  | [39389] 
llama-vulkan-radv  | [39389] init: using 31 threads for HTTP server
llama-vulkan-radv  | [39389] start: binding port with default address family
llama-vulkan-radv  | [39389] main: loading model
llama-vulkan-radv  | [39389] srv    load_model: loading model '/models/Qwen3-Next-80B-A3B-Instruct-UD-Q8_K_XL.gguf'
llama-vulkan-radv  | [39389] common_init_result: fitting params to device memory, for bugs during this step try to reproduce them with -fit off, or provide --verbose logs if the bug only occurs with -fit on
llama-vulkan-radv  | [39389] /opt/llama.cpp/ggml/src/ggml.c:3553: GGML_ASSERT(ggml_is_contiguous(a)) failed
llama-vulkan-radv  | [39389] /lib64/libggml-base.so.0(+0x35a5) [0x7f593929f5a5]
llama-vulkan-radv  | [39389] /lib64/libggml-base.so.0(ggml_print_backtrace+0x1eb) [0x7f593929f96b]
llama-vulkan-radv  | [39389] /lib64/libggml-base.so.0(ggml_abort+0x11f) [0x7f593929faef]
llama-vulkan-radv  | [39389] /lib64/libggml-base.so.0(ggml_reshape_2d+0xab) [0x7f59392a5d6b]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(_ZN19llm_build_qwen3next23build_layer_attn_linearEP18llm_graph_input_rsP11ggml_tensorS3_S3_S3_i+0xa66) [0x7f593cc21246]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(_ZN19llm_build_qwen3nextC1ERK11llama_modelRK16llm_graph_params+0x195) [0x7f593cc21785]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(_ZNK11llama_model11build_graphERK16llm_graph_params+0x755) [0x7f593cb39105]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(_ZN13llama_context13graph_reserveEjjjPK22llama_memory_context_ibPm+0x193) [0x7f593cac3b33]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(_ZN13llama_contextC2ERK11llama_model20llama_context_params+0x17f6) [0x7f593cac86c6]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(llama_init_from_model+0x113) [0x7f593cac92d3]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(+0x2664a) [0x7f593ca9f64a]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(+0x27767) [0x7f593caa0767]
llama-vulkan-radv  | [39389] /lib64/libllama.so.0(llama_params_fit+0x4e) [0x7f593caa442e]
llama-vulkan-radv  | [39389] /usr/bin/llama-server() [0x64e951]
llama-vulkan-radv  | [39389] /usr/bin/llama-server() [0x6512c9]
llama-vulkan-radv  | [39389] /usr/bin/llama-server() [0x50736a]
llama-vulkan-radv  | [39389] /usr/bin/llama-server() [0x46a86d]
llama-vulkan-radv  | [39389] /lib64/libc.so.6(+0x35b5) [0x7f5938d145b5]
llama-vulkan-radv  | [39389] /lib64/libc.so.6(__libc_start_main+0x88) [0x7f5938d14668]
llama-vulkan-radv  | [39389] /usr/bin/llama-server() [0x46ce85]
```

Som-anon:
@SirSchnobi try this #18761 and comment there. Commenting on merged pull requests is not the best way to report bugs...

ngxson:
will be fixed on https://github.com/ggml-org/llama.cpp/pull/18762

danielhanchen:
@ngxson Nice work will update today!

danielhanchen:
Updated Instruct at https://huggingface.co/unsloth/Qwen3-Next-80B-A3B-Instruct-GGUF and Thinking is ongoing! Had to redo imatirx cal :)

bartowski1182:
Both of mine have been updated, good work on this

https://huggingface.co/bartowski/Qwen_Qwen3-Next-80B-A3B-Thinking-GGUF

https://huggingface.co/bartowski/Qwen_Qwen3-Next-80B-A3B-Instruct-GGUF

PR #18692, opened 2026-01-08
debug : include LLAMA_POOLING_TYPE_UNSPECIFIED in pooling check

danbev:
This commit updates the pooling check in the debug example to also include LLAMA_POOLING_TYPE_UNSPECIFIED and not just LLAMA_POOLING_TYPE_NONE.

### DIFF
--- a/examples/debug/debug.cpp
+++ b/examples/debug/debug.cpp
@@ -71,7 +71,7 @@ struct output_data {
     float *                  data_ptr    = nullptr;
     int                      data_size   = 0;
     std::string              type_suffix;
-    std::vector<float>       storage;
+    std::vector<float>       embd_norm;
     std::string              prompt;
     std::vector<llama_token> tokens;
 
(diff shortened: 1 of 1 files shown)


ggerganov (approved):

ggerganov:
See my comments in https://github.com/ggml-org/llama.cpp/pull/18464. If my understanding is correct there, then we likely don't need to query the pooling here at all and this can be simplified.

danbev:
> See my comments in #18464. If my understanding is correct there, then we likely don't need to query the pooling here at all and this can be simplified.

Thanks, I'll take a look at this now :+1: 
@ggerganov I've added https://github.com/ggml-org/llama.cpp/pull/18692/commits/f31c38dd19d012750146945ed930637d6a1050f2 to address this. Is this what you had in mind?

ggerganov (approved):

PR #18694, opened 2026-01-08
Webui/file upload

ServeurpersoCom:

Fix restrictive file type validation in WebUI

The file upload filter was rejecting many valid text file formats (VHDL, TCL, Verilog, Makefiles without extensions, etc.) with no option to override the automatic detection

Changes:
- Make isFileTypeSupported() inclusive by default: fallback to treating unknown files as text
- Simplify file processing logic: remove redundant TEXT type check, let fallback handle all text files
- Images/Audio/PDF still route to specialized handlers, everything else attempts text extraction

This resolves the issue where users had to rename files to .txt just to upload them. If a file is actually binary, the UTF-8 read will gracefully fail and the file is added without content

No more :
Remove

Tests (specific file codepath unchanged, new unknown extension supported as a text file for LLM):
VHDL
Image

Fixes #18618
Related: #17556


ServeurpersoCom:
Potential and Optional improvement/simplification: the file picker could just offer "*.*" by default (removes a little more existing code) since the codepath is already done in the application, and any fallback to TEXT for the LLM if unknown :

Picker

allozaur:
> Potential and Optional improvement/simplification: the file picker could just offer "_._" by default (removes a little more existing code) since the codepath is already done in the application, and any fallback to TEXT for the LLM if unknown :
> Picker

sure, let's improve this then :)

ServeurpersoCom:
I carefully remove a lot of code! It will be simpler, while preserving the internal modality

tools/server/webui/src/lib/components/app/chat/ChatForm/ChatForm.svelte
```diff
@@ -148,37 +104,12 @@
...
 	}
 
-	function handleFileUpload(fileType?: FileTypeCategory) {
-		if (fileType) {
-			fileAcceptString = getAcceptStringForFileType(fileType);
-		} else {
-			fileAcceptString = undefined;
-		}
+	function handleFileUpload() {
+		fileAcceptString = undefined;
```
ServeurpersoCom: I can remove this !

pwilkin:
Yeah, I didn't want to mess too much in your mechanisms, but I feel like this is the best way to go.

ServeurpersoCom:
Yes, and it's like a cleanup PR, I like features with a negative line count! Basically, the dialog takes everything, whitelists handles the modalities redirection correctly, and anything not explicitly included inside it falls back to text. After all, LLMs love text :)

tools/server/webui/src/lib/components/app/chat/ChatForm/ChatForm.svelte
ServeurpersoCom: Done

allozaur (approved):

allozaur:
@ServeurpersoCom just one last thing to address is the failing checks ;) storybook tests are failing

ServeurpersoCom:
> @ServeurpersoCom just one last thing to address is the failing checks ;) storybook tests are failing

I'm looking at that!

PR #18695, opened 2026-01-08
scripts : pr2wt.sh reset to remote head

ggerganov:
cont #18644 

Pull from latest remote.

### DIFF
--- a/scripts/pr2wt.sh
+++ b/scripts/pr2wt.sh
@@ -10,6 +10,7 @@
 #   ./scripts/pr2wt.sh 12345
 #   ./scripts/pr2wt.sh 12345 opencode
 #   ./scripts/pr2wt.sh 12345 "cmake -B build && cmake --build build"
+#   ./scripts/pr2wt.sh 12345 "bash -l"
 
 function usage() {
     echo "usage: $0 <pr_number> [cmd]"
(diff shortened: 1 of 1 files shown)


scripts/pr2wt.sh
```diff
@@ -39,29 +39,37 @@ org_repo=${org_repo%.git}
...
 
 wt_path=$(cd ../$dir-pr-$PR && pwd)
 
 echo "git worktree created in $wt_path"
 
+cd $wt_path
+git branch --set-upstream-to=pr/$PR/$head_ref
+
+if [[ -z $(git status --porcelain) ]]; then
+    git reset --hard pr/$PR/$head_ref
```
CISC: What scenario is the hard reset for on a clean status?

CISC (approved):

scripts/pr2wt.sh
ggerganov: The goal is if the local head is behind the remote head to move forward. This can happen if the contributor has updated the PR after you pr2wtd it earlier.

CISC: Sure, the hard reset just seemed dramatic for something supposedly clean. :)

ggerganov: I could misunderstand how git work, but don't think the `--hard` flag is optional in this case. Without it, you will get a diff with staged changes.

@copilot Is the `--hard` flag necessary here?

ggerganov: @CISC Got it - changed to `git pull --ff-only`.

CISC:
What happened to setting upstream?

scripts/pr2wt.sh
```diff
@@ -39,29 +40,40 @@ org_repo=${org_repo%.git}
...
-git worktree add -b pr/$PR ../$dir-pr-$PR pr/$PR/${head_ref} 2> /dev/null
+git worktree add -b pr/$PR ../$dir-pr-$PR pr/$PR/$head_ref 2> /dev/null
 
 wt_path=$(cd ../$dir-pr-$PR && pwd)
 
 echo "git worktree created in $wt_path"
 
-# if a command was provided, execute it
+cd $wt_path
+git branch --set-upstream-to=pr/$PR/$head_ref
```
CISC, line 70: Hmmm, I notice that `git worktree add` is actually already setting up the tracking, not sure exactly under what conditions (if any) it won't as you're deleting the local branch first.

I guess it won't hurt either though.

CISC (approved):

PR #18696, opened 2026-01-08
common : add --license to display embedded licenses

angt:
This commit introduces a mechanism to embed all licenses directly into the compiled binaries.

This eliminates the need to distribute separate LICENSE files alongside the executable, making the binaries self-contained and simplifying deployment.


angt:
This automatically adds the license of cpp-httplib/boringssl/libressl only when they are used.

ngxson (approved):

ggerganov (approved):

PR #18239, opened 2025-12-21
ggml-cpu: parallelize tensor repacking with OpenMP

pestopoppa:
## Summary

Add OpenMP parallelization to tensor repack functions to significantly speed up model loading on many-core CPUs.

## Motivation

On systems with many cores (e.g., AMD EPYC), tensor repacking for AVX-512 optimization was single-threaded and became a significant bottleneck during model loading. The repack functions convert quantized tensors from storage layout to SIMD-optimized interleaved layout.

## Benchmark Results

Measured on AMD EPYC 9655 "Turin" (96 cores, 192 threads):

| Model Size | Before | After | Speedup |
|------------|--------|-------|---------|
| 6.8GB Q4_K | 5.0s   | 3.3s  | **1.5x** |
| 19GB Q4_K  | 11.9s  | 5.3s  | **2.2x** |

Speedup increases with model size as repack time dominates over I/O.

## Changes

- Convert pointer-increment loops to explicit indexing (parallelizable)
- Add `#pragma omp parallel for` to outer loops
- Move thread-local `dst_tmp` arrays inside parallel region
- Each thread processes independent row groups with no synchronization needed

## Functions Parallelized

- `repack_q4_0_to_q4_0_4_bl` - Q4_0 x4 interleave
- `repack_q4_K_to_q4_K_8_bl` - Q4_K models (most common)
- `repack_q2_K_to_q2_K_8_bl` - Q2_K models
- `repack_q4_0_to_q4_0_8_bl` - Q4_0 x8 interleave
- `repack_iq4_nl_to_iq4_nl_4_bl` - IQ4_NL x4
- `repack_iq4_nl_to_iq4_nl_8_bl` - IQ4_NL x8

## Testing

- Verified outputs match original implementation
- Tested on multiple Q4_K_M models
- Build verified with GCC 13.3.0 + OpenMP

PR #18698, opened 2026-01-08
Improving inference speed for the repack buffer type on NUMA architectures

zzjianhui:
DEPENDS ON PR #18239 
This PR implements improving inference speed for the repack buffer type on NUMA and is currently built on top of PR #18239 . Once PR #18239  is merged, I will rebase this branch.

When using repack buffer type, the physical memory allocation is dictated by the first-touch policy. Since the main thread performs repack, weight is often allocated on a single NUMA node, leading to uneven weight distribution.

Multi-threaded repack can alleviate this problem, but the threads are not bound to NUMA nodes.

This patch applies the same thread affinity strategy (--numa distribute) to the repacking phase. By binding the repack threads to the same nodes as the compute threads, we ensure that weights are written (and thus allocated) on the local NUMA node, minimizing cross-node memory access during inference.

Performance on Intel Xeon Silver 4514Y (32 core), set --numa distribute:
qwen3 8B Q4_K: 19.39 -> 26.92 t/s (+39%)
qwen3 32B Q4_K: 4.99 -> 7.38 t/s (+48%)

```shell
| model                  |      size |  params | backend | threads |  test |          t/s |
# master
| qwen3 8B Q4_K - Medium |  4.68 GiB |  8.19 B | CPU     |      32 | tg128 | 19.39 ± 0.23 |
# OMP Repack PR#18239
| qwen3 8B Q4_K - Medium |  4.68 GiB |  8.19 B | CPU     |      32 | tg128 | 21.81 ± 0.11 |
# bind core
| qwen3 8B Q4_K - Medium |  4.68 GiB |  8.19 B | CPU     |      32 | tg128 | 26.92 ± 0.16 |
```
```shell
| model                  |      size |  params | backend | threads |  test |          t/s |
# master
| qwen3 32B Q4_K - Medium | 18.40 GiB | 32.76 B | CPU     |      32 | tg128 |  4.99 ± 0.01 |
# OMP Repack PR#18239
| qwen3 32B Q4_K - Medium | 18.40 GiB | 32.76 B | CPU     |      32 | tg128 |  6.39 ± 0.03 |
# bind core
| qwen3 32B Q4_K - Medium | 18.40 GiB | 32.76 B | CPU     |      32 | tg128 |  7.38 ± 0.03|

### CODE REVIEW

**Comment by zzjianhui** [comment]
@ggerganov Could you please take a look at the PR?

File: ggml/src/ggml-cpu/repack.cpp [hunk 1]
```diff
@@ -2188,9 +2192,21 @@ template <typename BLOC_TYPE, int64_t INTER_SIZE, int64_t NB_COLS, ggml_type PAR
...
+        {
+            ggml_cpu_set_numa_thread_affinity(omp_get_thread_num());
+            int r = ggml::cpu::repack::repack<BLOC_TYPE, INTER_SIZE, NB_COLS>(t, data, data_size);
+            #pragma omp master
+            ret = r;
+        }
+#else
+        ret = ggml::cpu::repack::repack<BLOC_TYPE, INTER_SIZE, NB_COLS>(t, data, data_size);
+#endif
+        return ret;
```
ggerganov, line 2503: How do we know that this thread distribution corresponds to the one used during compute? AFAICT there is nothing to guarantee this and it really depends on how the repack/compute loops are written.

Also, when we create the threadpool, we use different indexing:

https://github.com/ggml-org/llama.cpp/blob/1051ecd28907d2ca0a15c135f190fe415d0a3d1b/ggml/src/ggml-cpu/ggml-cpu.c#L3131-L3137

ggerganov (commented):
Not really familiar with NUMA stuff and don't have appropriate hardware to test these changes. Try to find someone who is more familiar with the topic to review.

ggml/src/ggml-cpu/repack.cpp
zzjianhui: When the --numa distribute option is set, the set_numa_thread_affinity function is called prior to computation to bind threads. The binding pattern is T0-N0, T1-N1, T2-N0... (assuming two NUMA nodes). 

During OpenMP multi-threaded repacking, set_numa_thread_affinity is invoked to apply the same binding pattern. Consequently, memory allocated by threads 0, 2, 4... resides on NODE0, while memory allocated by threads 1, 3, 5... resides on NODE1.

Currently, scenarios other than OpenMP are not considered.

zzjianhui: The key point is that **physical thread identity does not need to be the same**  between repack and inference phases. What matters is the **logical thread ID  consistency**.

This works because of NUMA's **first-touch memory allocation policy**:

**During repack phase:**
- Logical thread 0 writes weight data → memory allocated on Node 0
- Logical thread 1 writes weight data → memory allocated on Node 1
- ...

**During inference phase:**
- Logical thread 0 (bound to Node 0) accesses memory written by repack's logical thread 0
- This ensures NUMA-local memory access

The `ggml_cpu_set_numa_thread_affinity(omp_get_thread_num())` and the threadpool's  binding both use the same NUMA distribution pattern based on `g_state.numa`,  so the logical-ID-to-NUMA-node mapping is consistent.

**However, you are right that the thread count must be consistent.** 

I plan to add a mechanism to set the repack thread count.

zzjianhui:
I tested the perplexity before and after the modification, and it remained unchanged.

```shell
#!/bin/bash

echo never > /sys/kernel/mm/transparent_hugepage/enabled
sysctl -w kernel.numa_balancing=0

./build_master/bin/llama-perplexity -m /home/zx/models/qwen3-32b-q4km.gguf \
-f ./wikitext-2-raw/wiki.test.raw --chunks 32 -t 32 --numa distribute 2>&1 | grep "Final estimate: PPL"

./build_p2/bin/llama-perplexity -m /home/zx/models/qwen3-32b-q4km.gguf \
-f ./wikitext-2-raw/wiki.test.raw --chunks 32 -t 32 --numa distribute 2>&1 | grep "Final estimate: PPL"

./build_master/bin/llama-perplexity -m /home/zx/models/qwen3-30b-a3b-thinking-2507-q4km.gguf \
-f ./wikitext-2-raw/wiki.test.raw --chunks 32 -t 32 --numa distribute 2>&1 | grep "Final estimate: PPL"

./build_p2/bin/llama-perplexity -m /home/zx/models/qwen3-30b-a3b-thinking-2507-q4km.gguf \
-f ./wikitext-2-raw/wiki.test.raw --chunks 32 -t 32 --numa distribute 2>&1 | grep "Final estimate: PPL"
```

Output:

```shell
kernel.numa_balancing = 0
Final estimate: PPL = 6.8095 +/- 0.20135
Final estimate: PPL = 6.8095 +/- 0.20135
Final estimate: PPL = 6.1121 +/- 0.17424
Final estimate: PPL = 6.1121 +/- 0.17424
```

PR #18700, opened 2026-01-08
server : use different seeds for child completions

ggerganov:
rel #18663

I realized that we want to have different RNG seeds for the `n_cmpl` completions - otherwise we will generate the same result for each child. This makes the assert in the `test_chat_completions_multiple_choices()` test invalid.

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -2928,6 +2928,7 @@ std::unique_ptr<server_res_generator> server_routes::handle_completions_impl(
                     server_task child = task.create_child(task.id, rd.get_new_id());
 
                     // use different sampling seed for each child
+                    // note: https://github.com/ggml-org/llama.cpp/pull/18700#discussion_r2675115723
                     if (child.params.sampling.seed != LLAMA_DEFAULT_SEED) {
                         child.params.sampling.seed += j + 1;
                     }
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -4,7 +4,6 @@
 #include "server-task.h"
 #include "server-queue.h"
 
-#include "arg.h"
 #include "common.h"
 #include "llama.h"
 #include "log.h"
@@ -16,7 +15,6 @@
 #include <cstddef>
 #include <cinttypes>
 #include <memory>
-#include <unordered_set>
 #include <filesystem>
 
 // fix problem with std::min and std::max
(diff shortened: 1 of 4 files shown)


ngxson (approved):

tools/server/server-context.cpp
```diff
@@ -2927,9 +2925,11 @@ std::unique_ptr<server_res_generator> server_routes::handle_completions_impl(
...
                 task.n_children = task.params.n_cmpl - 1;
                 for (size_t j = 0; j < task.n_children; j++) {
-                    server_task child = task.create_child(
-                        task.id,
-                        rd.get_new_id());
+                    server_task child = task.create_child(task.id, rd.get_new_id());
+
+                    // use different sampling seed for each child
+                    child.params.sampling.seed += j + 1;
+
```
ngxson, line 2935: Should we also handle the case where default seed is used?

ngxson: Ah nevermind, I think we don't need to care about that case here

ggerganov: Hm, you might be right. If we don't handle the default seed here, then 1 child will have random seed and the rest will get the same seeds.

CISC:
I think `server-windows` CI possibly broke:
https://github.com/ggml-org/llama.cpp/actions/runs/20844687013/job/59885629295
https://github.com/ggml-org/llama.cpp/actions/runs/20847242141/job/59893626577

ggerganov:
These 2 test expect a failure due to the cache (256 tokens) getting filled, but the assumption of the test is that the 4 parallel requests would start together at the same time:

https://github.com/ggml-org/llama.cpp/blob/53eb9435da3affa12a38a8b0fb29081698a8d1cc/tools/server/tests/unit/test_completion.py#L377-L378

So the logic is flaky - if the runner is under load like in these 2 cases, the first request starts much earlier than the others.

CISC:
@ggerganov BTW, `ggml-ci`s having issues again:
https://github.com/ggml-org/llama.cpp/actions?page=2&query=is%3Aqueued

ggerganov:
@CISC Do you mean that the workflows are not processed fast enough?

CISC:
> @CISC Do you mean that the workflows are not processed fast enough?

There were several queued from yesterday, though it seems to have cleared somewhat, now there are just 2 left.

CISC:
> > @CISC Do you mean that the workflows are not processed fast enough?
> 
> There were several queued from yesterday, though it seems to have cleared somewhat, now there are just 2 left.

Ah, I think that may just be because they have timed out.

CISC:
@ggerganov As you can see the nvidia runners seem to be stuck/extremely slow:
https://github.com/ggml-org/llama.cpp/actions?query=is%3Aqueued

PR #18701, opened 2026-01-08
[WIP] ggml-opencl: op args init refactoring

chraac:
Refactors the OpenCL kernel argument initialization process by introducing a template helper, significantly reducing boilerplate code and improving readability.

## Key Changes
- Introduced `cl_set_kernel_args`: A variadic template function that automatically iterates through arguments and sets them using `clSetKernelArg`.
- Type-Specific Setters: Added `cl_kernel_arg_setter` struct specializations to handle different types:
  - **ggml_tensor**: Automatically extracts the OpenCL buffer (`data_device`) and calculates the offset (`offset + view_offs`), setting both as kernel arguments.
  - **Arrays (ne, nb)**: Automatically unrolls `int64_t[4]` and `size_t[4]` arrays into individual scalar arguments (e.g., ne0, ne1, ne2, ne3).
  - *Primitives*: Direct handling for `int`, `float`, `cl_ulong`.


ggml/src/ggml-opencl/ggml-opencl.cpp
```diff
@@ -4994,32 +4997,9 @@ static void ggml_cl_add_id(ggml_backend_t backend, const ggml_tensor * src0, con
...
-    CL_CHECK(clSetKernelArg(kernel,  5, sizeof(cl_ulong), &offset2));
-    CL_CHECK(clSetKernelArg(kernel,  6, sizeof(cl_mem),   &extrad->data_device));
-    CL_CHECK(clSetKernelArg(kernel,  7, sizeof(cl_ulong), &offsetd));
-    CL_CHECK(clSetKernelArg(kernel,  8, sizeof(cl_ulong), &nb01));
-    CL_CHECK(clSetKernelArg(kernel,  9, sizeof(cl_ulong), &nb02));
-    CL_CHECK(clSetKernelArg(kernel, 10, sizeof(cl_ulong), &nb11));
-    CL_CHECK(clSetKernelArg(kernel, 11, sizeof(cl_ulong), &nb21));
-    CL_CHECK(clSetKernelArg(kernel, 12, sizeof(int),      &ne0));
-    CL_CHECK(clSetKernelArg(kernel, 13, sizeof(int),      &ne1));
+    cl_set_kernel_args(kernel, src0, src1, src2, dst, nb01, nb02, nb11, nb21, ne0, ne1);
```
chraac, line 5672: It might be worth considering a shared header for kernel signatures to prevent mismatches between the host and device. 
Knowing the refactoring effort would be high, but it would guarantee consistency.

lhez: Currently we use a simple script to read a kernel file and make it into a C++ string in a header file to embed the kernels into executable. Using headers in kernel files will greatly increase the complexity of embedding kernels.

chraac: > Currently we use a simple script to read a kernel file and make it into a C++ string in a header file to embed the kernels into executable.

Yeah, I saw the [embed_kernel.py](https://github.com/ggml-org/llama.cpp/blob/master/ggml/src/ggml-opencl/kernels/embed_kernel.py) script that converts the CL source into string literals. 
If we introduce a new header here, I think we could extend that script to handle `#include` directives by simply replacing them with the content of the header file from the same directory.

ggml/src/ggml-opencl/ggml-opencl.cpp
```diff
@@ -5231,43 +5185,35 @@ static void ggml_cl_div(ggml_backend_t backend, const ggml_tensor * src0, const
...
+                kernel,
+                src0,
+                src1,
+                dst,
+                src0->nb,
+                src1->ne,
+                src1->nb,
+                ne0,
+                dst->nb
+            );
```
chraac, line 5821: Heres a safer version of this function that'll check if the parameters match the cl kernel func signiture

chraac:
@lhez 
Also have a thought regarding `load_cl_kernels`: Consider refactoring this to use a data structure (like an attribute table) to store the program sources and kernel name. 
This approach would significantly reduce code duplication and improve maintainability as new kernels are added.

PR #18708, opened 2026-01-09
fix text spacing in print_info

ddh0:
In the `print_info:` output during model loading, there was a space missing after `rope_yarn_log_mul`:

```txt
print_info: n_ctx_orig_yarn  = 131072
print_info: rope_yarn_log_mul= 0.0000
print_info: rope_finetuned   = unknown
```

Now it's fixed:

```txt
print_info: n_ctx_orig_yarn   = 131072
print_info: rope_yarn_log_mul = 0.0000
print_info: rope_finetuned    = unknown
```

### DIFF
--- a/src/llama-model.cpp
+++ b/src/llama-model.cpp
@@ -7098,59 +7098,59 @@ void llama_model::print_info() const {
     };
 
     // hparams
-    LLAMA_LOG_INFO("%s: arch              = %s\n",     __func__, arch_name().c_str());
-    LLAMA_LOG_INFO("%s: vocab_only        = %d\n",     __func__, hparams.vocab_only);
-    LLAMA_LOG_INFO("%s: no_alloc          = %d\n",     __func__, hparams.no_alloc);
+    LLAMA_LOG_INFO("%s: arch                  = %s\n",     __func__, arch_name().c_str());
+    LLAMA_LOG_INFO("%s: vocab_only            = %d\n",     __func__, hparams.vocab_only);
+    LLAMA_LOG_INFO("%s: no_alloc              = %d\n",     __func__, hparams.no_alloc);
 
     if (!hparams.vocab_only) {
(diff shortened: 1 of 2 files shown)


CISC:
Ooof, yes, and this was why... :)

Since you're already touching the alignment everywhere AFAICR there are some conditionally printed ones that don't match.

Edit: Ah yes, all the arch-specific ones.

ddh0:
Are there some lines I missed?

CISC:
> Are there some lines I missed?

Not per se, but the arch-specific logs weren't aligned with the rest to begin with.

ddh0:
Gotcha, I will take around look this evening and fix those as well.

ngxson:
I think we can start to abstract this print into a new function, like:

```cpp
void print_info_item(const char * func_name, const char * item_name, ...) {
    // get va_arg and format to string
    LLAMA_LOG_INFO("%s: %-10s = %s", func_name, item_name, value_str);
}
```

so if something is longer in the future, we just need to bump `%-10s` to `%-11s`, etc

ddh0:
Aligned all the lines now. Tested on hybrid model (Jamba Mini), MoE (gpt-oss-120b), SWA (gemma-3-4b), and normal dense model (llama-3.3-70b).

ddh0:
@ngxson, @CISC

Will it be okay to merge like this to fix the immediate issue? Or do you want me to abstract it to a new function like ngxson said?

CISC:
> Will it be okay to merge like this to fix the immediate issue? Or do you want me to abstract it to a new function like ngxson said?

I'm fine either way, if you want to do it, I suggest doing it in this PR, if not we can merge this.

ddh0:
I'll leave the abstraction for someone else in another PR.

CISC (approved):

Issue #18656, opened 2026-01-07
Misc. bug: Vulkan validation error VUID-VkWriteDescriptorSet-descriptorType-00333 and VUID-vkCmdPushConstants-offset-01795


rillomas:
### Name and Version

λ build_vk_validation\bin\Release\llama-cli.exe --version
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = Intel(R) Arc(TM) B580 Graphics (Intel Corporation) | uma: 0 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 32768 | int dot: 1 | matrix cores: KHR_coopmat
version: 7640 (e443fbcfa)
built with MSVC 19.50.35721.0 for x64

### Operating systems

Windows

### Which llama.cpp modules do you know to be affected?

llama-bench

### Command line

```shell
llama-bench.exe -m gemma-3n-E2B-it-Q4_K_M.gguf -fa 1 -r 1 -p 512 -n 0
```

### Problem description & steps to reproduce

I'm seeing Vulkan validation errors on e443fbcfa. I didn't see this before so probably introduced in a recent change. Still need to track down the exact commit.

```
Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-00333 ] | MessageID = 0xf2fc081c
vkUpdateDescriptorSets(): pDescriptorWrites[0].pBufferInfo[0].range (1509949440) is greater than maxStorageBufferRange (1073741820) for descriptorType VK_DESCRIPTOR_TYPE_STORAGE_BUFFER.
The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_STORAGE_BUFFER or VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC, the range member of each element of pBufferInfo, or the effective range if range is VK_WHOLE_SIZE, must be less than or equal to VkPhysicalDeviceLimits::maxStorageBufferRange (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/descriptorsets.html#VUID-VkWriteDescriptorSet-descriptorType-00333)
Objects: 1
    [0] VkBuffer 0x70000000007

Validation Error: [ VUID-vkCmdPushConstants-offset-01795 ] | MessageID = 0x27bc88c6
vkCmdPushConstants(): is called with
[...]

PR #18709, opened 2026-01-09
vulkan: Check maxStorageBufferRange in supports_op

jeffbolznv:
For #18656.

This should reject ops that use tensors that would violate this limit. There's an exception for ops that use BDA, and there will need to be an exception for 64bit_indexing.


0cc4m:
Please rebase and add support for 64-bit indexing.

jeffbolznv:
> Please rebase and add support for 64-bit indexing.

done.

0cc4m (approved):

PR #18710, opened 2026-01-09
Test backend ops: add json printer

water-chika:
Add json prnter

### DIFF
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -462,7 +462,7 @@ enum test_mode {
 };
 
 // Output format support similar to llama-bench
-enum output_formats { CONSOLE, SQL, CSV };
+enum output_formats { CONSOLE, SQL, CSV, JSON };
 
 static const char * output_format_str(output_formats format) {
     switch (format) {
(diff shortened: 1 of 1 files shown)


am17an:
Do you have any downstream application which will use JSON output?

water-chika:
I'd like to add a script used to compare different backend ops perf: [commit](https://github.com/water-chika/llama.cpp/commit/8cfc833f64212a08e991f5d8246dab149acd8463)

am17an:
> I'd like to add a script used to compare different backend ops perf: [commit](https://github.com/water-chika/llama.cpp/commit/8cfc833f64212a08e991f5d8246dab149acd8463)

Why can't you do that with the SQL output for example?

water-chika:
Because I do not have SQL database and it is more convenient to use json for me.
Is there a script to compare SQL output? I'd like to use it if there is one.

am17an:
Yes there is, if you are just wanting to compare performance between two commits you can do something like this
`./scripts/compare-commits.sh commit1 commit2 test-backend-ops <rest of the backend-ops command>`

Internally the script calls `scripts/compare-llama-bench.py` which uses the SQL format to compare.

water-chika:
I want to compare performances of different backends, like Vulkan and CUDA.  Because these actually run in same GPU. 

But it might could get this comparation by modify ```scripts/compare-llama-bench.py```.
Will close this PR.

PR #18711, opened 2026-01-09
Support parsing JSON into grammar for schemas with no type and no properties

markrietveld:
Fixes #18712

These changes work for me when working with Home Assistant `2026.1.0`, and `2025.12.5`. I also haven't found any issues using the llama-server web UI


markrietveld:
I came across https://github.com/home-assistant/core/issues/160462 after putting this up. It looks like the Home Assistant changes are being reverted, so my PR here might be on the wrong path

PR #18719, opened 2026-01-09
model: Add VAETKI support

dororodoroddo:
## Summary

Add support for VAETKI architecture, a Korean MoE language model with MLA (Multi-head Latent Attention).

### Key Features
- **MLA Attention**: Low-rank KV compression for memory efficiency
- **MoE + Shared Experts**: 128 routed experts with sigmoid gating + 1 shared expert
- **Hybrid Attention**: Interleaved sliding window and full attention layers
- **4-Norm Structure**: Four RMSNorm per layer (input, post_attn, pre_mlp, post_mlp)
- **Vision Encoder**: CLIP-based with 2D RoPE for VL variant

### Models Supported
- VAETKI (100B, 48 layers)
- VAETKI-VL-7B-A1B (7B, 24 layers) - Vision-Language model

### Changes
- `convert_hf_to_gguf.py`: VAETKI model conversion with WBLRMSNorm handling
- `src/models/vaetki.cpp`: LLM inference graph with MLA and MoE
- `tools/mtmd/models/vaetki.cpp`: Vision encoder with 2D RoPE
- `src/llama-vocab.cpp`: Metaspace-based BPE tokenizer support

ref: https://huggingface.co/nc-ai-consortium/VAETKI
ref: https://huggingface.co/nc-ai-consortium/VAETKI-VL-7B-A1B
ref: https://huggingface.co/dororodoroddo/VAETKI-GGUF
ref: https://huggingface.co/dororodoroddo/VAETKI-VL-7B-A1B-GGUF

## Test Plan
- [x] Convert VAETKI-VL-7B-A1B to GGUF (F16, Q4_K_M)
- [x] Convert VAETKI-100B to GGUF (F16, Q4_K_M)
- [x] Build test binary with llama.cpp C++ API
- [x] VAETKI-VL-7B-A1B-Q4_K_M text generation (Korean/English, CPU/Metal)
- [x] VAETKI-100B-Q4_K_M text generation (Korean/English, CPU/Metal)

<details>
<summary><b>Test Results</b></summary>

# VAETKI llama.cpp Test Results

## Environment

- **Platform**: macOS Darwin arm64 (Apple Silicon)
- **Test Image**: 
At_the_ticket_gates

---
## VAETKI-VL-7B-A1B-Q4_K_M Text Korean [CPU]

**Prompt**: 안녕! 오늘 기분이 어때? 나는 요즘 일이 너무 많아서 지쳐있어. 뭔가 재미있는 얘기 해줄 수 있어?

**Response**:
```
저는 단순히 대답을 제공하는 것이 아니라, 이 분과 대화하고 소통할 수 있도록 도와드리기 위해 여기 있습니다. 어떤 메시지를 남겨주실지 모르겠지만, 메시지를 보내시면 다음과 같은 방식으로 답변을 드릴 수 있습니다:

1. 긍정적이고 격려적인 말로 기분을 끌어올리기
2. 재미있거나 독특한 이야기를 공유하여 기분을 좋게 하기
3. 업무 관련 문제나 고민을 나누고 도움을 줄 수 있는 방법을 제안하기

어떤 메시지를 남겨보시겠습니까? 😊
```

**Performance**: Prompt 191.0 tok/s | Generation 102.7 tok/s | 127 tokens

---

## VAETKI-VL-7B-A1B-Q4_K_M Text Korean [Metal]

**Performance**: Prompt 6,392.6 tok/s | Generation 156.8 tok/s | 256 tokens

---

## VAETKI-VL-7B-A1B-Q4_K_M Text English [CPU]

**Prompt**: Hey! How are you feeling today? I've been really exhausted from work lately. Can you tell me something fun or interesting?

**Response**:
```
It sounds like you're feeling a bit tired and drained. Let's take a brief break. I'll tell you a fun fact or a quirky story that you might find interesting.

So, you're feeling tired because of your job, but what if I told you that the world's favorite city is not in America, but rather in a distant land? It's a place called Tokyo, and it's where I just spent some time there.

Tokyo is a place where you can find a perfect blend of the ordinary and the extraordinary. It's where you can find a cup of coffee at a quaint café, or a cup of ramen at a street vendor's stall.
```

**Performance**: Prompt 207.5 tok/s | Generation 106.4 tok/s | 256 tokens

---

## VAETKI-VL-7B-A1B-Q4_K_M Text English [Metal]

**Performance**: Prompt 31,008.8 tok/s | Generation 165.8 tok/s | 119 tokens

---

## VAETKI-100B-Q4_K_M Text Korean [CPU]

**Prompt**: 안녕! 오늘 기분이 어때? 나는 요즘 일이 너무 많아서 지쳐있어. 뭔가 재미있는 얘기 해줄 수 있어?

**Response**:
```
안녕! 나는 AI라서 감정은 없지만, 네 기분을 이해하려고 노력할게! 😊 요즘 일이 많아서 지쳤다니, 정말 힘들었겠다. 나도 네가 휴식을 취하고 기운을 낼 수 있는 재미있는 이야기를 준비해봤어!

**한 번 들어볼래?**

---

**1. "코끼리와 핫도그" 이야기**
한 코끼리가 동물원에서 핫도그를 사 먹으려고 했어. 그런데 핫도그 가게 주인이 말했어.
"코끼리는 핫도그를 살 수 없어! 너는 너무 커서 핫도그를 다 먹지도 못할 거야!"
코끼리는 슬퍼하며 집에 돌아왔어. 그런데 다음 날, 가게 주인이 코끼리에게 전화를 걸었어.
"코끼리야! 너를 위한 특별 핫도그를 만들었어! 너를 위한 '초대형 핫도그'야!"
코끼리는 기뻐하며 핫도그를 받았어. 그런데 핫도그를 먹으려는 순간, 가게 주인이 말했어.
"잠깐! 코끼리는 핫도그를 먹으
```

**Performance**: Prompt 2.6 tok/s | Generation 12.7 tok/s | 256 tokens

---

## VAETKI-100B-Q4_K_M Text Korean [Metal]

**Performance**: Prompt 5,715.3 tok/s | Generation 46.2 tok/s | 256 tokens

---

## VAETKI-100B-Q4_K_M Text English [CPU]

**Prompt**: Hey! How are you feeling today? I've been really exhausted from work lately. Can you tell me something fun or interesting?

**Response**:
```
Hey there! I'm doing well, thanks for asking—though I'm just an AI, so I don't get tired, but I can totally relate to needing a fun break from work. 😄

Here's something fun and interesting for you:
Did you know that **an octopus has three hearts**? Two of them pump blood to the gills, while the third pumps it to the rest of the body. Even cooler: When they swim, the heart that feeds their body actually *stops beating* to conserve energy! They're basically underwater superheroes. 🐙

Hope that gives you a little smile and a fun fact to share. You've got this—take a breather when you can! 🌟
```

**Performance**: Prompt 15.4 tok/s | Generation 16.8 tok/s | 162 tokens

---

## VAETKI-100B-Q4_K_M Text English [Metal]

**Performance**: Prompt 15,799.3 tok/s | Generation 44.9 tok/s | 256 tokens

---

### VAETKI-VL-7B-A1B-Q4_K_M

| Test | CPU Prompt | Metal Prompt | CPU Gen | Metal Gen | Prompt Speedup | Gen Speedup |
|------|-----------|--------------|---------|-----------|----------------|-------------|
| Text Korean | 191.0 | 6,392.6 | 102.7 | 156.8 | **33x** | **1.5x** |
| Text English | 207.5 | 31,008.8 | 106.4 | 165.8 | **149x** | **1.6x** |

*(Units: tok/s)*

---

### VAETKI-100B-Q4_K_M

| Test | CPU Prompt | Metal Prompt | CPU Gen | Metal Gen | Prompt Speedup | Gen Speedup |
|------|-----------|--------------|---------|-----------|----------------|-------------|
| Text Korean | 2.6 | 5,715.3 | 12.7 | 46.2 | **2,198x** | **3.6x** |
| Text English | 15.4 | 15,799.3 | 16.8 | 44.9 | **1,026x** | **2.7x** |

*(Units: tok/s)*

</detail>


tools/mtmd/clip.cpp
```diff
@@ -3371,6 +3394,31 @@ bool clip_image_batch_encode(clip_ctx * ctx, const int n_threads, const clip_ima
                     }
                 }
 
+                set_input_i32("positions", positions);
+            } break;
+        case PROJECTOR_TYPE_VAETKI:
```
ngxson: merge this with the case above it. do not copy

tools/mtmd/clip.cpp
```diff
@@ -1318,6 +1329,7 @@ struct clip_model_loader {
         };
 
         model.class_embedding = get_tensor(TN_CLASS_EMBD, false);
+        model.class_pos_emb   = get_tensor(TN_CLASS_POS_EMBD, false);
```
ngxson: only your model need it, move this to the `case PROJECTOR_TYPE_VAETKI`

tools/mtmd/models/vaetki.cpp
```diff
@@ -0,0 +1,185 @@
...
+        cls_cos = ggml_cos(ctx0, cls_pos);
+        cls_sin = ggml_sin(ctx0, cls_pos);
+    }
+
+    if (model.pre_ln_w) {
+        inpL = build_norm(inpL, model.pre_ln_w, model.pre_ln_b, norm_t, eps, -1);
+        cb(inpL, "pre_ln", -1);
+    }
+
+    for (int il = 0; il < n_layer; il++) {
```
ngxson: instead of copy qwen's code, try using `build_vit` instead. see `glm4v.cpp`

tools/mtmd/models/vaetki.cpp
```diff
@@ -0,0 +1,185 @@
...
+                ggml_row_size(Kcur->type, d_head * n_head), 0);
+
+            ggml_tensor * Q_patch = ggml_view_3d(ctx0, Qcur, d_head, n_head, n_pos_patches,
+                ggml_row_size(Qcur->type, d_head),
+                ggml_row_size(Qcur->type, d_head * n_head),
+                ggml_row_size(Qcur->type, d_head * n_head));
+            ggml_tensor * K_patch = ggml_view_3d(ctx0, Kcur, d_head, n_head, n_pos_patches,
+                ggml_row_size(Kcur->type, d_head),
+                ggml_row_size(Kcur->type, d_head * n_head),
+                ggml_row_size(Kcur->type, d_head * n_head));
```
ngxson: this implementation is quite naive. I don't have time to optimize it, but just keep in mind that it is suboptimal and will be slower compared to transformers

tools/mtmd/clip.cpp
```diff
@@ -2842,6 +2863,7 @@ bool clip_image_preprocess(struct clip_ctx * ctx, const clip_image_u8 * img, str
         case PROJECTOR_TYPE_GLM_EDGE:
         case PROJECTOR_TYPE_GEMMA3:
         case PROJECTOR_TYPE_INTERNVL: // TODO @ngxson : support dynamic resolution
+        case PROJECTOR_TYPE_VAETKI:
```
ngxson: are you sure? this will force the model to use fixed size image

does the model support dynamic image size? if not, what's the reason behind copying qwen-vl implementation in the first place?

ngxson (changes requested):

src/models/vaetki.cpp
```diff
@@ -0,0 +1,194 @@
...
+            ggml_tensor * q = ggml_mul_mat(ctx0, model.layers[il].wq_a, cur);
+            cb(q, "q_a", il);
+
+            q = build_norm(q, model.layers[il].attn_q_a_norm, nullptr, LLM_NORM_RMS, il);
+            cb(q, "q_a_norm", il);
+
+            q = ggml_mul_mat(ctx0, model.layers[il].wq_b, q);
+            cb(q, "q", il);
+
+            ggml_tensor * q_nope =
```
pwilkin: This is all unnecessary. You're already passing the rope dimension parameter, which deals with partial rope internally (the `n_rot` parameter determines which part of the tensor is actually rotated). You don't have to copy the transformers logic line-by-line by splitting and then concatenating the tensors.

ngxson: hmm ok indeed this is copied from `deepseek2.cpp`. It's quite uncool not to disclosure where this code was copied from.

this split is mostly for deepseek MLA implementation where part of the Q/K is used for another purpose.

I think instead of blindly copy deepseek code, we can simply add `iswa` support via to deepseek2 arch. unless I missed something, this model seems like just deepseek2 + SWA

CISC: > unless I missed something, this model seems like just deepseek2 + SWA

I suspect so, but haven't had time to verify.

dororodoroddo: My apologies. Since the model appears to be DeepSeek-V2 based, I developed the converter by combining various architectures. That is why the implementation involves an integration of multiple structures.

dororodoroddo:
I've just added the missing encoder conversion code. This part is actually quite similar to Qwen2VL, with the key difference being that VAETKI (Rice ViT) uses a CLS token while Qwen2VL does not

tools/mtmd/clip.cpp
dororodoroddo: After merging, only added the line for VAETKI position size handling:
`const int pos_size = (model.proj_type == PROJECTOR_TYPE_VAETKI) ? num_patches : n_pos;`

tools/mtmd/clip.cpp
dororodoroddo: I missed that and have moved the code as requested.

tools/mtmd/models/vaetki.cpp
dororodoroddo: I have refactored the code to use build_vit as requested; could you please check if this approach is acceptable?

tools/mtmd/models/vaetki.cpp
dororodoroddo: I understand why it is slow, but I don't have the ability to improve it. I will keep that in mind.

tools/mtmd/clip.cpp
dororodoroddo: I missed that point as I was initially focused on cls and took the wrong direction. I've since moved towards the qwen2vl implementation.

src/models/vaetki.cpp
dororodoroddo: //
Thanks for the advice. I resolved this by reordering the Q weights from [nope | rope] to [rope | nope] during GGUF conversion to utilize ggml_rope_ext.

ngxson: Could you point out the differences between this model and deepseek2 (MLA) ?

tools/mtmd/clip.cpp
ngxson: I'm now a bit doubt if we really need to copy everything from qwen model.

Can you try using `build_rope_2d` instead of `ggml_rope_multi`? See `pixtral.cpp` for example

If it works, the current model can be moved to `PROJECTOR_TYPE_PIXTRAL` section

ngxson: Hmm nevermind, that won't work because you're using the same RoPE ordering as qwen-vl

tools/mtmd/clip.cpp
```diff
@@ -3482,11 +3506,14 @@ bool clip_image_batch_encode(clip_ctx * ctx, const int n_threads, const clip_ima
...
         case PROJECTOR_TYPE_QWEN3VL:
         case PROJECTOR_TYPE_GLM4V:
+        case PROJECTOR_TYPE_VAETKI:
             {
                 const int merge_ratio = hparams.n_merge;
                 const int pw = image_size_width  / patch_size;
                 const int ph = image_size_height / patch_size;
-                std::vector<int> positions(n_pos * 4);
+
+                const int pos_size = (model.proj_type == PROJECTOR_TYPE_VAETKI) ? num_patches : n_pos;
```
ngxson: Since you said that qwen doesn't use CLS token, it's ok to just change it to `num_patches` here

```suggestion
                const int pos_size = num_patches;
```

The use of `n_pos` is probably a mistake from another contributor, it's not valid with +1 value for CLS anyway

tools/mtmd/clip.cpp
```diff
@@ -1197,6 +1201,14 @@ struct clip_model_loader {
                         hparams.set_limit_image_tokens(8, 4096);
                         hparams.set_warmup_n_tokens(46*46); // avoid OOM on warmup
                     } break;
+                case PROJECTOR_TYPE_VAETKI:
+                    {
+                        hparams.rope_theta = 10000.0f;
+                        hparams.n_merge = 2;
+                        get_u32(KEY_SPATIAL_MERGE_SIZE, hparams.n_merge, false);
+                        hparams.set_limit_image_tokens(4, 3265);
+                        hparams.set_warmup_n_tokens(40*40);
```
ngxson: your model doesn't support dynamic image size, no need to set these params

```suggestion
```

tools/mtmd/clip.cpp
```diff
@@ -1545,6 +1557,16 @@ struct clip_model_loader {
                     model.mm_1_w = get_tensor(string_format(TN_LLAVA_PROJ, 2, "weight"));  // merger.mlp.2
                     model.mm_1_b = get_tensor(string_format(TN_LLAVA_PROJ, 2, "bias"));
                 } break;
+            case PROJECTOR_TYPE_VAETKI:
+                {
+                    model.class_pos_emb = get_tensor(TN_CLASS_POS_EMBD);
+                    model.mm_0_w = get_tensor(string_format(TN_MVLM_PROJ_MLP, 0, "weight"));
+                    model.mm_0_b = get_tensor(string_format(TN_MVLM_PROJ_MLP, 0, "bias"));
```
ngxson: this naming is incorrect, the numbered naming is reserved for FFN tensors. for input norm, use `mm.input_norm` instead

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,328 @@ def prepare_tensors(self):
...
+        # Handle class_embedding and class_pos_emb (keep model.visual. prefix for mapping)
+        if "class_embedding" in name or "class_pos_emb" in name:
+            return [(self.map_tensor_name(name), data_torch)]
+
+        # Strip model.visual. -> visual. for other tensors
+        if name.startswith("model.visual."):
+            name = name.replace("model.visual.", "visual.")
+
+        # Split fused QKV tensors
+        if ".qkv." in name:
```
ngxson: fused qkv is supported by `build_vit`. no need to split them

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,328 @@ def prepare_tensors(self):
...
+            suffix = "weight" if name.endswith(".weight") else "bias"
+            return [(f"mm.model.mlp.0.{suffix}", data_torch)]
+        elif "merger.mlp.0" in name:
+            # mlp.0 -> mm.model.mlp.1 (up projection)
+            suffix = "weight" if name.endswith(".weight") else "bias"
+            return [(f"mm.model.mlp.1.{suffix}", data_torch)]
+        elif "merger.mlp.2" in name:
+            # mlp.2 -> mm.model.mlp.3 (down projection)
+            suffix = "weight" if name.endswith(".weight") else "bias"
+            return [(f"mm.model.mlp.3.{suffix}", data_torch)]
```
ngxson: use `self.format_tensor_name` for it:
- merger.ln_q --> `V_MM_INP_NORM`
- `mm.model.mlp.1` --> `V_MM_UP`
- `mm.model.mlp.3.` --> `V_MM_DOWN`

src/models/vaetki.cpp
dororodoroddo: Here's a diff code showing the key differences
<details>
<summary>diff</summary>

```diff
  #include "models.h"

  llm_build_vaetki::llm_build_vaetki(const llama_model & model, const llm_graph_params & params) :
      llm_graph_context(params) {

- bool is_lite = (hparams.n_layer == 27 || hparams.n_layer == 26);  // deepseek2.cpp:6~12
- const bool is_mla = (hparams.n_embd_head_k_mla != 0 && hparams.n_embd_head_v_mla != 0);

- const int64_t n_embd_head_k = is_mla ? hparams.n_embd_head_k_mla : hparams.n_embd_head_k;
- const int64_t n_embd_head_v = is_mla ? hparams.n_embd_head_v_mla : hparams.n_embd_head_v;
+ const int64_t n_embd_head_k_mla = hparams.n_embd_head_k_mla;     //  always MLA (no lite variant)
+ const int64_t n_embd_head_v_mla = hparams.n_embd_head_v_mla;

      const int64_t n_embd_head_qk_rope = hparams.n_rot;
- const int64_t n_embd_head_qk_nope = n_embd_head_k - n_embd_head_qk_rope;  // deepseek2.cpp:15
+ const int64_t n_embd_head_qk_nope = n_embd_head_k_mla - n_embd_head_qk_rope;

      const uint32_t kv_lora_rank = hparams.n_lora_kv;

- GGML_ASSERT(ext_factor >= 0.0f);                                         // deepseek2.cpp:24~29
- const float attn_factor_org = attn_factor * (1.0f + 0.1f * logf(1.0f / freq_scale));
- const float mscale = attn_factor_org * (1.0f + 0.1f * hparams.rope_yarn_log_mul * logf(1.0f / freq_scale));
- const float kq_scale = 1.0f * mscale * mscale / sqrtf(float(n_embd_head_k));
+ const float kq_scale = 1.0f / sqrtf(float(n_embd_head_qk_nope + n_embd_head_qk_rope));

      ggml_tensor * cur;
      ggml_tensor * inpL;

      inpL = build_inp_embd(model.tok_embd);

- ggml_tensor * inp_attn_scale = nullptr;                           // deepseek2.cpp:38~41
- if (hparams.f_attn_temp_scale != 0.0f) {
-     inp_attn_scale = build_inp_attn_scale();
- }

      ggml_tensor * inp_pos = build_inp_pos();

- auto * inp_attn = build_attn_inp_kv();                            // deepseek2.cpp:46
+ auto * inp_attn = build_attn_inp_kv_iswa();                       //  iSWA

      ggml_tensor * inp_out_ids = build_inp_out_ids();

      for (int il = 0; il < n_layer; ++il) {
+ const float freq_base_l  = model.get_rope_freq_base(cparams, il);   //  per-layer RoPE for hybrid attention
+ const float freq_scale_l = model.get_rope_freq_scale(cparams, il);

          ggml_tensor * inpSA = inpL;

          cur = build_norm(inpL, model.layers[il].attn_norm, NULL, LLM_NORM_RMS, il);
          cb(cur, "attn_norm", il);

          // self_attention
          {
- ggml_tensor * q = NULL;                                           // deepseek2.cpp:59~72
- if (!is_lite) {
-     q = ggml_mul_mat(ctx0, model.layers[il].wq_a, cur);
-     cb(q, "q", il);
-     q = build_norm(q, model.layers[il].attn_q_a_norm, nullptr, LLM_NORM_RMS, il);
-     cb(q, "q", il);
-     q = ggml_mul_mat(ctx0, model.layers[il].wq_b, q);
-     cb(q, "q", il);
- } else {
-     q = ggml_mul_mat(ctx0, model.layers[il].wq, cur);
-     cb(q, "q", il);
- }
+ ggml_tensor * q = ggml_mul_mat(ctx0, model.layers[il].wq_a, cur); //  always MLA (no lite variant)
+ cb(q, "q_a", il);
+
+ q = build_norm(q, model.layers[il].attn_q_a_norm, nullptr, LLM_NORM_RMS, il);
+ cb(q, "q_a_norm", il);
+
+ q = ggml_mul_mat(ctx0, model.layers[il].wq_b, q);
+ cb(q, "q", il);

- // split into {n_embd_head_qk_nope, n_head, n_tokens}             // deepseek2.cpp:73~83
- ggml_tensor * q_nope = ggml_view_3d(ctx0, q, n_embd_head_qk_nope, n_head, n_tokens,
-     ggml_row_size(q->type, n_embd_head_k),
-     ggml_row_size(q->type, n_embd_head_k) * n_head, 0);
- cb(q_nope, "q_nope", il);
- // and {n_embd_head_qk_rope, n_head, n_tokens}
- ggml_tensor * q_pe = ggml_view_3d(ctx0, q, n_embd_head_qk_rope, n_head, n_tokens,
-     ggml_row_size(q->type, n_embd_head_k),
-     ggml_row_size(q->type, n_embd_head_k) * n_head,
-     ggml_row_size(q->type, n_embd_head_qk_nope));
- cb(q_pe, "q_pe", il);
+ // q is now [rope | nope] after weight reordering in conversion   //  for use ggml_rope_ext partial rotation
+ // reshape to {n_embd_head_k_mla, n_head, n_tokens}
+ q = ggml_reshape_3d(ctx0, q, n_embd_head_k_mla, n_head, n_tokens);

              ggml_tensor * kv_cmpr_pe = ggml_mul_mat(ctx0, model.layers[il].wkv_a_mqa, cur);
              cb(kv_cmpr_pe, "kv_cmpr_pe", il);

              // {kv_lora_rank, n_tokens}
              ggml_tensor * kv_cmpr = ggml_view_2d(ctx0, kv_cmpr_pe,
                  kv_lora_rank, n_tokens,
                  ggml_row_size(kv_cmpr_pe->type, kv_lora_rank + n_embd_head_qk_rope), 0);
              cb(kv_cmpr, "kv_cmpr", il);

              // {n_embd_head_qk_rope, 1, n_tokens}
              ggml_tensor * k_pe = ggml_view_3d(ctx0, kv_cmpr_pe,
                  n_embd_head_qk_rope, 1, n_tokens,
                  ggml_row_size(kv_cmpr_pe->type, kv_lora_rank + n_embd_head_qk_rope),
                  ggml_row_size(kv_cmpr_pe->type, kv_lora_rank + n_embd_head_qk_rope),
                  ggml_row_size(kv_cmpr_pe->type, kv_lora_rank));
              cb(k_pe, "k_pe", il);

- q_pe = ggml_rope_ext(ctx0, q_pe, inp_pos, nullptr, n_rot, rope_type, n_ctx_orig,  // deepseek2.cpp:101~107
-     freq_base, freq_scale, ext_factor, attn_factor, beta_fast, beta_slow);
- cb(q_pe, "q_pe", il);
-
- k_pe = ggml_rope_ext(ctx0, k_pe, inp_pos, nullptr, n_rot, rope_type, n_ctx_orig,
-     freq_base, freq_scale, ext_factor, attn_factor, beta_fast, beta_slow);
- cb(k_pe, "k_pe", il);
+ // apply rope - rotates first n_rot dims, copies rest unchanged   //  for use ggml_rope_ext(n_rot)
+ ggml_tensor * Qcur = ggml_rope_ext(ctx0, q, inp_pos, nullptr, n_rot, rope_type, n_ctx_orig,
+     freq_base_l, freq_scale_l, ext_factor, attn_factor, beta_fast, beta_slow);  //  per-layer freq
+ cb(Qcur, "Qcur", il);
+
+ k_pe = ggml_rope_ext(ctx0, k_pe, inp_pos, nullptr, n_rot, rope_type, n_ctx_orig,
+     freq_base_l, freq_scale_l, ext_factor, attn_factor, beta_fast, beta_slow);  //  per-layer freq
+ cb(k_pe, "k_pe_rope", il);

              kv_cmpr = build_norm(kv_cmpr, model.layers[il].attn_kv_a_norm, nullptr, LLM_NORM_RMS, il);
              cb(kv_cmpr, "kv_cmpr_norm", il);

- if (is_mla) {                                                     // deepseek2.cpp:112~150
-     // {n_embd_head_qk_nope, n_tokens, n_head}
-     q_nope = ggml_permute(ctx0, q_nope, 0, 2, 1, 3);
-     cb(q_nope, "q_nope_perm", il);
-
-     // {n_embd_head_qk_nope, kv_lora_rank, n_head} x {n_embd_head_qk_nope, n_tokens, n_head}
-     ggml_tensor * q_nope_absorbed = ggml_mul_mat(ctx0, model.layers[il].wk_b, q_nope);
-     cb(q_nope_absorbed, "q_nope_absorbed", il);
-
-     // {kv_lora_rank, n_head, n_tokens}
-     q_nope_absorbed = ggml_permute(ctx0, q_nope_absorbed, 0, 2, 1, 3);
-     cb(q_nope_absorbed, "q_nope_absorbed_perm", il);
-
-     // {n_embd_head_qk_rope + kv_lora_rank, n_head, n_tokens}
-     ggml_tensor * Qcur = ggml_concat(ctx0, q_pe, q_nope_absorbed, 0);
-     cb(Qcur, "Qcur", il);
-
-     kv_cmpr = ggml_reshape_3d(ctx0, kv_cmpr, kv_lora_rank, 1, n_tokens);
-     cb(kv_cmpr, "kv_cmpr_reshape", il);
-
-     // {n_embd_head_qk_rope + kv_lora_rank, 1, n_tokens}
-     ggml_tensor * Kcur = ggml_concat(ctx0, k_pe, kv_cmpr, 0);
-     cb(Kcur, "Kcur", il);
-
-     // {kv_lora_rank, 1, n_tokens}
-     ggml_tensor * Vcur = kv_cmpr;
-     cb(Vcur, "Vcur", il);
-
-     if (inp_attn_scale) {
-         Qcur = ggml_mul(ctx0, Qcur, inp_attn_scale);
-         cb(Qcur, "Qcur_attn_temp_scaled", il);
-     }
-
-     // MLA with the absorption optimization converts into MQA
-     cur = build_attn(inp_attn,
-             model.layers[il].wo, NULL,
-             Qcur, Kcur, Vcur, nullptr, nullptr, model.layers[il].wv_b, kq_scale, il);
- } else {
              ggml_tensor * kv = ggml_mul_mat(ctx0, model.layers[il].wkv_b, kv_cmpr);
              cb(kv, "kv", il);

              // {n_embd_head_qk_nope, n_head, n_tokens}
              ggml_tensor * k_nope = ggml_view_3d(ctx0, kv,
                  n_embd_head_qk_nope, n_head, n_tokens,
                  ggml_row_size(kv->type, n_embd_head_qk_nope + n_embd_head_v_mla),
                  ggml_row_size(kv->type, n_embd_head_qk_nope + n_embd_head_v_mla) * n_head, 0);
              cb(k_nope, "k_nope", il);

              // {n_embd_head_v_mla, n_head, n_tokens}
              ggml_tensor * Vcur = ggml_view_3d(ctx0, kv,
                  n_embd_head_v_mla, n_head, n_tokens,
                  ggml_row_size(kv->type, n_embd_head_qk_nope + n_embd_head_v_mla),
                  ggml_row_size(kv->type, n_embd_head_qk_nope + n_embd_head_v_mla) * n_head,
                  ggml_row_size(kv->type, n_embd_head_qk_nope));
              cb(Vcur, "Vcur", il);

              Vcur = ggml_cont(ctx0, Vcur);
              cb(Vcur, "Vcur_cont", il);

- ggml_tensor * Qcur = ggml_concat(ctx0, q_pe, q_nope, 0);          // deepseek2.cpp:173~177
- cb(Qcur, "Qcur", il);
- ggml_tensor * Kcur = ggml_concat(ctx0, ggml_repeat(ctx0, k_pe, q_pe), k_nope, 0);
- cb(Kcur, "Kcur", il);
+ ggml_tensor * q_pe_ref = ggml_view_3d(ctx0, Qcur,                 //  for use ggml_repeat shape ref
+     n_embd_head_qk_rope, n_head, n_tokens, Qcur->nb[1], Qcur->nb[2], 0);
+ ggml_tensor * Kcur = ggml_concat(ctx0, ggml_repeat(ctx0, k_pe, q_pe_ref), k_nope, 0);
+ cb(Kcur, "Kcur", il);

- if (inp_attn_scale) {                                             // deepseek2.cpp:179~183
-     Qcur = ggml_mul(ctx0, Qcur, inp_attn_scale);
-     cb(Qcur, "Qcur_attn_temp_scaled", il);
- }

              cur = build_attn(inp_attn,
                      model.layers[il].wo, NULL,
                      Qcur, Kcur, Vcur, nullptr, nullptr, nullptr, kq_scale, il);
- }  // end if (is_mla) else
          }

+ cur = build_norm(cur, model.layers[il].attn_post_norm, NULL, LLM_NORM_RMS, il);  //  4-norm structure
+ cb(cur, "attn_post_norm", il);

- if (il == n_layer - 1 && inp_out_ids) {                           // deepseek2.cpp:191~196
-     cur   = ggml_get_rows(ctx0, cur, inp_out_ids);
-     inpSA = ggml_get_rows(ctx0, inpSA, inp_out_ids);
- }
- ggml_tensor * ffn_inp = ggml_add(ctx0, cur, inpSA);
- cb(ffn_inp, "ffn_inp", il);
+ if (il == n_layer - 1 && inp_out_ids) {                           //  moved after post_norm
+     cur   = ggml_get_rows(ctx0, cur, inp_out_ids);
+     inpSA = ggml_get_rows(ctx0, inpSA, inp_out_ids);
+ }
+
+ ggml_tensor * ffn_inp = ggml_add(ctx0, cur, inpSA);
+ cb(ffn_inp, "ffn_inp", il);

          cur = build_norm(ffn_inp, model.layers[il].ffn_norm, NULL, LLM_NORM_RMS, il);
          cb(cur, "ffn_norm", il);

          if ((uint32_t) il < hparams.n_layer_dense_lead) {
              cur = build_ffn(cur,
                  model.layers[il].ffn_up, NULL, NULL,
                  model.layers[il].ffn_gate, NULL, NULL,
                  model.layers[il].ffn_down, NULL, NULL,
                  NULL, LLM_FFN_SILU, LLM_FFN_PAR, il);
              cb(cur, "ffn_out", il);
          } else {
              // MoE branch
              ggml_tensor * moe_out = build_moe_ffn(cur,
                  model.layers[il].ffn_gate_inp,
                  model.layers[il].ffn_up_exps,
                  model.layers[il].ffn_gate_exps,
                  model.layers[il].ffn_down_exps,
- model.layers[il].ffn_exp_probs_b,                                 // deepseek2.cpp:215
+ nullptr,
                  n_expert, n_expert_used,
                  LLM_FFN_SILU, hparams.expert_weights_norm,
                  hparams.expert_weights_scale, hparams.expert_weights_scale,
                  (llama_expert_gating_func_type) hparams.expert_gating_func,
                  il);
              cb(moe_out, "ffn_moe_out", il);

              // FFN shared expert
              ggml_tensor * ffn_shexp =
                  build_ffn(cur,
                      model.layers[il].ffn_up_shexp, NULL, NULL,
                      model.layers[il].ffn_gate_shexp, NULL, NULL,
                      model.layers[il].ffn_down_shexp, NULL, NULL,
                      NULL, LLM_FFN_SILU, LLM_FFN_PAR, il);
              cb(ffn_shexp, "ffn_shexp", il);

              cur = ggml_add(ctx0, moe_out, ffn_shexp);
              cb(cur, "ffn_out", il);
          }

+ cur = build_norm(cur, model.layers[il].ffn_post_norm, NULL, LLM_NORM_RMS, il);  //  4-norm structure
+ cb(cur, "ffn_post_norm", il);

          cur = ggml_add(ctx0, cur, ffn_inp);

          cur = build_cvec(cur, il);
          cb(cur, "l_out", il);

          inpL = cur;
      }
      cur = inpL;

      cur = build_norm(cur, model.output_norm, NULL, LLM_NORM_RMS, -1);

      cb(cur, "result_norm", -1);
      res->t_embd = cur;

      cur = ggml_mul_mat(ctx0, model.output, cur);

      cb(cur, "result_output", -1);
      res->t_logits = cur;

      ggml_build_forward_expand(gf, cur);
  }
```

</details>

tools/mtmd/clip.cpp
dororodoroddo: Updated as requested

tools/mtmd/clip.cpp
dororodoroddo: Oh, the previous code was a mistake, I’ve corrected it by switching to qwen2VL. 
Dynamic image size is indeed supported, and I’ve verified the behavior across minimum, maximum, and intermediate image sizes.

tools/mtmd/clip.cpp
dororodoroddo: Updated as requested

convert_hf_to_gguf.py
dororodoroddo: Updated as requested

convert_hf_to_gguf.py
dororodoroddo: Removed the split logic

dororodoroddo: I tried removing the split logic as suggested, and while the build was successful, I found that the model is no longer recognizing images correctly.

dororodoroddo: !! I found the root cause and fixed it. Please disregard my previous comment.

src/models/vaetki.cpp
ngxson: We expect you to explain with your words to proves that you understand it. Otherwise, it adds no values if you just give us the diff, which we can easily do it ourselves.

Also, if you used AI to generate part of the code, you MUST explicitly state that.

dororodoroddo: Oh, I apologize. 
To write this code(src/models/vaetki.cpp), I used AI to analyze the model's structure and search for similar structures, and AI was also used for formatting.
First, let me explain my understanding of the structure. Vaetki uses MLA and MOE to compress the KV cache, which I believe is similar to DEEPSEEK-V2. Based on this, I refined it through trial and error.
Then, I observed that the config values ​​for lite only had 24 and 48 layers (without 26/27), and that MLA was used. Therefore, I applied the code that always uses MLA.

Yarn was not used in the config, and the scaling formula was based on HF/modeling_vaetki_vl.py.

MLA Absorption was removed after errors occurred, and the 'line-by-line by splitting' code in the RoPE structure was taken from modeling_vaetki_vl.py.

The following is the basis for VAETKI's characteristics:

<details>

<summary>ref : technical_report 2.1 architecture</summary>

The VAETKI models are fundamentally based on a decoder-only Transformer architecture, following
an autoregressive language modeling formulation. Each model consists of a stack of Transformer
layers, where every layer is composed of a causal self-attention module and a Mixture-of-Experts
(MoE) based feed-forward network. This design enables efficient scaling while preserving strong
reasoning capabilities under sparse activation.
Attention Mechanism. Instead of conventional multi-head attention, we employ a refined
attention structure that integrates Multi-Latent Attention (MLA) with a Local–Global Interleaving
strategy to optimize both memory and computational efficiency. While MLA [1] significantly improves
memory efficiency by compressing the KV context and maintains a stronger representational capacity
compared to Grouped Query Attention (GQA) [2], it does not inherently address the quadratic
computational complexity of the attention mechanism.
To manage this, we incorporate an interleaving approach inspired by the structural efficiency of
Gemma 2 [3]. This design is based on the observation that expanding the receptive field inevitably
leads to the dispersion of softmax attention weights [4, 5], a structural characteristic that often
yields diminishing returns in long-context model performance. Recognizing this dispersion as a
fundamental property rather than a problem to be solved, we adopt a pragmatic stance by reducing
redundant global references. By strategically interleaving local sliding window attention with periodic global attention, we focus computational resources on the most essential dependencies,
effectively lowering the computational burden during both training and inference.
In practice, we follow the configuration established in Gemma 3 [6], which observed that the
interleaving frequency does not lead to meaningful degradation in language modeling perplexity;
accordingly, we retain a local-to-global ratio of 5:1. To support this dual-pattern approach, we apply
Rotary Positional Embeddings (RoPE) with decoupled base frequencies, specifically setting θ= 10K
for local attention to maintain high-resolution local focus and θ= 1M for global attention to preserve
long-range dependency capabilities. By integrating these cross-architectural optimizations, our final
configuration achieves an approximately 83% reduction in KV cache size compared to the standard
MLA implementation when handling a sequence length of 128K.
Empirical Evaluation of Local-Global Interleaved MLA. To evaluate the efficiency gains
of Local–Global Interleaved MLA, we conducted empirical experiments using eight NVIDIA H100
NVL PCIe GPUs. The evaluation was performed with a sequence length of 32k tokens and a hidden
dimension of 3072. The MLA hyperparameters follow standard settings, with qk dim set to 196
and v dim set to 128. Low-rank adaptation is applied with q lora rank set to 1536, corresponding
to half of the hidden dimension, and kv lora rank set to 512, corresponding to one-sixth of the
hidden dimension. and FlashAttention-3 [7] is used as the attention backend. Under this setup,
we observe that the configuration reaches a performance sweet spot at approximately the 36th
Transformer layer, beyond which the achieved TFLOPs per second per GPU becomes saturated.
With the evaluated 100B-scale model, this design yields an additional training efficiency gain of
approximately 40%, demonstrating the practical benefits of the proposed attention architecture for
large-scale MoE models. The empirical results are summarized in Table 2. Efficiency is computed as
(Tvanilla−Tinterleaved)/Tvanilla, where Tvanilla and Tinterleaved denote the time per iteration of Vanilla
MLA and Local–Global Interleaved MLA, respectively.
Feed-Forward Network and MoE Design. The feed-forward networks adopt a fine-grained
Mixture-of-Experts (MoE) architecture [8], which has become a dominant design choice for large-
scale language models. In the largest configuration, VAETKI-100B-A10B, each MoE layer consists
of 128 experts together with a single shared expert. For each token, 8 experts are dynamically
activated and their outputs are combined via a weighted sum based on routing probabilities. The
final output is then derived by additively combining this result with the output of the shared expert.
The resulting output representation is configured to be symmetric with respect to the dense FFN
intermediate size, ensuring consistent dimensionality across dense and sparse layers.
Expert routing is performed based on affinity scores computed using a sigmoid function. Instead of grouping-based top-k selection, we adopt the Global Auxiliary Loss [9] for expert routing, and
further augment it with a conventional auxiliary loss to encourage balanced expert utilization. Both
auxiliary losses are applied with a coefficient of 1 ×10−4
.
Normalization and FP8-ready Training Stability. Layer normalization applied to both
attention modules and feed-forward networks following the structure introduced in Gemma 2 [3],
which is also referred to as Peri-LN [10]. This structure has been widely adopted in recent open-
weight large language models such as Gemma 3 and OLMo 2 [11], to improve early-stage training
stability. Given that the scale of training tokens is critical to final performance, we utilize FP8
(E4M3) precision with NVIDIA Transformer Engine [12] for the linear layers within both MoE and
dense feed-forward networks to accelerate training throughput.
During our exploration of FP8 training recipes, we observed that the blockwise scaling strategy, as
presented in DeepSeek-V3 [13], could induce instabilities during the very early stages of training.
To address this, we employ delayed scaling during the first stage of our three-stage pre-training
curriculum (see Section 3.2), subsequently transitioning to blockwise scaling in the later stages to
improve training efficiency. Furthermore, we employ a Zero-Centered RMSNorm [14] formulation
to encourage activation distributions to cluster around zero, significantly enhancing numerical
robustness in large-scale MoE settings under low-precision constraints.

</details>
.
.
.
After running the model, I felt confident enough to submit a PR. During the PR process, I received additional suggestions and modifications beyond what I already understood. To address these issues, I briefly reviewed and understood the proposed function structure, and implemented the modifications. However, this was based on my limited understanding, and it would be more accurate to say I "didn't understand" or "couldn't directly apply it."

After rereading the Vaetki documentation and reviewing the code, I realized that I had merely copied what I thought I understood. Therefore, I thought it would be best to withdraw the PR and ask someone with more knowledge of GGML code and Vaetki (preferably a Vaetki developer) to submit a PR. However, I'd like to ask if withdrawing the PR would help minimize confusion.
(I will keep the forked repository and the branch/huggingface GGUF file for future reference.)

ngxson:
@dororodoroddo the `size.longest_edge/shortest_edge` in [preprocessor_config.json](https://huggingface.co/nc-ai-consortium/VAETKI-VL-7B-A1B/blob/main/preprocessor_config.json) refers to the total number of pixels, not the H or W dimension. I pushed the latest change to support this as min/max pixels, but I haven't had time to test it.

Would appreciate if you can try re-converting GGUF & test it to confirm if my commit works

---

Update: the preprocessor_config has min/max pixel as dedicated fields, so we can use it instead:
image

Added in my last commit below

dororodoroddo:
> @dororodoroddo the `size.longest_edge/shortest_edge` in [preprocessor_config.json](https://huggingface.co/nc-ai-consortium/VAETKI-VL-7B-A1B/blob/main/preprocessor_config.json) refers to the total number of pixels, not the H or W dimension. I pushed the latest change to support this as min/max pixels, but I haven't had time to test it.
> 
> Would appreciate if you can try re-converting GGUF & test it to confirm if my commit works
> 
> Update: the preprocessor_config has min/max pixel as dedicated fields, so we can use it instead: image
> 
> Added in my last commit below


Thank you for the fix.

I rebuilt both llama.cpp and gguf, and tested the following sizes, all of which rendered the images correctly.

Test sizes: min 3136px (56*56), max 2560000px (1600*1600), 560*560, 224*224, 112*112

src/models/vaetki.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+            // {n_embd_head_v_mla, n_head, n_tokens}
+            ggml_tensor * Vcur = ggml_view_3d(ctx0, kv,
+                n_embd_head_v_mla, n_head, n_tokens,
+                ggml_row_size(kv->type, n_embd_head_qk_nope + n_embd_head_v_mla),
+                ggml_row_size(kv->type, n_embd_head_qk_nope + n_embd_head_v_mla) * n_head,
+                ggml_row_size(kv->type, n_embd_head_qk_nope));
+            cb(Vcur, "Vcur", il);
+
+            Vcur = ggml_cont(ctx0, Vcur);
```
ngxson, line 98: Probably redundant, please try
```suggestion
```

src/llama-model.cpp
```diff
@@ -1124,6 +1125,43 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+                ml.get_key(LLM_KV_ATTENTION_Q_LORA_RANK,       hparams.n_lora_q);
+                ml.get_key(LLM_KV_ATTENTION_KV_LORA_RANK,      hparams.n_lora_kv);
+                ml.get_key(LLM_KV_ATTENTION_KEY_LENGTH_MLA,    hparams.n_embd_head_k_mla, false);
+                ml.get_key(LLM_KV_ATTENTION_VALUE_LENGTH_MLA,  hparams.n_embd_head_v_mla, false);
+
+                if (hparams.n_embd_head_k_mla != 0 && hparams.n_embd_head_v_mla != 0) {
+                    hparams.n_embd_head_k = hparams.n_embd_head_k_mla;
+                    hparams.n_embd_head_v = hparams.n_embd_head_v_mla;
+                    hparams.n_head_kv_arr = hparams.n_head_arr;
+                }
```
ngxson: these code paths are for compatibility with old deepseek2 GGUFs, it is likely redundant here
```suggestion
                ml.get_key(LLM_KV_ATTENTION_KEY_LENGTH_MLA,    hparams.n_embd_head_k_mla);
                ml.get_key(LLM_KV_ATTENTION_VALUE_LENGTH_MLA,  hparams.n_embd_head_v_mla);
                hparams.n_embd_head_k = hparams.n_embd_head_k_mla;
                hparams.n_embd_head_v = hparams.n_embd_head_v_mla;
                hparams.n_head_kv_arr = hparams.n_head_arr;
```

src/llama-model.cpp
```diff
@@ -1124,6 +1125,43 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+                if (hparams.expert_gating_func == LLAMA_EXPERT_GATING_FUNC_TYPE_NONE) {
+                    hparams.expert_gating_func = LLAMA_EXPERT_GATING_FUNC_TYPE_SIGMOID;
+                }
+
+                ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW, hparams.n_swa, false);
+                ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA, hparams.rope_freq_base_train_swa, false);
+                if (hparams.n_swa > 0) {
+                    hparams.swa_type = LLAMA_SWA_TYPE_STANDARD;
+                    ml.get_key_or_arr(LLM_KV_ATTENTION_SLIDING_WINDOW_PATTERN, hparams.swa_layers, hparams.n_layer);
+                }
```
ngxson: SWA should be enforced, it is not optional for this model

ngxson (commented):
the mtmd part looks ok, I'm passing to @CISC @pwilkin to review the rest

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+    _experts: list[dict[str, Tensor]] | None = None
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Flatten text_config parameters to top level
+        if "text_config" in self.hparams:
+            text_config = self.hparams["text_config"]
+            for key, value in text_config.items():
+                if key not in self.hparams:
+                    self.hparams[key] = value
```
```suggestion
```
Already done here:
https://github.com/ggml-org/llama.cpp/blob/32bf06ae342a1b98feaeaf27a5df4b54bc9a0bd2/convert_hf_to_gguf.py#L769-L771

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+        # Add merges (convert from [['a', 'b'], ...] to ['a b', ...] format)
+        if merges:
+            # tokenizer.json stores merges as list of pairs, GGUF expects space-separated strings
+            if isinstance(merges[0], list):
+                merges = [' '.join(pair) for pair in merges]
+            self.gguf_writer.add_token_merges(merges)
+
+        # Add special tokens
+        special_vocab = gguf.SpecialVocab(dir_model, load_merges=False)
+        special_vocab.add_to_gguf(self.gguf_writer)
```
CISC: This is super iffy, and only done to add `\u2581` as far as I can tell.

Support for this should rather be added to `SpecialVocab`.

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+
+        hparams = self.hparams
+        self.gguf_writer.add_block_count(hparams["num_hidden_layers"])
+        self.gguf_writer.add_context_length(hparams.get("max_position_embeddings", 32768))
+        self.gguf_writer.add_embedding_length(hparams["hidden_size"])
+        self.gguf_writer.add_feed_forward_length(hparams["intermediate_size"])
+        self.gguf_writer.add_head_count(hparams["num_attention_heads"])
```
```suggestion
```

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+        # For MLA without absorption, n_head_kv = n_head (full MHA after decompression)
+        self.gguf_writer.add_head_count_kv(hparams["num_attention_heads"])
+        self.gguf_writer.add_layer_norm_rms_eps(hparams.get("rms_norm_eps", 1e-5))
```
```suggestion
```

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+        self.gguf_writer.add_rope_dimension_count(hparams["qk_rope_head_dim"])
+
+        # VAETKI uses hybrid attention with different rope_theta per layer type:
+        # - sliding_attention layers use rope_theta (local, default 10000.0)
+        # - full_attention layers use rope_theta_global (global, default 1000000.0)
+        # In llama.cpp: rope_freq_base is for non-SWA (full), rope_freq_base_swa is for SWA (sliding)
+        rope_theta_local = hparams.get("rope_theta", 10000.0)
+        rope_theta_global = hparams.get("rope_theta_global", 1000000.0)
+        self.gguf_writer.add_rope_freq_base(rope_theta_global)  # for full_attention layers
+        self.gguf_writer.add_rope_freq_base_swa(rope_theta_local)  # for sliding_attention layers
```
```suggestion
```
These should already be set, but you must fix your config to use `rope_theta_local` instead of `rope_theta` (not correct, and thus will not be set and default to 10000, which coincides with your config anyway), or alternatively use the new `rope_parameters` in `transformers` `5.0`.

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+
+        # MoE parameters
+        self.gguf_writer.add_leading_dense_block_count(hparams.get("first_k_dense_replace", 1))
+        self.gguf_writer.add_expert_count(hparams["n_routed_experts"])
+        self.gguf_writer.add_expert_used_count(hparams["num_experts_per_tok"])
```
```suggestion
```

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+        self.gguf_writer.add_expert_shared_count(hparams.get("n_shared_experts", 1))
+        self.gguf_writer.add_expert_feed_forward_length(hparams["moe_intermediate_size"])
+        self.gguf_writer.add_expert_weights_scale(hparams.get("routed_scaling_factor", 1.0))
+        # VAETKI uses sigmoid gating function (WBLTopkRouter uses router_logits.sigmoid())
+        self.gguf_writer.add_expert_gating_func(gguf.ExpertGatingFuncType.SIGMOID)
+        # Normalize top-k probabilities (norm_topk_prob=true in config)
```
```suggestion
        if (routed_scale := hparams.get("routed_scaling_factor")) is not None:
            self.gguf_writer.add_expert_weights_scale(routed_scale)
```

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+
+            # Add sliding window pattern from layer_types
+            if "layer_types" in hparams:
+                # Convert layer_types to sliding_window_pattern (1 = sliding, 0 = full)
+                # Store as uint32 array to match llama.cpp hparams.swa_layers type
+                sliding_window_pattern = [1 if t == "sliding_attention" else 0 for t in hparams["layer_types"]]
+                self.gguf_writer.add_array(
+                    gguf.Keys.Attention.SLIDING_WINDOW_PATTERN.format(arch=self.gguf_writer.arch),
+                    sliding_window_pattern
+                )
```
```suggestion
        self.gguf_writer.add_sliding_window(hparams["sliding_window"])
        sliding_window_pattern = []
        for t in self.hparams["layer_types"]:
            sliding_window_pattern.append(t == "sliding_attention")
        self.gguf_writer.add_sliding_window_pattern(sliding_window_pattern)
```

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7634,314 @@ def prepare_tensors(self):
...
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        # Skip vision encoder tensors
+        if "vision_tower" in name or "vision_model" in name or "visual" in name:
+            return []
+        if name.startswith("model.vision_model.") or name.startswith("vision_model."):
+            return []
+
+        # Handle lm_head.weight (VAETKI does not use tied embeddings)
+        if name == "lm_head.weight":
+            return [(self.format_tensor_name(gguf.MODEL_TENSOR.OUTPUT), data_torch)]
```
```suggestion
```
No point, this is handled at the end.

src/llama-vocab.cpp
CISC, line 1: Ok, I don't think this is BPE any more, this vocab is not correct.

dororodoroddo: The tokenizer config.json says "BPE," so I based it on that, but I think my knowledge was lacking.
I'll have to delete all the tokenizer codes and start over.
I'll analyze how the Hugging Face Rust library parses the JSON line by line and find alternative workflows that follow the same order.

I'll try my best to achieve an ETA of (UTC): Wednesday, January 14, 2026, 2:59 PM.

src/models/vaetki.cpp
dororodoroddo: this caused an error
```
Loading model... |/vaetki/llama.cpp/src/llama-kv-cache.cpp:1109: GGML_ASSERT(ggml_row_size(v_cur->type, n_embd_head) == v_cur->nb[1]) failed
WARNING: Using native backtrace. Set GGML_BACKTRACE_LLDB for more info.
WARNING: GGML_BACKTRACE_LLDB may cause native MacOS Terminal.app to crash.
```

src/llama-model.cpp
dororodoroddo: This worked well

src/llama-model.cpp
dororodoroddo: Yes, I applied this

src/llama-vocab.cpp
dororodoroddo: After careful study, I discovered significant differences.
Only a few areas matched BPE, while many others differed.
The logic was as follows:

# encoding
1. normalizer : Replace " " to "_" (like SPM)
2. pre_tokenizer : Metaspace (" " to "_"  one more time)  -> Split "a_b" to ['a', '_b'] (like SPM + BPE)
3. Tokenize : BPE(?)
    a. rank based merge (BPE)
    b. Byte Fallback to <0x##> (like a SPM)
4. post_processor : ByteLevel

# decoding
1. ID to token str
2. Decoder Sequence 
    a. replace : "_" to " "
    b. ByteFallback : <0xXX> to UTF-8
    c. Fuse

This tokenizer didn't overlap with any other tokenizer, so I had no choice but to create a new logic, much like plamo.

I borrowed a lot from BPE, and since this code was solely for VAETKI, I used a loop instead of a regex

Some of the logic (detokenization) follows SPM or UGM's logic (BOS tokenization).

After applying all the code up to this point, I rebuilt llama.cpp and gguf. After testing with 7B-text, 7B-vision, and 100b q4_k_m-text, I found no issues.

convert_hf_to_gguf.py
dororodoroddo: Yes, I removed that code.

convert_hf_to_gguf.py
dororodoroddo: All of that part was removed and in the process of being rewritten, it was rewritten to load itself like plamo to use the new tokenizer logic.

convert_hf_to_gguf.py
dororodoroddo: Yes, I removed that code.

convert_hf_to_gguf.py
dororodoroddo: Yes, I removed that code too.

convert_hf_to_gguf.py
dororodoroddo: Yes, I wrote it according to the transformers 5.0 format.
```def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # Set rope_parameters for hybrid attention (transformers 5.0 format)
        self.rope_parameters = {
            "full_attention": {"rope_theta": self.hparams.get("rope_theta_global", 1000000.0)},
            "sliding_attention": {"rope_theta": self.hparams.get("rope_theta", 10000.0)}
        }
```

CISC: I meant in your model configs, adding it here is no better than hardcoding it.

CISC: Basically you are using the wrong config variable for sliding attention (oddly enough it's correctly named in your `configuration_vaetki_vl.py`).

dororodoroddo: ah, i understand. in config.json, It is listed as rope_theta in config.json
image

CISC: Yes, and that is incorrect, `rope_theta` is historically the global variable, and in `configuration_vaetki_vl.py` it is named `rope_theta_local` (which is correct), but setting it from `rope_theta` because `config.json` is incorrect. I suggest you fix that.

dororodoroddo: Aha, so what you're saying is that it will be fixed with rope_theta_local someday, so fix it.
```
self.rope_parameters = {
            "full_attention": {"rope_theta": self.hparams.get("rope_theta_global", 1000000.0)},
            "sliding_attention": {"rope_theta": self.hparams.get("rope_theta_local", 10000.0)}
        }
```

CISC: Sorry, I thought you were working for NC-AI. If that's not the case, just remove it all, `10000` is the default for SWA, if they ever fix it it will get automatically added by this code:
https://github.com/ggml-org/llama.cpp/blob/65e6f8b4abfb9f2fd2fbc7c57f046d18691bbe38/convert_hf_to_gguf.py#L778-L788

dororodoroddo: i remove it 🫡

convert_hf_to_gguf.py
dororodoroddo: Oh, I missed this code. But when I apply it, I get an error saying that the bool array cannot be read
```
llama_model_load: error loading model: error loading model hyperparameters: vaetki.attention.sliding_window_pattern is not a string/float32/uint32/int32 array
```

CISC: Ah, hm, it looks like @ngxson changed this recently, let me look into it.

dororodoroddo: Yes, I understand. This part, namely the rope_theta part in the comment above and the removal part in the comment below, is pending commit.

CISC: Ok, strangely enough it seems this has never worked (I guess only because no model has ever actually used the pattern), just change it to `int(t == "sliding_attention")`.

src/llama-vocab.cpp
```diff
@@ -609,9 +610,9 @@ struct llm_tokenizer_bpe_session {
                 if (token == LLAMA_TOKEN_NULL) {
                     for (auto j = str.begin(); j != str.end(); ++j) {
                         std::string byte_str(1, *j);
-                        auto token_multibyte = vocab.text_to_token(byte_str);
-                        if (token_multibyte != LLAMA_TOKEN_NULL) {
-                            output.push_back(token_multibyte);
+                        auto token_byte = vocab.text_to_token(byte_str);
+                        if (token_byte != LLAMA_TOKEN_NULL) {
+                            output.push_back(token_byte);
                         }
```
dororodoroddo, line 622: The above traces will be removed at the next push.

convert_hf_to_gguf.py
dororodoroddo: Yes, I modified it that way.

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7631,242 @@ def prepare_tensors(self):
...
+
+        # key_length_mla/value_length_mla are the MLA head dimensions (same as key/value for non-absorption)
+        self.gguf_writer.add_key_length_mla(hparams["qk_head_dim"])
+        self.gguf_writer.add_value_length_mla(hparams["v_head_dim"])
+        self.gguf_writer.add_rope_dimension_count(hparams["qk_rope_head_dim"])
+
+        self.rope_parameters = {
+            "full_attention": {"rope_theta": self.hparams.get("rope_theta_global", 1000000.0)},
+            "sliding_attention": {"rope_theta": self.hparams.get("rope_theta", 10000.0)}
+        }
```
```suggestion
```

dororodoroddo: Oh, there was a serious mistake. It has been removed.

convert_hf_to_gguf.py
```diff
@@ -7631,6 +7631,237 @@ def prepare_tensors(self):
...
+        self.gguf_writer.add_expert_feed_forward_length(hparams["moe_intermediate_size"])
+        if (routed_scale := hparams.get("routed_scaling_factor")) is not None:
+            self.gguf_writer.add_expert_weights_scale(routed_scale)
+        if hparams.get("norm_topk_prob", False):
+            self.gguf_writer.add_expert_weights_norm(True)
+
+        self.gguf_writer.add_sliding_window(hparams["sliding_window"])
+        sliding_window_pattern = []
+        for t in self.hparams["layer_types"]:
+            sliding_window_pattern.append(int(t == "sliding_attention"))
```
```suggestion
            sliding_window_pattern.append(t == "sliding_attention")
```
Fixed in #18850

dororodoroddo: The changes have been reflected after rebasing.

src/models/vaetki.cpp
ngxson: I think you're copying the "MLA without the absorption optimization" (`is_mla = false`) branch from deepseek2. The performance will be sub-optimal.

CISC:
@ngxson The vocab gives me great pause, I think this needs to be discussed properly, and if current implementation is the way to go we need to thoroughly test it and add it to `ggml-vocabs`.

@ggerganov Any takes on this?

src/models/vaetki.cpp
dororodoroddo: Yes, I followed the "MLA excluding absorption optimization" branch because I referenced the implementation in "modeling_vaetki.py."

The model architecture documentation mentions achieving the benefits of "MLA absorption," so it seems feasible.

However, simply switching branches resulted in errors, so I'm currently analyzing the architecture for this implementation.

It seems unlikely a quick solution will be found. i'll trying

ggerganov:
@CISC Are you concerned that the tokenizer is not correct? If so, we should wait for a comprehensive validation against a reference implementation to be demonstrated by @dororodoroddo or someone else.

This seems to be a quite young model family, so I think it's OK to keep this branch for a while and see how things develop. If there is increased interest/demand by the community for support and we get feedback that the implementation is working as expected, we can consider merging.

CISC:
> @CISC Are you concerned that the tokenizer is not correct?

My main concern is that it's set up as a BPE tokenizer in `transformers` with heavy alterations via pre-tokenizer flags. It's not the first case, and might actually require that we rethink our own BPE tokenizer, see #18756 for reference.

dororodoroddo:
In addition to the administrator's comments, I will pause the revisions and wait for the following reasons:
1. The Vaetki family needs to be monitored.
- The Vaetki project was one of the projects supported by the Korean government, but it recently lost support, and NC-AI has given up on the second round.
=> There may not be 'increased interest/demand by the community for support.'

2. Rather than writing a tokenizer specifically for Vaetki, we may need a tokenizer logic that allows for step-by-step customization, like the huggingface tokenizer.

I'm personally interested in the Vaetki architecture and plan to experiment with it. However, I’d appreciate everyone’s feedback on whether to keep this as is or close the pull request

PR #18720, opened 2026-01-09
kv-cache : optimize SWA slot reuse with forward-looking masking

pestopoppa:
## Summary

Optimizes SWA (sliding window attention) cache slot reuse by using forward-looking position masking. This enables efficient speculative decoding on ISWA models (Gemma-3 family) without requiring the `--swa-full` flag.

**Key change:** In `find_slot()`, instead of checking if cached cells are masked relative to the stored sequence max position, we now check against the incoming batch's max position. This allows reusing cells that will be outside the attention window *after* the batch is inserted.

## Results

Tested on Gemma-3-27B + Gemma-3-1B draft (speculative decoding):

| Metric | Before | After |
|--------|--------|-------|
| `--swa-full` required | Yes | No |
| SWA cache size | 10240 MiB | 624 MiB |
| Memory reduction | — | **94%** |
| Acceptance rate | N/A (crashed) | 42-81% |

## Technical Details

- For SWA caches (`n_swa > 0`), compute `pos_batch_max` from the batch before searching for slots
- Use `pos_batch_max + 1` instead of `cells.seq_pos_max(seq_id_cell) + 1` for the masking check
- Non-SWA caches are unaffected (the SWA masking check is skipped entirely)

## Test Plan

- [x] Tested on Gemma-3-27B-QAT + Gemma-3-1B draft without `--swa-full`
- [x] Verified SWA cache remains at 624 MiB (not 10240 MiB)
- [x] Verified 81-89% acceptance rate on coding prompts
- [x] Verified no regression on non-SWA models

Claude was used to research the codebase.

### DIFF
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -798,6 +798,14 @@ llama_kv_cache::slot_info llama_kv_cache::find_slot(const llama_ubatch & ubatch,
         // for non-continuous slots, we test the tokens one by one
         const uint32_t n_test = cont ? n_tokens : 1;
 
+        // For SWA caches: compute max position in this batch for sequence s
+        // This is used to determine which old cells are outside the future attention window
+        // (forward-looking: cells masked AFTER batch insertion can be reused)
+        llama_pos pos_batch_max = 0;
+        if (n_swa > 0) {
+            pos_batch_max = ubatch.pos[s * n_tokens + n_tokens - 1];
+        }
+
         while (true) {
             if (head_cur + n_test > cells.size()) {
                 n_tested += cells.size() - head_cur;
@@ -818,7 +826,7 @@ llama_kv_cache::slot_info llama_kv_cache::find_slot(const llama_ubatch & ubatch,
                 //  - the cell is empty
                 //  - the cell is occupied only by one sequence:
                 //    - (disabled) mask causally, if the sequence is the same as the one we are inserting
-                //    - mask SWA, using current max pos for that sequence in the cache
+                //    - mask SWA, using the batch's max pos (forward-looking)
                 //                always insert in the cell with minimum pos
                 bool can_use = cells.is_empty(idx);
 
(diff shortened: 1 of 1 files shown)


ggerganov:
> This enables efficient speculative decoding on ISWA models (Gemma-3 family) without requiring the --swa-full flag.

Must be confused - the `--swa-full` flag is not required. SWA speculative decoding has been supported with `llama-server` ever since https://github.com/ggml-org/llama.cpp/pull/14131

pestopoppa:
Apologies, I made a mistake in describing this PR properly and it created confusion. This PR optimizes SWA (Sliding Window Attention) KV cache slot reuse by using forward-looking masking when searching for available cells.

When `find_slot()` searches for cells to reuse in SWA caches, it checks if old cells are masked relative to the cached sequence max position. This is conservative - cells that will be outside the attention window after batch insertion cannot be reused until a future decode.

As a solution we check cell masking against the the new batch's max position instead of the cached max. This allows immediate reuse of cells that will be outside the sliding window after the batch completes.

This is a pure optimization with no behavioral changes - existing functionality is preserved.

ggerganov:
It does not work. When you evaluate the batch it is true that the last token would need to "see" just the `n_swa` tokens before that. But you also must consider what the first token in the batch is going to "see". SWA requires that every token has the `n_swa` context before it in order to be fully precise.

I also did similar experiments and noticed that if you apply such approximations, the behavior is still OK. Probably it would be very difficult to notice a regression through regular chat completions. But the main reason to not allow this is that we want to be mathematically exact in the SWA evaluation and prevent any potential side effects from this approximation.

pestopoppa:
> But the main reason to not allow this is that we want to be mathematically exact in the SWA evaluation and prevent any potential side effects from this approximation.

Thank you for this response. Maybe the repo's policy allow making the optimization explicitly opt-in (e.g. --swa-approx) so users can choose the tradeoff while default behavior remains exact? I'm guessing the answer is no otherwise you would've done it already yourself. Thanks again for the feedback! Much appreciated

pestopoppa:
@ggerganov On second thought ...  I thought about this more and realized that mathematical correctness can still be enforced by using the batch's min position instead of its max. The first token in a batch has the most demanding attention window (because it extends furthest back), so if we preserve everything it needs, all other tokens automatically have their full context. This guarantees mathematical exactness with no approximation. The memory savings are nearly identical - the only difference is (batch_size - 1) fewer cells can be reused, which is negligible (7 cells for an 8-token speculative batch vs thousands reclaimed). Testing with Gemma-3-12B generating 1504 tokens (47% beyond the n_swa=1024 window) - cache stayed bounded, ~50% acceptance rate, output verified as coherent.

I opened a fresh PR with the corrected approach: #18727

ggerganov:
> The first token in a batch has the most demanding attention window (because it extends furthest back), so if we preserve everything it needs, all other tokens automatically have their full context. This guarantees mathematical exactness with no approximation.

This is already the logic implemented on `master`. `cells.seq_pos_max() + 1` is equal to the smallest position in the incoming batch. I don't think #18727 has any improvement to that logic.

If you are trying to solve a specific issue, it would be easier to explain the problem that you encounter. I don't normally use SWA speculative decoding these days, but last I checked it was running fine and don't think there is anything that can be optimized.

PR #18723, opened 2026-01-09
llama: fix pooled embedding readback sizing/stride and state I/O

retr0reg:
When pooling, size and copy embd_seq using the
pooled tensor's actual width (t_embd->ne[0])
and row stride (t_embd->nb[1]) instead of n_embd.
This avoids ggml_backend_tensor_get_async bounds
assertions (abort) when n_embd_out < n_embd, and
prevents heap OOB reads when n_embd_out > n_embd
when callers treat pooled embeddings as length
n_embd_out. Also use get_n_embd_out() in
state_write_data() and output_reorder()

Also add GGML_ASSERT(seq_idx >= 0) before computing 
byte offsets for readback to prevent negative overflow.

Fix: `PWNO-0023`

### DIFF
--- a/src/llama-context.cpp
+++ b/src/llama-context.cpp
@@ -1207,12 +1207,21 @@ int llama_context::encode(const llama_batch & batch_inp) {
                     // extract sequence embeddings
                     auto & embd_seq_out = embd_seq;
 
+                    const uint32_t n_embd_out = (uint32_t) t_embd->ne[0];
+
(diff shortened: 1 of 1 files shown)


CISC:
I don't think this is the correct fix, at least not for `LLAMA_POOLING_TYPE_RANK`, should not `n_embd_out` and `n_cls_out` be sanity-checked instead?

PR #18727, opened 2026-01-09
kv-cache: optimize SWA slot reuse with forward-looking masking

pestopoppa:
## Summary

Optimize Sliding Window Attention (SWA) cache slot reuse by checking cell reusability against the incoming batch's **minimum** position rather than the cached sequence's maximum position. This enables more aggressive cache reclamation while maintaining mathematical exactness.

## Addressing PR #18720 Feedback

This addresses feedback from @ggerganov on #18720, which correctly noted that using the batch's **maximum** position could evict cells still needed by earlier tokens in the batch. The fix uses **minimum** position instead:

- The token at the batch's minimum position has the most demanding attention window (extends furthest back in history)
- By checking reusability against this position, we guarantee all tokens in the batch have their complete context
- A cell can only be reused if it's outside the attention window of **all** tokens in the batch
- Since the min-position token's window is the superset, this is mathematically exact

Memory impact is negligible: only `(batch_size - 1)` fewer cells can be reused compared to a max-based approach (e.g., 7 cells for an 8-token speculative batch), while correctness is guaranteed.

## Test Results

Tested with Gemma-3-12B (n_swa=1024) + Gemma-3-1B draft:

| Metric | Result |
|--------|--------|
| Tokens generated | 1504 (47% beyond window boundary) |
| SWA cache size | Stayed bounded at 1536 cells |
| Acceptance rate | ~50% |
| Output quality | ✓ Coherent technical documents |

## Changes

- `src/llama-kv-cache.cpp`: Modified `find_slot()` to compute batch min position and use it for SWA cell reusability checks

### DIFF
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -815,6 +815,14 @@ llama_kv_cache::slot_info llama_kv_cache::find_slot(const llama_ubatch & ubatch,
         // for non-continuous slots, we test the tokens one by one
         const uint32_t n_test = cont ? n_tokens : 1;
 
+        // For SWA caches: compute min position in this batch for sequence s
+        // This ensures all tokens in the batch have their full attention window
+        // (the token at min position has the most demanding context requirement)
+        llama_pos pos_batch_min = 0;
+        if (n_swa > 0) {
+            pos_batch_min = ubatch.pos[s * n_tokens];
+        }
+
         while (true) {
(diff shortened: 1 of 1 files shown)


pestopoppa:
You're right - `seq_pos_max + 1` equals `pos_batch_min` by construction since positions are assigned contiguously. My change computes the same value differently without functional improvement.

Thanks for the explanation. Can't believe I missed that. Closing.

PR #18729, opened 2026-01-10
lookup, lookahead: fix crash when n_ctx not specified

pestopoppa:
## Summary

Fixes a crash in `llama-lookup` and `llama-lookahead` when run without explicit `-c` flag:

```
GGML_ASSERT(batch.seq_id[batch.n_tokens] && "llama_batch size exceeded")
```

## Root Cause

Both examples use `params.n_ctx` directly for batch initialization:

```cpp
// lookup.cpp:109
llama_batch batch_tgt = llama_batch_init(params.n_ctx, 0, 1);

// lookahead.cpp:118
llama_batch batch = llama_batch_init(params.n_ctx, 0, W + G + 1);
```

Since #16653 changed the default `n_ctx` to 0 (for GPU auto-fitting), `params.n_ctx` remains 0 even after the context is properly initialized. This creates a zero-sized batch that crashes on the first `common_batch_add()`.

## Bug History

This bug was **dormant for 2+ years**:

| Date | PR | Default n_ctx | Effect |
|------|-----|---------------|--------|
| Nov 2023 | #4207 | 512 | lookahead.cpp created - **works** |
| Dec 2023 | #4484 | 512 | lookup.cpp created - **works** |
| Nov 2024 | #10136 | 4096 | Default increased - **works** |
| **Dec 2025** | **#16653** | **0** | Auto-fitting enabled - **CRASHES** |

The pattern was always incorrect, but only triggered when `n_ctx` default became 0.

## Fix

Use `llama_n_ctx(ctx)` to get the actual runtime context size:

```cpp
// Before
llama_batch batch_tgt = llama_batch_init(params.n_ctx, 0, 1);

// After
llama_batch batch_tgt = llama_batch_init(llama_n_ctx(ctx), 0, 1);
```

This matches:
- The pattern already used in `lookup.cpp:72` for `max_context_size`
- The pattern used in `speculative.cpp` and `speculative-simple.cpp`

## Testing

```bash
# Before fix (crashes):
llama-lookup -m model.gguf -f prompt.txt --draft 4 -n 50
# GGML_ASSERT(batch.seq_id[batch.n_tokens] && "llama_batch size exceeded")

# After fix (works):
llama-lookup -m model.gguf -f prompt.txt --draft 4 -n 50
# n_accept = 1, accept = 12.500%
```

## Note

`llama-lookahead` has a separate pre-existing issue with sequence initialization (`n_seq_max=1` when it needs `W+G+1`) that is unrelated to this batch size fix.

PR #14482, opened 2025-07-01
llama : reuse compute graphs

ggerganov:
target #14285

Reuse computation graphs from the previous ubatch when possible. Works with any batch size and any model.

> [!NOTE]
> To enable this functionality, there is a temporary requirement `LLAMA_SET_ROWS=1` to be set in your environment variable. In the future, this will become the default.

This functionality requires the `ggml_set_rows()` operator to be supported (see #14285). In order to be able to reuse a compute graph, its topology (shapes, strides, parameters, etc.) has to be entirely defined by the set of input tensors (e.g. `inp_embd`, `inp_pos`, `inp_attn`, etc.).

This PR adds logic to update a previous `llm_graph_result` by verifying that the new `llm_graph_params` would result in the same tensor shapes. For this to work, we should no longer preemptively reset the scheduler after processing a batch so that all buffers from the previous graph remain allocated and ready for reuse, in case the new `ubatch` is compatible. See the new `llm_graph_result::update()` method:

https://github.com/ggml-org/llama.cpp/blob/fc4fdf623c098d2b4d7699fdb7f2ea5ae1f63b57/src/llama-graph.h#L506-L525

The other change that is needed is to introduce a way to swap the `llama_memory_context` of all graph inputs, so that the new call to `llm_graph_result_i::set_inputs()` uses the correct context from the current `ubatch`. This is performed by calling the `llm_graph_input_i::update()` method of all input tensors.

To enable this feature, define the `LLAMA_SET_ROWS` environment variable.

To debug, define `LLAMA_GRAPH_RESULT_DEBUG=2` and add `-lv 1` to the CLI args.

### Tests

```bash
LLAMA_SET_ROWS=1 ./bin/llama-cli -m ../models/llama-3.2-3b-instruct/ggml-model-q8_0.gguf -p "I believe the meaning of life is" -n 32 --top-k 1 -fa

LLAMA_SET_ROWS=1 ./bin/llama-parallel -m ../models/qwen2.5-3b-coder/ggml-model-q8_0.gguf -np 8 -ns 128 -s 1 -c 4096 -fa -n 128
```

Benchmark on M2 Ultra:

```bash
[...]

PR #4207, opened 2023-11-24
lookahead : add example for lookahead decoding

ggerganov:
ref #4157 

Think this should implement the approach from: https://lmsys.org/blog/2023-11-21-lookahead-decoding/

- The approach requires large batches to be decoded, which in turn requires a lot of FLOPS even for single stream
- Haven't done extensive tests on NVIDIA GPUs yet, but on Apple Silicon I don't observe speed-ups with 7B F16 models, likely due to small compute
- As usual quantum models would require efficient batch CUDA kernels which we haven't implemented yet, so I don't expect speed-up there as well for the moment

In `llama.cpp` the required attention mask is achieved with the following `llama_batch`:

```
        // Example for W = 5, N = 4, G = 2:
        // (I = input, L = lookahead, V = verification)
        //
        // Batch:  0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20
        // T:        -2 -2 -2 -2 -1 -1 -1 -1 -1  0  0  0  0  0  0
        // Info:   I  L  L  L  L  L  L  L  L  L  L  L  L  L  L  V  V  V  V  V  V
        // Pos:    0  1  2  3  4  1  2  3  4  5  2  3  4  5  6  1  2  3  1  2  3   (+ n_past)
        // Logits: 1  0  0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1  1  1  1
        // ---------------------------------------------------------------------
        // Seq:    0
        //         1              1              1
        //         2  2              2              2
        //         3  3  3              3              3
        //         4  4  4  4              4              4
        //         5  5  5  5  5              5              5
        //         6                                            6  6  6
        //         7                                                     7  7  7
        // ---------------------------------------------------------------------
        //                                       |  |  |  |  |  |  |  |  |  |  |
        //                                       V  V  V  V  V  |  |  |  |  |  |
        //                                         j_tokens     |  |  |  |  |  |
        //                                                      V  V  V  V  V  V
        //                                                             id
```

TODO:

- [ ] Pass `W`, `N`, `G` from command-line (currently hardcoded)
- [ ] Decode the verification n-grams via tree-based decoding. Since most of the n-grams share common prefixes, we can ramp-up the number `G` at no additional computational cost. See `speculative` example for a demonstration of tree-based decoding

```bash
[...]

PR #18730, opened 2026-01-10
lookahead: fix n_seq_max and kv_unified configuration

pestopoppa:
## Summary

Fixes `llama-lookahead` configuration issues that have been broken since PR #14482 (July 2025).

**Note:** This PR depends on #18729 for the batch init fix. Both PRs are needed for lookahead to fully work.

## Root Cause

Two lookahead-specific configuration issues:

### 1. Sequence count (n_seq_max)
PR #14482 changed seq_id validation from `LLAMA_MAX_SEQ` (large constant) to `n_seq_max` (context-specific). Lookahead needs `W + G + 1 = 31` sequences for parallel Jacobi decoding, but `params.n_parallel` defaulted to 1.

### 2. KV unified mode
Batch splitting with "coupled sequences" requires unified KV cache. Lookahead didn't enable this, causing:
```
split_equal: sequential split is not supported when there are coupled sequences
```

## Fix

```cpp
// lookahead requires W + G + 1 sequences for parallel Jacobi decoding
params.n_parallel = W + G + 1;

// unified KV cache is required for coupled sequences in batch splitting
params.kv_unified = true;
```

## Bug Timeline

| Date | PR | Effect |
|------|-----|--------|
| Nov 2023 | #4207 | lookahead.cpp created - works |
| **July 2025** | **#14482** | seq_id validation changed - **breaks lookahead** |

## Testing

With both this PR and #18729 applied:

```
encoded    4 tokens in    0.138 seconds
decoded   51 tokens in   71.591 seconds
n_accept  = 13
```

## Dependencies

- **Requires #18729** for the batch init fix (`params.n_ctx` → `llama_n_ctx(ctx)`)

---
Bug history researched with Claude.

### DIFF
--- a/examples/lookahead/lookahead.cpp
+++ b/examples/lookahead/lookahead.cpp
@@ -50,6 +50,12 @@ int main(int argc, char ** argv) {
     const int N = 5;  // n-gram size
     const int G = 15; // max verification n-grams
 
+    // lookahead requires W + G + 1 sequences for parallel Jacobi decoding
+    params.n_parallel = W + G + 1;
+
+    // unified KV cache is required for coupled sequences in batch splitting
+    params.kv_unified = true;
+
     // init llama.cpp
     llama_backend_init();
     llama_numa_init(params.numa);


ngxson:
`kv_unified` is only handled by server, no use to add it here. please stop sending us low-quality PRs

ngxson:
the only thing that may need to make it work is to correctly specify `params.n_parallel`, you may merge your changes into one PR instead of spamming us with smaller one

ggerganov:
Actually, I think the unified KV is required for this example. Haven't ran it in a while, but it was developed when we only had unified cache. And probably forgot to update it when we switched to non-unified being the default.

ngxson:
hmm ok I see, I thought this is the speculative decoding example. it seems like the lookahead example rely extensively on "branching" the generation via parallel seq.

It's a valid use of unified cache here then, sorry for the oversight. @pestopoppa it's better if you can merge all changes related to lookahead into one PR, this way we can test it easier.

pestopoppa:
> @pestopoppa it's better if you can merge all changes related to lookahead into one PR, this way we can test it easier.

Thanks for taking another look and appreciatign the nuance. I've consolidated the fix from this PR into #18729 (commit b917cd2).

  The PR now includes:
  1. llama_n_ctx(ctx) fix for batch init (both lookup and lookahead)
  2. params.n_parallel = W + G + 1 for the 31 sequences
  3. params.kv_unified = true for coupled sequence batch splitting

  Ready for testing when you have a chance.

PR #18731, opened 2026-01-10
feat: add support for WeDLM architecture

feedseawave:
# model : add WeDLM architecture support

## Summary

Add support for **WeDLM** (Tencent's 8B instruction-tuned model).

WeDLM is architecturally identical to Qwen3 (both use QK Norm, GQA, SwiGLU, RoPE), so this PR reuses the existing `llm_build_qwen3` graph builder.

## Changes

| File | Changes |
|------|---------|
| `src/llama-arch.h` | Add `LLM_ARCH_WEDLM` enum |
| `src/llama-arch.cpp` | Add name mapping, share tensor list with Qwen3 |
| `src/llama-model.cpp` | Add case fallthrough to Qwen3 for params loading and graph building |
| `gguf-py/gguf/constants.py` | Add `MODEL_ARCH.WEDLM` and tensor list |
| `convert_hf_to_gguf.py` | Add `WeDLMModel` converter class inheriting from `Qwen3Model` |

## Model Information

- **Original**: https://huggingface.co/tencent/WeDLM-8B-Instruct
- **Parameters**: 8.19B
- **Layers**: 36
- **Context**: 16,384
- **Features**: QK Norm, GQA (32 heads / 8 KV), SwiGLU, RoPE (theta=1M)
- **Languages**: English, Chinese
- **License**: Apache 2.0

## Testing

- [x] Conversion: `convert_hf_to_gguf.py` works
- [x] Quantization: Q4_K_M, Q8_0 working
- [x] Inference: Correct outputs in English and Chinese
- [x] CPU: ~88 t/s pp, ~8 t/s tg (16 threads, Zen4)
- [x] CUDA: **1834 t/s pp, 37 t/s tg** (RTX 4060, Q4_K_M, ngl=99)

## Prompt Format

```
<|im_start|>system
You are a helpful AI assistant.<|im_end|>
<|im_start|>user
{message}<|im_end|>
<|im_start|>assistant
```

## AI Disclosure

AI tools were used in the following assistive way:
- Helped identify architectural similarities between WeDLM and Qwen3

All code has been manually reviewed. The implementation is straightforward:
- 5 files modified, 37 lines added
- Only involves registration and case fallthrough
- I can explain every line of my code，in fact, they are pretty straightforward
Local testing completed  with CPU and CUDA backends. Full CI not yet run due to the fact that these changes seem not to have a substantial impact. let me know if a Full CI is needed


pwilkin:
Out of curiosity - did you consider adapting `diffusion-cli.cpp` to run the diffusion algorithm for this arch?:

feedseawave:
> Out of curiosity - did you consider adapting `diffusion-cli.cpp` to run the diffusion algorithm for this arch?:

Hi, Pitor. It is indeed a very interesting question. theoretically, it should work, since it is a diffusion language model. But in config.json of WeDLM it goes  "mask_token_id": null,  and it would thus cause an assertion to fail in diffusion-cli.cpp （line 610     GGML_ASSERT(mask_token_id != LLAMA_TOKEN_NULL);）For now, I haven't find out what would be a proper mask token, a reuse of qwen would be a more minimal revision to llama and since the structure of qwen and weDLM are the same, it seems a very convinient way to make things run.

am17an:
How are people supposed to run this without `diffusion-cli.cpp` support?

feedseawave:
> How are people supposed to run this without `diffusion-cli.cpp` support?
Hi, Aman. your worry is indeed very keen. Yet the model is more of a dual mode. With the diffusion feature, on the model publishing page at huggingface, the team claims a roughly 1.5x better performance with better parallel decoding. That being said, with the method above, a working model with autoregressive mode can still be generated.  But yes, It would great to unleash the full potential of the model, I will see what I can do with a follow up PR.

am17an:
What is a "dual mode"? AFAIK it was trained in a certain way, and to get the same quality of outputs you would need to replicate the decoding used in training. You would implement something like this https://huggingface.co/tencent/WeDLM-8B-Instruct/blob/main/modeling_wedlm.py#L734. This looks similar to the existing block-based decoding in LLaDA but I haven't looked into in detail. 

A PR which would add the model in full fidelity would be useful to the community at large. As such merging this PR would only have degraded outputs for the model and people would be better off using other tools

feedseawave:
> What is a "dual mode"? AFAIK it was trained in a certain way, and to get the same quality of outputs you would need to replicate the decoding used in training. You would implement something like this https://huggingface.co/tencent/WeDLM-8B-Instruct/blob/main/modeling_wedlm.py#L734. This looks similar to the existing block-based decoding in LLaDA but I haven't looked into in detail.
> 
> A PR which would add the model in full fidelity would be useful to the community at large. As such merging this PR would only have degraded outputs for the model and people would be better off using other tools

Aman，Thank you for the detailed feedback. You are absolutely right—the dual-mode training and block-based decoding are what define WeDLM's unique identity. My initial approach was indeed a runable at minimal cost, but I now see that how quality is sacrificed during the process .

I’m going to dive deeper into the code and see what I can do.

I'll keep this PR as a draft

PR #18740, opened 2026-01-10
POC: group gate_exps and up_exps + fix mxfp4 alignment for PP boost

am17an:
Instead of doing the gate and up expert `MUL_MAT_ID` separately, we group these two together (interestingly, the HF model also has them together but interleaved), so do 1 larger sparse MM (+ add) than 2. I uploaded a [GGUF](https://huggingface.co/am17an/gpt-oss-20b-q8_0.gguf) with this change for just `gpt-oss-20b`

Results:

### Master

<details>

4090:
| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp2048 |      9858.26 ± 68.48 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp4096 |     10464.32 ± 44.26 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp8192 |     10591.82 ± 11.03 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |         pp16384 |     10229.44 ± 16.46 |

5090:

| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp2048 |    14544.29 ± 124.42 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp4096 |     15873.04 ± 81.31 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp8192 |     16310.87 ± 39.83 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |         pp16384 |     13459.03 ± 63.68 |

</details>

### This PR 

<details>

4090: 

| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp2048 |     10908.66 ± 16.40 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp4096 |     11663.88 ± 46.38 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp8192 |     11815.96 ± 22.13 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |         pp16384 |      11267.29 ± 4.64 |

5090:

| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp2048 |     16002.91 ± 12.40 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp4096 |     17792.65 ± 47.00 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp8192 |    17774.20 ± 127.46 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |         pp16384 |     14217.53 ± 88.59 |

</details>

So overall a gain of ~10% in PP. 

I tried another simple but backwards incompatible change, changing the `block_mxfp4` size from 17 bytes to 20  bytes to make it 4-byte aligned in 9099a1b175c804a092f6b5431d5e752386e47163, that stacked on top of this PR gives 

<details> 

4090: 

| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| gpt-oss 20B Q8_0               |  12.93 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp2048 |     11232.48 ± 20.12 |
| gpt-oss 20B Q8_0               |  12.93 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp4096 |     12063.57 ± 36.07 |
| gpt-oss 20B Q8_0               |  12.93 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp8192 |     12236.66 ± 24.62 |
| gpt-oss 20B Q8_0               |  12.93 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |         pp16384 |     11528.88 ± 30.15 |

5090:

| model                          |       size |     params | backend    | ngl | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | -: | --------------: | -------------------: |
| gpt-oss 20B Q8_0               |  12.93 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp2048 |    16994.97 ± 143.78 |
| gpt-oss 20B Q8_0               |  12.93 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp4096 |      18900.47 ± 6.48 |
| gpt-oss 20B Q8_0               |  12.93 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |          pp8192 |     18885.38 ± 74.63 |
| gpt-oss 20B Q8_0               |  12.93 GiB |    20.91 B | CUDA       |  99 |     2048 |  1 |         pp16384 |     14500.50 ± 85.92 |

</details> 

Another ~5-10% gain for the 5090 and 1-3% gain for 4090. I uploaded another [GGUF](https://huggingface.co/am17an/gpt-oss-20b-q8_0-mxfp4.gguf) for this. This increases the size by ~15% from 12.1 GB to 13.9 GB.

I forgot to mention it also helps TG about ~3-4%

---- 
Does it make sense for optimisation 1 to happen during model-load via some flag? (we could add something similar for QKV merging). For optimisation 2 the best would be a re-pack in the CUDA backend as discussed in #18427, but I'm finding it difficult to implement that. 

AI disclosure - since I don't intend to merge this code as is, this code is mostly written by AI for quick experiments (guided by me)


am17an:
@ggerganov does it make sense to have this optimization enabled via a flag? We can disable mmap in case the flag is supplied. The gains are significant (~10%), so at least opt-in should make sense to do?

ggerganov:
I maybe see a tiny improvement on Metal, though not very significant:

<details>
<summary>Un-grouped</summary>

| model                          |       size |     params | backend    | threads | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | -------: | -: | --------------: | -------------------: |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |           pp512 |       2393.20 ± 5.85 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |          pp2048 |       2705.33 ± 4.81 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |          pp4096 |       2627.99 ± 5.81 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |          pp8192 |       2442.56 ± 1.63 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |         pp16384 |       2128.52 ± 0.66 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |            tg32 |        131.96 ± 0.53 |
build: ad966b8a8 (7685)
</details>

<details>
<summary>Grouped</summary>

| model                          |       size |     params | backend    | threads | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | -------: | -: | --------------: | -------------------: |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |           pp512 |       2434.97 ± 4.40 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |          pp2048 |       2730.94 ± 3.54 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |          pp4096 |       2649.00 ± 2.06 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |          pp8192 |       2456.41 ± 3.80 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |         pp16384 |       2135.46 ± 1.04 |
| gpt-oss 20B Q8_0               |  11.27 GiB |    20.91 B | Metal,BLAS |      16 |     2048 |  1 |            tg32 |        133.36 ± 0.60 |
build: ad966b8a8 (7685)
</details>

> @ggerganov does it make sense to have this optimization enabled via a flag? We can disable mmap in case the flag is supplied. 

I think this is a similar case to the Q, K, V fusing scenario. Merging during convert seems like the better option.

am17an:
> Merging during convert seems like the better option.

So maybe `convert_hf_to_gguf.py` should have option `--fuse-glu` and `--fuse-qkv`? I'm not sure if we can do in a generic way in convert (e.g.  how `llama_graph::build_moe_ffn` does it)

ggerganov:
> So maybe convert_hf_to_gguf.py should have option --fuse-glu and --fuse-qkv?

Yes, that's what I imagine.

> I'm not sure if we can do in a generic way in convert (e.g. how llama_graph::build_moe_ffn does it)

Not sure. If it is not possible, then on a per-model basis also works IMO.

am17an:
> If it is not possible, then on a per-model basis also works IMO.

Okay what we can do is check the `--fuse-*` paths are implemented for a model, if not then we throw an error while converting. So we can start off with the popular ones like gpt-oss and qwen3

ggerganov:
Feels like since we can do it generically in `llama_graph::build_moe_ffn`, it should be possible to also do it also in a generic way in `convert_hf_to_gguf.py`. Though I don't know how difficult it is to implement this logic. My guess it is going to be similar to the logic that handles "multiple python tensor -> one GGUF tensor" (e.g. for data types such as FP8, MXFP4, etc.).

am17an:
cc @CISC if he has any ideas

CISC:
Honestly, I think we should just change conversion to do this for all models, no option, though that requires adding support everywhere of course.

CISC:
We could do it automatically by preprocessing all the tensors in `prepare_tensors`, but that might explode memory if tensor data is modified by `modify_tensors`. Safer to do it manually, though a lot more work/code.

Edit: If we ensure all `modify_tensors` call `super` we can do it in the base class, `experts`-style.

CISC:
> Edit: If we ensure all `modify_tensors` call `super` we can do it in the base class, `experts`-style.

That would also be a handy way to refactor the `experts` handling into the base class.

CISC:
> > Edit: If we ensure all `modify_tensors` call `super` we can do it in the base class, `experts`-style.
> 
> That would also be a handy way to refactor the `experts` handling into the base class.

cc/ @compilade

am17an:
I think this would also be beneficial in the dense case, though I don't have the numbers. But a flag would be nice, we can set it to default true once this is used more widely

am17an:
@CISC I think the simplest way to do this is model by model? I'm not greatly familiar with the convert script, so I will accept your judgement here. Can we have a post-processing step `fuse_tensors` which does this?

CISC:
> @CISC I think the simplest way to do this is model by model? I'm not greatly familiar with the convert script, so I will accept your judgement here. Can we have a post-processing step `fuse_tensors` which does this?

That would complicate `prepare_tensors`, better to just do it in base class `modify_tensors`.

CISC:
> > @CISC I think the simplest way to do this is model by model? I'm not greatly familiar with the convert script, so I will accept your judgement here. Can we have a post-processing step `fuse_tensors` which does this?
> 
> That would complicate `prepare_tensors`, better to just do it in base class `modify_tensors`.

So, basically ensure that all models pass on the relevant tensors like this:
https://github.com/ggml-org/llama.cpp/blob/ad966b8a8044b218fc358c28995aecbd526752ac/convert_hf_to_gguf.py#L2946

Then fuse them here:
https://github.com/ggml-org/llama.cpp/blob/ad966b8a8044b218fc358c28995aecbd526752ac/convert_hf_to_gguf.py#L516-L519

That will ensure that the following will receive fused tensors:
https://github.com/ggml-org/llama.cpp/blob/ad966b8a8044b218fc358c28995aecbd526752ac/convert_hf_to_gguf.py#L551

CISC:
For gate/up it can be a little tricky though as there might not be a gate, so you must ensure that you don't throw away up.

am17an:
It's not possible to intercept this loop because of the `yield`s right?

https://github.com/ggml-org/llama.cpp/blob/60591f01d433f3fc7603d5273fbe361bd05a3507/convert_hf_to_gguf.py#L555

Doesn't this seem like the best place to do it though, just create a buffer store and merge?

CISC:
> It's not possible to intercept this loop because of the `yield`s right?
> 
> https://github.com/ggml-org/llama.cpp/blob/60591f01d433f3fc7603d5273fbe361bd05a3507/convert_hf_to_gguf.py#L555

Not sure what you're asking, `modify_tensors` will either:
1. Return an array with a single tensor
2. Return an array with multiple tensors
3. Yield a series of tensors

CISC:
> Doesn't this seem like the best place to do it though, just create a buffer store and merge?

No, once you're here the tensor is about to get added to the GGUF.

am17an:
What I am asking is whether something like this is possible to do here, instead of the `modify_tensors` bit

```
def prepare_tensors(self):
    gate_up_buffer: dict[int, dict[str, tuple[str, Tensor]]] = {}
    
    for name, data_torch in chain(self.generate_extra_tensors(), self.get_tensors()):
        # ... existing code ...
        
        for new_name, data_torch in self.modify_tensors(data_torch, name, bid):
            
            # --- Gate/Up fusion ---
            if self.fuse_moe_gate_up and bid is not None:
                if "ffn_gate_exps" in new_name:
                    gate_up_buffer.setdefault(bid, {})["gate"] = (new_name, data_torch)
                    if "up" in gate_up_buffer.get(bid, {}):
                        # assume fuse_gate_up exists
                        fused_name, fused_data = self._fuse_gate_up(gate_up_buffer.pop(bid))
                        new_name, data_torch = fused_name, fused_data
                    else:
                        continue  # wait for partner
                elif "ffn_up_exps" in new_name:
                    # mirror logic
                    ...
            
            # ... rest of existing processing ...
 ```

**Comment by CISC** [comment]
> What I am asking is whether something like this is possible to do here, instead of the `modify_tensors` bit

It is, but you should do that in `modify_tensors`, it doesn't belong in `prepare_tensors`.

**Comment by am17an** [comment]
`modify_tensors` has early returns, that's why I was asking if there a single place where we can make the change rather than modify every return

**Comment by CISC** [comment]
> `modify_tensors` has early returns, that's why I was asking if there a single place where we can make the change rather than modify every return

You don't have to worry about the early returns, those are usually because some tensor is ignored.

**Comment by CISC** [comment]
BTW, a complicating factor (and why I'm saying you have to do this in `modify_tensors`) is because LoRA conversion uses these methods as well, which means we will have to fuse LoRA A/B tensors as well.

**Comment by compilade** [comment]
> I think this would also be beneficial in the dense case

If I understand correctly, in the dense case fusing FFN `gate` and `up` is only possible when using `LLM_FFN_PAR` with `build_ffn`.

In the MoE case it seems like parallel FFN is always used. (and so stacking can always be done)

> We could do it automatically by preprocessing all the tensors in `prepare_tensors`, but that might explode memory if tensor data is modified by `modify_tensors`.

The tensor data is only materialized when either written (in the `write_tensors_to_file` phase), or when converted to non-tensor data (e.g. when comparing tensors with `torch.equal`, which returns a `bool`).

Keeping references to previously-written tensors is a problem. But otherwise stacking modifications shouldn't be problematic.

Making `prepare_tensors` do multiple passes of processing should be fine, since the actual transformations are aggregated and done at the time of writing. But with `--no-lazy` that would be quite memory intensive.

> > Edit: If we ensure all `modify_tensors` call `super` we can do it in the base class, `experts`-style.
> 
> That would also be a handy way to refactor the `experts` handling into the base class.

`modify_tensors` currently does both tensor name mapping and shape modifications. But it *should* work fine even when the `name` is the `new_name` (since the tensor name mappings map the target names to themselves) (this could be useful to unify the expert stacking names).

But there's some models which have already stacked `gate` and `up` tensors, in which case a general transformation in the base `modify_tensors` would need to handle that (and directly map them when possible). (Although I think that mainly applies to the dense case)

I think it's more convenient to put general transformations in an outer function around `modify_tensors` to avoid the trouble of calling the base class for *all* the tensors coming out of `modify_tensors` (because it's very easy to forget, especially when splitting tensors). That could either be `prepare_tensors`, or a new function between the two.

> BTW, a complicating factor (and why I'm saying you have to do this in `modify_tensors`) is because LoRA conversion uses these methods as well

LoRA conversion uses `prepare_tensors` as well, ~~so modifying it directly would work too.~~

If it's more convenient to change `modify_tensors` and/or if there's model-specific special cases, then it should be handled in there, otherwise it doesn't necessarily have to be.

EDIT: LoRA splits up the tensors further, so maybe not possible directly like this. See next comment (<https://github.com/ggml-org/llama.cpp/pull/18740#issuecomment-3750122398>)

**Comment by compilade** [comment]
> BTW, a complicating factor (and why I'm saying you have to do this in modify_tensors) is because LoRA conversion uses these methods as well

Ah, I see what you mean, the LoRA's `modify_tensors` splits into A and B, so stacking `gate` and `up` would need to be done between the model's `modify_tensors` and LoRA's `modify_tensors` somehow (which is where the base `modify_tensors` can be called).

So in that case it makes the base `modify_tensors` the most convenient place to change, I think, unless we restructure the modification steps in some other way.

**Comment by pwilkin** [comment]
If I understand @CISC 's comments correctly, handling this in the base case of `modify_tensors()` with a branch for gate/no gate handles like 90% of the cases, so we could simply handle outliers in their separate model classes.|

I think we should definitely aim to support any optimizations out-of-the-box, especially if they're for free (i.e. obtained during the tensor conversion phase).

**Comment by am17an** [comment]
> You don't have to worry about the early returns, those are usually because some tensor is ignored.

That's not true, we would have to do it model by model basis. For example in gpt-oss, 
```python
        # split the gate_up into gate and up
        if "gate_up_proj" in name:
            if name.endswith("_bias"):
                name_up = name.replace("gate_up_proj_bias", "up_proj.bias")
                name_gate = name.replace("gate_up_proj_bias", "gate_proj.bias")
                gate_proj_bias, up_proj_bias = data_torch[..., ::2], data_torch[..., 1::2]
                return [
                    (self.map_tensor_name(name_gate), gate_proj_bias),
                    (self.map_tensor_name(name_up), up_proj_bias)
                ]
            elif "_blocks" not in name and "_scales" not in name:
                logger.warning(f"{name} is not in MXFP4, performance may be degraded")
                name_up = name.replace("gate_up_proj", "up_proj.weight")
                name_gate = name.replace("gate_up_proj", "gate_proj.weight")
                data_torch = data_torch.transpose(-1, -2)
                gate_proj_weight, up_proj_weight = data_torch[:, ::2, :], data_torch[:, 1::2, :]
                return [
                    (self.map_tensor_name(name_gate), gate_proj_weight),
                    (self.map_tensor_name(name_up), up_proj_weight)
                ]
            else:
                # otherwise, it should already be repacked to ggml MXFP4 format
                return []
        return [(self.map_tensor_name(name), data_torch)]
```

CISC:
> > You don't have to worry about the early returns, those are usually because some tensor is ignored.
> 
> That's not true, we would have to do it model by model basis. For example in gpt-oss,

Yes, obviously, hence `"ensure that all models pass on the relevant tensors"`, those should be made into super calls.

am17an:
Then we should just replace all returns rather than cherry picking, even though that's a slightly large change.

CISC:
> Then we should just replace all returns rather than cherry picking, even though that's a slightly large change.

All non-empty returns, sure.

PR #18745, opened 2026-01-10
fix: use actual tensor embedding dimension instead of model parameter

### LINKED ISSUE #18737: Eval bug: Tensor Out of Bounds in ggml_backend_tensor_get_async using Multi Modal Embedding with Pooling
### Name and Version

$ build/bin/llama-cli --version
version: 7693 (2656c0d26)

### Operating systems

Linux

### GGML backends

CPU, HIP

### Hardware

AMD Ryzen 7 8745H for CPU only, HIP backend tested against Radeon 780M Graphics.

### Models

Qwen3-VL-Embedding: https://huggingface.co/muzzy/Qwen3-VL-Embedding-GGUF/tree/main

When trying to load Qwen3-VL-Embedding, I'm getting a tensor out of bounds error.  If I set pooling to `none`, or omit the pooling argument entirely, the error does not occur, but when I try to call /embed/ I get a message about the model not supporting the pooling method.  Per this post https://huggingface.co/Qwen/Qwen3-Embedding-0.6B/discussions/42, I believe the pooling type for the Qwen3 Embedding models should be `last`.

The GGUF's + mmproj's linked above were created  as such: `python3 convert_hf_to_gguf.py --outtype q8_0  .` and `python3 convert_hf_to_gguf.py --mmproj --outtype q8_0  .`

### First Bad Commit

_No response_

### Relevant log output

Command run: ``` build/bin/llama-server --fit off -m ../Qwen3-VL-Embedding-GGUF/Qwen3-VL-Embedding-7.6B-Q8_0.gguf --embedding -c 16000 --host 0.0.0.0 --pooling last --mmproj ../Qwen3-VL-Embedding-GGUF/mmproj-Qwen3-VL-Embedding-Q8_0.gguf ```
<details>

[llama.log](https://github.com/user-attachments/files/24541649/llama.log)

chrismuzyn:
Replace hardcoded n_embd with n_embd_tensor in embedding extraction to handle cases where tensor dimensions differ from model parameters. 

This issue was only triggered when the pool type was changed.  This fixes https://github.com/ggml-org/llama.cpp/issues/18737.  

I have tested this fix extensively with `Qwen3-VL-Embedding-8b` with pooling set to `last`.  Of course, previously it didn't work at all so there was nothing to compare to.

So I also tested extensively with `llama-embed-nemotron-8b` with pooling set to `mean` both with and without my change.  The returned embeddings matched 100%.  (They matched when genned on the same backend, CPU - CPU and Vulkan - Vulkan, they interestingly did NOT match CPU - Vulkan, but I think that must have been the case previously?  Unfortunately I somehow broke my ROCM install so I was unable to retest HIP).

Please let me know if there are additional tests I can run or if I can provide my exact GGUFs.  Will try to retest HIP if I can get it working again, but the issue was reproducible solely with the CPU backend.

### DIFF
--- a/src/llama-context.cpp
+++ b/src/llama-context.cpp
@@ -1204,6 +1204,8 @@ int llama_context::encode(const llama_batch & batch_inp) {
             case LLAMA_POOLING_TYPE_CLS:
             case LLAMA_POOLING_TYPE_LAST:
                 {
+                    const int64_t n_embd_tensor = t_embd ? t_embd->ne[0] : hparams.n_embd_inp();
+
                     // extract sequence embeddings
                     auto & embd_seq_out = embd_seq;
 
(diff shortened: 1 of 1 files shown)


ngxson:
we recently added `n_embd_out` for this exact purpose, it should be used instead.

consider adding a GGUF metadata for it. the original Qwen3-VL-Embedding model is also missing 1_Pooling, I don't think it's actually ready to be used unless Qwen team fixed it (I already reached out to them, but got no responses)

vincenthawke:
> we recently added `n_embd_out` for this exact purpose, it should be used instead.

I am guessing the windows builds don't include that update. Don't see it after running `llama-server --help`

PR #18750, opened 2026-01-11
fix: OOB reads in UGM tokenizer (precompiled_charsmap handling)

hourhl:
## Summary
This PR fixes multiple buffer overflow vulnerabilities in charsmap-related operations by adding proper bounds validation before memory access.

## Issues Fixed
- Fixes GHSA-ppcr-mg43-5hq3: Overflow in [src/llama-vocab.cpp - llm_tokenizer_ugm]
- Fixes GHSA-4383-xr9f-c744: Overflow in [src/llama-vocab.cpp - llm_tokenizer_ugm_session]

## Changes
- Added `size >= sizeof(uint32_t)` validation
- Added boundary check  before strlen operation

## Testing
- Verified that malformed inputs now trigger proper error handling
- Existing tests continue to pass

## Security Impact
Both issues could lead to out-of-bounds reads, potentially causing crashes or information disclosure in edge cases. This PR ensures proper validation before all buffer operations.


src/llama-vocab.cpp
```diff
@@ -1103,7 +1106,13 @@ struct llm_tokenizer_ugm_session {
                 throw std::runtime_error("Index out of array bounds in precompiled charsmap!");
             }
             const char * prefix_replacement = &(tokenizer.prefix_replacements)[longest_prefix_offset];
-            return { prefix_replacement, strlen(prefix_replacement), longest_prefix_length };
+            // Use strnlen to safely bound the search within prefix_replacements
+            size_t max_len = tokenizer.prefix_replacements_size - longest_prefix_offset;
+            size_t repl_len = strnlen(prefix_replacement, max_len);
+            if (repl_len == max_len && prefix_replacement[max_len - 1] != '\0') {
```
retr0reg: `prefix_replacement[max_len - 1] != '\0'` seems redundant, `strnlen(prefix_replacement, max_len);` only returns `max_len` when no \0 exists in the first` max_len` bytes; if the last byte is '\0', it returns `max_len` - 1

src/llama-vocab.cpp
```diff
@@ -1103,7 +1106,13 @@ struct llm_tokenizer_ugm_session {
                 throw std::runtime_error("Index out of array bounds in precompiled charsmap!");
             }
             const char * prefix_replacement = &(tokenizer.prefix_replacements)[longest_prefix_offset];
-            return { prefix_replacement, strlen(prefix_replacement), longest_prefix_length };
+            // Use strnlen to safely bound the search within prefix_replacements
+            size_t max_len = tokenizer.prefix_replacements_size - longest_prefix_offset;
+            size_t repl_len = strnlen(prefix_replacement, max_len);
```
retr0reg: `strnlen` might fail on MSVC

hourhl:
Thanks for the review @retr0reg . Updated per feedback:

1. Removed the redundant last-byte check—you're correct that if the loop reaches max_len,  it means no \0 was found in the first max_len bytes.

2. Replaced strnlen with a simple while loop for better portability (MSVC compatibility).The loop scans for \0 within the safe bounds; if it reaches max_len without finding one, we throw.

The fix now bounds all reads within prefix_replacements_size and detects unterminated strings.

hourhl:
Hi, @CISC  @retr0reg 
Just following up on this security fix PR. It's been about 3 weeks since submission. 
Is there anything I can do to help move this forward? Happy to make any changes if needed.
Thanks!

CISC (approved):

PR #18752, opened 2026-01-11
security: narrow down the scope of what we consider a vulnerability

ngxson:
Changes in this PR:
- Narrow down directories covered by the policy, so problems with testing tools / learning examples like (e.g. `examples/**`) will be considered as general bugs
- Prohibit reports that are predominantly written by AI
- Requires a working PoC for each report

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,12 +1,48 @@
 # Security Policy
 
+ - [**Reporting a vulnerability**](#reporting-a-vulnerability)
+ - [**Requirements**](#requirements)
+ - [**Covered Topics**](#covered-topics)
  - [**Using llama.cpp securely**](#using-llamacpp-securely)
    - [Untrusted models](#untrusted-models)
    - [Untrusted inputs](#untrusted-inputs)
    - [Data privacy](#data-privacy)
    - [Untrusted environments or networks](#untrusted-environments-or-networks)
    - [Multi-Tenant environments](#multi-tenant-environments)
- - [**Reporting a vulnerability**](#reporting-a-vulnerability)
+
+## Reporting a vulnerability
+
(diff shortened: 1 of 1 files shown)


ServeurpersoCom:
THANKS !

ServeurpersoCom (approved):
LGTM. Clear scope, PoC requirement, and AI restrictions will improve signal-to-noise ratio on security reports.

SECURITY.md
```diff
@@ -1,12 +1,48 @@
...
+Maintainers reserve the right to close the report if these requirements are not fulfilled.
+
+## Covered Topics
+
+Only vulnerabilities that fall within these parts of the project are considered valid. For problems falling outside of this list, please report them as issues.
+
+- `src/**/*`
+- `ggml/**/*`
+- `gguf-py/**/*`
+- `tools/server/*` (note: Web UI is not covered)
```
ngxson, line 41: I want to make this a bit more clear, will push a new PR for that. Btw forgot to tell you @ServeurpersoCom but we should wait for @ggerganov approval too (I guess it's fine this time, no worries)

(to Georgi: if you need any modifications, fell free to comment on my next PR)

ggerganov: We get many reports about DoS (Denial-of-Service) where certain input crashes the server for example - I don't think these vulnerabilities are significant for the project at this stage, so we should add a separate point to treat these as bugs instead of vulnerabilities.

ngxson: Thanks, I added your point in https://github.com/ggml-org/llama.cpp/pull/18754

ServeurpersoCom: Absolutely, even if some DoS attacks have the potential to escalate, if they are fixed beforehand it poses no problem

PR #18753, opened 2026-01-11
tests : refactor test-backend-sampler

ggerganov:
cont #17004
fix https://github.com/ggml-org/llama.cpp/pull/18700#issuecomment-3730527891

The new backend sampler tests have been [clogging](https://github.com/ggml-org/llama.cpp/actions/runs/20890458903/job/60020736715#step:3:22077) the `ggml-ci` recently by adding almost ~10 extra min runtime to some workflows. Optimize and deslop the implementation:

- Avoid loading the model on every test
- Better names
- Better RAII
- Remove redundant CMake test declarations
- Add device "auto" - used by default
- Use smaller `Q4_0` model instead of `F16`

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -297,7 +297,8 @@ function gg_sum_test_scripts {
 }
 
 function gg_get_model {
-    local gguf_0="$MNT/models/qwen3/0.6B/ggml-model-f16.gguf"
+    #local gguf_0="$MNT/models/qwen3/0.6B/ggml-model-f16.gguf"
+    local gguf_0="$MNT/models/qwen3/0.6B/ggml-model-q4_0.gguf"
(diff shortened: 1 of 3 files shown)


danbev (approved):

CISC:
@reeselevine Broke WebGPU CI since `greedy` test actively fails now:
https://github.com/ggml-org/llama.cpp/blob/1ffc529cbc3e48cc3ee554ea960234092b437a9c/tests/test-backend-sampler.cpp#L1124-L1134
https://github.com/ggml-org/llama.cpp/actions/runs/20897547250/job/60038427586#step:4:21395

reeselevine:
Looks like we need to implement support for `ARGMAX`, I'll prioritize this!

PR #18765, opened 2026-01-11
doc: ban AI-generated PR descriptions [no ci]

JohannesGaessler:
We could maybe explicitly ban the use of language models to generate PR descriptions. In principle I think that this is maybe one of the more legitimate use cases for language models (if you wrote the code yourself and fully understand it). But basically all of the time-wasting slop PRs that we want to reject use language models to generate PR descriptions; the people pushing the submit button simply have no other way of creating the appearance of understanding the code they submit. So potentially this change would save maintainers a bit of time.

I personally never use language models for summarizing pull requests, and I would consider the potential time savings vs. actually writing the code negligible anyways (because I already know what it does). But I don't know the workflows of other maintainers/contributors. But we can also just write down strict rules to be followed "officially" and then make exceptions on a case-by-case basis for people who have proven themselves.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ If AI is used to generate any portion of the code, contributors must adhere to t
 1. Explicitly disclose the manner in which AI was employed.
 2. Perform a comprehensive manual review prior to submitting the pull request.
 3. Be prepared to explain every line of code they submitted when asked about it by a maintainer.
-4. Using AI to respond to human reviewers is strictly prohibited.
+4. Using AI to write pull request descriptions or to respond to human reviewers is strictly prohibited.
 
 For more info, please refer to the [AGENTS.md](AGENTS.md) file.
 


ngxson (approved):
Just keep in mind that things written here is only for reference, in case people ask us about the policy. Unfortunately I think some people will still lazy and ask AI to write the description.

On the bright side, obvious mistakes like [this](https://github.com/ggml-org/llama.cpp/pull/18747#issuecomment-3735944270) can be a good way to reject a PR. Models like Claude be quite both manipulative and stupid at the same time.

pwilkin:
Yes please, if someone wants to submit a PR I expect them to write the PR description by themselves. That at least tells us whether they understand anything from the PR (in case it's AI-generated). The worst thing out there is AI-generated descriptions for AI-generated PRs and people who don't really understand either.

taronaeo (approved):

PR #18776, opened 2026-01-12
server : add arg for disabling prompt caching

rgerganov:
We have a use case where we run end to end tests which include `llama-server` and we expect all responses to be strictly deterministic. Of course we are setting a fixed `seed` in the HTTP request but it turns out this is not enough if prompt caching is enabled. Unfortunately, we cannot use the `cache_prompt` request option because it is not OpenAI compatible.

This patch adds another command line arg for disabling prompt caching but I'd be happy to discard it if there is some other way to accomplish this. Even if there is no such way, I don't insist merging this if maintainers decide this adds more confusion or our use case is not a valid one.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2867,6 +2867,14 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_CACHE_REUSE"));
     add_opt(common_arg(
+        {"--cache-prompt"},
+        {"--no-cache-prompt"},
+        string_format("whether to enable prompt caching (default: %s)", params.cache_prompt ? "enabled" : "disabled"),
+        [](common_params & params, bool value) {
+            params.cache_prompt = value;
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}));
+    add_opt(common_arg(
         {"--metrics"},
         string_format("enable prometheus compatible metrics endpoint (default: %s)", params.endpoint_metrics ? "enabled" : "disabled"),
         [](common_params & params) {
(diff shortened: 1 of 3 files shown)


ngxson (approved):
I think it's ok to add this, I thought we already had this arg, but turns out we haven't had it.

Nits: it's better to move this arg right before `--cache-reuse`, while also update the help message of `--cache-reuse` saying that it depends on `--cache-prompt`

rgerganov:
@ngxson thanks for the review, let me know if any other docs need update

common/arg.cpp
```diff
@@ -2877,10 +2856,18 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
...
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_THREADS_HTTP"));
+    add_opt(common_arg(
+        {"--cache-prompt"},
+        {"--no-cache-prompt"},
+        string_format("whether to enable prompt caching (default: %s)", params.cache_prompt ? "enabled" : "disabled"),
+        [](common_params & params, bool value) {
+            params.cache_prompt = value;
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}));
```
ngxson:
```suggestion
    ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_CACHE_PROMPT"));
```

common/arg.cpp
```diff
@@ -2877,10 +2856,18 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
...
     add_opt(common_arg(
         {"--cache-reuse"}, "N",
         string_format(
-            "min chunk size to attempt reusing from the cache via KV shifting (default: %d)\n"
+            "min chunk size to attempt reusing from the cache via KV shifting (default: %d) if cache-prompt is enabled\n"
```
ngxson:
```suggestion
            "min chunk size to attempt reusing from the cache via KV shifting, requires prompt caching to be enabled (default: %d)\n"
```

ngxson (approved):

PR #18781, opened 2026-01-12
webui : send both backend_sampling == false/true

ggerganov:
I believe the webui does not send `backend_sampling == false` in the request when the checkbox is unchecked. The goal is when the server is started with `-bs` to be able to disable backend sampling from the client.

Also I think the checkbox is currently not synchronized with the server argument like the rest of the sampling parameters. But not sure how to fix this of if it is supported for checkboxes.

### DIFF
--- a/tools/server/public/index.html
+++ b/tools/server/public/index.html
@@ -18,7 +18,7 @@
 		<div style="display: contents">
 			<script>
 				{
-					__sveltekit_1ow7rhy = {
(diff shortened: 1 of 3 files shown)


allozaur:
I will take a closer look at this

allozaur (approved):
@ggerganov I've added the parameter sync for the setting

ServeurpersoCom (approved):

PR #18785, opened 2026-01-12
CUDA: Factor out and re-use `block_reduce` function

ORippler:
This was an open TODO from #17004 on CUDA side

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -537,12 +537,15 @@ struct block_reduce_policy;
 template <typename T, typename... Ts>
 constexpr bool is_any = (std::is_same_v<T, Ts> || ...);
 
+template<typename...>
+constexpr bool ggml_cuda_dependent_false_v = false;
+
 template <typename T> struct block_reduce_policy<block_reduce_method::SUM, T> {
     static __device__ T reduce(T val) {
         if constexpr(is_any<T, float, float2, half2, int>) {
             return warp_reduce_sum(val);
         } else {
-            static_assert(false, "Unsupported type for block reduce sum");
+            static_assert(ggml_cuda_dependent_false_v<T>, "Unsupported type for block reduce sum");
         }
     }
 
@@ -556,7 +559,7 @@ template <typename T> struct block_reduce_policy<block_reduce_method::SUM, T> {
         } else if constexpr (std::is_same_v<T, int>) {
             return 0;
         } else {
-            static_assert(false, "Unsupported type for block reduce sum");
+            static_assert(ggml_cuda_dependent_false_v<T>, "Unsupported type for block reduce sum");
         }
     }
 };
@@ -566,7 +569,7 @@ template <typename T> struct block_reduce_policy<block_reduce_method::MAX, T> {
         if constexpr (is_any<T, float, half2>) {
             return warp_reduce_max(val);
         } else {
-            static_assert(false, "Unsupported type for block reduce max");
+            static_assert(ggml_cuda_dependent_false_v<T>, "Unsupported type for block reduce max");
         }
     }
 
@@ -576,7 +579,7 @@ template <typename T> struct block_reduce_policy<block_reduce_method::MAX, T> {
         } else if constexpr (std::is_same_v<T, half2>) {
             return make_half2(-INFINITY, -INFINITY);
         } else {
-            static_assert(false, "Unsupported type for block reduce max");
+            static_assert(ggml_cuda_dependent_false_v<T>, "Unsupported type for block reduce max");
         }
     }
 };
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -535,10 +535,10 @@ template<block_reduce_method method_t, typename T>
 struct block_reduce_policy;
 
 template <typename T, typename... Ts>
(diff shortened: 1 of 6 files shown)


ggml/src/ggml-cuda/common.cuh
```diff
@@ -526,6 +526,50 @@ static __device__ __forceinline__ half2 warp_prefix_inclusive_sum(half2 a) {
 #endif // FP16_AVAILABLE
 }
 
+enum warp_reduce_method {
```
am17an:
```suggestion
enum class warp_reduce_method {
```

ggml/src/ggml-cuda/common.cuh
```diff
@@ -526,6 +526,50 @@ static __device__ __forceinline__ half2 warp_prefix_inclusive_sum(half2 a) {
...
+    WARP_REDUCE_MAX,
+    WARP_REDUCE_SUM,
+};
+
+template <warp_reduce_method reduce_method, const unsigned int block_size_template = 0> static __device__ float two_stage_warp_reduce(float val, float * shared_vals) {
```
am17an:
```suggestion
template <warp_reduce_method reduce_method, const unsigned int block_size_template = 0> 
static __device__ float two_stage_warp_reduce(float val, float * shared_vals) {
```

am17an: Also rename to `reduce_method_t` perhaps

ggml/src/ggml-cuda/common.cuh
```diff
@@ -526,6 +526,50 @@ static __device__ __forceinline__ half2 warp_prefix_inclusive_sum(half2 a) {
...
+
+    float (*reduce_fun)(float);
+    switch (reduce_method) {
+        case WARP_REDUCE_MAX:
+            reduce_fun = warp_reduce_max;
+            break;
+        case WARP_REDUCE_SUM:
+            reduce_fun = warp_reduce_sum;
+            break;
+    }
```
am17an: We can make this a trait

```
struct reduction_traits<MAX> {
   using reduce_fn_t = warp_reduce_max; 
   static constexpr float sentinel = -INFINITY;
}
```

ggml/src/ggml-cuda/common.cuh
am17an: shouldn't the name simply be `block_reduce`?

ggml/src/ggml-cuda/common.cuh
```diff
@@ -526,6 +526,50 @@ static __device__ __forceinline__ half2 warp_prefix_inclusive_sum(half2 a) {
...
+        }
+        __syncthreads();
+        switch (reduce_method) {
+            case WARP_REDUCE_MAX:
+                val = -INFINITY;
+                break;
+            case WARP_REDUCE_SUM:
+                val = 0.0f;
+                break;
+        }
```
am17an: Can be simplified using the reduce_method trait

ggml/src/ggml-cuda/common.cuh
```diff
@@ -526,6 +526,50 @@ static __device__ __forceinline__ half2 warp_prefix_inclusive_sum(half2 a) {
...
+        if (lane_id < (static_cast<int>(block_size) / WARP_SIZE)) {
+            val = shared_vals[lane_id];
+        }
+        return reduce_fun(val);
+    } else {
+        return val;
+    }
```
am17an:
```suggestion
    } 
    
    return val;
```

ggml/src/ggml-cuda/common.cuh
am17an: I think should be renamed something like
```
enum class block_reduce_method {
 MAX, 
 SUM
};
```

ORippler:
Added type traits and expanded to cover all types supported for `warp_reduce_sum`/`warp_reduce_max`.

@JohannesGaessler Do you know if there was a reason we kept rms_norm_back disabled for `ncols % WARP_SIZE != 0`? https://github.com/ggml-org/llama.cpp/pull/18785/commits/8bc326a59cbb33d2adfb0b2da3da7159ece234a8 locally tests pass for me on CUDA even for the above case, so I enabled support to see how it behaves in CI

ggml/src/ggml-cuda/common.cuh
```diff
@@ -526,6 +526,83 @@ static __device__ __forceinline__ half2 warp_prefix_inclusive_sum(half2 a) {
...
+        } else if constexpr (std::is_same_v<T, half2>) {
+            return make_half2(-INFINITY, -INFINITY);
+        } else {
+            static_assert(false, "Unsupported type for block reduce max");
+        }
+    }
+};
+
+template <block_reduce_method reduce_method_t, const unsigned int block_size_template = 0, typename T>
+static __device__ T block_reduce(T val, T * shared_vals) {
```
am17an, line 588: maybe this should be called `block_reduce_1d`, users might expect `block_reduce` to reduce any dimension of block. Or perhaps we can add an assert that blockDim.y == 1

am17an (approved):

JohannesGaessler (approved):
I think it would be good to have a more templated approach to reduction like this in the CUDA backend, but I think we should aim to do this consistently for both warp-wise and block-wise reductions.

I don't remember why I put in the restriction for the backwards pass, if the corresponding `test-backend-ops grad` passes it is fine to remove.

PR #18791, opened 2026-01-12
ci, tests : use cmake to download models and remove libcurl dependency

angt:


CMakeLists.txt
```diff
@@ -141,6 +141,23 @@ if (NOT DEFINED GGML_CUDA_GRAPHS)
...
+        file(MAKE_DIRECTORY "${DEST_DIR}")
+        file(DOWNLOAD
+            "https://huggingface.co/ggml-org/models/resolve/main/${NAME}?download=true"
+            "${DEST}"
+            TLS_VERIFY ON
+        )
+    endif()
+    set(LLAMA_DL_MODEL "${DEST}" PARENT_SCOPE)
+endfunction()
+
```
ggerganov: Does this work if the file has been partially downloaded for some reason?

The `LLAMA_DL_MODEL` name is a bit overloaded as we have `GGML_BACKEND_DL` which meaning of `DL` is completely different. Can simply using `LLAMA_DOWNLOAD_MODEL` to avoid ambiguity.

CISC: CMake will always redownload the whole file, though it can skip that if using `EXPECTED_HASH`, which could be retrieved from API or `blob/main` page, but I don't think we need that.

ngxson: Or another trick is to download it to `{filename}.tmp`, then remove the `.tmp` after download is finished

angt: Let's go for the most robust way with EXPECTED_HASH

CISC (approved):

CMakeLists.txt
```diff
@@ -141,6 +141,27 @@ if (NOT DEFINED GGML_CUDA_GRAPHS)
...
+        STATUS status
+    )
+    list(GET status 0 code)
+    if(NOT code EQUAL 0)
+        list(GET status 1 msg)
+        message(FATAL_ERROR "Failed to download ${NAME}: ${msg}")
+    endif()
+    set(LLAMA_DOWNLOAD_MODEL "${DEST}" PARENT_SCOPE)
+endfunction()
+
```
ggerganov: This helper function would be better to be moved in `cmake/common.cmake`

ggerganov (approved):

FireBurn:
Hi, I think this is causing attempted downloads even when tests are off:

`cmake -C /var/tmp/portage/sci-misc/llama-cpp-9999/work/llama-cpp-9999_build/gentoo_common_config.cmake -G Ninja -DCMAKE_INSTALL_PREFIX=/usr -DLLAMA_BUILD_TESTS=OFF -DLLAMA_BUILD_SERVER=ON -DCMAKE_SKIP_BUILD_RPATH=ON -DGGML_NATIVE=0 -DGGML_RPC=ON -DLLAMA_CURL=ON -DBUILD_NUMBER=1 -DGENTOO_REMOVE_CMAKE_BLAS_HACK=ON -DGGML_CUDA=OFF -DGGML_OPENCL=OFF -DGGML_OPENMP=OFF -DGGML_VULKAN=ON -DCMAKE_INSTALL_LIBDIR=/usr/lib64/llama.cpp -DCMAKE_INSTALL_RPATH=/usr/lib64/llama.cpp -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=/var/tmp/portage/sci-misc/llama-cpp-9999/work/llama-cpp-9999_build/gentoo_toolchain.cmake /var/tmp/portage/sci-misc/llama-cpp-9999/work/llama-cpp-9999`

```
-- Found CURL: /usr/lib64/libcurl.so (found version "8.17.0")
-- Downloading tinyllamas/stories15M-q4_0.gguf from ggml-org/models...
CMake Error at cmake/common.cmake:44 (file):
  file DOWNLOAD cannot compute hash on failed download

    from url: "https://huggingface.co/ggml-org/models/resolve/main/tinyllamas/stories15M-q4_0.gguf?download=true"
    status: [6;"Could not resolve hostname"]
Call Stack (most recent call first):
  examples/eval-callback/CMakeLists.txt:9 (llama_download_model)

```

angt:
Thanks @FireBurn, could you check with this PR #18838 to make sure it solves your issue ?

PR #18795, opened 2026-01-12
Clean up t5 input builders

gabe-l-hart:
## Description

This PR removes some dead code in the graph input setters for T5 models and embedding models. It should be a functional No-Op.

(nice find @hansolosan)


CISC:
See https://github.com/ggml-org/llama.cpp/pull/18528#discussion_r2657015653

ggerganov (approved):
Ok, let's remove this as it's just causing some confusion and likely won't get utilized any time soon.

CISC (approved):

Issue #18766, opened 2026-01-11
Eval bug: Model won't load if DirectIO is used


PANCHO7532:
### Name and Version

```
~/llama.cpp $ build/bin/llama-cli --version
version: 9 (0c3b7a9)
built with Clang 21.1.8 for Android armv7l
```

### Operating systems

Other? (Please let us know in description)

### GGML backends

CPU

### Hardware

CPU: MediaTek MT6737 ARM Cortex-A53
Motorola E4 Plus

### Models

Qwen 2.5 0.5b Instruct Q8_0 - [link](https://huggingface.co/Qwen/Qwen2.5-0.5B-Instruct-GGUF/resolve/main/qwen2.5-0.5b-instruct-q8_0.gguf?download=true)

### Problem description & steps to reproduce

OS: Android 7.1 Stock with Termux

1) Prepare llama.cpp build
`cmake -B build -DBUILD_SHARED_LIBS=OFF -DGGML_STATIC=OFF -DLLAMA_BUILD_EXAMPLES=OFF -DLLAMA_BUILD_TESTS=OFF`

2) Build llama.cpp
`cmake --build build -j3`

3) Start llama.cpp
`build/bin/llama-cli --model /mnt/m_external_sd/qwen2.5-0.5b-instruct-q80.gguf`

4) See this error
```
~/llama.cpp $ build/bin/llama-cli --model /mnt/m_external_sd/qwen2.5-0.5b-instruct-q80.gguf
Loading model... -llama_model_load: error loading model: read error: Invalid argument
llama_model_load_from_file_impl: failed to load model
|common_init_from_params: failed to load model '/mnt/m_external_sd/qwen2.5-0.5b-instruct-q80.gguf'
srv    load_model: failed to load model, '/mnt/m_external_sd/qwen2.5-0.5b-instruct-q80.gguf'

Failed to load the model
```
It would refuse to load any model regardless of where it is located, tried putting the model file either on the SD card and Internal Storage, even in the same directory where the binaries are and no luck. `llama-server` is also affected. Probably other tools as well.

Interestingly enough, if i try to load it with llama-completion, it throws the same error, but also throws a segmentation fault:

```
~/llama.cpp $ build/bin/llama-completion --model /mnt/m_external_sd/qwen2.5-0.5b-instruct-q80.gguf
build: 9 (0c3b7a9) with Clang 21.1.8 for Android armv7l
main: llama backend init
main: load the model and apply lora adapter, if any
[...]

Issue #18764, opened 2026-01-11
Misc. bug: Unable to run llama-bench.exe on Vulkan


alan-l:
### Name and Version

llama-bench.exe b7668

### Operating systems

Windows

### Which llama.cpp modules do you know to be affected?

llama-bench

### Command line

```shell
llama-bench.exe --mmap 1 --n-gpu-layers 0 --flash-attn 1 --model "gpt-oss-20b-UD-Q8_K_XL.gguf"
llama-bench.exe --mmap 1 --n-gpu-layers 0 --flash-attn 0 --model "gpt-oss-20b-UD-Q8_K_XL.gguf"
```

### Problem description & steps to reproduce

Vulkan OOM error, detailed in logs below. I only included the logs for GPT-OSS because the errors in the other tested models are the same.

Tested with gpt-oss-20b-UD-Q8_K_XL.gguf, Ministral-3-14B-Instruct-2512-UD-Q4_K_XL.gguf, and gemma-3-12b-it-qat-UD-Q4_K_XL.gguf.

Seems to work with models less than 10B parameters.

Oh, and Vulkan drivers are up-to-date.

### First Bad Commit

Error shows up in b7668, no problems in b7667.

### Relevant log output

<details>
<summary>Logs</summary>
********************
Model: C:\Models\OpenAI\gpt-oss-20b-GGUF\gpt-oss-20b-UD-Q8_K_XL.gguf
********************
Test #1 Params: --mmap 1 --n-gpu-layers 0 --flash-attn 1
...
load_backend: loaded RPC backend from C:\Llama\Vulkan\ggml-rpc.dll
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = Intel(R) Iris(R) Xe Graphics (Intel Corporation) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 32768 | int dot: 1 | matrix cores: none
load_backend: loaded Vulkan backend from C:\Llama\Vulkan\ggml-vulkan.dll
load_backend: loaded CPU backend from C:\Llama\Vulkan\ggml-cpu-icelake.dll
| model                          |       size |     params | backend    | ngl | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -: | --------------: | -------------------: |
ggml_vulkan: Device memory allocation of size 423624704 failed.
[...]

PR #18801, opened 2026-01-13
llama: fix direct-io loading fallback EOF exception

0cc4m:
The issue was that `len` can be larger than the file size due to alignment padding. This is apparently fine for the direct-io path, but in case that fails, the fallback uses `std::fread`, which just returns EOF if there are not enough bytes left in the file pointer to read another object.

To fix this, I check how many bytes are left in the file pointer and if this is smaller than `len`, use that as object size for `std::fread`. In case `to_read` is 0, `std::fread` will do nothing and return 0, so I also checked this case for the EOF exception.

@JTischbein Let me know if this looks correct to you.

I have an AMD Vulkan system affected by this loading issue, and this does fix it for me.

Related issues: #18741 and maybe #18746 #18764 #18766

### DIFF
--- a/src/llama-mmap.cpp
+++ b/src/llama-mmap.cpp
@@ -244,11 +244,14 @@ struct llama_file::impl {
         }
         errno = 0;
         if (fd == -1) {
-            std::size_t ret = std::fread(ptr, len, 1, fp);
+            const size_t curr_off = tell();
+            const size_t to_read = std::min(len, size - curr_off);
+
+            std::size_t ret = std::fread(ptr, to_read, 1, fp);
             if (ferror(fp)) {
                 throw std::runtime_error(format("read error: %s", strerror(errno)));
             }
-            if (ret != 1) {
+            if (to_read > 0 && ret != 1) {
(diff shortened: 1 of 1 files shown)


src/llama-mmap.cpp
```diff
@@ -244,11 +244,14 @@ struct llama_file::impl {
         }
         errno = 0;
         if (fd == -1) {
-            std::size_t ret = std::fread(ptr, len, 1, fp);
+            const size_t curr_off = tell();
+            const size_t to_read = std::min(len, size - curr_off);
+
+            std::size_t ret = std::fread(ptr, to_read, 1, fp);
```
JTischbein:
```suggestion
            std::size_t ret = std::fread(ptr, 1, len, fp);
```

JTischbein:
@0cc4m Thank you for opening the PR!

ggerganov (approved):
`std::fread` [returns zero and performs no other action](https://en.cppreference.com/w/cpp/io/c/fread.html) when `to_read == 0` so looks OK. No issues on Mac and DGX Spark.

Issue #18783, opened 2026-01-12
Eval bug: Regression in inference of 8 bit quantized models on Neoverse v2 CPU cores


izard:
### Name and Version

Any version since commit 669696e00d60e7ffa0223ee61242318a51f33a79 produces rubbish on Q8 models inference when running on ARM Neoverse V2 cores.
This can be fixed by disabling new vectorized dot product function from that commit  (ggml_gemv_q8_0_4x8_q8_0), e.g.
```
 - #if  defined(__aarch64__) && defined(__ARM_NEON) && defined(__ARM_FEATURE_DOTPROD)
 + #if 0

```

### Operating systems

Linux

### GGML backends

CPU

### Hardware

Nvidia Grace, Graviton 4.

### Models

Any Q8 model.

### Problem description & steps to reproduce

Any inference on any Q8 model on any Arm Neoverse v2 CPU produces garbage since that new AARCH64 Neon optimized dot product commit.

### First Bad Commit

669696e00d60e7ffa0223ee61242318a51f33a79

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

izard:
@Alcpz could you please have a look? It is just one function of that big commit with many other useful optimizations seems to be misbehaving on some AARCH64 CPUs

Alcpz:
Will do. I didn't find garbage when submitting the PR, can you point me to a model that produced the garbage for you?

Edit: Any model makes me guess either qwen or llama should get me the garbage right?

izard:
I checked with gpt-oss-20b-Q8_0.gguf, Llama-3.1-8B-Instruct-UD-Q8_K_XL.gguf, Qwen3-32B-UD-Q8_K_XL.gguf. Same models quantized to non-Q8 work fine. Maybe it is not processor btw but an issue with toolchain that has some issues with that particular combination of intrinsics. My test setup is using
/usr/bin/c++ --version
c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.

Alcpz:
[...]

PR #18802, opened 2026-01-13
ggml-cpu: __fp16 and pointer++ to float_t

Alcpz:
https://github.com/ggml-org/llama.cpp/issues/18783


Alcpz:
Closing, tests show this is not sufficient.

PR #18806, opened 2026-01-13
vocab : fix attribute overrides for harmony

ggerganov:
cont #18541
fix #18794

I have accidentally broken the harmony workaround with #18541. The goal of assigning the "user-defined" token attribute to the control tokens is to force them to render in the CLI. Restore that logic.

### DIFF
--- a/src/llama-vocab.cpp
+++ b/src/llama-vocab.cpp
@@ -2436,6 +2436,9 @@ void llama_vocab::impl::load(llama_model_loader & ml, const LLM_KV & kv) {
             auto & attr = id_to_token[t.second].attr;
 
             if (t.first == "<|channel|>" || t.first == "<|message|>" || t.first == "<|start|>" || t.first == "<|constrain|>") {
+                LLAMA_LOG_WARN("%s: setting token '%s' (%d) attribute to USER_DEFINED (%u), old attributes: %u\n",
+                        __func__, t.first.c_str(), t.second, LLAMA_TOKEN_ATTR_USER_DEFINED, attr);
+
                 attr = LLAMA_TOKEN_ATTR_USER_DEFINED;
             }
         }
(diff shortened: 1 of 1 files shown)


src/llama-vocab.cpp
```diff
@@ -2489,7 +2492,7 @@ void llama_vocab::impl::load(llama_model_loader & ml, const LLM_KV & kv) {
                 special_eog_ids.erase(end_id);
 
                 auto & attr = id_to_token[end_id].attr;
-                attr = (llama_token_attr) (attr | LLAMA_TOKEN_ATTR_USER_DEFINED);
+                attr = LLAMA_TOKEN_ATTR_USER_DEFINED;
 
                 LLAMA_LOG_WARN("%s: special_eog_ids contains both '<|return|>' and '<|call|>', or '<|calls|>' and '<|flush|>' tokens, removing '<|end|>' token from EOG list\n", __func__);
```
CISC, line 2497: ~I don't know if the `<|flush|>` token (added for Solar Open) is required to be rendered, but probably doesn't matter.~ Nvm, misread that.

CISC (approved):

ngxson (approved):

ngxson:
cc @pwilkin not sure you were affected by this while working on gpt-oss chat parsing

pwilkin:
@ngxson didn't notice, as in the GPT-OSS parser seems to work fine.

pwilkin:
I did notice a warning about double EOS being appended to the prompt at the start for some model, but it didn't have a noticeable effect on performance.

ggerganov:
I think this bug only affects token text rendering. I.e., token text not being visible in the CLI tools and `/completion` endpoint, but the actual tokens are still there. So it can maybe affect certain clients depending on the parsing logic.

PR #18810, opened 2026-01-13
licenses : add minja

angt:


CISC:
It's probably going to be removed soon though, see #18462

angt:
> It's probably going to be removed soon though, see #18462

Thanks! Let’s forget this PR. I was checking all licenses and saw this one was missing. cc @ngxson

PR #18811, opened 2026-01-13
sampling : remove sampling branching in output_reserve

danbev:
This commit updates output_reserve in llama-context.cpp to always allocate sampling buffers regardless of whether sampling is needed for the current batch.

The motivation for this is to avoid reallocations and branching based on the sampling requirements of the batch.

### DIFF
--- a/src/llama-context.cpp
+++ b/src/llama-context.cpp
@@ -853,7 +853,7 @@ float * llama_context::get_sampled_probs_ith(int32_t idx) {
 float * llama_context::get_sampled_logits_ith(int32_t idx) {
     output_reorder();
 
-    if (sampling.logits == nullptr) {
+    if (!has_sampled) {
         return nullptr;
     }
 
@@ -908,7 +908,7 @@ size_t llama_context::get_sampled_candidates_count(int32_t idx) {
 size_t llama_context::get_sampled_logits_count(int32_t idx) {
     output_reorder();
 
-    if (sampling.logits == nullptr) {
+    if (!has_sampled) {
         return model.vocab.n_tokens();
     }
 
@@ -1832,11 +1832,8 @@ uint32_t llama_context::output_reserve(int32_t n_outputs, const llama_batch & ba
         has_embd   = true;
     }
 
-    // Check which sampling modes are needed for the current batch.
-    // TODO: avoid this branching by working with the worst-case
-    bool has_sampling = false;
-    bool cpu_logits   = false;
-
+    has_sampled = false;
+    bool cpu_logits = false;
     if (batch.logits) {
         for (int32_t i = 0; i < batch.n_tokens; i++) {
             if (!batch.logits[i]) {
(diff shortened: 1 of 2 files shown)


danbev:
@ggerganov When you get a chance could you take a look at this and see if this is what you had in mind?

src/llama-context.cpp
ggerganov: I was thinking that this entire logic should be removed and allocate buffers so that they are enough regardless if backend sampling is used or not, or if we need CPU logits or not.

Eventually, this TODO would have to be fixed:

https://github.com/ggml-org/llama.cpp/blob/0440bfd1605333726ea0fb7a836942660bf2f9a6/tools/server/server-context.cpp#L1166-L1169

And for that we need both pre-sampling logits (currently, called "CPU logits") and post-sampling logits to be extracted from the context. So the worst case is that both buffers are allocated at the same time.

danbev: I've added https://github.com/ggml-org/llama.cpp/pull/18811/commits/6e61d40ad5a15a34849e454e6cc2453f1741e8a0 which now always allocates backend buffers if there are any backend samplers configured in the context, regardless if the current batch includes sequences that use backend buffers or not.

> And for that we need both pre-sampling logits (currently, called "CPU logits") and post-sampling logits to be extracted from the context.

I'm not sure I'm following this terminology, so `pre-sampling logits` is when no backend sampler processing has been performed, and `post-sampling` logits is when backend-sampling has been performed?

ggerganov: Yes, the `pre-sampling` logits (a.k.a. the raw logits) are the ones we get from the end of the transformer. When backend sampling is enabled, we never copy these logits back to the host - this is one of the main advantages of using the backend sampling.

However the server can accept requests that need these logits and currently this is in conflict with the backend sampling option (hence the TODO). Not sure yet how to handle it.

But in any case, I think the simplification in this PR is OK for now.

danbev: Ah got it, thanks for the clarification!

ggerganov: The problem now is that here we will extract the pre-sampling logits unconditionally even if all sequences are using backend sampling:

https://github.com/ggml-org/llama.cpp/blob/6e61d40ad5a15a34849e454e6cc2453f1741e8a0/src/llama-context.cpp#L1663-L1679

danbev: I'll take a look as this :+1:

src/llama-context.cpp
```diff
@@ -1456,6 +1456,23 @@ static void copy_tensor_async_candidates(
     }
 }
 
+static bool needs_cpu_logits(const llama_ubatch & ubatch, const std::map<llama_seq_id, llama_sampler *> & samplers) {
```
ggerganov: Let's use the term "raw logits" - think it fits better.

src/llama-context.cpp
```diff
@@ -1814,6 +1825,8 @@ int llama_context::decode(const llama_batch & batch_inp) {
 //
 
 uint32_t llama_context::output_reserve(int32_t n_outputs, const llama_batch & batch) {
+    GGML_UNUSED(batch);
```
ggerganov: We can now remove the `batch` argument from this method.

ggerganov (approved):
Should be OK to merge after the CI pass

Issue #17106, opened 2025-11-08
Misc. bug: Vulkan output is gibberish


mimi89999:
### Name and Version

```
$ ./llama-cli --version
load_backend: loaded RPC backend from /home/michel/llm/llama-b6989-bin-ubuntu-vulkan-x64/build/bin/libggml-rpc.so
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = Intel(R) Iris(R) Xe Graphics (TGL GT2) (Intel open-source Mesa driver) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 65536 | int dot: 1 | matrix cores: none
load_backend: loaded Vulkan backend from /home/michel/llm/llama-b6989-bin-ubuntu-vulkan-x64/build/bin/libggml-vulkan.so
load_backend: loaded CPU backend from /home/michel/llm/llama-b6989-bin-ubuntu-vulkan-x64/build/bin/libggml-cpu-icelake.so
version: 6989 (eeee367de)
built with cc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0 for x86_64-linux-gnu
```

### Operating systems

Linux

### Which llama.cpp modules do you know to be affected?

llama-cli

### Command line

```shell
./llama-cli -hf ggml-org/SmolLM3-3B-GGUF -p "Hello"
```

### Problem description & steps to reproduce

The output I got is gibberish:

```
user
Hello
assistant
upported D F;

 D R P23D323D4 PP
D3D3DDD
> 
```

When using the CPU backend bu setting `-ngl 0` or simply by removing the Vulkan module, my output is correct

### Relevant log output

```shell
load_backend: loaded RPC backend from /home/michel/llm/llama-b6989-bin-ubuntu-vulkan-x64/build/bin/libggml-rpc.so
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = Intel(R) Iris(R) Xe Graphics (TGL GT2) (Intel open-source Mesa driver) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 65536 | int dot: 1 | matrix cores: none
load_backend: loaded Vulkan backend from /home/michel/llm/llama-b6989-bin-ubuntu-vulkan-x64/build/bin/libggml-vulkan.so
[...]

PR #18814, opened 2026-01-13
vulkan: work around Intel ANV fp16 bug in MMQ

0cc4m:
This should work around the issue #17106 hopefully. It looks to me like a driver bug in ANV.


jeffbolznv (approved):

jeffbolznv:
Looks fine. I wonder if it goes away with spirv-opt disabled? Might be a spirv-tools bug rather than driver bug.

0cc4m:
> Looks fine. I wonder if it goes away with spirv-opt disabled? Might be a spirv-tools bug rather than driver bug.

But wouldn't it also affect AMD and Nvidia then?

jeffbolznv:
Probably, but it might depend how it's broken. It's fine to submit as-is.

PR #18815, opened 2026-01-13
CUDA: fix allignment on register spill for FA

JohannesGaessler:
Fixes the same issue as described in https://github.com/ggml-org/llama.cpp/pull/18799 (I think). I was unable to reproduce the issue but based on the reported information I think it has to do with register spilling. By default the arrays defined in the FA kernels are in registers where there are no alignment issues. However, if these arrays spill to local memory they are only guaranteed to be aligned to 8 bytes so the kernel can crash. Explicitly requiring an alignment of 16 bytes should fix those cases. @danbev please confirm whether this PR fixes your issues.

### DIFF
--- a/ggml/src/ggml-cuda/fattn-common.cuh
+++ b/ggml/src/ggml-cuda/fattn-common.cuh
@@ -59,7 +59,7 @@ static __device__ __forceinline__ float vec_dot_fattn_vec_KQ_f16(
 
 #pragma unroll
     for (int k_KQ_0 = 0; k_KQ_0 < D/2; k_KQ_0 += nthreads*cpy_ne) {
-        half2 tmp[cpy_ne];
+        __align__(16) half2 tmp[cpy_ne];
         ggml_cuda_memcpy_1<sizeof(tmp)>(tmp, K_h2 + k_KQ_0 + (threadIdx.x % nthreads)*cpy_ne);
 #pragma unroll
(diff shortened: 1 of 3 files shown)


danbev:
@JohannesGaessler Thanks for looking into this! 
The original crash is gone with the current commits but there is one more alignment to allow llama-batched-bench pass successfully:
```console
diff --git a/ggml/src/ggml-cuda/fattn-common.cuh b/ggml/src/ggml-cuda/fattn-common.cuh
index 6b55f784f..797100577 100644
--- a/ggml/src/ggml-cuda/fattn-common.cuh
+++ b/ggml/src/ggml-cuda/fattn-common.cuh
@@ -59,7 +59,7 @@ static __device__ __forceinline__ float vec_dot_fattn_vec_KQ_f16(
 
 #pragma unroll
     for (int k_KQ_0 = 0; k_KQ_0 < D/2; k_KQ_0 += nthreads*cpy_ne) {
-        half2 tmp[cpy_ne];
+        __align__(16) half2 tmp[cpy_ne];
         ggml_cuda_memcpy_1<sizeof(tmp)>(tmp, K_h2 + k_KQ_0 + (threadIdx.x % nthreads)*cpy_ne);
 #pragma unroll
         for (int k_KQ_1 = 0; k_KQ_1 < cpy_ne; ++k_KQ_1) {

```
With this change I'm able to run a debug build without the program crashing.

danbev (approved):

PR #17501, opened 2025-11-25
feat: support GGML_BACKEND_DIR environment variable

o7si:
Currently, CMake supports setting the `GGML_BACKEND_DIR` variable to specify the backend installation path:

https://github.com/ggml-org/llama.cpp/blob/583cb83416467e8abf9b37349dcf1f6a0083745a/ggml/CMakeLists.txt#L80

This path gets embedded into the code as a search path:

https://github.com/ggml-org/llama.cpp/blob/583cb83416467e8abf9b37349dcf1f6a0083745a/ggml/src/ggml-backend-reg.cpp#L521-L531

However, in some cases, users may need to adjust the backend path for various reasons, which would invalidate the embedded path in the code.

This PR adds an environment variable to improve this situation by allowing runtime path configuration.

Related Issue:
- #17491
- #15074

PR #18817, opened 2026-01-13
ggml-backend: Separate dynamic lib install and search paths, add relative search

DaAwesomeP:
I ran into some issues when trying to build a self-contained macOS `.app` bundle using GGML dynamic shared libraries. The directory to install the shared libraries may be different from the search path, and the search path needs to be relative. In general this just provides some more flexibility. 

This pull splits  the `GGML_BACKEND_DIR` CMake option into 3 separate options:
- `GGML_BACKEND_ABSOLUTE_SEARCH_DIR`: Absolute path to a directory to search for dyanmic shared libraries (same behavior as previous `GGML_BACKEND_DIR` but in a separate var from install)
- `GGML_BACKEND_RELATIVE_TO_EXE_SEARCH_DIR`: Allows specifying a dynamic shared library search path relative to the executable (new feature)
- `GGML_BACKEND_INSTALL_DIR`: Absolute path to install the shared libraries to with during the CMake install phase (same behavior as previous `GGML_BACKEND_DIR` but in a separate var from search path)

I left `GGML_BACKEND_DIR` in place to avoid breaking change, though it now shows a deprecation warning.

Issues this solves:

1. The dynamic libraries need to be stored in `./Contents/Frameworks/`, but the executable is in `./Contents/MacOS/`. This requires a relative search path of `../Frameworks/` from the executable. The existing `GGML_BACKEND_DIR` only works for absolute paths, which does not work for a portable `.app` bundle.
    
    Here is the directory tree of an example `.app` bundle for reference:
    ```
    MyApp.app
    └── Contents
        ├── Frameworks
        │   ├── libggml-base.0.9.5.dylib
        │   ├── libggml-base.0.dylib -> libggml-base.0.9.5.dylib
        │   ├── libggml-cpu-apple_m1.so
        │   ├── libggml-cpu-apple_m2_m3.so
        │   ├── libggml-cpu-apple_m4.so
        │   ├── libggml-metal.so
        │   ├── libggml.0.9.5.dylib
        │   ├── libggml.0.dylib -> libggml.0.9.5.dylib
        │   └── libomp.dylib
        ├── Info.plist
        ├── MacOS
        │   └── MyApp
        ├── PkgInfo
        └── Resources
            └── Icon.icns
    ```
2. Since these are dynamically loaded shared libraries and are not linked from the main executable, they are not automatically pulled into the bundle when running CMake's `fixup_bundle()` function to make it portable. So, we need to copy these into the bundle during the install phase. However, this absolute install path is only used by CMake and should not be baked into the binary as a search path.


ggerganov:
Looks like an improvement for macOS. @ckastner @o7si any concerns, suggestions?

o7si:
Since `llama.cpp` loads backends using the `dlopen` mechanism at runtime, traditional environment variables like `LD_LIBRARY_PATH` or `DYLD_LIBRARY_PATH` don't take effect. The current code uses the `GGML_BACKEND_DIR` macro (injected by CMake) to set the search path, allowing dlopen to locate the corresponding `.so` or `.dylib` files.

While adding new macro definitions through CMake can certainly meet the requirements, I think using environment variables might be a more elegant solution. 

In many scenarios (especially in companies with strict policies), different applications need to ensure `hash consistency for their shared libraries`. The macro injection approach can break this hash consistency, requiring each application that wants to integrate llama.cpp to recompile.

I previously submitted a PR (#17501) to try to address similar issues.

ckastner:
I concur with the problem, but I think there is a simpler solution: `GGML_BACKEND_DIR` could simply support relative paths as well. Easy to implement, and avoids two new complex options + a deprecation.

"Relative" can reasonably mean either to exe, lib, or cwd. I think it would be good enough to simply assume exe for now like this PR does, this can always be extended to the other options later if needed.

Regarding `GGML_BACKEND_INSTALL_DIR`, if I'm understanding this correctly, then the installation path on macOS would be deterministic (`../Content/Frameworks`), so I think this option could be avoided, too.

I think having an envvar with the ability to override this is a good idea, will comment on the other PR for this.

DaAwesomeP:
> While adding new macro definitions through CMake can certainly meet the requirements, I think using environment variables might be a more elegant solution.

When running from a `.app`, does macOS add the bundle's `Frameworks` dir to `DYLD_LIBRARY_PATH`? If so, then I will do that instead and remove the extra search path vars. I think the install path still needs to be separate though (see below)

If you mean a user-specified env var, I don't think that works since the goal is for a user to open a `.app` from Finder. I would not expect a user to set an env var prior to double-clicking on a self-contained app.

> `GGML_BACKEND_DIR` could simply support relative paths as well. Easy to implement, and avoids two new complex options + a deprecation.

I see, so I could check if the first path segment contains only `.` characters (or doesn't begin with a `/`?) and case based on that. As you said it is less clear what it is relative from, but if you'd prefer I can switch to that.

> Regarding GGML_BACKEND_INSTALL_DIR, if I'm understanding this correctly, then the installation path on macOS would be deterministic (../Content/Frameworks), so I think this option could be avoided, too.

So this is more tricky. Because CMake provides multiple ways to build app bundles--or a user could build it themselves in an install script--finding the base path of where the bundle should install to is hard to determine from the dependency. When GGML is built as a dependency of another project, I'm not sure the dependency can detect at all if it should be added to a bundle and where that bundle is. We could have `MACOS_BUNDLE_PATH_OPTION`, but I think this is less flexible and less clear. The best solution I can come up with is that the parent app that defines the bundle should tell GGML where to install the dylibs to.

Regardless of the `.app` scenario, I think the search and install vars need to split. Specifying an absolute install path does not necessarily mean that absolute path should get baked into the binary for searching. That install path can be specific to the build/install process (i.e. has a `Users` or `/home` or `builds/Release` in it) to get the binaries in the right place, which should then be found using relative means.

o7si:
> If you mean a user-specified env var, I don't think that works since the goal is for a user to open a .app from Finder. I would not expect a user to set an env var prior to double-clicking on a self-contained app.

However, with environment variable support, the `.app` might be able to programmatically set the search path internally via system APIs - this would be completely transparent to users who still just double-click as usual.

ckastner:
> I see, so I could check if the first path segment contains only `.` characters (or doesn't begin with a `/`?) and case based on that. 

Call `is_relative()` on the path,  it's simpler and more portable.

> As you said it is less clear what it is relative from, but if you'd prefer I can switch to that.

It's significantly less cognitive load than adding two new options. I think relative-to-exe is the intuitive choice but one could simply also additionally try relative-to-lib if ever needed. I would not try relative-to-cwd for security reasons.

> Regardless of the .app scenario, I think the search and install vars need to split. Specifying an absolute install path does not necessarily mean that absolute path should get baked into the binary for searching.

The search path can already be controlled via `GGML_BACKEND_DIR`, making it support a relative path is reasonable and simple, so we can check that off.

If the install path is deterministic, something like `<APP_BUNDLE_ROOT>/Content/Frameworks/`, than I would adjust the install code for just for that case. Why introduce a new variable when it can only take one correct value anyway.

ckastner:
> It's significantly less cognitive load than adding two new options. I think relative-to-exe is the intuitive choice but one could simply also additionally try relative-to-lib if ever needed

Just realized: there can be only one relative path if we're also using `GGML_BACKEND_DIR` during install, because we can only install to one destination.

Though personally I'd lean towards that one destination until an actual need arises.

DaAwesomeP:
Just a note--I'm using GGML as a dependency in another project (not llama.cpp, but the GGML PRs go in this repo). 

> However, with environment variable support, the .app might be able to programmatically set the search path internally via system APIs - this would be completely transparent to users who still just double-click as usual.

Can you give an example of this? This seems a lot less clear/simple than an implementer setting a CMake var.

> If the install path is deterministic, something like <APP_BUNDLE_ROOT>/Content/Frameworks/, than I would adjust the install code for just for that case. Why introduce a new variable when it can only take one correct value anyway.

Yeah the problem is figuring out that `<APP_BUNDLE_ROOT>`. The parent project depending on GGML knows it, but GGML can't figure it out. The `./Content/Frameworks/` part should be deterministic, but the `<APP_BUNDLE_ROOT>` part depends entirely on the build config.

o7si:
@DaAwesomeP  Something like this:
```swift
// Get the executable path and compute the Frameworks directory
// macOS app bundle structure: MyApp.app/Contents/MacOS/MyApp
// We need to find backends in: MyApp.app/Contents/Frameworks/
char exe_path[PATH_MAX];
uint32_t size = sizeof(exe_path);
if (_NSGetExecutablePath(exe_path, &size) == 0) {
    std::string path(exe_path);
    // Replace /MacOS/ with /Frameworks/ to get the backend directory
    auto pos = path.rfind("/MacOS/");
    if (pos != std::string::npos) {
        std::string frameworks_dir = path.substr(0, pos) + "/Frameworks";
        
        // Set GGML_BACKEND_DIR so dlopen can find the backend .so files
        setenv("GGML_BACKEND_DIR", frameworks_dir.c_str(), 1);  // <-- NOTE HERE

        // Load all available backends (CPU, Metal, BLAS, etc.)
        ggml_backend_load_all();
    }
}
```
The app discovers its bundle path at runtime and sets `GGML_BACKEND_DIR` automatically.

DaAwesomeP:
@o7si I see, though this feels a lot more complicated than simply concatenating `../Frameworks` to the existing executable search method if you know this is necessary at build time. Now the implementer has to define their own method based on `_NSGetExecutablePath()` while GGML is already performing the same task.

I also don't think passing a value like this at runtime as an env var is very intuitive. If GGML was its own separate CLI interface called with command line arguments then sure, but it is being used as a CPP library here.

Separately: From some initial searching it doesn't look like the `DYLD_LIBRARY_PATH` includes the bundle local `../Frameworks` folder. The app encodes each bundled linked library path individually like `@executable_path/../Frameworks/libexample.dylib`, so bundled libraries never touch `DYLD_LIBRARY_PATH`.

I guess I'm not fully understanding the hesitation with encoding an optional relative search path from `GGML_BACKEND_DIR`, especially since GGML already supports absolute search paths with this method.

o7si:
@DaAwesomeP  There are essentially two approaches here:

- Adding a new macro definition (injected via `CMake`).
- Adding an environment variable.

The argument for the former is **usability** (ease of integration), while the latter prioritizes **flexibility**.

@ggerganov what are your thoughts on this?

Edit: Or we could implement both – they're not mutually exclusive. :D

DaAwesomeP:
Hi @ggerganov, checking in on this. Thanks!

DaAwesomeP:
Hi @ggerganov checking in again!

PR #18822, opened 2026-01-14
hexagon: support for OP_CPY, host buffers now optional, hvx-utils refactoring and optimizations  

max-krasnyansky:
This a bit loaded but for a good reason :)
The Hexagon backend is now able to do nearly-complete offload (all Ops) for Qwen and Llama family of models with Q4/Q8/MXFP4/F16/F32 tensors.

- OP_CPY is now supported for all f32 <-> f16 flavors
- Environment variable handling has been made consistent 
   `GGML_HEXAGON_FEATURE_X=1` means enabled, `..._X=0` means disabled
   previously we sometimes checked only for the presence (i.e. `getenv()` returning non null)
- Host buffers are now optional and can be disabled for most models with `GGML_HEXAGON_HOSTBUF=0`
   If host buffers are disabled we no longer register the REPACK buffers.
   The intent is to make this the default in the next iteration.
- As part of the OP_CPY work I refactored, optimized and cleaned up `hvx-utils` into a nice header-only library.
   This removes a lot of code duplication, makes naming consistent and in general improves HVX code.
   
All changes have been tested on Snapdragon Gen3,4,5 devices with Qwen3, Llama-3.2, LFM2, and GPT-OSS-20B.
We're passing all previously supported `test-backend-ops` tests and some new ones `CPY`.


lhez (approved):
Looks good (it is indeed a bit loaded).

PR #16701, opened 2025-10-21
model : add PaddleOCR

ngxson:
This is a very early WIP

Progress:

- ~~Only the language model is working now. The vision encoder is not yet implemented~~
- ~~Vision encoder is added, but not yet numerically correct~~
- Model generate hallucinated text, likely because of the projector being incorrect

PR #18825, opened 2026-01-14
model: Add PaddleOCR-VL model support

megemini:
## Edit 20260310

The template and model files have been updated and uploaded by PaddleOCR team: https://huggingface.co/PaddlePaddle/PaddleOCR-VL-1.5-GGUF

---

## Edit 20260228

If encountering `llama-server` with error but `llama-cli` works fine, try to use `--chat-template-file` with `llama-server`

```shell
./build/bin/llama-server -m /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-1.5-GGUF.gguf \
  --mmproj /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-1.5-GGUF-mmproj.gguf \
  --port 8111 --host 0.0.0.0 --ctx-size 131072 -n 4096 --temp 0 --jinja \
  --chat-template-file /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-1.5/chat_template_llama.jinja
```
 
`chat_template_llama.jinja` for PaddleOCR-VL-1.5 uploaded here: https://modelscope.cn/models/megemini/PaddleOCR-VL-1.5-GGUF/files

for PaddleOCR-VL uploaded here: https://modelscope.cn/models/megemini/PaddleOCR-VL-GGUF/files

this issue is reported.

---

Add PaddleOCR-VL model support.

Test with some images:

1. A receipt

with command:

``` shell
./build/bin/llama-cli -m /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-GGUF.gguf \
  --mmproj /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-GGUF-mmproj.gguf \
  --color on\
  --image /home/shun/Pictures/1640.jpeg \
  --prompt "OCR:"
```

Screenshot from 2026-01-14 13-54-35

2. A table

paddleocr

with command:

``` shell
./build/bin/llama-cli -m /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-GGUF.gguf \
  --mmproj /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-GGUF-mmproj.gguf \
  --color on\
  --image /home/shun/Pictures/paddleocr.jpg \
  --prompt "Table Recognition:"
```

Screenshot from 2026-01-14 13-56-26

can be formatted:

Screenshot from 2026-01-14 13-56-53

p.s. Thanks to @ngxson https://github.com/ggml-org/llama.cpp/pull/16701 

### Update

The model converted by commands:

``` shell
python3 convert_hf_to_gguf.py /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL \
  --outfile /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-GGUF.gguf \
  --outtype bf16 \
  --verbose
python3 convert_hf_to_gguf.py /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL \
  --mmproj \
  --outfile /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-GGUF-mmproj.gguf \
  --outtype bf16 \
  --verbose
```

or, can be downloaded from : https://modelscope.cn/models/megemini/PaddleOCR-VL-GGUF/summary


gguf-py/gguf/constants.py
```diff
@@ -284,6 +284,8 @@ class Clip:
     class ClipVision:
         PROJECTOR_TYPE      = "clip.vision.projector_type" # for mixed modality models
         IMAGE_SIZE          = "clip.vision.image_size"
+        MAX_PIXELS          = "clip.vision.max_pixels"
+        MIN_PIXELS          = "clip.vision.min_pixels"
```
ngxson: Please use the correct naming from https://github.com/ggml-org/llama.cpp/pull/18719

src/models/paddleocr.cpp
ngxson, line 1: what's the different between this and `qwen2vl.cpp`? seems like 100% identical

convert_hf_to_gguf.py
```diff
@@ -3855,6 +3866,49 @@ def prepare_tensors(self):
                 raise ValueError(f"Unprocessed experts: {experts}")
 
 
+@ModelBase.register("PaddleOCRVLForConditionalGeneration")
+class PaddleOCRModel(Ernie4_5Model):
+    model_arch = gguf.MODEL_ARCH.PADDLEOCR
```
ngxson, line 3864: I don't ge why you need to make this overcomplicated, what's wrong with my code on https://github.com/ggml-org/llama.cpp/pull/16701 where I just reuse the same `Ernie4_5Model` arch?

ngxson (changes requested):
this PR looks a bit suspicious, please explicitly specify if you're using AI to generate it

megemini:
> this PR looks a bit suspicious, please explicitly specify if you're using AI to generate it

@ngxson 

First, thanks for your review, but, NO! Absolutely not!

The initial commit was based on your PR: https://github.com/ggml-org/llama.cpp/pull/16701.

However, as you mentioned, `Model generate hallucinated text, likely because of the projector being incorrect.`

Therefore, I tried to compare the code with the implementation from https://github.com/huggingface/transformers/pull/42178, which adds PaddleOCR-VL support to transformers.

To be frank, I’ve only been working with the llama.cpp source code for about two weeks, so I DO USE AI to help me understand the code. However, since llama.cpp is a C++ project, the AI isn't very helpful, so I'm forced to work on it all by myself!

PaddleOCR-VL is almost identical to ERNIE4.5, but there are two main differences from your PR:

- PaddleOCR-VL uses `mrope` instead of `rope`, which requires `rope_sections`.
- The `positions` tensor is different from ERNIE4.5 or Qwen2VL.

https://github.com/ggml-org/llama.cpp/blob/9d5a701ebe06ac300049754167ec88c00bc78bd0/tools/mtmd/clip.cpp#L3529-L3550

Additionally, ERNIE4.5 had a minor issue where `add_prefix_space` needs to be added. 

https://github.com/ggml-org/llama.cpp/blob/9d5a701ebe06ac300049754167ec88c00bc78bd0/convert_hf_to_gguf.py#L3728-L3736

Or, PaddleOCR-VL prompt is changed from `OCR:` to `_OCR:`, which makes PaddleOCR-VL error-prone, because it is trained on `OCR:`, and the generative model is only 0.3B, which may not be so smart.

> what's the different between this and qwen2vl.cpp? seems like 100% identical

The `src/models/paddleocr.cpp` is based on `src/models/ernie4-5.cpp`, but uses `ggml_rope_multi` instead of `ggml_rope_ext`.

While `src/models/paddleocr.cpp`, `src/models/ernie4-5.cpp` and `src/models/qwen2vl.cpp` are almost identical, but, 

https://github.com/ggml-org/llama.cpp/blob/9d5a701ebe06ac300049754167ec88c00bc78bd0/src/models/paddleocr.cpp#L73-L75

https://github.com/ggml-org/llama.cpp/blob/36f0132464096e49ed344cdeeee65e39e2b43b14/src/models/qwen2vl.cpp#L68-L70

ERNIE4.5 and PaddleOCR-VL do not require `model.layers[il].bo`, although they behave the same if `model.layers[il].bo` is `NULL`.

> I don't ge why you need to make this overcomplicated, what's wrong with my code on https://github.com/ggml-org/llama.cpp/pull/16701 where I just reuse the same Ernie4_5Model arch?

As I mentioned above, PaddleOCR-VL requires `mrope` and `rope_sections`

https://github.com/ggml-org/llama.cpp/blob/9d5a701ebe06ac300049754167ec88c00bc78bd0/src/llama-model.cpp#L8318-L8320

ngxson:
Thanks for the explanation, that's useful. I think it should be put onto the PR description to make reviewing easier.

Indeed, it seems like the model is just qwen2vl under the hood, minus small difference in tokenizer. At least for the language model, it can just reuse the same qwen2vl arch if the bo tensor is not present in the weight.

I'll come back to this after https://github.com/ggml-org/llama.cpp/pull/18719 being merged

gguf-py/gguf/constants.py
megemini: done ~

megemini:
### Update 20260129

[PaddleOCR-VL-1.5](https://huggingface.co/PaddlePaddle/PaddleOCR-VL-1.5) released.

Model exported: https://modelscope.cn/models/megemini/PaddleOCR-VL-1.5-GGUF/summary

Test with new type of image:

seal

with command and result:

Screenshot from 2026-01-29 20-33-22

Compare with gt:

Screenshot from 2026-01-29 20-38-38

ngxson (changes requested):
It does not work in my test:

```sh
gh pr checkout 18825

# download https://huggingface.co/PaddlePaddle/PaddleOCR-VL
# then convert
python convert_hf_to_gguf.py PaddleOCR-VL --outtype f16 --outfile mmproj.gguf --mmproj
python convert_hf_to_gguf.py PaddleOCR-VL --outtype f16 --outfile model-f16.gguf --mmproj

# test
./build/bin/llama-mtmd-cli -m ../models/PaddleOCR-VL/model-f16.gguf -mm ../models/PaddleOCR-VL/mmproj.gguf --image ./tools/mtmd/test-1.jpeg -p "OCR:" --temp 0 -n 1024 --jinja
```

Generated:

```
main: loading model: ../models/PaddleOCR-VL/model-f16.gguf
WARN: This is an experimental CLI for testing multimodal capability.
      For normal use cases, please use the standard llama-cli
encoding image slice...
image slice encoded in 464 ms
decoding image batch 1/1, n_tokens_batch = 391
image decoded (batch 1/1) in 349 ms

The New York Times
V
OL
C
E
X
X
X
X
X
X
X
X
X
X
X
X
X
X
```

Image included in the repo: https://github.com/ggml-org/llama.cpp/blob/master/tools/mtmd/test-1.jpeg

---

With image placed after prompt: `-p "OCR:<__media__>"`

Better, but still goes into non-stop loop:

```
image slice encoded in 288 ms
decoding image batch 1/1, n_tokens_batch = 391
image decoded (batch 1/1) in 4 ms

The New York Times
VOL.C.E.VII. No. 46.711
N E Y O R . M O D . Y . J U L . T . I . M . 1 . 9 . X
10 CENTS
MEN WALK ON MOON
ASTRONAUTS LAND ON PLAIN; COLLECT ROCKS, PLANT FLAG
Voice From Moon: 'Eagle Has Landed'
B O E D E S T H E L A R E M O D E S T H O N T, T R O N Q U L I E D E S T H E L A R E M O D E S T H O N T, WE ARE ON THE
B O E D E S T H E L A R E M O D E S T H O N T, WE ARE ON THE
B O E D E S T H E L A R E M O D E S T H O N T, WE ARE ON THE
B O E D E S T H E L A R E M O D E S T H O N T, WE ARE ON THE
B O E D E S T H E L A R E M O D E S T H O N T, WE ARE ON THE
```

megemini:
The possible reasons for this situation could be:
When PaddleOCR-VL is called through PaddleOCR, it is a two-stage task. 
In the first stage, the PP-DocLayout model is used to cut the original image and capture the parts that need to be recognized, such as the title, body text, charts, etc. 
In the second stage, the PaddleOCR-VL model is used to recognize each subgraph. This is the complete process of PaddleOCR calling PaddleOCR-VL.
However, tools such as llama.cpp and transformers hand over the entire image to PaddleOCR-VL for recognition. Therefore, when the layout in the image is messy, there may be a decrease in the model's recognition ability.
Referring to this [test image](https://github.com/ggml-org/llama.cpp/blob/master/tools/mtmd/test-1.jpeg), the layout is quite complex, and the small font parts are too small to distinguish with the naked eye, which may reduce the recognition ability of the model.

As a reference, I also used transformers to recognize this image, and the reference code is as follows:

``` python

from transformers import AutoProcessor, AutoModelForImageTextToText, BitsAndBytesConfig, TextIteratorStreamer
import torch
from threading import Thread

path = "/home/aistudio/paddleocr-vl/PaddleOCR-VL"
processor = AutoProcessor.from_pretrained(path, local_files_only=True, use_fast=True)

model = AutoModelForImageTextToText.from_pretrained(
    path,
    device_map="cuda:0",
    local_files_only=True
)

messages = [
    {
        "role": "user",
        "content": [
            {"type": "image", "url": "/home/aistudio/paddleocr-vl/test-1.jpeg"},
            {"type": "text", "text": "OCR:"},
        ]
    }
]

inputs = processor.apply_chat_template(
    messages,
    add_generation_prompt=True,
    tokenize=True,
    return_dict=True,
    return_tensors="pt",
).to(model.device)

streamer = TextIteratorStreamer(processor, skip_prompt=True, skip_special_tokens=True)

def generate_stream():
    model.generate(
        **inputs,
        max_new_tokens=1024,
        streamer=streamer,
        temperature=0.7,
        top_k=50,
        top_p=0.95,
        do_sample=True
    )

thread = Thread(target=generate_stream, daemon=True)
thread.start()

print("Output: ", end="", flush=True)
try:
    for text in streamer:
        print(text, end="", flush=True)
    print()
except KeyboardInterrupt:
    print("\n[Interrupted]")
    thread.join(timeout=1)


```

The recognition result may also exhibit a 'non-stop loop' phenomenon:

Screenshot from 2026-01-29 23-36-46

megemini:
I test another image from https://modelscope.cn/models/PaddlePaddle/PaddleOCR-VL-1.5 :

spotting

with command

``` shell
./build/bin/llama-mtmd-cli -m /home/aistudio/paddleocr-vl/model-f16-1.5.gguf -mm /home/aistudio/paddleocr-vl/mmproj-1.5.gguf --image /home/aistudio/paddleocr-vl/spotting.png -p "Spotting:" --temp 0 -n 1024 --jinja
```

the result:

Screenshot from 2026-01-29 23-53-58

can be formatted:

Screenshot from 2026-01-29 23-59-16


@ngxson Can you try some other pictures, preferably clearer？

ngxson:
I cannot the results in your last comment:

image

But in anyways, imagine if user try this using the exact command found on your OP, there is a chance it will be gibberish results. I don't think "just tell user to change the image" is a good option either.

Instead, you must test it more thoroughly and document any abnormal things on your OP. Currently, by reading your OP, any users can think that the model should "just work" out of the box without anything else.

megemini:
I completely agree with what you said! I have updated the links of the pictures in the last comment. Can you see them now?

I am trying to contact the PaddleOCR team and ask them to help test it. Additionally, if the limitation issue is indeed due to the usage method of the model, how should we remind users in the document?

yimlin:
> **[megemini](/megemini) **

Do you have a compiled version available? We are looking for a llama.cpp that supports paddleocr-vl, as we will complete the document preprocessing.

megemini:
> > **[megemini](/megemini) **
> 
> Do you have a compiled version available? We are looking for a llama.cpp that supports paddleocr-vl, as we will complete the document preprocessing.

no, sorry ... ... you can try pull my branch and compile it ~

Remember2015:
In my opinion, it would be acceptable if documentation of this model clarifies that users need to run paddle layout analysis model first (maybe need some instructions), and then calling llama.cpp for OCR. The only concern is whether this PR supports the latest model, as I found that PaddleOCR-VL has already released v1.5 version

megemini:
> In my opinion, it would be acceptable if documentation of this model clarifies that users need to run paddle layout analysis model first (maybe need some instructions), and then calling llama.cpp for OCR. The only concern is whether this PR supports the latest model, as I found that PaddleOCR-VL has already released v1.5 version

try this https://modelscope.cn/models/megemini/PaddleOCR-VL-1.5-GGUF/summary

yimlin:
when this pull request will be merged into the master branch? is there any other questions?

Remember2015:
> > In my opinion, it would be acceptable if documentation of this model clarifies that users need to run paddle layout analysis model first (maybe need some instructions), and then calling llama.cpp for OCR. The only concern is whether this PR supports the latest model, as I found that PaddleOCR-VL has already released v1.5 version
> 
> try this https://modelscope.cn/models/megemini/PaddleOCR-VL-1.5-GGUF/summary

thank you , i'll try it~

zhang-prog:
@ngxson It has been verified that the accuracy aligns with the official figures provided by PaddleOCR:
PaddleOCRVL: 
92.80%(llama-server) vs 92.86%(vllm)
PaddleOCRVL1.5: 
94.42%(llama-server) vs 94.50%(vllm)

As mentioned by @megemini , PaddleOCR-VL employs a two-stage approach. For certain complex cases (such as the non-stop loop case mentioned above), we recommend referring to [PaddleOCR-VL](https://www.paddleocr.ai/latest/en/version3.x/pipeline_usage/PaddleOCR-VL.html) to utilize both the layout detection model and the VLM for document parsing. (Once this PR is merged, we will also add instructions for llama-server usage to the [PaddleOCR-VL](https://www.paddleocr.ai/latest/en/version3.x/pipeline_usage/PaddleOCR-VL.html))

ngxson:
Sounds good then, I'll attempt to merge this today or tmr.

@zhang-prog @megemini  If possible, please give me the link to the recommended llama.cpp guides, so that I can put it somewhere for visibility.

ngxson:
the chat template is broken, should be fixed via https://github.com/ggml-org/llama.cpp/pull/19419

IIIIIllllIIIIIlllll:
I encountered an error after building with the latest code. Could you please tell me how to resolve it? Thank you very much.

```
2026-02-08 16:20:05.648 - warmup: flash attention is enabled
2026-02-08 16:20:05.648 - srv    load_model: loaded multimodal model, '/home/mark/Models/Test/PaddleOCR-VL-1.5-BF16/mmproj.gguf'
2026-02-08 16:20:05.648 - srv    load_model: initializing slots, n_slots = 1
2026-02-08 16:20:05.660 - no implementations specified for speculative decoding
2026-02-08 16:20:05.660 - slot   load_model: id  0 | task -1 | speculative decoding context not initialized
2026-02-08 16:20:05.660 - slot   load_model: id  0 | task -1 | new slot, n_ctx = 32768
2026-02-08 16:20:05.660 - srv    load_model: prompt cache is enabled, size limit: 8192 MiB
2026-02-08 16:20:05.660 - srv    load_model: use `--cache-ram 0` to disable the prompt cache
2026-02-08 16:20:05.660 - srv    load_model: for more info see https://github.com/ggml-org/llama.cpp/pull/16391
2026-02-08 16:20:05.661 - srv          init: init: chat template parsing error: 
2026-02-08 16:20:05.661 - ------------
2026-02-08 16:20:05.661 - While executing For at line 34, column 13 in source:
2026-02-08 16:20:05.661 - ...age["role"] == "system" -%}↵        {%- for content in message["content"] -%}↵  ...
2026-02-08 16:20:05.661 -                                            ^
2026-02-08 16:20:05.661 - Error: Expected iterable or object type in for loop: got String
2026-02-08 16:20:05.661 - srv          init: init: please consider disabling jinja via --no-jinja, or use a custom chat template via --chat-template
2026-02-08 16:20:05.661 - srv          init: init: for example: --no-jinja --chat-template chatml
2026-02-08 16:20:05.661 - srv    operator(): operator(): cleaning up before exit...
2026-02-08 16:20:05.662 - main: exiting due to model loading error
```

megemini:
> I encountered an error after building with the latest code. Could you please tell me how to resolve it? Thank you very much.
> 

wait https://github.com/ggml-org/llama.cpp/pull/19419 merged, or use template below:

``` jinja

{%- if not add_generation_prompt is defined -%}
    {%- set add_generation_prompt = true -%}
{%- endif -%}
{%- if not cls_token is defined -%}
    {%- set cls_token = "<|begin_of_sentence|>" -%}
{%- endif -%}
{%- if not eos_token is defined -%}
    {%- set eos_token = "</s>" -%}
{%- endif -%}
{{- cls_token -}}
{%- for message in messages -%}
    {%- if message["role"] == "user" -%}
        {{- "User: " -}}

      {%- if message["content"] is string -%}
        {{- message["content"] }}
      {%- else -%}

        {%- for content in message["content"] -%}
            {%- if content["type"] == "image" -%}
                {{ "<|IMAGE_START|><|IMAGE_PLACEHOLDER|><|IMAGE_END|>" }}
            {%- endif -%}
        {%- endfor -%}
        {%- for content in message["content"] -%}
            {%- if content["type"] == "text" -%}
                {{ content["text"] }}
            {%- endif -%}
        {%- endfor -%}

      {%- endif -%}


        {{ "\n" -}}
    {%- elif message["role"] == "assistant" -%}
        {{- "Assistant: " -}}

      {%- if message["content"] is string -%}
        {{- message["content"] }}
      {%- else -%}

        {%- for content in message["content"] -%}
            {%- if content["type"] == "text" -%}
                {{ content["text"] }}
            {%- endif -%}
        {%- endfor -%}

      {%- endif -%}


        {{ eos_token -}}
    {%- elif message["role"] == "system" -%}

      {%- if message["content"] is string -%}
        {{- message["content"] }}
      {%- else -%}

        {%- for content in message["content"] -%}
            {%- if content["type"] == "text" -%}
                {{ content["text"] + "\n" }}
            {%- endif -%}
        {%- endfor -%}

      {%- endif -%}

    {%- endif -%}
{%- endfor -%}
{%- if add_generation_prompt -%}
    {{- "Assistant: " -}}
{%- endif -%}

```

with command:

``` shell

./build/bin/llama-cli -m /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-GGUF.gguf \
  --mmproj /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-GGUF-mmproj.gguf \
  --color on\
  --image /home/shun/Pictures/1640.jpeg \
  --prompt "OCR:" \
  --chat-template-file /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL/chat_template_llama.jinja

```

zhang-prog:
@ngxson Is there any documentation for specific models in llama.cpp? (like [PaddleOCR-VL vLLM recipes](https://docs.vllm.ai/projects/recipes/en/latest/PaddlePaddle/PaddleOCR-VL.html))? If there is, we’ll document how to use the model there; otherwise, we’ll put it in the [PaddleOCR-VL](https://www.paddleocr.ai/latest/en/version3.x/pipeline_usage/PaddleOCR-VL.html) docs.

zhang-prog:
@ngxson @megemini 
We have added documentation for `llama-server` to the [PaddleOCR-VL documentation](https://www.paddleocr.ai/latest/en/version3.x/pipeline_usage/PaddleOCR-VL.html#313-launch-service-directly-using-inference-acceleration-frameworks).

image

lilydjwg:
@zhang-prog I have a simple question: to parse an image from start to result, is CUDA still needed to run the whole process despite llama.cpp has a Vulkan backend?

ngxson:
The chat template fix has just been merged, I'll give it a try soon. Hopefully things work this time and I'll be able to merge this

ro99:
hello @zhang-prog @ngxson, I can see that PP-DocLayoutV3 is an intermediate step in the inference pipeline.

 Will PP-DocLayoutV3 work with this? Or it has to be executed separated (e.g. with transformers)?

ngxson (approved):

convert_hf_to_gguf.py
```diff
@@ -3739,6 +3746,10 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         if (head_dim := self.hparams.get("head_dim")) is None:
             head_dim = self.hparams["hidden_size"] // num_heads
 
+        if "mlp_AR" in name or "vision_model" in name:
+            # skip vision model and projector tensors
+            return []
```
```suggestion
            return
```

convert_hf_to_gguf.py
```diff
@@ -3848,6 +3859,49 @@ def prepare_tensors(self):
...
+
+        if "vision_model" in name or "mlp_AR" in name:
+            if "packing_position_embedding" in name:
+                return [] # unused
+            elif "vision_model.head" in name:
+                # we don't yet support image embeddings for this model
+                return []
+            else:
+                return [(self.map_tensor_name(name), data_torch)]
+        return [] # skip other tensors
```
```suggestion
        name = name.replace("visual.", "model.")

        if "vision_model" in name or "mlp_AR" in name:
            if "packing_position_embedding" in name:
                return # unused
            elif "vision_model.head" in name:
                # we don't yet support image embeddings for this model
                return
            else:
                yield from super().modify_tensors(data_torch, name, bid)
        return # skip other tensors
```

megemini:
> hello @zhang-prog @ngxson, I can see that PP-DocLayoutV3 is an intermediate step in the inference pipeline.
> 
> Will PP-DocLayoutV3 work with this? Or it has to be executed separated (e.g. with transformers)?

PP-DocLayoutV3 not included in llama.cpp

CISC (approved):

jiabochao:
> the result:
> 
> Screenshot from 2026-01-29 23-53-58
> can be formatted:
> 
> Screenshot from 2026-01-29 23-59-16

@megemini Hi, I have a question: how is the LLM Spotting format converted to this document? Are there any libraries or tools that can convert this format to DOCX or Markdown?

IIIIIllllIIIIIlllll:
I'm encountering this error when using the latest version (B8115): 
`tokenize: error: number of bitmaps (1) does not match number of markers (0). `
It seems that it only works with the previous chat template. Is this as expected?
the chat template:

```
{%- if not add_generation_prompt is defined -%}
    {%- set add_generation_prompt = true -%}
{%- endif -%}
{%- if not cls_token is defined -%}
    {%- set cls_token = "<|begin_of_sentence|>" -%}
{%- endif -%}
{%- if not eos_token is defined -%}
    {%- set eos_token = "</s>" -%}
{%- endif -%}
{{- cls_token -}}
{%- for message in messages -%}
    {%- if message["role"] == "user" -%}
        {{- "User: " -}}

      {%- if message["content"] is string -%}
        {{- message["content"] }}
      {%- else -%}

        {%- for content in message["content"] -%}
            {%- if content["type"] == "image" -%}
                {{ "<|IMAGE_START|><|IMAGE_PLACEHOLDER|><|IMAGE_END|>" }}
            {%- endif -%}
        {%- endfor -%}
        {%- for content in message["content"] -%}
            {%- if content["type"] == "text" -%}
                {{ content["text"] }}
            {%- endif -%}
        {%- endfor -%}

      {%- endif -%}


        {{ "\n" -}}
    {%- elif message["role"] == "assistant" -%}
        {{- "Assistant: " -}}

      {%- if message["content"] is string -%}
        {{- message["content"] }}
      {%- else -%}

        {%- for content in message["content"] -%}
            {%- if content["type"] == "text" -%}
                {{ content["text"] }}
            {%- endif -%}
        {%- endfor -%}

      {%- endif -%}


        {{ eos_token -}}
    {%- elif message["role"] == "system" -%}

      {%- if message["content"] is string -%}
        {{- message["content"] }}
      {%- else -%}

        {%- for content in message["content"] -%}
            {%- if content["type"] == "text" -%}
                {{ content["text"] + "\n" }}
            {%- endif -%}
        {%- endfor -%}

      {%- endif -%}

    {%- endif -%}
{%- endfor -%}
{%- if add_generation_prompt -%}
    {{- "Assistant: " -}}
{%- endif -%}
```

megemini:
> @megemini Hi, I have a question: how is the LLM Spotting format converted to this document? Are there any libraries or tools that can convert this format to DOCX or Markdown?

https://github.com/PaddlePaddle/PaddleOCR/blob/main/docs/version3.x/algorithm/PaddleOCR-VL/PaddleOCR-VL-1.5.en.md#core-features

`text spotting (text-line localization and recognition)`  with prompt `Spotting:` gives result in `text-line recognition`

image

with prompt `OCR:` gives result reorganized

image

I think this is the main difference.

As for the conversion of formats, can refer https://github.com/PaddlePaddle/PaddleX/blob/release/3.4/paddlex/inference/pipelines/paddleocr_vl/pipeline.py

megemini:
> I'm encountering this error when using the latest version (B8115): `tokenize: error: number of bitmaps (1) does not match number of markers (0). ` It seems that it only works with the previous chat template. Is this as expected? the chat template:

b8115 https://github.com/ggml-org/llama.cpp/releases/download/b8115/llama-b8115-bin-ubuntu-x64.tar.gz works fine without `--chat-template-file`

image

IIIIIllllIIIIIlllll:
> > I'm encountering this error when using the latest version (B8115): `tokenize: error: number of bitmaps (1) does not match number of markers (0). ` It seems that it only works with the previous chat template. Is this as expected? the chat template:
> 
> b8115 https://github.com/ggml-org/llama.cpp/releases/download/b8115/llama-b8115-bin-ubuntu-x64.tar.gz works fine without `--chat-template-file`
> 
> image

oh, sorry, I'm using llama-server. 
`C:\Users\Mark\App\llama-server-v0.4.2-b8112-windows-vulkan\llamacpp\win-vulkan\llama-server.exe -m C:\Users\Mark\Models\PaddleOCR-VL-1.5\PaddleOCR-VL-1.5.gguf --port 8083 --mmproj C:\Users\Mark\Models\PaddleOCR-VL-1.5\mmproj-PaddleOCR-VL-1.5.gguf --ctx-size 252144 --flash-attn on --no-mmap --temp 0 --top-p 0.95 --top-k 40 --min-p 0.05 --presence-penalty 0.0 --repeat-penalty 1.0 --frequency-penalty 0.0 --batch-size 2048 --ubatch-size 2048 `

megemini:
> > > I'm encountering this error when using the latest version (B8115): `tokenize: error: number of bitmaps (1) does not match number of markers (0). ` It seems that it only works with the previous chat template. Is this as expected? the chat template:
> > 
> > 
> > b8115 https://github.com/ggml-org/llama.cpp/releases/download/b8115/llama-b8115-bin-ubuntu-x64.tar.gz works fine without `--chat-template-file`
> > image
> 
> oh, sorry, I'm using llama-server. `C:\Users\Mark\App\llama-server-v0.4.2-b8112-windows-vulkan\llamacpp\win-vulkan\llama-server.exe -m C:\Users\Mark\Models\PaddleOCR-VL-1.5\PaddleOCR-VL-1.5.gguf --port 8083 --mmproj C:\Users\Mark\Models\PaddleOCR-VL-1.5\mmproj-PaddleOCR-VL-1.5.gguf --ctx-size 252144 --flash-attn on --no-mmap --temp 0 --top-p 0.95 --top-k 40 --min-p 0.05 --presence-penalty 0.0 --repeat-penalty 1.0 --frequency-penalty 0.0 --batch-size 2048 --ubatch-size 2048 `

The easiest way to fix this issue is add `<__media__>` before prompt, e.g. change `OCR:` to `<__media__>OCR:` when using `llama-server` , and do NOT use `--chat-template-file`, because there is a slight difference between PaddleOCR-VL and PaddleOCR-VL-1.5 templates.

@ngxson I found that, with `llama-server` command, the `prompt` of `body_parsed` after `oaicompat_chat_params_parse` should be something like `"<|begin_of_sentence|>User: <__media__>Formula Recognition:\nAssistant:\n"` other than `"<|begin_of_sentence|>User: Formula Recognition:\nAssistant:\n"`, which means `<__media__>` not been added properly. 

https://github.com/ggml-org/llama.cpp/pull/19419 changed `common_chat_msgs_to_json_oaicompat` to `render_message_to_json`, should `oaicompat_chat_params_parse` in `server-common` also be updated?

0xFS0CIETY:
why i got error in llama-server but cli works fine did i done something wrong

megemini:
> why i got error in llama-server but cli works fine did i done something wrong 

use `<__media__>OCR:` instead of `OCR:` , same issue https://github.com/ggml-org/llama.cpp/pull/18825#issuecomment-3934457275

0xFS0CIETY:
thanks its works 
image

IIIIIllllIIIIIlllll:
Thank you very much! However, as shown in the image, I'm unsure if my usage is correct.
I used llama-server, and the prompt was: '<__media__>Spotting:'.
In any case, the identified mathematical formulas all contained errors.

image
image

megemini:
> Thank you very much! However, as shown in the image, I'm unsure if my usage is correct. I used llama-server, and the prompt was: '<**media**>Spotting:'. In any case, the identified mathematical formulas all contained errors.
> 

I think the best way to process this test image is to first preprocess it using PP-DocLayoutV3, and then identify the text and formula parts separately. Otherwise, it should not be expected to obtain results that are 100% consistent with using PaddleOCR with PaddleOCR-VL model. Especially the formula in the middle of the picture.

p.s. @IIIIIllllIIIIIlllll @0xFS0CIETY the `<__media__>XXX:` is just a temporary solution, not a feature

congson1293:
@megemini I tried using `PaddleOCR-VL-1.5` from https://modelscope.cn/models/megemini/PaddleOCR-VL-1.5-GGUF with `llama-server`, but it always generated very short text result. I checked with `llama-cli`, and the result is correct.

megemini:
> @megemini I tried using `PaddleOCR-VL-1.5` from https://modelscope.cn/models/megemini/PaddleOCR-VL-1.5-GGUF with `llama-server`, but it always generated very short text result. I checked with `llama-cli`, and the result is correct.

try to use `--chat-template-file` with `llama-server`

```shell
./build/bin/llama-server -m /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-1.5-GGUF.gguf \
  --mmproj /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-1.5-GGUF-mmproj.gguf \
  --port 8111 --host 0.0.0.0 --ctx-size 131072 -n 4096 --temp 0 --jinja \
  --chat-template-file /media/shun/bigdata/Models/PaddleOCR_VL_SFT/PaddleOCR-VL-1.5/chat_template_llama.jinja
```
 
`chat_template_llama.jinja` uploaded https://modelscope.cn/models/megemini/PaddleOCR-VL-1.5-GGUF/files

congson1293:
Thank you @megemini. The problem was that I required structured output, which made the result too short. I removed the constraint and the result is now correct.

wangziren1:
@megemini Hello, I have a problem. When I use the following command to convert the model, it reports an error.
```
(llama.cpp) (base) zy@ps:~/data8tb/wangziren/llama.cpp$ python convert_hf_to_gguf.py models_paddleocr_vl/paddleocr_vl/ --mmproj  --outfile models_paddleocr_vl/PaddleOCR-VL-GGUF-mmproj_bf1
6.gguf   --outtype bf16   --verbose
INFO:hf-to-gguf:Loading model: paddleocr_vl
WARNING:hf-to-gguf:Failed to load model config from models_paddleocr_vl/paddleocr_vl: The repository models_paddleocr_vl/paddleocr_vl contains custom code which must be executed to correctly load the model. You can inspect the repository content at /home/zy/data8tb/wangziren/llama.cpp/models_paddleocr_vl/paddleocr_vl .
 You can inspect the repository content at https://hf.co/models_paddleocr_vl/paddleocr_vl.
Please pass the argument `trust_remote_code=True` to allow custom code to be run.
WARNING:hf-to-gguf:Trying to load config.json instead
INFO:hf-to-gguf:Model architecture: SiglipVisionModel
ERROR:hf-to-gguf:Model SiglipVisionModel is not supported
```
Without the --mmproj option, the model can be exported normally. With --mmproj, an error occurs.
I couldn't find the definition of SiglipVisionModel in convert_hf_to_gguf.py

megemini:
> I couldn't find the definition of SiglipVisionModel in convert_hf_to_gguf.py

Your model file and its configuration file are of an older version and need to be downloaded with the latest version.

wangziren1:
Thank you. I downloaded it from modelscope, which is different from the one on huggingface—the version on huggingface is the latest.

PR #18828, opened 2026-01-14
refactor : remove libcurl, use OpenSSL when available

angt:


.github/workflows/release.yml
```diff
@@ -358,7 +351,7 @@ jobs:
       - name: Build
         id: cmake_build
         run: |
-          cmake -S . -B build ${{ matrix.defines }} -DGGML_NATIVE=OFF -DGGML_CPU=OFF -DGGML_BACKEND_DL=ON -DLLAMA_CURL=OFF
+          cmake -S . -B build ${{ matrix.defines }} -DGGML_NATIVE=OFF -DGGML_CPU=OFF -DGGML_BACKEND_DL=ON -DLLAMA_OPENSSL=OFF
```
```suggestion
          cmake -S . -B build ${{ matrix.defines }} -DGGML_NATIVE=OFF -DGGML_CPU=OFF -DGGML_BACKEND_DL=ON -DLLAMA_BUILD_BORINGSSL=ON
```

.github/workflows/release.yml
```diff
@@ -412,7 +405,7 @@ jobs:
             -DGGML_NATIVE=OFF ^
             -DGGML_CPU=OFF ^
             -DGGML_CUDA=ON ^
-            -DLLAMA_CURL=OFF ^
+            -DLLAMA_OPENSSL=OFF ^
```
```suggestion
            -DLLAMA_BUILD_BORINGSSL=ON ^
```

.github/workflows/release.yml
```diff
@@ -481,7 +474,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=Release ^
             -DGGML_BACKEND_DL=ON -DBUILD_SHARED_LIBS=ON ^
             -DGGML_CPU=OFF -DGGML_SYCL=ON ^
-            -DLLAMA_CURL=OFF
+            -DLLAMA_OPENSSL=OFF
```
```suggestion
            -DLLAMA_BUILD_BORINGSSL=ON
```

.github/workflows/release.yml
```diff
@@ -608,7 +601,7 @@ jobs:
             -DAMDGPU_TARGETS="${{ matrix.gpu_targets }}" `
             -DGGML_HIP_ROCWMMA_FATTN=ON `
             -DGGML_HIP=ON `
-            -DLLAMA_CURL=OFF
+            -DLLAMA_OPENSSL=OFF
```
```suggestion
            -DLLAMA_BUILD_BORINGSSL=ON
```

CISC (approved):

.github/workflows/release.yml
```diff
@@ -649,7 +642,7 @@ jobs:
           cmake -B build -G Xcode \
             -DGGML_METAL_USE_BF16=ON \
             -DGGML_METAL_EMBED_LIBRARY=ON \
-            -DLLAMA_CURL=OFF \
+            -DLLAMA_BUILD_BORINGSSL=ON \
```
CISC: Mistake?

angt: absolutely, sorry

angt:
@CISC we can also enable boringssl in build-linux-cross.yml ?

CISC:
> @CISC we can also enable boringssl in build-linux-cross.yml ?

They all probably have working `libssl`, so if anything, just remove `-DLLAMA_OPENSSL=OFF` and install `libssl-dev`.

angt:
> > @CISC we can also enable boringssl in build-linux-cross.yml ?
> 
> They all probably have working `libssl`, so if anything, just remove `-DLLAMA_OPENSSL=OFF` and install `libssl-dev`.

That would be very surprising, they all do cross-building 👀

CISC:
> > > @CISC we can also enable boringssl in build-linux-cross.yml ?
> > 
> > 
> > They all probably have working `libssl`, so if anything, just remove `-DLLAMA_OPENSSL=OFF` and install `libssl-dev`.
> 
> That would be very surprising, they all do cross-building 👀

Why? There's cross-built dependencies. Anyway, doesn't matter, just leave them, might as well have test-cases for non-ssl build.

angt:
Let's go ?

cc @ggerganov @ngxson

CMakeLists.txt
```diff
@@ -111,11 +111,16 @@ option(LLAMA_BUILD_SERVER   "llama: build server example" ${LLAMA_STANDALONE})
 option(LLAMA_TOOLS_INSTALL  "llama: install tools"        ${LLAMA_TOOLS_INSTALL_DEFAULT})
 
 # 3rd party libs
-option(LLAMA_CURL       "llama: use libcurl to download model from an URL" ON)
 option(LLAMA_HTTPLIB    "llama: if libcurl is disabled, use httplib to download model from an URL" ON)
```
ngxson:
```suggestion
option(LLAMA_HTTPLIB    "llama: httplib for downloading functionality" ON)
```

not 100% sure what's the best wording, because we will use it for things other than just downloading in the future (for example, reversed proxy). But I think this should be enough for now

common/download.cpp
```diff
@@ -1151,7 +815,7 @@ int common_download_file_single(const std::string &,
     throw std::runtime_error("download functionality is not enabled in this build");
 }
 
-#endif // LLAMA_USE_CURL || LLAMA_USE_HTTPLIB
+#endif // defined(LLAMA_USE_HTTPLIB)
```
ngxson, line 818: small note here, but we should probably move the `#ifdef` inside each function, as it make things easier to follow.

(this should be a follow-up PR)

ngxson (approved):
this make the downloading code much cleaner now 🚀

ggerganov (approved):

timwalls:
This will be why my CI/CD built versions of `llama-server` started failing at runtime with "`https' scheme is not supported" then... :-)

For the next person scratching their head: `libssl-dev` is now a build dependency, without which the build won't fail but model downloads will stop working.

angt:
Thanks @timwalls.
I agree that the message provided by `cpp-httplib` is not helpful enough in our case.
I've opened a small PR to improve it.

PR #18838, opened 2026-01-14
examples : add test only when LLAMA_BUILD_TESTS=ON

angt:
Related to https://github.com/ggml-org/llama.cpp/pull/18791#issuecomment-3749365389.


examples/eval-callback/CMakeLists.txt
```diff
@@ -4,10 +4,12 @@ install(TARGETS ${TARGET} RUNTIME)
...
-else()
-    llama_download_model("tinyllamas/stories15M-be.Q4_0.gguf" SHA256=9aec857937849d976f30397e97eb1cabb53eb9dcb1ce4611ba8247fb5f44c65d)
+if(LLAMA_BUILD_TESTS)
+    set(TEST_TARGET test-eval-callback)
+    if(NOT ${CMAKE_SYSTEM_PROCESSOR} MATCHES "s390x")
+        llama_download_model("tinyllamas/stories15M-q4_0.gguf" SHA256=66967fbece6dbe97886593fdbb73589584927e29119ec31f08090732d1861739)
+    else()
+        llama_download_model("tinyllamas/stories15M-be.Q4_0.gguf" SHA256=9aec857937849d976f30397e97eb1cabb53eb9dcb1ce4611ba8247fb5f44c65d)
+    endif()
+    add_test(NAME ${TEST_TARGET} COMMAND llama-eval-callback -m "${LLAMA_DOWNLOAD_MODEL}" --prompt hello --seed 42 -ngl 0)
```
CISC, line 14: Hmmm, this won't do, I think you need to add download as a dependency to the test.

angt: I've tested like this:
```
$ rm -rf build && cmake -B build -DLLAMA_BUILD_TESTS=ON 2>/dev/null | grep tiny
-- Downloading tinyllamas/stories15M-q4_0.gguf from ggml-org/models...
```
and
```
$ rm -rf build && cmake -B build -DLLAMA_BUILD_TESTS=OFF 2>/dev/null | grep tiny
```
Did i miss something else ?

CISC: Yes, we don't want to download anything just because we're building the tests, the download is only needed when running the test.

FireBurn:
Thanks, this sorts things here

examples/eval-callback/CMakeLists.txt
angt: Do you know an easy way to achieve what you want ? I can create a `download.cmake` and add a test that call it but that would add lot of complexity to avoid the download when building tests.

For now the current PR fixes the reported issue.

CISC: Using `set_tests_properties` you can either add `DEPENDS` or `FIXTURES_REQUIRED` (with an accompanying `FIXTURES_SETUP`).

angt: That’s what I was thinking of, but that requires an external `.cmake` with a dedicated test. I’ll do a new PR.

PR #18841, opened 2026-01-14
Changing default values of mmap and direct_io to false in llama-bench

JTischbein:
When both are enabled by default the DirectIO path is used. By specifying mmap explicitly this behavior does not change. Therefore, with both parameters set to false as default, the DirectIO path and MMAP path can be enabled using the corresponding flag, without the need to disable the other. The default setting now uses `std::fread`. 

Before `--mmap 1 -dio 0` was needed to use mmap, now only `--mmap 1`.


JTischbein:
cc @0cc4m

0cc4m (approved):

PR #18842, opened 2026-01-14
kv-cache : optimize KQ mask construction

### LINKED ISSUE #18631: Eval bug: on long context, GPU utilization jumps seen on Mamba2 Hybrid models (Granite 4-h Small, Nemotron 30B A3B)
### Name and Version

$ llama-cli --version
  Device 0: NVIDIA A100-PCIE-40GB, compute capability 8.0, VMM: yes
version: 7635 (da143b994)

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

A100 40Gb PCIe

### Models

https://huggingface.co/ibm-granite/granite-4.0-h-small-GGUF?show_file_info=granite-4.0-h-small-Q6_K.gguf

https://huggingface.co/unsloth/Nemotron-3-Nano-30B-A3B-GGUF?show_file_info=Nemotron-3-Nano-30B-A3B-Q5_K_M.gguf

When I run either of the listed models with a context of 500k tokens, I get jumps of GPU utilization (observed with nvitop) to low values, eventually 0, during prefill. On the Granite they start after ~100k tokens. On the Nemotron they start earlier and become deeper - after ~250k tokens I see 0 most of the time with peak jumps up. 

Command lines are:

$ llama-server -m models/granite-4.0-h-small-Q6_K.gguf -c 510000 -ngl 99 

$ llama-server -m models/Nemotron-3-Nano-30B-A3B-Q5_K_M.gguf -c 510000 -ngl 99 

(I picked the highest quants that still fit a 510k KV cache in the 40Gb VRAM)

Eventual completion times (very short answers so it's basically just prefill:

Granite: 200k tokens took 241s, 500k tokens took 1196 s

Nemotron: 200k took 174s, 500k took 1016s

I would appreciate guidance about root-causing. #18599 is merged in the code I built.

### First Bad Commit

_No response_

### Relevant log output

no error messages

ggerganov:
fix #18631

Speed-up the loop for initializing the KQ mask array on the host. Take advantage of the fact that the 1D masks for different tokens of the same sequence are very similar, so we can minimize the checks by keeping track of the mask cells that can potentially change due to causal masking, SWA etc. based on their positions compared to the positions in the batch.

[verification](https://github.com/ggml-org/llama.cpp/commits/gg/kv-mask-opt-verify/)

### DIFF
--- a/src/llama-hparams.cpp
+++ b/src/llama-hparams.cpp
@@ -200,42 +200,6 @@ uint32_t llama_hparams::n_layer_kv() const {
     return res;
 }
 
-bool llama_hparams::is_masked_swa(uint32_t n_swa, llama_swa_type swa_type, llama_pos p0, llama_pos p1) {
-    assert(p0 >= 0 && p1 >= 0);
-
-    switch (swa_type) {
-        case LLAMA_SWA_TYPE_NONE:
-            {
-            } break;
-        case LLAMA_SWA_TYPE_STANDARD:
-            {
-                if (p1 - p0 >= (int32_t) n_swa) {
-                    return true;
-                }
-            } break;
-        case LLAMA_SWA_TYPE_CHUNKED:
-            {
-                const llama_pos pos_chunk_start = (p1 / n_swa) * n_swa;
-
-                if (p0 < pos_chunk_start) {
-                    return true;
-                }
-            } break;
-        case LLAMA_SWA_TYPE_SYMMETRIC:
-            {
-                const int32_t half_n_swa = (int32_t) n_swa / 2;
-                const int32_t pos_diff = p1 - p0;
-
-                // Mask if outside the symmetric window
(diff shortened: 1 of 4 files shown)


am17an:
This PR gives a ~7% perf boost on PP in CUDA as well. Nice!

ggerganov:
Hm, it should make a difference only at very large contexts, e.g. somewhere beyond 64k. In which case did you observe an improvement?

am17an:
Strange, it gives me a PP boost at empty context and 2048 batch size. 

Edit: Tested on gpt-oss-20b on 5090 with fa on.

lexasub:
my experiments with boosting set_input_kq_mask_impl https://github.com/lexasub/llama.cpp/tree/d-flash-prefil (last commits rewrite set_input_kq_mask_impl) - in work, stuck in std::copy - need more optimal algo without copy (via fill on example, but my implementations with fill now have smaller perfomance that std::copy)

PR #18844, opened 2026-01-14
fit-params : print signed int for -ngl param

ggerganov:
```bash
# before
$ llama-fit-params -m model.gguf
-c 0 -ngl 4294967295

# after
$ llama-fit-params -m model.gguf
-c 0 -ngl -1
```


ggerganov:
Hm, although not sure if it is actually the intended thing to print it as unsigned int?

PR #18858, opened 2026-01-15
ggml-cpu: add q4_0 repack support for wasm

aviallon:
Add LLM written WASM simd128 implementations for `ggml_quantize_mat_q8_0_4x4`, `ggml_quantize_mat_q8_0_4x8` and `ggml_gemv_q4_0_4x4_q8_0`, `ggml_gemm_q4_0_4x4_q8_0`, `ggml_gemv_q4_0_8x8_q8_0` and `ggml_gemm_q4_0_8x8_q8_0`.
Tested from a custom Wllama.

### DIFF
--- a/ggml/src/ggml-cpu/CMakeLists.txt
+++ b/ggml/src/ggml-cpu/CMakeLists.txt
@@ -543,7 +543,10 @@ function(ggml_add_cpu_backend_variant_impl tag_name)
         ggml_add_cpu_backend_features(${GGML_CPU_NAME} s390 ${ARCH_DEFINITIONS})
     elseif (CMAKE_SYSTEM_PROCESSOR MATCHES "wasm")
         message(STATUS "Wasm detected")
-        list (APPEND GGML_CPU_SOURCES ggml-cpu/arch/wasm/quants.c)
+        list (APPEND GGML_CPU_SOURCES
+            ggml-cpu/arch/wasm/quants.c
+            ggml-cpu/arch/wasm/repack.cpp
+        )
     else()
         message(WARNING "Unknown CPU architecture. Falling back to generic implementations.")
         list(APPEND ARCH_FLAGS -DGGML_CPU_GENERIC)
--- a/ggml/src/ggml-cpu/arch/wasm/repack.cpp
+++ b/ggml/src/ggml-cpu/arch/wasm/repack.cpp
@@ -1,0 +1,631 @@
+#define GGML_COMMON_IMPL_CPP
+#define GGML_COMMON_DECL_CPP
+#include "ggml-common.h"
+#include "ggml-backend-impl.h"
+
+#include "ggml-impl.h"
+#include "ggml-cpu.h"
+#include "ggml-cpu-impl.h"
+#include "simd-mappings.h"
+#include "traits.h"
+
+#include <cmath>
+#include <cstring>
+#include <cassert>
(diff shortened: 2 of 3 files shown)


aviallon:
@ngxson this may be of interest to you. Note: it is required to modify wllama's build slightly to use this, as emscripten overrides `CMAKE_SYSTEM_PROCESSOR`.
Needed options:
```cmake
set(GGML_CPU_REPACK ON CACHE BOOL "enable ggml CPU repack optimizations")
set(LLAMA_WASM_MEM64 OFF CACHE BOOL "disable MEMORY64 for wllama")
```
+ `emcmake cmake -DEMSCRIPTEN_SYSTEM_PROCESSOR=wasm …`

ngxson:
If I understand correctly, this requires MEM64 to be disabled, right?

CC @reeselevine for visibility

reeselevine:
Interesting, looks like the change in https://github.com/ngxson/wllama/pull/201/commits/492c4233bd7af5400eb24c2fe036a0c8d306c392 to move the flags to CMakeLists.txt is what caused the 64-bit build to be enabled by default, since `LLAMA_WASM_MEM64` is on by default in llama.cpp right now. So the cache directive is needed to override it, unless it's specified on the command line. 

I don't think the simd implementations in this PR would require disabling 64-bit generally, right? It's just that right now, wllama doesn't support the 64-bit builds yet.

reeselevine:
Or actually, I realize https://github.com/ngxson/wllama/pull/200 doesn't include that flag, because the WebGPU integration PR hasn't been merged yet. Maybe the cache directive actually isn't needed?

aviallon:
> Interesting, looks like the change in https://github.com/ngxson/wllama/pull/201/commits/492c4233bd7af5400eb24c2fe036a0c8d306c392 to move the flags to CMakeLists.txt is what caused the 64-bit build to be enabled by default, since `LLAMA_WASM_MEM64` is on by default in llama.cpp right now. So the cache directive is needed to override it, unless it's specified on the command line. 
> 
> I don't think the simd implementations in this PR would require disabling 64-bit generally, right? It's just that right now, wllama doesn't support the 64-bit builds yet.

I actually tried building with 64-bit support enabled, and got errors even when running with node.js directly.

aviallon:
For the record, with that + the llama.cpp version bump and `-ffast-math -fno-finite-math-only`, I get ~50% faster PP compared to current wllama.

aviallon:
@reeselevine I'm planning to do a pass on this some time in the future. What do you think should be improved?
I think I'll make builds with WebAssembly + WebGPU support in the future, but that means I'll need a fast-enough CPU fallback for my tiny models to work (hello Qwen3.5-0.8B !)

reeselevine:
I don't have any experience working with WASM SIMD, so not sure I can give any recommendations on those implementations. It'd be great to have fast CPU-fallback for devices/cases where WebGPU doesn't work well, but also just want to note that Qwen3.5 should be supported by the WebGPU backend now.

aviallon:
> I don't have any experience working with WASM SIMD, so not sure I can give any recommendations on those implementations. It'd be great to have fast CPU-fallback for devices/cases where WebGPU doesn't work well, but also just want to note that Qwen3.5 should be supported by the WebGPU backend now.

Yes, the issue is not WebGPU support by llama.cpp, but by the target browsers…

PR #8526, opened 2024-07-17
llama : simplify Mamba with advanced batch splits

compilade:
As promised in <https://github.com/ggerganov/llama.cpp/pull/7531#issuecomment-2209585346>, I've been extracting the advanced batch splits out of the Jamba PR (#7531).

I've also backported the contiguous allocation of recurrent state slots, which makes it possible to *also* include the changes from #7531 which simplify the `ggml` operators used specifically for Mamba. Hopefully this isn't too much at once.

See <https://github.com/ggerganov/llama.cpp/pull/7531#discussion_r1620997020> for an explanation of the batch splits.

# Summary

- `ggml.c`
  - Simplify `ggml_ssm_conv` and `ggml_ssm_scan` by assuming batched sequences have the same number of new tokens, and that the states are contiguous and ordered correctly.
  - Allow `ggml_concat` to work with a non-contiguous second argument.
    - The CPU implementation already supported this, but it was guarded with an assertion. Meanwhile, I *think* the CUDA implementation already supports this too, and does *not* prevent its usage (not totally sure), so I did not change it.
- `llama.cpp`
  - Advanced batch splits handled with `lctx.sbatch` for persistent buffers
    - Refactor "helpers for smoother batch API transition", by handling them in `llama_sbatch`, which allows avoiding repeated allocations by re-using the same buffers.
    - Simple batch splits should be equivalent to the previous behavior and are made with `lctx.sbatch.split_simple(n_tokens)` to build a `llama_ubatch` with a max size of `n_tokens`.
    - Equal-sequence-lengths splits are made with `lctx.sbatch.split_equal(n_tokens)`, and are used to simplify the operators of recurrent models.
    - Add `llama_ubatch`. Similar to `llama_batch`, but aware of equal-length sequences.
      - Make `llama_set_inputs` (and others) use `llama_ubatch` instead of `llama_batch`.
  - Make recurrent state slot allocation contiguous in `llama_kv_cache_find_slot`
  - Add `llm_build_mamba` to build a Mamba block, used for Mamba, and will be used for Jamba
  - Add `llm_build_copy_mask_state` (maybe not a good name) to abstract away the shuffling and masking of recurrent states. Used for Mamba, and it should be usable for other recurrent architectures too.
  - Simplify the sanity checks for `qs.n_attention_wv` in `llama_model_quantize_internal` to make it future proof for hybrid models.
[...]

PR #18862, opened 2026-01-15
llama : remove write/read of output ids/logits/embeddings

danbev:
This commit removes the write/read of output ids, logits and embeddings from the llama context state.

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -1805,14 +1805,13 @@ bool common_prompt_batch_decode(
                                int & n_past,
                                int   n_batch,
             const std::string_view & state_path,
-                              bool   save_state,
-                              bool   is_last_batch) {
+                              bool   save_state) {
     const int n_eval = tokens.size();
     if (n_eval == 0) {
         return true;
     }
 
-    if (save_state && is_last_batch && n_eval > 1) {
+    if (save_state && n_eval > 1) {
         const int n_tokens_before_last = n_eval - 1;
 
         GGML_ASSERT(n_eval <= n_batch);
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -1788,3 +1788,56 @@ float lr_opt::get_lr(float epoch) const {
     LOG_INF("epoch %.2g lr=%.2g\n", epoch, r);
     return r;
 }
+
+bool common_prompt_batch_decode(
+              struct llama_context * ctx,
+    const std::vector<llama_token> & tokens,
+                               int & n_past,
+                               int   n_batch,
+       const std::filesystem::path & state_path,
+                              bool   save_state,
+                              bool   is_last_batch) {
+    const int n_eval = tokens.size();
+    if (n_eval == 0) {
+        return true;
+    }
+
+    if (save_state && is_last_batch && n_eval > 1) {
+        const int n_tokens_before_last = n_eval - 1;
+
+        GGML_ASSERT(n_eval <= n_batch);
+
+        // Decode all but the last token so we can save the memory state before decoding the last token.
+        // This is done so we can restore the session state later and replay the last token.
+        // Memory implementations in recurrent/hybrid models don't support removing tokens from their
+        // memory, so we can't just remove the last token from the memory and replay the last token which
+        // is the reason for this logic.
+        if (llama_decode(ctx, llama_batch_get_one(const_cast<llama_token*>(tokens.data()), n_tokens_before_last))) {
+            LOG_ERR("%s : failed to eval\n", __func__);
+            return false;
+        }
+        n_past += n_tokens_before_last;
+
+        llama_state_save_file(ctx, state_path.string().c_str(), tokens.data(), n_tokens_before_last);
+        LOG_INF("saved session before last token to %s, n_tokens = %d\n", state_path.string().c_str(), n_tokens_before_last);
+
+        llama_token last_token = tokens.back();
+        llama_batch batch = llama_batch_get_one(&last_token, 1);
+        int32_t pos = n_past;
+        batch.pos = &pos;
+
+        if (llama_decode(ctx, batch)) {
+            LOG_ERR("%s : failed to eval last token\n", __func__);
+            return false;
+        }
+        n_past++;
+    } else {
+        if (llama_decode(ctx, llama_batch_get_one(const_cast<llama_token*>(tokens.data()), n_eval))) {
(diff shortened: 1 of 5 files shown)


ggerganov:
Initially, I was thinking that since the samplers can now be part of the context state, we should also store this information. But it should include also the sampler states. And it gets very complicated.

But now I am wondering if we should instead remove the output ids, the logits and the embeddings from the state and only store the model info and the memory. I can't think of a meaningful use case for storing this information. And even if its needed, one can simply run the last token through `llama_decode` to obtain the necessary logits/embeddings. So maybe this is the better option as it will simplify the read/write logic.

danbev:
> And even if its needed, one can simply run the last token through `llama_decode` to obtain the necessary logits/embeddings.

I've updated [llama-completion](https://github.com/ggml-org/llama.cpp/pull/18862/commits/1d241c517858cab604e3a09938e1c6126ffd93e5) to "replay" the last token. Could you take a look and this if this is correctly handled for recurrent/hybrid models?

tools/completion/completion.cpp
```diff
@@ -387,6 +387,32 @@ int main(int argc, char ** argv) {
...
+        // Logits are not stored as part of the session state so we need to
+        // "replay" the last token to get logits for sampling.
+        if (!session_tokens.empty() && n_match > 0 && n_match == session_tokens.size()) {
+            llama_token last_token = session_tokens.back();
+            int32_t pos;
+
+            if (llama_model_is_recurrent(model) || llama_model_is_hybrid(model)) {
+                LOG_INF("%s: recurrent/hybrid model: decode using next position: %d\n", __func__, (int)n_match);
+                pos = n_match; // use next position for decoding
+            } else {
```
ggerganov: Hm this does not look right. It will evaluate token `n_match - 1` with position `n_match`.

danbev: For recurrent models their hidden state is cumulative so we can't undo the last tokens effect on the state.
Would it perhaps be alright to store the logits just for recurrent/hybrid models? 
Or do you have any other alternatives for this that I might not have thought of?

ggerganov: The solution here is to save the session at least one token before the full prompt is processed. We already do this in server:

https://github.com/ggml-org/llama.cpp/blob/eef375ce167cf36d8ed2cf347be95a7a7a90fb5e/tools/server/server-context.cpp#L2522-L2526

I looked at some point how to make a similar logic in `completion.cpp`, but seemed a bit complicated. Likely need to do some refactoring to simplify the logic first.

Ideally, we have to find a common pattern that is easy to apply - for example the `save-load-state.cpp` would need similar logic.

> Would it perhaps be alright to store the logits just for recurrent/hybrid models?

The few branches in the logic is always better, so prefer to avoid that.

src/llama-context.cpp
```diff
@@ -2500,64 +2500,6 @@ size_t llama_context::state_write_data(llama_io_write_i & io) {
...
-        const auto & output_ids = this->output_ids;
-
-        std::vector<int32_t> w_output_pos;
-
-        w_output_pos.resize(n_outputs);
-
-        // build a more compact representation of the output ids
-        for (size_t i = 0; i < n_batch(); ++i) {
-            // map an output id to a position in the batch
-            int64_t pos = output_ids[i];
```
compilade, line 2499: The main reason for output ids in the first place was when using `llama-perplexity` with e.g. the hellaswag benchmark, which uses parallel sequences of different lengths, which are re-ordered by the batch splitting logic for recurrent models to get equal-length sequences within ubatches.

(The initial reason was to make the output buffer smaller (see <https://github.com/ggml-org/llama.cpp/pull/6122>), but it allowed tackling batch splitting afterwards (https://github.com/ggml-org/llama.cpp/pull/8526))

I'm not sure whether simply replaying the last token will be sufficient in the case where more logits than that are required.

compilade: Whoops nevermind, this concern doesn't apply here, session saving doesn't affect things like `llama-perplexity`, I think.

danbev: I'll take a look just to make sure as I not familiar how llama-perplexity stores its information but I was under the impression that this format is different and not related.

common/common.cpp
```diff
@@ -1863,3 +1863,66 @@ float lr_opt::get_lr(float epoch) const {
...
+}
+
+bool common_prompt_batch_decode(
+              struct llama_context * ctx,
+    const std::vector<llama_token> & tokens,
+                               int & n_past,
+                               int   n_batch,
+       const std::filesystem::path & state_path,
+                              bool   save_state,
+                              bool   is_last_batch) {
```
ggerganov: The API can be simpler:

```suggestion
bool common_prompt_batch_decode(
              struct llama_context * ctx,
    const std::vector<llama_token> & tokens,
                               int & n_past,
                               int   n_batch,
                 const std::string & state_fname) {
    const bool save_state = !state.fname.empty();
    ...
```

Don't think it's worth bringing `std::filesystem` here, so simple `std::string` should be fine.

common/common.h
```diff
@@ -779,6 +779,23 @@ void common_batch_add(
...
+//
+// Note: We save state before the last token so that we can replay it to ensure
+// compatibility with all memory types. Recurrent/hybrid models cannot remove
+// tokens from memory, so this approach works across all model architectures.
+bool common_prompt_batch_decode(
+              struct llama_context * ctx,
+    const std::vector<llama_token> & embd,
+                               int & n_past,
+                               int   n_batch,
+            const std::string_view & state_path,
```
ggerganov: I think the preferred way in C++ usually is to pass `std::string_view` by value:


```suggestion
                  std::string_view   state_path,
```

ggerganov (approved):
I think the change is OK. Just double check that `save-load-state` works correctly with few different model types.

We should think how to exercise this in the CI. If we are able to generate small test models of different architectures as Johannes mentioned in https://github.com/ggml-org/llama.cpp/pull/19378#issuecomment-3899326129, we can add extra `save-load-state` and `server` tests to the CI in order to cover recurrent and hybrid arches.

danbev:
@ggerganov I've tested `save-load-state` with recurrent, hybrid, and standard transformer models.  

I'll take a look at https://github.com/ggml-org/llama.cpp/pull/19802 and follow up regarding adding tests for this to CI.

fairydreaming:
@danbev if you have some time please take a look at #23400

PR #18870, opened 2026-01-15
feat: Add file descriptor based model loading for Android SAF support

Siddhesh2377:
Summary

  This PR adds native file descriptor (FD) based model loading to llama.cpp, enabling Android applications to load GGUF models without requiring the MANAGE_EXTERNAL_STORAGE permission.

  Motivation

  On Android 11+, Google heavily restricts the MANAGE_EXTERNAL_STORAGE permission - it's only granted to file managers and antivirus apps. Regular apps must use the Storage Access Framework (SAF) to access files outside their sandbox.

  SAF provides access to files via content:// URIs and file descriptors, not file paths. The current llama.cpp API only accepts file paths, which:
  - Doesn't work with SAF URIs
  - The /proc/self/fd/N symlink trick fails because Android blocks open() on these paths even though stat()/fstat() work

  Solution

  Add a new public API that accepts a file descriptor directly:

  LLAMA_API struct llama_model * llama_model_load_from_fd(
      int fd,
      size_t file_size, 
      struct llama_model_params params);

  The implementation:
  1. Uses fdopen() to convert FD to FILE* for GGUF parsing
  2. Uses dup() so the caller retains ownership of their FD
  3. Reads model data via standard file I/O (no mmap support for FD loading)

 Limitations

  - No mmap support (FD loading reads into memory)
  - No split model support
  - Windows not supported (returns nullptr with error log)

  Usage Example (Android JNI)
```cpp
  // Get FD from Android's ContentResolver via JNI
  int fd = env->CallIntMethod(parcelFileDescriptor, detachFdMethod);

  llama_model_params params = llama_model_default_params();
  params.use_mmap = false;  // Required for FD loading

  llama_model* model = llama_model_load_from_fd(fd, file_size, params);
```
  Testing

  Tested on Android 14 with SAF-provided file descriptors from Google Drive and local storage


src/llama-model-loader.cpp
```diff
@@ -367,13 +366,7 @@ namespace GGUFMeta {
                 result[i] = value;
             }
         } else {
-            if (arr_info.gt == GGUF_TYPE_BOOL) {
-                std::transform((const bool *)arr_info.data, (const bool *)arr_info.data + arr_info.length, result.begin(), [](bool x) {
-                    return static_cast<T>(x);
-                });
-            } else {
-                std::copy((const T*)arr_info.data, (const T *)arr_info.data + arr_info.length, result.begin());
-            }
```
ggerganov, line 376: This was accidentally reverted?

ggerganov (commented):
The `llama.cpp` and `llama-model-loader.cpp` changes should be possible to deduplicate significantly.

src/llama-model-loader.cpp
Siddhesh2377: I will have a look into it, and let you know

Siddhesh2377:
@ggerganov  Check now

src/llama-model-loader.cpp
```diff
@@ -492,6 +501,133 @@ namespace GGUFMeta {
...
+    }
+    tensor_buft_overrides = param_tensor_buft_overrides_p;
+}
+
+void llama_model_loader::load_architecture_and_metadata() {
+    get_key(llm_kv(LLM_KV_GENERAL_ARCHITECTURE), arch_name, false);
+    llm_kv = LLM_KV(llm_arch_from_string(arch_name));
+}
+
+void llama_model_loader::determine_file_type_and_dump_metadata(int trace, const char * source_name) {
```
ggerganov, line 522: This is not used anywhere.

Did you even looked at your changes, or just expect me to fix everything for you?

Siddhesh2377: Hey @ggerganov 
I am Sorry for the Mess, Actully I don't have a good understanding Of C++, The Feature is very Great for Android Developers, thus for making it quick i coded it with Claude AI
That Said, i will Fix it, and Try to Raise A Clean PR for you to include it, Hope You Understand 
Thanks : )

PR #18880, opened 2026-01-16
ggml webgpu: support for backend sampling

reeselevine:
The refactor to backend-sampler tests in https://github.com/ggml-org/llama.cpp/pull/18753 broke WebGPU CI, because we hadn't implemented necessary operations yet and something about that refactor caused the tests to automatically run and fail on whatever backend is available. I suppose I could have just modified the tests to not run on the WebGPU backend, but instead I implemented all the operations necessary to run the sampler tests via WebGPU.

So, this PR adds the following:
* Support for CLAMP, LOG, FILL, PAD, ARGMAX, ARGSORT, TOP_K, CUMSUM, SUM, SUM_ROWS, as well as a few unary operators that weren't needed for sampling, but which were written by @abhijitramesh and which I included as part of my unary ops refactor to JIT compilation while supporting CLAMP.
*  Support for I32 indices for SET_ROWS and I32 variants of CPY.
* Implements compilation of new operations in a JIT manner, so we only compile variants of shaders that are actually used by the models. WebGPU compilation always happens at runtime because it needs to know what platform it's running on, but many of the existing shaders are compiled at startup, which reduces flexibility and increases resource usage for no good reason (in my opinion). We're moving to this new approach generally.
* Update ops support


CISC (approved):
Awesome!

The CI failure is just corrupt `ccache`, I have deleted the cache and will rerun it once CIs are done.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -3069,6 +3071,9 @@ static bool ggml_backend_webgpu_device_supports_op(ggml_backend_dev_t dev, const
         case GGML_OP_FILL:
             supports_op = op->type == GGML_TYPE_F32 && src0->type == GGML_TYPE_F32;
             break;
+        case GGML_OP_LOG:
```
CISC, line 3074: Update the `ops` docs.

CISC:
`LOG` fails:
```console
2026-01-16T21:14:19.3151243Z 35: Failing tests:
2026-01-16T21:14:19.3154644Z 35:   LOG(type=f16,ne=[10,5,4,3])
2026-01-16T21:14:19.3158114Z 35:   LOG(type=f16,ne=[7,1,5,3])
2026-01-16T21:14:19.3161950Z 35:   Backend WebGPU: WebGPU: [1;31mFAIL[0m
```

reeselevine:
> `LOG` fails:
> 
> ```
> 2026-01-16T21:14:19.3151243Z 35: Failing tests:
> 2026-01-16T21:14:19.3154644Z 35:   LOG(type=f16,ne=[10,5,4,3])
> 2026-01-16T21:14:19.3158114Z 35:   LOG(type=f16,ne=[7,1,5,3])
> 2026-01-16T21:14:19.3161950Z 35:   Backend WebGPU: WebGPU: �[1;31mFAIL�[0m
> ```

looks like it's numerical issues, let me try casting to f32 for the actual operation, shouldn't affect performance much if at all

jeffbolznv:
> looks like it's numerical issues, let me try casting to f32 for the actual operation, shouldn't affect performance much if at all

Could be a rounding issue on the result - for ggml-vulkan we've had to force RTNE to make a lot of these f16 tests pass.

reeselevine:
> > looks like it's numerical issues, let me try casting to f32 for the actual operation, shouldn't affect performance much if at all
> 
> Could be a rounding issue on the result - for ggml-vulkan we've had to force RTNE to make a lot of these f16 tests pass.

Ah interesting, if this doesn't fix it I'll look into that. I don't think WebGPU supports the same features though.

reeselevine:
Looks like the higher precision worked!

PR #18883, opened 2026-01-16
jinja : attribute support for join, map and sort

CISC:
Adds attribute support (`int` for arrays and `str` for objects), for `join`, `map` and `sort` filters.

Adds support for negative array index and default in `value->at()`.

### DIFF
--- a/common/jinja/value.cpp
+++ b/common/jinja/value.cpp
@@ -775,19 +775,30 @@ const func_builtins & value_array_t::get_builtins() const {
             if (!is_val<value_array>(args.get_pos(0))) {
                 throw raised_exception("join() first argument must be an array");
             }
-            value val_delim     = args.get_kwarg_or_pos("d",         1);
-            value val_attribute = args.get_kwarg_or_pos("attribute", 2);
-            if (!val_attribute->is_undefined()) {
-                throw not_implemented_exception("array attribute join not implemented");
-            }
+            value val_delim = args.get_kwarg_or_pos("d",         1);
+            value attribute = args.get_kwarg_or_pos("attribute", 2);
             const auto & arr = args.get_pos(0)->as_array();
-            std::string delim = is_val<value_string>(val_delim) ? val_delim->as_string().str() : "";
+            const bool attr_is_int = is_val<value_int>(attribute);
+            if (!attribute->is_undefined() && !is_val<value_string>(attribute) && !attr_is_int) {
+                throw raised_exception("join() attribute must be string or integer");
+            }
+            const int64_t attr_int = attr_is_int ? attribute->as_int() : 0;
+            const std::string delim = val_delim->is_undefined() ? "" : val_delim->as_string().str();
+            const std::string attr_name = attribute->is_undefined() ? "" : attribute->as_string().str();
             std::string result;
             for (size_t i = 0; i < arr.size(); ++i) {
-                if (!is_val<value_string>(arr[i]) && !is_val<value_int>(arr[i]) && !is_val<value_float>(arr[i])) {
+                value val_arr = arr[i];
+                if (!attribute->is_undefined()) {
+                    if (attr_is_int && is_val<value_array>(val_arr)) {
+                        val_arr = val_arr->at(attr_int);
+                    } else if (!attr_is_int && !attr_name.empty() && is_val<value_object>(val_arr)) {
+                        val_arr = val_arr->at(attr_name);
+                    }
+                }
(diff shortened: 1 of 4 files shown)


common/jinja/value.cpp
```diff
@@ -775,19 +775,30 @@ const func_builtins & value_array_t::get_builtins() const {
             if (!is_val<value_array>(args.get_pos(0))) {
                 throw raised_exception("join() first argument must be an array");
             }
-            value val_delim     = args.get_kwarg_or_pos("d",         1);
-            value val_attribute = args.get_kwarg_or_pos("attribute", 2);
-            if (!val_attribute->is_undefined()) {
-                throw not_implemented_exception("array attribute join not implemented");
-            }
+            value val_delim = args.get_kwarg_or_pos("d",         1);
+            value attribute = args.get_kwarg_or_pos("attribute", 2);
```
ngxson, line 779: both map, sort, join can have `attribute`, probably we can find a way to abstract out the `attribute` handling?

common/jinja/value.cpp
```diff
@@ -847,29 +862,35 @@ const func_builtins & value_array_t::get_builtins() const {
...
                 throw raised_exception("sort: first argument must be an array");
             }
-            bool reverse = args.get_kwarg("reverse", mk_val<value_undefined>())->as_bool();
-            value attribute = args.get_kwarg("attribute", mk_val<value_undefined>());
-            std::string attr = attribute->is_undefined() ? "" : attribute->as_string().str();
+            value val_reverse = args.get_kwarg_or_pos("reverse",        1);
+            value val_case    = args.get_kwarg_or_pos("case_sensitive", 2);
+            value attribute   = args.get_kwarg_or_pos("attribute",      3);
+            // FIXME: sorting is case sensitive
+            //const bool case_sensitive = val_case->as_bool(); // undefined == false
```
ngxson, line 873: maybe throw `not_implemented_exception` if `case_sensitive` is set?

common/jinja/value.h
```diff
@@ -168,8 +168,20 @@ struct value_t {
...
+    virtual value & at(int64_t index, value & default_val) {
+        if (index < 0) {
+            index += val_arr.size();
+        }
+        if (index < 0 || static_cast<size_t>(index) >= val_arr.size()) {
+            return default_val;
+        }
+        return val_arr[index];
+    }
+    virtual value & at(int64_t index) {
```
ngxson, line 180: I'm not quite sure why I placed the `at(...)` inside `value_t` in the first place, probably better to place them inside `value_array_t` and `value_object_t` respectively - let's have a look in a follow-up PR

common/jinja/value.cpp
```diff
@@ -847,29 +862,35 @@ const func_builtins & value_array_t::get_builtins() const {
...
-                    if (!is_val<value_object>(a) || !is_val<value_object>(b)) {
-                        throw raised_exception("sort: items are not objects");
+                    if (attr_is_int && is_val<value_array>(a) && is_val<value_array>(b)) {
+                        val_a = a->at(attr_int);
+                        val_b = b->at(attr_int);
+                    } else if (!attr_is_int && !attr_name.empty() && is_val<value_object>(a) && is_val<value_object>(b)) {
+                        val_a = a->at(attr_name);
+                        val_b = b->at(attr_name);
+                    } else {
+                        throw raised_exception("sort: unsupported object attribute comparison");
```
ngxson, line 890: maybe also log the type of a and b in the message, so that it's easier to debug in the future

ngxson (approved):

common/jinja/value.cpp
CISC: It's the other way around, `case_sensitive=False` is default, but we are always case sensitive ATM.

common/jinja/value.cpp
CISC: I didn't want to add `std::format` here since it's not used anywhere else.

common/jinja/value.cpp
CISC: Maybe, but didn't immediately see how since they are applied differently.

CISC:
Merging now, let's address open suggestions if applicable in follow-ups.

common/jinja/value.cpp
ngxson: I mean you can just use string concat `+`, we already doing that in multiple places. Not the best, but it makes debugging much easier.

CISC: I think I spaced out a little, thinking I needed to format values or something. :)

I'll fix it in the next PR which will touch this code anyway.

PR #18039, opened 2025-12-14
[Speculative decoding] feat: add EAGLE3 speculative decoding support

ruixiang63:
> [!IMPORTANT]
> The old PR has been backed up in this branch: https://github.com/ruixiang63/llama.cpp/tree/eagle3-v1-backup
The new commits in this PR have been rebased onto the latest master branch, refactored to use the new speculative API, cherry-picked from https://github.com/ggml-org/llama.cpp/pull/22728, and made compatible with MTP.

> [!TIP]
> New Eagle3 models for Gemma4 are now supported. With reasoning enabled, speedup can exceed 2x. With reasoning disabled, it can reach over 3x. See https://github.com/ggml-org/llama.cpp/pull/18039#issuecomment-4500827273
With `Q4_K_M` quantization, the speedup still looks good. https://github.com/ggml-org/llama.cpp/pull/18039#issuecomment-4507174617

As discussed in https://github.com/ggml-org/llama.cpp/discussions/15902, Eagle3 represents the current SOTA in speculative decoding and is widely adopted across the industry. Integrating Eagle3 into llama.cpp enhances its performance and strengthens its competitiveness among leading inference frameworks. With Eagle3 speculative decoding now integrated into llama.cpp, inference performance has been significantly improved, achieving a 2–3× speedup.
This enhancement is the result of close collaboration between the NVIDIA and GGML teams, showcasing a strong technical partnership. 

The following provides a brief overview of this PR:

EAGLE3 is an encoder-decoder based speculative decoding method:
- Extracts features from target model at specific layers
- Uses feature fusion layer to compress target features
- Generates draft tokens with single-layer decoder
- Maps draft vocabulary to target vocabulary via d2t tensor

Key changes:
- Add LLM_ARCH_EAGLE3 architecture
- Add EAGLE3 encoder/decoder graph (src/models/eagle3.cpp)
- Add feature extraction from target model layers
- Add g_embeddings handling for decoder input
- Add GGML_TENSOR_FLAG_SYNC for GPU synchronization
- Add --eagle3 flag for speculative-simple example
- Add EAGLE3 model conversion in convert_hf_to_gguf.py

EAGLE3 Architecture Overview :
```
┌─────────────────────────────────────────────────────────────────┐
│                    EAGLE3 Overview                              │
└─────────────────────────────────────────────────────────────────┘

  Target Model          EAGLE3 Encoder         EAGLE3 Decoder
  (LLaMA 8B)              (FC Layer)           (1-layer Transformer)
[...]

PR #12648, opened 2025-03-29
tts : implement sesame CSM + Mimi decoder

ngxson:
Related to #12392

Tbh it is more complicated than expected.

This PR only contains the backbone + decoder:
- Input: text
- Output: audio codes, which can then converted to waveform via https://github.com/ggml-org/llama.cpp/pull/12636

## How to try this?

By default, all GGUF files are downloaded from [ggml-org Hugging Face's account](https://huggingface.co/ggml-org/sesame-csm-1b-GGUF)

```sh
# build (make sure to have LLAMA_CURL enabled)
cmake -B build -DLLAMA_CURL=ON
cmake --build build -j --target llama-tts-csm

# run it
./build/bin/llama-tts-csm -p "[0]Hi, my name is Xuan Son. I am software engineer at Hugging Face."
```

Alternatively, GGUF files can be converted using `convert_mimi_to_gguf.py` and `convert_csm_to_gguf.py` under `example/tts` directory. These script uses `transformers.AutoModel` under the hood, so they will also handle downloading safetensors file automatically.

Note: it pronounces "Xuan" incorrectly, but the rest is OK

https://github.com/user-attachments/assets/fffcec71-6b75-4e34-8179-dd718e4c47fa

## How Sesame CSM works?

The model contains a backbone and a decoder, both are based on llama 3.x architecture (auto-aggressive).

1. The input text will firstly be processed by backbone, the output is (1) a RVQ semantic code and (2) the raw embedding from last layer, after norm
2. These 2 output from backbone then get passed into decoder as input. The decoder then generate the next 31 RVQ acoustic tokens
3. At this point, 32 RVQ are generated, it then get "squash" back into one single vector, then pass back the the backbone
4. Repeat from step 1 to generate the next codes

```mermaid
flowchart TD
    A[Input Text, vocab 128_256 tokens] -- prompt input --> B

    subgraph Backbone
        B[Backbone transformer]
        B --> C[Output logits, vocab 65632 tokens]
        B --> D[Output Raw embd, vector of 2048 elem]
    end

    D -- vector input --> Proj
    C -- sampling --> Stoken[RVQ semantic token]
    Stoken --> Fin
    Stoken --> H

    subgraph Decoder
        Proj[Projector, reduce size to 1024]
        Fin[Input vocab: 65632 tokens] -- vector dim 2048 --> Proj
        Proj --> F[Decoder transformer]
        F --> G[Output logits: vocab 2051 tokens]
    end

[...]

PR #18886, opened 2026-01-16
llama : add MTP API

ngxson:
Nothing here is working, but I'm pushing this to discuss the direction of MTP (multi-token prediction) integration in `libllama`

Currently, this implementation is designed around `llama_cross`, which make it a bit similar to know `llama_encode|decode` works. I'm targeting GLM support in the first iteration because the weight already had the `nextn` tensors.

For the explanation of the API, see the comment below.

In theory, this should work the same way for deepseek3, xiaomi mimo, GLM, eagle3 (see [this comment](https://github.com/ggml-org/llama.cpp/pull/18039#discussion_r2699272434))


ngxson:
I spent more time thinking on this and I decide to change the approach:
- We now using 2 different `llama_context`, one for MTP and one for the main LLM
- Upon doing a `llama_decode()` on main LLM, a `llama_mtp_start()` call will follow to copy the state from main model to MTP model. I'm not 100% sure if this is needed, but adding for clarity for now.

The user code will look like this with this implementation:

```cpp
// load model
llama_model * model_llm = llama_model_load_from_file("main-model.gguf", ...);
llama_model * model_mtp = model_llm; // by default, MTP is built-in
// optionally, may load MTP from another file
if (!llama_model_has_mtp(model_llm)) {
    model_mtp = llama_model_load_from_file("mtp-model.gguf", ...);
}

llama_context * ctx_llm = llama_init_from_model(model_llm, ...);
llama_context * ctx_mtp = llama_init_from_model(model_mtp, ...);

// example generating one token with main LLM

common_batch_add(batch, last_token, ...);
llama_decode(batch);
llama_token curr_token = common_sampler_sample(...);

llama_mtp_start(ctx_llm, ctx_mtp);
// at this point, MTP state is populated to ctx_mtp->cross

std::vector<llama_token> mtp_tokens;
llama_token last_mtp_token = curr_token;
for (i = 0; i < n_draft_max; i++) {
    common_batch_clear(batch);
    common_batch_add(batch, last_mtp_token, ...);
    llama_decode(ctx_mtp, batch); // use state from ctx_mtp->cross, and update it afterwards
    last_mtp_token = common_sampler_sample(...);
    mtp_tokens.push_back(last_mtp_token);
}
```

While on the surface, this approach doesn't seem much different from https://github.com/ggml-org/llama.cpp/pull/18039 or https://github.com/ggml-org/llama.cpp/pull/15225, it does have some important advantages:
1. Because the internal MTP state is never leaked to the user code, the implementation is more expandable in the future without breaking the API.
2. One of the main problem currently is that because the MTP model is very small. The synchronization CPU <--> backend between each token generated becomes a bottleneck in this case. In the long-term, we can optimize this by **not** letting `common_sampler_sample` to call `llama_synchronize()`. This allow generating MTP tokens to be fully asynchronous, and since sampling can be done on the backend, the backend will never need to pause until it finishes generating enough `n_draft_max` tokens.

ngxson:
> 2\. This allow generating MTP tokens to be fully asynchronous, and since sampling can be done on the backend, the backend will never need to pause until it finishes generating enough `n_draft_max` tokens.

Elaborating this point a bit more: while working on https://github.com/ggml-org/llama.cpp/pull/12648, I have noticed that the synchronization has significant impact on generation speed of a small model.

The sesame CSM contains of 2 parts: a main LLM that generate embeddings and a smaller decoder that generates 12 audio tokens for a given embedding. It is pretty much close to a speculative model with `n_draft_max = 12`. My implementation was quite slow compared to pytorch version because on pytorch version, they can generate all 12 tokens without interrupting the backend (in pytorch version, they explicitly implemented backend sampling to by pass the CPU sampler).

So I hope the same idea can be implemented to MTP model at some point, as backend sampling is already there in llama.cpp.

include/llama.h
```diff
@@ -370,6 +370,7 @@ extern "C" {
         bool kv_unified;  // use a unified buffer across the input sequences when computing the attention
                           // try to disable when n_seq_max > 1 for improved performance when the sequences do not share a large prefix
                           // ref: https://github.com/ggml-org/llama.cpp/pull/14363
+        bool is_mtp;      // create context for Multi-Token Prediction (MTP)
```
ggerganov: Something to consider is moving `llm_graph_type` to the public interface and changing this to:

```c
llama_graph_type gtype;
```

If we do it like this, we can even rework the `llama_encode()` and `llama_decode()` into a single `llama_process()`.

src/llama-context.cpp
```diff
@@ -3203,6 +3253,12 @@ float * llama_get_embeddings_seq(llama_context * ctx, llama_seq_id seq_id) {
     return ctx->get_embeddings_seq(seq_id);
 }
 
+int32_t llama_mtp_start(llama_context * ctx_llm, llama_context * ctx_mtp) {
+    ctx_llm->synchronize();
+
+    return ctx_llm->cpy_mtp_state(*ctx_mtp);
+}
+
```
ggerganov, line 3186: Also not sure if this is the best way, but seems OK for now. Would look into generalizing somehow to not be too MTP specific. I.e. a more generic mechanism for sharing data between contexts.

ggerganov (commented):
Overall looks good. I agree that pt.2 from https://github.com/ggml-org/llama.cpp/pull/18039#discussion_r2701873697 should be tested and see if it works to make the Eagle3 context decoder-only.

src/llama-context.cpp
ngxson: I'm thinking about another version of `llama_cross` that will encapsulate multiple `llama_context` inside it:

```cpp
struct llama_cross {
    llama_context * ctx_enc; // text encoder models like T5
    llama_context * ctx_llm;
    llama_context * ctx_mtp;
    llama_context * ctx_mtmd;
};
```

Such that when a `llama_process()` is called on one context, it will propagate the state to another context.

For now I cannot think of a better way to avoid having purpose-specific naming like `mtmd`, `mtp` because the data between 2 contexts can vary depending on the task. But I think we can iterate from this idea.

For the current PR, I think I can proceed with the `llama_mtp_start` because it will easy to adapt to whatever API we may come up with in the future.

jukofyork:
This was just linked on Reddit today:

https://z-lab.ai/projects/dflash/

https://github.com/z-lab/dflash

and seems worth thinking about for any future MPT/Eagle API:

> We demonstrate that a free lunch does exist. Our key insight is that the large AR target model’s hidden features implicitly contain information about future tokens, a phenomenon also observed by [4].
>
> Instead of asking a tiny diffusion model to reason from scratch, DFlash conditions the draft model on context features extracted from the target model. This fuses the deep reasoning capabilities of the large model with the parallel generation speed of the small diffusion drafter.

ngxson:
I was quite distracted with other things lately, going back to this PR now. Hopefully get GLM working first, then Eagle

Deepseek should be the same as GLM, but need to modify the conversion script, which will need to be done in a dedicated PR

ggerganov:
Alright, I've also just restarted work on https://github.com/ggml-org/llama.cpp/pull/18039. I think we will meet both implementations at some point and be able to support both MTP and Eagle.

forforever73:
@ngxson  I noticed in [#15225](https://github.com/ggml-org/llama.cpp/pull/15225) that their MTP implementation keeps a full KV cache for the draft model, while in this PR the draft context is cleared before each draft (llama_memory_clear(ctx_dft, false)).
  Does this design risk degrading model quality, since the MTP context no longer has the full historical KV?

ngxson:
@forforever73 it was meant for testing. I will need to find a way to copy KV cache between main LLM and MTP layer without copying through host

gundemirbas:
is there any process on this draft? Currently qwen3.5 cannot use either MTP nor self spec

am17an:
@ngxson are you still planning to work on this? If not I can take this up, I think it will be a good feature to add

JMPSequeira:
This, along side tensor parallelism, is early xmas

PR #18887, opened 2026-01-16
DirectIO Model Loading: Extend and fix Fallback

JTischbein:
Due to issues with the DirectIO model loading path on Android this PR adds `EINVAL` errors to the fallback condition. Also there was a bug in the fallback to `mmap` in case `open` with the DirectIO flag fails.


JTischbein:
Fixes #18766

PR #18890, opened 2026-01-17
fit-params : Handle n_ctx 0 for models that entirely fit with n_ctx_train

### LINKED ISSUE #18889: Eval bug: llama_model_fit results in zero context size
### Name and Version

HEAD

### Operating systems

Linux

### GGML backends

HIP

### Hardware

various

### Models

Small ones whose total weights and maximum context would fit onto the card(s)

The new llama_model_fit system is great and prevents a lot of hand tuning. I noticed it sometimes has issues, especially for models that are much smaller than available VRAM, resulting in a context size of 0, rather than the n_ctx_train, which would be expected for a model with plenty of headroom.

It looks like this is because there is an early return path when the provide cparams fit without adjustment. At least through the C API, this means the context size of 0 isn't special if the model fits, but is special if it doesn't.

Will attach a PR with a fix that resolves the issue for me, but may need more discussion of course.

### First Bad Commit

n/a

### Relevant log output

n/a

65a:
During initial fitting, if the n_ctx is 0, try to fit using n_ctx_train.
If initial fitting was successful, make sure to set n_ctx to n_ctx_train instead of leaving it at zero.

This could be done differently, but this approach avoids making cparams mutable deeper in the stack, and then having to clean it back to 0 if we aren't going to early return.

Fixes #18889

### DIFF
--- a/src/llama.cpp
+++ b/src/llama.cpp
@@ -81,7 +81,11 @@ static std::vector<llama_device_memory_data> llama_get_device_memory_data(
         throw std::runtime_error("failed to load model");
     }
 
-    llama_context * ctx = llama_init_from_model(model, *cparams);
+    llama_context_params cparams_copy = *cparams;
+    if (cparams_copy.n_ctx == 0)
+        cparams_copy.n_ctx = model->hparams.n_ctx_train;
+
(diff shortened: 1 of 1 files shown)


65a:
This problem still exists at HEAD

PR #18892, opened 2026-01-17
examples: llama evaluation tool for mmlu, aime, gsm8k

gatbontonpc:
# Overview
This is a bare minimum implementation of a llama evaluation tool written in python. It runs against an OpenAI-compatible endpoint (e.g. llama-server). 

**Feedback welcome!**
## Supported Tests

- GSM8K — grade-school math (free-form, numeric answers)
- AIME — competition math (integer answers)
- MMLU — multi-domain multiple-choice
- HellaSwag — commonsense reasoning (multiple-choice)
- ARC — grade-school science (multiple-choice)
- WinoGrande — commonsense coreference (multiple-choice)

## Example output

```
./llama-server -m model.gguf --port 8033
python examples/llama-eval/llama-eval.py \
  --path_server http://localhost:8033 \
  --prompt_source arc \
  --n_prompts 100
```
```
llama-eval duration:           2025.81 s

=== llama-eval suite summary ===
Task                 Acc  Correct    Total  Invalid    Error
-----------------------------------------------------------------
aime               0.067        6       90       14        0
arc                0.950       95      100        0        0
gsm8k              0.350       35      100       10        0
hellaswag          0.720       72      100        0        0
mmlu               0.670       67      100        0        0
winogrande         0.690       69      100        0        0
-----------------------------------------------------------------
ALL                0.583      344      590
```

## Codex Generated summary

This PR introduces llama-eval, a deliberately lean evaluation tool intended to make running common LLM benchmarks against an OpenAI-compatible endpoint (e.g. llama-server) trivial, portable, and easy to reason about.

The design is explicitly motivated by the discussion in #18195, where existing evaluator frameworks (e.g. NVIDIA Evaluator, GPT-OSS evals) were found to be overly complex, operationally heavy, and difficult to run locally—especially on macOS.

llama-eval prioritizes:

- Zero infrastructure overhead 
- Near-zero dependencies
- Single-command execution
- Direct HTTP interaction with the model server

The result is a tool that can be cloned and run immediately, without Docker, schedulers, config files, or external services.

### DIFF
--- a/examples/llama-eval/llama-eval.py
+++ b/examples/llama-eval/llama-eval.py
@@ -1,0 +1,358 @@
+#!/usr/bin/env python3
+
+import re
+import argparse
+import json
+import os
+import random
+import subprocess
+from time import sleep, time
+from typing import Optional, Union
+
+import datasets
+import logging
+import requests
+from tqdm.contrib.concurrent import thread_map
+from typing import Iterator
+from abc import ABC
+
+logging.basicConfig(level=logging.INFO, format='%(message)s')
+logger = logging.getLogger("llama-eval")
+
+
+MATH_TEMPLATE = """
+{question}
+Put your final answer within \\boxed{{}}.
+"""
+
+MC_FROM_INT = {
+    0: "A",
+    1: "B",
+    2: "C",
+    3: "D",
+}
+
+
+def format_multiple_choice(prompt: str, choices: list[str]):
+    QUERY_TEMPLATE_MULTICHOICE = """
+    {question}
+
+    (A) {A}
+    (B) {B}
+    (C) {C}
+    (D) {D}
+
+    Express your final answer as the corresponding option 'A', 'B', 'C', or 'D'. Put your final answer within \\boxed{{}}.
+
+    """.strip()
+    A_str = choices[0]
+    B_str = choices[1]
+    C_str = choices[2]
+    D_str = choices[3]
+    query = QUERY_TEMPLATE_MULTICHOICE.format(
+        question=prompt, A=A_str, B=B_str, C=C_str, D=D_str
+    )
+    return query
+
+
+# Preprocess hellaswag
+def preprocess(text):
+    text = text.strip()
+    # NOTE: Brackets are artifacts of the WikiHow dataset portion of HellaSwag.
+    text = text.replace(" [title]", ". ")
+    text = re.sub("\\[.*?\\]", "", text)
+    text = text.replace("  ", " ")
+    return text
+
+
+def hellaswag_process_doc(doc):
+    ctx = doc["ctx_a"] + " " + doc["ctx_b"].capitalize()
+    question = preprocess(doc["activity_label"] + ": " + ctx)
+    proc_answers = [preprocess(answer) for answer in doc["endings"]]
+    prompt = format_multiple_choice(question, proc_answers)
+    out_doc = {
+        "prompt": prompt,
+        "gold": MC_FROM_INT[int(doc["label"])],
+    }
+    return out_doc
+
+
+def mmlu_process_doc(doc):
+    prompt = format_multiple_choice(doc["question"], doc["choices"])
+    out_doc = {
+        "prompt": prompt,
+        "gold": MC_FROM_INT[int(doc["answer"])],
+    }
+    return out_doc
+
+
+def extract_boxed_text(text):
+    pattern = r"boxed{(.*?)}|framebox{(.*?)}"
+    matches = re.findall(pattern, text, re.DOTALL)
+    logger.debug(matches)
+    if matches:
+        for match in matches[::-1]:
+            for group in match:
+                if group != "":
+                    return group.split(",")[-1].strip()
+    logger.warning(
+        "Could not extract boxed text. Using last integer. Maybe expand context window"
+    )
+    pattern = r"\d+"  # get the last integer if no pattern found
+    matches = re.findall(pattern, text, re.DOTALL)
+    if matches:
+        return matches[-1]
+
+    return ""
+
+
+def get_prompts_text(
+    dataset_name: str, ds: datasets.Dataset
+) -> Optional[tuple[list[str], list[str]]]:
+    ret = []
+    if dataset_name.lower() == "mmlu":
+        ds = ds.map(mmlu_process_doc)
+        ret = ds["prompt"], ds["gold"]
+    elif dataset_name.lower() == "hellaswag":
+        ds = ds.map(hellaswag_process_doc)
+        ret = ds["prompt"], ds["gold"]
+    elif dataset_name.lower() == "aime":
+        ds = ds.map(
+            lambda k: {
+                "prompt": MATH_TEMPLATE.format(
+                    question=k["problem"],
+                )
+            }
+        )
+        ret = ds["prompt"], ds["answer"]
+    elif dataset_name.lower() == "gsm8k":
+        ds = ds.map(lambda k: {"prompt": MATH_TEMPLATE.format(question=k["question"])})
+        la = []
+        for answer in ds["answer"]:
+            la.append(answer.split("### ")[-1].rstrip())
+        ret = ds["prompt"], la
+    else:
+        return None
+
+    return ret
+
+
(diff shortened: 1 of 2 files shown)


strawberrymelonpanda:
Personally I'd really like something like this. As ggerganov implied in #18195, I find pretty much any and all of the evaluation harnesses out there to just be too much effort for too little payoff. They probably work great for researchers.

"Do not include any explanation." might lower scores on models that need a bit of time to "think" out loud? 
Is there any randomization of the correct choice?

Can this easily support custom benchmarks? I've written my own evaluation scripts to test new LLMs, but I'd love to scrap it in favor of something simpler like this. If so, I could probably reverse engineer the format from an existing, but if this goes to production a quick note on writing benchmarks would be welcome.

Code evaluation tests would be nice-to-have, but sadly that's probably when you start getting into Docker containers for safety and/or dependencies for things like [Bubblewrap](https://github.com/containers/bubblewrap), so multiple-choice probably makes sense to sidestep that complexity.

jeffbolznv:
I like this because it gives a more objective quality assessment for evaluation than just "it doesn't become incoherent". It also does some more irregular/batched workloads that we don't hit in llama-bench/llama-cli.

examples/llama-eval/llama-eval.py
```diff
@@ -0,0 +1,703 @@
...
+    task = TASK_DICT.get(case.task)
+    if task is None:
+        result["error"] = f"unknown_task: {case.task}"
+        return result
+    logger.debug(case.prompt)
+
+    json_data = build_request(case, data["n_predict"])
+    res_json = {}
+    try:
+        response = session.post(f"{server_address}/v1/completions", json=json_data)
```
ggerganov, line 444: The evaluation should use the `/v1/chat/completions` endpoint since we are going to be mainly evaluating models fine tuned for a specific chat template.

The `/v1/completions` support could be added later, but for now it's not important to support it.

examples/llama-eval/llama-eval.py
```diff
@@ -0,0 +1,703 @@
...
+
+TASK_DICT: dict[str, type[TaskSpec]] = {
+    "mmlu": MMLU_Task,
+    "aime": Aime_Task,
+    "gsm8k": Gsm8k_Task,
+    "hellaswag": Hellaswag_Task,
+    "arc": ARC_Task,
+    "winogrande": WinoGrande_Task,
+}
+
```
ggerganov, line 398: Let's keep it simple and start with just AIME2025 for now, until we polish the UX. As a reference, use the `gpt-oss` evals from https://github.com/openai/gpt-oss/tree/main/gpt_oss/evals.

For example, the correct prompt is here:

https://github.com/openai/gpt-oss/blob/599476783c6f88508dab8577808b5ead5cbee8d2/gpt_oss/evals/aime_eval.py#L11-L16

examples/llama-eval/llama-eval.py
```diff
@@ -0,0 +1,703 @@
...
+    pattern = r"boxed{(.*?)}|framebox{(.*?)}"
+    matches = re.findall(pattern, text, re.DOTALL)
+    logger.debug(matches)
+    if matches:
+        for match in matches[::-1]:
+            for group in match:
+                if group != "":
+                    return group.split(",")[-1].strip()
+    logger.debug("Could not extract boxed text. Maybe expand context window")
+
```
ggerganov, line 54: We should abstract this to support an external "grader" or "judge" to avoid problems that I described in: https://huggingface.co/openai/gpt-oss-120b/discussions/132

Regexing will never work well enough, but post-processing the final answer with an LLM should be solid.

examples/llama-eval/llama-eval.py
```diff
@@ -0,0 +1,703 @@
...
+        )
+    finally:
+        if session is not None:
+            session.close()
+
+    t1 = time()
+    logger.info(f"\nllama-eval duration:           {t1-t0:.2f} s")
+    results.extend(checkpoint_results)
+    pertask_results = aggregate_by_task(results)
+    print_summary(pertask_results)
```
ggerganov, line 632: We need a way to see the current success rate in real-time and not wait for the entire eval to finish. By default, we should see "correct / not correct" for each task. With extra verbosity, we should also see produced answer vs expected answer for each task as soon as it completes.

ggerganov (commented):
Thanks for starting this.

My general recommendation is to start things simple and focus on supporting a single eval until we figure out exactly what we want in terms of user experience. This will simplify the review process and from there we can more easily expand with more evals.

As a first objective, we should aim to support AIME2025 as I am most familiar with it.

See my comments below for more info.

examples/llama-eval/llama-eval.py
```diff
@@ -0,0 +1,703 @@
...
+    "arc": ARC_Task,
+    "winogrande": WinoGrande_Task,
+}
+
+
+def build_request(case: Case, n_predict: int) -> dict[str, Any]:
+    json_data = {
+        "n_predict": n_predict,
+        "max_tokens": n_predict,
+        "temperature": 0,
```
ggerganov, line 404: By default, no sampling parameters should be passed. We will rely on server-driven sampling configuration in such cases.

Optionally the user of this script can specify common sampling parameters such as: `temp`, `top-k`, `top-p`, `min-p`. These should be printed prominently at the start of the run.

examples/llama-eval/llama-eval.py
```diff
@@ -0,0 +1,703 @@
...
+        action="store_false",
+        help="Disble resuming from last state stored in checkpoint file",
+    )
+    parser.add_argument(
+        "--checkpoint-file",
+        type=Path,
+        dest="checkpoint_file",
+        default="./llama-eval-checkpoint.jsonl",
+        help="Checkpoint file to read last state from",
+    )
```
ggerganov, line 689: The checkpoint system should be a bit more strongly typed compared to the current ad-hoc json. I think first we have to implement an object that contains all the information about an eval run (tasks, task states, unique ID of the run, sampling config, etc.). And then the checkpointing system would simply save/load these state objects. For now we can remove all of the checkpoint logic and we will come back to that after we design the eval state.

examples/llama-eval/llama-eval.py
```diff
@@ -0,0 +1,703 @@
...
+        "To pass arguments such as the model path to the server, set the corresponding environment variables (see llama-server --help). "
+        "The reported numbers are the speeds as observed by the Python script and may differ from the performance reported by the server, "
+        "particularly when the server is fast vs. the network or Python script (e.g. when serving a very small model)."
+    )
+    parser.add_argument(
+        "--path_server",
+        type=str,
+        default="http://localhost:8033",
+        help="llama-server url",
+    )
```
ggerganov, line 648: We have to support a list of endpoints so we can distribute evals on multiple servers. For each endpoint, we will specify a separate number of worker threads submitting tasks.

ggerganov (commented):
So to wrap up my feedback for now is to:

- Simplify as much as possible and focus on one eval
- Implement an "eval state" object:
  - ID
  - list of tasks
  - task states
  - sampling config
- Implement a "processor" object
  - list of endpoints
  - threads per endpoints
  - grade/judge type (regex, endpoint or cli tool)
- The processor accepts an eval state and starts processing it. It will be responsible for dumping the eval state from time to time as it progresses

examples/llama-eval/llama-eval.py
```diff
@@ -0,0 +1,703 @@
...
+import json
+import threading
+
+logging.basicConfig(level=logging.INFO, format='%(message)s')
+logger = logging.getLogger("llama-eval")
+
+MATH_TEMPLATE = """
+{question}
+Do not include any explanation. Put your final answer within \\boxed{{}}.
+"""
```
JohannesGaessler, line 26: The OAI API supports outputs structured as e.g. JSON. I think it makes more sense to use that for extracting the answer rather than to use something like this. In any case, you should be aware that [in my testing](https://github.com/JohannesGaessler/elo_hellm) it is much harder for models to immediately output the correct answer compared to first letting them reason about it. The approach that I took was to let the model first answer normally and to then prompt it for a final answer using a structured output.

PR #18898, opened 2026-01-17
graph : utilize `ggml_build_forward_select()` to avoid reallocations

ggerganov:
target #18550 
cont #17617

Extracted the usage of the new `ggml_build_forward_select()` from #18550 into a separate PR in order to more clearly demonstrate how it can be applied to avoid graph reallocations.

Here we utilize it to avoid reallocations when switching between different types of inputs (tokens or embeddings) for most models.

Also enable the server CI to report errors if unexpected reallocations occur during the server tests.

### DIFF
--- a/src/llama-graph.cpp
+++ b/src/llama-graph.cpp
@@ -34,8 +34,8 @@ void llm_graph_input_embd::set_input(const llama_ubatch * ubatch) {
 bool llm_graph_input_embd::can_reuse(const llm_graph_params & params) {
     bool res = true;
 
-    res &= (!tokens && !params.ubatch.token) || (tokens && tokens->ne[0] == params.ubatch.n_tokens);
-    res &= (!embd   && !params.ubatch.embd)  || (embd   &&   embd->ne[1] == params.ubatch.n_tokens);
+    res &= (!params.ubatch.token) || (tokens && tokens->ne[0] == params.ubatch.n_tokens);
+    res &= (!params.ubatch.embd)  || (embd   &&   embd->ne[1] == params.ubatch.n_tokens);
 
     return res;
 }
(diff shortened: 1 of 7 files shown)


src/llama-graph.cpp
```diff
@@ -1298,19 +1310,40 @@ ggml_tensor * llm_graph_context::build_inp_embd(ggml_tensor * tok_embd) const {
 
             cur = ggml_add(ctx0, cur, inpL_delta);
         }
-    } else {
-        inp->embd = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, n_embd, ubatch.n_tokens);
-        ggml_set_input(inp->embd);
+
+        if (hparams.n_deepstack_layers > 0) {
```
ngxson: I think the more generic condition here is `n_embd_inp != n_embd`

src/llama-graph.cpp
```diff
@@ -1298,19 +1310,40 @@ ggml_tensor * llm_graph_context::build_inp_embd(ggml_tensor * tok_embd) const {
...
+
+    // vector embeddings path (ubatch.embd != nullptr)
+    {
+        auto & cur = inps[1];
 
         cur = inp->embd;
+
+        if (hparams.n_deepstack_layers > 0) {
+            cur = ggml_view_2d(ctx0, cur, hparams.n_embd, n_tokens, cur->nb[1], 0);
+            cur = ggml_cont   (ctx0, cur); // makes the shape of this node the same as the ubatch.token path
```
ngxson: Instead of resizing input embeddings row size to `n_embd`, I'm wondering if we can/should do the reverse: pad the input token embedding to `n_embd_inp` using `ggml_pad`

ngxson: Note that doing this way will allow us to remove the `ggml_build_forward_select` in model cgraph, since the `ggml_add(ctx0, cur, ds)` path will always be taken. Although, I'm a bit worry if `ggml_pad` will have negative impact on performance

ggerganov: Good idea - applied in c84637d808cd366a2a14ee2ee3e38d5c3b037e3b

PR #18900, opened 2026-01-17
jinja : add missing tojson filter for bool

CISC:
Boolean type was missing `tojson`.

Add more literal tests.

PR #18901, opened 2026-01-17
jinja : fix lexing of float literals with sign

CISC:
The lexer was only checking for integers and not floats when prepended with a unary operator.

Fixes https://github.com/ggml-org/llama.cpp/pull/18900#issuecomment-3764067660

### DIFF
--- a/common/jinja/lexer.cpp
+++ b/common/jinja/lexer.cpp
@@ -259,6 +259,11 @@ lexer_result lexer::tokenize(const std::string & source) {
 
                     // Check for numbers following the unary operator
                     std::string num = consume_while(is_integer);
+                    if (pos < src.size() && src[pos] == '.' && pos + 1 < src.size() && is_integer(src[pos + 1])) {
+                        ++pos; // Consume '.'
+                        std::string frac = consume_while(is_integer);
+                        num += "." + frac;
+                    }
                     std::string value = std::string(1, ch) + num;
                     token::type t = num.empty() ? token::unary_operator : token::numeric_literal;
                     // JJ_DEBUG("consumed unary operator or numeric literal: '%s'", value.c_str());
--- a/common/jinja/lexer.cpp
+++ b/common/jinja/lexer.cpp
@@ -91,6 +91,16 @@ lexer_result lexer::tokenize(const std::string & source) {
         return str;
     };
 
+    auto consume_numeric = [&]() -> std::string {
+        std::string num = consume_while(is_integer);
+        if (pos < src.size() && src[pos] == '.' && pos + 1 < src.size() && is_integer(src[pos + 1])) {
+            ++pos; // Consume '.'
+            std::string frac = consume_while(is_integer);
+            num += "." + frac;
+        }
+        return num;
+    };
+
     auto next_pos_is = [&](std::initializer_list<char> chars, size_t n = 1) -> bool {
         if (pos + n >= src.size()) return false;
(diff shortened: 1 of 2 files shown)


tests/test-jinja.cpp
```diff
@@ -247,6 +247,12 @@ static void test_expressions(testing & t) {
         "Bob"
     );
 
+    test_template(t, "negative float (not dot notation)",
+        "{{ -1.0 }}",
+        json::object(),
+        "-1.0"
+    );
```
aldehir, line 254:
```suggestion
    test_template(t, "negative float (not dot notation)",
        "{{ -1.0 }}",
        json::object(),
        "-1.0"
    );

    test_template(t, "negative float no frac (not dot notation)",
        "{{ -1. }}",
        json::object(),
        "-1.0"
    );
```

common/jinja/lexer.cpp
```diff
@@ -259,6 +259,11 @@ lexer_result lexer::tokenize(const std::string & source) {
 
                     // Check for numbers following the unary operator
                     std::string num = consume_while(is_integer);
+                    if (pos < src.size() && src[pos] == '.' && pos + 1 < src.size() && is_integer(src[pos + 1])) {
+                        ++pos; // Consume '.'
+                        std::string frac = consume_while(is_integer);
+                        num += "." + frac;
+                    }
```
ngxson: while looking at this, I'm wondering if this case is currently supported: `.5` and `-.5` (they are valid python expressions)

we can probably extract this code to in to a new function `is_fraction`, and reuse the function to parse this case (ofc, can be done in a follow-up PR)

ngxson (approved):

common/jinja/lexer.cpp
```diff
@@ -259,6 +259,11 @@ lexer_result lexer::tokenize(const std::string & source) {
 
                     // Check for numbers following the unary operator
                     std::string num = consume_while(is_integer);
+                    if (pos < src.size() && src[pos] == '.' && pos + 1 < src.size() && is_integer(src[pos + 1])) {
```
aldehir:
```suggestion
                    if (pos < src.size() && src[pos] == '.') {
```

Should probably do this for the case when it doesn't follow a unary operator as well.

common/jinja/lexer.cpp
aldehir: I think a `consume_numeric()` would be appropriate here.

common/jinja/lexer.cpp
CISC: Needs to check that previous character was integer though.

common/jinja/lexer.cpp
CISC: Ok, needs refinement, worth doing in this PR I think.

CISC: Actually, it seems jinja does not adhere to this, gives an error with no leading digit.

common/jinja/lexer.cpp
CISC: Ok, checked this as well, jinja does not support `1.` floats, gives an error.

aldehir: Ah, ok. My mistake.

aldehir (approved):

common/jinja/lexer.cpp
ngxson: hmm ok TIL not all python syntax are considered as valid jinja syntax..

PR #18904, opened 2026-01-17
jinja : fix object item order (and properly implement dictsort)

CISC:
Objects' `as_object` returned the (map sorted) values instead of the (insertion ordered) values.

This had the side-effect of making `dictsort` a no-op, so fixing this allowed me to implement `reverse` and `by='value'`.

### DIFF
--- a/common/jinja/value.h
+++ b/common/jinja/value.h
@@ -146,7 +146,7 @@ struct value_t {
     virtual string as_string() const { throw std::runtime_error(type() + " is not a string value"); }
     virtual bool as_bool() const { throw std::runtime_error(type() + " is not a bool value"); }
     virtual const std::vector<value> & as_array() const { throw std::runtime_error(type() + " is not an array value"); }
-    virtual const std::vector<std::pair<std::string, value>> & as_object() const { throw std::runtime_error(type() + " is not an object value"); }
+    virtual const std::vector<std::pair<std::string, value>> & as_ordered_object() const { throw std::runtime_error(type() + " is not an object value"); }
     virtual value invoke(const func_args &) const { throw std::runtime_error(type() + " is not a function value"); }
     virtual bool is_none() const { return false; }
     virtual bool is_undefined() const { return false; }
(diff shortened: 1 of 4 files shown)


common/jinja/value.h
```diff
@@ -299,7 +302,7 @@ struct value_object_t : public value_t {
         val_obj.insert(key, val);
     }
     virtual std::string type() const override { return "Object"; }
-    virtual const std::map<std::string, value> & as_object() const override { return val_obj.unordered; }
+    virtual const std::vector<std::pair<std::string, value>> & as_object() const override { return val_obj.ordered; }
```
ngxson: I think this should be renamed to `as_ordered_object()`

Also, `value_to_json_internal` and `dictsort` are currently reading `val->val_obj.ordered` directly, it should be changed to use `as_ordered_object()`

CISC: `as_ordered_object` could be misunderstood to mean sorted though.

ngxson: I think it should be fine, should be equivalent to `nlohmann::ordered_json`. I believe the sorted version called [flat_map](https://en.cppreference.com/w/cpp/container/flat_map.html) though I've never used it.

CISC: ~Doesn't read like it:~ Oh, nvm.
> The class template flat_map acts as a wrapper to the two underlying containers, passed as objects of type KeyContainer and MappedContainer respectively. The first container is sorted, and for each key its corresponding value is in the second container at the same index (offset).

CISC: It's not that simple for `dictsort` as it copies and modifies the object, this is as clean as I could get it without introducing a `sort` member.

ngxson: Ah yeah, it's ok then. I thought it only read the keys.

ngxson (approved):

CISC:
Bah, and now this needs fixing:
```
20: While executing BinaryExpression at line 27, column 52 in source:
20: ...ns.interval is none and thinking_budget <= k -%}↵        {%- set ns.interval = v...
20:                                            ^
20: Error: Unknown operator "<=" between Integer and String
```

ngxson:
There is a flag `is_key_numeric` in `value_t::map`, you just need to make sure it's passed down to the newly created object (I know it's quite a hacky way, but currently there is only one single template having numeric as object key)

CISC:
> There is a flag `is_key_numeric` in `value_t::map`, you just need to make sure it's passed down to the newly created object (I know it's quite a hacky way, but currently there is only one single template having numeric as object key)

Ahhh...

CISC:
> There is a flag `is_key_numeric` in `value_t::map`, you just need to make sure it's passed down to the newly created object (I know it's quite a hacky way, but currently there is only one single template having numeric as object key)

Shouldn't this be done at insertion instead?

CISC:
> > There is a flag `is_key_numeric` in `value_t::map`, you just need to make sure it's passed down to the newly created object (I know it's quite a hacky way, but currently there is only one single template having numeric as object key)
> 
> Shouldn't this be done at insertion instead?

Actually, we should probably store the key as `value` in the `vector` and `string` in the `map` and handle this properly, I'll look into it.

PR #18909, opened 2026-01-18
WebUI: New server loading page

dariusjlukas:
### Description

Update the webui loading page (loading.html) to be more consistent with the rest of the UI.

### Changes

Style loading.html similar to the Svelte UI and add a loading animation. Respect user-selected color theme.

| Description | Before | After |
| ----------------- | ---------- | ------- |
| Loading (dark mode) | existing-loading-screen | https://github.com/user-attachments/assets/8680b2d2-f9a9-4772-b943-67ba0cfd41d6 |
| Loading (light mode) | existing-loading-screen | https://github.com/user-attachments/assets/c2e69c84-b3f9-465e-9dd6-fc6ea672a055 |

### Generative AI Use

Use of generative AI for development was limited to code review and providing suggestions.


ngxson (changes requested):
This is not a good approach, it adds too much redundant code for a small feature that is used less than 1% of the time.

Instead, we can just remove `loading.html` and handle it on svelte webui, it can ping the `/models` or `/props` endpoint until HTTP code 200 is received

allozaur:
> This is not a good approach, it adds too much redundant code for a small feature that is used less than 1% of the time.
> 
> Instead, we can just remove `loading.html` and handle it on svelte webui, it can ping the `/models` or `/props` endpoint until HTTP code 200 is received

yep, and we actually had planned in internal roadmap. Closing this one, we'll revisit this in near future

PR #18915, opened 2026-01-18
fix: Use `tabular-nums` for chat message statistics

nathanlesage:
This PR implements a small change to the Server WebUI that utilizes the CSS font property `font-variant-numeric: tabular-nums` using the corresponding tailwind class `tabular-nums` to reduce the jittering that can be experienced with the fast updating of the chat message statistics.

Without this property, the WebUI's font uses variable-width numerical glyphs which will make the entire statistics "jitter" visually as the model generates a response. With this property, changing individual numbers does not change the width of the various informative messages.

This makes the statistics view much more stable and visually less stressful (?) when the model generates tokens.

## Changes

* Added the tailwind class `tabular-nums` to the `info` container of the assistant message component

### Before

https://github.com/user-attachments/assets/718c7b0a-e63e-4045-8f89-dacab19219b4

### After

https://github.com/user-attachments/assets/cd8c3d96-f291-49d4-8e98-42a4dad96244


allozaur (changes requested):
Great catch! :) Please just update the static build by running `npm run build` and pushing the updated `tools/server/public/index.html.gz` file

nathanlesage:
Whoops, missed that one. I think it worked…? (It did log a few warnings, I don't know if that is expected?)

allozaur:
> Whoops, missed that one. I think it worked…? (It did log a few warnings, I don't know if that is expected?)

Yeah, don't worry about the warnings, they are related to SCSS for KaTeX using deprecated syntax which we will address along the way.

nathanlesage:
Great! Let me know if anything else is wonky, so I can address it.

allozaur (approved):

PR #18916, opened 2026-01-18
ci : run test-jinja -py on high perf

CISC:
Adds new `ctest` label; `python`.

Create new test; `test-jinja-py`, runs on high perf where Python env is activated and dependencies are installed.

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -254,7 +254,7 @@ function gg_run_ctest_release {
     (time make -j$(nproc)                                    ) 2>&1 | tee -a $OUT/${ci}-make.log
 
     if [ -z ${GG_BUILD_LOW_PERF} ]; then
-        (time ctest --output-on-failure -L main ) 2>&1 | tee -a $OUT/${ci}-ctest.log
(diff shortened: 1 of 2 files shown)


CISC:
@ngxson gentle ping

ngxson (approved):

ngxson:
hmm, the CI fails because jinja2 is not available, I'm wondering if we should add the CI step to install it

CISC:
> hmm, the CI fails because jinja2 is not available, I'm wondering if we should add the CI step to install it

Which CI? It should only run on high perf, which installs all requirements (`transformers` installs `jinja2`).

ngxson:
This one for example: https://github.com/ggml-org/llama.cpp/actions/runs/21152102787/job/60830061989

```
26/41 Test #26: test-jinja-py .....................***Failed    1.99 sec
whitespace control
  trim_blocks removes newline after tag
    Python script failed with exit code: 1
    Output: Traceback (most recent call last):
  File "<string>", line 2, in <module>
ModuleNotFoundError: No module named 'jinja2'
```

CISC:
Ah, no, that's because pip install failed:
https://github.com/ggml-org/llama.cpp/actions/runs/21152102787/job/60830061989#step:5:29

PR #18921, opened 2026-01-18
fix(server): prevent log thread from blocking when child process dies

tdevelope:
Closes #18912

### Problem
In router mode, if a child server process crashes unexpectedly (e.g., OOM kill), the log thread can block on the child's stdout pipe. 
This prevents subprocess_join() from cleaning up the child process, potentially leaving zombie (<defunct>) processes.

### Solution
Modify the log thread to use non-blocking I/O with select() and a timeout. 
The thread now periodically checks if the child process is alive and exits cleanly if the child dies. 
subprocess_join() remains the responsibility of the parent thread, avoiding deadlocks.

### Note
This fix is Linux-only, as it relies on POSIX pipes and select(). 
Windows support is not included and can be added in a future update.

### DIFF
--- a/tools/server/server-models.cpp
+++ b/tools/server/server-models.cpp
@@ -507,21 +507,51 @@ void server_models::load(const std::string & name) {
         FILE * stdout_file = subprocess_stdout(child_proc.get()); // combined stdout/stderr
 
         std::thread log_thread([&]() {
-            // read stdout/stderr and forward to main server log
-            // also handle status report from child process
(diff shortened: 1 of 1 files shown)


ngxson:
See: https://github.com/ggml-org/llama.cpp/issues/18912#issuecomment-3768806323

~~Will re open this if someone confirm if the bug is valid~~

Edit: this is not a valid solution, won't reopen this

PR #18924, opened 2026-01-18
jinja : fix undefined keys and attributes and int/float as bool

CISC:
Return `undefined` on undefined keys and attributes.
Integers and floats can be represented as bools.

Added `falsy` tests.
Added `truthy` tests.

### DIFF
--- a/tests/test-jinja.cpp
+++ b/tests/test-jinja.cpp
@@ -264,7 +264,7 @@ static void test_conditionals(testing & t) {
         "yes"
     );
 
-    test_template(t, "is 1.0 thruthy",
+    test_template(t, "is 1.0 truthy",
         "{{ 'yes' if y else 'no' }}",
         {{"y", 1.0}},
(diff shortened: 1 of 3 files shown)


tests/test-jinja.cpp
```diff
@@ -191,6 +191,42 @@ static void test_conditionals(testing & t) {
...
+        "yes"
+    );
+
+    test_template(t, "is empty object falsy",
+        "{{ 'yes' if not y else 'no' }}",
+        {{"y", json::object()}},
+        "yes"
+    );
+
+    test_template(t, "is empty string falsy",
```
pwilkin, line 225: Zero should be falsy as well, right?

CISC: Sure, I'll add truthy tests as well...

ngxson (approved):

aldehir (approved):

PR #18929, opened 2026-01-19
convert : use n_groups instead of hardcoded values in reshape

danbev:
This commit modifies the conversion script for NemotronHModel to use the `n_groups` hyperparameter, and allow Python to calculate the the last dimension, using -1, when reshaping the `mixer.norm.weight` tensor.

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -9183,8 +9183,7 @@ class NemotronHModel(GraniteHybridModel):
                 return [(mapped_name, reshaped_data)]
 
             if name.endswith("mixer.norm.weight"):
-                n_groups = self.hparams["n_groups"]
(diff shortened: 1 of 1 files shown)


convert_hf_to_gguf.py
```diff
@@ -9183,7 +9183,8 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
                 return [(mapped_name, reshaped_data)]
 
             if name.endswith("mixer.norm.weight"):
-                reshaped_data = data_torch.reshape(8, 512)
+                n_groups = self.hparams["n_groups"]
+                reshaped_data = data_torch.reshape(n_groups, -1)
```
```suggestion
                reshaped_data = data_torch.reshape(self.n_group, -1)
```

CISC (approved):

PR #18934, opened 2026-01-19
ggml-cuda: enable cuda-graphs for `n-cpu-moe`

am17an:
Add piece-wise cuda graph for the multiple split case. Currently cuda graphs get disabled when there are splits as we only keep 1 cuda graph per device. Multiple updates with different sized splits/shapes triggers the disable. 
This PR adds a cuda graph per split (a split is keyed via the first node in the split)

Tested on 2x4090 and 1x 5090. My guess is the speed-up will much larger on windows due to high kernel-launch latency

| Model                               |   n_cpu_moe | Test   |   t/s 3d55846a5 |   t/s n-cpu-moe-piecewise |   Speedup |
|:------------------------------------|------------:|:-------|----------------:|--------------------------:|----------:|
| glm4moe 106B.A12B IQ4_XS - 4.25 bpw |           8 | tg128  |           60.84 |                     63.12 |      1.04 |
| glm4moe 106B.A12B IQ4_XS - 4.25 bpw |          16 | tg128  |           48.25 |                     50.75 |      1.05 |
| glm4moe 106B.A12B IQ4_XS - 4.25 bpw |          32 | tg128  |           32.84 |                     35.03 |      1.07 |
| glm4moe 106B.A12B IQ4_XS - 4.25 bpw |          64 | tg128  |           25.08 |                     27.49 |      1.10 |
| gpt-oss 120B MXFP4 MoE              |           8 | tg128  |           95.96 |                    100.93 |      1.05 |
| gpt-oss 120B MXFP4 MoE              |          16 | tg128  |           70.42 |                     75.44 |      1.07 |
| gpt-oss 120B MXFP4 MoE              |          32 | tg128  |           44.80 |                     48.47 |      1.08 |
| gpt-oss 120B MXFP4 MoE              |          64 | tg128  |           40.87 |                     44.90 |      1.10 |

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2964,18 +2964,25 @@ static bool ggml_cuda_graph_node_properties_match(ggml_tensor * node, ggml_cuda_
     return true;
 }
 
+static const void * ggml_cuda_graph_get_key(ggml_cgraph * cgraph) {
+    return cgraph->nodes[0];
+}
+
 static bool ggml_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx, ggml_cgraph * cgraph) {
 
     bool res = false;
 
-    if (cuda_ctx->cuda_graph->instance == nullptr) {
+    const void * graph_key = ggml_cuda_graph_get_key(cgraph);
+    ggml_cuda_graph * graph = cuda_ctx->cuda_graph(graph_key);
+
+    if (graph->instance == nullptr) {
         res = true;
     }
 
     // Check if the graph size has changed
-    if (cuda_ctx->cuda_graph->props.size() != (size_t)cgraph->n_nodes + cgraph->n_leafs) {
+    if (graph->props.size() != (size_t)cgraph->n_nodes + cgraph->n_leafs) {
         res = true;
-        cuda_ctx->cuda_graph->props.resize(cgraph->n_nodes + cgraph->n_leafs);
+        graph->props.resize(cgraph->n_nodes + cgraph->n_leafs);
     }
 
     // Loop over nodes in GGML graph to determine if CUDA graph update is required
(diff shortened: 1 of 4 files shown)


DocShotgun:
Interesting. Does this also apply to manual `-ot` assignment of routed expert layers to CPU, or is it required to use `--n-cpu-moe`?

am17an:
It should work with `ot` as well (although I didn't test that)

jeffbolznv:
Is this just keyed by a pointer to a node? What happens if the pointer is freed and reused?

am17an:
Yes, that's ggml_tensor *. My understanding is that those pointers don't change during the course of a program. The underlying data and shapes can change

jeffbolznv:
I'm pretty sure you can allocate tensors and then reuse the memory to make new tensors and a new graph. It's even pretty likely to reuse the same pointers given how the tensors are allocated out of a ggml_context.

ggerganov:
Yeah, we don't have a good way atm to identify graphs robustly, atm.

Btw, although the node pointer is not robust and can fail as @jeffbolznv noted, do I understand correctly that we would still going to perform the "node properties" comparison. So even if the node pointer match is false-positive, the detailed "node properties" check would catch that?

am17an:
Yeah, we still do that check for node properties. If the graph is exactly the same if the node pointer is recycled, then the saved CUDA graph can still be used. I think the only downside would be if the pointer has changed but the graph remains the same, in that case we could have re-used the graph but would create a new one. So I don't see a correctness issue

Panchovix:
I wonder, does this change increases performance when using multiGPU + offloading to CPU?

am17an:
@Panchovix yes I tried with 3 devices

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3806,7 +3818,14 @@ static void ggml_backend_cuda_event_wait(ggml_backend_t backend, ggml_backend_ev
...
 
-    const bool use_cuda_graph = ggml_cuda_graph_set_enabled(cuda_ctx);
+#ifdef USE_CUDA_GRAPH
+    const void * graph_key = ggml_cuda_graph_get_key(cgraph);
+    const bool use_cuda_graph = ggml_cuda_graph_set_enabled(cuda_ctx, graph_key);
+#else
+    const bool use_cuda_graph = false;
+    GGML_UNUSED(cuda_ctx);
+    GGML_UNUSED(cgraph);
+#endif
```
JohannesGaessler, line 3828:
```suggestion
#endif // USE_CUDA_GRAPH
```

JohannesGaessler (approved):
Based on static code analysis I agree with Aman regarding that these changes are logically correct. As of right now we have a single CUDA graph for which we check compatibility with the current `ggml_cgraph` and re-use it if possible. With this PR we store one CUDA graph per first node pointer and re-use that one if possible.

ORippler:
> Yeah, we don't have a good way atm to identify graphs robustly, atm.
> Btw, although the node pointer is not robust and can fail as @jeffbolznv noted, do I understand correctly that we would still going to perform the "node properties" comparison. So even if the node pointer match is false-positive, the detailed "node properties" check would catch that?

> Based on static code analysis I agree with Aman regarding that these changes are logically correct. As of right now we have a single CUDA graph for which we check compatibility with the current ggml_cgraph and re-use it if possible. 

Afaik we are interested in exposing `ggml_cgraph` reuse to the backends in the future (was planned to be part of graphplan API). In such cases, we would skip node properties comparison in the backend and rely on validity of node pointer matching. In such a use-case we could get semantically difficult-to-catch bugs.

am17an:
Since this code touches cuda graphs, multi-gpu and tensor offload code, none of which are tested in the CI - I ran this through my own AWS runner which tests logprobs with different scenarios.
https://github.com/am17an/llama.cpp/actions/runs/21310867142 and the results are fine. Hopefully no breakages re tensor offload or multi-gpu. Running this test is kind of expensive though (because AWS only gives on-demand instances to me instead of spot) and I will only run it for stuff like this

ggml/src/ggml-cuda/ggml-cuda.cu
am17an: Sorry I missed applying this suggestion. Will do it in a follow up PR

Panchovix:
Just to confirm the improvement. On DeepSeek V3 0324, with 7 GPUs (2x4090,3x5090,A6000 and A40) and 70GB RAM on a 9900X, before this commit:

```
prompt eval time =   10301.69 ms /  4325 tokens (    2.38 ms per token,   419.83 tokens per second)
       eval time =   32475.95 ms /   443 tokens (   73.31 ms per token,    13.64 tokens per second)
```

After this commit

```
prompt eval time =   10085.26 ms /  4325 tokens (    2.33 ms per token,   428.84 tokens per second)
       eval time =   35617.65 ms /   510 tokens (   69.84 ms per token,    14.32 tokens per second)
```

So about 2.1% better perf on PP and 5% on TG.

PR #18936, opened 2026-01-19
support Glm4MoeLite

ddh0:
Support for `Glm4MoeLiteForCausalLM`, which seems to be just a renamed version of DeepseekV3 with some code moved around. Thanks to @ngxson for the help.

ref:
- 🤗 [zai-org/GLM-4.7-Flash](https://huggingface.co/zai-org/GLM-4.7-Flash)
- 🤗 [config.json](https://huggingface.co/zai-org/GLM-4.7-Flash/raw/main/config.json)
- 🤗 [modular](https://github.com/huggingface/transformers/blob/main/src/transformers/models/glm4_moe_lite/modular_glm4_moe_lite.py)

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -1078,6 +1078,9 @@ class TextModel(ModelBase):
         if chkhsh == "b3d1dd861f1d4c5c0d2569ce36baf3f90fe8a102db3de50dd71ff860d91be3df":
             # ref: https://huggingface.co/aari1995/German_Semantic_V3
             res = "jina-v2-de"
+        if chkhsh == "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267":
+            # ref: https://huggingface.co/zai-org/GLM-4.7-Flash
+            res = "glm4"
         if chkhsh == "0ef9807a4087ebef797fc749390439009c3b9eda9ad1a097abbe738f486c01e5":
             # ref: https://huggingface.co/meta-llama/Meta-Llama-3-8B
             res = "llama-bpe"
@@ -1255,9 +1258,6 @@ class TextModel(ModelBase):
         if chkhsh == "6c81ce329e0802883b22eabab0d3fa48357337ef1ecb45443828bf1f6254833f":
             # ref: https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B
             res = "exaone-moe"
-        if chkhsh == "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267":
-            # ref: https://huggingface.co/zai-org/GLM-4.7-Flash
-            res = "glm4"
 
         if res is None:
             logger.warning("\n")
@@ -7462,7 +7462,7 @@ class DeepseekModel(TextModel):
     "KimiVLForConditionalGeneration",
     "YoutuForCausalLM",
     "YoutuVLForConditionalGeneration",
-    "Glm4MoeLiteForCausalLM"
+    "Glm4MoeLiteForCausalLM",
 )
 class DeepseekV2Model(TextModel):
     model_arch = gguf.MODEL_ARCH.DEEPSEEK2
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -7462,7 +7462,6 @@ class DeepseekModel(TextModel):
     "KimiVLForConditionalGeneration",
     "YoutuForCausalLM",
     "YoutuVLForConditionalGeneration",
-    "Glm4MoeLiteForCausalLM",
 )
 class DeepseekV2Model(TextModel):
     model_arch = gguf.MODEL_ARCH.DEEPSEEK2
@@ -8448,6 +8447,32 @@ class Glm4MoeModel(TextModel):
             experts = [k for d in self._experts for k in d.keys()]
             if len(experts) > 0:
                 raise ValueError(f"Unprocessed experts: {experts}")
+
+
+@ModelBase.register("Glm4MoeLiteForCausalLM")
+class Glm4MoeLiteModel(DeepseekV2Model):
+    model_arch = gguf.MODEL_ARCH.DEEPSEEK2
+
+    # copied from Glm4MoeModel
+    def set_vocab(self):
+        from transformers import AutoTokenizer
+
+        tokenizer = AutoTokenizer.from_pretrained(self.dir_model)
+        special_vocab = gguf.SpecialVocab(self.dir_model, load_merges=True)
+        tokens, toktypes, tokpre = self.get_vocab_base()
+        self.gguf_writer.add_tokenizer_model("gpt2")
+        self.gguf_writer.add_tokenizer_pre(tokpre)
+        self.gguf_writer.add_token_list(tokens)
+        self.gguf_writer.add_token_types(toktypes)
+
+        # Special tokens
+        # Note: Using <|endoftext|> (151329) for eot causes endless generation
+        special_vocab._set_special_token("bos", tokenizer.get_added_vocab()["[gMASK]"])  # 151331
+        special_vocab._set_special_token("eot", tokenizer.get_added_vocab()["<|user|>"])  # 151336
+        special_vocab._set_special_token("unk", tokenizer.get_added_vocab()["<|endoftext|>"]) # 151329
+        special_vocab._set_special_token("eom", tokenizer.get_added_vocab()["<|observation|>"])  # 151338
+
+        special_vocab.add_to_gguf(self.gguf_writer)
 
 
 @ModelBase.register("GlmForCausalLM", "ChatGLMModel", "ChatGLMForConditionalGeneration")
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -1257,7 +1257,7 @@ class TextModel(ModelBase):
             res = "exaone-moe"
         if chkhsh == "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267":
             # ref: https://huggingface.co/zai-org/GLM-4.7-Flash
-            res = "glm-4.7-flash"
+            res = "glm4"
 
         if res is None:
             logger.warning("\n")
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -1255,6 +1255,9 @@ class TextModel(ModelBase):
         if chkhsh == "6c81ce329e0802883b22eabab0d3fa48357337ef1ecb45443828bf1f6254833f":
             # ref: https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B
             res = "exaone-moe"
+        if chkhsh == "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267":
+            # ref: https://huggingface.co/zai-org/GLM-4.7-Flash
+            res = "glm-4.7-flash"
 
         if res is None:
             logger.warning("\n")
@@ -7458,7 +7461,8 @@ class DeepseekModel(TextModel):
     "DeepseekV3ForCausalLM",
     "KimiVLForConditionalGeneration",
     "YoutuForCausalLM",
-    "YoutuVLForConditionalGeneration"
+    "YoutuVLForConditionalGeneration",
+    "Glm4MoeLiteForCausalLM"
 )
 class DeepseekV2Model(TextModel):
     model_arch = gguf.MODEL_ARCH.DEEPSEEK2
--- a/convert_hf_to_gguf_update.py
+++ b/convert_hf_to_gguf_update.py
@@ -170,7 +170,7 @@ pre_computed_hashes = [
     {"name": "grok-2",    "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/alvarobartt/grok-2-tokenizer", "chkhsh": "66b8d4e19ab16c3bfd89bce5d785fb7e0155e8648708a1f42077cb9fe002c273"},
     # jina-v2-de variants
     {"name": "jina-v2-de", "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/aari1995/German_Semantic_V3", "chkhsh": "b3d1dd861f1d4c5c0d2569ce36baf3f90fe8a102db3de50dd71ff860d91be3df"},
-    {"name": "glm-4.7-flash", "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/zai-org/GLM-4.7-Flash", "chkhsh": "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267"},
+    {"name": "glm4", "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/zai-org/GLM-4.7-Flash", "chkhsh": "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267"},
 ]
 
 def download_file_with_auth(url, token, save_path):
--- a/convert_hf_to_gguf_update.py
+++ b/convert_hf_to_gguf_update.py
@@ -173,6 +173,7 @@ pre_computed_hashes = [
     {"name": "glm4", "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/zai-org/GLM-4.7-Flash", "chkhsh": "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267"},
 ]
 
+
 def download_file_with_auth(url, token, save_path):
     headers = {"Authorization": f"Bearer {token}"} if token else None
     response = sess.get(url, headers=headers)
--- a/convert_hf_to_gguf_update.py
+++ b/convert_hf_to_gguf_update.py
@@ -148,7 +148,6 @@ models = [
     {"name": "youtu",            "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/tencent/Youtu-LLM-2B", },
     {"name": "solar-open",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/upstage/Solar-Open-100B", },
     {"name": "exaone-moe",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B", },
-    {"name": "glm-4.7-flash",    "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/zai-org/GLM-4.7-Flash", },
 ]
 
 # some models are known to be broken upstream, so we will skip them as exceptions
@@ -171,8 +170,8 @@ pre_computed_hashes = [
     {"name": "grok-2",    "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/alvarobartt/grok-2-tokenizer", "chkhsh": "66b8d4e19ab16c3bfd89bce5d785fb7e0155e8648708a1f42077cb9fe002c273"},
     # jina-v2-de variants
     {"name": "jina-v2-de", "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/aari1995/German_Semantic_V3", "chkhsh": "b3d1dd861f1d4c5c0d2569ce36baf3f90fe8a102db3de50dd71ff860d91be3df"},
+    {"name": "glm-4.7-flash", "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/zai-org/GLM-4.7-Flash", "chkhsh": "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267"},
 ]
-
 
 def download_file_with_auth(url, token, save_path):
     headers = {"Authorization": f"Bearer {token}"} if token else None
--- a/convert_hf_to_gguf_update.py
+++ b/convert_hf_to_gguf_update.py
@@ -148,6 +148,7 @@ models = [
     {"name": "youtu",            "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/tencent/Youtu-LLM-2B", },
     {"name": "solar-open",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/upstage/Solar-Open-100B", },
     {"name": "exaone-moe",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B", },
+    {"name": "glm-4.7-flash",    "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/zai-org/GLM-4.7-Flash", },
 ]
 
 # some models are known to be broken upstream, so we will skip them as exceptions
--- a/src/llama-vocab.cpp
+++ b/src/llama-vocab.cpp
@@ -2423,6 +2423,7 @@ void llama_vocab::impl::load(llama_model_loader & ml, const LLM_KV & kv) {
                     || t.first == "_<EOT>"
                     || t.first == "<|end_of_text|>"
                     || t.first == "<end_of_utterance>" // smoldocling
+                    || t.first == "<|user|>" // glm-4.7-lite
                ) {
                 special_eog_ids.insert(t.second);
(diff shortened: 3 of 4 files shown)


ddh0:
Would close #18931

convert_hf_to_gguf_update.py
```diff
@@ -148,6 +148,7 @@ class TOKENIZER_TYPE(IntEnum):
     {"name": "youtu",            "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/tencent/Youtu-LLM-2B", },
     {"name": "solar-open",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/upstage/Solar-Open-100B", },
     {"name": "exaone-moe",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B", },
+    {"name": "glm-4.7-flash",    "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/zai-org/GLM-4.7-Flash", },
```
ngxson: this should be placed in `pre_computed_hashes`, like the other `glm4` tokenizers

also, we technically don't have to change the name `glm4` as it will requires C++ code changes. but I'm not quite sure if this tokenizer is different from other GLM 4.x models

ngxson: it seems like the same tokenizer as GLM-4.7 (the bigger variant), minus some special tokens:
- https://huggingface.co/zai-org/GLM-4.7-Flash/blob/main/tokenizer_config.json
- https://huggingface.co/zai-org/GLM-4.7/blob/main/tokenizer_config.json

so seems to be fine to reuse `glm4` tokenizer

fizzAI:
its actually a renamed version of GLM4Moe with DeepseekV3Attention (uses MLA) and an added dense expert at the start, but i'm not sure if any of that is relevant here

convert_hf_to_gguf_update.py
ddh0: Thanks, I moved it there in [38e4882](https://github.com/ggml-org/llama.cpp/pull/18936/commits/38e4882469a57eab0d022b1ae1de7bdb33eec4be). The pre-tokenizer `chkhsh` is different from any of the others so I added a new pre-tokenizer for it. Going to re-convert and try again now.

ngxson:
> with DeepseekV3Attention (uses MLA) and an added dense expert at the start

deepseekv3 also have some dense layer at the start, so unless I missed something, it seems like just deepseek renamed (not GLM4Moe renamed)

fizzAI:
oh i didn't know mb, i guess just the logic for picking dense layers is different. in that case then yeah from the GGUF perspective it's just renamed deepseek

ddh0:
Freshly converted, imatrix'd from Q8_0, and quantized to IQ4_XS. Output is coherent for multi-turn. PPL for imatrix is a bit high at ~15 - 17 but I think this is correct now. Should be ready for review.

ubergarm:
> Output is coherent for multi-turn. PPL for imatrix is a bit high at ~15 - 17 but I think this is correct now. Should be ready for review.

Thanks for sharing this quickly! I can confirm that I converted the safetensors to bf16 GGUF and it runs on 2x RTX A6000's with coherent multi-turn chat.

imatrix with [my usual corpus](https://gist.github.com/ubergarm/edfeb3ff9c6ec8b49e88cdf627b0711a) on the bf16 ended up at `Final estimate: PPL = 7.1406 +/- 0.04441`, and when running your https://huggingface.co/ddh0/imatrices/raw/main/ddh0_imat_calibration_data_v2.txt on the bf16 i get comparable to your q8_0 results: `Final estimate: PPL = 15.5197 +/- 0.33298`

<details>

<summary>👈 convert command </summary>

```bash
$ python \
    convert_hf_to_gguf.py \
    --outtype bf16 \
    --split-max-size 50G \
    --outfile /mnt/raid/models/ubergarm/GLM-4.7-Flash-GGUF/ \
    /mnt/raid/models/zai-org/GLM-4.7-Flash/
```

</details>

<details>

<summary>👈 imatrix command </summary>

```bash
model=/mnt/raid/models/ubergarm/GLM-4.7-Flash-GGUF/GLM-4.7-Flash-64x2.6B-BF16-00001-of-00002.gguf

    #-f ddh0_imat_calibration_data_v2.txt \

./build/bin/llama-imatrix \
    --model "$model"\
    -f ubergarm-imatrix-calibration-corpus-v02.txt \
    -o /mnt/raid/models/ubergarm/GLM-4.7-Flash-GGUF/imatrix-GLM-4.7-Flash-BF16.dat \
    --ctx-size 512 \
    -ub 4096 -b 4096 \
    -fit off \
    -ngl 99 \
    -ts 40,48 \
    --threads 1 \
    --no-mmap \
    --output-format dat
```

</details>

Only odd thing I saw was this warning disabling flash attention even with `-fit off` hrmm:

```
llama_context: layer 0 is assigned to device CUDA0 but the Flash Attention tensor is assigned to device CPU (usually due to missing support)
llama_context: Flash Attention was auto, set to disabled
```

ddh0:
I assume flash attention being disabled is due to MLA? Not sure

convert_hf_to_gguf.py
```diff
@@ -1255,6 +1255,9 @@ def get_vocab_base_pre(self, tokenizer) -> str:
         if chkhsh == "6c81ce329e0802883b22eabab0d3fa48357337ef1ecb45443828bf1f6254833f":
             # ref: https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B
             res = "exaone-moe"
+        if chkhsh == "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267":
+            # ref: https://huggingface.co/zai-org/GLM-4.7-Flash
+            res = "glm4"
```
CISC: This is placed wrong, run `convert_hf_to_gguf_update.py` again.

convert_hf_to_gguf.py
```diff
@@ -7458,7 +7461,8 @@ def prepare_tensors(self):
     "DeepseekV3ForCausalLM",
     "KimiVLForConditionalGeneration",
     "YoutuForCausalLM",
-    "YoutuVLForConditionalGeneration"
+    "YoutuVLForConditionalGeneration",
+    "Glm4MoeLiteForCausalLM"
```
```suggestion
    "Glm4MoeLiteForCausalLM",
```

ngxson (approved):
Nice 🚀 (can be merged once @CISC approves)

CISC (approved):

ddh0:
I don't have permissions to merge it myself, can @ngxson or @CISC hit the button for me please?

Thanks both of you for your help! It's easy to add new models when they're just renamed existing models :)

CISC:
> I don't have permissions to merge it myself, can @ngxson or @CISC hit the button for me please?

Will do once CI finishes.

BTW, we should consider including MTP layer in DeepSeek models soon.

EvilFreelancer:
I suspect glm4 tokenizer mismatch in GLM-4.7-Flash, cause its vocab is 154,856 vs GLM-4.5-Air vocab 151,365, i.e. +3,491 tokens and non-matching some special token IDs.

The GLM-4.7-Flash model converted under the glm4 identifier on infenrence stage goes into infinite generation (likely EOS/stop IDs don’t match, so stop conditions never trigger) and also doesn’t emit/recognize the <think> tag, which I guess points to incorrect special-token mapping.

Probably need some additional testing.

ddh0:
The `glm4` is only the pre-tokenizer, the actual tokenizer will work just like any other model, as far as I understand. LMK if you find a concrete issue.

Aaryan-Kapoor:
This might need additional testing.

Following up with @EvilFreelancer's concern - this might be a real bug. See: https://huggingface.co/ngxson/GLM-4.7-Flash-GGUF/discussions/2

GLM-4.7-Flash's <|user|> token is declared as EOS in the original tokenizer config (to stop generation at end of assistant turn), but the GGUF only sets <|endoftext|> as EOS. The model doesn't stop at <|user|>, so it generates past its turn and loops into self-conversation:

```
Wait - looking at `User` (the first line of my internal thought block)...
Ah, I see. In this specific simulation/turn-based interface:
1. User says "hello"
2. Model responds with weirdness
3. **Current Turn**: The user is sending the exact same command to a new model instance or just repeating it because they didn't get an answer?
```

ddh0:
> This might need additional testing.
> 
> Following up with @EvilFreelancer's concern - this might be a real bug. See: https://huggingface.co/ngxson/GLM-4.7-Flash-GGUF/discussions/2
> 
> GLM-4.7-Flash's <|user|> token is declared as EOS in the original tokenizer config (to stop generation at end of assistant turn), but the GGUF only sets <|endoftext|> as EOS. The model doesn't stop at <|user|>, so it generates past its turn and loops into self-conversation:
> 
> ```
> Wait - looking at `User` (the first line of my internal thought block)...
> Ah, I see. In this specific simulation/turn-based interface:
> 1. User says "hello"
> 2. Model responds with weirdness
> 3. **Current Turn**: The user is sending the exact same command to a new model instance or just repeating it because they didn't get an answer?
> ```

Does the issue occur with GGUFs generated with this PR? I did not upload those other ones, and cannot vouch for them.

EvilFreelancer:
@ddh0 you may try https://huggingface.co/ngxson/GLM-4.7-Flash-GGUF

ddh0:
> @ddh0 you may try https://huggingface.co/ngxson/GLM-4.7-Flash-GGUF

You misunderstand. I'm saying that you should not test with those GGUFs since they were not made with this PR. The issue might be related, or it might not. I would like to narrow it down, so please try with GGUFs generated with this PR.

ddh0:
@ngxson - PTAL, it's possible I am missing something

DocShotgun:
> I assume flash attention being disabled is due to MLA? Not sure

Odd, because this doesn't happen when running big Deepseek with MLA. Perhaps there aren't CUDA flash MLA kernels configured for these tensor dimensions or something?

noctrex:
> This might need additional testing.
> 
> Following up with @EvilFreelancer's concern - this might be a real bug. See: https://huggingface.co/ngxson/GLM-4.7-Flash-GGUF/discussions/2
> 
> GLM-4.7-Flash's <|user|> token is declared as EOS in the original tokenizer config (to stop generation at end of assistant turn), but the GGUF only sets <|endoftext|> as EOS. The model doesn't stop at <|user|>, so it generates past its turn and loops into self-conversation:
> 
> ```
> Wait - looking at `User` (the first line of my internal thought block)...
> Ah, I see. In this specific simulation/turn-based interface:
> 1. User says "hello"
> 2. Model responds with weirdness
> 3. **Current Turn**: The user is sending the exact same command to a new model instance or just repeating it because they didn't get an answer?
> ```

Created a GGUF with this PR, and the same happens to me. It never stops, and loops itself

Aaryan-Kapoor:
Can someone with the GGUF try:
```
./llama-cli -m GLM-4.7-Flash.gguf --jinja -p "hello" -n 512 \
  --reverse-prompt "<|user|>" \
  --reverse-prompt "<|observation|>"
```
@noctrex

I'm converting one myself right now

EvilFreelancer:
@ddh0 I applied the fix on my side, then converted the model to GGUF and re-tested. The runtime behavior matches what’s observed with the Hugging Face GGUF release here: [ngxson/GLM-4.7-Flash-GGUF](https://huggingface.co/ngxson/GLM-4.7-Flash-GGUF) (couple examples of response in discussions).

Tested with command like this, on llama.cpp from commit `3d55846a5c626e2e608db8c24fa9ee6defaacca9`.
```
./build/bin/llama-cli -m ./models-local/zai-org/fp16.gguf
```

Couple tests [here](https://gist.github.com/EvilFreelancer/c449055168939ec15f966f131ea5eade).

EvilFreelancer:
@Aaryan-Kapoor see [here](https://gist.github.com/EvilFreelancer/c449055168939ec15f966f131ea5eade#file-test3-txt).

ddh0:
Thanks guys. It seems like something is definitely wrong then. I will mark this back as draft and take a closer look over the next few hours.

noctrex:
> Can someone with the GGUF try:
> 
> ```
> ./llama-cli -m GLM-4.7-Flash.gguf --jinja -p "hello" -n 512 \
>   --reverse-prompt "<|user|>" \
>   --reverse-prompt "<|observation|>"
> ```
> 
> @noctrex
> 
> I'm converting one myself right now

Loop still happening unfortunately.

ngxson:
it should work with latest commit, but reasoning parsing seems to be a bit buggy

image

noctrex:
> it should work with latest commit, but reasoning parsing seems to be a bit buggy
> 
> image

Yes, with the addition to llama-vocab.cpp, I can now converse with it. But as you said it dumps all the thinking also in the conversation. Maybe a template fix needed.

ddh0:
Latest commit also adds `<|observation|>` to EOGs (since it is specified [here](https://huggingface.co/zai-org/GLM-4.7-Flash/blob/main/config.json#L8))

Aaryan-Kapoor:
Thanks for the quick fix @ngxson! Adding `<|user|>` as EOG should address the infinite generation issue.

A couple notes:

1. **Missing `<|observation|>`** - GLM-4.7-Flash's config.json declares 3 EOG tokens:
   ```json
   "eos_token_id": [154820, 154827, 154829]
   ```
   - 154820: `<|endoftext|>` 
   - 154827: `<|user|>`
   - 154829: `<|observation|>`

   Should `<|observation|>` be added too?

2. **Alternative approach** - Rather than hardcoding in `llama-vocab.cpp`, could the conversion script read the `eos_token_id` array from config.json and set all as EOG in the GGUF metadata? That way other inference engines would also pick them up correctly.

ngxson:
The reasoning problem was because the template force adding `<think>`:

```
    <|assistant|>{{- '</think>' if (enable_thinking is defined and not enable_thinking) else '<think>' -}}
```

Quite strange because GLM-4.7 should also have the same problem: https://huggingface.co/zai-org/GLM-4.7/blob/main/chat_template.jinja

Aaryan-Kapoor:
@ddh0 thanks for the commit!

ddh0:
I will wait before marking this PR back as ready for review until it is confirmed that it's working as expected. I will also try it out myself locally with the latest changes.

ngxson:
I'm a bit doubt if the change in `llama-vocab` is actually correct. Looking at `chat.cpp`, the logic for handling this template is already there, not quite sure why it's not correct for GLM 4.7 Flash

ddh0:
I agree, I don't understand why it would be necessary

ngxson:
ohh I see, because the `DeepseekV2Model` conversion class doesn't have the special vocab logic

ngxson:
sorry, need to regenerate gguf now

arch-btw:
Thanks @ngxson. 

Possible stupid question but could part of it be because \<eop\> isn't being used?
Because the Jinja template starts with \<sop\> but then the seemingly related \<eop\> is never used to close the prompt?

``
[gMASK]<sop>
``

https://huggingface.co/zai-org/GLM-4.7-Flash/blob/main/tokenizer_config.json#L35-L49

ngxson:
does the chat template use `<eos>` ?

arch-btw:
@ngxson I don't see it in the Jinja.

ngxson:
then it should be irrelevant if chat template doesn't use it

ngxson:
the new GGUF should works:

image

noctrex:
Confirm, it works now.

ddh0:
Confirmed working with the latest changes (though, it was also working fine for me before). Marked as ready for review again.

EvilFreelancer:
Confirm, let's merge )

CISC:
:100: Nice work everyone! :tada:

ddh0:
> Odd, because this doesn't happen when running big Deepseek with MLA. Perhaps there aren't CUDA flash MLA kernels configured for these tensor dimensions or something?

~~For the record @DocShotgun, flash attention is working fine for me. Not sure what going on earlier but seems OK now. Feel free to open a new issue and link it here if there are still problems.~~

Nevermind, it does seem that some operations are not supported on CUDA... this problem is above my pay-grade at this point, it will need to be addressed in another PR :(

ggerganov:
Quick question about [interleaved/preserved thinking](https://docs.z.ai/guides/capabilities/thinking-mode):
If I pass a kwarg to the chat template `“clear_thinking”: false`, would this work to enable "preserved" thinking? Do we need changes to the client also? If yes, should we support this in the `llama-server` WebUI?

cc @ngxson @aldehir @pwilkin

ngxson:
@ggerganov there was a discussion around this: https://github.com/ggml-org/llama.cpp/discussions/18368

webui change is needed because the `reasoning_content` must be put back into the message object, as a dedicated field.

but as mentioned in the discussion, it can be a bit tricky as not all models support this. we can probably extend the `/props` to reflect if the model support putting back the reasoning into message, and show the option "Preserve reasoning" in webui

ggerganov:
@ngxson Btw, regarding https://github.com/ggml-org/llama.cpp/pull/18936#issuecomment-3770096320, I think there is still an issue. Consider this short conversation consisting of 2 messages:

image

When I dump the last prompt from the second message, I see this:

```bash
curl http://127.0.0.1:8013/slots?model="glm-4.7-flash-hf" | jq .[2]

{
  ...
  "prompt": "[gMASK]<sop><|user|>Say \"hi\"<|assistant|></think>Hello! How can I help you today?<|user|>Say \"bye\"<|assistant|><think>",
  ...                                                     ^ problem here?
}
```

Notice that there isn't a corresponding `<think>` token for the assistant response of the first message. This look strange - not sure if it is necessarily wrong, but to me it looks bogus.

ngxson:
It's actually a feature of the [chat template](https://huggingface.co/zai-org/GLM-4.7-Flash/blob/main/chat_template.jinja) (I know that looks strange, but I was surprised too):

```
{%- if add_generation_prompt -%}
    <|assistant|>{{- '</think>' if (enable_thinking is defined and not enable_thinking) else '<think>' -}}
{%- endif -%}
```

For ref, chat template of [GLM 4.6](https://huggingface.co/zai-org/GLM-4.6/blob/main/chat_template.jinja) doesn't do that: 

```
{%- if add_generation_prompt -%}
    <|assistant|>{{- '\n<think></think>' if (enable_thinking is defined and not enable_thinking) else '' -}}
{%- endif -%}
```

ngxson:
Note that for messages that are not the last one, they also skip the opening tag `<think>`:

```
{%- if ((clear_thinking is defined and not clear_thinking) or loop.index0 > ns.last_user_index) and reasoning_content -%}
{{ '<think>' + reasoning_content.strip() +  '</think>'}}
{%- else -%}
{{ '</think>' }}
{%- endif -%}
```

ggerganov:
> Note that for messages that are not the last one, they also skip the opening tag <think>:

Hm yes, looks like it is intentional. Weird.

Anyway, I encountered a situation where during a longer conversation the model didn't close it's reasoning with `</think>` before generating the final response, so I started looking why that could be. Thought this might be related, but maybe there is something else going on. Thanks.

noctrex:
Seems to be quantization sensitive. I created this quant and it seems to perform great:
[https://huggingface.co/noctrex/GLM-4.7-Flash-MXFP4_MOE-GGUF](https://huggingface.co/noctrex/GLM-4.7-Flash-MXFP4_MOE-GGUF)
I even use it in Kilo Code and it works. Used the prompt:
`write python code to import a csv, get the data type for each column using low memory = false, for fields that contain text summarize the number of records per value, for numbers give the average min and max, don't do anything with dates or other fields`
And it created the script, a test file, a readme, and tested it, and it actually works.

image
image

ngxson:
@ggerganov Thanks for looking into this. Indeed, I couldn't test compare logprobs with vLLM because the model hasn't been properly supported by vLLM (would appreciate if someone can do)

I compared line-by-line the HF implementation of [this](https://github.com/huggingface/transformers/blob/main/src/transformers/models/glm4_moe_lite/modeling_glm4_moe_lite.py) vs [deepseek_v3](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v3/modeling_deepseek_v3.py) but so far I couldn't find any differences that could affect the result.

ggerganov:
Yeah, we should definitely check against a reference implementation to make sure. Overall it performs quite well so far - been running it today for a while. But this was one-off surprise where the reasoning was not properly closed. I would guess it's some tokenizer/template issue, but will do more experiments to see if I can narrow it down.

@noctrex I am using the full precision so it's not a quantization issue.

noctrex:
@ggerganov Weird, as I've seen reports of users that try out the Q4 and below quants reporting strange issues, such as starting to think in chinese or thinking endlessly, and just returning gibberish answers.
Anyway, Thanks for your great work!

pwilkin:
@ggerganov you can use the autoparser branch, it has support for reasoning and tool calling for this one properly already :>

ngxson:
Ok so thanks to @bartowski1182 I got the logprobs from vllm vs F16 on llama.cpp. There are indeed some differences, and I just realized that since the prompt is repeated in `compare_logprobs.py`, it makes the long context tokens pretty much predictable (the script will need to be fixed)

Looking deeper into the code now..

| idx  | logits_llama.log | logprob_1 | logits_other.log | logprob_2 | diff (abs) |
|------|------------------|-----------|------------------|-----------|------------|
| 1    | 'S'              | -2.6846   | '\\'             | -2.5100   | 0.1746     |
| 2    | ' here'          | -0.8800   | ' here'          | -1.6765   | 0.7964     |
| 3    | ' AI'            | -0.6684   | ' AI'            | -0.9069   | 0.2385     |
| 4    | ' AI'            | -0.5882   | ' AI'            | -0.9261   | 0.3379     |
| 5    | ' assistant'     | -0.4866   | ' assistant'     | -0.6661   | 0.1795     |
| 6    | ' designed'      | -1.1536   | ' designed'      | -1.3800   | 0.2264     |
| 7    | ' of'            | -0.0046   | ' of'            | -0.0041   | 0.0004     |
| 8    | 'S'              | -2.6461   | '\\'             | -2.7089   | 0.0628     |
| 9    | ' tools'         | -2.0654   | ' the'           | -1.9421   | 0.1233     |
| 10   | ' to'            | -1.0288   | '.'              | -1.0014   | 0.0274     |
| 1011 | ' you'           | -0.0000   | ' you'           | -0.0000   | 0.0000     |
| 1012 | ' need'          | -0.0000   | ' need'          | -0.0000   | 0.0000     |
| 1013 | ' to'            | -0.0001   | ' to'            | -0.0000   | 0.0001     |
| 1014 | '\n'             | -2.1092   | ' use'           | -0.0000   | 2.1092     |
| 1015 | ' a'             | -0.0000   | ' a'             | -0.0000   | 0.0000     |
| 1016 | ' tool'          | -0.0000   | ' tool'          | -0.0000   | 0.0000     |
| 1017 | ' output'        | -0.0000   | ' output'        | -0.0000   | 0.0000     |
| 1018 | ' the'           | -0.0000   | ' the'           | -0.0000   | 0.0000     |
| 1019 | ' call'          | -0.0000   | ' call'          | -0.0000   | 0.0000     |
| 1020 | ' in'            | -0.0000   | ' in'            | -0.0000   | 0.0000     |
| 5021 | ' requires'      | -0.0000   | ' requires'      | -0.0000   | 0.0000     |
| 5022 | ' external'      | -0.0000   | ' external'      | -0.0000   | 0.0000     |
| 5023 | ' data'          | -0.0000   | ' data'          | -0.0000   | 0.0000     |
| 5024 | ' computation'   | -0.0000   | ' computation'   | -0.0000   | 0.0000     |
| 5025 | ' or'            | -0.0000   | ' or'            | -0.0000   | 0.0000     |
| 5026 | ' actions'       | -0.0000   | ' actions'       | -0.0000   | 0.0000     |
| 5027 | ' beyond'        | -0.0000   | ' beyond'        | -0.0000   | 0.0000     |
| 5028 | ' your'          | -0.0000   | ' your'          | -0.0000   | 0.0000     |
| 5029 | ' internal'      | -0.0000   | ' internal'      | -0.0000   | 0.0000     |
| 5030 | ' knowledge'     | -0.0000   | ' knowledge'     | -0.0000   | 0.0000     |

aldehir:
> but as mentioned in the discussion, it can be a bit tricky as not all models support this. we can probably extend the /props to reflect if the model support putting back the reasoning into message, and show the option "Preserve reasoning" in webui

An option based off `/props` is certainly a cleaner approach, but I also don't think there's any harm in adding the reasoning content back. If a template doesn't use it, it'll simply be ignored.

@ggerganov Interleaved thinking isn't leveraged in basic chat. This check will prune any reasoning prior to the last user message, and in a chat session that's guaranteed to always prune:
```
loop.index0 > ns.last_user_index
```

It's only leveraged during tool calling. Something to consider when implementing MCP support.

PR #18937, opened 2026-01-19
server : refactor oai_parser_opt, move it to server_chat_params

ngxson:
In this PR:
- Rename `oaicompat_parser_options` --> `server_chat_params`
- Store `common_chat_templates_ptr` inside it
- `ctx_server.get_meta()` now returns `common_chat_templates_ptr`
- For CLI: chat template is now applied in `cli.cpp`; this should allow defining tools from CLI in the future

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -534,8 +534,7 @@ public:
     server_queue    queue_tasks;
     server_response queue_results;
 
-    common_chat_templates_ptr chat_templates;
-    oaicompat_parser_options  oai_parser_opt;
+    server_chat_params chat_params;
 
     ~server_context_impl() {
         if (!sleeping) {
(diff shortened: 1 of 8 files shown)


tools/server/server-context.cpp
```diff
@@ -907,6 +874,42 @@ struct server_context_impl {
...
+
+            // thinking is enabled if:
+            // 1. It's not explicitly disabled (reasoning_budget == 0)
+            // 2. The chat template supports it
+            const bool enable_thinking = params_base.use_jinja && params_base.reasoning_budget != 0 && common_chat_templates_support_enable_thinking(chat_templates.get());
+            SRV_INF("thinking = %d\n", enable_thinking);
+
+            // print sample chat example to make it clear which template is used
+            // @ngxson modern templates are too long, spam the logs; printing the example is enough
+            LOG_INF("%s: chat template, example_format: '%s'\n", __func__,
```
pwilkin: Maybe move this to after line 881 instead of calling example twice? If we fallback to chatml we don't really need to print the example.

pwilkin (approved):
Just a minor one.

tools/server/server-context.cpp
ngxson: nice catch, I removed the chatml fallback logic altogether

PR #13033, opened 2025-04-20
quantize: improve pattern matching for allowed tensors

EAddario:
This PR implements @slaren's regex matching recommendation for allowed tensors. For example: `--tensor-type attn=q4_k` will now apply to all tensors named \*attn\*

PR #12511, opened 2025-03-22
quantize: Handle user-defined quantization levels for additional tensors

EAddario:
This PR adds the ability to quantize other tensors, beyond token-embedding and output-tensor. It handles most of the supported architectures. ~~except Mamba, RWKV6, RWKV6QWEN2 and T5 to avoid having too many command options, but can add as well if maintainers request it.~~

For full background on the PR, please see: [Squeezing Tensor Bits: the quest for smaller LLMs](https://medium.com/@eaddario/squeezing-tensor-bits-the-quest-for-smaller-llms-86b23bd052ca)

PR #18952, opened 2026-01-20
quant : manual overrides of tensor types take precedence

ggerganov:
cont #12511, #13033

When user provides a tensor type with the `--tensor-type` CLI argument, it takes precedence  over the automatic logic. The shape-compatibility check is moved after the new tensor type is determined, so now it covers both the manual and automatic paths.


ggerganov:
cc @EAddario

PR #18955, opened 2026-01-20
jinja : implement mixed type object keys

CISC:
Allow all hashable types as object keys, taking care to replicate special python/jinja behavior between `int`/`float`/`bool`.

Fixed array/object output with `string` filter.

Fixed object `tojson` output (did not properly escape key string).

Fixed object item order when replacing an item.


common/jinja/value.h
```diff
@@ -108,21 +109,75 @@ struct value_t {
...
-        bool is_key_numeric = false;
         std::map<std::string, value> unordered;
-        std::vector<std::pair<std::string, value>> ordered;
-        void insert(const std::string & key, const value & val) {
-            if (unordered.find(key) != unordered.end()) {
-                // if key exists, remove from ordered list
-                ordered.erase(std::remove_if(ordered.begin(), ordered.end(),
-                    [&](const std::pair<std::string, value> & p) { return p.first == key; }),
-                    ordered.end());
+        std::vector<std::tuple<std::string, value, value>> ordered;
```
ngxson: IMO having key type as string here can be quite redundant, we can already check the type via `is_val`

common/jinja/value.h
```diff
@@ -108,21 +109,75 @@ struct value_t {
...
+            const auto is_hashable = [&](const value & val) -> bool {
+                const std::string val_type = val->type();
+                return !key->is_undefined() && val_type != "Array" && val_type != "Object";
+            };
+            if (key_type == "Array") {
+                const auto & vec = key->as_array();
+                if (std::all_of(vec.begin(), vec.end(), is_hashable)) {
+                    key_type = "Tuple";
+                    key_str = key->as_repr();
+                }
```
ngxson: maybe better to implement hash function and `is_hashable` as struct member, same idea as `as_string()`, `as_bool()`, etc

common/jinja/value.h
```diff
@@ -108,21 +109,75 @@ struct value_t {
...
+            } else if (key_type == "Object") {
+                const auto & vec = key->as_ordered_object();
+                if (std::all_of(vec.begin(), vec.end(), [&](auto ikv) -> bool {
+                    return is_hashable(std::get<2>(ikv));
+                })) {
+                    key_type = "NamedTuple";
```
ngxson: I think kwarg is a good alternative to `NamedTuple`

tests/test-jinja.cpp
```diff
@@ -401,6 +401,26 @@ static void test_set_statement(testing & t) {
...
         "1"
     );
+
+    test_template(t, "set dict with mixed type keys",
+        "{% set d = {0: 1, none: 2, 1.0: 3, '0': 4, (0, 0): 5, false: 6, 1: 7} %}{{ d[(0, 0)] + d[0] + d[none] + d['0'] + d[false] + d[1.0] + d[1] }}",
+        json::object(),
+        "37"
+    );
+
+    // TODO: Print tuples correctly
```
ngxson: not sure if this helps, but you can also print the tuple manually, example:

```
{% for k, v in d %}{{ k|string }} -> {{ v|string }}{% endfor %}
```

common/jinja/caps.cpp
```diff
@@ -97,7 +97,7 @@ caps caps_get(jinja::program & prog) {
             return json{nullptr};
         },
         [&](bool, value & messages, value &) {
-            auto & content = messages->at(0)->at("content");
+            auto & content = messages->at(0)->at(mk_val<value_string>("content"));
```
ngxson: I think we can make an overload version `at(const std::string &)`

ngxson (commented):
I'm rethinking about this approach a bit, probably we should drop the `map::unordered`, keep all the data inside `std::vector<std::pair<value, value>> ordered`, and always perform a linear search when we access an object.

Ofc, that will be slower, but realistically a template in the wild never has more than 50 or even 100 keys inside an object, so it's probably fine.

ngxson:
Btw, the `value.h` header is becoming larger. We should move long function definitions into `value.cpp`, while keeping one-liner definition inside the header (for more visibility)

common/jinja/value.h
CISC: It's not the type, it's the unique lookup key, I store it here as well so I don't have to regenerate it excessively when checking if it's the right key.

common/jinja/value.h
CISC: Hmmm, possibly.

tests/test-jinja.cpp
CISC: No, I mean as in `(1, 2, ...)` to match jinja output.

common/jinja/caps.cpp
CISC: I have to forward declare `value_string` to be allowed to do that.

CISC: ..and moving the `at`s into array/object doesn't work either...

common/jinja/value.h
ngxson: I see, but is it possible to rely on `std::hash` instead? Ofc there can be collisions, but we can adapt the logic so that it also check for `==` before returning the object, otherwise continue to next item.

I think an implementation based on `std::hash` should be cleaner overall

common/jinja/caps.cpp
ngxson: > I have to forward declare `value_string` to be allowed to do that.

you can also move the function implementation into cpp code, that way everything is already defined.

> ..and moving the `at`s into array/object doesn't work either...

hmm yeah.. that doesn't allow chained call `->at(...)->at(...)`

another idea: you can also define the function `value_t::at()` to check if the current value is an object or array, then `cast_val` to the correct type and call `at(...)` of the correct type

ngxson: > another idea: you can also define the function `value_t::at()` to check if the current value is an object or array, then `cast_val` to the correct type and call `at(...)` of the correct type

probably don't need that at all: if a `jinja::value` is an object, then `val->at()` will always call the `value_object_t::at()` - same for array type

ngxson: this is what I mean: https://github.com/ggml-org/llama.cpp/compare/master...ngxson:xsn/jinj_at?expand=1

CISC: Yeah, that worked, not sure what I was doing earlier. :confused:

CISC: Ended up having to do forward declaration for array/object anyway, ohwell.

common/jinja/value.h
CISC: Refactored, not sold on going through all items yet though. Any real upside to not having the map?

ngxson: > Any real upside to not having the map?

Less code to sync between vector <--> map, less error-prone I think

common/jinja/value.h
```diff
@@ -293,17 +303,44 @@ struct value_array_t : public value_t {
...
+        }
+        if (index < 0 || static_cast<size_t>(index) >= val_arr.size()) {
+            throw std::runtime_error("Index " + std::to_string(index) + " out of bounds for array of size " + std::to_string(val_arr.size()));
+        }
+        return val_arr[index];
+    }
     virtual const func_builtins & get_builtins() const override;
+    virtual bool is_hashable() const override {
+        if (std::all_of(val_arr.begin(), val_arr.end(), [&](auto & val) -> bool {
+            return !val->is_undefined() && !is_val<value_array>(val) && !is_val<value_object>(val);
```
ngxson: checking `is_val` here is redundant because the instance must already be the correct type, before calling this `virtual`

if you remove this `is_val` check, you no longer need to forward declare these types

plus, I think this can be moved to the cpp file

CISC: Uh, no, I'm not checking the array, I'm checking items in the array.

common/jinja/value.h
```diff
@@ -247,6 +252,8 @@ struct value_string_t : public value_t {
         return val_str.length() > 0;
     }
     virtual const func_builtins & get_builtins() const override;
+    virtual bool is_hashable() const override { return true; }
+    virtual std::string unique_hash() const { return type() + "." + as_string().str(); }
```
ngxson: The downside of this is that we now need double the memory to store a string. Also, this is not exactly a hash anymore because the output data is variable in length

> A hash function is a function that takes a variable-length input and produces a fixed-length output

I still prefer doing the `std::hash` way, which uses `size_t` as hash value

ngxson: Also, since `std::map` allow using custom `std::hash`, this can save us some headaches

CISC: Yeah, I know it's not, just thought I'd save the bother of renaming it later. :)

common/jinja/value.h
ngxson: Hmm, in this case, I think it's probably not very useful to support hashable array (i.e. tuple), I think no real template actually using it.

Instead, let's simplify this a bit by not allow hashing array for now.

common/jinja/value.h
CISC: > Also, since `std::map` allow using custom `std::hash`, this can save us some headaches

Ah, interesting, I'll look into this tomorrow.

ngxson: Small correction, it was `std::unordered_map` that support `std::hash`

The `map` uses red-black tree, so it requires `std::less` to compare between 2 nodes. `unordered_map` in the other hand use hash table, so it uses `std::hash`

Note that I've also looked into `std::flat_map` but it doesn't seem to be a good candidate. The problem is that it relies on binary search with `std::less` to search for a key, so we cannot use the insertion order to index the key.

So in conclusion, `std::unordered_map` + `std::hash` is the best choice here, while the key ordering can be stored in a `std::vector`; we will also need to implement `operator==`

CISC: How about vendoring `tsl::ordered_map`? It seems to align perfectly with our need:
https://github.com/Tessil/ordered-map?tab=readme-ov-file#key-features

ngxson: I think we can just reinvent it, it's quite minor to be vendored.

The implementation under the hood seems to be just a hash map + a vector to preserve the order, which is basically what I suggested above: https://github.com/Tessil/ordered-map/blob/3f7ee9abc68ce35047250f3723d7760a5362481e/include/tsl/ordered_hash.h#L271-L284

So it's quite trivial to reimplement. And since we don't need to polyfill all `std::map` functions like `rend`, `rbegin`, etc, we can keep our implementation smaller, I expect no more than 100 LOCs (not counting the `std::hash` implementation of `jinja::value`)

ngxson: Nevertheless, either using `tsl::ordered_map` or writing our own implementation, we will still need to implement `std::hash` and `operator==()`, which seems to be a more complex task

CISC: An added bonus of the hash implementation would be a much better value comparison and probably even cross type equivalence.

ngxson: Yeah, in most cases, I would expect the hash function to be derived straight from primitive type:

Something like:

```cpp
size_t jinja::value_int::hash() {
    return std::hash<int64_t>(val_int);
}
```

So whatever using the hash will see the jinja type as native cpp type.

CISC:
Phew, this led me down quite a few rabbit holes (how to use hash specialization in a class of the specialized type, c++ standard lacking hash_combine drama, virtual equality overload pitfalls, what have you), now it finally all makes sense and even compiles, but unfortunately does not work, more fun for tomorrow! :)

CISC:
Massive refactoring done, now has proper value hashing and equality operators (equivalence and strict (non-)equal). Added real `value_tuple` so that we can allow only immutable types as object keys.

Will be easy to add further operators for sorting in follow-up PR.

Lessons learned:
* You can't override specialized hash of `shared_ptr`
* You can't forward declare hashing functor of `unordered_map`
* The rest has been repressed, never to be mentioned again

Hopefully not too many mistakes made, ~but for some reason `test-chat-template` fails, so obviously some mistakes present, will investigate further unless someone brighter spots my error first~. :)

CISC:
> for some reason `test-chat-template` fails, so obviously some mistakes present, will investigate further unless someone brighter spots my error first. :)

Actually, it's the template that's faulty (not sure why this worked before?), we set `tools = []`, and the `Mistral-Large-Instruct-2407` template does this:
```jinja
{%- if not tools is defined %}
    {%- set tools = none %}
{%- endif %}
{%- if tools is not none and (message == user_messages[-1]) %}
    {{- "[AVAILABLE_TOOLS] [" }}
    {%- for tool in tools %}
        ...
    {%- endfor %}
    {{- "[/AVAILABLE_TOOLS]" }}
{%- endif %}
```

common/jinja/value.h
```diff
@@ -247,19 +294,48 @@ struct value_string_t : public value_t {
         return val_str.length() > 0;
     }
     virtual const func_builtins & get_builtins() const override;
+    virtual bool is_hashable() const override { return true; }
+    virtual size_t unique_hash() const noexcept override {
+        return hash_bytes(
+            type().data(), type().size(),
+            as_string().str().data(), as_string().str().size()
+        );
```
ngxson: IMO I'm a bit worry about the performance of this function. The `str()` function will concatenate all string parts, and even if the string is single-part, it will need to copy that part into a new memory area.

Instead, I think we can try implement a `string::hash` function that updates the hash while iterating the list of parts.

Probably not very important for now, but highly recommended to avoid `.str()`

common/jinja/value.h
```diff
@@ -146,76 +129,121 @@ struct value_t {
...
     virtual string as_string() const override { return std::to_string(val_int); }
     virtual bool as_bool() const override {
         return val_int != 0;
     }
     virtual const func_builtins & get_builtins() const override;
+    virtual bool is_numeric() const override { return true; }
+    virtual bool is_hashable() const override { return true; }
+    virtual size_t unique_hash() const noexcept override {
+        return hash_bytes(
+            type().data(), type().size(),
```
ngxson: [type_info::hash_code](https://en.cppreference.com/w/cpp/types/type_info/hash_code.html) can also be a good seed for hashing (?)

maybe something like: `hash_bytes<seed_type>(...)` that automatically generate seed based on `<seed_type>`

common/jinja/value.h
```diff
@@ -436,7 +695,18 @@ struct value_func_t : public value_t {
...
+    virtual bool is_hashable() const override { return true; }
+    virtual size_t unique_hash() const noexcept override {
+        return hash_bytes(
+            as_repr().data(), as_repr().size()
+        );
+    }
+protected:
+    virtual bool equivalent(const value_t & other) const override {
+        const value_func_t & other_val = static_cast<const value_func_t &>(other);
+        return type() == other.type() && val_func.target_type() == other.val_func.target_type() && val_func.target<value(const func_args &)>() == other.val_func.target<value(const func_args &)>() && typeid(*this) == typeid(other) && arg0 == other_val.arg0;
```
ngxson: In most cases, `type()` should be used for debugging only. If we need to compare type, it's best to use RTTI. I think it's redundant if we already check for `typeid` here

common/jinja/value.h
```diff
@@ -287,64 +384,217 @@ struct value_array_t : public value_t {
...
+            return val->is_immutable() && val->is_hashable();
+        })) {
+            return true;
+        }
+        return false;
+    }
+    virtual size_t unique_hash() const noexcept override {
+        return hash_bytes(
+            type().data(), type().size(),
+            as_repr().data(), as_repr().size()
```
ngxson: `as_repr` should also be reserved for debugging only. Using this here will result in significantly more memory usage and lower performance. I think it's best "chain" the element hash (see my review comment)

ngxson (commented):
Pretty close! Just need to optimize the a little bit by avoiding relying too much on string.

Btw, your `hash_bytes` with `seed` is in some ways an implementation of `hash_combine`, you can combine hash of multiple elements by:

```cpp
size_t cur_hash = 0; // or maybe something else
for (auto elem : arr) {
  cur_hash = hash_bytes(/* seed */ cur_hash, elem.unique_hash());
}
return cur_hash;
```

Doing `str()` or `as_repr()` uses a lot more memory. An example can be:

```py
my_tuple = ("some string")
my_outer_tuple = (my_arr, my_arr, my_arr, my_arr, my_arr)
```

Calculating hash of `my_outer_tuple` will now use 5 times more memory than needed, because `as_repr` need to allocate memory for all strings, even when they points to the same memory

CISC:
> Btw, your `hash_bytes` with `seed` is in some ways an implementation of `hash_combine`, you can combine hash of multiple elements by:
> 
> ```c++
> size_t cur_hash = 0; // or maybe something else
> for (auto elem : arr) {
>   cur_hash = hash_bytes(/* seed */ cur_hash, elem.unique_hash());
> }
> return cur_hash;
> ```

Yep, that was the point, but I see I made a footgun, never start with an initial seed (unless it is a previous hash).


> Calculating hash of `my_outer_tuple` will now use 5 times more memory than needed, because `as_repr` need to allocate memory for all strings, even when they points to the same memory

Very good point, I'll look into your suggestions and see what can be done.

common/jinja/value.h
```diff
@@ -287,64 +384,217 @@ struct value_array_t : public value_t {
...
+        std::ostringstream ss;
+        ss << "{";
+        for (size_t i = 0; i < val_obj.size(); i++) {
+            if (i > 0) ss << ", ";
+            auto & [key, val] = val_obj.at(i);
+            ss << value_to_string_repr(key) << ": " << value_to_string_repr(val);
+        }
+        ss << "}";
+        return ss.str();
+    }
```
ngxson, line 513: just note that I initially don't allow `as_repr` or `to_string` to be recursive, because it can go into an infinitive loop if the object/array entity is pointed back to itself:

```py
obj = {}
obj["a"] = obj

# or even harder to detect, nested circular
obj["a"] = {"b": obj}
```

a bit ironically that this is not actually classified as a vulnerability. just sometimes program are actually coded this way, and it is indeed a very common practice in high-level languages like javascript or python:

```js
var node = {"parent": null}
node["child"] = {
  "parent" : node,
  "child": null,
};
// now, JSON.stringify(node) will throw:
// TypeError: Converting circular structure to JSON
```

btw, may worth a fix for `to_json` as it can also stuck on this case. IIRC javascript simply throw an error if it detect circular in an object

CISC: Yeah, very aware of this, decided to ignore it for now. :)

CISC:
@ngxson Ok, be80c622a542ff72965c5ac6d02b65ec5130f800 should improve things, though I don't see how we can avoid composing strings as the parts may not be the same while the whole is.

Edit: I mean for equality, for hash it is fine processing parts due to the nature of FNV-1a.

common/jinja/value.h
CISC: Nested circulars are usually bypassed by keeping a reference of every encountered object that can nest when processing and simply skip/print `...` when coming across one that has already been visited. Will follow up if no-one else does.

ngxson:
I'll work on this a bit, will push some optimizations directly here

common/jinja/utils.h
```diff
@@ -46,4 +47,38 @@ static std::string fmt_error_with_source(const std::string & tag, const std::str
...
+    for (; c < e; ++c) {
+        seed = (seed ^ *c) * prime;
+    }
+
+    if constexpr (sizeof...(args) > 0) {
+        seed = hash_bytes(seed, std::forward<Args>(args)...);
+    }
+
+    return seed;
+}
```
ngxson: just thinking out loud in math terms: if we consider `hash_bytes(seed, data, size)` as a (surjective) function `f(s, x)` with `s = seed` and `x = tuple(data, size)`

given input data `x0`, `x1`, `x2`:

```
s = initial_seed
hash = f(f(f(s, x0), x1), x2)
```

and your "convenient" function `hash_bytes(x0, x1, ...)` can be defined as g:

```
g() = s
g(x0) = f(s, x0)
g(x0, x1) = f(f(s, x0), x1)
...
```

therefore, `g` is still surjective and `g(x0, x1) != g(x1, x0)` which is so far so good.

but now, consider: `f(x0 ~ x1) == g(x0, x1)` with `~` the "concatenation" operation: as long as `x0 ~ x1 == x2 ~ x3`, then `g(x0, x1) == g(x2, x3)`, which is expected when calculating hash of string parts (where each part is different but concatenated version is the same). technically says, this make `g` no longer a **good** hash function, since we have a known set of collisions. but in our context this is acceptable. `FNV-1a` is not that good anyway.

(note: the property `f(x0 ~ x1) == g(x0, x1)` can be derived from the implementation and can be considered as a postulation in this context)

in practice, to prevent this, most hash functions use some kind of internal state such that: `hash(s, x) = output(s, mix(x))`, so `hash(x0 ~ x1) != hash(x1 ~ x0)`. we can implement this easily but it's not really necessary. we are not doing cryptographic hash anyway.

end of the math part. now, in term of speed, what I think can be nice & fun to make `hash_bytes` to operate on blocks of data instead of one byte at a time. because one update depends on the prior state, CPU will have hard time with out-of-order execution. a simple vectorization should help a lot. just the case of `x0 ~ x1` will be a bit complicated. I'll see how to do that.

CISC: FNV-1a certainly has its flaws, it was mainly chosen for its simplicity and chainability, it performs well enough in terms of quality and speed here I think.

Feel free to improve upon it if you wish though. :)

ngxson: So I ended up refactor this into a new struct called `hasher()`, it is inspired by nodejs's `crypto.createHash()`

```cpp
// Old notion:
size_t seed = hash_bytes(data0, len0);
       seed = hash_bytes(data1, len1);
       // ...
size_t output = hash_bytes(seed, dataN, lenN);

// New notion
hasher hash = hasher().update(data0, len0).update(data1, len1); // ...
size_t output = hash.digest();
```

With this type of notion, we can in theory implement the notion of "internal state", though I won't add it because it's unnecessary (just mentioning here for completeness)

The new notion should have the exact same mathematical properties as the old one (reflected by the new test cases in `test-jinja`)

The result will be different from the old one though, because we are processing block of N bytes at once (on 64-bit system, we process 64/8 = 8 bytes). I expect most compilers will know how to vectorize it. If data is not a multiple of the block size, they will be buffered.

I'll do another pass tomorrow to see if there are any other clean up needed. In the meantime, feel free to review my last commit and lmk if anything seems off to you.

common/jinja/utils.h
```diff
@@ -46,4 +47,92 @@ static std::string fmt_error_with_source(const std::string & tag, const std::str
...
+struct hasher {
+    static constexpr auto size_t_digits = std::numeric_limits<size_t>::digits;
+    static constexpr size_t prime = size_t_digits == 64 ? 0x100000001b3 : 0x01000193; // seed
+    static constexpr auto block_size = size_t_digits / 8; // in bytes; allowing the compiler to vectorize the computation
+
+    static_assert(size_t_digits == 64 || size_t_digits == 32);
+
+    uint8_t buffer[block_size];
+    size_t idx = 0; // current index in buffer
+    size_t state = prime;
```
CISC:
```suggestion
    static constexpr size_t prime = size_t_digits == 64 ? 0x100000001b3 : 0x01000193;
    static constexpr size_t seed = size_t_digits == 64 ? 0xcbf29ce484222325 : 0x811c9dc5;
    static constexpr auto block_size = size_t_digits / 8; // in bytes; allowing the compiler to vectorize the computation

    static_assert(size_t_digits == 64 || size_t_digits == 32);

    uint8_t buffer[block_size];
    size_t idx = 0; // current index in buffer
    size_t state = seed;
```
The `prime` is an actual prime, should not be used as seed.

common/jinja/utils.h
```diff
@@ -3,6 +3,7 @@
 #include <string>
 #include <sstream>
 #include <algorithm>
+#include <limits>
```
CISC:
```suggestion
#include <cstdint>
#include <limits>
```

common/jinja/value.h
```diff
@@ -287,64 +391,224 @@ struct value_array_t : public value_t {
...
     virtual const func_builtins & get_builtins() const override;
+    virtual bool is_hashable() const override {
+            if (std::all_of(val_obj.begin(), val_obj.end(), [&](auto & pair) -> bool {
+                const auto & val = pair.second;
+                return val->is_immutable() && val->is_hashable();
+            })) {
+                return true;
+            }
+            return false;
+    }
```
CISC, line 572:
```suggestion
    virtual bool is_hashable() const override {
        if (std::all_of(val_obj.begin(), val_obj.end(), [&](auto & pair) -> bool {
            const auto & val = pair.second;
            return val->is_immutable() && val->is_hashable();
        })) {
            return true;
        }
        return false;
    }
```
Not sure what happened here?

common/jinja/value.h
```diff
@@ -287,64 +391,224 @@ struct value_array_t : public value_t {
...
+        }
+        return false;
+    }
+    virtual hasher unique_hash() const noexcept override {
+        const auto type_hash = typeid(*this).hash_code();
+        auto hash = hasher().update(&type_hash, sizeof(type_hash));
+        for (const auto & val : val_arr) {
+            const size_t val_hash = val->unique_hash().digest();
+            hash.update(&val_hash, sizeof(size_t));
+        }
```
CISC, line 446: We should allow `unique_hash` to be passed a `hasher` so that arrays and objects can just update it instead of hashing the hashes.

ngxson: forgot leave a comment here, but that won't work in this case: `["ab", "c"]` vs `["a", "bc"]`

the case of reusing state is mostly used by string parts, I don't think it's valid anywhere else

common/jinja/value.h
```diff
@@ -436,7 +708,27 @@ struct value_func_t : public value_t {
         return val_func(new_args);
     }
     virtual std::string type() const override { return "Function"; }
-    virtual std::string as_repr() const override { return type(); }
+    virtual std::string as_repr() const override { return type() + "<" + name + ">(" + (arg0 ? arg0->as_repr() : "") + ")"; }
+    virtual bool is_hashable() const override { return true; }
+    virtual hasher unique_hash() const noexcept override {
```
ngxson, line 711: a quick search points out that function in python is distinguished by the pointer, so I think this logic may not be correct.

~~also, the current impl will make things like `upper(string)` and `lower(string)` to be the same.~~ hmm no, we do compare the `val_func` pointer

I think it's better to remove this and just throw an error saying that function cannot be used as dict key. templates should never use it anyway. (overall I'm not quite comfortable about adding too much codes for functionalities that no one uses in practice)

ngxson: nevermind, it's easier to just compare the function pointer

common/jinja/utils.h
ngxson: actually we don't need `<limits>` because type size in bits can be calculated by `sizeof(...) * 8`

common/jinja/value.h
CISC: I don't think that's actually a problem as `typeid` is hashed inbetween those.

common/jinja/value.h
ngxson: It fails anyway because there was a line in `member_expression` that tries to convert the property into string: `property->as_string().str()`

but I think we shouldn't spend too much time on this. using function as key is a bad programming practice anyway, I don't think anyone gonna do it in practice.

I simply changed `value_func_t::is_hashable` to `false` in the last commit (probably better to also remove the `unique_hash()` implementation? I'm not sure)

ngxson (approved):
Feel free to merge when you're ready

common/jinja/value.h
CISC: I agree that this is not likely to be used in templates, it's fine not supporting it. Being able to hash it might be useful though?

CISC:
> Feel free to merge when you're ready

Seems to need `cstring` include, also still think https://github.com/ggml-org/llama.cpp/pull/18955#discussion_r2731023033 is a good idea (see my last comment).

common/jinja/value.h
ngxson: I don't have strong opinions whether to keep its hash function, but I think it will be dead code if it's never called in practice.

I would prefer even throw an exception whenever its hash function is call, just to see in which situations it is used. But the hash is currently declared as `noexcept` so I cannot really do that

CISC: Same thing with `undefined` really, is also technically dead code.

common/jinja/value.h
ngxson: mathematically say, hashing the digest of elements vs letting each element to update are (quasi-)equivalent in our case, because the `digest()` only do one job: to add padding to the input data.

my current version is still mathematically equivalent to adding a `typeid` in-between, while being more simple than having to add a version of `unique_hash` that takes another `hasher` as input.

common/jinja/value.h
ngxson: the `undefined_t` hash is needed only because it won't compile without one (kinda a dummy for now)

but yeah maybe it's ok to keep hash function of `func_t` because it doesn't take much code. the runtime already thrown an error if type is use as key && it is not hashable.

common/jinja/value.h
ngxson: also this code path is quite rarely used in practice (in case of using tuple as key), so I prefer to keep it simple for now. it should still be efficient enough doing this way.

CISC:
Ok, all good, will merge when CI is done.

PR #18956, opened 2026-01-20
Support TranslateGemma text+image translation in llama-server (lang codes + image url)

xiaobing318:

## Goal/Motivation
Support TranslateGemma text and image translation in llama-server, and allow request-level default language codes and message-level language codes to coexist, overcoming the previous limitation where only template default values could be used.

### Approach 1 
Pass `source_lang_code`/`target_lang_code` only via `chat_template_kwargs` as global defaults. This approach does not modify the `common_chat_msg_content_part` struct in `common/chat.h`. However, it causes language code fields written by downstream developers/users in the official Google/TranslateGemma recommended request body to be ignored, leaving only the defaults. Below is an example request/response using only `chat_template_kwargs`:

**Request Body**
```json
{
  "model": "translategemma-4b-it",
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "欢迎，翻译Gemma模型。"
        }
      ]
    }
  ],
  "chat_template_kwargs": {
    "source_lang_code": "zh",
    "target_lang_code": "en"
  },
  "temperature": 0,
  "max_tokens": 1024
}
```
**Response Body**
```json
{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "Welcome! I am Gemma, a translation model."
      }
    }
  ],
  "created": 1768903118,
  "model": "translategemma-4b-it.Q8_0.gguf",
  "system_fingerprint": "b7763-addc3aac5",
  "object": "chat.completion",
  "usage": {
    "completion_tokens": 11,
    "prompt_tokens": 82,
    "total_tokens": 93
  },
  "id": "chatcmpl-C3AkwTeEEC3dyioOrNvP3DD47kT1M0kE",
  "timings": {
    "cache_n": 0,
    "prompt_n": 82,
    "prompt_ms": 1354.697,
    "prompt_per_token_ms": 16.520695121951217,
    "prompt_per_second": 60.53014068828676,
    "predicted_n": 11,
    "predicted_ms": 1775.798,
    "predicted_per_token_ms": 161.4361818181818,
    "predicted_per_second": 6.194398236736387
  }
}
```

### Approach 2

Extend the message structure to preserve `messages[0].content[0].source_lang_code`/`target_lang_code`. This approach ensures that the language code fields written by downstream developers/users in the official Google/TranslateGemma recommended request body are not ignored. Below is an example request/response compatible with the official message format:

**Request Body**
```json
{
  "model": "translategemma-4b-it",
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "source_lang_code": "en",
          "target_lang_code": "zh",
          "text": "welcome, translategemma model."
        }
      ]
    }
  ],
  "temperature": 0,
  "max_tokens": 1024
}
```
**Response Body**
```json
{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "欢迎，翻译Gemma模型。"
      }
    }
  ],
  "created": 1768905402,
  "model": "translategemma-4b-it.Q8_0.gguf",
  "system_fingerprint": "b7764-4dd087a5f",
  "object": "chat.completion",
  "usage": {
    "completion_tokens": 8,
    "prompt_tokens": 82,
    "total_tokens": 90
  },
  "id": "chatcmpl-b0wPKXp6EROLUvHU8mkWZcYw9ntT7Hxd",
  "timings": {
    "cache_n": 0,
    "prompt_n": 82,
    "prompt_ms": 1314.545,
    "prompt_per_token_ms": 16.031036585365854,
    "prompt_per_second": 62.37899805636171,
    "predicted_n": 8,
    "predicted_ms": 1244.181,
    "predicted_per_token_ms": 155.522625,
    "predicted_per_second": 6.429932622343533
  }
}
```

## Choice and Rationale
The current change adopts a combined approach of “extend the message structure + preserve `chat_template_kwargs` defaults,” for the following reasons:
- **Compatible with the official request format:** The official format places language codes on the content part (`message[0].content[0].source_lang_code/target_lang_code`). Without extending the structure, these fields are ignored during parsing and only the default language codes can be used.
- **Avoid losing fields during template concatenation:** When the template does not require typed content (i.e., `requires_typed_content` is `false`), `common_chat_msgs_to_json_oaicompat` concatenates `content_parts` into plain text. Only when language codes are explicitly marked in the part will typed content be preserved, ensuring the fields enter template rendering.
- **Keep a simple calling path:** You can still set default language codes only via `chat_template_kwargs`, as the simplest request body and a backward-compatible path.

## Main Changes
- `chat.h` extends `common_chat_msg_content_part` to store language codes and marker flags.
- `chat.cpp` reads/writes language codes in OAI parsing/serialization and avoids losing them during concatenation (without this change, `messages[0].content[0].source_lang_code/target_lang_code` would be ignored at parse time and fall back to the default language codes). Meanwhile, `common_chat_params_init_translategemma` supports reading default language codes from `chat_template_kwargs` and, when content is an image marker, constructs the `image` structure required by TranslateGemma.
- `server-common.cpp` adds a parsing branch for `/v1/chat/completions` to handle `content[0].type = "image"` (the `url` field) as a compatibility extension for `image_url`, and uniformly converts it to an MTMD marker.

## Compatibility/Behavior

**Command used for debugging:**
```bash
llama-server -m "E:/data/llama.cpp-data/gguf-models/translategemma-4b-it.Q8_0.gguf" -mm "E:/data/llama.cpp-data/gguf-models/translategemma-4b-it.mmproj-Q8_0.gguf" -ngl 99 --jinja --media-path "E:/data/llama.cpp-data/translategemma-data" -c 2048
```

Images take effect only when mmproj is enabled and `allow_image` is true; otherwise the same error is returned consistently. Text-only models and non-multimodal requests are unaffected. Language code precedence is: “message-level fields > `chat_template_kwargs` defaults > built-in fallback.”

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -295,6 +295,14 @@ std::vector<common_chat_msg> common_chat_msgs_parse_oaicompat(const json & messa
                         common_chat_msg_content_part msg_part;
                         msg_part.type = type;
                         msg_part.text = part.at("text");
+                        if (part.contains("source_lang_code") && part.at("source_lang_code").is_string()) {
+                            msg_part.source_lang_code = part.at("source_lang_code").get<std::string>();
+                            msg_part.has_source_lang_code = true;
+                        }
+                        if (part.contains("target_lang_code") && part.at("target_lang_code").is_string()) {
+                            msg_part.target_lang_code = part.at("target_lang_code").get<std::string>();
+                            msg_part.has_target_lang_code = true;
+                        }
                         msg.content_parts.push_back(msg_part);
                     }
                 } else if (!content.is_null()) {
@@ -363,7 +371,16 @@ json common_chat_msgs_to_json_oaicompat(const std::vector<common_chat_msg> & msg
         if (!msg.content.empty()) {
             jmsg["content"] = msg.content;
         } else if (!msg.content_parts.empty()) {
-            if (concat_typed_text) {
+            bool keep_typed_content = false;
+            if (concat_typed_text && msg.role == "user") {
+                for (const auto & part : msg.content_parts) {
+                    if (part.has_source_lang_code || part.has_target_lang_code) {
+                        keep_typed_content = true;
+                        break;
+                    }
+                }
+            }
+            if (concat_typed_text && !keep_typed_content) {
                 std::string text;
                 for (const auto & part : msg.content_parts) {
                     if (part.type != "text") {
@@ -379,10 +396,17 @@ json common_chat_msgs_to_json_oaicompat(const std::vector<common_chat_msg> & msg
             } else {
                 auto & parts = jmsg["content"] = json::array();
                 for (const auto & part : msg.content_parts) {
-                    parts.push_back({
+                    json part_obj = {
                         {"type", part.type},
                         {"text", part.text},
-                    });
+                    };
+                    if (part.has_source_lang_code) {
+                        part_obj["source_lang_code"] = part.source_lang_code;
+                    }
+                    if (part.has_target_lang_code) {
+                        part_obj["target_lang_code"] = part.target_lang_code;
+                    }
+                    parts.push_back(std::move(part_obj));
                 }
             }
         } else {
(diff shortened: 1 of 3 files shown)


tools/server/server-common.cpp
```diff
@@ -922,6 +922,18 @@ json oaicompat_chat_params_parse(
...
+                    throw std::runtime_error("image input is not supported - hint: if this is unexpected, you may need to provide the mmproj");
+                }
+
+                handle_media(out_files, p, opt.media_path);
+
+                // replace this chunk with a marker
+                p["type"] = "text";
+                p["text"] = mtmd_default_marker();
+                p.erase("url");
+
```
ngxson, line 936: what's wrong with `image_url` ?

common/chat.h
ngxson, line 1: I don't get why we need this change

common/chat.cpp
```diff
@@ -2691,6 +2715,183 @@ static common_chat_params common_chat_params_init_exaone_moe(const common_chat_t
     return data;
 }
 
+static common_chat_params common_chat_params_init_translategemma(
+    const common_chat_template & tmpl,
+    const struct templates_params & inputs) {
+    common_chat_params data;
+
+    static const std::string kMtmdMediaMarker = "<__media__>";
+    static const std::string kMtmdImageMarker = "<__image__>";
```
ngxson, line 2724: do not hard code these values

ngxson (changes requested):
This PR introduce quite a lot of hacky code for a single model, we cannot accept it as-is.

I'll revisit it if more users report problem with it, but for now it seems like just a nice-to-have. I bet the model still works fine without all of these changes, just the user code have to handle more things.

xiaobing318:
> This PR introduce quite a lot of hacky code for a single model, we cannot accept it as-is.
> 
> I'll revisit it if more users report problem with it, but for now it seems like just a nice-to-have. I bet the model still works fine without all of these changes, just the user code have to handle more things.

Yes, these changes are currently intended to adapt to the recommended request body format. If the request body is modified, only user-side code needs to be added.

common/chat.h
xiaobing318: https://github.com/ggml-org/llama.cpp/pull/18956/changes/BASE..4dd087a5f55fcdf9a0f2db1d977cda9c6da2167b#diff-2580689a73e0d42f06cf4b5ed02e7cf87c6fc874343bce9c6f0e1ebf190d95b7R295

If the `source_lang_code` and `target_lang_code` fields are not added, the language codes in `messages` will be discarded during the parsing phase, rendering this request syntax ineffective, and only the global default values ​​will remain.

This approach did introduce quite a lot of hacky code; it was just an experiment.

tools/server/server-common.cpp
xiaobing318: Taking the `translategemma-4b-it` model from the `Translategemma` family as an example:

1. From the documentation at [translategemma-4b-it](https://huggingface.co/google/translategemma-4b-it), we know that A 'type' property where the value must be either 'text' or 'image'.
3. The template at [chat_template.jinja](https://huggingface.co/google/translategemma-4b-it/blob/main/chat_template.jinja) only supports the 'image' field, not the 'image_url' field.

xiaobing318:
> This PR introduce quite a lot of hacky code for a single model, we cannot accept it as-is.
> 
> I'll revisit it if more users report problem with it, but for now it seems like just a nice-to-have. I bet the model still works fine without all of these changes, just the user code have to handle more things.

For the TranslateGemma series models, if the code in llama-server is not modified, then the model's built-in request body format and chat template need to be modified/mapped. This can be implemented in downstream applications. I'm wondering if implementing such modifications or mappings (e.g., image--->image_url) in downstream applications would increase the burden on downstream developers. This is because downstream developers (some of whom may be unfamiliar with llama-server) expect an out-of-the-box experience when running TranslateGemma series models using llama-server, meaning they only need to focus on the recommended request body format and chat template of the TranslateGemma series models themselves, without needing to modify or map them in downstream applications, as this requires a certain level of understanding of llama-server.

**Sample request body that downstream developers need to modify/map**
```json
{
  "model": "translategemma-4b-it",
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "image_url",
          "image_url": {
            "url": "https://c7.alamy.com/comp/2YAX36N/traffic-signs-in-czech-republic-pedestrian-zone-2YAX36N.jpg",
            "detail": "auto"
          }
        }
      ]
    }
  ],
  "chat_template_kwargs": {
    "source_lang_code": "en",
    "target_lang_code": "zh"
  },
  "temperature": 0,
  "max_tokens": 2048
}
```
**Sample request body that downstream developers do not need to modify/map**
```json
{
  "model": "translategemma-4b-it",
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "image",
          "source_lang_code": "en",
          "target_lang_code": "zh",
          "url": "https://c7.alamy.com/comp/2YAX36N/traffic-signs-in-czech-republic-pedestrian-zone-2YAX36N.jpg"
        }
      ]
    }
  ],
  "temperature": 0,
  "max_tokens": 2048
}
```
@ngxson I’d like to hear your thoughts on this.

ngxson:
I don't see why it can be a burden for downstream devs.

The first example (manual mapping / without this PR) is flexible and can support any models in the future.

Your proposal here is bias towards the **current** version of TranslateGemma, there is no guarantee that their next model will be the same, or even someone else invent a new translation model, they will use another thing. So, we cannot allow adding API that is too model-specific.

If you want devs to be aware of this, push your example into the model's repo on HF instead.

xiaobing318:
> I don't see why it can be a burden for downstream devs.
> 
> The first example (manual mapping / without this PR) is flexible and can support any models in the future.
> 
> Your proposal here is bias towards the **current** version of TranslateGemma, there is no guarantee that their next model will be the same, or even someone else invent a new translation model, they will use another thing. So, we cannot allow adding API that is too model-specific.
> 
> If you want devs to be aware of this, push your example into the model's repo on HF instead.

Understood, thank you for sharing.

PR #18957, opened 2026-01-20
common, server : use the same User-Agent by default

angt:
This commit also ensures that if a custom User-Agent is used, it will be the only one sent.

### DIFF
--- a/common/common.h
+++ b/common/common.h
@@ -57,6 +57,8 @@ extern const char * LLAMA_COMMIT;
 extern const char * LLAMA_COMPILER;
 extern const char * LLAMA_BUILD_TARGET;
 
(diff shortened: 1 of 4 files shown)


ngxson (commented):
the HF backend is hard-coded to only return certain info when User-Agent == "llama-cpp", so this PR will break that.

I think we should simply remove the build_info from UA, don't think it's useful for now

angt:
> the HF backend is hard-coded to only return certain info when User-Agent == "llama-cpp", so this PR will break that.
> 
> I think we should simply remove the build_info from UA, don't think it's useful for now

But I've tested before submitting the PR 🤔

ngxson (approved):

angt:
But I was worried about the use of `llama.cpp` in `llama-server`, hope it doesn't change the behavior in that case

PR #18961, opened 2026-01-20
cli : fix reasoning responses in CLI

ngxson:
The chat format was not populate to task state in CLI, so reasoning content was not parsed correctly

With this PR, GLM-4.7 now works correctly on CLI:

image

TODO in follow-up PR: clean up `common_chat_parser_params`

### DIFF
--- a/common/chat.h
+++ b/common/chat.h
@@ -145,7 +145,7 @@ struct common_chat_templates_inputs {
     std::vector<common_chat_tool> tools;
     common_chat_tool_choice tool_choice = COMMON_CHAT_TOOL_CHOICE_AUTO;
     bool parallel_tool_calls = false;
-    common_reasoning_format reasoning_format = COMMON_REASONING_FORMAT_NONE;
+    common_reasoning_format reasoning_format = COMMON_REASONING_FORMAT_NONE; // TODO: refactor this to "bool enable_thinking"
     bool enable_thinking = true;
     std::chrono::system_clock::time_point now = std::chrono::system_clock::now();
     std::map<std::string, std::string> chat_template_kwargs;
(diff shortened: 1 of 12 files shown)


common/chat.h
```diff
@@ -165,14 +165,21 @@ struct common_chat_params {
     std::string                         parser;
 };
 
-struct common_chat_syntax {
+// per-message parsing syntax
+// should be derived from common_chat_params
+struct common_chat_parser_params {
     common_chat_format       format                = COMMON_CHAT_FORMAT_CONTENT_ONLY;
-    common_reasoning_format  reasoning_format      = COMMON_REASONING_FORMAT_NONE;
+    common_reasoning_format  reasoning_format      = COMMON_REASONING_FORMAT_NONE; // TODO: refactor this to "bool parse_reasoning"
```
ngxson, line 172: The whole point of `common_reasoning_format` is to tell which API response style we want. It is almost useless outside of server. This should be refactored in the next PR

pwilkin: Yes, 100% agree this should be a purely API server flag.

pwilkin (approved):
This is OK as a stopgap, but the structures regarding the parser params have become a mess again, need to rethink this.

tests/test-chat-parser.cpp
```diff
@@ -54,113 +54,109 @@ static void assert_throws(const std::function<void()> & fn, const std::string &
...
-        /* .format = */ COMMON_CHAT_FORMAT_CONTENT_ONLY,
-        /* .reasoning_format = */ COMMON_REASONING_FORMAT_NONE,
-        /* .reasoning_in_content = */ false,
-        /* .thinking_forced_open = */ false,
-    });
+    common_chat_parser_params params;
+    params.format = COMMON_CHAT_FORMAT_CONTENT_ONLY;
+    params.reasoning_format = COMMON_REASONING_FORMAT_NONE;
+    params.reasoning_in_content = false;
+    params.thinking_forced_open = false;
```
ngxson, line 61: @pwilkin just FYI, this ended up touching more code than I expected, not sure if this introduce more conflicts for your autoparser PR

pwilkin: No worries, I'll rebase

PR #18963, opened 2026-01-20
server: fix json_schema response_format handling

### LINKED ISSUE #10732: Misc. bug: server accepts response_format: json_object, but not response_format: json_schema
### Name and Version

on latest commit ce8784bdb153ff7794dde5a50b0ebfa51baa6171

but have been noticing it for several days now

### Operating systems

_No response_

_No response_

I have been trying to follow the steps for structured json output using json_schema in server [here](https://github.com/ggerganov/llama.cpp/tree/ce8784bdb153ff7794dde5a50b0ebfa51baa6171/examples/server#post-v1chatcompletions-openai-compatible-chat-completions-api)

however, I was not able to get any combination of `json_schema` to work. I *was* able to get `json_object` to work, doing effectively the same thing, but since this differs from the OpenAI API (which I suppose server is striving for) I suppose it's a bug. The official server docs also mention json_schema is supported

Does not work, does not apply any structured json schema at all, and does not indicate any failure or warning (simply returns unstructured output), taken directly from [OpenAI's docs](https://platform.openai.com/docs/guides/structured-outputs?lang=curl#chain-of-thought) (click "curl" for the curl version)

```json
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "math_reasoning",
        "schema": {
          "type": "object",
          "properties": {
            "steps": {
              "type": "array",
              "items": {
                "type": "object",
                "properties": {
                  "explanation": {
                    "type": "string"
                  },
                  "output": {
                    "type": "string"
                  }
                },
                "required": [
                  "explanation",
                  "output"
                ],
                "additionalProperties": false
              }
            },
            "final_answer": {
              "type": "string"
            }
          },
          "required": [
            "steps",
            "final_answer"
          ],
          "additionalProperties": false
        },
        "strict": true
      }
    }
```

however, "json_object" works as expected

```json
 "response_format": {
    "type": "json_object",
    "schema": {
        "type": "object",
        "properties": {
            "steps": {
                "type": "array",
                "items": {
                    "type": "object",
                    "properties": {
                        "explanation": {
                            "type": "string"
                        },
                        "output": {
                            "type": "string"
                        }
                    },
                    "required": [
                        "explanation",
                        "output"
                    ],
                    "additionalProperties": false
                }
            },
            "final_answer": {
                "type": "string"
            }
        },
        "required": [
            "steps",
            "final_answer"
        ],
        "additionalProperties": false
    }
}
```

### First Bad Commit

_No response_

### Relevant log output

_No response_

khasinski:
## Summary
- Add validation that `json_schema.schema` is present when using `response_format.type: "json_schema"`. Previously, a missing schema would silently result in an empty schema `{}` being used, which matches any JSON and doesn't constrain the output.
- Update the error message for invalid response_format types to include "json_schema" as a valid option (was only listing "text" and "json_object").

## Test plan
- [x] Built and tested with existing test suites (`test-json-schema-to-grammar`, `test-chat`)
- [x] Verified that valid `json_schema` format now correctly extracts the schema
- [x] Verified that missing `json_schema.schema` now throws a clear error

Fixes #10732

### DIFF
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -870,10 +870,15 @@ json oaicompat_chat_params_parse(
         if (response_type == "json_object") {
             json_schema = json_value(response_format, "schema", json::object());
         } else if (response_type == "json_schema") {
+            // https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format
+            // OpenAI expects: response_format.json_schema.schema
             auto schema_wrapper = json_value(response_format, "json_schema", json::object());
-            json_schema = json_value(schema_wrapper, "schema", json::object());
+            if (!schema_wrapper.contains("schema")) {
+                throw std::invalid_argument("response_format type \"json_schema\" requires \"json_schema.schema\" to be set");
+            }
+            json_schema = schema_wrapper.at("schema");
         } else if (!response_type.empty() && response_type != "text") {
-            throw std::invalid_argument("response_format type must be one of \"text\" or \"json_object\", but got: " + response_type);
+            throw std::invalid_argument("response_format type must be one of \"text\", \"json_object\", or \"json_schema\", but got: " + response_type);
         }
     }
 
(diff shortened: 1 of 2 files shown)


ngxson:
please add a test case for it

khasinski:
I think those added tests should cover most use cases, but I'm willing to write more if necessary

khasinski:
Can I get an update on this? :)

PR #18930, opened 2026-01-19
CUDA: Replace init_offsets kernel with iterators in cub-based argsort

ORippler:
This is mostly a QOL improvement, saving us the cost of materializing the iterator.

--- before

```
PS C:\Users\osimons\llama.cpp> $env:GGML_CUDA_DISABLE_GRAPHS=1; .\build-cuda\bin\RelWithDebInfo\test-backend-ops.exe -o ARGSORT perf
Backend 1/2: CUDA0
  Device description: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition
  Device memory: 97886 MB (96048 MB free)

  ARGSORT(type=f32,ne=[65000,16,1,1],order=0):                  4130 runs -   393.42 us/run -     8125 kB/run -   19.70 GB/s
  ARGSORT(type=f32,ne=[200000,1,1,1],order=0):                  8192 runs -   129.78 us/run -     1562 kB/run -   11.48 GB/s
  ARGSORT(type=f32,ne=[200000,16,1,1],order=0):                 1343 runs -  1100.22 us/run -    25000 kB/run -   21.67 GB/s
````

--- after

```
Backend 1/2: CUDA0
  Device description: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition
  Device memory: 97886 MB (96048 MB free)

  ARGSORT(type=f32,ne=[65000,16,1,1],order=0):                  4130 runs -   394.60 us/run -     8125 kB/run -   19.64 GB/s
  ARGSORT(type=f32,ne=[200000,1,1,1],order=0):                 16384 runs -   119.31 us/run -     1562 kB/run -   12.49 GB/s
  ARGSORT(type=f32,ne=[200000,16,1,1],order=0):                 1343 runs -  1097.34 us/run -    25000 kB/run -   21.73 GB/s
```

PR #18964, opened 2026-01-20
CUDA: Fix builds for older CCCL versions by ifdefing strided_iterator

ORippler:
Strided iterator was added in [CCCL 3.1](https://github.com/NVIDIA/cccl/releases/tag/v3.1.0), which is packaged into [CTK
13.1](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#id5)

Should fix #18960

### DIFF
--- a/ggml/src/ggml-cuda/argsort.cu
+++ b/ggml/src/ggml-cuda/argsort.cu
@@ -44,14 +44,14 @@ void argsort_f32_i32_cuda_cub(ggml_cuda_pool & pool,
     const dim3 grid_size((ncols + block_size - 1) / block_size, nrows);
     init_indices<<<grid_size, block_size, 0, stream>>>(temp_indices, ncols, nrows);
 
-#    ifdef STRIDED_ITERATOR_AVAILABLE
+#ifdef STRIDED_ITERATOR_AVAILABLE
     auto offset_iterator = cuda::make_strided_iterator(cuda::make_counting_iterator(0), ncols);
-#    else
+#else
     ggml_cuda_pool_alloc<int> offsets_alloc(pool, nrows + 1);
     int *                     offset_iterator = offsets_alloc.get();
     const dim3                offset_grid((nrows + block_size - 1) / block_size);
     init_offsets<<<offset_grid, block_size, 0, stream>>>(offset_iterator, ncols, nrows);
-#    endif
+#endif
     CUDA_CHECK(cudaMemcpyAsync(temp_keys, x, ncols * nrows * sizeof(float), cudaMemcpyDeviceToDevice, stream));
 
     size_t temp_storage_bytes = 0;
--- a/ggml/src/ggml-cuda/argsort.cu
+++ b/ggml/src/ggml-cuda/argsort.cu
@@ -2,6 +2,9 @@
 
 #ifdef GGML_CUDA_USE_CUB
 #    include <cub/cub.cuh>
+#    if (CCCL_MAJOR_VERSION >= 3 && CCCL_MINOR_VERSION >= 1)
+#        define STRIDED_ITERATOR_AVAILABLE
+#    endif
 using namespace cub;
 #endif  // GGML_CUDA_USE_CUB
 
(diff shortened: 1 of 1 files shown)


ggml/src/ggml-cuda/argsort.cu
```diff
@@ -33,8 +44,14 @@ void argsort_f32_i32_cuda_cub(ggml_cuda_pool & pool,
...
 
+#    ifdef STRIDED_ITERATOR_AVAILABLE
     auto offset_iterator = cuda::make_strided_iterator(cuda::make_counting_iterator(0), ncols);
-
+#    else
+    ggml_cuda_pool_alloc<int> offsets_alloc(pool, nrows + 1);
+    int *                     offset_iterator = offsets_alloc.get();
+    const dim3                offset_grid((nrows + block_size - 1) / block_size);
+    init_offsets<<<offset_grid, block_size, 0, stream>>>(offset_iterator, ncols, nrows);
+#    endif
```
am17an:
```suggestion
#ifdef STRIDED_ITERATOR_AVAILABLE
auto offset_iterator = cuda::make_strided_iterator(cuda::make_counting_iterator(0), ncols);
#else
    ggml_cuda_pool_alloc<int> offsets_alloc(pool, nrows + 1);
    int *                     offset_iterator = offsets_alloc.get();
    const dim3                offset_grid((nrows + block_size - 1) / block_size);
    init_offsets<<<offset_grid, block_size, 0, stream>>>(offset_iterator, ncols, nrows);
#endif //STRIDED_ITERATOR_AVAILABLE
```

am17an (commented):
If there is no noticeable performance gain from doing this, we can just remove this for now and add a TODO when it's supported. As it is the CUDA code is full of ifdefs, but those are mostly required for performance reasons

ggml/src/ggml-cuda/argsort.cu
ORippler: The indentation of the preprocessor directives came from the automated code formatter, but if preferred I can also unindent this.

ORippler:
> If there is no noticeable performance gain from doing this, we can just remove this for now and add a TODO when it's supported. As it is the CUDA code is full of ifdefs, but those are mostly required for performance reasons

This is already supported, it just requires CCCL >= 3.1, which is bundled with Cuda Toolkit (CTK) >= 13.1. Given the long range of supported CTK versions by our cuda backend (11.8-13.1), it's going to be a long time until CCCL >= 3.1 can be assumed to be present for every build (and thus compile-time checks be avoided).

While the performance gain in this specific case is not significant, I'd personally still favor generating visibility of new CCCL features over removing it to increase their eventual adoption by CUDA devs.

elfarolab:
> > If there is no noticeable performance gain from doing this, we can just remove this for now and add a TODO when it's supported. As it is the CUDA code is full of ifdefs, but those are mostly required for performance reasons
> 
> This is already supported, it just requires CCCL >= 3.1, which is bundled with Cuda Toolkit (CTK) >= 13.1. Given the long range of supported CTK versions by our cuda backend (11.8-13.1), it's going to be a long time until CCCL >= 3.1 can be assumed to be present for every build (and thus compile-time checks be avoided).
> 
> While the performance gain in this specific case is not significant, I'd personally still favor generating visibility of new CCCL features over removing it to increase their eventual adoption by CUDA devs.

@ORippler 

Hopefully in the future it will be as easy as you say..

It is difficult to switch to new releases of CUDA on embedded Linux platforms like Jetson. For older hardware (JetPack 4.x series), you can't update CUDA without breaking the system, you must wait for the next JetPack release to use newer CUDA libraries. On newer Jetpack, such as 6.x, you have limited new releases of updates available via apt packages. CUDA is deeply integrated with NVIDIA embedded Linux hardware, so independent upgrades are impossible. I've tried this for years, but after months in production, something always fails. So I stick to official JetPack releases to avoid deployment issues.

ggml/src/ggml-cuda/argsort.cu
JohannesGaessler: It's preferable to indent as outlined by Aman, please also add an `GGML_ABORT("fatal error")` if there is an attempt to execute this code without the feature being available.

ORippler: > It's preferable to indent as outlined by Aman

Done

> please also add an GGML_ABORT("fatal error") if there is an attempt to execute this code without the feature being available.

This is not necessary as we have a fallback path for CCCL <= 3.1 (which was erroneously removed in #18930 and is re-added in this PR as I was previously unaware of when `cuda::strided_iterator` was added to CCCL)

JohannesGaessler (approved):

ddh0:
If my understanding is correct, this needs to be merged to fix CUDA compilation, which is currently broken. See #18960 and #18930. (Shouldn't these types of things be caught by the CI?)

Edit: And also #18959

pwilkin:
Going to merge this since the builds are broken without it, if any problems come out we can address in a follow-up.

PR #18967, opened 2026-01-20
ggml-zdnn: Mark zDNN buffers as non-host

AlekseiNikiforovIBM:
While buffers reside in host memory,
additional transformation is needed to use buffers with zDNN.

This transformation was missing when using direct io.
Marking buffers as non-host forces use of ggml_backend_tensor_set function,
which in turn properly initializes and transforms buffers.

Fixes #18848

### DIFF
--- a/ggml/src/ggml-zdnn/ggml-zdnn.cpp
+++ b/ggml/src/ggml-zdnn/ggml-zdnn.cpp
@@ -368,7 +368,8 @@ static size_t ggml_backend_zdnn_buffer_type_get_alignment(ggml_backend_buffer_ty
 }
 
 static bool ggml_backend_zdnn_buffer_type_is_host(ggml_backend_buffer_type_t buft) {
-    return true;
+    /* while it resides in host memory, additional transformation is needed */
+    return false;
 
     GGML_UNUSED(buft);
 }


AlekseiNikiforovIBM:
@CISC , please take a look.

CISC:
I'm not really familiar with how the backend code works, but AFAICT Direct IO didn't change this codepath?

AlekseiNikiforovIBM:
Without direct io or before direct io I'm getting `use_mmap == true`, and this code path is used with call to `ggml_backend_tensor_set`:

https://github.com/ggml-org/llama.cpp/blob/37c35f0e1c625831687b146cbb0a57654ef88ca2/src/llama-model-loader.cpp#L1121

With direct io I get `use_mmap == false` and code ends in this code path for zDNN backend:

https://github.com/ggml-org/llama.cpp/blob/37c35f0e1c625831687b146cbb0a57654ef88ca2/src/llama-model-loader.cpp#L1127-L1133

There's unfortunately no call to `ggml_backend_tensor_set` or other similar functions. zDNN-specific tensor structures are not created.

With this change when direct io is used I'm aiming for function to go into this branch and call `ggml_backend_tensor_set`, or `ggml_backend_tensor_set_async` could work as well:

https://github.com/ggml-org/llama.cpp/blob/37c35f0e1c625831687b146cbb0a57654ef88ca2/src/llama-model-loader.cpp#L1189-L1195

CISC:
So, does that mean it used to fail with `--no-mmap`?

AlekseiNikiforovIBM:
I think so. I tried building current llama.cpp without this change and ran it with `--no-direct-io --no-mmap` and it failed same way as without these two options, i.e. with direct io.

CISC:
> I think so. I tried building current llama.cpp without this change and ran it with `--no-direct-io --no-mmap` and it failed same way as without these two options, i.e. with direct io.

Right, so probably always broken, wonder if it had some other sideeffects too...

CISC (approved):

taronaeo (approved):

PR #18968, opened 2026-01-20
CANN: Remove unnecessary wrapper for `gml_backend_buft_is_cann`

rauletorresc:
**Description of the problem**

There is an unnecessary indirection caused by the wrapper function `ggml_backend_buffer_is_cann` calling `gml_backend_buft_is_cann` internally, as the buffer type info can be accessed directly from the buffer info without having to make two nested function calls.

 **Proposed solution**

- Call `gml_backend_buft_is_cann` directly while passing the buffer type info, which is accessible.
- Remove `ggml_backend_buffer_is_cann` function.
- Move all necessary logic before its usage, in order to avoid having forward declarations and duplicated documentation.


hipudding (approved):
Looks good to me. this wrapper is not necessary.

rauletorresc:
@hipudding I guess is ready for merging :)

PR #18970, opened 2026-01-20
opencl: enable the general fp mm for non-cont input and as a fallback for specialized kqv kernel for adreno

lhez:
The specialized kqv kernel for Adreno uses `image1d_buffer_t` to wrap `dst` to utilize texture processor. However, `image1d_buffer_t` has dimension limit. The limit can be reached when there are enough tokens in kv cache.

For example, providing gpt-oss-20b with a prompt of about 4k tokens using `llama-completion` with `-ub 1024` will hit the limit, resulting in assert failure.

This PR utilizes the general mm for floating point by copying to `src0` and `src1` to contiguous if necessary. I believe this has been used in vulkan backend. A check is added to ensure that the arguments are within the limit of `image1d_buffer_t` for the specialized kqv kernel.


max-krasnyansky (approved):
Looks good. 
Didn't get a chance to try it on my setup yet. I'm testing some Hexagon changes and will give this a go along with that.

PR #18972, opened 2026-01-20
Add Devstral-2 (Ministral3ForCausalLM) architeture to  convert_hf_to_gguf.py

emsi:
This adds support for newer architectures like Devstral-2 (addresses [#17987](https://github.com/ggml-org/llama.cpp/issues/17987))

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -2979,7 +2979,11 @@ class Llama4VisionModel(MmprojModel):
         return []
 
 
-@ModelBase.register("Mistral3ForConditionalGeneration")
+@ModelBase.register(
+    "Mistral3ForConditionalGeneration",
+    "Ministral3ForCausalLM",
+)
+
 class Mistral3Model(LlamaModel):
     model_arch = gguf.MODEL_ARCH.MISTRAL3
 
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -2983,7 +2983,6 @@ class Llama4VisionModel(MmprojModel):
     "Mistral3ForConditionalGeneration",
     "Ministral3ForCausalLM",
 )
-
 class Mistral3Model(LlamaModel):
     model_arch = gguf.MODEL_ARCH.MISTRAL3
 


CISC:
Given https://github.com/ggml-org/llama.cpp/issues/17987#issuecomment-3650621568 I would like you to confirm that the converted model works and produces coherent output.

emsi:
> Given [#17987 (comment)](https://github.com/ggml-org/llama.cpp/issues/17987#issuecomment-3650621568) I would like you to confirm that the converted model works and produces coherent output.

I tested it on `mistralai/Devstral-2-123B-Instruct-2512`. No issues at all. Produces long, coherent outputs. I guess [#17987 (comment)](https://github.com/ggml-org/llama.cpp/issues/17987#issuecomment-3650621568) might be a template issue or something else (see the possible FP8 caveat at the bottom).


## What I did:
Because I'm on A100 hence no FP8 and had to convert to 16 bits (BF16) first:

```
from transformers import AutoModelForCausalLM, FineGrainedFP8Config, AutoTokenizer 
import torch

model_id = "mistralai/Devstral-2-123B-Instruct-2512"
out_dir = "Devstral-2-123B-Instruct-2512-bf16"

qcfg = FineGrainedFP8Config(dequantize=True)

model = AutoModelForCausalLM.from_pretrained(
    model_id,
    device_map="cpu",
    dtype=torch.bfloat16, 
    quantization_config=qcfg,
    low_cpu_mem_usage=True,
)

# Important: remove quantization_config before saving so reload doesn't try FP8
if hasattr(model.config, "quantization_config"):
    model.config.quantization_config = None

model.save_pretrained(out_dir, safe_serialization=True, save_original_format=False)
```

Then I made sure I'm using the python env with proper dependencies. For that I used a docker container built off of axolotl image with following changes:

```
FROM axolotlai/axolotl:main-latest

RUN git fetch
RUN git checkout transformers-v5
RUN pip install -e .

RUN python scripts/cutcrossentropy_install.py | sh
RUN pip install 'mistral-common[opencv]==1.8.5'
```

Using above container I converted the model to GGUF using the patched conversion script:
`python convert_hf_to_gguf.py  Devstral-2-123B-Instruct-2512-bf16 --outtype bf16 --outfile Devstral-2-123B-Instruct-2512-bf16.gguf`

Quantized (no need to use special python env for that ofc):
`llama-quantize Devstral-2-123B-Instruct-2512-bf16.gguf Devstral-2-123B-Instruct-2512-bf16-Q4_K_M.gguf  Q4_K_M`

I served it like that:
`llama-server  --model Devstral-2-123B-Instruct-2512-bf16-Q4_K_M.gguf --alias "Devstral-2-123B"     --threads -1     --n-gpu-layers 999     --prio 3     --min_p 0.01     --ctx-size 16384  --port 8001     --jinja --parallel 1`


I can't vouch it works with other paths, especially when working in FP8 because I have no access to FP8 capable hardware atm. That's the only caveat. On the other hand I'm pretty convinced that the change is safe because the architecture matches and the conversion produced working model when using proper env.

convert_hf_to_gguf.py
```diff
@@ -2979,7 +2979,11 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         return []
 
 
-@ModelBase.register("Mistral3ForConditionalGeneration")
+@ModelBase.register(
+    "Mistral3ForConditionalGeneration",
+    "Ministral3ForCausalLM",
+)
+
```
```suggestion
```

CISC (approved):

PR #18976, opened 2026-01-20
Split shared state (webgpu_context) into global state and per-thread state

nikhilJain17:
Right now, the WebGPU backend has a global `webgpu_context` struct with all the information required to instantiate and run a WebGPU graph.

We want to split up the `webgpu_context` struct as follows:  
- Move `get_tensor_sharing_buf` to global state, along with the `mutex`
- Move WebGPU handles like `wgpu::Device`, `wgpu::Adapter`, and `wgpu::Instance` to global state in the registration api
- Move WebGPU device capabilities to global state
- Make `webgpu_context` a per-thread struct, with pipelines and buffers as needed to run a graph (i.e. error buffers, timing buffers, debug buffers)


ggml/src/ggml-webgpu/wgsl-shaders/embed_wgsl.py
```diff
@@ -18,6 +18,13 @@ def parse_decls(decls_text):
         decls[name.strip()] = code.strip()
     return decls
 
+def replace_repl_placeholders(variant, template_map):
```
reeselevine: looks like this function was accidentally duplicated

ggml/src/ggml-webgpu/wgsl-shaders/unary_op.wgsl
```diff
@@ -0,0 +1,587 @@
+#define(REPL_TEMPLATES)
```
reeselevine: I think this file was accidentally added back in

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -267,25 +267,25 @@ struct webgpu_command {
 #endif
 };
 
+// Forward declaration
+struct webgpu_global_context;
```
reeselevine: can we avoid forward declarations as much as possible? I think this struct could simply be moved higher, lmk if I'm wrong though.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2547,302 +2577,492 @@ static void ggml_webgpu_init_mul_mat_pipeline(webgpu_context & webgpu_ctx) {
...
+        "device_desc: %s\n",
+        info.vendorID, std::string(info.vendor).c_str(), std::string(info.architecture).c_str(), info.deviceID, 
+        std::string(info.device).c_str(), std::string(info.description).c_str());
+    return true;
+}
+
+
+// Fetch the webgpu_context for this thread (indexed by device for multiple devices running per thread)
+// If it hasn't been created yet, then initiailize one.
+static webgpu_context get_per_thread_webgpu_context(ggml_backend_dev_t dev) {
```
reeselevine: could all of this logic be moved into `device_init` below (which should actually be renamed to `backend_init`)? And `ggm_backend_t` still holds a reference to the per-thread `webgpu_context`? Then, we wouldn't need a map from dev_t -> webgpu_context, the init_once flag, and this function in general. Let me know if I'm missing something.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2854,7 +3074,8 @@ static ggml_backend_buffer_type_t ggml_backend_webgpu_device_get_buffer_type(ggm
         },
         /* .device  = */
         dev,
-        /* .context = */ NULL,
+        /* .context = */
+        NULL
```
reeselevine, line 3106: nit: wondering about formatting change here, did you run `clang-format` on this file?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
nikhilJain17: I originally tried putting the `webgpu_context` in the backend context, but the issue is that both the buffer interface and backend interface require a `webgpu_context`, and sometimes the buffer interface gets initialized before the backend interface. When I tried this, `test-thread-safety` was failing because `...alloc_buffer()` was called before the backend context was created, so there was no `webgpu_context`. Maybe the long term solution is splitting `webgpu_context` into `webgpu_compute_context` and `webgpu_buffer_context`, but I didn't look into how feasible that would be.

nikhilJain17: Also updated function name from `device_init` to `backend_init`

ggml/src/ggml-webgpu/ggml-webgpu.cpp
nikhilJain17: I originally did not, but ran and committed the changes just now.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
nikhilJain17: Sure, fixed

ggml/src/ggml-webgpu/wgsl-shaders/unary_op.wgsl
nikhilJain17: Just fixed

ggml/src/ggml-webgpu/wgsl-shaders/embed_wgsl.py
nikhilJain17: Fixed

reeselevine (approved):

PR #18953, opened 2026-01-20
CUDA: add gqa_ratio 4 for GLM 4.7 flash

am17an:
Enable FA for GLM 4.7, I'm not sure it's optimal but at least it does not go to CPU. Fixes #18944

PR #13435, opened 2025-05-10
CUDA: faster Deepseek FA, add Turing support

JohannesGaessler:
This PR optimizes the CUDA FlashAttention kernel for better Deepseek performance. Since there seem to still be issues with my last FA PR on master I would suggest only merging this PR after they are sorted out.

* The batch sizes for K, V, and combining the results can now be set as a function of compute capability and the number of Q columns per CUDA block. This enables running the kernel on Turing with 64 kiB SRAM per streaming multiprocessor (vs. 99+ kiB on Ampere and newer). I don't have a Turing GPU for testing but it should at least be faster than running the kernel on the CPU. For Deepseek I found parameters that are faster for batch size 1 than master on Ampere or newer. For Gemma prompt processing on Turing should now also be a bit faster.
* Following https://github.com/ggml-org/llama.cpp/pull/13306#issuecomment-2858540263 I re-wrote the kernel to load both the K and V data from the K tensor. For Deepseek the V tensor is in fact now not being used at all by the CUDA code. For batch size 1 100% of the V data load can be skipped, for other batch sizes 50% can be skipped. Going forward I would suggest re-writing the code in other backends as well to use only the K tensor. The KV cache size could then be reduced by ~47% by simply not allocating and filling the V cache. At least as long as FlashAttention is used this should be relatively simple. So for a first version it would I think be fine to only deduplicate K and V if FA is used. @jukofyork are you interested in working on this?
* Following https://github.com/ggml-org/llama.cpp/pull/13306#discussion_r2081496830 I tried to make the suggested change to `ggml_cuda_unroll`. However, the code then only compiles if a [compile flag for relaxed constexpr](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#expt-relaxed-constexpr-expt-relaxed-constexpr) is used. Notably this compile flag is declared experimental with no guarantee of being forwards compatible. Since the compiler was able to do the correct loop optimization with the code as it is now I would suggest keeping it that way.

<details>
<summary>Performance</summary>

| GPU      | Model              | Microbatch size | Test    | t/s master |  t/s PR | Speedup |
| :------- | :----------------- | --------------: | :------ | ---------: | ------: | ------: |
| RTX 3090 | deepseek2 16B Q4_0 |               1 | pp16384 |     146.30 |  163.07 |    1.11 |
| RTX 3090 | deepseek2 16B Q4_0 |               2 | pp16384 |     143.80 |  145.86 |    1.01 |
[...]

PR #18986, opened 2026-01-21
mla : make the V tensor a view of K

ggerganov:
cont https://github.com/ggml-org/llama.cpp/pull/18953

This is based on top of the changes in https://github.com/ggml-org/llama.cpp/pull/18953.

Currently, the CUDA FA implementation has certain hardcoded assumptions about the layout of the K and V tensors when MLA is involved (see https://github.com/ggml-org/llama.cpp/pull/13435). The goal of this change is to make things more generic and avoid these assumptions:

- Change the concat layout of the MLA cache. The old was `[pe, kv]`. The new one is `[kv, pe]`. This has certain implications for backends such as Vulkan and overall it's better layout in terms of memory alignment
- Update the graph build code to pass the `V` tensor as a view of `K`. This can be used as a signal for the CUDA (and later other) backends to avoid loading extra `V` data during compute. _(the elimination of the `V` component from the `llama_kv_cache` will be done in follow-up PR - for now it's just a redundant data)_
- Added tests to exercise the "V is view of K" path of the FA - ~currently these tests will still fail (for CUDA only) because we changed the layout (see the point above). This needs to be taken into account in the CUDA implementation. For more info, see the comments in the `test-backend-ops.cpp`~ (CUDA impl fixed in [f07c65b](https://github.com/ggml-org/llama.cpp/pull/18986/commits/f07c65b4efb29be4197c0156d44e8ee0885194fd))

Next PRs:

- Do not allocate V cache when MLA is used
- Add regular "V is view of K" FA tests in `test-backend-ops` and avoid hardcoded logic
- ~Expand CUDA kernel coverage for quantized MLA cache and general improvement of the MLA-related logic~ (resolved in https://github.com/ggml-org/llama.cpp/pull/19023)
- Try to take advantage of "V is view of K" in other backends

### DIFF
--- a/ggml/src/ggml-cuda/fattn-common.cuh
+++ b/ggml/src/ggml-cuda/fattn-common.cuh
@@ -778,11 +778,14 @@ void launch_fattn(
 ) {
     constexpr int ncols = ncols1 * ncols2;
 
-    const bool is_mla = DV == 512; // TODO better parameterization
-
     const ggml_tensor * Q = dst->src[0];
     const ggml_tensor * K = dst->src[1];
     const ggml_tensor * V = dst->src[2];
+
+    // TODO: make this more generic by removing the notion of "MLA".
+    //       for example "is V a view of K?" so we can skip loading it.
+    //       V strides should be driven by V itself and avoid assumption of the data layout
+    const bool is_mla = V->op == GGML_OP_VIEW && V->src[0] == K;
 
     GGML_ASSERT(V || is_mla);
 
--- a/ggml/src/ggml-cuda/fattn-mma-f16.cuh
+++ b/ggml/src/ggml-cuda/fattn-mma-f16.cuh
@@ -794,7 +794,8 @@ static __device__ __forceinline__ void flash_attn_ext_f16_iter(
     // For MLA K and V have the same data.
     // Therefore, iterate over V in reverse and re-use the data if possible.
     static_assert(!mla || nstages <= 1, "combination of MLA and multi-stage loading not implemented");
-    constexpr int reusable_cutoff = mla ? (DV - 1) - (DV - 1) % (2*nbatch_K2) : DV;
+    // constexpr int reusable_cutoff = mla ? (DV - 1) - (DV - 1) % (2*nbatch_K2) : DV;
+    constexpr int reusable_cutoff = DV; // TODO implement properly
 #if defined(AMD_WMMA_AVAILABLE) && !defined(LDMATRIX_TRANS_AVAILABLE)
     T_A_VKQ A_identity;
     make_identity_mat(A_identity);
--- a/ggml/src/ggml-cuda/fattn-mma-f16.cuh
+++ b/ggml/src/ggml-cuda/fattn-mma-f16.cuh
@@ -794,7 +794,7 @@ static __device__ __forceinline__ void flash_attn_ext_f16_iter(
     // For MLA K and V have the same data.
     // Therefore, iterate over V in reverse and re-use the data if possible.
     static_assert(!mla || nstages <= 1, "combination of MLA and multi-stage loading not implemented");
-    constexpr int reusable_cutoff = mla ? (DKQ - 1) - (DKQ - 1) % (2*nbatch_K2) - (DKQ - DV) : DV;
+    constexpr int reusable_cutoff = mla ? (DV - 1) - (DV - 1) % (2*nbatch_K2) : DV;
 #if defined(AMD_WMMA_AVAILABLE) && !defined(LDMATRIX_TRANS_AVAILABLE)
     T_A_VKQ A_identity;
     make_identity_mat(A_identity);
@@ -1552,7 +1552,7 @@ static __global__ void flash_attn_ext_f16(
             (const half *) (mask + nb33*(sequence % ne33));
         float2       * dstk   = ((float2 *) dst) + (sequence*ne01.z*ne02 + head0) * (DV/2);
 
-        const half2 * V_h2 = mla ? K_h2 + (DKQ/2 - DV/2) : (const half2 *) (V + nb23*sequence + nb22*(head0 / gqa_ratio));
+        const half2 * V_h2 = mla ? K_h2 : (const half2 *) (V + nb23*sequence + nb22*(head0 / gqa_ratio));
         const float * sinks_f = sinks ? (const float *) sinks + head0 : nullptr;
 
         const float slope = ncols2 == 1 ? get_alibi_slope(max_bias, head0, n_head_log2, m0, m1) : 1.0f;
@@ -1596,7 +1596,7 @@ static __global__ void flash_attn_ext_f16(
         (const half *) (mask + nb33*(sequence % ne33));
     float2       * dstk   = ((float2 *) dst) + (sequence*ne01.z*ne02 + head0) * (DV/2);
 
-    const half2 * V_h2 = mla ? K_h2 + (DKQ/2 - DV/2) : (const half2 *) (V + nb23*sequence + nb22*(head0 / gqa_ratio));
+    const half2 * V_h2 = mla ? K_h2 : (const half2 *) (V + nb23*sequence + nb22*(head0 / gqa_ratio));
     const float * sinks_f = sinks ? (const float *) sinks + head0 : nullptr;
 
     const float slope = ncols2 == 1 ? get_alibi_slope(max_bias, head0, n_head_log2, m0, m1) : 1.0f;
--- a/src/models/deepseek2.cpp
+++ b/src/models/deepseek2.cpp
@@ -124,14 +124,14 @@ llm_build_deepseek2::llm_build_deepseek2(const llama_model & model, const llm_gr
 
                 // {n_embd_head_qk_rope + kv_lora_rank, n_head, n_tokens}
                 // note: rope must go first for in-place context shifting in build_rope_shift()
-                ggml_tensor * Qcur = ggml_concat(ctx0, q_pe, q_nope_absorbed, 0);
+                ggml_tensor * Qcur = ggml_concat(ctx0, q_nope_absorbed, q_pe, 0);
                 cb(Qcur, "Qcur", il);
 
                 kv_cmpr = ggml_reshape_3d(ctx0, kv_cmpr, kv_lora_rank, 1, n_tokens);
                 cb(kv_cmpr, "kv_cmpr_reshape", il);
 
                 // {n_embd_head_qk_rope + kv_lora_rank, 1, n_tokens}
-                ggml_tensor * Kcur = ggml_concat(ctx0, k_pe, kv_cmpr, 0);
+                ggml_tensor * Kcur = ggml_concat(ctx0, kv_cmpr, k_pe, 0);
                 cb(Kcur, "Kcur", il);
 
                 // {kv_lora_rank, 1, n_tokens}
--- a/src/models/deepseek2.cpp
+++ b/src/models/deepseek2.cpp
@@ -169,11 +169,10 @@ llm_build_deepseek2::llm_build_deepseek2(const llama_model & model, const llm_gr
                 Vcur = ggml_cont(ctx0, Vcur);
                 cb(Vcur, "Vcur_cont", il);
 
-                // note: rope must go first for in-place context shifting in build_rope_shift()
-                ggml_tensor * Qcur = ggml_concat(ctx0, q_pe, q_nope, 0);
+                ggml_tensor * Qcur = ggml_concat(ctx0, q_nope, q_pe, 0);
                 cb(Qcur, "Qcur", il);
 
-                ggml_tensor * Kcur = ggml_concat(ctx0, ggml_repeat(ctx0, k_pe, q_pe), k_nope, 0);
+                ggml_tensor * Kcur = ggml_concat(ctx0, k_nope, ggml_repeat(ctx0, k_pe, q_pe), 0);
                 cb(Kcur, "Kcur", il);
 
                 if (inp_attn_scale) {
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -6122,7 +6122,24 @@ struct test_flash_attn_ext : public test_case {
         ggml_tensor * k = create_permuted(type_KV,       hsk_padded, kv, nh,         nr23[1], true); // the K tensor is usually a view of the K cache
         ggml_set_name(k, "k");
 
-        ggml_tensor * v = create_permuted(type_KV,       hsv_padded, kv, nh,         nr23[1], true); // the V tensor is usually a view of the V cache
+        ggml_tensor * v = nullptr;
+        if (hsk_padded == 576 && hsv_padded == 512) {
+            // TODO: this branch should become a separate test case parameter instead of hardcoding this for these head shapes
+
+            // in this branch, the V cache is sub-view of the K cache. this is used by some MLA-based models
+            // for more info:
+            //   - https://github.com/ggml-org/llama.cpp/pull/13435
+            //   - https://github.com/ggml-org/llama.cpp/pull/18953#issuecomment-3774948392
+            v = ggml_view_4d(ctx, k, hsv_padded, kv, nh, nr23[1], k->nb[1], k->nb[2], k->nb[3], 0);
+
+            // note: this is the currently assumed layout by the CUDA FA implementation
+            //       however, this layout is problematic, because the offset can become very inconveniet for quantized KV types (i.e. not multiple of 16)
+            //       so the models/deepseek2.cpp implementation has been updated to concat the PE data at the end of the row, resulting in the layout above
+            //       the CUDA implementation has to be updated to not make an assumption abou the layout, and instead use the V tensor as it is
+            //v = ggml_view_4d(ctx, k, hsv_padded, kv, nh, nr23[1], k->nb[1], k->nb[2], k->nb[3], ggml_row_size(type_KV, hsk_padded - hsv_padded)/ggml_blck_size(type_KV));
+        } else {
+            v = create_permuted(type_KV,       hsv_padded, kv, nh,         nr23[1], true); // the V tensor is usually a view of the V cache
+        }
         ggml_set_name(v, "v");
 
         ggml_tensor * m = nullptr;
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -6130,13 +6130,8 @@ struct test_flash_attn_ext : public test_case {
             // for more info:
             //   - https://github.com/ggml-org/llama.cpp/pull/13435
             //   - https://github.com/ggml-org/llama.cpp/pull/18953#issuecomment-3774948392
+            //   - https://github.com/ggml-org/llama.cpp/pull/18986
             v = ggml_view_4d(ctx, k, hsv_padded, kv, nh, nr23[1], k->nb[1], k->nb[2], k->nb[3], 0);
-
-            // note: this is the currently assumed layout by the CUDA FA implementation
-            //       however, this layout is problematic, because the offset can become very inconveniet for quantized KV types (i.e. not multiple of 16)
-            //       so the models/deepseek2.cpp implementation has been updated to concat the PE data at the end of the row, resulting in the layout above
-            //       the CUDA implementation has to be updated to not make an assumption abou the layout, and instead use the V tensor as it is
-            //v = ggml_view_4d(ctx, k, hsv_padded, kv, nh, nr23[1], k->nb[1], k->nb[2], k->nb[3], ggml_row_size(type_KV, hsk_padded - hsv_padded)/ggml_blck_size(type_KV));
         } else {
             v = create_permuted(type_KV,       hsv_padded, kv, nh,         nr23[1], true); // the V tensor is usually a view of the V cache
         }
--- a/src/llama-graph.cpp
+++ b/src/llama-graph.cpp
@@ -1565,6 +1565,11 @@ ggml_tensor * llm_graph_context::build_attn_mha(
             v = ggml_transpose(ctx0, v);
         }
 
+        // TODO: update llama_kv_cache to not store V cache in the MLA case and automatically return a view of K
+        if (v_mla) {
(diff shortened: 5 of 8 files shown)


ggml/src/ggml-cuda/fattn-common.cuh
```diff
@@ -778,12 +778,15 @@ void launch_fattn(
...
-
     const ggml_tensor * Q = dst->src[0];
     const ggml_tensor * K = dst->src[1];
     const ggml_tensor * V = dst->src[2];
 
+    // TODO: make this more generic by removing the notion of "MLA".
+    //       for example "is V a view of K?" so we can skip loading it.
+    //       V strides should be driven by V itself and avoid assumption of the data layout
+    const bool is_mla = V->op == GGML_OP_VIEW && V->src[0] == K;
+
```
ggerganov, line 789: @JohannesGaessler Let me know if this is clear. The proposal is that when `V = ggml_view(K)`, the implementation can use this information, for example to avoid extra loads of data, etc. But technically, this is completely optional to do and the implementation can also just read from V directly.

This way, if the user code insists on passing different `V` data, then that should also work. In that case `V` won't be a view of `K` so it will be treated as a regular `V` tensor.

JohannesGaessler: You should adjust the kernel selection logic in `fattn.cu` to only use the MMA kernel if this boolean is true. The MMA kernel has the MLA-specific optimization of re-using the K data that was previously loaded for calculation of KQ as V data when calculating VKQ. But the tile kernel simply loads the data from the K and V pointers with no re-use. So for now I would suggest that we condition the use of the MMA kernel on this boolean and use the tile kernel as a fallback.

We could in principle compile multiple template specializations but currently they would be unused for real models.

ggml/src/ggml-cuda/fattn-mma-f16.cuh
```diff
@@ -794,7 +794,7 @@ static __device__ __forceinline__ void flash_attn_ext_f16_iter(
     // For MLA K and V have the same data.
     // Therefore, iterate over V in reverse and re-use the data if possible.
     static_assert(!mla || nstages <= 1, "combination of MLA and multi-stage loading not implemented");
-    constexpr int reusable_cutoff = mla ? (DKQ - 1) - (DKQ - 1) % (2*nbatch_K2) - (DKQ - DV) : DV;
+    constexpr int reusable_cutoff = mla ? (DV - 1) - (DV - 1) % (2*nbatch_K2) : DV;
```
ggerganov: I am not sure I completely understand the logic for reusing the K data, but to the best of my understanding this is the correct version now. Extra look is need to confirm.

To clarify, on `master`, the K/V data is layout like this in memory:

```bash
# K - 16 cells of K-only data
# S - 16 cells of shared data between both K and V

# a single latent head of 576 cells 
kkKKssssssssssssssssSSSSSSSSSSSSSSSS
```

In this PR, we change the layout so the K data is at the end of the row:

```bash
# K - 16 cells of K-only data
# S - 16 cells of shared data between both K and V

# a single latent head of 576 cells 
ssssssssssssssssSSSSSSSSSSSSSSSSkkKK
```

src/models/deepseek2.cpp
```diff
@@ -124,14 +124,14 @@ llm_build_deepseek2::llm_build_deepseek2(const llama_model & model, const llm_gr
 
                 // {n_embd_head_qk_rope + kv_lora_rank, n_head, n_tokens}
                 // note: rope must go first for in-place context shifting in build_rope_shift()
-                ggml_tensor * Qcur = ggml_concat(ctx0, q_pe, q_nope_absorbed, 0);
+                ggml_tensor * Qcur = ggml_concat(ctx0, q_nope_absorbed, q_pe, 0);
```
ngxson, line 127: a side-effect of this change is that `build_rope_shift` will no longer work correctly (may need to use a `view` when shifting)

ggml/src/ggml-cuda/fattn-mma-f16.cuh
JohannesGaessler: If we change the memory layout on master the loop structure in the MMA kernel needs to be done differently. On master the loop for K is going forward, the one for V is going backward. But if the reusable data is at the beginning the loop for K should be backward and the loop for V should be forward.

src/models/deepseek2.cpp
ggerganov: This should fix it:

```diff
diff --git a/src/llama-kv-cache.cpp b/src/llama-kv-cache.cpp
index fd9f97d52..2cc9efed8 100644
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -1614,10 +1614,10 @@ ggml_cgraph * llama_kv_cache::build_graph_shift(llm_graph_result * res, llama_co
 
         ggml_tensor * k =
             ggml_view_3d(ctx, layer.k,
-                n_embd_head_k, n_head_kv, get_size()*n_stream,
+                n_embd_head_k - hparams.n_lora_kv, n_head_kv, get_size()*n_stream,
                 ggml_row_size(layer.k->type, n_embd_head_k),
                 ggml_row_size(layer.k->type, n_embd_k_gqa),
-                0);
+                ggml_row_size(layer.k->type, hparams.n_lora_kv));
 
         ggml_tensor * cur = build_rope_shift(cparams, ctx, k, inp->k_shift, rope_factors, freq_base_l, freq_scale_l);
 
```

ggml/src/ggml-cuda/fattn-mma-f16.cuh
ggerganov: Hm, I am not sure I understand. Is this just for performance reasons? The current implementation works correctly and passes the tests.

ggerganov: Ah, I think I know what you are thinking. I updated the comment to better reflect the layout. So to clarify - the order of the shared data is the same, it's just shifted forward. And the K-only data is also ordered the same - it's just moved at the end of the buffer.

The same transform is also being applied to the Q data. So at the end, the dot products of QK and QKV are the same. Does that clarify?

JohannesGaessler: The MMA kernel uses only the K pointer which was meant as an assist for a more general implementation of deduplicating the KV cache for DeepSeek that never materialized. But it is also re-using the K data that is already in SRAM to avoid having to load it again. For that the loop over V is done in reverse because on consumer GPUs there simply isn't enough SRAM to hold the entire tile of K at once. So yes, that part is only for better performance.

src/models/deepseek2.cpp
ggerganov: The rope shift logic for PLM and minicpm3 has been broken since it didn't take into account the "nope" portion of the K embeddings. This is fixed now with [69d4fd7](https://github.com/ggml-org/llama.cpp/pull/18986/commits/69d4fd7e8fd8012a60f1e76dc412c71c7a2d2517)

ggerganov:
This should be the minimum changes needed to stabilize the CI and allow further work on the MLA code paths in `libllama` and the corresponding backend FA implementations. I've added some notes about the next steps.

ggml/src/ggml-cuda/fattn-mma-f16.cuh
JohannesGaessler:
```suggestion
    // constexpr int reusable_cutoff = mla ? (DV - 1) - (DV - 1) % (2*nbatch_K2) : DV;
    constexpr int reusable_cutoff = DV; // TODO implement properly
```

I'm not sure the way you implemented it is correct in general. I would suggest that for now we simply go with the safe option and I will make a follow-up PR.

tests/test-backend-ops.cpp
```diff
@@ -6122,7 +6122,24 @@ struct test_flash_attn_ext : public test_case {
...
+            // TODO: this branch should become a separate test case parameter instead of hardcoding this for these head shapes
+
+            // in this branch, the V cache is sub-view of the K cache. this is used by some MLA-based models
+            // for more info:
+            //   - https://github.com/ggml-org/llama.cpp/pull/13435
+            //   - https://github.com/ggml-org/llama.cpp/pull/18953#issuecomment-3774948392
+            v = ggml_view_4d(ctx, k, hsv_padded, kv, nh, nr23[1], k->nb[1], k->nb[2], k->nb[3], 0);
+
+            // note: this is the currently assumed layout by the CUDA FA implementation
+            //       however, this layout is problematic, because the offset can become very inconveniet for quantized KV types (i.e. not multiple of 16)
```
JohannesGaessler: I'm not sure what you mean here. Moving the K-only data to the front does not fix any issues with memory alignment. On master the check for memory alignment causes quantized KV cache to fall back to CPU but that is a bug with the check, see https://github.com/ggml-org/llama.cpp/pull/19023 .

ggerganov: The comment is no longer relevant - I removed it.

For reference, the problem is that the Vulkan backend does not support views with `offset % 16 != 0`. At least when I tried, it asserted here:

https://github.com/ggml-org/llama.cpp/blob/3c5151387c33473da13271068b0ec2c03b3fd977/ggml/src/ggml-vulkan/ggml-vulkan.cpp#L6084-L6110

But this is no longer relevant, because we change the memory layout in this PR and we no longer need to offset the view.

JohannesGaessler (approved):

LostRuins:
I have noticed a coherency regression when used with GLM-4-32B-0414 when using context shifting (`llama_memory_seq_rm` + `llama_memory_seq_add`) after merging this PR. Details at https://github.com/ggml-org/llama.cpp/issues/19292

ggerganov:
Could you provide logs during the loading with the model parameters (head size, n_rot, etc)? It's not obvious why this model would break from the changes here.

LostRuins:
Sure, here's a long of loading the model with vulkan

```
ggml_vulkan: 0 = Intel(R) RaptorLake-S Mobile Graphics Controller (Intel Corporation) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 32768 | int dot: 1 | matrix cores: none
ggml_vulkan: 1 = NVIDIA GeForce RTX 4090 Laptop GPU (NVIDIA) | uma: 0 | fp16: 1 | bf16: 1 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: NV_coopmat2
load_backend: loaded Vulkan backend from C:\Users\user\Desktop\llama-b7921-bin-win-vulkan-x64\ggml-vulkan.dll
load_backend: loaded CPU backend from C:\Users\user\Desktop\llama-b7921-bin-win-vulkan-x64\ggml-cpu-haswell.dll
build: 7921 (e9a859db3) with Clang 19.1.5 for Windows x86_64
main: llama backend init
main: load the model and apply lora adapter, if any
llama_params_fit_impl: projected to use 19583 MiB of device memory vs. 15278 MiB of free device memory
llama_params_fit_impl: cannot meet free memory target of 1024 MiB, need to reduce device memory by 5328 MiB
llama_params_fit_impl: context size reduced from 32768 to 4096 -> need 1708 MiB less memory in total
llama_params_fit_impl: filling dense layers back-to-front:
llama_params_fit_impl:   - Vulkan1 (NVIDIA GeForce RTX 4090 Laptop GPU): 49 layers,  14204 MiB used,   1073 MiB free
llama_params_fit: fitting params to free memory took 0.25 seconds
llama_model_load_from_file_impl: using device Vulkan1 (NVIDIA GeForce RTX 4090 Laptop GPU) (0000:01:00.0) - 15278 MiB free
llama_model_loader: loaded meta data with 44 key-value pairs and 613 tensors from D:\ExtDrive\KoboldCpp\models\GLM-4-32B-0414-Q4_K_S.gguf (version GGUF V3 (latest))
llama_model_loader: Dumping metadata keys/values. Note: KV overrides do not apply in this output.
llama_model_loader: - kv   0:                       general.architecture str              = glm4
llama_model_loader: - kv   1:                               general.type str              = model
llama_model_loader: - kv   2:                               general.name str              = Glm-4-32B-0414
llama_model_loader: - kv   3:                            general.version str              = 0414
llama_model_loader: - kv   4:                           general.basename str              = Glm-4-32B-0414
llama_model_loader: - kv   5:                       general.quantized_by str              = Unsloth
llama_model_loader: - kv   6:                         general.size_label str              = 32B
llama_model_loader: - kv   7:                            general.license str              = mit
llama_model_loader: - kv   8:                           general.repo_url str              = https://huggingface.co/unsloth
llama_model_loader: - kv   9:                   general.base_model.count u32              = 1
llama_model_loader: - kv  10:                  general.base_model.0.name str              = GLM 4 32B 0414
llama_model_loader: - kv  11:               general.base_model.0.version str              = 0414
llama_model_loader: - kv  12:          general.base_model.0.organization str              = THUDM
llama_model_loader: - kv  13:              general.base_model.0.repo_url str              = https://huggingface.co/THUDM/GLM-4-32...
llama_model_loader: - kv  14:                               general.tags arr[str,2]       = ["unsloth", "text-generation"]
llama_model_loader: - kv  15:                          general.languages arr[str,2]       = ["zh", "en"]
llama_model_loader: - kv  16:                           glm4.block_count u32              = 61
llama_model_loader: - kv  17:                        glm4.context_length u32              = 32768
llama_model_loader: - kv  18:                      glm4.embedding_length u32              = 6144
llama_model_loader: - kv  19:                   glm4.feed_forward_length u32              = 23040
llama_model_loader: - kv  20:                  glm4.attention.head_count u32              = 48
llama_model_loader: - kv  21:               glm4.attention.head_count_kv u32              = 2
llama_model_loader: - kv  22:                        glm4.rope.freq_base f32              = 10000.000000
llama_model_loader: - kv  23:      glm4.attention.layer_norm_rms_epsilon f32              = 0.000010
llama_model_loader: - kv  24:                  glm4.attention.key_length u32              = 128
llama_model_loader: - kv  25:                glm4.attention.value_length u32              = 128
llama_model_loader: - kv  26:                  glm4.rope.dimension_count u32              = 64
llama_model_loader: - kv  27:                       tokenizer.ggml.model str              = gpt2
llama_model_loader: - kv  28:                         tokenizer.ggml.pre str              = glm4
llama_model_loader: - kv  29:                      tokenizer.ggml.tokens arr[str,151552]  = ["!", "\"", "#", "$", "%", "&", "'", ...
llama_model_loader: - kv  30:                  tokenizer.ggml.token_type arr[i32,151552]  = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, ...
llama_model_loader: - kv  31:                      tokenizer.ggml.merges arr[str,318088]  = ["Ġ Ġ", "Ġ ĠĠĠ", "ĠĠ ĠĠ", "...
llama_model_loader: - kv  32:                tokenizer.ggml.eos_token_id u32              = 151336
llama_model_loader: - kv  33:            tokenizer.ggml.padding_token_id u32              = 151330
llama_model_loader: - kv  34:                tokenizer.ggml.eot_token_id u32              = 151336
llama_model_loader: - kv  35:            tokenizer.ggml.unknown_token_id u32              = 151329
llama_model_loader: - kv  36:                tokenizer.ggml.bos_token_id u32              = 151329
llama_model_loader: - kv  37:                    tokenizer.chat_template str              = [gMASK]<sop>\n{%- if tools -%}\n<|syste...
llama_model_loader: - kv  38:               general.quantization_version u32              = 2
llama_model_loader: - kv  39:                          general.file_type u32              = 14
llama_model_loader: - kv  40:                      quantize.imatrix.file str              = GLM-4-32B-0414-GGUF/imatrix_unsloth.dat
llama_model_loader: - kv  41:                   quantize.imatrix.dataset str              = unsloth_calibration_GLM-4-32B-0414.txt
llama_model_loader: - kv  42:             quantize.imatrix.entries_count u32              = 366
llama_model_loader: - kv  43:              quantize.imatrix.chunks_count u32              = 724
llama_model_loader: - type  f32:  245 tensors
llama_model_loader: - type q4_K:  356 tensors
llama_model_loader: - type q5_K:   11 tensors
llama_model_loader: - type q6_K:    1 tensors
print_info: file format = GGUF V3 (latest)
print_info: file type   = Q4_K - Small
print_info: file size   = 17.40 GiB (4.59 BPW)
load: 0 unused tokens
load: special_eos_id is not in special_eog_ids - the tokenizer config may be incorrect
load: printing all EOG tokens:
load:   - 151329 ('<|endoftext|>')
load:   - 151336 ('<|user|>')
load: special tokens cache size = 14
load: token to piece cache size = 0.9710 MB
print_info: arch                  = glm4
print_info: vocab_only            = 0
print_info: no_alloc              = 0
print_info: n_ctx_train           = 32768
print_info: n_embd                = 6144
print_info: n_embd_inp            = 6144
print_info: n_layer               = 61
print_info: n_head                = 48
print_info: n_head_kv             = 2
print_info: n_rot                 = 64
print_info: n_swa                 = 0
print_info: is_swa_any            = 0
print_info: n_embd_head_k         = 128
print_info: n_embd_head_v         = 128
print_info: n_gqa                 = 24
print_info: n_embd_k_gqa          = 256
print_info: n_embd_v_gqa          = 256
print_info: f_norm_eps            = 0.0e+00
print_info: f_norm_rms_eps        = 1.0e-05
print_info: f_clamp_kqv           = 0.0e+00
print_info: f_max_alibi_bias      = 0.0e+00
print_info: f_logit_scale         = 0.0e+00
print_info: f_attn_scale          = 0.0e+00
print_info: n_ff                  = 23040
print_info: n_expert              = 0
print_info: n_expert_used         = 0
print_info: n_expert_groups       = 0
print_info: n_group_used          = 0
print_info: causal attn           = 1
print_info: pooling type          = 0
print_info: rope type             = 0
print_info: rope scaling          = linear
print_info: freq_base_train       = 10000.0
print_info: freq_scale_train      = 1
print_info: n_ctx_orig_yarn       = 32768
print_info: rope_yarn_log_mul     = 0.0000
print_info: rope_finetuned        = unknown
print_info: model type            = 32B
print_info: model params          = 32.57 B
print_info: general.name          = Glm-4-32B-0414
print_info: vocab type            = BPE
print_info: n_vocab               = 151552
print_info: n_merges              = 318088
print_info: BOS token             = 151329 '<|endoftext|>'
print_info: EOS token             = 151336 '<|user|>'
print_info: EOT token             = 151336 '<|user|>'
print_info: UNK token             = 151329 '<|endoftext|>'
print_info: PAD token             = 151330 '[MASK]'
print_info: LF token              = 198 'Ċ'
print_info: EOG token             = 151329 '<|endoftext|>'
print_info: EOG token             = 151336 '<|user|>'
print_info: max token length      = 1024
load_tensors: loading model tensors, this can take a while... (mmap = true, direct_io = false)
load_tensors: offloading output layer to GPU
load_tensors: offloading 48 repeating layers to GPU
load_tensors: offloaded 49/62 layers to GPU
load_tensors:   CPU_Mapped model buffer size =  4129.59 MiB
load_tensors:      Vulkan1 model buffer size = 13692.96 MiB
................................................................................................
common_init_result: added <|endoftext|> logit bias = -inf
common_init_result: added <|user|> logit bias = -inf
llama_context: constructing llama_context
llama_context: n_seq_max     = 1
llama_context: flash_attn    = auto
llama_context: kv_unified    = false
llama_context: freq_base     = 10000.0
llama_context: n_ctx_seq (4096) < n_ctx_train (32768) -- the full capacity of the model will not be utilized
llama_context: Vulkan_Host  output buffer size =     0.58 MiB
llama_kv_cache:        CPU KV buffer size =    52.00 MiB
llama_kv_cache:    Vulkan1 KV buffer size =   192.00 MiB
llama_kv_cache: size =  244.00 MiB (  4096 cells,  61 layers,  1/1 seqs), K (f16):  122.00 MiB, V (f16):  122.00 MiB
sched_reserve: reserving ...
sched_reserve: Flash Attention was auto, set to enabled
sched_reserve:    Vulkan1 compute buffer size =   320.03 MiB
sched_reserve: Vulkan_Host compute buffer size =    32.01 MiB
sched_reserve: graph nodes  = 2081
sched_reserve: graph splits = 158 (with bs=512), 2 (with bs=1)
sched_reserve: reserve took 289.36 ms, sched copies = 1
```

To clarify: the inference itself is fine. The regression only happens after a KV shift is applied, like so

```
//extract the unwanted tokens out from context and KV from trimstart

int diff = found - trimstart;
llama_memory_seq_rm(llama_get_memory(ctx), 0, trimstart, trimstart + diff);
llama_memory_seq_add(llama_get_memory(ctx), 0, trimstart + diff, -1, -diff);
```

if no shifting is performed, there inference both before and after this PR is normal. If shifting is performed, there is a large degradation observed after this specific PR.

LostRuins:
Also I suspect it's not just this old model but also GLM-4.7-Flash and possibly others affected, but as GLM-4.7-Flash was going through many other separate unrelated fixes I didn't want to use that as the model.

ggerganov:
Does it work correctly with CPU-only run?

LostRuins:
Surprisingly, Pure CPU takes forever, but it works correctly!

Shifting produces the expected results both before and after this PR with pure CPU. Which feels even stranger. 

I did a few more tests:

Pure CPU = Shifting works fine. Super, super slow PP. Fully coherent.
Vulkan with FA and 0 layers offloaded = Shifting works fine, Output is fully coherent. Total VRAM usage at 0 layers is 1.1GB.
Vulkan with FA and 10/62 layers offloaded = Mild incoherence. Example of the mild incoherence:
```
It sparkled with a sense of fear and fear. He noticed a large crowd gathered near the edge of the forest, huddled together, casting long, dancing shadows that danced away from the flickering torches that lined the forest's edge. 
```
Vulkan with FA and 50/62 layers offloaded = Very incoherent. Example of extreme incoherence:
```
When he reached the middle of the clearing field that stretched out in front of him began to look like it was part of the town map had been pushed slightly to the east now, and now he would see the figure-shapedI-shaped man walked to the ground where the lights were coming out of the ground, which was a short spursky, but he could not make himself
```


Trying without FA, same result, coherence is related to how many layers are offloaded. At 50 layers:
```
the smell of the town turned into the main road ended up to the west of the town for a long time, making sure that he was no longer behind the other man in the town's watch, which was inky, checking the time to check if he could get some water from the fountain near the spiggetreenged the way he had been there for a while
```

ggerganov:
Seems like an issue in the Vulkan backend - track #19296 for more info.

PR #18995, opened 2026-01-21
jinja: support none|string

### LINKED ISSUE #18974: Eval bug: Inference fails with certain inputs for Granite and Nemotron
### Name and Version

b7783

### Operating systems

Linux

### GGML backends

Vulkan

### Hardware

Intel ARC Pro B50
AMD RX 7600XT

### Models

unsloth/Nemotron-3-Nano-30B-A3B-UD-Q4_K_XL.gguf
unsloth/granite-4.0-h-small-UD-Q4_K_XL.gguf

Pretty simple prompt - I'm copying and pasting the llama-server README.md file into a prompt and asking the model a simple question about it ("explain adaptive-p sampling"). This is happening in Open WebUI and in the llama.cpp web UI.

Nemotron fails immediately with this error message in Open WebUI:

```
------------
While executing FilterExpression at line 8, column 66 in source:
...~ json_key ~ '>' ~ (json_dict[json_key] | string) ~ '</' ~ json_key ~ '>' }}↵   ...
                                           ^
Error: Unknown (built-in) filter 'string' for type None
```

Granite starts giving me gibberish:

```
cision (I’m literally the same, but I’m not sure if you’re aware of the way to do it. I’m not sure if you’re aware of the way to do it, but you’ll have to download [I stopped it at this point]
```

Running this with `unsloth/LFM2.5-1.2B-Instruct-UD-Q6_K_XL.gguf` works just fine.

### First Bad Commit

Not sure

### Relevant log output

Pasted above. Not seeing any other relevant log info

pwilkin:
Would partially fix #18974

### DIFF
--- a/common/jinja/value.cpp
+++ b/common/jinja/value.cpp
@@ -1005,7 +1005,7 @@ const func_builtins & value_none_t::get_builtins() const {
     static const func_builtins builtins = {
         {"default", default_value},
         {"tojson", tojson},
-        {"string", [](const func_args &) -> value { return mk_val<value_string>(""); }}
+        {"string", [](const func_args &) -> value { return mk_val<value_string>("None"); }}
     };
     return builtins;
 }
--- a/common/jinja/value.cpp
+++ b/common/jinja/value.cpp
@@ -1005,6 +1005,7 @@ const func_builtins & value_none_t::get_builtins() const {
     static const func_builtins builtins = {
         {"default", default_value},
         {"tojson", tojson},
+        {"string", [](const func_args &) -> value { return mk_val<value_string>(""); }}
     };
     return builtins;
 }
--- a/tests/test-jinja.cpp
+++ b/tests/test-jinja.cpp
@@ -613,7 +613,7 @@ static void test_filters(testing & t) {
     test_template(t, "none to string",
         "{{ x|string }}",
         {{"x", nullptr}},
-        ""
+        "None"
     );
 }
 
--- a/tests/test-jinja.cpp
+++ b/tests/test-jinja.cpp
@@ -609,6 +609,12 @@ static void test_filters(testing & t) {
         json::object(),
         "hello"
     );
+
+    test_template(t, "none to string",
+        "{{ x|string }}",
+        {{"x", nullptr}},
+        ""
+    );
 }
 
 static void test_literals(testing & t) {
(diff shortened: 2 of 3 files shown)


common/jinja/value.cpp
```diff
@@ -1005,6 +1005,7 @@ const func_builtins & value_none_t::get_builtins() const {
     static const func_builtins builtins = {
         {"default", default_value},
         {"tojson", tojson},
+        {"string", [](const func_args &) -> value { return mk_val<value_string>(""); }}
```
CISC:
```suggestion
        {"string", [](const func_args &) -> value { return mk_val<value_string>("None"); }}
```

tests/test-jinja.cpp
```diff
@@ -609,6 +609,12 @@ static void test_filters(testing & t) {
         json::object(),
         "hello"
     );
+
+    test_template(t, "none to string",
+        "{{ x|string }}",
+        {{"x", nullptr}},
+        ""
```
CISC:
```suggestion
        "None"
```

CISC (approved):
You should always test with `test-jinja -py`. :)

CISC:
One could argue that `value_none_t` should have `as_string` too, and I think there are some other places that don't treat this correctly.

ngxson:
> One could argue that `value_none_t` should have `as_string` too, and I think there are some other places that don't treat this correctly.

I think `as_string` is currently never used in `none`, so no strong opinion if it should be added. In most places, we have logic to check if input is null, and if yes, we just either skip it or throw an exception.

ngxson (approved):

CISC:
> > One could argue that `value_none_t` should have `as_string` too, and I think there are some other places that don't treat this correctly.
> 
> I think `as_string` is currently never used in `none`, so no strong opinion if it should be added. In most places, we have logic to check if input is null, and if yes, we just either skip it or throw an exception.

Here's the one I was thinking of:
https://github.com/ggml-org/llama.cpp/blob/0be21c4986518194c2e2cc583a921075ca808017/common/jinja/runtime.h#L574-L575

ngxson:
not quite remember why it doesn't work, but IIRC because some statements are treated as functions that returns `None`, so sometimes `None` appears out of nowhere in the rendered content. but yeah, currently I think this simple case: `{{ None }}` won't match what jinja2 would render

CISC:
> not quite remember why it doesn't work, but IIRC because some statements are treated as functions that returns `None`, so sometimes `None` appears out of nowhere in the rendered content. but yeah, currently I think this simple case: `{{ None }}` won't match what jinja2 would render

Ah, worth looking into at some point, would have been useful to know which template. :)

PR #18998, opened 2026-01-21
mtmd: add `mtmd_get_output_embd_size`

giladgd:
It appears that `llama_model_n_embd(model) * mtmd_input_chunk_get_n_tokens(chunk)` is no longer valid to calculate the size of the mtmd output embedding, so this PR adds another method to get the size of the embedding directly.


ngxson:
This is not a valid solution. You should use `llama_model_n_embd_inp` instead

PR #19012, opened 2026-01-22
ggml-cpu: Use tiled FA for prompt-processing

am17an:
the FA performance is gimped on CPU for long contexts because it essentially uses a vector kernel. This PR adds a tiled FA for PP. Perf tuning for tile sizes done on a AMD EPYC single-socket 64-c machine. The code is kept fairly simple to leave room for incremental optimizations. According to perf, most of the time is spent in `ggml_vec_dot_f16` and `ggml_vec_mad_f32`, and about ~10% of the time in `ggml_lookup_f16_to_f32` 

| Model                 | Test         |   t/s a14b960bc |   t/s tile-fa-cpu |   Speedup |
|:----------------------|:-------------|----------------:|------------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512        |          237.25 |            244.39 |      1.03 |
| gpt-oss 20B MXFP4 MoE | pp512@d1024  |          205.31 |            224.61 |      1.09 |
| gpt-oss 20B MXFP4 MoE | pp512@d2048  |          171.80 |            209.63 |      1.22 |
| gpt-oss 20B MXFP4 MoE | pp512@d4096  |          134.60 |            185.10 |      1.38 |
| gpt-oss 20B MXFP4 MoE | pp512@d8192  |           59.69 |            143.43 |      2.40 |
| gpt-oss 20B MXFP4 MoE | pp512@d16384 |           25.12 |             97.29 |      3.87 |
| llama 8B Q4_K_M       | pp512        |          201.83 |            199.66 |      0.99 |
| llama 8B Q4_K_M       | pp512@d1024  |          160.35 |            175.60 |      1.10 |
| llama 8B Q4_K_M       | pp512@d2048  |          134.40 |            158.26 |      1.18 |
| llama 8B Q4_K_M       | pp512@d4096  |           57.76 |            130.85 |      2.27 |
| llama 8B Q4_K_M       | pp512@d8192  |           28.01 |             95.79 |      3.42 |
| llama 8B Q4_K_M       | pp512@d16384 |           14.24 |             62.50 |      4.39 |

TODO:
- [ ] perf tuning on ARM

### DIFF
--- a/ggml/src/ggml-cpu/ggml-cpu.c
+++ b/ggml/src/ggml-cpu/ggml-cpu.c
@@ -2871,8 +2871,8 @@ struct ggml_cplan ggml_graph_plan(
                         const int64_t DV = node->src[2]->ne[0];
 
                         // Tiled flash attention scratch (tile sizes defined in common.h)
-                        // Per-thread: Q_q + KQ + VKQ32 + V32 + padding
-                        cur = sizeof(float)*(GGML_FA_TILE_Q*DK + GGML_FA_TILE_Q*GGML_FA_TILE_KV + GGML_FA_TILE_Q*DV + GGML_FA_TILE_KV*DV + CACHE_LINE_SIZE_F32)*n_tasks;
+                        // Per-thread: Q_q + KQ + mask + VKQ32 + V32 + padding
+                        cur = sizeof(float)*(GGML_FA_TILE_Q*DK + 2*GGML_FA_TILE_Q*GGML_FA_TILE_KV + GGML_FA_TILE_Q*DV + GGML_FA_TILE_KV*DV)*n_tasks;
                     } break;
                 case GGML_OP_FLASH_ATTN_BACK:
                     {
--- a/ggml/src/ggml-cpu/ggml-cpu.c
+++ b/ggml/src/ggml-cpu/ggml-cpu.c
@@ -14,6 +14,7 @@
 #include "vec.h"
 #include "ops.h"
 #include "ggml.h"
+#include "common.h"
 
 #if defined(_MSC_VER) || defined(__MINGW32__)
 #include <malloc.h> // using malloc.h with MSC/MINGW
@@ -2866,10 +2867,12 @@ struct ggml_cplan ggml_graph_plan(
                     } break;
                 case GGML_OP_FLASH_ATTN_EXT:
                     {
-                        const int64_t ne10 = node->src[1]->ne[0]; // DK
-                        const int64_t ne20 = node->src[2]->ne[0]; // DV
+                        const int64_t DK = node->src[1]->ne[0];
+                        const int64_t DV = node->src[2]->ne[0];
 
-                        cur = sizeof(float)*(1*ne10 + 2*ne20)*n_tasks; // 1x head size K + 2x head size V (per thread)
+                        // Tiled flash attention scratch (tile sizes defined in common.h)
+                        // Per-thread: Q_q + KQ + VKQ32 + V32 + padding
+                        cur = sizeof(float)*(GGML_FA_TILE_Q*DK + GGML_FA_TILE_Q*GGML_FA_TILE_KV + GGML_FA_TILE_Q*DV + GGML_FA_TILE_KV*DV + CACHE_LINE_SIZE_F32)*n_tasks;
                     } break;
                 case GGML_OP_FLASH_ATTN_BACK:
                     {
--- a/ggml/src/ggml-cpu/ops.cpp
+++ b/ggml/src/ggml-cpu/ops.cpp
@@ -8394,8 +8394,6 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
 
         memset(VKQ32, 0, Q_TILE_SZ * DV * sizeof(float));
 
-        const ggml_fp16_t * mp = mask ? (ggml_fp16_t *)((char *) mask->data + iq1*mask->nb[1] + (iq2%mask->ne[2])*mask->nb[2] + (iq3%mask->ne[3])*mask->nb[3]) : NULL;
-
         // k indices
         const int ik3 = iq3 / rk3;
         const int ik2 = iq2 / rk2;
@@ -8414,6 +8412,26 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
         }
 
         for (int64_t ic = 0; ic < nek1; ic += KV_TILE_SZ) {
+
+            // skip the tile entirely if all the masks are -inf
+            if (mask) {
+                bool can_skip = true;
+                for (int tq = 0; tq < tile_rows; tq++) {
+                    const ggml_fp16_t * mp_row = (const ggml_fp16_t *)((const char *) mask->data + (iq1 + tq)*mask->nb[1] + (iq2%mask->ne[2])*mask->nb[2] + (iq3%mask->ne[3])*mask->nb[3]);
+                    for (int tk = 0; tk < KV_TILE_SZ; tk++) {
+                        const float mv = slope * GGML_CPU_FP16_TO_FP32(mp_row[ic + tk]);
+                        if (mv != -INFINITY) {
+                            can_skip = false;
+                            break;
+                        }
+                    }
+                }
+
+                if (can_skip) {
+                    continue;
+                }
+            }
+
             for (int tq = 0; tq < Q_TILE_SZ; tq++) {
                 const void * q_row = (const char *)Q_q + tq * DK * kv_type_size;
                 for (int tk = 0; tk < KV_TILE_SZ; tk++) {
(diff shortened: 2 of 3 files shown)


ggml/src/ggml-cpu/ops.cpp
```diff
@@ -8271,6 +8272,251 @@ static void ggml_compute_forward_flash_attn_ext_f16_one_chunk(
...
+            if (mp) {
+                for (int tq = 0; tq < Q_TILE_SZ; tq++) {
+                    const ggml_fp16_t * mp_row = (const ggml_fp16_t *)((const char *) mask->data + (iq1 + tq)*mask->nb[1] + (iq2%mask->ne[2])*mask->nb[2] + (iq3%mask->ne[3])*mask->nb[3]);
+                    for (int tk = 0; tk < KV_TILE_SZ; tk++) {
+                        const float mv = slope * GGML_CPU_FP16_TO_FP32(mp_row[ic + tk]);
+                        KQ[tq * KV_TILE_SZ + tk] += mv;
+                    }
+                }
+            }
+
```
ggerganov, line 8455: It might be worth to pre-load the mask tile at the start of the loop and skip the iteration if all values are -INF. This helps for unified kv cache with multiple sequences. Probably you should see a benefit in the prompt processing at larger `-np` with this command:

```bash
llama-batched-bench -m [model] -c 33792 -npp 2048 -ntg 32 -npl 1,2,4 -kvu
```

Can be done in follow-up PR - just leave a TODO for now.

ggml/src/ggml-cpu/ops.cpp
```diff
@@ -8271,6 +8272,251 @@ static void ggml_compute_forward_flash_attn_ext_f16_one_chunk(
...
+            if (logit_softcap != 0.0f) {
+                ggml_vec_tanh_f32(Q_TILE_SZ * KV_TILE_SZ, KQ, KQ);
+                ggml_vec_scale_f32(Q_TILE_SZ * KV_TILE_SZ, KQ, logit_softcap);
+            }
+
+            if (mp) {
+                for (int tq = 0; tq < Q_TILE_SZ; tq++) {
```
ggerganov: Fixes out of bounds access:

```suggestion
                for (int tq = 0; tq < tile_rows; tq++) {
```

am17an:
Not sure why the server test case is crashing backend sampling on

ggerganov:
I can reproduce locally with a CPU-only build. I think there is a bug in the FA kernel that produces nan in some cases.

```diff
diff --git a/ggml/src/ggml-cpu/ops.cpp b/ggml/src/ggml-cpu/ops.cpp
index e4228f232..3f5f97945 100644
--- a/ggml/src/ggml-cpu/ops.cpp
+++ b/ggml/src/ggml-cpu/ops.cpp
@@ -8511,6 +8511,10 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
 
             // permute(0, 2, 1, 3)
             memcpy((char *) dst->data + (i3*ne2*ne1 + i2 + i1*ne1)*nb1, VKQ32 + tq * DV, nb1);
+
+            for (int k = 0; k < DV; k++) {
+                GGML_ASSERT(!isnan(VKQ32[tq * DV + k]));
+            }
         }
 
         ir += tile_rows;
```

Server test that fails:

```bash
LLAMA_ARG_BACKEND_SAMPLING=1 LLAMA_LOG_VERBOSITY=10 LLAMA_SERVER_BIN_PATH=/path/to/build-cpu/bin/llama-server ./tests.sh -s unit/test_completion.py::test_completion_prompt_cache
```

Haven't traced what is the cause yet.

am17an:
@ggerganov the failure was actually what would be the optimization you mentioned anyway. When the row is all masked, we were accumulating V still which leads to NaNs. I added the change to skip those rows and confirmed locally that the test passes

ggml/src/ggml-cpu/ops.cpp
am17an: actually maybe you meant we can skip the entire tile?

ggerganov: Yes, the idea is to load the mask tile before the K*Q product and `continue` the entire `ic` iteration if the mask tile it's full of -INF.

This corresponds to this optimization in the vec kernel:

https://github.com/ggml-org/llama.cpp/blob/81ab64f3c858c0db8c7c3a6bccd4cbbe624f52a3/ggml/src/ggml-cpu/ops.cpp#L8167-L8172

ggml/src/ggml-cpu/ops.cpp
```diff
@@ -8271,6 +8272,265 @@ static void ggml_compute_forward_flash_attn_ext_f16_one_chunk(
...
+        float * base  = (float *) params->wdata + ith*(Q_TILE_SZ*DK + Q_TILE_SZ*KV_TILE_SZ + Q_TILE_SZ*DV + KV_TILE_SZ*DV + CACHE_LINE_SIZE_F32);
+
+        void  * Q_q    = base;
+        float * KQ     = (float *)((char *)base + Q_TILE_SZ * DK * sizeof(float));
+        float * VKQ32  = KQ + Q_TILE_SZ * KV_TILE_SZ;
+        float * V32    = VKQ32 + Q_TILE_SZ * DV;  // F32 buffer for V tile
+
+        memset(VKQ32, 0, Q_TILE_SZ * DV * sizeof(float));
+
+        const ggml_fp16_t * mp = mask ? (ggml_fp16_t *)((char *) mask->data + iq1*mask->nb[1] + (iq2%mask->ne[2])*mask->nb[2] + (iq3%mask->ne[3])*mask->nb[3]) : NULL;
```
ggerganov: This pointer is unused since we recalculate the mask row pointer on line 8434.

ggml/src/ggml-cpu/ops.cpp
```diff
@@ -8271,6 +8272,283 @@ static void ggml_compute_forward_flash_attn_ext_f16_one_chunk(
...
+                            can_skip = false;
+                            break;
+                        }
+                    }
+                }
+
+                if (can_skip) {
+                    continue;
+                }
+            }
```
ggerganov, line 8435: Wouldn't it be better to store the converted F32 mask tile in the work data and avoid converting F16->FP32 below again on line 8454?

am17an: Yes that would be better. I was just being conservative with the work size

ggml/src/ggml-cpu/ggml-cpu.c
```diff
@@ -2866,10 +2867,12 @@ struct ggml_cplan ggml_graph_plan(
...
                     {
-                        const int64_t ne10 = node->src[1]->ne[0]; // DK
-                        const int64_t ne20 = node->src[2]->ne[0]; // DV
+                        const int64_t DK = node->src[1]->ne[0];
+                        const int64_t DV = node->src[2]->ne[0];
 
-                        cur = sizeof(float)*(1*ne10 + 2*ne20)*n_tasks; // 1x head size K + 2x head size V (per thread)
+                        // Tiled flash attention scratch (tile sizes defined in common.h)
+                        // Per-thread: Q_q + KQ + VKQ32 + V32 + padding
+                        cur = sizeof(float)*(GGML_FA_TILE_Q*DK + GGML_FA_TILE_Q*GGML_FA_TILE_KV + GGML_FA_TILE_Q*DV + GGML_FA_TILE_KV*DV + CACHE_LINE_SIZE_F32)*n_tasks;
```
ggerganov: The cache line size is automatically taken into account at the end of this function on line 2911, so no need to add it here too.

ggml/src/ggml-cpu/ops.cpp
```diff
@@ -8271,6 +8272,279 @@ static void ggml_compute_forward_flash_attn_ext_f16_one_chunk(
...
+                    S[tq] *= ms;
+                }
+                M[tq] = Mnew;
+
+                if (tile_max == -INFINITY) {
+                    skip[tq] = true;
+                    continue;
+                }
+
+                S[tq] += ggml_vec_soft_max_f32(KV_TILE_SZ, kq_row, kq_row, Mnew);
```
ggerganov, line 8480: I think this is slightly better:

```suggestion
                float tile_max;
                ggml_vec_max_f32(KV_TILE_SZ, &tile_max, kq_row);

                if (tile_max == -INFINITY) {
                    skip[tq] = true;
                    continue;
                }

                const float Mold = M[tq];
                const float Mnew = fmaxf(Mold, tile_max);

                if (Mnew > Mold) {
                    const float ms = expf(Mold - Mnew);
                    ggml_vec_scale_f32(DV, VKQ32 + tq * DV, ms);
                    S[tq] *= ms;
                    M[tq] = Mnew;
                }

                S[tq] += ggml_vec_soft_max_f32(KV_TILE_SZ, kq_row, kq_row, Mnew);
```

ggerganov (approved):

PR #19014, opened 2026-01-22
server : add option to skip chat template verification

Myp3a:
This PR adds an ability to skip chat template verification on internal chat presets with `--no-chat-template-verify` argument. 

Tested:
- Template with broken syntax: server exits with an error independently of the new argument;
- Template with strict checks: server exits with an error by default, allows the usage if the argument is specified;
- Running server, invalid request for the template: 500 error with a Jinja error (expected);
- Running server, valid request: proper response.

Should fix #18895


ngxson:
This is not a good solution. Better to fix the template or the handling instead, see `chat.cpp`

PR #19017, opened 2026-01-22
timegm api absent in AIX

mehendarkarprajwal:
The timegm api is not part of AIX libc. Hence we want to use the alternative available ie. mktime

### DIFF
--- a/vendor/cpp-httplib/httplib.cpp
+++ b/vendor/cpp-httplib/httplib.cpp
@@ -117,6 +117,8 @@ time_t parse_http_date(const std::string &date_str) {
 
 #ifdef _WIN32
   return _mkgmtime(&tm_buf);
+#elif defined _AIX
+  return mktime(&tm_buf);
 #else
   return timegm(&tm_buf);
 #endif


taronaeo:
This PR should be directed to the vendor's repository. Once it is accepted upstream, we will re-sync it here with a vendor file update.

I don't think I can accept this PR.

taronaeo:
Here's a quicklink to the repo: https://github.com/yhirose/cpp-httplib

mehendarkarprajwal:
Thank you @taronaeo !!
I had raised a PR in that repo also. Looks like it is merged now -> https://github.com/yhirose/cpp-httplib/pull/2336
Ok will wait for re-sync of cpp-httplib here.

taronaeo:
Great! I will close this PR and create another to sync the vendor files. Will ref this PR.

taronaeo:
Sorry it looks like we're using tagged releases for `cpp-httplib` and will need to wait for the next release to be out from their end. Feel free to ping me again if you notice a new version is out.

https://github.com/ggml-org/llama.cpp/blob/a14b960bc70a0b48405409bbe3e0d6238473a0f8/scripts/sync_vendor.py#L15-L16

PR #19019, opened 2026-01-22
chat : fix translategemma crash on common_chat_format_example

ngxson:
fix https://github.com/ggml-org/llama.cpp/issues/18895

not sure if this is a good solution, we can also maintain a llama.cpp-specific template, but I found it quite difficult to communicate to end-users


aldehir (approved):
Given the variety of templates, at a certain point you have to do some form of transformation. I agree this is the best place to do that in llama.cpp.

PR #19028, opened 2026-01-22
server: do not log certain endpoints (avoid log spam)

ngxson:
Avoid spamming the log with requests on `/health` or `/models` API

### DIFF
--- a/tools/server/server-http.cpp
+++ b/tools/server/server-http.cpp
@@ -28,14 +28,20 @@ server_http_context::server_http_context()
 server_http_context::~server_http_context() = default;
 
 static void log_server_request(const httplib::Request & req, const httplib::Response & res) {
-    // skip GH copilot requests when using default port
-    if (req.path == "/v1/health") {
+    // skip logging requests that are regularly sent, to avoid log spam
+    if (req.path == "/health"
+        || req.path == "/v1/health"
+        || req.path == "/models"
+        || req.path == "/v1/models"
+        || req.path == "/props"
+        || req.path == "/metrics"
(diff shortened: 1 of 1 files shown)


CISC (approved):
Would it make sense to at least add a `SRV_DBG` so you can see the requests being made when debug is enabled?

ngxson:
In my experience, I haven't found any occasions where I find these lines (and the request logging in general) helpful. Mostly because this function only log a request when it's already finished.

Probably I will remove this or refactor into something more useful (i.e. log before the request is processed)

aviallon:
@ngxson How about the `/metrics` endpoint? I don't find it useful to log either.

tools/server/server-http.cpp
```diff
@@ -28,14 +28,20 @@ server_http_context::server_http_context()
...
 static void log_server_request(const httplib::Request & req, const httplib::Response & res) {
-    // skip GH copilot requests when using default port
-    if (req.path == "/v1/health") {
+    // skip logging requests that are regularly sent, to avoid log spam
+    if (req.path == "/health"
+        || req.path == "/v1/health"
+        || req.path == "/models"
+        || req.path == "/v1/models"
+        || req.path == "/props"
+        || req.path == "/metrics"
```
ngxson, line 37: @aviallon should be skipped too, but I think we will get rid of this logic soon

aviallon: By the way, I saw you work a lot on the HTTP code recently.
By chance, did you fix the `/health` requests starvation when the server is overloaded?

PR #19029, opened 2026-01-22
Add build instruction to use Vulkan backend on macOS

chanbengz:
It might sound weird but someone might need to use Vulkan on Mac (possibly in the absence of Linux machine while requiring Vulkan development). So I tried to fill in this missing part of documentation.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -487,6 +487,37 @@ Finally, after finishing your build, you should be able to do something like thi
 # ggml_vulkan: Using Intel(R) Graphics (ADL GT2) | uma: 1 | fp16: 1 | warp size: 32
 ```
 
+### For Mac users:
+
+Generally, follow LunarG's [Getting Started with the MacOS Vulkan SDK](https://vulkan.lunarg.com/doc/sdk/latest/mac/getting_started.html) guide for installation and setup of the Vulkan SDK. There are two options of Vulkan drivers on macOS, both of which implement translation layers to map Vulkan to Metal. They can be hot-swapped by setting the `VK_ICD_FILENAMES` environment variable to point to the respective ICD JSON file.
+
+Check the box for "KosmicKrisp" during the LunarG Vulkan SDK installation.
+
+Set environment variable for the LunarG Vulkan SDK after installation (and optionally add to your shell profile for persistence):
+```bash
+source /path/to/vulkan-sdk/setup-env.sh
+```
+
+#### Using MoltenVK
+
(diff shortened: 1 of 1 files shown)


jeffbolznv (approved):

docs/build.md
```diff
@@ -487,6 +487,40 @@ Finally, after finishing your build, you should be able to do something like thi
...
+```bash
+export VK_ICD_FILENAMES=$VULKAN_SDK/share/vulkan/icd.d/libkosmickrisp_icd.json
+```
+
+#### Build
+
+This is the only step different from [above](#common-steps) instructions.
+```bash
+cmake -B build -DGGML_VULKAN=1 -DGGML_METAL=OFF
+cmake --build build --config Release
```
0cc4m, line 518: Isn't `--config Release` in this step a Windows-thing? I don't remember doing that on Mac, I think I did the same workflow as on Linux, which would be 
```
cmake -B build -DGGML_VULKAN=1 -DGGML_METAL=0 -DCMAKE_BUILD_TYPE=Release
cmake --build build -j
```
I don't have a Mac anymore to check, I was just wondering.

chanbengz: Not really I guess. It works on my Mac. I think `--config Release` is following CPU and Metal build (Metal build is default on Mac so both commands are the same).

docs/build.md
```diff
@@ -487,6 +487,40 @@ Finally, after finishing your build, you should be able to do something like thi
...
+Check the box for "KosmicKrisp" during the LunarG Vulkan SDK installation.
+
+Set environment variable for the LunarG Vulkan SDK after installation:
+```bash
+export VULKAN_SDK=/path/to/Vulkan/macOS
+export VK_ICD_FILENAMES=$VULKAN_SDK/share/vulkan/icd.d/MoltenVK_icd.json
+export VK_LAYER_PATH=$VULKAN_SDK/share/vulkan/explicit_layer.d
+export PATH=$VULKAN_SDK/bin:$PATH
+export DYLD_LIBRARY_PATH=$VULKAN_SDK/lib:$DYLD_LIBRARY_PATH
+```
```
0cc4m, line 499: Overall this is fine if it works, but I think you can skip some or all of these steps if you source the script provided by the SDK. Can you check if that is true? Once that is done I'll merge.

chanbengz: Yeah it works perfectly. And thanks for pointing out because I found there're missing env variables.

0cc4m (approved):

0cc4m:
Thank you!

PR #19031, opened 2026-01-22
cli : load parser definition

aldehir:
fix #19018 

@ngxson I would like to refactor the common interface eventually to avoid serializing the definition. For now, we have to deserialize the PEG parser definition because the server passes params through a `nholmann::json` object.


pwilkin (approved):
Go for it :)

ngxson (approved):

PR #19045, opened 2026-01-23
completion : fix prompt cache for recurrent models

ggerganov:
fix #19041

Recurrent memory does not support `llama_memory_seq_rm()`. Perform it only if necessary.


danbev (approved):
Verified locally that this fixes https://github.com/ggml-org/llama.cpp/issues/19041.

PR #19048, opened 2026-01-23
common : use two decimal places for float arg help messages

danbev:
This commit updates the help messages for various command-line arguments in arg.cpp to display floating-point default values with two decimal places instead of one.

The motivation for this changes is that currently only having one decimal place means that values generated using --help or llama-gen-docs will not display the correct values.

For example, currently the value of top-p in tools/server/README.md is `0.9`, but the default value is actually '0.95'. And running llama-gen-docs does not update this value as it uses the output from the help message, which shows only one decimal place, so the values look like they are unchanged.


CISC (approved):
Update docs.

ngxson (approved):

PR #19056, opened 2026-01-23
Add workaround for templates requiring non-null content

pwilkin:
As in topic, even though OpenAI standard is null content when assistant message contains tool calls, some templates explicitly require it to be non-null or they'll fail with an error.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -2783,8 +2783,8 @@ static void system_message_not_supported(json & messages) {
 static void requires_non_null_content(json & messages) {
     GGML_ASSERT(messages.is_array());
     for (auto & message : messages) {
-        if (message.contains("tool_calls") && !messages.contains("content")) {
-            messages["content"] = "";
+        if (message.contains("tool_calls") && !message.contains("content")) {
+            message["content"] = "";
         }
     }
 }
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -2894,7 +2894,10 @@ static common_chat_params common_chat_templates_apply_jinja(
         workaround::system_message_not_supported(params.messages);
     }
 
-    if (!tmpl.original_caps().requires_non_null_content) {
+    if (tmpl.original_caps().supports_tool_calls) {
+        // some templates will require the content field in tool call messages
+        // to still be non-null, this puts an empty string everywhere where the 
+        // content field is null
         workaround::requires_non_null_content(params.messages);
     }
 
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -2780,6 +2780,15 @@ static void system_message_not_supported(json & messages) {
     }
 }
 
+static void requires_non_null_content(json & messages) {
+    GGML_ASSERT(messages.is_array());
+    for (auto & message : messages) {
+        if (message.contains("tool_calls") && !messages.contains("content")) {
+            messages["content"] = "";
+        }
+    }
+}
+
 static void func_args_not_string(json & messages) {
     GGML_ASSERT(messages.is_array());
     for (auto & message : messages) {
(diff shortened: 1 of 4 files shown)


aldehir:
This was a problem with a few templates, #18485 defaults to the empty string when null. I'm guessing the problem is during capability evaluation? If so, we should just pass in `"content": ""` to every message, instead of a whole workaround. Seems unnecessary, IMO.

pwilkin:
> This was a problem with a few templates, #18485 defaults to the empty string when null. I'm guessing the problem is during capability evaluation? If so, we should just pass in `"content": ""` to every message, instead of a whole workaround. Seems unnecessary, IMO.

If I remember correctly, other templates on the other hand DO expect content to be null on tool_call messages because that's the OpenAI spec, so I would settle for a workaround.

pwilkin:
Although if I'm wrong, then yeah, scrapping the capability check and just keeping the workaround would be fine IMO.

aldehir:
It's already happening, and no one has complained so far 😊

pwilkin:
Aight, will keep just the workaround.

tests/test-jinja.cpp
```diff
@@ -949,6 +949,19 @@ static void test_tests(testing & t) {
...
+        "no"
+    );
+
+    test_template(t, "null is undefined",
+        "{% if null is not defined %}yes{% else %}no{% endif %}",
+        json::object(),
+        "yes"
+    );
+
 }
```
CISC, line 965:
```suggestion
}
```
What are you trying to test here, we already have this test (`null` is a JSON thing), you can change `y` to `null` if you want:
https://github.com/ggml-org/llama.cpp/blob/20db3d4f2e3cfb48797e0e3ef094cefc30dafbce/tests/test-jinja.cpp#L189-L193

pwilkin: Yeah, I wanted to do the explicit "null constant is alias for undefined" check.

common/jinja/caps.cpp
```diff
@@ -160,7 +160,7 @@ caps caps_get(jinja::program & prog) {
                     {"content", "Assistant message"},
                     {"tool_calls", json::array({
                         {
-                            {"id", "call1"},
+                            {"id", "call00001"},
```
ngxson, line 163: probably useful to leave a comment header explaining why the ID must be at a certain length (IIRC there was one or two templates enforces this?)

just in case we may come up with a better way in the future, we can come back and test the problematic template

common/chat.cpp
```diff
@@ -2885,6 +2894,13 @@ static common_chat_params common_chat_templates_apply_jinja(
         workaround::system_message_not_supported(params.messages);
     }
 
+    if (tmpl.original_caps().supports_tool_calls) {
+        // some templates will require the content field in tool call messages
+        // to still be non-null, this puts an empty string everywhere where the 
+        // content field is null
+        workaround::requires_non_null_content(params.messages);
+    }
```
ngxson, line 2902: I'm not 100% sure if this is expected by all templates. in theory, there can be regression in these cases:
- template check: `if content is not none: (do something) else (do other things)`
- template that wraps content inside special tokens, example: `<message><tool>...</tool><content></content></message>` --> in such case, content maybe expected to be none

but that's just in theory, for now I can't think of a way to verify it.

common/jinja/caps.cpp
pwilkin: Right, I'll add.

common/chat.cpp
pwilkin: I thought the same, but @aldehir talked me out of it 😀 can always add the cap code if we change our mind.

aldehir: Trying to recall, but I believe this is where simply checking "non-null" is insufficient:
https://github.com/ggml-org/llama.cpp/blob/bb02f74c612064947e51d23269a1cf810b67c9a7/models/templates/Qwen-Qwen3-0.6B.jinja#L40-L42
This branch is only explored with a tool call + reasoning content. The Qwen3 template was not triggering the `requires non-null` check in Minja, so we were passing `null` content and it was throwing an exception. I think most templates are defensive against empty content, more than null content. A default of an empty string seems more reasonable to me. We are already passing in an empty string as of #18485. The capability checks do not align with this.

PR #19059, opened 2026-01-24
WebUI: Add localization using i18n

ronaldmannak:
As discussed in https://github.com/ggml-org/llama.cpp/discussions/19032 this PR uses i18n to translate the WebUI in 39 new languages, including Chinese, Spanish, and French. Some translations will probably not be perfect, but those can fixed in other PRs (or in this PR, happy to update it).

AI Disclosure
I used AI for the following things:
- Codex replaced all strings in the WebUI that needed to be translated with keys
- Codex created en.json with all strings in WebUI that needed to be translated
- Codex for setting up i18n
- I used GPT 5.2 to translate en.json in 39 languages

I understand that is not quite conform the guidelines, but this PR wouldn't be possible without using AI.


allozaur:
Hey! Thanks for the effort, but this PR would need rebasing after we merge few other things that change the architecture and structure. Additionally I had a bit different idea for how we'd manage i18n.

So let's come back to this in a bit.

ronaldmannak:
@allozaur sounds good. Should I close this PR for now?

allozaur:
> @allozaur sounds good. Should I close this PR for now?

Yep, let's revisit i18n in near future.

PR #18866, opened 2026-01-15
convert_hf_to_gguf.py: refactor modify_tensors to call super

am17an:
Prepare for #18740 and possibly other weight merges. This PR is just a refactor, and shouldn't change any functionality

PR #19064, opened 2026-01-24
Fix conversion for models inherit from LlamaModel calling super().modify_tensors()

bartowski1182:
After #18866 modify_tensors was altered to call the super function

Previously, Afmoe and CogVLM both returned the tensors directly, bypassing the `super().modify_tensors()` function

LlamaModel causes `undo_permute` to be set to `True`, which breaks the model by applying the Q/K permutation when the super function is called, as well as other potential side effects

This change replaces super() with ModelBase in order to bypass inheritance

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -10719,7 +10719,7 @@ class CogVLMVisionModel(MmprojModel):
         if not name.startswith("model.vision."):
             return
 
-        yield from ModelBase.modify_tensors(self, data_torch, name, bid)
+        yield from super().modify_tensors(data_torch, name, bid)
 
 
 @ModelBase.register("CogVLMForCausalLM")
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -2688,6 +2688,7 @@ class ArceeModel(LlamaModel):
 @ModelBase.register("AfmoeForCausalLM")
 class AfmoeModel(LlamaModel):
     model_arch = gguf.MODEL_ARCH.AFMOE
+    undo_permute = False
 
     def set_gguf_parameters(self):
         super().set_gguf_parameters()
@@ -10725,6 +10726,7 @@ class CogVLMVisionModel(MmprojModel):
 @ModelBase.register("CogVLMForCausalLM")
 class CogVLMModel(LlamaModel):
     model_arch = gguf.MODEL_ARCH.COGVLM
+    undo_permute = False
 
     def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
         # block vision tensors
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -2688,7 +2688,6 @@ class ArceeModel(LlamaModel):
 @ModelBase.register("AfmoeForCausalLM")
 class AfmoeModel(LlamaModel):
     model_arch = gguf.MODEL_ARCH.AFMOE
-    undo_permute = False
 
     def set_gguf_parameters(self):
         super().set_gguf_parameters()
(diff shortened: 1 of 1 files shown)


bartowski1182:
Get another issue with CogVLM presumably also because of the change

KeyError: "could not find any of: ['n_kv_heads', 'num_key_value_heads']"

Can't figure out what it should be..

am17an:
I think the problem is with models with multiple inheritance (e.g. CogVLM inherits from LlamaModel). So I think instead of `yield from super()`, it should be `yield from ModelBase.modify_tensors()`

bartowski1182:
@am17an Currently there are other models that inherent from `LlamaModel` (Mistral, SmolLM3, Apertus)

Mistral currently includes the same `undo_permute = False`

Are you suggesting no models should inherit from `LlamaModel` or that Afmoe specifically should not have?

Just want to be clear on what's the "best" solution at this time

am17an:
No I mean currently we're calling the immediate super class's `modify_tensors`, I was suggesting that we call the ModelBase's `modify_tensors` to bypass all inheritance

bartowski1182:
Oh I understand what you're saying, that was clear in your original comment too I just blanked on understanding it (I'll blame it on being up at 1am)

Yes that seems reasonable! Avoid any side effects from Llama for models that weren't originally calling it

am17an:
@CISC what do you think?

bartowski1182:
For the record I tested your proposed fix and it seems to work, PPL is normal again

CISC:
> I think the problem is with models with multiple inheritance (e.g. CogVLM inherits from LlamaModel). So I think instead of `yield from super()`, it should be `yield from ModelBase.modify_tensors()`

Yes, if the model always returned directly before it should yield from base.

am17an:
So, `s/yield from super()/yield from ModelBase/g`?

CISC:
> So, `s/yield from super()/yield from ModelBase/g`?

Yes, in the following models:
- AfmoeModel
- CogVLMModel
- GraniteHybridModel
- NemotronHModel (the final super() is ok)

bartowski1182:
@am17an do you want to make the change in a new PR or should I update this one

am17an:
@bartowski1182 you can update this PR. Thank you for fixing!

bartowski1182:
@am17an updated

convert_hf_to_gguf.py
```diff
@@ -9101,12 +9101,12 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
                         merged_name = f"model.layers.{bid}.mlp.experts.{w_name}.weight"
 
-                        yield from super().modify_tensors(data_torch, merged_name, bid)
+                        yield from ModelBase.modify_tensors(self, data_torch, merged_name, bid)
                     return
                 else:
                     return
 
-        yield from super().modify_tensors(data_torch, name, bid)
+        yield from ModelBase.modify_tensors(self, data_torch, name, bid)
```
```suggestion
        yield from super().modify_tensors(data_torch, name, bid)
```
This one was correct:
https://github.com/ggml-org/llama.cpp/blob/77078e80e5a9c00fc06539f27c0ae23259511ff4/convert_hf_to_gguf.py#L9251

convert_hf_to_gguf.py
```diff
@@ -10719,7 +10719,7 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         if not name.startswith("model.vision."):
             return
 
-        yield from super().modify_tensors(data_torch, name, bid)
+        yield from ModelBase.modify_tensors(self, data_torch, name, bid)
```
```suggestion
        yield from super().modify_tensors(data_torch, name, bid)
```
This one is also ok as it inherits from `MmprojModel`.

convert_hf_to_gguf.py
```diff
@@ -10756,7 +10756,7 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         elif name.startswith('language_model.'):
             name = name.replace('language_model.', '')
 
-        yield from super().modify_tensors(data_torch, name, bid)
+        yield from ModelBase.modify_tensors(self, data_torch, name, bid)
```
```suggestion
        yield from super().modify_tensors(data_torch, name, bid)
```
This one should be ok:
https://github.com/ggml-org/llama.cpp/blob/77078e80e5a9c00fc06539f27c0ae23259511ff4/convert_hf_to_gguf.py#L10946

CISC (approved):

bartowski1182:
Merged your fixes in the laziest way possible, sorry about that lol

CISC:
> Merged your fixes in the laziest way possible, sorry about that lol

No worries. :)

PR #19067, opened 2026-01-24
kv-cache : support V-less cache

ggerganov:
cont #18986 

Support V-less KV cache. This is useful for MLA models such as DeepSeek and GLM 4.7 Flash where we store combined latent data represented by the K cache. Results in almost x2 less memory for the KV cache.

Also:

- Add `llama_hparams::is_mla()`
- Add `llama_hparams::n_embd_head_k_mla()`
- Add `llama_hparams::n_embd_head_v_mla()`
- Rename `llama_hparams::get_n_embd_out()` -> `llama_hparams::n_embd_out()`
- Add `class llm_graph_input_attn_k` - similar to `class llm_graph_input_attn_kv`, but only K data

### DIFF
--- a/src/llama-graph.cpp
+++ b/src/llama-graph.cpp
@@ -407,6 +407,27 @@ bool llm_graph_input_attn_kv::can_reuse(const llm_graph_params & params) {
     return res;
 }
 
+void llm_graph_input_attn_k::set_input(const llama_ubatch * ubatch) {
+    mctx->set_input_k_idxs(self_k_idxs, ubatch);
+
+    mctx->set_input_kq_mask(self_kq_mask, ubatch, cparams.causal_attn);
+}
+
+bool llm_graph_input_attn_k::can_reuse(const llm_graph_params & params) {
+    const auto * mctx = static_cast<const llama_kv_cache_context *>(params.mctx);
+
+    this->mctx = mctx;
+
+    bool res = true;
+
+    res &= self_k_idxs->ne[0] == params.ubatch.n_tokens;
+
+    res &= self_kq_mask->ne[0] == mctx->get_n_kv();
+    res &= self_kq_mask->ne[1] == params.ubatch.n_tokens;
+
+    return res;
+}
+
 void llm_graph_input_attn_kv_iswa::set_input(const llama_ubatch * ubatch) {
     mctx->get_base()->set_input_k_idxs(self_k_idxs, ubatch);
     mctx->get_base()->set_input_v_idxs(self_v_idxs, ubatch);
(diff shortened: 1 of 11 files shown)


John-Dekka:
The commit computes `is_mla` from the model hparams `(hparams.n_embd_head_k_mla != 0 && hparams.n_embd_head_v_mla != 0)` and when true, the kv-cache will skip allocating V tensors and the graph uses the new k-only input path. That decision is driven by the model/hparams at load time.

GLM-4.7-Flash config.json does not include the `n_embd_head_k_mla / n_embd_head_v_mla` fields.

Was hoping to squeeze my GLM model a bit more. :crying_cat_face: 


---

Nevermind.

deepseek2.attention.key_length_mla
deepseek2.attention.value_length_mla

ggerganov:
> GLM-4.7-Flash config.json does not include the n_embd_head_k_mla / n_embd_head_v_mla fields.

@John-Dekka These are `llama.cpp`-specific parameters, they don't have to be present in the `config.json`. This patch applies to GLM-4.7-Flash.

JohannesGaessler (approved):
The CUDA changes are correct, the changes in the llama.cpp user code seem correct to me though I am not as familiar with that part of the codebase.

eapache:
Will the `—fit` algorithm pick up the changes in memory requirements here, or does it need to be adjusted as well to expect the smaller KV cache for these models?

JohannesGaessler:
`llama_kv_cache` has a member `std::vector<std::pair<ggml_context_ptr, ggml_backend_buffer_ptr>> ctxs_bufs;`. As long as the KV cache is allocated in `ctxs_bufs` it should work correctly with `-fit` regardless of the specific tensors.

jacekpoplawski:
very cool

master: 
```
llama_params_fit_impl: projected to use 57339 MiB of device memory vs. 71537 MiB of free device memory
llama_kv_cache: size = 19525.56 MiB (200192 cells,  47 layers,  1/1 seqs), K (f16): 10337.06 MiB, V (f16): 9188.50 MiB
```
PR:
```
llama_params_fit_impl: projected to use 48150 MiB of device memory vs. 71537 MiB of free device memory
llama_kv_cache: size = 10337.06 MiB (200192 cells,  47 layers,  1/1 seqs), K (f16): 10337.06 MiB, V (f16):    0.00 MiB
```

src/llama-graph.cpp
```diff
@@ -1868,6 +1886,93 @@ ggml_tensor * llm_graph_context::build_attn(
...
+    ggml_tensor * cur = build_attn_mha(q, k, v, kq_b, kq_mask, sinks, v_mla, kq_scale, il);
+    cb(cur, "kqv_out", il);
+
+    if (wo) {
+        cur = build_lora_mm(wo, cur);
+        if (arch == LLM_ARCH_GLM4 || arch == LLM_ARCH_GLM4_MOE) {
+            // GLM4 and GLM4_MOE seem to have numerical issues with half-precision accumulators
+            ggml_mul_mat_set_prec(cur, GGML_PREC_F32);
+        }
+    }
```
ggerganov, line 1967: We might need to add `LLM_ARCH_DEEPSEEK2` here in case we suspect similar numerical issues with GLM 4.7 Flash - something too keep in mind. cc @jeffbolznv @JohannesGaessler

ddh0: IIRC @ngxson tried this during our PR and it made no difference in his testing

ggerganov: At that time, the wrong gating function was used, so can't conclude based on this. Plus this is somewhat backend-specific - e.g. it's not a problem for Metal since we always accumulate in F32.

PR #19070, opened 2026-01-24
llama-fit-params: keep explicit --ctx-size 0

### LINKED ISSUE #18376: Misc. bug: Maximum Context Size

JohannesGaessler:
Fixes https://github.com/ggml-org/llama.cpp/issues/18376 , alternative to https://github.com/ggml-org/llama.cpp/pull/18567 . If the user explicitly sets `--ctx-size 0`, set the minimum context size to `UINT32_MAX` to prevent `llama_params_fit` from reducing it.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -1231,6 +1231,10 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         string_format("size of the prompt context (default: %d, 0 = loaded from model)", params.n_ctx),
         [](common_params & params, int value) {
             params.n_ctx = value;
+            if (value == 0) {
+                // disable context reduction in llama_params_fit if the user explicitly requests the full context size:
+                params.fit_params_min_ctx = UINT32_MAX;
+            }
         }
     ).set_env("LLAMA_ARG_CTX_SIZE"));
     add_opt(common_arg(
--- a/include/llama.h
+++ b/include/llama.h
@@ -489,6 +489,7 @@ extern "C" {
     //   - returns true if the parameters could be successfully modified to fit device memory
     //   - this function is NOT thread safe because it modifies the global llama logger state
     //   - only parameters that have the same value as in llama_default_model_params are modified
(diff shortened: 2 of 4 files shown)


ggerganov (approved):
@JohannesGaessler On a related note: for LlamaBarn we need a mechanism to query how much memory would be needed for a context of size `-c N`. Would you accept if we extend `llama-fit-params` to provide such information and if yes, can you recommend a way to implement it in terms of `llama-fit-params` UX and `libllama` API modifications (if needed)?

JohannesGaessler:
As of right now `llama_params_fit` internally sets `llama_model_params::no_alloc = True`, creates `llama_model` and `llama_context` instances, and then calls `llama_context::memory_breakdown`. I would suggest that rather than touch `llama_params_fit` an external project should simply retrieve this information itself. The problem is that currently the return format of `llama_context::memory_breakdown` is `std::map<ggml_backend_buffer_type_t, llama_memory_breakdown_data>` which is not C-compatible. So that functionality needs to either be exposed in a C-compatible way or it needs to be exposed in `llama-cpp.h`

PR #19077, opened 2026-01-24
Added qwen3-omni (only text)

rodial:
Tested locally
Model https://huggingface.co/rodial/Qwen3-Omni-30B-A3B-Thinking-GGUF


convert_hf_to_gguf.py
```diff
@@ -4516,6 +4516,27 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         yield from super().modify_tensors(data_torch, name, bid)
 
 
+@ModelBase.register("Qwen3OmniMoeForConditionalGeneration")
+class Qwen3OmniMoeModel(Qwen2MoeModel):
+    model_arch = gguf.MODEL_ARCH.QWEN3OMNIMOE
```
CISC, line 4521: I think you can just register this to `Qwen3VLMoeTextModel` and add the extra tensor checks there.

convert_hf_to_gguf.py
```diff
@@ -4516,6 +4516,27 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+
+    def set_vocab(self):
+        super().set_vocab()
+        self.gguf_writer.add_eos_token_id(151645)  # from Qwen3-VL, because null in config.json
+        self.gguf_writer.add_bos_token_id(151643)  # from Qwen3-VL, because null in config.json
```
CISC: BOS is not needed, and probably not EOS either (`<|im_end|>` is automatically added as EOT).

rodial: BOS may not be needed, but if EOS is not specified, the model aborts the response.

rodial: removed set BOS

convert_hf_to_gguf.py
CISC: You misunderstood, I didn't mean you should change inheritance, I meant you don't need a new arch.

ngxson:
I don't think this PR is useful as-is, as said, no need for new arch. The text model is just a normal Qwen3VL.

Closing this PR because it's duplicated with another one. Make sure to do your research before pushing a PR.

PR #19095, opened 2026-01-25
tests : add GQA=20 FA test

ggerganov:
Might be a good idea to have a test that exercises GQA=20 in order to catch any potential regressions.


JohannesGaessler (approved):

JohannesGaessler:
The test failure here seems to be due to attention sinks in a configuration that does not appear in actual models. But there also seems to be a more serious issue.

ggerganov:
Let me know if you prefer to adjust the tests in some way in this PR, or feel free to push directly here.

JohannesGaessler:
In terms of correctness the new functionality should be covered with the test case added in https://github.com/ggml-org/llama.cpp/pull/19115 . The kernel selection logic specific for GLM 4.7 Flash should only affect performance.

ggerganov:
> The test failure here seems to be due to attention sinks in a configuration that does not appear in actual models.

@JohannesGaessler How would you like to handle the sinks for this shape? Probably it's fine to add a check in `support_op` and not accept sinks until an actual model appears.

PR #19098, opened 2026-01-26
CUDA: fix FA rowsum + fp16 softmax offset

rmatif:
While testing Flux2 klein in sdcpp, I hit two separate failures that show up as NaNs or wildly off outputs

- Fully masked rows can end up with rowsum == 0. The MMA/WMMA paths (and the stream‑k fixup path) divide by rowsum and propagate NaNs. I simply guarded the divide and write zeros for those rows

- In the np == 1 MMA path, fp16 accumulation can overflow when softmax weights are summed over the full batch. The existing offset wasn’t enough once nbatch_fa grows, so the exp weights saturate. I add an additional computed offset of ln(nbatch_fa) for np == 1, which keeps the summed weights <= 1 in fp16

Both issues are reproducible in sdcpp

```shell
./build/bin/sd-cli \
  --diffusion-model models/flux-2-klein-base-4b-Q8_0.gguf \
  --vae models/flux2-vae.safetensors \
  --llm models/Qwen3-4B-Q8_0.gguf \
  -p "Replace the duck with a cat" \
  --sampling-method euler \
  --steps 20 \
  --cfg-scale 2.5 \
  --diffusion-fa \
  -W 1216 \
  -H 832 \
  -s 42 \
  -v \
  --vae-tiling \
  -r duck.png
```

### DIFF
--- a/ggml/src/ggml-cuda/fattn-common.cuh
+++ b/ggml/src/ggml-cuda/fattn-common.cuh
@@ -716,7 +716,11 @@ static __global__ void flash_attn_stream_k_fixup(
     }
 
     // Write back final result:
-    *dst = dst_val / rowsum;
+    if (!(rowsum > 0.0f)) {
+        *dst = 0.0f;
+    } else {
+        *dst = dst_val / rowsum;
+    }
 }
 
 template<int D> // D == head size
--- a/ggml/src/ggml-cuda/fattn-mma-f16.cuh
+++ b/ggml/src/ggml-cuda/fattn-mma-f16.cuh
@@ -561,6 +561,16 @@ static __device__ __forceinline__ void flash_attn_ext_f16_iter(
         KQ_max_new[col] = KQ_max[col];
     }
     float KQ_rowsum_add[cols_per_thread] = {0.0f};
+    constexpr int log2_nbatch_fa =
+        nbatch_fa == 256 ? 8 :
+        nbatch_fa == 128 ? 7 :
+        nbatch_fa ==  64 ? 6 :
+        nbatch_fa ==  32 ? 5 :
+        nbatch_fa ==  16 ? 4 :
+        nbatch_fa ==   8 ? 3 : 0;
+    static_assert(log2_nbatch_fa != 0, "unexpected nbatch_fa");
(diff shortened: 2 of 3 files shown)


ggml/src/ggml-cuda/fattn-common.cuh
```diff
@@ -716,7 +716,11 @@ static __global__ void flash_attn_stream_k_fixup(
     }
 
     // Write back final result:
-    *dst = dst_val / rowsum;
+    if (!(rowsum > 0.0f)) {
+        *dst = 0.0f;
+    } else {
+        *dst = dst_val / rowsum;
+    }
```
JohannesGaessler, line 723:
```suggestion
    *dst = dst_val == 0.0f && rowsum == 0.0f ? 0.0f : dst_val / rowsum;
```

This should also cover the edge case of a fully masked-out row. I think a more narrow condition like this would be preferable to avoid accidentally masking bugs. Also you missed some spots where this fix is needed: the vector kernel, the tile kernel, and the function for combining the results of the non-MMA kernels.

ggml/src/ggml-cuda/fattn-mma-f16.cuh
```diff
@@ -561,6 +561,16 @@ static __device__ __forceinline__ void flash_attn_ext_f16_iter(
...
     }
     float KQ_rowsum_add[cols_per_thread] = {0.0f};
+    constexpr int log2_nbatch_fa =
+        nbatch_fa == 256 ? 8 :
+        nbatch_fa == 128 ? 7 :
+        nbatch_fa ==  64 ? 6 :
+        nbatch_fa ==  32 ? 5 :
+        nbatch_fa ==  16 ? 4 :
+        nbatch_fa ==   8 ? 3 : 0;
+    static_assert(log2_nbatch_fa != 0, "unexpected nbatch_fa");
```
JohannesGaessler, line 571: Sorry, but this patch makes no sense to me. `nbatch_fa` should not affect how much of an offset is needed to avoid overflows in the VKQ accumulators so I think that you are misdiagnosing the problem.

PR #19101, opened 2026-01-26
Check if ctx or model is null before calling sampler

gagankonana:
Add a check in main in completion.cpp to check if ctx or model is null before calling sampler. This validates that common_init_from_params was successful 
Resolves #19065

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -1195,9 +1195,6 @@ llama_context * common_init_result::context() {
 }
 
 common_sampler * common_init_result::sampler(llama_seq_id seq_id) {
-    if (pimpl->samplers.size() <= (size_t) seq_id) {
-        return nullptr;
-    }
     return pimpl->samplers[seq_id].get();
 }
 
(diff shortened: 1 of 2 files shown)


danbev (approved):

ggerganov:
This seems to hide some other problem - it's not a valid solution

danbev:
> This seems to hide some other problem - it's not a valid solution

Ah yes, it looks like this is because the model is failing to load but we are calling llama_init->sampler(0) without checking the context here: https://github.com/ggml-org/llama.cpp/blob/56f3ebf38ea5757447e7ea167cda15e1eebf9601/tools/completion/completion.cpp#L144-L146

We should move [check](https://github.com/ggml-org/llama.cpp/blob/56f3ebf38ea5757447e7ea167cda15e1eebf9601/tools/completion/completion.cpp#L148) of the ctx  before this call.

gagankonana:
Updated the PR. I followed the pattern from other files. Let me know wdyt

danbev:
> Updated the PR. I followed the pattern from other files. Let me know wdyt

Thanks! This looks good to me. Can you also take a look at the [EditorConfig Checker](https://github.com/ggml-org/llama.cpp/actions/runs/21424693358/job/61694051276?pr=19101) CI failure?

PR #19107, opened 2026-01-26
ci : use new 1vCPU runner for lightweight jobs

CISC:
See https://github.blog/changelog/2026-01-22-1-vcpu-linux-runner-now-generally-available-in-github-actions/

### DIFF
--- a/.github/workflows/check-vendor.yml
+++ b/.github/workflows/check-vendor.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   check-vendor:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
 
     steps:
       - name: Checkout
--- a/.github/workflows/close-issue.yml
+++ b/.github/workflows/close-issue.yml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   close-issues:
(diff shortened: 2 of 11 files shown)


ggerganov (approved):

CISC:
Whoa, guess running pyright is out of the question.

CISC:
> Whoa, guess running pyright is out of the question.

..makes a mental note to check out [ty](https://docs.astral.sh/ty/) again some day...

ggerganov:
Btw, I am wondering if I should add 1 or 2 extra `ggml-ci` runners for CUDA/Vulkan, or if we should try to optimize the current `ggml-ci` runtime. It's a bit slow lately.

CISC:
> Btw, I am wondering if I should add 1 or 2 extra `ggml-ci` runners for CUDA/Vulkan, or if we should try to optimize the current `ggml-ci` runtime. It's a bit slow lately.

Yeah, it doesn't take much activity for the runner backlog to get swamped, if we can get more runners that would be great (also optimizing runtime of course).

CISC:
Hmmm, odd failure:
https://github.com/ggml-org/llama.cpp/actions/runs/21385925341/job/61562252524

PR #19109, opened 2026-01-26
llama : disable Direct IO by default

ggerganov:
ref https://github.com/ggml-org/llama.cpp/issues/19035#issuecomment-3798971944
cont #18012 

- Update `llama_model_params::use_direct_io == false` by default
- Update `common_params::use_direct_io == false` by default

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2198,18 +2198,15 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
     add_opt(common_arg(
         {"--mmap"},
         {"--no-mmap"},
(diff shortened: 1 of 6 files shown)


CISC:
Perhaps makes sense to just disable mmap when `--direct-io` is used (and available) as mmap is on by default, a bit silly to have to also use `--no-mmap`?

ggerganov:
@CISC I've updated the logic as recommended.

The CUDA/Vulkan CI runs are a bit faster now:

Screenshot 2026-01-27 at 6 27 13 PM

And on master:

Screenshot 2026-01-27 at 6 27 58 PM

CISC (approved):

taronaeo (approved):

PR #19113, opened 2026-01-26
common : add specific warning for multiple -m options

danbev:
This commit updates the deprecation warning for multiple model options with a specific message that does not mention comma-separated values as an option.

The motivation for this is that currently it can be confusing if multiple model options are specified with llama-server as this would display the following warning:
```console
DEPRECATED: argument '--model' specified multiple times, use comma-separated values instead (only last value will be used)
```
But llama-server does not support a comma-separated models option.

With this change the user is informed that multiple models can be supported in router mode if using llama-server:
```console
DEPRECATED: multiple --model options are not supported (only last value will be used). Use router mode for llama-server (see --help)
```

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -430,6 +430,26 @@ static bool parse_bool_value(const std::string & value) {
 // CLI argument parsing functions
 //
 
+static void check_duplicate_arg(const std::string & arg) {
+    static const std::set<std::string> multi_value_options = {
+        "--lora",
+        "--lora-scaled",
+        "--control-vector",
+        "--control-vector-scaled",
+        "-f", "--file",
+        "--context-file",
+        "--api-key",
+        "--logit-bias",
+        "--tensor-filter"
+    };
+
+    if (multi_value_options.find(arg) != multi_value_options.end()) {
+        LOG_INF("DEPRECATED: argument '%s' specified multiple times, use comma-separated values instead (only last value will be used)\n", arg.c_str());
+    } else {
+        throw std::invalid_argument(string_format("error: argument '%s' cannot be specified multiple times", arg.c_str()));
+    }
+}
+
 static bool common_params_parse_ex(int argc, char ** argv, common_params_context & ctx_arg) {
     common_params & params = ctx_arg.params;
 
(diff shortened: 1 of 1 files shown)


ngxson:
Unless I missed something, we never support multiple -m, so using the word "deprecated" here make one thinks that it used to be, but now no longer supported

danbev:
> Unless I missed something, we never support multiple -m, so using the word "deprecated" here make one thinks that it used to be, but now no longer supported

Yeah good point, and I was not sure about using "deprecated" here but thought I'd perhaps it was supported at some point that I was not aware of. Let me know and I'll change this to something else, perhaps just `Warning or WARNING`.

ngxson:
But to start, in which case people may specify multiple `-m`?

I personally believe that multiple `-m` has never been an expected behavior since the beginning of this project, so it makes more sense to just throw an error instead.

danbev:
> But to start, in which case people may specify multiple `-m`?

This was suggested to me by an AI actually (can't recall which model) which is why I looked into this as I did not think this was supported.

> so it makes more sense to just throw an error instead.

I agree, and if that is alright to do that here it would be preferred.

ngxson:
> This was suggested to me by an AI actually (can't recall which model) which is why I looked into this as I did not think this was supported.

In such case, I think there is no need to change anything at all.

If something need to changed, I expect to throw an error on any args that are not supposed to have multiple values. We have plenty of them: `-c, -n, --temp, --top-k, -ngl, ...`

danbev:
> In such case, I think there is no need to change anything at all.

I actually think this is a reason to change this as it might be suggested to users, and it is somewhat confusing as it is right now. 

But I agree that throwing an error for all the options that don't actually support being specified multiple times would be nicer. I'll add a commit with a suggestion for this:

For example, using multiple --model options would result in an error:
```console
$ llama-completion -m models/Qwen2.5-0.5B-Instruct.gguf --model models/Qwen2.5-0.5B-Instruct.gguf
error: argument '--model' cannot be specified multiple times
```
But using multiple --file options would only result in a warning (like it currently does):
```console
$ llama-completion -m models/Qwen2.5-0.5B-Instruct.gguf -f test_file.txt -f test_file.txt
DEPRECATED: argument '-f' specified multiple times, use comma-separated values instead (only last value will be used)
...
```

danbev:
Closing as this is out of date.

PR #19092, opened 2026-01-25
CUDA: faster FA for GQA > 1 but not power of 2

JohannesGaessler:
This PR generalizes the CUDA MMA FlashAttention kernel to enable the GQA optimizations for models where the ratio between the number of Q heads and the number of K/V heads is not a power of 2. This is done by simply padding the Q columns per CUDA block to the next higher power of 2. This wastes a bit of compute but particularly for small batch sizes the kernel is I/O-bound anyways.

On Ampere or newer this improves performance of GLM 4.7 Flash as well as some random models like Granite 3.0 with a GQA ratio of 3. On Volta the new code path is slower than master so it's disabled. On RDNA4 it seems to be faster but as of right now the performance of the MMA kernel is bad on RDNA for head sizes > 128 so there is no benefit for GLM 4.7 Flash.

<details>
<summary>Performance changes</summary>

| GPU        | Model                 |     Microbatch size | Test           |     t/s master |     t/s 12116ba4f |     Speedup |
| :--------- | :-------------------- | ------------------: | :------------- | -------------: | ----------------: | ----------: |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                   1 | pp512@d32768   |          52.43 |             77.70 |        1.48 |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                   2 | pp512@d32768   |          73.10 |             85.88 |        1.17 |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                   4 | pp512@d32768   |         124.81 |            132.09 |        1.06 |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                   8 | pp512@d32768   |         201.59 |            201.04 |        1.00 |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                  16 | pp512@d32768   |         312.73 |            312.58 |        1.00 |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                  32 | pp512@d32768   |         387.78 |            389.53 |        1.00 |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                  64 | pp512@d32768   |         441.64 |            437.98 |        0.99 |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                 128 | pp512@d32768   |         455.55 |            462.53 |        1.02 |
| RTX 3090   | GLM 4.7 Flash Q2_K_M  |                 256 | pp512@d32768   |         569.98 |            566.91 |        0.99 |
[...]

PR #19115, opened 2026-01-26
CUDA: fix padding of GQA to power of 2 in FA

JohannesGaessler:
Fixes https://github.com/ggml-org/llama.cpp/issues/19112 , the issue was introduced with https://github.com/ggml-org/llama.cpp/pull/19092 .

The MMA CUDA FlashAttention kernel uses a stream-k decomposition to treat the four-dimensional input tensors as one continuous dimension to split across streaming multiprocessors. However, in conjunction with the GQA-specific optimizations in the MMA kernel this is only correct if the number of Q columns per CUDA block exactly divide `n_gqa`. Otherwise the wrong Q and K/V heads will be associated and the result will be wrong (if there is only a single K/V head this doesn't matter so it was not detected in testing).

This PR extends the 4D space on master to a 5D space by splitting the "z" dimension with the number of Q heads into one dimension for the number of K/V heads and another dimension for the number of Q heads per K/V head. This then makes it possible to simply pad the Q columns per CUDA block to a power of 2.

I modified one of the test cases in `test-backend-ops` to check for this fix. On master `n_gqa` is set to 1, 4, and 16. I chose these values to check for no GQA optimizations, GQA optimizations with a single CUDA block in z direction, and GQA optimizations with >1 CUDA blocks per z direction. By changing the last value from 16 to 12 it will still cover that case while also checking for the correct padding logic.

### DIFF
--- a/ggml/src/ggml-cuda/fattn-common.cuh
+++ b/ggml/src/ggml-cuda/fattn-common.cuh
@@ -629,8 +629,8 @@ static __global__ void flash_attn_mask_to_KV_max(
 template<int D, int ncols1, int ncols2> // D == head size
 __launch_bounds__(D, 1)
 static __global__ void flash_attn_stream_k_fixup(
-        float * __restrict__ dst, const float2 * __restrict__ dst_fixup, const int ne01, const int ne02, const int ne03, const int ne11,
-        const int nbatch_fa) {
+        float * __restrict__ dst, const float2 * __restrict__ dst_fixup, const int ne01, const int ne02, const int ne03,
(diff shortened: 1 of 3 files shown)


ggerganov (approved):

RodriMora:
Confirmed working good now with https://huggingface.co/bartowski/MiniMaxAI_MiniMax-M2.1-GGUF at Q6. 
It was outputting giberish with the latest commit https://github.com/ggml-org/llama.cpp/commit/8f80d1b254aef70a0959e314be368d05debe7294

PR #19139, opened 2026-01-27
llama: Add option to merge gate and exp weights

am17an:
Continuing on #18740 and #18866, add option `--fuse_gate_up_exps` to `convert_hf_to_gguf.py`. 
I've just added the gate_up tracking for deepseek2 (GLM 4.7 flash) and gpt-oss - although for gpt-oss we need even more changes (it goes through the `generate_extra_tensors` for generating expert weights). This PR is not complete as we would need to add this check in all MoE models and their tensors, but putting it out there in any case.  

on 5090:

Master
| deepseek2 30B.A3B Q4_K - Medium |  16.88 GiB |    29.94 B | CUDA       |  99 |  1 |          pp2048 |      6269.31 ± 11.73 |
| deepseek2 30B.A3B Q4_K - Medium |  16.88 GiB |    29.94 B | CUDA       |  99 |  1 |          pp4096 |      5833.01 ± 11.12 |
| deepseek2 30B.A3B Q4_K - Medium |  16.88 GiB |    29.94 B | CUDA       |  99 |  1 |          pp8192 |      5137.33 ± 16.16 |
| deepseek2 30B.A3B Q4_K - Medium |  16.88 GiB |    29.94 B | CUDA       |  99 |  1 |           tg128 |        168.98 ± 0.68 |

this PR (with the fused GGUF)

| deepseek2 30B.A3B Q4_K - Medium |  16.88 GiB |    29.94 B | CUDA       |  99 |  1 |          pp2048 |      6846.90 ± 10.22 |
| deepseek2 30B.A3B Q4_K - Medium |  16.88 GiB |    29.94 B | CUDA       |  99 |  1 |          pp4096 |      6303.35 ± 15.73 |
| deepseek2 30B.A3B Q4_K - Medium |  16.88 GiB |    29.94 B | CUDA       |  99 |  1 |          pp8192 |       5502.23 ± 9.50 |
| deepseek2 30B.A3B Q4_K - Medium |  16.88 GiB |    29.94 B | CUDA       |  99 |  1 |           tg128 |        167.92 ± 0.75 |


CISC:
> although for gpt-oss we need even more changes (it goes through the `generate_extra_tensors` for generating expert weights). 

Ah, should probably also go through those to ensure they are all generators and yielding from base then.

convert_hf_to_gguf.py
```diff
@@ -514,8 +518,36 @@ def set_gguf_parameters(self):
...
+        if self.fuse_gate_up_exps and bid is not None:
+            if self.match_model_tensor_name(new_name, gguf.MODEL_TENSOR.FFN_GATE_EXP, bid):
+                self._gate_exp_buffer[bid] = data_torch
+                # Check if up_exps is already buffered for this layer
+                if bid in self._up_exp_buffer:
+                    gate_data = self._gate_exp_buffer.pop(bid)
+                    up_data = self._up_exp_buffer.pop(bid)
+                    # gate/up shape: (n_expert, n_ff, n_embd), concatenate to (n_expert, n_ff*2, n_embd)
+                    fused_data = torch.cat([gate_data, up_data], dim=1)
+                    fused_name = f"blk.{bid}.ffn_gate_up_exps.weight"
```
```suggestion
                    fused_name = self.format_tensor_name(gguf.MODEL_TENSOR.FFN_GATE_UP_EXP, bid)
```

convert_hf_to_gguf.py
```diff
@@ -514,8 +518,36 @@ def set_gguf_parameters(self):
...
+                return []  # Wait for up_exps
+            elif self.match_model_tensor_name(new_name, gguf.MODEL_TENSOR.FFN_UP_EXP, bid):
+                self._up_exp_buffer[bid] = data_torch
+                # Check if gate_exps is already buffered for this layer
+                if bid in self._gate_exp_buffer:
+                    gate_data = self._gate_exp_buffer.pop(bid)
+                    up_data = self._up_exp_buffer.pop(bid)
+                    # gate/up shape: (n_expert, n_ff, n_embd), concatenate to (n_expert, n_ff*2, n_embd)
+                    fused_data = torch.cat([gate_data, up_data], dim=1)
+                    fused_name = f"blk.{bid}.ffn_gate_up_exps.weight"
```
```suggestion
                    fused_name = self.format_tensor_name(gguf.MODEL_TENSOR.FFN_GATE_UP_EXP, bid)
```

gguf-py/gguf/constants.py
```diff
@@ -3115,6 +3117,7 @@ class MODEL_TENSOR(IntEnum):
         MODEL_TENSOR.ATTN_OUT,
         MODEL_TENSOR.ATTN_SINKS,
         MODEL_TENSOR.FFN_GATE_INP,
+        MODEL_TENSOR.FFN_GATE_UP_EXP,
```
CISC: You need to add it to deepseek too.

convert_hf_to_gguf.py
```diff
@@ -10023,20 +10055,23 @@ def transform_nibble_layout(self, tensor):
...
+        """Repack blocks and scales into MXFP4 format, returns tensor."""
         assert blocks.dtype == torch.uint8
         assert scales.dtype == torch.uint8
         scales = scales.unsqueeze(-1)
         assert len(blocks.shape) == 4
         assert len(scales.shape) == 4
         blocks = self.transform_nibble_layout(blocks)
         new_data = torch.concat((scales, blocks), dim=-1)
-        new_shape = [new_data.shape[0], new_data.shape[1], new_data.shape[2] * 32]
-        logger.info(f"Repacked {new_name} with shape {new_shape} and quantization MXFP4")
```
ngxson, line 10705: as long as the input tensor is still 2D, this log line should still be correct, isn't it?

convert_hf_to_gguf.py
```diff
@@ -10023,20 +10055,23 @@ def transform_nibble_layout(self, tensor):
...
         # flatten last dim
         new_data = new_data.view(new_data.shape[0], new_data.shape[1], new_data.shape[2] * new_data.shape[3])
-        new_data = new_data.numpy()
-        self.gguf_writer.add_tensor(new_name, new_data, raw_dtype=gguf.GGMLQuantizationType.MXFP4)
+        return new_data
+
+    def tensor_force_quant(self, name, new_name, bid, n_dims):
+        if any(x in new_name for x in ("ffn_gate_exps", "ffn_up_exps", "ffn_down_exps", "ffn_gate_up_exps")):
+            return gguf.GGMLQuantizationType.MXFP4
+        return super().tensor_force_quant(name, new_name, bid, n_dims)
```
ngxson: this change seems quite magic to me, I'm doubt that it will actually work.

IIRC the conversion code always dequantize the tensor to F32 then quantize back the the type specified here. that's why `repack_mxfp4` does not return a tensor, but adds it directly to GGUF.

**please test this change very carefully**, because dequant --> quant round trip can degrade the quality

am17an: Sorry I didn't mean to push this change. Got pushed in the rebase. Will remove

convert_hf_to_gguf.py
```diff
@@ -514,8 +518,36 @@ def set_gguf_parameters(self):
...
-        return [(self.map_tensor_name(name), data_torch)]
+        new_name = self.map_tensor_name(name)
+
+        # Handle gate/up expert tensor fusion if enabled
+        if self.fuse_gate_up_exps and bid is not None:
+            if self.match_model_tensor_name(new_name, gguf.MODEL_TENSOR.FFN_GATE_EXP, bid):
+                self._gate_exp_buffer[bid] = data_torch
+                # Check if up_exps is already buffered for this layer
+                if bid in self._up_exp_buffer:
+                    gate_data = self._gate_exp_buffer.pop(bid)
```
ngxson: I think this code can be simplified to avoid duplicate 2 branches:

```py
if self.match_model_tensor_name(new_name, gguf.MODEL_TENSOR.FFN_GATE_EXP, bid):
    self._gate_exp_buffer[bid] = data_torch
elif self.match_model_tensor_name(new_name, gguf.MODEL_TENSOR.FFN_UP_EXP, bid):
    self._up_exp_buffer[bid] = data_torch

if bid in self._gate_exp_buffer and bid in self._up_exp_buffer:
    # merge them
```

ngxson:
Still, I think there are some problems with the conversion script:
- For gpt-oss, the model already come with the gate_up fused. It's split and repacked inside `generate_extra_tensors`, so in theory, up/gate won't be passed to `modify_tensors`. Instead, `GptOssModel.generate_extra_tensors()` method must be aware of either `fuse_gate_up_exps` is enabled or not, and repack it accordingly without going through `modify_tensors` (which is known to do dequant --> requant roud trip)
- The cpp code currently only support gpt-oss and deepseek2, so I think the conversion script must throw an error on any other models. This is to avoid user to produce a non-working GGUF

am17an:
I tried to address 2) above. Full disclosure I used AI for this to make a list of MoE models and make all the edits. 

For gpt-oss, I don't know what is the correct way to do it, one way is what I added in #18740, I can add the same here.

CISC:
> I tried to address 2) above.

You must update all arches in ~`tensor_mapping.py`~`constants.py` too, otherwise conversion will fail (so, really, this point was already addressed :)).

am17an:
> You must update all arches in tensor_mapping.py too

I don't understand what you mean. Please assume I have no experience with conversion.

CISC:
> > You must update all arches in tensor_mapping.py too
> 
> I don't understand what you mean. Please assume I have no experience with conversion.

Sorry, I meant `constants.py`, since we changed it to use `format_tensor_name`:
https://github.com/ggml-org/llama.cpp/blob/566884c5dd90892d7c07a240093f6c5651cc7f18/convert_hf_to_gguf.py#L489-L491

ngxson:
IMO it's probably better to find a way to generalize this, rather than apply the change to each models. Also, maybe only apply this to a limited number of models first, then see if it actually works.

What I'm worry is that contributors may not aware of this change and some models will eventually forget to support this. Also, it addes a bit more boilerplate to new models' code. We should find a way to establish a new rule that new models should use merged gate/up by default

am17an:
My initial PR #18740 did this only for 1 model. The discussion was that it should be done for *all* models, for this we needed to call `ModelBase::modify_tensors` from each model's `modify_tensors`. See #18866, #19084, #19091. That was to lay the groundwork for doing this. This PR is also initially did it for 1 model (GLM 4.7 flash) but it would be weird to support a flag for one model and I can foresee people asking for support for other models.  

> Also, it addes a bit more boilerplate to new models' code.

The llama-arch.cpp file is full of boilerplate code. I don't see what this PR changes. A function can be made to abstract the if statements if that's you mean. I was planning to do that for the Q,K,V merge anyway.

With that said, I don't care if all models get support, especially since it involves a lot of code changes which aren't exactly fun or foolproof. My aim is extract maximum performance out of a select few models which are proven to have many users like gpt-oss, qwen3 and now glm 4.7 flash. So the solution I see is this PR with some arch's which have been verified to work, the rest can either fail through `constants.py` or we can hardcode the supported arch set.

ngxson:
> The llama-arch.cpp file is full of boilerplate code. I don't see what this PR changes. A function can be made to abstract the if statements if that's you mean. I was planning to do that for the Q,K,V merge anyway.

Most of them are not really boilerplate code, for example: you may see most models copy the same FFN/FFN_EXPS code, but this is because the tensor dimension must be specified and some models indeed can have different shapes for it, this is totally up to the original model (or the people who created the model) to decide. So, they are not actually boilerplate code, but they are configurations.

Definition of merged gate/up and QKV are not the same because they can be inferred from the existing code. In terms of functional programming: you can have a pure function to transform from the unmerged "definition" to merged one.

My suggestion is that we can avoid adding llama.cpp-specific code paths like what you're doing into the model "definition". Instead, it can be implemented as a transformation inside `create_tensor`, logic will be somewhat like:
- if the tensor name `tn` is gate or up, see if we already had another up or gate added
- if yes, look up in the list of tensors if we have `gate_up` merged tensor
- if no, treat is as normal

The same logic can be reused for QKV merged. We can potentially generalize it further so more merged FFN and merged QKV can reuse exactly the same code path with minor modifications.

Otherwise, if we only support the 3 models as your said:

> My aim is extract maximum performance out of a select few models which are proven to have many users like gpt-oss, qwen3 and now glm 4.7 flash.

I (more) agree with this because it makes things easier to test and to track down issues. As pointed out in my comments earlier, there were some issues regarding gpt-oss. So it makes more sense to scale down the scope of this PR, and put more efforts into testing it thoroughly

ngxson:
> My suggestion is that we can avoid adding llama.cpp-specific code paths like what you're doing into the model "definition". Instead, it can be implemented as a transformation inside `create_tensor`

Otherwise, another way to do is to abstract the loading of FFN into a new function, `create_tensor_ffn`. This can already significantly reduce the amount of code, while making it more aligned with `build_ffn`

am17an:
> Definition of merged gate/up and QKV are not the same because they can be inferred from the existing code. In terms of functional programming: you can have a pure function to transform from the unmerged "definition" to merged one.

Not sure what you mean. We can merge Q,K,V into a one big matrix and same for gate and up. I'm not able to see the difference and even your proposed implementation uses the same path for both.

> Otherwise, if we only support the 3 models as your said:

Re the `create_tensors` part, I'm not willing to take an entirely different implementation approach to the current PR. So we can limit the scope to the 2-3 models and we can move on

am17an:
And btw this pattern already exists in `llama-model.cpp`, checking for the fused tensor and fallback if not

https://github.com/ggml-org/llama.cpp/blob/362bff9773cd4fc1ea85bb78ebed29bf7d1e17c9/src/llama-model.cpp#L3297-L3309

CISC:
> And btw this pattern already exists in `llama-model.cpp`, checking for the fused tensor and fallback if not

A `create_tensor_ffn`/`create_tensor_qkv` sounds like a good idea, though since it would have to update `layers[i]` directly that would probably not be a good name for it.

Edit: Either way, leave it for a follow-up.

ngxson:
> Not sure what you mean. We can merge Q,K,V into a one big matrix and same for gate and up. I'm not able to see the difference and even your proposed implementation uses the same path for both.

What I mean was that when writing the tensor loading code, the `create_tensor` is merely to describe what are the shape of the QKV tensor and some addition flags.

The detail of whether they are merged or not should be abstracted out, because it's has nothing with the model definition itself.

In terms of functional programming, it doesn't make sense to write something like:

```
q = load_q(n_embd, n_dim_k)
k = load_k(n_embd, n_dim_k)
v = load_v(n_embd, n_dim_v)
if (!q && !k && !v)
  qkv = load_qkv(n_embd * 3, n_dim_qkv)
```

Instead, `load_qkv` can already be derived from `load_q | k | v`, duplicating the code is redundant. It can be abstracted out, something like `create_tensor_qkv`

It doesn't change the implementation under the hood, but the point is that: model definition should ONLY convey what make up the model, NOT what we added to make it faster

am17an:
Sorry I kind of abandoned this PR because it seemed like we can't reach a consensus. What is something we can all agree on that should be mergeable? I don't think this sizeable speedup which is relatively free should just go unmerged. @ggerganov maybe you can weigh in as well

ngxson:
I think the first step could be to abstract out a `create_tensor_qkv` as I suggested above, it should be a dedicated PR for visibility (and to allow `git bisect` if anything goes wrong)

And as explained earlier, GPT-OSS repacking won't work as-is, it need to be adapted somehow. I'll have a look if I have time, but for now I think we can leave out the GPT-OSS from this PR (I suspect it will currently be dequant --> quant round trip)

In short, I'm just trying to reduce the scope of this PR, as it still touching quite many models.

ggerganov:
For my understanding, the python changes are good, correct?

ngxson:
@ggerganov the code suggests that python code won't work correctly with GPT-OSS because of the custom MXFP4 repacking logic, though @am17an please correct if I'm wrong

am17an:
> I think the first step could be to abstract out a create_tensor_qkv as I suggested above

Okay I will refactor to create a function `create_tensor_up_gate_exps` which does a similar pattern. I agree that GPT-OSS can be handled separately as a special case. 

> For my understanding, the python changes are good, correct?

Yes, as @ngxson said, any custom packing logic would need a special case. Although it seems like we don't have a generic way of converting mxfp4 models, adding that might fix this more generally

ggerganov:
Ok got it. The gpt-oss needs some extra work since it utilizes the `generate_extra_tensors()` mechanism in order to create `mxfp4` tensors directly. We can address this later on, shouldn't be too difficult.

The `create_tensor_up_gate_exps` and `create_tensor_qkv` sound good to reduce repetitiveness. Thought it should be fine to do it in a follow-up PR too. Either way I think is fine.

am17an:
I made the change to reduce the number of models with this change. This reduces the amount of changes to just deepseek2 (GLM Flash) and qwen3next and qwen3.5

Before:
 Device 0: NVIDIA GeForce RTX 5090, compute capability 12.0, VMM: yes
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |          pp2048 |      2959.99 ± 11.51 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           tg128 |        133.57 ± 1.21 |

After:
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |          pp2048 |       3326.12 ± 7.60 |
| qwen3next 80B.A3B Q2_K - Medium |  27.12 GiB |    79.67 B | CUDA       |  99 |           tg128 |        132.90 ± 1.35 |

src/llama-model.cpp
```diff
@@ -7458,9 +7465,8 @@ bool llama_model::load_tensors(llama_model_loader & ml) {
                         }
 
                         layer.ffn_gate_inp  = create_tensor(tn(LLM_TENSOR_FFN_GATE_INP,  "weight", i), { n_embd, n_expert }, 0);
-                        layer.ffn_gate_exps = create_tensor(tn(LLM_TENSOR_FFN_GATE_EXPS, "weight", i), { n_embd, n_ff_exp, n_expert }, 0);
                         layer.ffn_down_exps = create_tensor(tn(LLM_TENSOR_FFN_DOWN_EXPS, "weight", i), { n_ff_exp, n_embd, n_expert }, 0);
-                        layer.ffn_up_exps   = create_tensor(tn(LLM_TENSOR_FFN_UP_EXPS,   "weight", i), { n_embd, n_ff_exp, n_expert }, 0);
+                        create_tensor_gate_up_exps(layer, i, n_embd, n_ff_exp, n_expert, 0);
```
ngxson, line 7469: could be better to have a `create_tensors_ffn`, so that will be aligned with `build_ffn`

but that can be another PR, this is good for now

src/models/deepseek2.cpp
```diff
@@ -218,7 +218,9 @@ llm_build_deepseek2::llm_build_deepseek2(const llama_model & model, const llm_gr
                 LLM_FFN_SILU, hparams.expert_weights_norm,
                 hparams.expert_weights_scale, hparams.expert_weights_scale,
                 (llama_expert_gating_func_type) hparams.expert_gating_func,
-                il);
+                il,
+                nullptr,
+                model.layers[il].ffn_gate_up_exps);
```
ngxson, line 223: probably better to regroup all `ffn` tensors into a new `model.layers[il].ffn` struct, so that `build_ffn` / `build_moe_ffn` will only take one single param for weight.

I'll do that in a dedicated PR

ngxson (approved):

am17an:
Ok I will merge this, the one issue I see is that `--fuse_gate_up_weights` is available for every model, but the GGUF will only work for the models given above which have the cpp part implemented, there is currently no easy way to verifying that the support is only limited to so and so models (apart from hardcoding it), it will be best to do the whole thing at once. We can do it once we are sure this doesn't cause any unforeseen issues.

ubergarm:
People are asking me already if I used this new `convert_hf_to_gguf.py --fuse-gate-up-exps` (there is no `--fuse_gate_up_weights`) for recent qwen3 models...

My concern is if someone generates a bf16 GGUF using `--fuse-gate-up-exps` and calculates imatrix on it, then it will be missing data if someone tries to use the same imatrix file on a bf16 GGUF that was not fused (and vice versa maybe?)...

Not sure if it is possible to do the fusions at graph compile time instead of baking them into the GGUFs?

ngxson:
> Not sure if it is possible to do the fusions at graph compile time instead of baking them into the GGUFs?

I think it will quite tricky doing that way. It will require dequantizing FFN tensors in order to merge them.

Maybe we can add a pre-check in `llama-quant` in this case? And ideally find a way to merge imatrix data to match the new tensor layout? (Idk, just speculating)

ubergarm:
> And ideally find a way to merge imatrix data to match the new tensor layout?

fortunately, [ik was accommodating and resolved the discrepancies](https://github.com/ikawrakow/ik_llama.cpp/pull/1419) introduced with this PR so we again have parity in the worlds

thanks and happy weekend!

PR #19147, opened 2026-01-27
jinja : undefined should be treated as sequence/iterable (return string/array) by filters/tests

CISC:
Fixes #19130

### DIFF
--- a/common/jinja/value.cpp
+++ b/common/jinja/value.cpp
@@ -115,6 +115,18 @@ static T slice(const T & array, int64_t start, int64_t stop, int64_t step = 1) {
 }
 
 template<typename T>
+static value empty_value_fn(const func_args &) {
+    if constexpr (std::is_same_v<T, value_int>) {
+        return mk_val<T>(0);
+    } else if constexpr (std::is_same_v<T, value_float>) {
+        return mk_val<T>(0.0);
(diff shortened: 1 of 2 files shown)


CISC:
@pwilkin I think maybe this fixes some of your template issues in #18675?

pwilkin:
@CISC probably :) I'll tell you, but I'm doing a big refactoring to make the code more structured and maintainable and haven't re-merged with master in a couple of days.

common/jinja/value.cpp
```diff
@@ -1012,9 +1017,86 @@ const func_builtins & value_none_t::get_builtins() const {
 const func_builtins & value_undefined_t::get_builtins() const {
     static const func_builtins builtins = {
         {"default", default_value},
-        {"tojson", [](const func_args & args) -> value {
-            args.ensure_vals<value_undefined>();
-            return mk_val<value_string>("null");
+        {"capitalize", [](const func_args &) -> value {
```
ngxson: we can also define each function once, then use map it here to make the code more concise. example:

```cpp
static const auto make_str = [](const func_args &) -> value { return mk_val<value_string>(); }
...
{"capitalize", make_str},
{"first", make_str},
```

ngxson (approved):

PR #19150, opened 2026-01-28
hexagon: enable offloading to Hexagon on Windows on Snapdragon

tboinovski1:
GGML Hexagon backend updates to support Windows on Snapdragon.
 
Features:

- Support for building and offloading to NPU on WoS.
- A common/wrapper interface for Android and Windows DSP drivers.

 
NPU artifacts must be signed before running on a Windows device. For details on building and running refer to _docs/backend/snapdragon/README.md_.
For details on driver installation, signing generated artifacts and preparing your Windows device refer to _docs/backend/snapdragon/windows.md_.
 
Side effects:

- Refactors the dynamic loading code (dl_load_library and friends) in ggml-backend-reg.cpp into a separate module to allow reuse in Hexagon backend too.

 
Known issues:

- Intermittent instability when running multi-session.


max-krasnyansky:
@ggerganov @CISC @slaren 
Please let us know if you guys are OK with factoring out `ggml-backend-dl`.
We need to load stuff dynamically from the Hexagon backend.

ggml/src/ggml-hexagon/CMakeLists.txt
```diff
@@ -25,56 +35,71 @@ else()
...
 endif()
 
-link_custom_library(htp_iface cdsprpc)
-link_custom_library(htp_iface rpcmem)
-
 set(TARGET_NAME ggml-hexagon)
 ggml_add_backend_library(${TARGET_NAME}
-    ggml-hexagon.cpp htp-utils.c htp-utils.h ../../include/ggml-hexagon.h)
+    ggml-hexagon.cpp
+    ../ggml-backend-dl.cpp
```
ggerganov: Is the `ggml-backend-dl.cpp` source necessary to be specified here? You already link the backend library to `libggml`

max-krasnyansky: This did look suspicious to me earlier but I convinced myself we needed it :) 

```
lld-link: error: undefined symbol: struct HINSTANCE__* __cdecl dl_load_library(class std::filesystem::path const &)
>>> referenced by C:/Users/HCKTest/src/llama.cpp/ggml/src/ggml-hexagon/htp-drv.cpp
>>>               ggml/src/ggml-hexagon/CMakeFiles/ggml-hexagon.dir/htp-drv.cpp.obj
```

Looks like we're only linking ggml-base

```
C:\windows\system32\cmd.exe /C "cd . && C:\PROGRA~1\LLVM\bin\CLANG_~1.EXE --target=arm64-pc-windows-msvc -nostartfiles -nostdlib -march=armv8.7a+fp16 -fvectorize -ffp-model=fast -flto -D_GNU_SOURCE -O3 -DNDEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrt  -shared -fuse-ld=lld-link -o bin\ggml-hexagon.dll  -Xlinker /MANIFEST:EMBED -Xlinker /implib:ggml\src\ggml-hexagon\ggml-hexagon.lib -Xlinker /pdb:bin\ggml-hexagon.pdb -Xlinker /version:0.9 ggml/src/ggml-hexagon/CMakeFiles/htp_iface.dir/htp_iface_stub.c.obj ggml/src/ggml-hexagon/CMakeFiles/ggml-hexagon.dir/ggml-hexagon.cpp.obj ggml/src/ggml-hexagon/CMakeFiles/ggml-hexagon.dir/htp-drv.cpp.obj  
ggml/src/ggml-base.lib  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -loldnames  && cd ."
```

Which comes from `ggml_add_backend_library`
```
    target_link_libraries(${backend} PRIVATE ggml-base)
```

max-krasnyansky (approved):

ggml/src/ggml-hexagon/CMakeLists.txt
max-krasnyansky: @ggerganov 
Ok to merge or do you want to iterate on how the dl_ stuff is handled?

ggerganov: I think it's better to copy the dl code in the backend. Including sources across library boundaries is not a good idea. You can keep the dl extraction in separate files as you suggest, as it is OK from refactoring standpoint. But the hexagon backend should have its own copy of this code.

max-krasnyansky: Sounds good. We'll duplicate it for now. Maybe we can add the `dl_` stuff to `ggml-base` or perhaps introduce something like `ggml-backend-base` if other backends end up needing it.

ggerganov: Moving to `libggml-base` would make it have to link to `libdl`. Don't think it's a good enough reason to add this dependency for now.

PR #19157, opened 2026-01-28
convert : yield Mamba2Model/GraniteMoeModel modify_tensors

danbev:
This commit updates the `GraniteHybridModel` class' modify_tensors function to properly delegate to `Mamba2Model.modify_tensors` and `GraniteMoeModel.modify_tensors` using 'yield from' instead of 'return'.

The motivation for this is that modify_tensors is a generator function (it uses 'yield from'), but the two calls above use return statements but don't yield anything which means that the the caller of this function will not receive any yielded values from it. And this causes layer tensors to be silently dropped during conversion.

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -8912,7 +8912,8 @@ class GraniteHybridModel(Mamba2Model, GraniteMoeModel):
             name.endswith("block_sparse_moe.input_linear.weight")
             or "shared_mlp" in name
         ):
-            return GraniteMoeModel.modify_tensors(self, data_torch, name, bid)
+            yield from return GraniteMoeModel.modify_tensors(self, data_torch, name, bid)
+            return
 
         # Determine whether this is a mamba layer or an attention layer
         if bid in self._ssm_layers:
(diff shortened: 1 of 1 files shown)


convert_hf_to_gguf.py
CISC: Missed this one.

CISC (approved):
Oof, the multiple inheritance models were a pain. :)

convert_hf_to_gguf.py
danbev: Thanks, I've updated it now.

convert_hf_to_gguf.py
```diff
@@ -8912,13 +8912,16 @@ def modify_tensors(
             name.endswith("block_sparse_moe.input_linear.weight")
             or "shared_mlp" in name
         ):
-            return GraniteMoeModel.modify_tensors(self, data_torch, name, bid)
+            yield from return GraniteMoeModel.modify_tensors(self, data_torch, name, bid)
```
```suggestion
            yield from GraniteMoeModel.modify_tensors(self, data_torch, name, bid)
```

danbev: Thanks again!

PR #19178, opened 2026-01-29
Finish “TODO: avoid using atexit() here by making `console` a singleton”

CoraBlack:
## Changed:
- move functions in namespace console to static class console
- clean up the console after cli program finished automatically by RAII(static single instance)

## note:
- without test in Windows (because my Windows development environment is not complete)

### DIFF
--- a/tools/cli/cli.cpp
+++ b/tools/cli/cli.cpp
@@ -92,10 +92,10 @@ struct cli_context {
         }
 
         // wait for first result
-        console::spinner::start();
+        console::start();
         server_task_result_ptr result = rd.next(should_stop);
 
-        console::spinner::stop();
+        console::stop();
         std::string curr_content;
         bool is_thinking = false;
 
@@ -211,7 +211,7 @@ int main(int argc, char ** argv) {
 
     // TODO: avoid using atexit() here by making `console` a singleton
     console::init(params.simple_io, params.use_color);
-    atexit([]() { console::cleanup(); });
+    // atexit([]() { console::cleanup(); });
 
     console::set_display(DISPLAY_TYPE_RESET);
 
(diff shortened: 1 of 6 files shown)


tools/cli/cli.cpp
```diff
@@ -230,14 +228,14 @@ int main(int argc, char ** argv) {
...
-    console::spinner::start();
+    console::start();
     if (!ctx_cli.ctx_server.load_model(params)) {
-        console::spinner::stop();
+        console::stop();
         console::error("\nFailed to load the model\n");
         return 1;
     }
 
-    console::spinner::stop();
```
ngxson, line 240: I don't get it. why spinner is removed?

ngxson (changes requested):

tools/cli/cli.cpp
CoraBlack: Oh...that is my problem. At the time I thought it just used for grouping methods.
I thought i misunderstood its meaning in code and i just want to make it look like a whole

I wll reread the code and make modifications, thanks for your review!

I have fixed it. Add a struct console::spinner to replace the original namespace console::spinner and restore start() and stop() methods in it. And I have fixed the compiling problem from github bots(I didn't notice it at that time)

CoraBlack: sorry, i don't notice the build information from the github bot. I have fixed the compiling problem.

CoraBlack:
@ggerganov Excuse me, could you review my code?
(I am not sure the process of pr review. I never make any pr before)

ggerganov:
This is not how the singleton pattern is done in C++.

Given that you don't have a lot of experience, I would recommend contributing to some other project first to learn the basics. llama.cpp can be quite difficult for beginners.

CoraBlack:
Thanks for your review.

PR #19179, opened 2026-01-29
ggml-backend: fix async set/get fallback sync

JohannesGaessler:
While working on an implementation for backend-agnostic tensor parallelism I found what I believe to be a bug in the ggml backend code. For a minimal implementation I did at first not implement `set_tensor_async` and `get_tensor_async` assuming that I could just rely on the synchronous fallback and implement those later. However, `set_tensor_async` and `get_tensor_async` do not call `ggml_backend_synchronize` for their fallback so I got incorrect results. This PR adds the corresponding calls.


slaren (approved):
This change looks good to me. The copy should happen after all previously queued operations, so a synchronization is necessary if the backend does not implement async get/set.

PR #16817, opened 2025-10-28
Implement SparseK Attention mechanism — new GGML operator with CPU backend (GPU planned next)

yael-works:
###  New Attention Mechanism: SparseK Dynamic Attention (CPU, Graph-Level Prototype)

# PR Description — 
This PR integrates an experimental **SparseK dynamic attention mechanism** into the llama.cpp compute graph on the **CPU execution path**, built entirely from existing GGML operations.  
**No new GGML operator or low-level CPU kernels** are introduced in this PR.

The purpose of this PR is to establish **correct graph logic**, **metadata handling**, and **test coverage**, before adding optimized kernels or GPU/SYCL support in a follow-up PR.

---

##  Overview

SparseK introduces selective sparsity into attention using:

- **Top-K filtering** (fully implemented in this PR)
- **Local window** (metadata + config only; implementation planned)
- **Global stride** (metadata + config only; implementation planned)

At runtime, SparseK refines the base KQ mask (causal / cross / SWA) by selectively allowing only the strongest or relevant attention positions.

---

### **1. Graph-level SparseK mask building**

**Files:** `llama-graph.cpp`, `llama-graph.h`

This PR adds two new graph functions:

- `build_sparsek_mask(q, k, base_mask, il)`
- `maybe_apply_sparsek_mask(base_mask, q, k, n_kv, n_rows, n_stream, il)`

### **Implemented behavior:**

- Compute content-based scores via `ggml_mul_mat(k, q)`
- Reshape to 2D and apply `ggml_top_k`
- Construct a SparseK 0 / -INF mask using:
  - `ggml_get_rows`
  - `ggml_set_rows`
  - reshaping between 2D / 3D / 4D
- Combine the SparseK mask with the existing base mask
- Call `ggml_flash_attn_ext` *once* with the final mask

### **Important**
- Only **Top-K sparsity** is implemented at this stage.  
- Window and stride are **loaded and carried through**, but **not yet applied** in mask generation.

---

### **2. SparseK metadata & hyperparameters**

**Files:**  
`llama-hparams.h`, `llama-model.cpp`, `llama-model-loader.cpp`

Added optional HParams:

- `sparsek_enable`
- `sparsek_topk`
- `sparsek_window`
- `sparsek_stride`

These are read from GGUF keys if present:

- `llama.sparsek.enable`
- `llama.sparsek.top_k`
- `llama.sparsek.window`
- `llama.sparsek.stride`

Runtime graph receives these values and applies them consistently.

---

[...]

PR #19183, opened 2026-01-29
[RFC] implementing Expected Attention in llama.cpp?

ddh0:
Recently, NVIDIA put out a paper describing a method to reduce the size of the KV cache. They call their preferred method _Expected Attention_.

From my reading of [the paper](https://arxiv.org/pdf/2510.00636v1), the thought process is like this:
- Previous works have shown that the hidden states (`cur`) at each layer follow a gaussian distribution (and they re-confirmed this here too)
- The query projection tensor in each layer (`wq`) can be thought of as a linear transformation on `cur`, which maintains this same distribution for the resulting `Qcur` projection
- Using the callback `cb()`, we can observe the query projections `Qcur` in all layers over an interval, in order to model a distribution of the _expected future queries_ (the paper also takes into account RoPE / positional encoding for future positions, but I could not figure out how to do that, so I skipped it for now)
- Using the expected future queries (query?) we can compute the expected importance of each entry currently in the KV cache. They assign an importance score to all KV pairs based on the **magnitude of the expected contribution** of that pair (assuming the expected future query).
- They sort the KV entries based on the importance scores, and discard the bottom X% of entries.

So for example if the user specified a "compression ratio" of 20%, we would discard the 20% of KV entries that are expected to have the least impact on the hidden state in the future.

---

I had hoped to be able to get this working and submit a PR for it on my own, but there are some problems that I don't know how to solve:
- How to get the actual KV entries from the `llama_context`? The `llama_memory` structure abstracts away the implementation of the different cache types, so this is not straightforward.
- How to efficiently compute the expected attention scores for all KV entries? (It would probably be best to score per-layer and combine all those scores into one unified importance ranking somehow.)
- The paper re-computes the expected attention 512 tokens of inference. It might be better/easier to only re-compute when we are done with prompt processing, and before we start text generation? I don't know.
- What do we do once we have the positions we want to discard? Would it be possible to use the existing `defrag` logic for this?

I am posting this thread in hopes of starting a discussion to see if this method will even be feasible to implement in llama.cpp, or if it would be too complicated. The code is certainly **NOT** in a state that is ready for use, or even guaranteed to work at all. I am just including my demo code as a starting point. I would be pleasantly surprised if this PR ever leaves the draft stage. :)


ddh0:
additional references:
- <https://github.com/NVIDIA/kvpress>
- <https://github.com/NVIDIA/kvpress/blob/main/kvpress/presses/expected_attention_press.py>

ngxson:
If I understand correctly, the idea can be explained in more simple terms:
- Whenever KV cache is used, we track the attention score of the KV entries
- If a given KV entry is used less than < X% of the time, then we simple remove/prune it (we define "being used" by compare attention score against a threshold)

If we view the model as a pure function `f(x, kv) = (y, kv_new)` where `x` the input token, `kv` the current cache state and `y` the next token, then what I understand is that we now need `f(x, kv) = (y, kv_new, attn_score)` for this to work, so the simple way is to add `attn_score` as an output node, much like how some models can output token and/or embeddings.

However, it might sounds more complicated than that. First off, flash attention won't expose the attention score to the outside world. And secondly, this may make it more tricky to reuse cache cross multiple sequences, same problem with KV shifting/SWA.

Also, just to mention that there were some similar works in https://github.com/ggml-org/llama.cpp/pull/16817 , but might not very useful as the other PR is AI-coded and the author doesn't seem to fully understand it.

PR #19188, opened 2026-01-29
Bump cmake max version (needed for Windows on Snapdragon builds)

max-krasnyansky:
`arm64-windows-snapdragon-release` builds were failing due to an older policy in CMake.
We need 3.28 but bumping min version breaks CI on other platforms. 
Bumping min version via Presets doesn't work.
So bumping max version seems like a good compromise.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14) # for add_link_options and implicit target directories.
+cmake_minimum_required(VERSION 3.14...3.28) # for add_link_options and implicit target directories.
 project("llama.cpp" C CXX)
 include(CheckIncludeFileCXX)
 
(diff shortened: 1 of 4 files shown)


max-krasnyansky:
@CISC @ggerganov 
Ok with this?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14) # for add_link_options and implicit target directories.
+cmake_minimum_required(VERSION 3.14...3.28) # for add_link_options and implicit target directories.
```
ggerganov: Isn't it better to do it in `ggml/CMakeLists.txt` instead of here?

max-krasnyansky: Yep. Good idea. Moved.

ggerganov (approved):

PR #17108, opened 2025-11-08
vulkan: fix mmq out of bounds reads

0cc4m:
Fixes #17056

I also cleaned up some legacy indexing and buffer size code in the matmul functions.

Draft because I caused a mul_mat_id regression, still trying to find the reason.

PR #19194, opened 2026-01-30
Correctly fetch q8_1 quantize pipeline in test as needed by 8a3519b

sredman:
#17108 removed the `pipeline_quantize_q8_1` from the vk_device struct. Based on other changes in that PR, I believe this is the correct way to retrieve the pipeline now.

This requires building with the `GGML_VULKAN_RUN_TESTS` option to expose the previous build break.

This PR was not prepared with the help of AI, but still please don't ask me too many hard questions or else the cracks will appear :smiling_face_with_tear:.


jeffbolznv (approved):

0cc4m (approved):

0cc4m:
Thank you

PR #19195, opened 2026-01-30
Remove pipeline cache mutexes

nikhilJain17:
Now that `webgpu_context` is per-thread, we can remove mutexes from pipeline caches. We cannot remove mutexes from `webgpu_buf_pool` since they are allocated and freed in callback threads, and we cannot remove the mutex from the memset buffer pool since it is shared by all ggml buffers.


reeselevine (approved):
nice!

PR #19207, opened 2026-01-30
server : wrap around the "id_slot" parameter

ggerganov:
```c++
    server_slot * get_slot_by_id(int id_slot) {
        // note: allow id_slot to be out of bounds (wrap around)
        id_slot = id_slot % slots.size();

    ...
```

This allows to distribute tasks without knowing the number of slots that the server has.

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -1763,7 +1763,7 @@ private:
                         break;
                     }
 
-                    int id_slot = task.slot_action.id_slot;
+                    const int id_slot = task.slot_action.id_slot;
                     server_slot * slot = get_slot_by_id(id_slot);
                     if (slot == nullptr) {
                         send_error(task, "Invalid slot ID", ERROR_TYPE_INVALID_REQUEST);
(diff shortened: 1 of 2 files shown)


ngxson (approved):

tools/server/server-context.cpp
```diff
@@ -155,7 +155,7 @@ struct server_slot {
     double t_prompt_processing; // ms
     double t_token_generation;  // ms
 
-    std::function<void(int /* slot_id */)> callback_on_release;
+    std::function<void(int /* id_slot */)> callback_on_release;
```
aviallon, line 158: Would it make sense to convert the slot id to be an unsigned integer of known size instead? Like a uint32_t? Or a size_t?

Because if we wrap around the max size of a signed int, isn't that UB?

What I'd like to do is to be able to hash a conversation_id to a fixed-width integer, in order to more accurately map conversations to slots from my middleware.

Would thar be possible?

ngxson: I think there is a case somewhere in the code where `slot_id == -1` means "use whatever free slot"

But we can replace it with `SIZE_MAX` if needed, should not be too complicated. The only problem is that I don't know how `nlohmann::json` handle large int values.

aviallon: From the docs, it looks like, by default, they are converted to `std::int64`

https://json.nlohmann.me/api/basic_json/number_integer_t/

PR #19211, opened 2026-01-30
support MiniCPM-o 4.5(vision only)

tc-mb:
Add support for MiniCPM‑o 4.5 and document new versioning scheme

1. We plan to open-source MiniCPM‑o 4.5 (o45) next week, a new generation of Omni model, which has excellent performance.
2. We’re moving away from the previous sequential version scheme. Apologies for the earlier imperfect design. New scheme:
v series: use the major.minor number directly (e.g., v4.5 → 45)
o series: prefix with 1000 to distinguish (e.g., o4.5 → 100045)
No behavior changes for existing versions. This only adds support for the new 100045 code.

### DIFF
--- a/tools/mtmd/clip.cpp
+++ b/tools/mtmd/clip.cpp
@@ -1005,6 +1005,8 @@ struct clip_model_loader {
                         hparams.minicpmv_query_num = 64;
                     } else if (hparams.minicpmv_version == 6) {
                         hparams.minicpmv_query_num = 64;
+                    } else if (hparams.minicpmv_version == 100045) {
+                        hparams.minicpmv_query_num = 64;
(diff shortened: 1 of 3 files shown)


ngxson (approved):
Would you consider reimplementing this in `convert_hf_to_gguf.py`? Most models are already migrated, but unfortunately I don't have much time to work on minicpm-v

tc-mb:
> Would you consider reimplementing this in `convert_hf_to_gguf.py`? Most models are already migrated, but unfortunately I don't have much time to work on minicpm-v

Yes, what you say is very reasonable. I have been busy developing this version of the omni model before, and I will modify it after it is released.

Issue #19221, opened 2026-01-31
Misc. bug: Intel IGPU Perf Regression and Possible Faulty GPU Detection Logic


virajwad:
### Name and Version

llama-cli.exe --version
version: 7860 (88d23ad51)
built with Clang 19.1.5 for Windows x86_64

### Operating systems

Windows

### Which llama.cpp modules do you know to be affected?

llama-bench

### Command line

```shell
llama-bench.exe -p 512 -r 10 -n 0 -m DeepSeek-R1-0528-Qwen3-8B-Q4_K_M.gguf
```

### Problem description & steps to reproduce

**Model used in test**
https://huggingface.co/unsloth/DeepSeek-R1-0528-Qwen3-8B-GGUF/blob/main/DeepSeek-R1-0528-Qwen3-8B-Q4_K_M.gguf

**Problem Summary:**
Seems there is a perf regression on Intel IGPU when going from b7858 --> b7860

**How Regression was found:**
I noticed an Intel regression between b7657 and b7885, and painstakingly binary searched llama.cpp releases until I could narrow down two versions. From there I swapped back and forth between versions, ran many times, and made sure the regression was reproducible.

**Reproduce steps:**
- In CMD, use command above on b7858 and b7860 with the DeepSeek-Qwen3 8B Q4_K_M model linked in this issue. Run a few times. Not sure if it shows up on other hardware, but it is showing up on Intel IGPU.
- Performance in llama-bench seems to have higher variance on b7860 and is around 10-13% slower, when compared to b7858
- Check GPU.Z or Task Manager - The IGPU's memory space uses around 4-5% more allocated memory in b7860 case.

### First Bad Commit

88d23ad515e60c64db8df22b3de23cba00d044ab

### B7858 Case:
load_backend: loaded RPC backend from C:\Users\dungeon\Downloads\llama-b7858-bin-win-vulkan-x64\ggml-rpc.dll
**_ggml_vulkan: Found 1 Vulkan devices:_**
ggml_vulkan: 0 = Intel(R) Graphics <redacted driver> (Intel Corporation) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: KHR_coopmat
load_backend: loaded Vulkan backend from C:\Users\dungeon\Downloads\llama-b7858-bin-win-vulkan-x64\ggml-vulkan.dll
[...]

PR #19222, opened 2026-01-31
vulkan: fix GPU deduplication logic.

okuvshynov:
As reported in https://github.com/ggml-org/llama.cpp/issues/19221, the (same uuid, same driver) logic is problematic for windows+intel igpu.

Let's just avoid filtering for MoltenVK which is apple-specific, and keep the logic the  same as before 88d23ad5 - just dedup based on UUID.

Verified that MacOS + 4xVega still reports 4 GPUs with this version.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -5538,18 +5538,19 @@ static void ggml_vk_instance_init() {
                 // Check if there are two physical devices corresponding to the same GPU
                 // This handles the case where the same GPU appears with different drivers (e.g., RADV + AMDVLK on Linux),
                 // see https://github.com/ggml-org/llama.cpp/pull/7582 for original deduplication.
-                // However, for MoltenVK on macOS, multiple GPUs on the same card may report the same UUID,
-                // see https://github.com/KhronosGroup/MoltenVK/issues/2683. Until this is fixed, we'll only deduplicate
-                // when drivers differ (same driver + same UUID = likely different GPUs)
+                // MoltenVK on macOS may report the same UUID for distinct GPUs on multi-GPU cards,
+                // see https://github.com/KhronosGroup/MoltenVK/issues/2683. Skip dedup for MoltenVK.
                 auto old_device = std::find_if(
                     vk_instance.device_indices.begin(),
                     vk_instance.device_indices.end(),
                     [&devices, &new_id, &new_driver](const size_t k){
+                        if (new_driver.driverID == vk::DriverId::eMoltenvk) {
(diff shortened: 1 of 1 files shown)


virajwad:
Thanks @okuvshynov !!

I checked on latest llama.cpp master and your PR, with Intel IGPU system. I can confirm it fixes the issue :) 

[Previous Perf](https://github.com/ggml-org/llama.cpp/issues/19221)

**Perf with this PR:**
ggml_vulkan: 0 = Intel(R) Graphics (Intel Corporation) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: KHR_coopmat
| qwen3 8B Q4_K - Medium         |   4.68 GiB |     8.19 B | Vulkan     |  99 |           pp512 |       966.37 ± 18.38 |

build: 9676add96 (7897)

**Checking GPU.Z:**
Went from _2 Vulkan devices found_ to _1 Vulkan device found_
Peak memory used during GPU runtime went from 5932 MB --> 5730 MB after fix

The variance issue is also resolved, I'm getting pretty consistent results when running `llama-bench.exe -p 512 -r 10 -n 0 -m <model>` again and again.

Thanks for making the fix so quickly! Looks good to me.
@0cc4m

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -5538,18 +5538,19 @@ static void ggml_vk_instance_init() {
...
-                // However, for MoltenVK on macOS, multiple GPUs on the same card may report the same UUID,
-                // see https://github.com/KhronosGroup/MoltenVK/issues/2683. Until this is fixed, we'll only deduplicate
-                // when drivers differ (same driver + same UUID = likely different GPUs)
+                // MoltenVK on macOS may report the same UUID for distinct GPUs on multi-GPU cards,
+                // see https://github.com/KhronosGroup/MoltenVK/issues/2683. Skip dedup for MoltenVK.
                 auto old_device = std::find_if(
                     vk_instance.device_indices.begin(),
                     vk_instance.device_indices.end(),
                     [&devices, &new_id, &new_driver](const size_t k){
+                        if (new_driver.driverID == vk::DriverId::eMoltenvk) {
```
0cc4m: I think there's a chance this deduplication code may be required on MacOS for KosmicKrisp and MoltenVK, so best only apply this if new and old driverID is MoltenVK.

okuvshynov: Makes sense, will do

0cc4m (approved):
Thank you!

PR #19225, opened 2026-01-31
ISWA: Switch to base cache for position tracking

cmp-nct:
seq_pos_min/max should report logical range: updated position tracking to use base cache instead of SWA cache

The API should return seq position range but returns internal SWA-cache state instead of logical sequence range

During kv cache manipulations (seq_rm etc) I've ran into the bug that alternating swa model (gpt-oss-20b) wrongly freed up tokens from kv_swa that were in use (before the 128 window

This fix solved my issues (no more token losses) but I'm not familiar with that area of the codebase.

### DIFF
--- a/src/llama-kv-cache-iswa.cpp
+++ b/src/llama-kv-cache-iswa.cpp
@@ -107,12 +107,14 @@ void llama_kv_cache_iswa::seq_div(llama_seq_id seq_id, llama_pos p0, llama_pos p
 }
 
 llama_pos llama_kv_cache_iswa::seq_pos_min(llama_seq_id seq_id) const {
-    // the base cache is a superset of the SWA cache, so we can just check the SWA cache
-    return kv_swa->seq_pos_min(seq_id);
+    // base cache preserves all positions (no SWA cell reuse), making it the ground truth for position tracking
+    // SWA cache can lose older positions when cells are reused for new tokens outside the SWA window
+    return kv_base->seq_pos_min(seq_id);
 }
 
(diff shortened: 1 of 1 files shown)


cmp-nct:
Another related PR: https://github.com/cmp-nct-org/llama.cpp/pull/9
That one is against find_slot corrupting the kv cache of other sequences when restoring a kv state.
It solved the corruption in my case. Though someone familiar with iswa should take a look if I am hunting ghosts.

ggerganov:
The change is not valid - these functions intentionally work with the SWA cache since we have lost information about all tokens that are not in the SWA cache.

cmp-nct:
> The change is not valid - these functions intentionally work with the SWA cache since we have lost information about all tokens that are not in the SWA cache.

Thanks for your review.
I guess the problem I ran into is that I just can hardcode p_min to be always zero in application level programming.
p_max is already the end in both kv_base and kv_swa.
The returned seq_min is invalid to be used for application level (because it reports swa min, not the actual min_pos of kv_base
I overcomplicated it., though a comment in the API might be useful to inform that ISWA min_pos is SWA POV


@ggerganov  If you have a spare minute:
https://github.com/cmp-nct-org/llama.cpp/pull/9
I had find_slot corrupt my KV on state restoration when multiple sequences exist.
Maybe also wrong, but that change solved the corruption.

ggerganov:
Yes, the current logic of restoring sequences can erase other "out-of-SWA-scope" sequence data when SWA is involved. This is by design - one of the main advantages of SWA is that it reduces the KV cache, so the logic is implemented around that.

You can use split KV cache (`--no-kv-unifed --swa-full`) if you really want to keep the full SWA sequences.

PR #19234, opened 2026-01-31
chat: add thinking/reasoning support for Kimi K2

htxsrl:
## Summary

- Pass `enable_thinking` to the Kimi K2 chat template via `additional_context`, so the Jinja template can render thinking-related tokens correctly
- Handle `<think>` tags at the end of the generated prompt: when thinking is disabled, immediately close with `</think>`; when enabled, set `thinking_forced_open = true`
- Enables proper reasoning mode support for Kimi 2.5 / K2 models (consistent with how other models like DeepSeek and QwQ handle thinking)

## Context

Currently the Kimi K2 handler in `common_chat_params_init_kimi_k2` does not pass the `enable_thinking` flag to the template, and does not manage the `<think>` tag lifecycle. This means reasoning mode doesn't work correctly for Kimi 2.5 models — the model either always thinks or never does, regardless of the `--reasoning` flag.

This patch aligns Kimi K2's thinking support with the pattern already used by other model handlers in `chat.cpp`.

## Test plan

- [ ] Run `llama-server` with a Kimi 2.5 GGUF and `--reasoning` enabled, verify `<think>...</think>` blocks appear in output
- [ ] Run without `--reasoning`, verify thinking is suppressed (prompt ends with `</think>` immediately)
- [ ] Verify tool calling still works correctly with and without reasoning enabled

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1882,8 +1882,21 @@ static common_chat_params common_chat_params_init_kimi_k2(const common_chat_temp
     common_chat_params data;
     data.grammar_lazy = params.tools.is_array() && !params.tools.empty() && params.tool_choice != COMMON_CHAT_TOOL_CHOICE_REQUIRED;
 
-    data.prompt = apply(tmpl, params);
+    json additional_context = {
+        {"thinking", params.enable_thinking},
+    };
+
+    data.prompt = apply(tmpl, params, /* messages_override= */ std::nullopt, /* tools_override= */ std::nullopt, additional_context);
     data.format = COMMON_CHAT_FORMAT_KIMI_K2;
+
+    // Handle thinking tags based on prompt ending
+    if (string_ends_with(data.prompt, "<think>\n") || string_ends_with(data.prompt, "<think>")) {
+        if (!params.enable_thinking) {
+            data.prompt += "</think>";
+        } else {
+            data.thinking_forced_open = true;
+        }
+    }
 
     data.preserved_tokens = {
         "<think>",


pwilkin:
Need someone to test it with a live K2 model.

Lissanro:
I tried this patch with Q4_X quant of K2.5, but both in Roo Code and SillyTavern, it still produces thinking even though I ran without the `--reasoning` flag and "Enable Reasoning Effort" unchecked in Roo Code. Tool calling works though, but reasoning stays enabled. I tried with --jinja flag, and the original Moonshot chat template. If I missed a step to test this correct, please let me know.

pwilkin:
@Lissanro I think you have to run the server with `--reasoning-budget 0` to disable thinking.

Lissanro:
@pwilkin OK, with `--reasoning-budget 0` it worked. But then I do not understand the purpose of this patch... because `--reasoning-budget 0` already works without it, I tested with the official Moonshot template from their K2.5 huggingface repo.

I looked at the patch more closely and I think it does not implemented the described `--reasoning` flag that the summary explains and suggests testing with. The summary also implies changing the default behavior to non-reasoning ("Run without --reasoning, verify thinking is suppressed"), but it does not do that either.

So my question is, what is the issue this patch supposed to solve? Is there something that does not work without it as expected and what are the steps to reproduce? Because, like I mentioning, `--reasoning-budget 0` worked for me both with or without the patch. I am happy to help by testing any edge case (as long as there are clear testing steps to follow).

Lissanro:
I experimented with a bit further I think the real difference is that without the patch, `<think></think>` tokens are not produce if `--reasoning-budget 0` is used, which causes the model reply more in "thinking style".

With the patch, `<think></think>` block is produced with `--reasoning-budget 0`, making LLM replies more natural, like non-thinking LLMs would reply. Tool calls still work: I have tested non-reasoning mode with Roo Code and tool calls worked with the patch applied and `--reasoning-budget 0` set, making K2.5 spend much less tokens, somewhat similar like the old non-thinking K2 0905 used to work, which is expected result with the reasoning disabled.

Without the patch, K2.5 even with `--reasoning-budget 0` would use much more tokens to explain its actions instead of just doing them. For example, without the patch it would explain why it wants to read a file, with the patch it would just read the file without any explanation at all, but still can give me one if I ask.

Overall, I think this patch is useful after all, allows to save tokens if I know upfront the task does not really need reasoning. But patch's summary was very confusing, making it hard to understand the purpose of the patch and how to test.

htxsrl:
Hi, I proposed the patch, because without the patch, K2.5's default output mixes thinking and response together - the reasoning "leaks" into the main output even in normal usage.

pwilkin:
> Hi, I proposed the patch, because without the patch, K2.5's default output mixes thinking and response together - the reasoning "leaks" into the main output even in normal usage.

That might also be a bug in the reasoning detection logic currently in the server.

pwilkin:
@htxsrl Could you please try building the autoparser PR branch and see if thinking renders properly there?

htxsrl:
> @htxsrl Could you please try building the autoparser PR branch and see if thinking renders properly there?

Yes, I will try tomorrow

htxsrl:
> @htxsrl Could you please try building the autoparser PR branch and see if thinking renders properly there?

I built and tested the autoparser PR branch and I’m still seeing the same issue.
I compiled llama.cpp from the autoparser PR branch and launched llama-server with Kimi-K2.5 Q3_K_M.
With default parameters (-n -1 -cnv --ctx-size 8192 --threads 16 --n-gpu-layers 999 --batch-size 512 --mlock) the server mixes reasoning and answer content, exactly as I experienced on the main branch when connecting to localhost:8080.
Launching the server with --reasoning-format deepseek does not change the behavior.
I also tried forcing the format at request time via curl:
"reasoning_format": "deepseek"
but the thinking content is still mixed into the final answer.
Below is a more detailed write-up of the tests and findings.

Testing Autoparser PR #18675 with Kimi-K2.5
Environment
	•	OS: macOS Darwin arm64 (Apple M3 Ultra)
	•	Model: Kimi-K2.5 (Q3_K_M quantization, ~456 GB)
	•	Branch: PR #18675 (autoparser)
Test Procedure
	1	Built llama-server from the autoparser PR branch
	2	Started the server with the Kimi-K2.5 model
	3	Sent chat completion requests using reasoning_format: "deepseek"

Issue Found
The autoparser incorrectly detects reasoning_mode::TOOLS_ONLY for the Kimi-K2.5 template.

Running llama-debug-template-parser on the model reports:
```
reasoning_mode: TOOLS_ONLY
reasoning_start: '<think>'
reasoning_end: '</think>'
```
Because of this, the reasoning content is not extracted into message.reasoning_content. Instead, the </think> tag appears verbatim in message.content, for example:
```
{
  "role": "assistant",
  "content": "The user is asking... I should present this clearly.</think>Here's the step-by-step calculation..."
}
```


Expected Behavior
The reasoning block should be extracted into a separate reasoning_content field, leaving message.content clean.

Root Cause Analysis
In chat-auto-parser-generator.cpp:88:

`if (extract_reasoning && enable_thinking && analysis.reasoning != reasoning_mode::NONE) {`

The problem originates earlier: the differential analyzer (chat-diff-analyzer.cpp:477) classifies the template as reasoning_mode::TOOLS_ONLY.

This happens because it detects reasoning markers as conditional on tool usage in test renders.
However, Kimi-K2.5’s template always starts generation with a <think> tag, as defined in the template’s add_generation_prompt section, regardless of whether tools are used.

Suggested Fix
The Kimi-K2.5 template should be detected as:
	•	reasoning_mode::FORCED_OPEN or
	•	reasoning_mode::TAG_BASED
instead of TOOLS_ONLY, since the <think> tag is unconditionally emitted at generation start.

chrislawso:
.

devrandom:
I was going to submit an almost identical PR.  I can confirm this fixes the K2.5 issue in our setup.

Lissanro:
This #19234 patch fixes https://github.com/ggml-org/llama.cpp/issues/20008 - currently with Kimi K2.5 it fails correctly handle thinking block if `<think>` or `<think>\n` was at the end of the prompt. After applying this patch, Kimi K2.5 can correctly handle thinking in all cases - works everywhere I tried, including Roo Code or direct curl requests (please refer to https://github.com/ggml-org/llama.cpp/issues/20008 for more details about the issue).

To be honest, I forgot about this patch after a month, and when recently got back to using K2.5 with llama.cpp, I encountered the #20008 bug and ended up writing very similar patch https://github.com/ggml-org/llama.cpp/compare/master...Lissanro:llama.cpp:kimi-k2.5-think-fix and only remembered when was double checking open PRs searching for "Kimi" - but aside from minor style differences and comments that I copied from nearby similar code, this and mine patch are identical + this patch also handles thinking parameter for the chat template, while mine does not, so I decided not to submit my own PR, in favor of this one.

My suggestion would be to to consider merging this #19234 patch. I looked carefully into the file which this patch updates and as far as I can tell it correctly follows pattern established for other thinking models, which handle this in a similar way.

PR #19237, opened 2026-02-01
nix: fix allowUnfreePredicate for packages with multiple licenses

typedrat:
The `allowUnfreePredicate` in `pkgsCuda` wraps `p.meta.license` in a list, but `meta.license` can already be a list for packages with multiple licenses. This causes flake evaluation to fail when it tries to access `.free` on the nested list.

Fix by using `lib.toList` which handles both single licenses and license lists.


philiptaron (approved):
Lots of ways to do this, but this works.

PR #19239, opened 2026-02-01
jinja : add missing 'in' test to template engine (#19004)

sidmohan0:
Hey all! 

**Note: Fixes #19004**

## Context:

The jinja template parser was missing the `'in'` test from `global_builtins()`, causing templates using `reject("in", ...), select("in", ...), or 'x is in(y)' `to fail with `selectattr: unknown test 'in' `

This broke tool-calling for Qwen3-Coder and would break for any other model whose chat template uses the `in` test.

## What I did:

- Added `test_is_in` supporting array, string, and object containment checks, mirroring the existing 'in' operator logic in runtime.cpp.
- Includes test cases for all three containment types plus reject/select filter usage.

## Local Test environment:
  - Windows 11 26200 (x64), Intel i7-10700K, 64GB RAM
  - MSVC 17.14.25 (VS Build Tools 2022)
  - CMake 4.2.0-rc4
  - All 279 jinja tests pass (1241 assertions, 0 failures, 0 exceptions)
  
## Comments:

- Shouldn't impact existing entries as we're just adding a new entry; the only way this code runs is if a template explicitly uses the in test

## AI Disclosure
Built with Claude (AI-assisted). Fully tested locally. I understand what the code does.

Appreciate the look - let me know if you have any questions!

### DIFF
--- a/common/jinja/runtime.cpp
+++ b/common/jinja/runtime.cpp
@@ -144,6 +144,13 @@ value binary_expression::execute_impl(context & ctx) {
         return false;
     };
 
+    auto test_is_in = [&]() -> bool {
+        func_args args(ctx);
+        args.push_back(left_val);
+        args.push_back(right_val);
+        return global_builtins().at("test_is_in")(args)->as_bool();
+    };
+
     // Handle undefined and null values
     if (is_val<value_undefined>(left_val) || is_val<value_undefined>(right_val)) {
         if (is_val<value_undefined>(right_val) && (op.value == "in" || op.value == "not in")) {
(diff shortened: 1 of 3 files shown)


ngxson:
This PR appears to contain AI-generated code without proper disclosure or watermarking. Per our [contribution guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md), we cannot accept it in its current form.

sidmohan0:
> This PR appears to contain AI-generated code without proper disclosure or watermarking. Per our [contribution guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md), we cannot accept it in its current form.

Thank you for flagging - just updated the PR description.  Sincere apologies for the oversight - genuinely did not mean to mislead.

ngxson (approved):

PR #19242, opened 2026-02-01
sycl: implement GGML_OP_TOP_K

tdevelope:
Implements GGML_OP_TOP_K for the SYCL backend (F32).

The implementation finds the top K values per row and returns their indices, matching the CPU behavior.

### Limitations
- Supports k ≤ 32 due to register constraints
- Returns indices only (int32), not values

### Testing
Tested with `./build/bin/test-backend-ops support` on Intel Iris Xe GPU:
- k=1 to k=15: all tests pass
- k≥100: correctly reports as unsupported

Part of #14909


NeoZhangJianyu (approved):
It's great!
I will merge the PR after CI is passed.

Thank you!

PR #18467, opened 2025-12-29
vulkan: support buffer_from_host_ptr

jeffbolznv:
For https://github.com/ggml-org/llama.cpp/pull/18166#issuecomment-3695731015.

~Draft, the model loader change needs testing/polishing and to be merged with the other pending change.~ The change only affects the vulkan backend now.

PR #19254, opened 2026-02-01
Add test for vk_buffer from host memory

sredman:
Add a test to validate external host memory can correctly be allocated, written, used, and read.

I already had this test for my own external memory buffer solution, but then I noticed #18467 had done the work. I thought the test might still be useful.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -1569,6 +1569,7 @@ struct ggml_vk_garbage_collector {
 static void ggml_vk_preallocate_buffers(ggml_backend_vk_context * ctx, vk_context subctx);
 static void ggml_vk_load_shaders(vk_device& device);
 static void ggml_pipeline_allocate_descriptor_sets(ggml_backend_vk_context * ctx);
+static vk_buffer ggml_vk_buffer_from_host_ptr(vk_device & device, void * ptr, size_t size);
 
 static bool vk_memory_logger_enabled = false;
 
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -4484,7 +4484,9 @@ static void ggml_vk_load_shaders(vk_device& device) {
 
 static bool ggml_vk_khr_cooperative_matrix_support(const vk::PhysicalDeviceProperties& props, const vk::PhysicalDeviceDriverProperties& driver_props, vk_device_architecture arch);
 
+#if defined(GGML_VULKAN_RUN_TESTS)
 static void ggml_vk_test_single_device_buffer_from_host_ptr(vk_device& device);
+#endif
 
 static vk_device ggml_vk_get_device(size_t idx) {
     VK_LOG_DEBUG("ggml_vk_get_device(" << idx << ")");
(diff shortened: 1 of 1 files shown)


0cc4m:
I don't mind adding further backend-internal tests, but I think we should move them out of the main file. I plan to look into splitting it up in the future. Maybe we can start here by adding a ggml-vulkan-tests.cpp and .hpp? You don't have to move my tests as well, but you can if you want.

sredman:
> I don't mind adding further backend-internal tests, but I think we should move them out of the main file. I plan to look into splitting it up in the future. Maybe we can start here by adding a ggml-vulkan-tests.cpp and .hpp? You don't have to move my tests as well, but you can if you want.

Do you have the shape of a plan for this?

When I was originally working on this test, I tried to make test file in the ggml/tests path. That would have required an extensive refactoring in order to be able to reference the internal functions. We would need to mark any functions we want to call as non-static and export them. However, the more messy problem was that we'd need to extract all the struct definitions which we want to use, which is quite an undertaking. It could certainly be done, but would definitely need to be prepared and checked in quickly to avoid mass merge conflicts :slightly_smiling_face:

0cc4m:
That makes sense, I hadn't looked into it enough yet. Then I'll postpone that until I get to it myself. However, please move your test into the existing `#ifdef GGML_VULKAN_RUN_TESTS` area where the other tests are, including the call. The file is already way too large, I would like to avoid test code spread around everywhere.

sredman:
@0cc4m , I have moved the test declaration into the preexisting block of tests, and the call into the preexisting block of calls (in `ggml_vk_preallocate_buffers`).

ggml/src/ggml-vulkan/ggml-vulkan.cpp
sredman, line 12224: Note that this `abort` needs to be commented out in order to run the new tests.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -12095,6 +12096,89 @@ static void ggml_vk_test_dequant_matmul(ggml_backend_vk_context * ctx, size_t m,
...
     free(d_chk);
 }
+
+static void ggml_vk_test_single_device_buffer_from_host_ptr(vk_device& device) {
+    static std::vector<vk_device> tested_devices;
+    if (std::find(tested_devices.begin(), tested_devices.end(), device) == tested_devices.end()) {
+        tested_devices.push_back(device);
+    } else{
+        return; // Already tested this device, skip
+    }
```
0cc4m, line 12106: The existing tests just run once and abort, it's not intended to keep them active during runtime, so this test isn't necessary.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -12095,6 +12096,89 @@ static void ggml_vk_test_dequant_matmul(ggml_backend_vk_context * ctx, size_t m,
...
+
+    VK_LOG_DEBUG("ggml_vk_test_single_device_external_memory(" << device->name << ") - Single-device external host memory test");
```
0cc4m, line 12108: The debug name is inconsistent with the function name, and the additional text is unusual. Was that to make it easier to spot in the debug log?

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -12095,6 +12096,89 @@ static void ggml_vk_test_dequant_matmul(ggml_backend_vk_context * ctx, size_t m,
...
+    const size_t aligned_size = ((test_buffer_size + min_alignment - 1) / min_alignment) * min_alignment;
+
+    VK_LOG_DEBUG("ggml_vk_test_single_device_external_memory: min_alignment = " << min_alignment << ", aligned_size = " << aligned_size);
+
+    // Initialize host memory, which we will use as input to the device
+    auto input_deleter = [](float* p) { free(p); };
+    std::unique_ptr<float[], decltype(input_deleter)> input_host_memory(
+        static_cast<float*>(aligned_alloc(min_alignment, aligned_size)),
+        input_deleter
+    );
```
0cc4m, line 12127: This is a little convoluted compared to just going `float * input_host_memory = (float *)aligned_alloc(...)` and adding the `free()` call to the end, but if you prefer this, I don't mind.

However, to make it easier to understand what is going on, please check if the alloc succeeded and print an error message if not. Same for the other alloc below.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -12095,6 +12096,89 @@ static void ggml_vk_test_dequant_matmul(ggml_backend_vk_context * ctx, size_t m,
...
+    for (size_t i = 0; i < test_buffer_ne; i++) {
+        result_host_memory[i] = 0.0f; // Ensure we know the original contents
+    }
+
+    // Allocate an operation on-device destination buffer
+    vk_buffer device_dst_buffer = ggml_vk_create_buffer_device(device, test_buffer_size);
+
+    // Allocate an external memory buffer on device, importing the host memory
+    vk_buffer external_source_buffer = ggml_vk_buffer_from_host_ptr(device, input_host_memory.get(), test_buffer_size);
```
0cc4m, line 12146: Same thing as above, to make this easier to understand please check if this actually succeeded and print an error if not. In my test it returned a nullptr and this just crashed the copy_async functions below, so I needed a debugger to figure out which function failed.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -12095,6 +12096,89 @@ static void ggml_vk_test_dequant_matmul(ggml_backend_vk_context * ctx, size_t m,
     free(d);
     free(d_chk);
 }
+
+static void ggml_vk_test_single_device_buffer_from_host_ptr(vk_device& device) {
```
0cc4m, line 12100:
```suggestion
static void ggml_vk_test_device_buffer_from_host_ptr(vk_device& device) {
```

ggml/src/ggml-vulkan/ggml-vulkan.cpp
0cc4m: Feel free to remove it, I must have accidentally merged it. You can also move your test to the top, provided it fails non-fatally. The matmul tests take a while to run, and I haven't really needed them anymore in a long time.

0cc4m (commented):
Sorry about the delay. I ran the test and it works. It showed me that buffer_from_host_ptr does not work on my Intel iGPU, at least not in this way. I went through the code and added some comments. Once those are resolved it's good to merge.

PR #19268, opened 2026-02-02
ci: add test-backend-ops test for CPU

am17an:
add `test-backend-ops -b CPU` path for checking optimised implementations against reference ones. Will be implemented for FA in #19209


am17an:
Looks like it ran fine https://github.com/ggml-org/llama.cpp/actions/runs/21591212726/job/62211565628?pr=19268#logs

ggerganov (approved):

PR #19270, opened 2026-02-02
common : add common_speculative_is_compat()

ggerganov:
fix #19267

Memory modules that do not support removing the last tokens from the context (such as recurrent modules) cannot perform speculative decoding. Add new `common_speculative_is_compat()` to query this functionality and use it in `llama-server` to disable speculative decoding for those contexts.

### DIFF
--- a/common/speculative.cpp
+++ b/common/speculative.cpp
@@ -797,6 +797,42 @@ enum common_speculative_type common_speculative_type_from_name(const std::string
     return it->second;
 }
 
+bool common_speculative_is_compat(llama_context * ctx_tgt) {
+    auto * mem = llama_get_memory(ctx_tgt);
+    if (mem == nullptr) {
+        return false;
+    }
+
+    bool res = true;
+
+    llama_memory_clear(mem, true);
+
+    // eval 2 tokens to check if the context is compatible
+    std::vector<llama_token> tmp;
+    tmp.push_back(0);
+    tmp.push_back(0);
+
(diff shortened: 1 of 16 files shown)


danbev (approved):

ngxson (commented):
Just wondering if we can do this inside `common_speculative_init` instead.

For example, `common_speculative_init` can try to evaluate 2 tokens, then remove the first one. If `llama_memory_seq_rm` returns error, then we throw an error saying the model is not compatible.

Btw, I think it's better to throw an error and exit, rather than a warning.

coder543:
I just ran into #19267, and it would be cool if there were a way to make this compatible rather than just disabling it, but disabling it is better than crashing. With Qwen3-Coder-Next, ngram-mod could provide large speedups during coding workflows.

ggerganov:
@ngxson Implemented this idea in a new `common_speculative_is_compat()` helper function.

> Btw, I think it's better to throw an error and exit, rather than a warning.

Do you have something specific in mind? In my server config, I want to set a default ngram-based spec decoding and have it applied for all routed models. When a routed model does not support it, it still continues to work. So I think a warning is better.

ggerganov:
@ngxson Gentle ping

ngxson (approved):
Yeah sorry I missed the notif. LGTM!

Just wondering if we should also do the same check for draft model.

ggerganov:
Yes, I think we can do that. Will follow up in next PR.

PR #19271, opened 2026-02-02
Step3.5 support

pwilkin:
Since the model creators put up a llama.cpp fork, but they did it by bare directory upload and the llama.cpp version is quite far outdated, I attempted to pull a clean PR from their code. Not guaranteeing it's all correct, but putting it up since there's probably going to be lots of interest in the new model.


ggerganov:
I was just downloading it and will give it a try soon.

src/models/step35-iswa.cpp
```diff
@@ -0,0 +1,178 @@
...
+                cb(gate, "attn_gate_sigmoid", il);
+
+                // reshape + broadcast to [n_embd_head_v, n_head_l, n_tokens]
+                ggml_tensor * attn_3d = ggml_reshape_3d(ctx0, attn_out, n_embd_head_v, n_head_l, n_tokens);
+                ggml_tensor * gate_3d = ggml_reshape_3d(ctx0, gate,       1,          n_head_l, n_tokens);
+                gate_3d = ggml_repeat(ctx0, gate_3d, attn_3d);
+                cb(gate_3d, "attn_gate_bcast", il);
+
+                attn_3d = ggml_mul(ctx0, attn_3d, gate_3d);
+                cb(attn_3d, "attn_gated_3d", il);
```
ggerganov: This looks a bit redundant - think the repeat can be removed.

pwilkin:
@ggerganov Any idea what the deal is with this?

/usr/bin/ld: ../../bin/libllama.so.0.0.1: undefined reference to `bool llama_model_loader::get_key_or_arr<float, 512ul>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::array<float, 512ul>&, unsigned int, bool)'

Getting it locally too, not just on CI.

src/llama-model.cpp
```diff
@@ -2450,6 +2450,47 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+
+                hparams.has_rope_freq_base_per_layer = ml.get_key_or_arr(
+                    format("%s.rope.freq_base_per_layer", ml.get_arch_name().c_str()),
+                    hparams.rope_freq_base_per_layer,
+                    hparams.n_layer,
+                    false
+                );
+
+                type = LLM_TYPE_UNKNOWN;
+            } break;
```
ggerganov, line 2493: Maybe one of these does not compile correctly. On mac it builds OK btw.

pwilkin: Got it.

src/models/step35-iswa.cpp
pwilkin: Yeah should be auto-broadcast.

forforever73:
@pwilkin  Thanks a lot for putting this together so quickly, and for taking the time to extract the changes into a clean PR — we really appreciate the interest and effort around step3.5. We’re the authors of the step3.5 model, and we already have an official llama.cpp integration prepared, which we’re planning to submit very shortly (likely tomorrow).
﻿
For context, the earlier bare directory upload was on us — that was a mistake on our side, and we should have prepared a proper PR from the start. Thanks for jumping in and helping turn that into a PR.

Since the upcoming PR will be the reference implementation we’ll be maintaining going forward, we’d prefer to have that land as the primary integration. For that reason, we were hoping this PR could be held off from merging for now.
﻿
Thanks again for the quick turnaround and for helping move support for the new model forward.

src/llama-model.cpp
```diff
@@ -2450,6 +2450,47 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+                ml.get_key_or_arr(LLM_KV_SWIGLU_LIMITS,        hparams.swiglu_limits,        hparams.n_layer);
+                ml.get_key_or_arr(LLM_KV_SWIGLU_LIMITS_SHARED, hparams.swiglu_limits_shared, hparams.n_layer);
+
+                // Optional: Step35-only gating for applying rope scaling (HF: yarn_only_types).
+                // Default is 3 (apply on all layers) if the key is absent.
+                ml.get_key(
+                    format("%s.rope.scaling.apply_mask", ml.get_arch_name().c_str()),
+                    hparams.rope_scaling_apply_mask,
+                    false
+                );
```
ngxson, line 2483: I think you will need something like this (`const char*` instead of `std::string`):

```
template bool llama_model_loader::get_key_or_arr<float, 512>(const char *, std::array<float, 512> & result, uint32_t n, bool required);
```

But I think it's better to just put this as a `LLM_KV_*`, like all other metadata keys

convert_hf_to_gguf.py
```diff
@@ -7693,6 +7693,187 @@ def prepare_tensors(self):
...
+        # bit1 -> apply on sliding_attention (SWA layers)
+        yarn_only_types = self.hparams.get("yarn_only_types", None)
+        apply_mask = 0x3  # default: apply on all layers (backwards compatible)
+        if isinstance(yarn_only_types, list):
+            apply_mask = 0
+            if "full_attention" in yarn_only_types:
+                apply_mask |= 0x1
+            if "sliding_attention" in yarn_only_types:
+                apply_mask |= 0x2
+        self.gguf_writer.add_uint32(f"{self.gguf_writer.arch}.rope.scaling.apply_mask", int(apply_mask))
```
ngxson, line 7770: Better to add this as a `self.gguf_writer.add_*` function

src/llama-model.cpp
pwilkin: Already did it :>

convert_hf_to_gguf.py
```diff
@@ -7693,6 +7693,187 @@ def prepare_tensors(self):
                 raise ValueError(f"Unprocessed experts: {experts}")
 
 
+@ModelBase.register("Step35ForCausalLM")
```
ggerganov, line 7696: Should this be:


```suggestion
@ModelBase.register("Step3p5ForCausalLM")
```

pwilkin:
Closing as duplicate of #19283

PR #19282, opened 2026-02-03
[WIP] ggml-hexagon: convert f32 to f16 - fa opt part3

chraac:
### Key changes
*   **Optimization:** Implemented in-place F32-to-F16 conversion for row data, eliminating duplicate conversions inside the inner loop.
*   **New Compute Primitives:** Added fused multiply-add (MAD) support (`hvx_mad_f32_f16_aa_rx2`) for FP16 inputs with separate scaling factors.

### Implementation Details
*   **In-place F32 to F16 Conversion:** The row conversion logic was hoisted out of the inner loop.
*   **`hvx_mad_f32_f16_aa_rx2`:** A /new HVX intrinsic wrapper allowing efficient accumulation of two FP16 vectors.

### Performance
TODO: Add benchmarks comparisons (e.g., token/s improvements on target hardware).


ggml/src/ggml-hexagon/htp/flash-attn-ops.c
```diff
@@ -414,8 +354,12 @@ static void flash_attn_ext_f16_thread(struct htp_ops_context * octx, int ith, in
             }
         }
 
-        const uint8_t * q_ptr_vtcm = dma_queue_pop(dma).dst;
+        uint8_t * q_ptr_vtcm = dma_queue_pop(dma).dst;
+        if (is_q_fp32) {
+            hvx_copy_f16_f32_aa(q_ptr_vtcm, q_ptr_vtcm, DK);  // inplace convert f32 to f16
```
chraac, line 359: Pre-conversion to F16: Converted the row to F16 upfront to avoid repeated on-the-fly conversion in the code below.

max-krasnyansky:
Looks good to me. I'm not seeing any perf improvements but the changes make sense. 
That in-place f32->f16 conversion is clever and makes sense even if perf is about same.

I'm seeing a bunch of new failures in `test-backend-ops` though. All the smaller tensor sizes fail.
The large ones are all good. Let's fix that and it should be good to merge.

btw While we're at it we can also use f16 when we apply the slope

```
                if (mask) {
                    const __fp16 * mp = m_base + ic;
                    HVX_Vector m_vals_f16 = *(const HVX_UVector *) mp;

                    HVX_Vector one_f16 = Q6_Vh_vsplat_R(0x3c00);
                    HVX_VectorPair m_vals_f32_pair = Q6_Wqf32_vmpy_VhfVhf(Q6_Vh_vshuff_Vh(m_vals_f16), one_f16);

                    HVX_Vector m_vals_f32 = Q6_Vsf_equals_Vqf32(Q6_V_lo_W(m_vals_f32_pair));

                    HVX_Vector slope_vec = hvx_vec_splat_f32(slope);
                    HVX_Vector add_val = Q6_Vqf32_vmpy_VsfVsf(m_vals_f32, slope_vec);
                    scores = Q6_Vqf32_vadd_VsfVsf(scores, Q6_Vsf_equals_Vqf32(add_val));
                    scores = Q6_Vsf_equals_Vqf32(scores);
                }
```

Instead of splatting 1.0 to convert into fp32 just splat the slope value with hvx_vec_splat_f16(slope) and multiply with that.  That'd reduce the number of multiplies we do.

chraac:
> I'm seeing a bunch of new failures in test-backend-ops though. All the smaller tensor sizes fail.
The large ones are all good. Let's fix that and it should be good to merge.

hmm, seems to be a device-specific issue, as the FA tests are all passing on my 8gen2 device, gonna have a look on the failures later

chraac:
> Instead of splatting 1.0 to convert into fp32 just splat the slope value with hvx_vec_splat_f16(slope) and multiply with that.

Interesting, slightly concerned that casting the slope to f16 might affect precision a bit, but lemme have a try

max-krasnyansky:
I tracked down `test-backend-ops` failures.
There were due to incomplete nloe handling. We need to mask out both inputs otherwise nan/-nan * 0 is still nan.
So depending on what we have in VTCM or DDR buffers we might get nans. 

I've been doing updates to all the Ops to use local context structs and to precompute stuff as much as possible.
So I did a bunch of updates on top of your commits and started bigger PR #19583
I think we can close this one.

PR #19288, opened 2026-02-03
ggml-cpu: use LUT for converting e8->f32 scales on x86

am17an:
`perf` showed the e8m0->f32 function as a bottleneck. Use a LUT instead. Tested only on x86

| Model                 | Test   |   t/s topk-cuda-refactor |   t/s mxfp4-cpu-scale |   Speedup |
|:----------------------|:-------|-------------------------:|----------------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp1024 |                   237.74 |                257.53 |      1.08 |
| gpt-oss 20B MXFP4 MoE | pp2048 |                   228.16 |                246.40 |      1.08 |
| gpt-oss 20B MXFP4 MoE | pp4096 |                   211.92 |                227.59 |      1.07 |
| gpt-oss 20B MXFP4 MoE | pp8192 |                   185.53 |                197.05 |      1.06 |

### DIFF
--- a/ggml/src/ggml-cpu/arch/x86/quants.c
+++ b/ggml/src/ggml-cpu/arch/x86/quants.c
@@ -268,9 +268,9 @@ static inline __m256 quad_fp16_delta_float(const float x0, const float y0, const
                            _mm_set1_ps(GGML_CPU_FP16_TO_FP32(x0) * GGML_CPU_FP16_TO_FP32(y0)));
 }
 
-static inline __m256 quad_mx_delta_float(const int8_t x0, const float y0, const int8_t x1, const float y1) {
-    return _mm256_set_m128(_mm_set1_ps(GGML_E8M0_TO_FP32_HALF(x1) * GGML_CPU_FP16_TO_FP32(y1)),
-                           _mm_set1_ps(GGML_E8M0_TO_FP32_HALF(x0) * GGML_CPU_FP16_TO_FP32(y0)));
+static inline __m256 quad_mx_delta_float(const uint8_t x0, const float y0, const uint8_t x1, const float y1) {
+    return _mm256_set_m128(_mm_set1_ps(ggml_table_f32_e8m0_half[x1] * GGML_CPU_FP16_TO_FP32(y1)),
+                           _mm_set1_ps(ggml_table_f32_e8m0_half[x0] * GGML_CPU_FP16_TO_FP32(y0)));
 }
 #endif
 #elif defined(__SSSE3__)
@@ -782,6 +782,7 @@ void ggml_vec_dot_mxfp4_q8_0(int n, float * GGML_RESTRICT s, size_t bs, const vo
 
     __m256 accum1 = _mm256_setzero_ps();
     __m256 accum2 = _mm256_setzero_ps();
+
     for (; ib + 1 < nb; ib += 2) {
         const __m128i q4bits_1 = _mm_loadu_si128((const __m128i*)x[ib + 0].qs);
         const __m128i q4bits_2 = _mm_loadu_si128((const __m128i*)x[ib + 1].qs);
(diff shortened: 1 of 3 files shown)


ggerganov:
On M2 Ultra it does not make a difference:

| Model                 | Test   |   t/s master |   t/s pr/19288 |   Speedup |
|:----------------------|:-------|-------------:|---------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512  |       182.14 |         182.79 |      1.00 |
| gpt-oss 20B MXFP4 MoE | pp1024 |       178.52 |         178.44 |      1.00 |
| gpt-oss 20B MXFP4 MoE | pp2048 |       170.37 |         170.38 |      1.00 |

ggml/src/ggml-cpu/arch/x86/quants.c
```diff
@@ -268,9 +268,9 @@ static inline __m256 quad_fp16_delta_float(const float x0, const float y0, const
                            _mm_set1_ps(GGML_CPU_FP16_TO_FP32(x0) * GGML_CPU_FP16_TO_FP32(y0)));
 }
 
-static inline __m256 quad_mx_delta_float(const int8_t x0, const float y0, const int8_t x1, const float y1) {
-    return _mm256_set_m128(_mm_set1_ps(GGML_E8M0_TO_FP32_HALF(x1) * GGML_CPU_FP16_TO_FP32(y1)),
-                           _mm_set1_ps(GGML_E8M0_TO_FP32_HALF(x0) * GGML_CPU_FP16_TO_FP32(y0)));
+static inline __m256 quad_mx_delta_float(const uint8_t x0, const float y0, const uint8_t x1, const float y1) {
+    return _mm256_set_m128(_mm_set1_ps(ggml_table_f32_e8m0_half[x1] * GGML_CPU_FP16_TO_FP32(y1)),
+                           _mm_set1_ps(ggml_table_f32_e8m0_half[x0] * GGML_CPU_FP16_TO_FP32(y0)));
```
ggerganov: Instead of using the table explicitly, define 2 macros:

```c
#define GGML_COMPUTE_E8M0_TO_FP32_HALF(...) ...
#define GGML_E8M0_TO_FP32_HALF(x) ggml_table_f32_e8m0_half[(x)]
```

This way the existing code keeps using `GGML_E8M0_TO_FP32_HALF` which redirects to the table lookup.

am17an:
> On M2 Ultra it does not make a difference:

I only made the change for x86, M2 ultra will be ARM?

ggml/src/ggml-cpu/arch/x86/quants.c
am17an: I didn't do this because other parts of the code use this function also, not sure what the performance will be for them

ggerganov: I see. You can still use the same pattern as we do here:

https://github.com/ggml-org/llama.cpp/blob/e9a859db3cbad0d3dd10deb0527ddfcecce5e78d/ggml/src/ggml-cpu/simd-mappings.h#L115-L134

And forward the `GGML_E8M0_TO_FP32_HALF` to lookup only if `__AVX__` is defined for example.

am17an:
@ggerganov I have a M4 macbook air. When I run llama-bench on vanilla settings I get a `Metal,BLAS` backend. Is there a way to force ARM? Or maybe I am not understanding something w.r.t unified memory. 
Also one more question: is it there a way to skip the depth computation when doing llama-bench. I find it takes a long time for the context to fill up, I was wondering if there is a way to fill the kv-cache randomly with a specified depth

ggerganov:
To disable GPU and AMX acceleration, build with:

```bash
cmake -DGGML_METAL=OFF -DGGML_BLAS=OFF ...
```

For the `compare-commit.sh` script, use something like:

```bash
CMAKE_OPTS="-DGGML_METAL=OFF -DGGML_BLAS=OFF" scripts/compare-commits.sh master pr/19288 llama-bench -m ~/models/gpt-oss-20b/ggml-model-mxfp4.gguf -fa 1 -ub 2048 -p 0 -n 0 -r 3 -mmp 1 -t 16 -n 32,32,32
```

> Also one more question: is it there a way to skip the depth computation when doing llama-bench.

The `-d` argument is the best we have atm. It computes the context up to `d` once and then can run multiple `-p` and `-n` with that pre-computed context. It still needs to be computed one time.

Filling with random data is not going to work because it will not activate the correct experts for MoE models.

ggerganov (approved):

PR #19075, opened 2026-01-24
Vulkan Flash Attention Coopmat1 Refactor

0cc4m:
I finally had the time to go through Jeff's Flash Attention shaders in detail and used the chance to refactor the Coopmat1 for AMD. It started out as an attempt to use Coopmats for the Softmax * V matrix multiplication as well and then escalated into a refactor of the whole shader structure.

It now uses coopmats for the Softmax result * V matrix multiplication, and I vectorized some variables, changed how shared memory is used, load K and V directly from global memory if possible, otherwise streamed through a shared memory cache.

Tests are passing. Performance is up significantly on AMD RX 8060S (Strix Halo). Draft because there is a regression on Nvidia. Let me know if you see anything obvious @jeffbolznv. More tuning is likely required.

AMD 8060S:
| model                 | size      | params  | ngl | fa | test          | t/s (ROCm)      | t/s (before)    | t/s (after)     | diff   |
|-----------------------|-----------|---------|-----|----|---------------|-----------------|-----------------|-----------------|--------|
| llama 8B Q4_K - Small | 4.36 GiB  | 8.03 B  | 99  | 1  | pp512         | 1087.56 ± 41.81 | 1004.49 ± 44.19 | 1020.70 ± 46.87 | +1.6%  |
| llama 8B Q4_K - Small | 4.36 GiB  | 8.03 B  | 99  | 1  | tg128         | 40.38 ± 0.28    | 43.02 ± 0.25    | 43.91 ± 0.44    | +2.1%  |
| llama 8B Q4_K - Small | 4.36 GiB  | 8.03 B  | 99  | 1  | pp512 @ d8192 | 604.32 ± 4.11   | 418.94 ± 1.64   | 556.71 ± 3.25   | +32.9% |
| llama 8B Q4_K - Small | 4.36 GiB  | 8.03 B  | 99  | 1  | tg128 @ d8192 | 33.96 ± 0.04    | 33.34 ± 0.08    | 35.88 ± 0.14    | +7.6%  |
| gpt-oss 20B MXFP4 MoE | 11.27 GiB | 20.91 B | 99  | 1  | pp512         | 1037.80 ± 33.87 | 1316.13 ± 23.00 | 1285.66 ± 30.56 | -2.3%  |
| gpt-oss 20B MXFP4 MoE | 11.27 GiB | 20.91 B | 99  | 1  | tg128         | 70.82 ± 0.06    | 76.86 ± 0.12    | 76.17 ± 1.36    | -0.9%  |
[...]

PR #19290, opened 2026-02-03
vulkan: disable coopmat1 flash attention on Nvidia Turing

0cc4m:
See https://github.com/ggml-org/llama.cpp/pull/19075#issuecomment-3820716090

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -254,6 +254,7 @@ enum vk_device_architecture {
     AMD_RDNA3,
     INTEL_XE2,
     NVIDIA_PRE_TURING,
+    NVIDIA_TURING,
 };
 
 static vk_device_architecture get_device_architecture(const vk::PhysicalDevice& device) {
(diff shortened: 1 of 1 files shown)


ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -336,18 +337,34 @@ static vk_device_architecture get_device_architecture(const vk::PhysicalDevice&
...
+        if (sm_builtins) {
+            vk::PhysicalDeviceProperties2 props2;
+            vk::PhysicalDeviceShaderSMBuiltinsPropertiesNV sm_props;
+
+            props2.pNext = &sm_props;
+
+            device.getProperties2(&props2);
+
+            // Turing has 32, following architectures have 48
+            if (sm_props.shaderWarpsPerSM == 32) {
```
jeffbolznv, line 364: I'm not confident that this will be true forever, but maybe it's OK for now. There aren't a whole lot of feature differences between Turing and Ampere... one other one I can find is that primitiveFragmentShadingRateWithMultipleViewports is supported on Ampere but not Turing.

jeffbolznv (approved):

ggml/src/ggml-vulkan/ggml-vulkan.cpp
0cc4m: It was the first difference I found. But we can hopefully remove this workaround again when the driver is fixed, before a following architecture breaks this assumption.

PR #19291, opened 2026-02-03
ci : add sanitizer runs for server

ggerganov:
Reenable the server sanitizer builds + runs. The thread sanitizer is quite slow, so remains disabled for now.

https://github.com/ggerganov/tmp2/actions/runs/21629674042

### DIFF
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -32,4 +32,27 @@ function(llama_add_compile_flags)
             set(CXX_FLAGS "" PARENT_SCOPE)
         endif()
     endif()
+
+    if (NOT MSVC)
+        if (LLAMA_SANITIZE_THREAD)
+            message(STATUS "Using -fsanitize=thread")
+
+            add_compile_options(-fsanitize=thread)
+            link_libraries     (-fsanitize=thread)
+        endif()
+
+        if (LLAMA_SANITIZE_ADDRESS)
+            message(STATUS "Using -fsanitize=address")
+
+            add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
+            link_libraries     (-fsanitize=address)
+        endif()
+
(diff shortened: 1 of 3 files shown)


CISC (approved):

CISC:
Oh, interesting:
https://github.com/ggml-org/llama.cpp/actions/runs/21630912707/job/62343121265
```console
2026-02-03T17:10:51.0689076Z The following tests FAILED:
2026-02-03T17:10:51.0689668Z 	 32 - test-thread-safety (Failed)                       main
2026-02-03T17:10:51.0690260Z 	 41 - test-barrier (Failed)                             main
2026-02-03T17:10:51.0690796Z Errors while running CTest
```

cmake/common.cmake
ngxson, line 1: I got false-positive ASAN errors after this change (triggered by `httplib.h`), wondering if the sanitizer flag is still propagated correctly

CISC: Something strange is definitely going on.

CISC:
I wonder why these runners get randomly terminated, searching for it seem to suggest this is either because of high CPU or RAM usage, though not sure how this would be higher than on other runners.

ngxson:
If it's terminated due to high RAM, I think it will have a specific exit code (OOM kill code, usually 137)

I don't think it's terminated by high CPU though, I suspect that the compilation step can use 100% consistently, but the same can't be said with the server test. As we stop and start server instance in-between tests, in theory there will be a gap in CPU usage

CISC:
> I don't think it's terminated by high CPU though, I suspect that the compilation step can use 100% consistently, but the same can't be said with the server test. As we stop and start server instance in-between tests, in theory there will be a gap in CPU usage

It happens during compilation step, so most likely #19411 is the correct solution.

Issue #19128, opened 2026-01-27
Eval bug: GLM-4.7-Flash gibberish output with FA = on


engrtipusultan:
### Name and Version

```
./llama-server --version
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = AMD Radeon Graphics (RADV RENOIR) (radv) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 64 | shared memory: 65536 | int dot: 0 | matrix cores: none
version: 7842 (7afdfc9b8)
built with GNU 13.3.0 for Linux x86_64
```

### Operating systems

Linux

### GGML backends

Vulkan, BLAS

### Hardware

[AMD Ryzen 7 5825U](https://www.gmktec.com/products/amd-ryzen-7-5825u-mini-pc-nucbox-m5-plus)
RAM: 64Gb DDR4
gfx_target_version gfx90c
AOCL: aocl-linux-gcc-5.2.0
VULKAN SDK: LunarG 1.4.335.0
VULKAN Driver: mesa-vulkan-drivers: amd64 25.3.3 kisak1n amd64
OpenGL Driver Version Mesa 25.3.3 - kisak-mesa PPA

Make options used:

```
cmake -S "$SOURCE_DIR" -B "$BUILD_DIR" -DGGML_NATIVE=ON -DSD_VULKAN=ON -DGGML_BLAS=ON -DGGML_BLAS_VENDOR=AOCL -DBLAS_INCLUDE_DIRS=/opt/AMD/aocl/aocl-linux-gcc-5.2.0/gcc/include -DCMAKE_PREFIX_PATH=/opt/AMD/aocl/aocl-linux-gcc-5.2.0/gcc
cmake --build "$BUILD_DIR" --config Release --clean-first --parallel $(nproc)
```

### Models

GLM-4.7-Flash-Q8_0 from Unsloth updated

### Problem description & steps to reproduce

When GLM-4.7-Flash is run with FA = on, then on second or third prompt it starts producing gibberish output. 

For example:
prompt 1
Give me quick sort in python.
prompt 2 
What are possible improved versions ?
prompt 3
provide me code for IntroSort also share its time and space complexity.
FA = off coherent output.

FA = on garbage output. (happens when context reaches few thousand tokens) 

<img width="594" height="856" alt="Image" src="https://github.com/user-attachments/assets/3670b4ab-c5ed-4960-a51b-562932b2edfa" />

### First Bad Commit

[...]

PR #19296, opened 2026-02-03
tests : add non-cont, inplace rope tests

ggerganov:
ref https://github.com/ggml-org/llama.cpp/pull/18986#issuecomment-3841942982
ref https://github.com/ggml-org/llama.cpp/issues/19128#issuecomment-3807441909
ref https://github.com/ggml-org/llama.cpp/issues/19292

This type of rope is needed during rope shift:

https://github.com/ggml-org/llama.cpp/blob/cf36c21b221aac7c7ce6dca9def898311415e56b/src/llama-kv-cache.cpp#L1626-L1635

We are updating the KV cache data, so the op is inplace. Since the changes in #18986 the tensor is now also non-contiguous.

Based on the discussion in https://github.com/ggml-org/llama.cpp/pull/18986#issuecomment-3841942982, it appears that the Vulkan backend currently does not support this case. cc @jeffbolznv @0cc4m

### DIFF
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -8032,6 +8032,7 @@ static std::vector<std::unique_ptr<test_case>> make_test_cases_eval() {
         for (int mode : {GGML_ROPE_TYPE_NORMAL, GGML_ROPE_TYPE_NEOX, GGML_ROPE_TYPE_MROPE, GGML_ROPE_TYPE_IMROPE, GGML_ROPE_TYPE_VISION}) {
             for (bool ff : {false, true}) {
                 test_cases.emplace_back(new test_rope(type, {128,  32, 2, 1}, 128, mode, 512, 1.4245f, 0.7465f, 1.4245f, ff, 0, true, true));
+                test_cases.emplace_back(new test_rope(type, {128,  32, 2, 1}, 128, mode, 512, 1.4245f, 0.7465f, 1.4245f, ff, 1, true, true));
             }
         }
     }
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -8032,6 +8032,7 @@ static std::vector<std::unique_ptr<test_case>> make_test_cases_eval() {
         for (int mode : {GGML_ROPE_TYPE_NORMAL, GGML_ROPE_TYPE_NEOX, GGML_ROPE_TYPE_MROPE, GGML_ROPE_TYPE_IMROPE, GGML_ROPE_TYPE_VISION}) {
             for (bool ff : {false, true}) {
                 test_cases.emplace_back(new test_rope(type, {128,  32, 2, 1}, 128, mode, 512, 1.4245f, 0.7465f, 1.4245f, ff, 0, true, true));
+                test_cases.emplace_back(new test_rope(type, {128,  32, 2, 1}, 128, mode, 512, 1.4245f, 0.7465f, 1.4245f, ff, 1, true, true));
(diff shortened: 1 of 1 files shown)


jeffbolznv:
Yeah, the new tests are failing in the vulkan backend. I'll look into it.

LostRuins:
Thank you. If it works I will close https://github.com/ggml-org/llama.cpp/issues/19292

tests/test-backend-ops.cpp
```diff
@@ -8032,6 +8032,7 @@ static std::vector<std::unique_ptr<test_case>> make_test_cases_eval() {
         for (int mode : {GGML_ROPE_TYPE_NORMAL, GGML_ROPE_TYPE_NEOX, GGML_ROPE_TYPE_MROPE, GGML_ROPE_TYPE_IMROPE, GGML_ROPE_TYPE_VISION}) {
             for (bool ff : {false, true}) {
                 test_cases.emplace_back(new test_rope(type, {128,  32, 2, 1}, 128, mode, 512, 1.4245f, 0.7465f, 1.4245f, ff, 0, true, true));
+                test_cases.emplace_back(new test_rope(type, {128,  32, 2, 1}, 128, mode, 512, 1.4245f, 0.7465f, 1.4245f, ff, 1, true, true));
```
jeffbolznv: Might as well also exercise dim 3.


```suggestion
                test_cases.emplace_back(new test_rope(type, {128,  32, 2, 3}, 128, mode, 512, 1.4245f, 0.7465f, 1.4245f, ff, 1, true, true));
```

ggerganov:
Huh, adding the dim 3 somehow hided the CUDA failures from earlier:

https://github.com/ggml-org/llama.cpp/actions/runs/21636838841/job/62364670686#step:3:48919

I guess I have to add both tests for `ne[3] == 1` and `ne[3] > 1`

LostRuins:
Hi all, I can confirm that Jeff's patch fixes https://github.com/ggml-org/llama.cpp/issues/19292 for me on Vulkan! 

Works fine and fixes the incoherence-after-shifting regression on GLM-4-32B-0414 and GLM-4.7-Flash.

Much appreciated to all.

However is anyone able to check this test case is good for CUDA too? I have heard that a similar issue affected a CUDA RTX 5xxx user on GLM-4.7-Flash, would be nice if someone could confirm the test passes on CUDA.

ggerganov:
It's very likely broken with CUDA in a similar way - see the test failure from earlier: https://github.com/ggml-org/llama.cpp/actions/runs/21636838841/job/62364670686#step:3:48919

Once we stabilize the tests, we'll notify CUDA maintainers to take a look.

LostRuins:
Thanks! Once there is a fix for cuda i'll let the cuda user try it out again.

ggerganov:
@JohannesGaessler Note that some of the newly added rope tests are currently failing with CUDA: https://github.com/ggml-org/llama.cpp/actions/runs/21664340312/job/62455946368

ORippler:
> @JohannesGaessler Note that some of the newly added rope tests are currently failing with CUDA: https://github.com/ggml-org/llama.cpp/actions/runs/21664340312/job/62455946368

Should be fixed by https://github.com/ggml-org/llama.cpp/pull/19338

pwilkin:
The new tests are failing with Vulkan as well @jeffbolznv @0cc4m

jeffbolznv:
The vulkan fix is at https://github.com/ggml-org/llama.cpp/pull/19299

PR #19299, opened 2026-02-03
vulkan: fix non-contig rope

jeffbolznv:
For #19296.


LostRuins (approved):
Can confirm it fixes https://github.com/ggml-org/llama.cpp/issues/19292 for me! 

Works fine and fixes the incoherence-after-shifting regression on GLM-4-32B-0414 and GLM-4.7-Flash.

Much appreciated.

0cc4m (approved):

PR #19306, opened 2026-02-03
sycl: add F16 support for GGML_OP_CEIL

NechamaKrashinski:
Adds F16 support for the `GGML_OP_CEIL` operation in the SYCL backend. 

The implementation matches CPU semantics for element-wise ceil operations and enables both F32 and F16 types to pass functional tests.

Tests:
* ./build-sycl/bin/test-backend-ops test -b SYCL0 -o CEIL (14/14 passed)

Part of #14909


NeoZhangJianyu (approved):
It's great!

Thank you!

PR #19310, opened 2026-02-04
ggml-webgpu: JIT compile binary operators and handle binding overlaps

abhijitramesh:
This PR incorporates the following changes for the webGPU backend:

- Binary operators are now JIT compiled.
- Generates shaders to handle overlapping regions of buffers to multiple bind group entries for binary operators


ggml/src/ggml-webgpu/wgsl-shaders/binary.wgsl
```diff
@@ -0,0 +1,166 @@
...
+#define DataType f16
+#endif
+
+@group(0) @binding(0)
+var<storage, read_write> src0: array<DataType>;
+
+#if defined(INPLACE) && defined(OVERLAP)
+
+fn update(dst_i: u32, src0_i: u32, src1_i: u32) {
+#ifdef OP_ADD
```
reeselevine: I think there's a way to avoid the duplication of all these ifdefs for different versions of inplace/overlap. Not sure if what I'm thinking is best macro practice though... something like one statement of 

```
DST[dst_i] = SRC_A[src0_i] OP SRC_B[src1_i]
```

where `DST`, `SRC_A`, `SRC_B`, and `OP` are all defined based on the type.

Also, are all the combinations of src0/src1/dst necessary right now?

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -465,4 +465,83 @@ inline ggml_webgpu_processed_shader ggml_webgpu_preprocess_unary_shader(
     return result;
 }
 
+/** Binary **/
+
+struct ggml_webgpu_binary_pipeline_key {
+    int  type;
+    int  op;
+    bool inplace;
+    bool overlap;
```
reeselevine, line 474: can this name be more descriptive, since we have multiple types of overlap?

ggml/src/ggml-webgpu/wgsl-shaders/binary.wgsl
abhijitramesh: After running a few more models, I realized there are really only two kinds of overlaps. 
1. src0==dst
2. src1==dst

We are calling the first case INPLACE, and I have now renamed the second case to OVERLAP. With these considerations in mind. I have refactored the code to only detect and generate shaders for these three cases. 
1. Inplace 
2. overlap 
3. default (where all buffers are separate)

I have removed all the other dead code. Also, modularized the ops based on your suggestion above.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
abhijitramesh: Now, overlap means src1==dst, and inplace means src0==dst. The special case of overlap was redundant and hence removed.

reeselevine (approved):

PR #19315, opened 2026-02-04
[WebGPU] Plug memory leaks and free resources on shutdown

nikhilJain17:
This diff destroys `wgpu::Buffer`s and buffer pools on shutdown. It also fixes memory leaks on the heap, where we allocate `backend`, `backend_ctx`, `buffer_ctx`, and `decisions` on the heap but never delete them. These are either explicitly deleted or changed to be smart pointers. 

We implement destructors for our buffer pool structs, `webgpu_context` struct and `webgpu_global_context` struct. Since `webgpu_global_context` is a refcounted smart pointer, it will destruct automatically when all thread contexts have been destroyed. 

Screenshot 2026-02-03 at 3 56 11 PM

We call free on all the buffers we allocate, and we explicitly free our buffer pools and debug/error/staging buffers. 
Also, since we explicitly wait on all our callbacks (except for the DeviceLost callback), we do not have to worry about waiting for callbacks while shutting down.

Screenshot 2026-02-03 at 8 14 21 PM

Memory leak on the heap before. 

Screenshot 2026-02-03 at 8 15 30 PM

No memory leak on the heap after.


ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -194,7 +194,7 @@ inline ggml_webgpu_processed_shader ggml_webgpu_preprocess_flash_attn_shader(
     ggml_webgpu_processed_shader result;
     result.wgsl                                         = preprocessor.preprocess(shader_src, defines);
     result.variant                                      = variant;
-    ggml_webgpu_flash_attn_shader_decisions * decisions = new ggml_webgpu_flash_attn_shader_decisions();
```
nikhilJain17, line 197: I changed this into a `shared_ptr` because this was leaking on the heap since we never deallocated it.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -319,8 +329,20 @@ struct webgpu_global_context_struct {
     wgpu::Buffer debug_host_buf;
     wgpu::Buffer debug_dev_buf;
 #endif
+
+    ~webgpu_global_context_struct() {
+        this->get_tensor_staging_buf.Destroy();
+#ifdef GGML_WEBGPU_DEBUG
+        debug_host_buf.Destroy();
+        debug_dev_buf.Destroy();
```
nikhilJain17: I believe other `wgpu` members, like `Instance`, `Device`, `Queue`, and `Pipeline`, are refcounted and delete automatically when all references are deleted. But `Buffers` need to be explicitly destroyed.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -792,6 +814,9 @@ static void ggml_backend_webgpu_free(ggml_backend_t backend) {
 #if !defined(GGML_WEBGPU_CPU_PROFILE) && !defined(GGML_WEBGPU_GPU_PROFILE)
     GGML_UNUSED(ctx);
 #endif
+
+    delete ctx;
+    delete backend;
```
nikhilJain17, line 816: These are both allocated on the heap and leak if we don't delete them. Maybe we could turn them into `shared_ptr` but I don't know how it would behave once the pointer is passed around in the ggml lifecycle.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
nikhilJain17: Also since `webgpu_global_context` is a `shared_ptr`, its destructor is automatically called once all references to it are deleted.

reeselevine (approved):

reeselevine:
thanks, this looks great!

PR #19317, opened 2026-02-04
cleanup `llama-quantize --help` output

ddh0:
More pleasant formatting for the output of `llama-quantize --help`.

Before this PR:

Screenshot 2026-02-03 at 10 34 37 PM

---

After this PR:

Screenshot 2026-02-03 at 10 50 30 PM (this image was previously wrong, updated it to match the current code)


tools/quantize/quantize.cpp
```diff
@@ -119,27 +119,48 @@ static bool try_parse_ftype(const std::string & ftype_str_in, llama_ftype & ftyp
...
-    printf("  --token-embedding-type ggml_type: use this ggml_type for the token embeddings tensor\n");
-    printf("  --tensor-type TENSOR=TYPE: quantize this tensor to this ggml_type. example: --tensor-type attn_q=q8_0\n");
-    printf("      Advanced option to selectively quantize tensors. May be specified multiple times.\n");
-    printf("  --tensor-type-file tensor_type.txt: list of tensors to quantize to specific ggml_type. example: --tensor-type-file tensor_type_list.txt\n");
-    printf("      Advanced option to selectively quantize a long list of tensors. Format to be tensor_name=ggml_type, separated by spaces/newline.\n");
-    printf("  --prune-layers L0,L1,L2...comma-separated list of layer numbers to prune from the model\n");
-    printf("      Advanced option to remove all tensors from the given layers\n");
-    printf("  --keep-split: will generate quantized model in the same shards as input\n");
+    printf("  --allow-requantize\n");
+    printf("                                      allow requantizing tensors that have already been quantized\n");
```
compilade, line 126: This is a very minor comment, which can be ignored if you want.

This kind of padding might cause unnecessary additional modifications when either adding a longer option (and args) than what exists currently, or when removing options or rewording args.
It's probably *not really* a problem, though.

But if we'd like to fix that anyway, how about having a constant for the padding (e.g. `HELP_PAD` or something), and using 

```c
    printf("%-*s%s\n", HELP_PAD, "--some-option and args", "some description");
    printf("%-*s%s\n", HELP_PAD, "", "rest of the description");
```

which is using the `*` syntax of fields in `printf` to get the padding from a variable. (And the `-` is for right-padding)

And then in the future, when padding is updated, we'll only need to change the value of the padding variable.

There could even be a macro for it (if desired):

```c
#define HELP_PAD 42 // for example
#define PRINT_HELP(option, desc) printf("%-*s%s", HELP_PAD, (option), (desc))
```

Although if using macros, it's also possible to directly inscribe a compile-time known padding inside the `printf` call.

```c
#define HELP_PAD "42" // notice it's a string
#define PRINT_HELP(option, desc) printf("%-" HELP_PAD "s%s", (option), (desc))
```
This uses the concatenation of adjacent literal strings in C, but then `HELP_PAD` (or equivalent) needs to be a literal string.

Keep in mind this is probably not necessary. Manual padding can work fine too. Like I said, feel absolutely free to ignore this suggestion.

Also note that the advantage of not padding is when using very narrow terminals (e.g. when using `llama.cpp` in Termux on Android). Padded line-wrapping can look weird (although it's already padded in the other examples which use the `common/arg.cpp` help printing)

ddh0: Thanks for the feedback - I will leave it as-is for now, but happy to change it to use a constant or a macro if others think that would be more appropriate.

ggerganov (approved):

PR #14200, opened 2025-06-15
gguf-py: Make sentencepiece optional

Ahajha:
Sentencepiece has not had a new release in over a year, and users are running into easily fixable install issues like CMake minimum versions, missing headers, and missing wheels for newer python versions. See recent issues: https://github.com/google/sentencepiece/issues, like this one for example: https://github.com/google/sentencepiece/issues/1113

Sentencepiece is only actually used if users are already using it, so it seems pointless to depend on it directly. This updates the support to raise an error if sentencepiece isn't installed.

PR #5664, opened 2024-02-22
build(nix): Package gguf-py

ditsuke:
## What and why
Currently, the Nix build does not ship with the `gguf-py` lib, rendering scripts like `convert-hf-to-gguf` unusable.

This is an attempt to package `gguf-py` with `buildPythonPackage`. Very barebones at the moment and needs some work to exclude the Nix built `gguf` from the devShells, and some thought into other things like packaging the heavier python dependencies (we don't want them to be included by default, but we need a way to opt-in).

CC: @SomeoneSerge

PR #19319, opened 2026-02-04
gguf-py: Bump sentencepiece version

Ahajha:
There's a new version that's been out for a while that addresses the issues mentioned in https://github.com/ggml-org/llama.cpp/pull/14200. There's a long chain of reasons I would like this change, but the short version is that it allows people who use both `sentencepiece` and `gguf` to take advantage of these fixes. On conda-forge, currently, it locks the version (since there is no notion of optional dependencies).

Regardless, I don't think this should be too controversial.

### DIFF
--- a/gguf-py/pyproject.toml
+++ b/gguf-py/pyproject.toml
@@ -23,7 +23,7 @@ numpy = ">=1.17"
 tqdm = ">=4.27"
 pyyaml = ">=5.1"
 requests = ">=2.25"
-sentencepiece = { version = ">=0.1.98,<=0.2.0", optional = true }
+sentencepiece = { version = ">=0.1.98,<=0.2.1", optional = true }
 PySide6 = { version = "^6.9", python = ">=3.9,<3.14", optional = true }
 
 [tool.poetry.dev-dependencies]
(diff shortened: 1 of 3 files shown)


CISC (approved):
Thank you, could you also update `requirements/requirements-convert_legacy_llama.txt`?

ehoogeveen-medweb:
The range for `sentencepiece` in `pyproject.toml` was updated from
`sentencepiece = ">=0.1.98,<0.2.0"`
to
`sentencepiece = ">=0.1.98,<=0.2.0"`
in https://github.com/ggml-org/llama.cpp/pull/5664

But I wonder if this was intentional or if the intent was simply to allow `0.2.0` (the latest version at the time) and the actual range should have been
`sentencepiece = ">=0.1.98,<0.3.0"`
to align with semantic versioning (where versions <1.0 use the convention `0.major.minor`, so `0.3.0` would be a major release with breaking changes while `0.2.x` are minor releases).

CISC:
@Ahajha gentle ping

Ahajha:
So sorry, I completely missed these. I will update the PR in a bit.

Ahajha:
@CISC Updated, bumped the version and also updated the third file, PTAL

Ahajha:
Also should this be a new patch release for `gguf`?

CISC:
> Also should this be a new patch release for `gguf`?

No need, will make a minor release soonish, there have been a lot of changes.

PR #19321, opened 2026-02-04
qwen3next : fix chunking

ggerganov:
rel #19305

Based on my analysis in https://github.com/ggml-org/llama.cpp/issues/19305#issuecomment-3845730491 I am just restoring the old chunking logic. This seems to resolve the reported issue.

Note that I am blindly restoring the old code, and still haven't understood in details what this logic actually does. So extra look, logprobs tests and validation would be needed before we merge this. Draft for now.

### DIFF
--- a/src/models/qwen3next.cpp
+++ b/src/models/qwen3next.cpp
@@ -197,7 +197,7 @@ std::pair<ggml_tensor *, ggml_tensor *> llm_build_qwen3next::build_delta_net_chu
     ggml_tensor * g_cumsum = ggml_cumsum(ctx0, g);
     cb(g_cumsum, "g_cumsum", il); // shape: (chunk_size, 1, n_chunks, H_v * n_seqs)
 
-    ggml_tensor * gcs_i = g_cumsum; // ggml_reshape_4d(ctx0, g_cumsum, chunk_size, 1, n_chunks, H_v * n_seqs);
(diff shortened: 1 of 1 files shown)


ggerganov:
Superseded by #19324

src/models/qwen3next.cpp
```diff
@@ -268,85 +268,102 @@ std::pair<ggml_tensor *, ggml_tensor *> llm_build_qwen3next::build_delta_net_chu
...
             ggml_reshape_4d(ctx0, kgdmulvnew, kgdmulvnew->ne[0], kgdmulvnew->ne[1], H_v, n_seqs));
     }
 
-    // truncate padded tokens
     ggml_tensor * output_tokens = ggml_view_4d(ctx0, core_attn_out,
             S_v, n_tokens, H_v, n_seqs,
             ggml_row_size(core_attn_out->type, S_v),
             ggml_row_size(core_attn_out->type, S_v * chunk_size * n_chunks),
             ggml_row_size(core_attn_out->type, S_v * chunk_size * n_chunks * H_v), 0);
-    output_tokens = ggml_cont(ctx0, output_tokens);
```
ggerganov, line 342: @ngxson Btw, this `cont` seems redundant still.

Anyway, not very important. I think there is a lot to improve in this graph - will take a look in the next days.

PR #19322, opened 2026-02-04
vendor : add missing llama_add_compile_flags

CISC:
~Hopefully fixes CI~Ensure `httplib` and `boringssl`/`libressl` are built with sanitizer options, see https://github.com/ggml-org/llama.cpp/pull/19291#discussion_r2761613566


ggerganov:
I am not yet 100% sure that this is necessary. The workflows on `master` are passing.

I think the one that are failing such as https://github.com/ggml-org/llama.cpp/actions/runs/21664340188/job/62455945468?pr=19296 are when the PR is created before the new workflow file and Github Actions gets confused in some way.

CISC:
> I am not yet 100% sure that this is necessary. The workflows on `master` are passing.
> 
> I think the one that are failing such as https://github.com/ggml-org/llama.cpp/actions/runs/21664340188/job/62455945468?pr=19296 are when the PR is created before the new workflow file and Github Actions gets confused in some way.

Possibly, though it makes sense to add it here as we are building `httplib` independently.

CISC:
Cool, "works", but `BoringSSL` isn't compatible with `-Werror` I guess:
https://github.com/ggml-org/llama.cpp/actions/runs/21667745579/job/62467477296?pr=19322#step:4:1835

ggerganov (approved):

CISC:
Well, this is annoying, not sure how to fix this...

danbev:
@CISC I was able to reproduce the locally on my M3 and perhaps this will work for x64 as well:
```console
diff --git a/vendor/cpp-httplib/CMakeLists.txt b/vendor/cpp-httplib/CMakeLists.txt
index 3d938d9f3..19cad0800 100644
--- a/vendor/cpp-httplib/CMakeLists.txt
+++ b/vendor/cpp-httplib/CMakeLists.txt
@@ -31,6 +31,16 @@ target_compile_definitions(${TARGET} PRIVATE
 set(OPENSSL_NO_ASM ON CACHE BOOL "Disable OpenSSL ASM code when building BoringSSL or LibreSSL")

 if (LLAMA_BUILD_BORINGSSL)
+
+    if(LLAMA_FATAL_WARNINGS)
+        set(LLAMA_FATAL_WARNINGS_SAVED ${CMAKE_CXX_FLAGS})
+        string(REPLACE "-Werror" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+        string(REPLACE "-Werror" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wunused-parameter")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wunused-parameter")
+    endif()
+
     set(FIPS OFF CACHE BOOL "Enable FIPS (BoringSSL)")

     set(BORINGSSL_GIT "https://boringssl.googlesource.com/boringssl" CACHE STRING "BoringSSL git repository")
@@ -73,6 +83,10 @@ if (LLAMA_BUILD_BORINGSSL)
     set(CPPHTTPLIB_OPENSSL_SUPPORT TRUE)
     target_link_libraries(${TARGET} PUBLIC ssl crypto)

+    if(LLAMA_FATAL_WARNINGS)
+        set(CMAKE_CXX_FLAGS ${LLAMA_FATAL_WARNINGS_SAVED})
+    endif()
+
 elseif (LLAMA_BUILD_LIBRESSL)
     set(LIBRESSL_VERSION "4.2.1" CACHE STRING "LibreSSL version")
```
Setting the string replacements to `-Wno-unused-parameter` allowed the build to pass.

CISC:
> @CISC I was able to reproduce the locally on my M3 and perhaps this will work for x64 as well:

Thanks, normally you aren't supposed to fiddle with `CMAKE_CXX_FLAGS`, but nothing else seems to work. :P

CISC:
> Setting the string replacements to `-Wno-unused-parameter` allowed the build to pass.

So in theory just adding that should make the build pass?

CISC:
@danbev Doesn't seem to work. :(

danbev:
> @danbev Doesn't seem to work. :(

Sorry about wasting  your time. I really thought that would work :( 

I gave this a try on my fork and it seems to work [there](https://github.com/danbev/llama.cpp/actions/runs/21678054875/job/62504042341).

CISC:
> > @danbev Doesn't seem to work. :(
> 
> Sorry about wasting your time. I really thought that would work :(
> 
> I gave this a try on my fork and it seems to work [there](https://github.com/danbev/llama.cpp/actions/runs/21678054875/job/62504042341).

No worries, I see the reason it worked for you though, you didn't add `llama_add_compile_flags()`. :)

I'll experiment some more...

PR #19323, opened 2026-02-04
ci : fix sanitize workflow to enable ggml sanitizers too

ggerganov:
fix https://github.com/ggml-org/llama.cpp/pull/19291#issuecomment-3843594581

We need to sanitize also the ggml when sanitizing the llama.

Success: https://github.com/ggerganov/tmp2/actions/runs/21667569266/job/62466865225


CISC (approved):

CISC:
It looks like the `ggml-ci-mac` runners have stalled btw.

Edit: Hmmm, or they just have a really weird way of processing the queue, there are still 3 unfinished runs from yesterday...

ggerganov:
It's a single Mac Mini atm and it's struggling a bit lately to keep up.

PR #19334, opened 2026-02-04
Add missing ggml.pc.in

WhyNotHugo:
When compiling libggml in standalone mode, installation then fails due to a missing ggml.pc.in which is referenced in CmakeList.

Add the missing file, using llama.pc.in as a reference.

### DIFF
--- a/ggml/ggml.pc.in
+++ b/ggml/ggml.pc.in
@@ -1,0 +1,10 @@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=@CMAKE_INSTALL_PREFIX@
+libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+
+Name: ggml
+Description: Tensor library for machine learning
+Version: @GGML_VERSION@
+Libs: -L${libdir} -lggml -lggml-base
+Cflags: -I${includedir}


ggerganov:
The file is present in the `ggml` repository: https://github.com/ggml-org/ggml

WhyNotHugo:
I'm packaging libggml and llama.cpp (as separate packages) downstream. Is it preferrible to use the version vendored in the repository, or the ggml repository?

ggerganov:
I think the libggml package should be packaged from the ggml repo. The llama.cpp package should package it's own internal libggml from this repo, link to it and not expose it publicly. Would that make sense?

WhyNotHugo:
If we ship a standalone libggml package and llama.cpp ships its own libggml, we'd end up with two versions of `/usr/lib/libggml.so`, and both packages would conflict with each other.

The libggml vendored in this repository is dramatically different from the standalone one? Generally downstream distributions prefer to ship a single version of each library.

WhyNotHugo:
I'll just package the `.so` in `/usr/lib/llama.cpp`. I see that the vendored ggml has a different API from the standalone repo.

Thanks for the guidance!

ggerganov:
> I'll just package the .so in /usr/lib/llama.cpp. I see that the vendored ggml has a different API from the standalone repo.

Yes, that was my idea. Please provide feedback if this does not work out for some reason.

WhyNotHugo:
Seems to work fine. It's not ideal downstream, since the typical preference is to build and ship a single copy of each library, but given that there's divergence in this case, there's no way around that.

PR #19357, opened 2026-02-05
ggml : add int2 quantization and kleidiai sme2 gemm/gemv kernels

chaxu01:
This patch adds support for INT2 quantization and KleidiAI SME2 kernels. The implementation targets SME2-capable devices, such as MacBook M4 and Vivo X300. Support for the corresponding NEON kernels will follow in a subsequent patch.


chaxu01:
Hi @ggerganov, I wanted to follow up on this PR and check if you had a chance to take a look.

This adds INT2 quantization support and KleidiAI SME2 kernels targeting SME2-capable devices.

There is one CI failure in ggml-ci-x64-nvidia-cuda related to a ROPE op, which doesn’t appear to be related to this change (no CUDA / ROPE modifications in the PR).

Appreciate any feedback when you get a chance — thanks!

ggml/src/ggml-common.h
```diff
@@ -255,6 +255,13 @@ typedef struct {
 } block_tq2_0;
 static_assert(sizeof(block_tq2_0) == sizeof(ggml_half) + QK_K / 4, "wrong tq2_0 block size/padding");
 
+#define QKQ2_0C 512
+typedef struct {
+    ggml_half d;
+    uint8_t qs[QKQ2_0C / 4];
+} block_q2_0c;
+static_assert(sizeof(block_q2_0c) == sizeof(ggml_half) + (QKQ2_0C / 4), "wrong q2_0c block size/padding");
+
```
ggerganov, line 296: What are the arguments for introducing this new data type over using the existing `block_tq2_0`?

chaxu01: Hi Georgi,

Sorry for the late response, and thanks for the question.

The motivation for introducing block_q2_0c is that it supports a new INT2 quantization format used by a recently announced model from Tencent: https://huggingface.co/AngelSlim/HY-1.8B-2Bit, which has different value mapping and block layout requirements compared to the existing block_tq2_0.

From a technical perspective:

Although both are INT2, the quantization semantics and packing layout are not compatible with tq2_0 (different centroids / collapsed mapping and block organization).

The SME2 kernels are written against this specific layout, so reusing block_tq2_0 would either require changing its meaning or adding special-case handling in the kernels.

Introducing a separate type keeps the existing tq2_0 path unchanged and makes the distinction explicit.

ggerganov: Have you tried mapping the data of this model to `Q2_K` and comparing the performance of the new `Q2_0C` to it? The `block_q2_k` layout should be able to map the values precisely. Even though it is not as efficient storage-wise, it has the advantage of being widely supported across the ggml backends.

chaxu01: Q2_0C turns out to be a very good fit for SME execution. Its layout and scaling work particularly well with SME kernels, allowing the inner loops to avoid extra shuffles and bookkeeping and making the kernels more efficient in practice.

At the moment, we don’t have a dedicated SME kernel that consumes Q2_K, so Q2_K naturally goes through the generic (non-SME) paths. From that perspective, the two formats aren’t directly comparable in terms of current kernel support.

That said, it’s an interesting idea, and adding SME support for Q2_K could be a worthwhile direction to explore longer-term.

ggerganov: The main problem with this proposal is that this is a very ARM-specific change and it is hard to justify the extra maintenance of such new type, given that it will work on very specific type of hardware. Generally, it would be fine to add `Q2_0C` at some point, but we need strong evidence that it will be worth it and the new Tencent models are going to get traction. Before that there is little reason to merge it to `master`.

I think the better approach is to map these models to `Q2_K`. Note that I don't mean straightforward quantizing from BF16 -> Q2_K. Rather, map the int2 values to q2_k blocks by ustilizing the `dmin` offset and repeating the scales as necessary. This would give broad exposure to all ggml/llama.cpp users and increase potential of these models getting adopted. Only after the adoption, we can think about customized SME implementations. Also, it would be necessary to consider if repacking q2_k to q2_0c would be possible - this is much easier to justify merging if it works.

Little0o0: Hi Georgi @ggerganov , I’ve been following the discussion and wanted to add some context from our end at AngelSlim. I totally get the preference for using Q2_K to keep the codebase clean, but forcing these models into a per-block format really makes the model size suffer compared to what we're seeing with Q2_0C.

The main thing is that Q2_0C is just much more friendly for CPU execution. Beyond inference, our recent research into QAT shows that a per-channel approach is significantly faster and more stable during training than the per-block methods used in Q2_K. It’s not just us, either industry SOTA 2-bit methods like Meta’s ParetoQ and MobileLLM are also moving toward per-channel QAT for these exact reasons.

Since these 2-bit models are specifically optimized for per-channel efficiency, Q2_0C is the most direct way to preserve those performance gains without making compromises. We really think this reflects where the research is heading for high-efficiency models, and having native support in master would be a huge help for anyone looking to deploy them effectively.

chaxu01: @ggerganov,  Thanks for the feedback — I’d like to clarify one important point about the proposed 2-bit format.

This format is not SME-specific and not Arm-specific. It was not introduced to support SME kernels in particular. The intent is broader: it is designed to be efficient and applicable across all compute backends, not just CPUs.

The key motivation for introducing this format is that existing 2-bit formats do not allow us to reliably determine whether we can apply per-channel-style acceleration. The proposed format is still per-block, but it explicitly guarantees that all scales within each block are identical. This gives backends the flexibility to:
- Run a per-block routine when appropriate, or
- Switch to a per-channel optimized kernel safely, since the scale invariance within the block is guaranteed.

This is not tied to SME — any backend that can benefit from reduced correction frequency and more regular scaling behavior can leverage this property. In fact, unless a compute engine has a fully hardwired data path, larger quantization granularity can reduce the rate of required corrections and simplify kernel logic, which may benefit GPUs and other accelerators as well.

Additionally, the LUT can be configured to minimize accuracy loss, so the format is not only structurally useful but also practical from a model-quality standpoint.

So the goal is to define a backend-agnostic format that enables broader optimization opportunities across architectures, rather than something tied to a specific hardware feature.

chaxu01: @ggerganov just a quick follow-up.

In my previous comment, I tried to clarify that the proposed 2-bit format is not intended to be Arm- or SME-specific, but rather backend-agnostic and applicable across architectures. From your perspective, does that address the concern about it being too hardware-specific, or are there still structural reasons you think it would be difficult to justify merging at this stage?

I’d really appreciate your guidance on what would make this proposal more acceptable.

ggerganov: I am still thinking about this - thanks for keeping the discussion going.

The main concern that I have is that IMO this new type has very high probability of not getting adopted at all. Even if we merge today, it is not going to be used by our users and downstream projects. Atm, we (the llama.cpp project) do not provide KleidiAI releases and I am not aware of any major llama.cpp-based project shipping a KliediAI-enabled builds to their users. So my assumption is that this functionality will end up used by ARM-specific applications and at the same time I don't see a strong feedback loop yet to make it worth it. Under these assumption, it seems more reasonable to keep this on a specialized branch until we address these concerns?

On the other hand, the changes are not too complex in terms of the core ggml library. I agree that support can be added for other backends. So it's also possible to merge and hope for some improvement in the future.

To summarize, my thoughts and recommendations for making this more acceptable are along the lines:

- Add support for `Q2_K` so that more people give these low-bit models a try
- With more people trying, there will be more incentive to support a new datatype that improves the performance of these models across all backends
- It would be useful to extend CI workflows for example with the server test suite running on Arm to increase confidence that the implementation is correct. (see for example the [Metal-based workflow](https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/server-metal.yml) that I added recently). We do have some CI established which is great, but for example I personally have no idea how KleidiAI behaves in a typical server-based workflow (e.g. chats, connecting with 3rd party apps, etc.) since we don't get feedback about such kind of applications
- Should we create KleidiAI-enabled releases of llama.cpp so more people can exercise the backend?

Let me know what you think about this PoV.

ggml/src/ggml-common.h
```diff
@@ -255,6 +255,13 @@ typedef struct {
 } block_tq2_0;
 static_assert(sizeof(block_tq2_0) == sizeof(ggml_half) + QK_K / 4, "wrong tq2_0 block size/padding");
 
+#define QKQ2_0C 512
```
ggerganov, line 290: This large block size could impose significant limitations on the models that can support this data type. Is it necessary to be so big?

ggml/src/ggml-common.h
chaxu01: @ggerganov thanks for the thoughtful feedback and for outlining your concerns and recommendations clearly.

You’re right that this is now more about adoption and maintenance strategy than just the technical implementation. The suggestions around Q2_K support, stronger CI coverage (especially server workflows on Arm), and potentially KleidiAI-enabled releases are all very reasonable.

We’ll discuss this internally and come back to you.

ggml/src/ggml-common.h
Lqlsoftware: I don't think 512 is a good choice either. But under the current quantization definition, there really can't find a good way to support per‑channel quantization. With smaller block size, the overhead per block (redundant scales) goes up significantly, increases the bpw and hurts GEMM/GEMV performance.

Is it possible to introduce dynamic block size to current quantization definition? s.t. other model with smaller hidden size can support this data type.

chaxu01: The 512 block size was chosen mainly to amortize the per-block metadata (scale/zero and related information) and to better match wide vector kernels, so we can avoid frequent corrections and extra shuffles in the inner loop. For kernels operating on large tiles, this improves runtime efficiency and reduces metadata bandwidth overhead.

That said, I agree that 512 may be too restrictive as a hard-coded default, especially for models with smaller hidden sizes.

In theory, a dynamic block size could help address this. However, in the current ggml/gguf design, block size is closely tied to the quantization layout and kernel implementation. Making it dynamic would require broader changes across the stack and could add complexity to the codebase. While possible, it would not be a lightweight change and would come with additional maintenance overhead.

ggml/src/ggml-common.h
chaxu01: @ggerganov Thanks again for the constructive discussion and thoughtful feedback.

Given the concerns around adoption and long-term maintenance, we'll keep this work on a branch for now. In the meantime, we'll focus on increasing KleidiAI's exposure and maturity — including further performance improvements, exploring KleidiAI-enabled release artifacts, and expanding CI coverage (including server-style workflows) to improve confidence and visibility for the backend.

We'll continue gathering usage data and feedback, and once we have a clearer adoption path, we can revisit upstreaming this work.

Really appreciate the open discussion — and we'll keep this thread open for any updates, further comments or suggestions.

PR #19361, opened 2026-02-05
server: Add cached_tokens info to oaicompat responses

percontation:
Adds OpenAI and Anthropic compatible information about the number of prompt tokens that came from cache.

https://platform.openai.com/docs/api-reference/chat/object#chat-object-usage-prompt_tokens_details-cached_tokens

https://platform.claude.com/docs/en/build-with-claude/prompt-caching#tracking-cache-performance
^ reference for subtracting the newly reported "cache_read_input_tokens" from "input_tokens"


ngxson (commented):
Can you add a test case for it? See https://github.com/ggml-org/llama.cpp/blob/master/tools/server/tests/unit/test_chat_completion.py

Btw, it will be better to re-group the `usage` object into a new struct, to avoid having to copy-paste it. Should be similar to the `result_prompt_progress` struct. We can either do it in this PR, or a follow-up PR.

percontation:
Test added.

I just looked at adding a `struct result_usage`, but, it felt unwieldily for two reasons:

1. The format "usage" block differs; `struct result_usage` would need three different `to_json` methods (oaicompat, oaicompat_resp, anthropic). Not the worst, but, it doesn't really save much repetition as those other structs.
2. The struct data members would be a subset of what's already in `struct result_timings` (cache_n, prompt_n, predicted_n). I realize there's already duplicated data around (e.g. between timings and n_decoded, n_prompt_tokens); but, `struct result_usage` would take that to a new level with the same bit of info now being repeated in 3 different places.

So instead, I opted to just write a helper `usage_json_oaicompat()` function, since that format of "usage" gets used 3 times. Thoughts?

ggerganov (approved):

Abioy:
Hi, any update on this PR? Looking forward for it.

ngxson (approved):

PR #19363, opened 2026-02-05
llama : rename llama-sampling to llama-sampler

danbev:
This commit addresses the TODO in llama-sampling.h to rename that header and the implementation to llama-sampler.


ggerganov (approved):

danbev:
I was able to build `examples/llama.android` locally using gradle so I don't think that failure is related to this PR so I'll merge this and follow up if the error persists.

PR #19368, opened 2026-02-05
llama : refactor sampling_info to use buffer_view template

danbev:
This commit updates the sampling_info struct in llama-context to use a buffer_view template for the logits, probs, sampled tokens, and candidates buffers.

The motivation for this is to simplify the code, improve type safety and readability.

### DIFF
--- a/src/llama-context.h
+++ b/src/llama-context.h
@@ -238,6 +238,16 @@ public:
 
     bool set_sampler(llama_seq_id seq_id, llama_sampler * sampler);
 
+    template <typename T>
+    struct buffer_view {
+        T * data;
+        size_t size = 0;
+
+        bool has_data() const {
+            return data && size > 0;
+        }
+    };
+
 private:
     llm_graph_params graph_params(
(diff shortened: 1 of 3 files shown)


src/llama-context.h
```diff
@@ -238,6 +238,16 @@ struct llama_context {
...
+    template <typename T>
+    struct buffer_view {
+        T * data;
+        size_t size = 0;
+
+        bool has_data() const {
+            return data && size > 0;
+        }
+    };
+
```
ggerganov: Better to be in `llama-impl.h`

danbev: @ggerganov I've moved this into llama-impl.h now. Please take a another look when you get a chance and let me know if there is anything else that needs changing.

ggerganov (commented):
For consistency, the non-sampling logits and embeddings could be wrapped in buffer view as well

ggerganov (approved):
Can merge if the CI is green

PR #19370, opened 2026-02-05
cmake: add variable to skip installing tests

WhyNotHugo:
When packaging downstream, there's usually little point in installing test. The default behaviour remains the same.


danbev (approved):

WhyNotHugo:
I don't think CI failures are due to my changes.

WhyNotHugo:
Rebased

PR #18334, opened 2025-12-24
[WIP] webui: simple MCP (Model Context Protocol) support (tool calls)

ochafik:
## Summary

_Closed as redundant w/ other efforts / PRs (see discussion below), but this works well for me_

This PR adds MCP (Model Context Protocol) support to llama-server's web UI:
- Tool calls with streaming support
- Config in `~/.llama.cpp/mcp.json` (reloaded automatically):
  - **Stdio servers**: WebSocket passthrough proxy (spawned as subprocesses by llama-server)
  - **Remote HTTP servers**: Streamable HTTP proxy
- Tool calls & their results displayed in collapsible blocks
- Web client creates and maintains its connections for active conversation (one for each enabled server, w/ autoconnect + manual (dis)connection support)

Screenshot 2025-12-24 at 00 33 25
Screenshot 2025-12-24 at 00 33 37
Screenshot 2025-12-24 at 00 33 44
Screenshot 2025-12-24 at 00 49 54
Screenshot 2025-12-24 at 00 51 38

### New CLI Options

```bash
# Enable MCP support with default config location (~/.llama.cpp/mcp.json)
./llama-server -m model.gguf --webui-mcp

# Or specify config path
./llama-server -m model.gguf --webui-mcp --mcp-config /path/to/mcp.json
```

### Configuration Example

```json
{
  "mcpServers": {
    "brave-search": {
      "command": "npx",
      "args": ["-y", "@brave/brave-search-mcp-server", "--transport", "stdio"],
      "env": {
        "BRAVE_API_KEY": "..."
      }
    },
    "fetch": {
      "command": "uvx",
      "args": ["mcp-server-fetch"]
    },
    "chrome-devtools": {
      "command": "npx",
      "args": ["chrome-devtools-mcp@latest"]
    },
    "python": {
      "command": "uvx",
      "args": ["mcp-run-python", "--deps", "numpy,pandas,pydantic,requests,httpx,sympy,aiohttp", "stdio"]
    },
    "some-remote-server": {
      "url": "http://127.0.0.1:38180/mcp",
      "headers": {
        "Authorization": "Bearer YOUR_API_KEY"
      }
    }
  }
}
```

### Why WebSocket for stdio?

  The WebSocket <-> stdio passthrough is the simplest possible implementation:
  - **No MCP knowledge required**: Server just forwards lines of text both ways
  - **No Streamable HTTP complexity**: WebSocket is naturally bidirectional (vs SSE + POST)
[...]

PR #19373, opened 2026-02-05
MCP for llama-cli

pwilkin:
So while the boys are having fun with MCP for the WebUI, I figured I'd take the other route. I've created *minimal* MCP support for `llama-cli`, using the already existing subprocess library, so the code needed is in fact quite small.

Features:
`--mcp-config`: pass a standard Cursor-style JSON config with MCP servers
`--mcp-yolo`: as the name suggests, disables approval queries for all tool calls

Attached is an MCP config for Serena, a very good plugin that instantly can make your `llama-cli` a vibe-coding agent!

Note: this is just proof of concept, no support for the key production features such as:
* FANCY COLORS FOR EVERYTHING
* memory
* session save/restore
* history
and the like.

EDIT: Ah yeah, AI disclosure. OF COURSE it's vibe coded, I mean, c'mon...


tools/cli/mcp.hpp
```diff
@@ -0,0 +1,519 @@
...
+        read_buffer(std::move(other.read_buffer)) {
+        // Zero out the source process to prevent double-free
+        other.process = {};
+        other.running = false;
+    }
+
+    mcp_server & operator=(mcp_server && other) noexcept {
+        if (this != &other) {
+            stop(); // Clean up current resources
+
```
ngxson: wondering why this is needed. it looks quite dangerous to me.

in general, `y = std::move(x)` should not stop anything. the job of stopping things should be for the deconstructor.

tools/cli/mcp.hpp
```diff
@@ -0,0 +1,519 @@
...
+    std::mutex                        mutex;
+    int                               next_id = 1;
+    std::map<int, std::promise<json>> pending_requests;
+
+    // Buffer for reading
+    std::string read_buffer;
+
+  public:
+    mcp_server(const std::string &                        name,
```
ngxson: the design of this class is quite mixed up IMO

low-level thing like stdio and subprocess managements and high-level things like jsonrpc are being mixed up

also, why do we need jsonrpc? AFAIK it's only useful if the client use the official MCP library from anthropic

tools/cli/cli.cpp
ngxson, line 1: too many non-related changes in this file

ngxson (commented):
I only have time to look at only 2 files in this PR, but it doesn't give me a good feeling in term of architecture design.

> OF COURSE it's vibe coded, I mean, c'mon...

The problem is not about vibe-coding, but AFAICT, vibe-designing doesn't go well in most cases.

What I fear is that the code base of `llama-cli` is current very simple and anything added on top of it can potentially blow up its complexity very quickly. That's why I want to wait for the MCP on web UI be released first, that way we can already observe what goes well and what goes wrong, before bringing it to the CLI.

> Note: this is just proof of concept

If that's the case, should this PR be set to draft?

tools/cli/mcp.hpp
pwilkin: Huh? No, JSON-RPC is literally *the* communication protocol for MCP Stdio transport:
https://modelcontextprotocol.io/specification/2025-06-18/basic/transports

It's not optional, that's the way MCP Stdio works - you do a fork + exec of the child process, then communicate with it by sending JSON-RPC messages over stdin/stdout.

tools/cli/cli.cpp
pwilkin: I'll roll back the syntax changes.

tools/cli/mcp.hpp
pwilkin: It's needed, because otherwise you'd orphan a lot of things.

The move operator reuses a lot of resources from the other `mcp_server` instance. But the instance that you're overwriting with the operator itself has a lot of handles, including process handles that have to be stopped. If you omit `stop()` here, for example:

```
read_thread = std::move(other.read_thread);
```

will overwrite the handle to the *current* read thread and therefore the thread will be orphaned.

pwilkin:
> The problem is not about vibe-coding, but AFAICT, vibe-designing doesn't go well in most cases.

It's *not* vibe-designed though. I explicitly designed it to minimize the code overhead. The entire functionality fits in basically a single class with 500 lines of code.

> What I fear is that the code base of `llama-cli` is current very simple and anything added on top of it can potentially blow up its complexity very quickly. That's why I want to wait for the MCP on web UI be released first, that way we can already observe what goes well and what goes wrong, before bringing it to the CLI.

But the communication protocol is *completely* different for the WebUI MCP and the CLI MCP.

The WebUI MCP communicates via standard HTTP. There is no spawning of any subprocesses and the MCP relies on the servers being already active.

The CLI MCP communicates via stdio, which requires the subprocess management component, but on the other hand doesn't require streaming / SSE or HTTP support.

I don't see why we'd really want to merge those two paths.

> > Note: this is just proof of concept
> 
> If that's the case, should this PR be set to draft?

Well, it's a usable proof-of-concept and I guess people might want to have MCP support in llama-cli anyways (since terminal agents are very popular these days).

tools/cli/mcp.hpp
ngxson: I don't quite care about how the official docs says about it. Instead, I recommend not to rely on the docs because it's misleading about what is a transport layer.

I hope you do you what is the OSI model. Transport layer, as the name suggests, has only one role: to transport bytes from one place to another place. Encapsulating data is usually at another layer.

That's why I said earlier that things are being mixed up here. The transport layer and the application layer (jsonrpc) should be 2 different components

pwilkin: Yeah, but you said "AFAIK it's only useful if the client use the official MCP library from anthropic" - that's not the case, you just communicate using JSON-RPC messages with any STDIO MCP server.

I can of course separate the raw byte read and the JSON parsing, but for such a small class didn't really feel it was necessary.

pwilkin:
@ngxson aight, I've split the communication into a transport layer and a messaging layer, with the transport layer handling the low-level comms and the messaging layer in charge of the JSON-RPC. If you feel it's too much for one file, I can move it out of an `.hpp` structure into a standard header/source file structure.

ngxson:
> But the communication protocol is _completely_ different for the WebUI MCP and the CLI MCP.

I don't see how they are different. The protocol being used is JSON-RPC. Just the transport is different.

Let me bring up the OSI model for the second time:
- MCP is an application layer protocol based on JSON-RPC
- Everything that can deliver bytes from one place to another can be considered as "transport", including stdio, HTTP streamable, HTTP SSE, websocket. Even TCP over [this RFC 1149 joke](https://en.wikipedia.org/wiki/IP_over_Avian_Carriers) is a completely valid transport method

> The WebUI MCP communicates via standard HTTP. There is no spawning of any subprocesses and the MCP relies on the servers being already active.

Except for yes, server can and will have an implementation to spawn subprocess while bridging the data between stdio <--> HTTP SSE

The overall abstraction will look like [nodejs Stream interface](https://nodejs.org/api/stream.html) (but much more light-weight), and the implementation under the hood (i.e. stdio, HTTP, pigeon) is irrelevant

ngxson:
Btw, I think this PR has the same misconception as https://github.com/ggml-org/llama.cpp/pull/18334, about what is a "MCP transport"

I recommend discussing thoroughly about this first, in order to avoid any potential double efforts with the on-going MCP integration on web UI.

pwilkin:
I do understand the OSI separation :) I just thought that the idea behind the WebUI part was to do the part of MCP that does not require spawning any additional processes outside the browser.

The streaming proxy you mentioned in #18334 certainly does make sense as a unifying approach. I assume your idea is that the `server.cpp` launches and handles the proxy and then both the CLI and the WebUI just communicate with it?

I agree MCP specs are a total mess. The issue here is that the so-called stdio transport is both a transport *and* an interprocess management/communication layer, which is why, although the protocol (JSON-RPC) is the same, the implementation differs (because, by its nature, stdio is not platform-agnostic, i.e. you can't really use stdio transport without the ability to spawn processes). I just wanted a "quick win" since you already used the subprocess library which does most of IPC work, so I though I'd do a lightweight demo on how easy it is to transform cli into a full-fledged agentic client. But if you think this will interact with the common MCP integration, I'm fine to fork it into a separate tool until the full-fledged integration is done.

am17an:
I think llama-server already works quite well with OpenCode, we just need the model that "just works"

jacekpoplawski:
> I think llama-server already works quite well with OpenCode, we just need the model that "just works"

I have great experiences with GLM-4.7-Flash (+llama.cpp), coded a retro-shooter in C++ without big issues

ngxson:
> The streaming proxy you mentioned in #18334 certainly does make sense as a unifying approach. I assume your idea is that the `server.cpp` launches and handles the proxy and then both the CLI and the WebUI just communicate with it?

Yes, the long-term plan is to have one single `--mcp config.json` that shares between both server and CLI. The overview of the system will be like:

```mermaid
graph TD
    Frontend <--Custom SSE,HTTP--> Server 
    CLI <--C++ interface--> Server

    subgraph Server
        Stream_proxy
    end

    Stream_proxy <-- stdio --> Child_process
    Stream_proxy <-- SSE,HTTP,WS --> External_MCP_server
```

The on-going MCP work on web UI already provide the HTTP client/proxy implementation, which in theory should already allow CLI to communicate with an MCP server via HTTP SEE. Although, the implementation is still very basic.

If you want to proceed with this, I think having an abstraction of `tool_registry` is the first good step. This allow having tools not just from MCP, but also built-in tools, which already allow bunch of agentic usages. A list of basic `tool` like `list_files`, `read_file`, `write_file`, `shell_command` is already enough for start. On top of that, a `tool` can also encapsulate a [skill](https://agentskills.io/home), which is just a simple prompt stored in a file.

pwilkin:
All right. All this feedback makes sense, so here's what I'm gonna do:
-> change this back into draft
-> decouple this from `llama-cli` for now, add `llama-vibe` for people who want to still experiment with the simple CLI+MCP model
-> start working on `tool-registry` as @ngxson suggested, maybe also add a test to verify whether a given model will call those tools correctly, implement `register_tool` and `call_tool` abstractions

PR #19374, opened 2026-02-05
WebUI hide models in router mode

crsawyer:
When using `llama-server` with presets in router mode add the ability to completely hide certain models from the WebUI.

## Summary
 Adding `no-webui = true` to a model's preset configuration will exclude it from the WebUI.

## Why?
I tend to use the preset capability to load a group of models for offline vibe coding, but I don't want my embedding, reranking, and FIM models to be selectable via the WebUI. I also wanted it so that if there is only one available model after filtering, it would be automatically selected. I chose to reuse the `no-webui` argument for this. So that any model in the preset file that has `no-webui = true` will be excluded from the WebUI model selection list. *This also means that the excluded models cannot be loaded or unloaded via the WebUI*.

## Example:

**`models.ini`**
```ini
[*]
ngl = 999
threads = -1

; chat, tools
[gpt-oss-120b]
hf = ggml-org/gpt-oss-120b-GGUF
load-on-startup = true
jinja = true
flash-attn = 1
ubatch-size = 2048
batch-size = 32768
parallel = 2
; ctx-size = 131072*params.n_parallel
ctx-size = 262144
temp = 1.0
;min-p = 0.0
min-p = 0.01
top-p = 1.0
top-k = 0.0
; --- Speculative
spec-type = ngram-mod
spec-ngram-size-n = 24
draft-min = 48
draft-max = 64

; FIM
[Qwen2.5-Coder-7B-Q8_0]
load-on-startup = true
hf = ggml-org/Qwen2.5-Coder-7B-Q8_0-GGUF:Q8_0
flash-attn = 1
ubatch-size = 1024
batch-size = 1024
ctx-size = 0
cache-reuse = 256
parallel = 2
; ctx-size = 32768*params.n_parallel
ctx-size = 65536
no-webui = true

; embedding
[Qwen3-Embedding-0.6B]
load-on-startup = true
hf = Qwen/Qwen3-Embedding-0.6B-GGUF:Q8_0
flash-attn = 1
embedding = true
pooling = last
ubatch-size = 8192
verbose-prompt = true
parallel = 2
; ctx-size = 8192*params.n_parallel
ctx-size = 16384
no-webui = true

; reranking
[Qwen3-Reranker-0.6B]
load-on-startup = true
hf = ggml-org/Qwen3-Reranker-0.6B-Q8_0-GGUF:Q8_0
flash-attn = 1
rerank = true
parallel = 2
; ctx-size = 4096*params.n_parallel
ctx-size = 8192
no-webui = true
```

And now run it:
```bash
llama-server --offline --log-colors on --log-prefix --log-timestamps --no-models-autoload --models-preset models.ini
```

*Note: I haven't tested the ini file as written above. I pre-downloaded the models and use `model = modelfile.gguf` instead of the listed `hf = org/model`. Aside from that, this ini file was what I tested against.*


tools/server/webui/src/lib/components/app/models/ModelsSelector.svelte
```diff
@@ -47,7 +47,13 @@
 		upToMessageId
 	}: Props = $props();
 
-	let options = $derived(modelOptions());
+	let options = $derived(
+		modelOptions().filter((option) => {
+			// Exclude models that are marked as not for webui use
+			const modelProps = modelsStore.getModelProps(option.model);
+			return modelProps?.webui !== false;
```
allozaur: Comment here is redundant, the code is self-explanatory.

```suggestion
			const modelProps = modelsStore.getModelProps(option.model);

			return modelProps?.webui !== false;
```

tools/server/webui/src/lib/components/app/models/ModelsSelector.svelte
```diff
@@ -330,6 +336,10 @@
 			return options.find((option) => option.id === activeId);
 		}
 
+		if (options.length === 1) {
+			// Only one option - show it
+			return options[0];
+		}
```
allozaur, line 256:
```suggestion
		if (options.length === 1) {
			return options[0];
		}
```

allozaur (changes requested):

crsawyer (commented):
Changes applied as requested

allozaur (changes requested):
@crsawyer please rebase this branch and rebuild the webui static output. then we're good to go

allozaur (approved):

Issue #19379, opened 2026-02-05
server: router mode subprocess still occupies GPU after sleep-idle-seconds


leonardcser:
### Feature Description

In router mode, when `--sleep-idle-seconds` triggers, the child subprocess unloads the model from VRAM but the process remains alive and attached to the GPU, consuming ~600MiB per idle subprocess:

```
# Active
467282 dev 0 Compute  0% 10386MiB 11% 824MiB llama-server ...

# After sleep-idle triggers — process still on GPU
467282 dev 0 Compute N/A   614MiB  1% 369MiB llama-server ...
```

### Motivation

Idle subprocesses should not remain as GPU processes when they are not needed. With multiple models in router mode, the residual ~600MiB per dormant process wastes significant VRAM.

### Relation to #18189

Follow-up to #18189. PR #18228 implemented `--sleep-idle-seconds`, but it only unloads the model within the living process — it does not terminate the subprocess. The original request was closed as stale without this being addressed.

### Possible Implementation

A new option (e.g. `--unload-idle-seconds`) that triggers full subprocess termination in router mode via the existing `unload()` path. The building blocks are already there:

- `server_queue` already tracks idle time
- `server_models::unload()` already handles graceful shutdown → force-kill
- These two just need to be wired together, with the router re-spawning the process on the next request (same as `--models-max` LRU eviction already does)

IMbackK:
What kind of gpu? If its rocm, rocblas uses ~500MB of vram for itself to load its solutions, this is out of controll of the applicaiton and happens when the first rocblas call is made and is only freed when the application exits.

I could imagine cublas might have similar behavior.

Just as suggestion where the allocation here could be coming from

leonardcser:
The card in question is NVIDIA RTX PRO with cuda13.

But even if we cannot control the memory, I feel it would make more sense to stop the processes running on the GPU.

Currently we "unload" the vram but we still have a dormant process attached.

apunkt:
using b8215 I was able to manually unload models and no GPU processes remained
using b8261 I was able to manually unload models when sleeping, but GPU processes remain

using CUDA 13 and 12

apunkt:
b8405 works for me

leonardcser:
Not working on b8720.

On startup, the router reserves ~600MiB for "unloaded" models.
When model is loaded and goes back to sleep, it is still attached to the GPU processes with ~600MiB.
I was able to manually unload models and no GPU processes remained.

```
[...]

Issue #18189, opened 2025-12-19
Feature Request: Idle model unload timeout (router mode / config.ini)


sammcj:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

It would be great if llama-server when run in router mode and/or with a config.ini could unload models that have been idle for a given time.

For example, you could default it to 300 (seconds) and have models that haven't performed any inference in 5 minutes be unloaded to free up resources.

### Motivation

Models current sit loaded until the max loaded models are met, using up system resources.

[llama-swap provides](https://github.com/mostlygeek/llama-swap#configuration) this as a `ttl` setting.

Somewhat related:
- https://github.com/ggml-org/llama.cpp/issues/11703
- https://github.com/ggml-org/llama.cpp/issues/4598

### Possible Implementation

```ini
unload-idle-seconds = 300

[some-custom-important-model]
m = /models/custom.gguf
unload-idle-seconds = 3600
```

```bash
llama-server --unload-idle-seconds 300 --models-dir /models
```

ggerganov:
I agree this functionality is quite useful. The Metal backend already indirectly supports it thanks to Apple's wired memory collector (#17766) and I find it very convenient that my models are "unloaded" after a configurable amount of time (currently 3 minutes by default).

ngxson:
This requires https://github.com/ggml-org/llama.cpp/issues/18163 , so will be done after the parent issue is resolved

ngxson:
@ggerganov not sure if this idea worth exploring, but I've been thinking about having different options of "sleeping" in side libllama: the one option of sleeping can be to unload the model weight (maybe the full `llama_model` or just a selection of tensors), another option can be to also free KV memory. or, 2 options can be used together.

ngxson:
I added a "sleeping" feature for both single-model and router modes: https://github.com/ggml-org/llama.cpp/pull/18228

In the future, it can be quite interesting to allow configure the "level" of sleep (as mentioned in the last comment). We can also allow customize unloading only the main model, draft model, multimodal encoder, or any combination of them.

[...]

PR #18228, opened 2025-12-20
server: add auto-sleep after N seconds of idle

ngxson:
## Sleeping on Idle

The server supports an automatic sleep mode that activates after a specified period of inactivity (no incoming tasks). This feature, introduced in [PR #18228](https://github.com/ggml-org/llama.cpp/pull/18228), can be enabled using the `--sleep-idle-seconds` command-line argument. It works seamlessly in both single-model and multi-model configurations.

When the server enters sleep mode, the model and its associated memory (including the KV cache) are unloaded from RAM to conserve resources. Any new incoming task will automatically trigger the model to reload.

Note that the following endpoints are exempt from being considered as incoming tasks. They do not trigger model reloading and do not reset the idle timer:
- `GET /health`
- `GET /props`

---

## Implementation

The implementation of this feature consists of 3 main parts:
- `server_queue` sleeping state
- `server_context` sleeping state
- `server_res_generator` hook

The main loop inside `server_queue` acts as a watchdog timer (so we can avoid spawning a dedicated thread just for the watchdog). Upon timing condition passed, it signals to `server_context` to unload the model.

`server_res_generator` hooks on any incoming request, and will ask the `server_queue` to resume if it is in sleeping state. Note that some requests like `/health` bypass this check (they can only access read-only data of `server_context`)

Upon requested to resume, `server_queue` signals `server_context` to reload models, then unblock `server_res_generator` to proceed with the rest of the request.

IMPORTANT: for thread-safety reason, any requests to server will be blocked/delayed by `server_res_generator` until the server exits sleeping state

PR #19380, opened 2026-02-05
server: add `--stop-idle-seconds` for router mode

leonardcser:
## Summary
- Adds `--stop-idle-seconds` CLI flag that fully terminates idle model subprocesses in router mode after N seconds of inactivity
- On next request, `ensure_model_loaded()` re-spawns the process automatically (when `--models-autoload` is enabled)
- Unlike `--sleep-idle-seconds` which unloads VRAM/RAM within a child process, this flag kills the subprocess entirely

Refs #19379, follow-up to #18189

## Test plan
- [x] Build and verify `--stop-idle-seconds` appears in `--help`
- [x] Run router mode with `--stop-idle-seconds 5 --models-dir <dir>`, send a request, wait >5s, verify process is terminated
- [x] Send another request and verify the model re-spawns via autoload
- [x] Verify existing router tests still pass

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2998,6 +2998,16 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_MODELS_AUTOLOAD"));
     add_opt(common_arg(
+        {"--stop-idle-seconds"}, "SECONDS",
+        string_format("for router server, fully terminate model instances after N seconds of inactivity (default: %d; -1 = disabled)", params.stop_idle_seconds),
+        [](common_params & params, int value) {
+            if (value == 0 || value < -1) {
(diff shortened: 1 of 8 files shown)


ngxson:
I disagree with this feature, it sounds redundant to me.

The auto-sleep functionality added in https://github.com/ggml-org/llama.cpp/pull/18228 should already allow model weight to be unloaded after a timeout, while also allowing static endpoints like `/props`, `/models` to be accessed without waking up the server. The current PR is missing that, and adding it will just duplicate a lot of code.

If the server instance need to be unloaded, it's up to the downstream application to decide when to unload it.

leonardcser:
Thanks for the feedback

apunkt:
> If the server instance need to be unloaded, it's up to the downstream application to decide when to unload it.

agree, however unloading from downstream app now still leaves ps on gpu as mentioned in [#19379](https://github.com/ggml-org/llama.cpp/issues/19379#issuecomment-4032146553)_

Issue #19256, opened 2026-02-01
Misc. bug: Llama-cli saving output to file using ">" is not working anymore


anthonybudd:
### Name and Version

I am trying to save the output of llama-cli to a text file using ">"

Example:
`llama-cli -m model.gguf -p "what is the fastsest car" > file.txt`

I compiled `lama-cli` from source on ARM running Debain Trixie.

Notes:
While trying to figure out what was going on I tried calling llama-cli using Node.js `child_process`. If you try and capture the output using the below code `stdout` and `stderr` are never called, although the model output is clearly printed on the screen. so am making a bug report.

```ts
const childProcess = spawn('llama-cli', [
                '-m', 'model.gguf',
                '-p', '"what is the fastest car?"'
            ]);

            childProcess.stdout.setEncoding('utf8');
            childProcess.stdout.on('data', (chunk: Buffer | string) => {
                const text = typeof chunk === 'string' ? chunk : chunk.toString('utf8');
                console.log(`[llama.cpp stdout]:`, text);
            });

            childProcess.stderr.setEncoding('utf8');
            childProcess.stderr.on('data', (chunk: Buffer | string) => {
                const text = typeof chunk === 'string' ? chunk : chunk.toString('utf8');
                console.log(`[llama.cpp stderr]:`, text);
            });

            childProcess.stdin.write(`${sanitizedPrompt}\n`, 'utf8', (error?: Error | null) => {
                if (error) {
                    console.error('[LLM] Error writing to stdin:', error);
                } else {
                    console.log('[LLM] Prompt written to stdin successfully');
                }
            });
```

### Operating systems

Linux

### Which llama.cpp modules do you know to be affected?

llama-cli

### Command line

```shell
llama-cli -m model.gguf -p "what is the fastest car?" > file.txt
```

### Problem description & steps to reproduce

Run a command with a prompt
`llama-cli -m model.gguf -p "what is the fastest car?" > file.txt`

Then output the contents of the file with
`cat file.txt`

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

taronaeo:
After the `llama-cli` refactor, stdout redirection is no longer officially supported. IIRC the correct flag is now `-o` for output file.

MB7979:
I have this issue as well and it’s wrecked my workflow. I get an invalid argument error for --output-file and -o. 

Using “>” still works in llama-completion.

NechamaKrashinski:
[...]

PR #19384, opened 2026-02-06
console: fix output redirection by detecting TTY

MengAiDev:
When stdout is redirected (not a TTY), automatically disable advanced display mode to allow output to be captured by shell redirection and programmatic capture (e.g., Node.js child_process).

Previously, console output was always written to /dev/tty when advanced display was enabled, preventing redirection from working. This fix checks if stdout is a TTY on both POSIX (isatty) and Windows (GetConsoleMode) systems, and falls back to simple I/O mode when output is redirected.

Fixes issue#19256

### DIFF
--- a/common/console.cpp
+++ b/common/console.cpp
@@ -87,6 +87,23 @@ namespace console {
     void init(bool use_simple_io, bool use_advanced_display) {
         advanced_display = use_advanced_display;
         simple_io = use_simple_io;
+
+        // If stdout is redirected (not a TTY), disable advanced display to allow output redirection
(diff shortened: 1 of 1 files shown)


ngxson:
As commented on https://github.com/ggml-org/llama.cpp/issues/19256, the better way is to implement `-o` in CLI.

The CLI is not meant to be captured by other programs, much like how other CLI like claude code or gpt codex works

inferrna:
> much like how other CLI like claude code or gpt codex works

Nice. Big corps teach unix guys how their CLI should work.

PR #19400, opened 2026-02-06
rpc : update from common.cpp

angt:
Not sure why we want to avoid the linking 🤔


rgerganov:
> Not sure why we want to avoid the linking 🤔

The comment is misleading, the actual problem is that there is no `libcommon.so` that we can link with.

rgerganov (approved):

PR #19411, opened 2026-02-07
ci : use less jobs when building with sanitizers

ggerganov:
ref https://github.com/ggml-org/llama.cpp/pull/19291#issuecomment-3862954095

I think this might help with sanitizer jobs being randomly terminated (such as https://github.com/ggml-org/llama.cpp/actions/runs/21761592313/job/62786390511).

PR #19415, opened 2026-02-07
ci : enable ggml sanitizers for webui

CISC:
cont #19323

### DIFF
--- a/.github/workflows/server-webui.yml
+++ b/.github/workflows/server-webui.yml
@@ -107,12 +107,12 @@ jobs:
 
     strategy:
       matrix:
-        sanitizer: [ADDRESS, UNDEFINED] # THREAD is broken
+        sanitizer: [ADDRESS, UNDEFINED] # THREAD is very slow
         build_type: [RelWithDebInfo]
         include:
           - build_type: Release
             sanitizer: ""
-      fail-fast: false # While -DLLAMA_SANITIZE_THREAD=ON is broken
+      fail-fast: false
 
     steps:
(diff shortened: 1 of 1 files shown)


ngxson:
Maybe I missed something, but it seems like building llama-server and run server tests in webui workflow is redundant. I suppose @allozaur copied it from the normal server CI?

FWIW, there is just one single test in the whole test suite `tools/server/tests` that is related to webui, but it only test if `-no-webui` work correctly, so completely unrelated to webui build test.

I suppose the correct way is that: webui workflow only need to test if the svelte code is (1) well-formatted and (2) `npm run build` without any errors.

CISC:
> Maybe I missed something, but it seems like building llama-server and run server tests in webui workflow is redundant. I suppose @allozaur copied it from the normal server CI?

I guess, didn't really look too closely when I split it into its own workflow.

> FWIW, there is just one single test in the whole test suite `tools/server/tests` that is related to webui, but it only test if `-no-webui` work correctly, so completely unrelated to webui build test.

It does seem to run different tests than `server.yml` though, maybe they need to be moved over there?

ngxson:
> It does seem to run different tests than `server.yml` though, maybe they need to be moved over there?

It's not really different, but I guess because this code was copied from the version when we still use `server/tests/tests.sh` script in the workflow, so it sticking around until now. In any cases, I think it's safe to remove any pytest-related logic here and move them to `server.yml` workflow

CISC:
Will follow up on #19411 instead.

PR #19416, opened 2026-02-07
Update ROCm docker container to 7.2 release

superm1:
The current release (7.2) also includes binaries for gfx1150 (Strix Point).

https://rocm.docs.amd.com/projects/radeon-ryzen/en/docs-7.2/docs/compatibility/compatibilityryz/native_linux/native_linux_compatibility.html


superm1:
In my local testing this causes problems on gfx908 though, I'll work through those and open a new PR.

AFAICT it seems like conflicting template specializations for gfx908 with `__half` vs `hfloat16_t`.

PR #19419, opened 2026-02-07
chat: fix case where template accepts type content only

ngxson:
Fix chat template of PaddleOCR-VL, which requires content to be an array (see https://github.com/ggml-org/llama.cpp/pull/18825)

This should be able to handle these case:
- Template supports ONLY string content
- Template supports ONLY array content
- Template supports both string AND array content

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -3033,7 +3033,6 @@ static json render_message_to_json(const std::vector<common_chat_msg> & msgs, co
                     }
                 });
             }
-            printf("%s\n", jmsg.dump().c_str());
             messages.push_back(jmsg);
         } else {
             json jmsg = msg.to_json_oaicompat(/* concat_typed_text= */ false);
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -380,14 +380,44 @@ std::vector<common_chat_msg> common_chat_msgs_parse_oaicompat(const json & messa
     return msgs;
 }
 
-// DEPRECATED: only used in tests
-json common_chat_msgs_to_json_oaicompat(const std::vector<common_chat_msg> & msgs, bool concat_typed_text) {
+static json render_message_to_json(const std::vector<common_chat_msg> & msgs, const jinja::caps & c) {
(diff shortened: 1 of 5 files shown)


ngxson:
Pinging @pwilkin @aldehir @CISC if you can have a quick look on this (currently a roadblock for PaddleOCR). Thanks!

common/chat.cpp
```diff
@@ -3009,6 +3010,39 @@ static void use_generic_schema(json & messages) {
...
+            json jmsg = msg.to_json_oaicompat(/* concat_typed_text= */ false);
+            if (jmsg.at("content").is_string()) {
+                jmsg["content"] = json::array({
+                    json{
+                        {"type", "text"},
+                        {"text", jmsg.at("content").get<std::string>()},
+                    }
+                });
+            }
+            printf("%s\n", jmsg.dump().c_str());
```
pwilkin: Remove or convert to LOG_*

common/chat.h
```diff
@@ -240,6 +240,8 @@ bool common_chat_templates_support_enable_thinking(const common_chat_templates *
 
 // Parses a JSON array of messages in OpenAI's chat completion API format.
 std::vector<common_chat_msg> common_chat_msgs_parse_oaicompat(const nlohmann::ordered_json & messages);
+
+// DEPRECATED: only used in tests
```
pwilkin, line 244: Wouldn't it be a good idea just to replace `common_chat_msgs_to_json_oaicompat` with the body of the new render_message_to_json (with the extended signature with full caps) and make the old common_chat_msgs_to_json_oaicompat call it with just the single cap set?

pwilkin (approved):

common/chat.h
ngxson: Good idea, I implemented in https://github.com/ggml-org/llama.cpp/pull/19419/commits/6edb330ffa4090d3e9c6b45d2a263356e12c6cdc , `test-chat` is OK

CISC (approved):

aldehir (approved):

PR #19422, opened 2026-02-07
ggml-cpu: FA add GEMM microkernel

am17an:
This PR contains the following improvements for the tiled FA kernel
- Add a simd gemm for float32 in the tiled FA kernel. 
- Tune tile sizes for larger context
- Remove condition that kv depth % KV_TILE_SZ == 0

Future work would be adding a f16 version for hardware that supports it.

Results on 64c EPYC server, similar speed-ups for 16/32 cores. 

| Model                 | Test         |   t/s master |   t/s opt-fa-micro-gemm |   Speedup |
|:----------------------|:-------------|---------------------:|------------------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512        |               234.75 |                  237.96 |      1.01 |
| gpt-oss 20B MXFP4 MoE | pp512@d1024  |               201.78 |                  219.03 |      1.09 |
| gpt-oss 20B MXFP4 MoE | pp512@d2048  |               190.70 |                  207.59 |      1.09 |
| gpt-oss 20B MXFP4 MoE | pp512@d4096  |               167.51 |                  195.65 |      1.17 |
| gpt-oss 20B MXFP4 MoE | pp512@d8192  |               131.76 |                  163.66 |      1.24 |
| gpt-oss 20B MXFP4 MoE | pp512@d16384 |                88.44 |                  130.36 |      1.47 |
| gpt-oss 20B MXFP4 MoE | pp512@d32768 |                56.46 |                   91.62 |      1.62 |
| llama 8B Q4_K_M       | pp512        |               184.26 |                  180.40 |      0.98 |
| llama 8B Q4_K_M       | pp512@d1024  |               170.39 |                  182.02 |      1.07 |
| llama 8B Q4_K_M       | pp512@d2048  |               155.95 |                  162.77 |      1.04 |
| llama 8B Q4_K_M       | pp512@d4096  |               124.17 |                  132.95 |      1.07 |
| llama 8B Q4_K_M       | pp512@d8192  |                90.52 |                  103.16 |      1.14 |
| llama 8B Q4_K_M       | pp512@d16384 |                56.15 |                   79.88 |      1.42 |
| llama 8B Q4_K_M       | pp512@d32768 |                31.33 |                   49.53 |      1.58 |

AI disclosure: I wrote the register blocked micro kernel for AVX2, but I let AI handle rest of the kernel.

### DIFF
--- a/ggml/src/ggml-cpu/common.h
+++ b/ggml/src/ggml-cpu/common.h
@@ -6,8 +6,8 @@
 #include "ggml-impl.h"
 #include "simd-mappings.h"
 
-#define GGML_FA_TILE_Q  32
-#define GGML_FA_TILE_KV 16
+#define GGML_FA_TILE_Q  64
+#define GGML_FA_TILE_KV 64
 
 #ifdef __cplusplus
 
--- a/ggml/src/ggml-cpu/ops.cpp
+++ b/ggml/src/ggml-cpu/ops.cpp
@@ -3,6 +3,7 @@
 #include "ggml-cpu.h"
 #include "ggml-impl.h"
 #include "binary-ops.h"
+#include "simd-gemm.h"
 #include "ggml.h"
 #include "unary-ops.h"
 #include "vec.h"
@@ -8389,10 +8390,6 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
     GGML_ASSERT(k->type == v->type);
     const ggml_type kv_type = k->type;
 
-    const auto * kv_type_traits_cpu = ggml_get_type_traits_cpu(kv_type);
-    const ggml_from_float_t kv_from_float = kv_type_traits_cpu->from_float;
-    const ggml_vec_dot_t    kv_vec_dot    = kv_type_traits_cpu->vec_dot;
-    const size_t kv_type_size = ggml_type_size(kv_type);
 
     // broadcast factors
     const int64_t rk2 = neq2/nek2;
@@ -8423,8 +8420,9 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
 
     static constexpr int Q_TILE_SZ  = ggml_fa_tile_config::Q;
     static constexpr int KV_TILE_SZ = ggml_fa_tile_config::KV;
-
-    GGML_ASSERT(nek1 % KV_TILE_SZ == 0 && "KV sequence length must be divisible by KV_TILE_SZ");
+#ifdef GGML_SIMD
+    GGML_ASSERT(DV % GGML_F32_EPR == 0);
+#endif
 
     int ir = ir0;
     while (ir < ir1) {
@@ -8452,18 +8450,20 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
         }
 
         // Per-thread scratch layout:
-        // Q_q:    Q_TILE_SZ * DK (converted Q tile in KV type)
+        // Q_q:    Q_TILE_SZ * DK (converted Q tile — F32 for GEMM, KV type for scalar)
         // KQ:     Q_TILE_SZ * KV_TILE_SZ (attention scores in float)
         // mask:   Q_TILE_SZ * KV_TILE_SZ (mask in float)
         // VKQ32:  Q_TILE_SZ * DV (FP32 output accumulator)
-        // V32:    KV_TILE_SZ * DV (F32 buffer for V tile - used for f166 conversion)
-        float * base  = (float *) params->wdata + ith*(Q_TILE_SZ*DK + 2*Q_TILE_SZ*KV_TILE_SZ + Q_TILE_SZ*DV + KV_TILE_SZ*DV + CACHE_LINE_SIZE_F32);
+        // V32:    KV_TILE_SZ * DV (F32 buffer for V tile)
+        // K_f32:  KV_TILE_SZ * DK (F32 buffer for K tile — GEMM path)
+        float * base  = (float *) params->wdata + ith*(Q_TILE_SZ*DK + 2*Q_TILE_SZ*KV_TILE_SZ + Q_TILE_SZ*DV + KV_TILE_SZ*DV + KV_TILE_SZ*DK + CACHE_LINE_SIZE_F32);
 
         void  * Q_q    = base;
         float * KQ     = (float *)((char *)base + Q_TILE_SZ * DK * sizeof(float));
         float * mask32 = KQ + Q_TILE_SZ * KV_TILE_SZ;
         float * VKQ32  = mask32 + Q_TILE_SZ * KV_TILE_SZ;
-        float * V32    = VKQ32 + Q_TILE_SZ * DV;  // F32 buffer for V tile
+        float * V32    = VKQ32 + Q_TILE_SZ * DV;
+        float * K_f32  = V32 + KV_TILE_SZ * DV;
 
         memset(VKQ32, 0, Q_TILE_SZ * DV * sizeof(float));
         memset(mask32, 0, Q_TILE_SZ * KV_TILE_SZ * sizeof(float));
@@ -8476,27 +8476,34 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
         const int iv3 = iq3 / rv3;
         const int iv2 = iq2 / rv2;
 
-        for (int tq = 0; tq < tile_rows; tq++) {
-            const float * pq = (const float *) ((char *) q->data + ((iq1 + tq)*nbq1 + iq2*nbq2 + iq3*nbq3));
-            kv_from_float(pq, (char *)Q_q + tq * DK * kv_type_size, DK);
-        }
-        // Zero-pad remaining rows
-        for (int tq = tile_rows; tq < Q_TILE_SZ; tq++) {
-            memset((char *)Q_q + tq * DK * kv_type_size, 0, DK * kv_type_size);
+        {
+            float * Q_f32 = (float *)Q_q;
+            for (int tq = 0; tq < tile_rows; tq++) {
+                const float * pq = (const float *) ((char *) q->data + ((iq1 + tq)*nbq1 + iq2*nbq2 + iq3*nbq3));
+                memcpy(Q_f32 + tq * DK, pq, DK * sizeof(float));
+            }
+            for (int tq = tile_rows; tq < Q_TILE_SZ; tq++) {
+                memset(Q_f32 + tq * DK, 0, DK * sizeof(float));
+            }
         }
 
         for (int64_t ic = 0; ic < nek1; ic += KV_TILE_SZ) {
+            const int kv_tile = (int)std::min((int64_t)KV_TILE_SZ, nek1 - ic);
 
             // skip the tile entirely if all the masks are -inf
             if (mask) {
                 bool can_skip = true;
                 for (int tq = 0; tq < tile_rows; tq++) {
                     const ggml_fp16_t * mp_row = (const ggml_fp16_t *)((const char *) mask->data + (iq1 + tq)*mask->nb[1] + (iq2%mask->ne[2])*mask->nb[2] + (iq3%mask->ne[3])*mask->nb[3]);
-                    for (int tk = 0; tk < KV_TILE_SZ; tk++) {
+                    for (int tk = 0; tk < kv_tile; tk++) {
                         mask32[tq * KV_TILE_SZ + tk] = slope * GGML_CPU_FP16_TO_FP32(mp_row[ic + tk]);
                         if (mask32[tq * KV_TILE_SZ + tk] != -INFINITY) {
                             can_skip = false;
                         }
+                    }
+                    // Pad remaining mask entries with -inf
+                    for (int tk = kv_tile; tk < KV_TILE_SZ; tk++) {
+                        mask32[tq * KV_TILE_SZ + tk] = -INFINITY;
                     }
                 }
 
@@ -8505,13 +8512,33 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
                 }
             }
 
-            for (int tq = 0; tq < Q_TILE_SZ; tq++) {
-                const void * q_row = (const char *)Q_q + tq * DK * kv_type_size;
-                for (int tk = 0; tk < KV_TILE_SZ; tk++) {
-                    const void * k_row = (const char *) k->data + ((ic + tk)*nbk1 + ik2*nbk2 + ik3*nbk3);
-                    float s;
-                    kv_vec_dot(DK, &s, 0, k_row, 0, q_row, 0, 1);
-                    KQ[tq * KV_TILE_SZ + tk] = s * scale;
+            // Pack K tile transposed: K_f32[dk][kv] so KV_TILE is contiguous (SIMD dim)
+            // Zero-pad the last tile so the GEMM always operates on KV_TILE_SZ columns
+            memset(K_f32, 0, DK * KV_TILE_SZ * sizeof(float));
+            for (int tk = 0; tk < kv_tile; tk++) {
+                const char * k_data = (const char *)k->data + (ic + tk)*nbk1 + ik2*nbk2 + ik3*nbk3;
+                if (kv_type == GGML_TYPE_F16) {
+                    const ggml_fp16_t * k_f16 = (const ggml_fp16_t *)k_data;
+                    for (int64_t dk = 0; dk < DK; dk++) {
+                        K_f32[dk * KV_TILE_SZ + tk] = GGML_CPU_FP16_TO_FP32(k_f16[dk]);
+                    }
+                } else {
+                    const float * k_f32_src = (const float *)k_data;
+                    for (int64_t dk = 0; dk < DK; dk++) {
+                        K_f32[dk * KV_TILE_SZ + tk] = k_f32_src[dk];
+                    }
+                }
+            }
+            memset(KQ, 0, Q_TILE_SZ * KV_TILE_SZ * sizeof(float));
+            simd_gemm(KQ, (const float *)Q_q, K_f32, Q_TILE_SZ, DK, KV_TILE_SZ);
+            ggml_vec_scale_f32(Q_TILE_SZ * KV_TILE_SZ, KQ, scale);
+
+            // Set padded KQ entries to -inf so softmax gives them zero weight
+            if (kv_tile < KV_TILE_SZ) {
+                for (int tq = 0; tq < Q_TILE_SZ; tq++) {
+                    for (int tk = kv_tile; tk < KV_TILE_SZ; tk++) {
+                        KQ[tq * KV_TILE_SZ + tk] = -INFINITY;
+                    }
                 }
             }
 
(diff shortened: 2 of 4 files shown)


ggml/src/ggml-cpu/ops.cpp
```diff
@@ -8476,42 +8473,69 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
...
-            for (int tq = 0; tq < Q_TILE_SZ; tq++) {
-                const void * q_row = (const char *)Q_q + tq * DK * kv_type_size;
-                for (int tk = 0; tk < KV_TILE_SZ; tk++) {
-                    const void * k_row = (const char *) k->data + ((ic + tk)*nbk1 + ik2*nbk2 + ik3*nbk3);
-                    float s;
-                    kv_vec_dot(DK, &s, 0, k_row, 0, q_row, 0, 1);
-                    KQ[tq * KV_TILE_SZ + tk] = s * scale;
+            // Pack K tile transposed: K_f32[dk][kv] so KV_TILE is contiguous (SIMD dim)
+            // Zero-pad the last tile so the GEMM always operates on KV_TILE_SZ columns
+            memset(K_f32, 0, DK * KV_TILE_SZ * sizeof(float));
```
ggerganov: I think you can move these memsets to be performed once prior the `ic` loop?

am17an: just this one, for KQ I need to clear the array after every step

ggerganov:
Hm, it looks like the "CPU reference" mechanism that we implemented in #19209 does not actually work as I thought it would. I was thinking that we can run:

```bash
test-backend-ops -b CPU
```

And it would compare the reference vs non-reference CPU implementation. But this is not the case, because the `ggml_backend_cpu_set_use_ref()` applies to both `backend1` and `backend2` - they are the same backends.

How do you test the non-reference implementation against the reference?

am17an:
That's what I use and it correctly fails if there is some bug. For example on master if I do something like this 
```cpp
diff --git a/ggml/src/ggml-cpu/ops.cpp b/ggml/src/ggml-cpu/ops.cpp
index ed4535020..6223b202a 100644
--- a/ggml/src/ggml-cpu/ops.cpp
+++ b/ggml/src/ggml-cpu/ops.cpp
@@ -8247,7 +8247,7 @@ static void ggml_compute_forward_flash_attn_ext_f16_one_chunk(
 
             if (s > M) {
                 ms = expf(M - s);
-                M = s;
+                //M = s;
                 ggml_vec_scale_f32(DV, VKQ32, ms);
             } else {
                 vs = expf(s - M);
```

I get failures of the form 

`[FLASH_ATTN_EXT] ERR = 0.135610349 > 0.000500000   FLASH_ATTN_EXT(hsk=40,hsv=40,nh=4,nr23=[4,1],kv=512,nb=1,mask=0,sinks=1,max_bias=0.000000,logit_softcap=0.000000,prec=f32,type_KV=f16,permute=[0,1,2,3]): FAIL`

when running `test-backend-ops -b CPU`

When I print pointers in `make_test_cases_eval`, I get different pointers for `backend` and `backend_cpu`

ggerganov:
Ah correct - this works as expected.

The problem is that the tiled version is never exercised atm. `use_tiled` is always `false` because in all tests the batch size is less than `Q_TILE_SZ`.

am17an:
Ah I see, you're right. So I can change in `test-backend-ops`:

```
-                                            for (int nb : { 1, 3, 32, 35, }) {
+                                            for (int nb : { 1, 3, 32, 65, }) {
```
To exercise this path, though would have to be careful when tuning Q_TILE_SZ

ggml/src/ggml-cpu/ops.cpp
```diff
@@ -8551,33 +8576,23 @@ static void ggml_compute_forward_flash_attn_ext_tiled(
...
-                for (int tq = 0; tq < Q_TILE_SZ; tq++) {
-                    if (skip[tq]) continue;
-                    float * vkq_row = VKQ32 + tq * DV;
-                    for (int tk = 0; tk < KV_TILE_SZ; tk++) {
-                        const float p = KQ[tq * KV_TILE_SZ + tk];
-                        ggml_vec_mad_f32(DV, vkq_row, V32 + tk * DV, p);
-                    }
+            // V accumulation: VKQ32 += softmax(KQ) * V
+            // Pack V tile to contiguous F32, zero-padded
+            memset(V32, 0, KV_TILE_SZ * DV * sizeof(float));
```
ggerganov: This `memset` can also be moved outside the loop.

ggml/src/ggml-cpu/simd-gemm.h
```diff
@@ -0,0 +1,130 @@
...
+
+// TODO: add support for sizeless vector types
+#if defined(GGML_SIMD) && !defined(__ARM_FEATURE_SVE) && !defined(__riscv_v_intrinsic)
+
+// TODO: untested on avx512 and arm
+// These are in units of GGML_F32_EPR
+#if defined(__AVX512F__) || defined (__ARM_NEON__)
+    static constexpr int GEMM_RM = 6;
+    static constexpr int GEMM_RN = 4; // 24+4+1 = 29/32
+#elif defined(__AVX2__) || defined(__AVX__)
```
ggerganov, line 18: On M2 Ultra, I get better results using:

```bash
# GGML_F32_EPR = 4

GEMM_RM = 4
GEMM_RN = 4
```

Here is before and after:

- 6x4

| Model                 | Test        |   t/s master |   t/s pr/19422 |   Speedup |
|:----------------------|:------------|-------------:|---------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512@d1024 |       166.40 |         177.12 |      1.06 |
| gpt-oss 20B MXFP4 MoE | pp512@d4096 |       134.16 |         155.12 |      1.16 |
| qwen2 1.5B F16        | pp512@d1024 |       298.68 |         307.97 |      1.03 |
| qwen2 1.5B F16        | pp512@d4096 |       226.77 |         247.86 |      1.09 |
| qwen2 3B F16          | pp512@d1024 |       143.64 |         151.70 |      1.06 |
| qwen2 3B F16          | pp512@d4096 |       112.81 |         131.39 |      1.16 |

- 4x4

| Model                 | Test        |   t/s master |   t/s pr/19422 |   Speedup |
|:----------------------|:------------|-------------:|---------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512@d1024 |       167.38 |         178.00 |      1.06 |
| gpt-oss 20B MXFP4 MoE | pp512@d4096 |       134.28 |         160.09 |      1.19 |
| qwen2 1.5B F16        | pp512@d1024 |       299.04 |         313.51 |      1.05 |
| qwen2 1.5B F16        | pp512@d4096 |       226.24 |         259.23 |      1.15 |
| qwen2 3B F16          | pp512@d1024 |       144.24 |         152.88 |      1.06 |
| qwen2 3B F16          | pp512@d4096 |       113.59 |         135.72 |      1.19 |

ggml/src/ggml-cpu/simd-gemm.h
```diff
@@ -0,0 +1,130 @@
+#pragma once
+
+// Computes C[M x N] += A[M x K] * B[K x N]
+
+#include "ggml-cpu-impl.h"
+#include "vec.h"
+#include "common.h"
```
ggerganov, line 7: I think you need just `simg-mappings.h` here

ggerganov (approved):

ggml/src/ggml-cpu/simd-gemm.h
am17an: Thanks for testing, can you try one with a very high depth like 16384, that's where it would be really clear

am17an: Also for me, after this PR FA=1 is always faster than FA=0 at least for PP. For TG the results are better if there are more threads. I guess our GEMV implementation can be improved

ggerganov: Here is comparison of 6x4 (current) vs 4x4 (new):

| Model                 | Test         |   t/s 8debab310 |   t/s dce1b0911 |   Speedup |
|:----------------------|:-------------|----------------:|----------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512@d1024  |          176.48 |          178.33 |      1.01 |
| gpt-oss 20B MXFP4 MoE | pp512@d4096  |          156.05 |          159.91 |      1.02 |
| gpt-oss 20B MXFP4 MoE | pp512@d16384 |          106.38 |          113.16 |      1.06 |
| qwen2 1.5B F16        | pp512@d1024  |          307.25 |          313.73 |      1.02 |
| qwen2 1.5B F16        | pp512@d4096  |          247.65 |          258.67 |      1.04 |
| qwen2 1.5B F16        | pp512@d16384 |          136.64 |          153.64 |      1.12 |
| qwen2 3B F16          | pp512@d1024  |          151.11 |          152.89 |      1.01 |
| qwen2 3B F16          | pp512@d4096  |          130.56 |          133.45 |      1.02 |
| qwen2 3B F16          | pp512@d16384 |           85.33 |           91.88 |      1.08 |

I.e. it's better also for 16k context.

ggml/src/ggml-cpu/simd-gemm.h
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+}
+
+// C[M x N] += A[M x K] * B[K x N]
+static void simd_gemm(
+    float       * GGML_RESTRICT C,
+    const float * GGML_RESTRICT A,
+    const float * GGML_RESTRICT B,
+    int M, int64_t K, int64_t N)
```
ggerganov: The int types are a bit inconsistent. Either use `int` or `int64_t` everywhere:

```suggestion
    int64_t M, int64_t K, int64_t N)
```

ggml/src/ggml-cpu/simd-gemm.h
```diff
@@ -0,0 +1,139 @@
...
+    static constexpr int GEMM_RN = 2; // 12+2+1 = 15/16
+#else
+    static constexpr int GEMM_RM = 2;
+    static constexpr int GEMM_RN = 2;
+#endif
+
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Waggressive-loop-optimizations"
+#endif
```
ggerganov, line 29: I noticed the [compile warnings](https://github.com/ggml-org/llama.cpp/actions/runs/21982329215/job/63507789161#step:3:893) in the CI. Are we confident these are false-positives and safe to ignore?

am17an: Yes, it was complaining about overflowing iteration when ii is 2^58

ggml/src/ggml-cpu/simd-gemm.h
```diff
@@ -0,0 +1,139 @@
...
+#include "simd-mappings.h"
+
+// TODO: add support for sizeless vector types
+#if defined(GGML_SIMD) && !defined(__ARM_FEATURE_SVE) && !defined(__riscv_v_intrinsic)
+
+// TODO: untested on avx512
+// These are in units of GGML_F32_EPR
+#if defined(__AVX512F__) || defined (__ARM_NEON__)
+    static constexpr int GEMM_RM = 4;
+    static constexpr int GEMM_RN = 4; // 16+4+1 = 25/32
```
Djip007, line 17: you can try
```cpp
static constexpr int GEMM_RM = 6;
static constexpr int GEMM_RN = 4;   // 24 + 4 + 1 = 29 ...
```
that is what is used by AMD.

ggml/src/ggml-cpu/simd-gemm.h
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    for (int64_t kk = 0; kk < K; kk++) {
+        GGML_F32_VEC Bv[RN];
+        for (int r = 0; r < RN; r++) {
+            Bv[r] = GGML_F32_VEC_LOAD(B + kk * N + jj + r * KN);
+        }
+        for (int i = 0; i < RM; i++) {
+            GGML_F32_VEC p = GGML_F32_VEC_SET1(A[(ii + i) * K + kk]);
```
Djip007, line 54: this is nice for x86 to do it like that.

but on ARM (neon) I remember that there is some OP:
`regC += regB * regA[i]`  for FMA
it is possible to load a full register for A. 
for FP16 : with 32 register on neon you can have:
- RN = 1
- RM = 16  // => 1 vector load

for fp32 it may be
1x8  / 2x8 / 1x16

[edit] but may need some transpose on A for that.

ggml/src/ggml-cpu/simd-gemm.h
am17an: It was this before https://github.com/ggml-org/llama.cpp/pull/19422#discussion_r2803075159, but changed for ARM_NEON, will create a separate branch for AVX512

Djip007:
```cpp
#if defined(__AVX512F__)
    static constexpr int GEMM_RM = 4;
    static constexpr int GEMM_RN = 6; // 24+4+2 = 30/32  (+2 for pre-load)
#elif defined (__ARM_NEON__)
    static constexpr int GEMM_RM = 4;
    static constexpr int GEMM_RN = 4; // 16+4+1 = 25/32

```

ggml/src/ggml-cpu/common.h
```diff
@@ -6,8 +6,8 @@
 #include "ggml-impl.h"
 #include "simd-mappings.h"
 
-#define GGML_FA_TILE_Q  32
-#define GGML_FA_TILE_KV 16
+#define GGML_FA_TILE_Q  64
+#define GGML_FA_TILE_KV 64
```
Djip007, line 10: did this not need to be adjust with GEMM_RM/GEMM_RN size?

but I don't know if it is related to GEMM_RM or GEMM_RN (or both?)

ggml/src/ggml-cpu/simd-gemm.h
```diff
@@ -0,0 +1,139 @@
...
+
+    // Tail rows: one at a time
+    for (; ii < M; ii++) {
+        int64_t jj = 0;
+        for (; jj + GEMM_RN * KN <= N; jj += GEMM_RN * KN) {
+            simd_gemm_ukernel<1, GEMM_RN>(C, A, B, K, N, ii, jj);
+        }
+        for (; jj + KN <= N; jj += KN) {
+            simd_gemm_ukernel<1, 1>(C, A, B, K, N, ii, jj);
+        }
```
Djip007, line 105: for some gain (or not...) 
tinyblas use some encoding  0xMN to have a switch on "all" MxN possible
so we can use some <2,GEMM_RN> ... 

https://github.com/ggml-org/llama.cpp/blob/184c694f454d8b5f241f79aec3cc9fdfb6ddb824/ggml/src/ggml-cpu/llamafile/sgemm.cpp#L1224

but yes it need more dev / time / ...

ggml/src/ggml-cpu/common.h
am17an: It should be adjusted acc to ISA available, I just tuned on Zen 2 (AMD Rome) since that's the only one I have available. I think scratch space should be close to L1 cache size so maybe that's one factor to tune this

Djip007: yes, but I think something else: it is bloc of bloc so
```cpp
GGML_FA_TILE_Q   = 16*GEMM_RM;  // or 16*GEMM_RN 
GGML_FA_TILE_KV = 16*GEMM_RM;  // or 16*GEMM_RN
```
so we use at most the most efficient uGEMM blick size ?

Note: I did not take time to look it GGML_FA_TILE_Q and GGML_FA_TILE_KV is related to GEMM_RM or GEMM_RN:

And yes for L1/L2/L3 cache size, but to be best we need block on K not only NxM

am17an: Right, I guess for AVX2 it works out because GEMM_RM=4, 4*16 = 64. Let me try to tune it according to this and see the difference

Djip007: what bench cmd did you use ?

am17an: `-m llama_8b_q4_0.gguf,gpt-oss-20b.mxfp4 -fa 1 -d 0,8196,16348 -t 16,32,64 -n 0 -p 512`

Djip007: no `-ctk f32 -ctv f32 ` is needed ?

am17an: No because it converts for f32 from f16

Djip007:
I did some bench with zen5x16  (a IA-MAX+ 395)

- the AOCL use aocl AMD hight optimized gemm full BF16.
- FA-0_BF16 use std tinyblas with KV as BF16 no FA
- FA-0_FP16 use std tinyblas with KV as FP16 no FA
les nexts use FA with FP16 KV: I test some TILE size :
 GGML_FA_TILE_Q/GGML_FA_TILE_KV  GEMM_RMxGEMM_RN

| model           |            test |   AOCL_BF16 t/s |    FA-0_BF16 t/s |   FA-0_FP16  t/s |   64/64 6x4  t/s |  64/64 4x4 t/s |   48/64 6x4  t/s |  480/128 6x4 t/s |  300x64 6x4 t/s | 384/64 4x4 t/s |
| ----------------| --------------: | --------------: |----------------: | ---------------: | ---------------: | -------------: | ---------------: | ---------------: | --------------: | -------------: |
| llama 8B BF16   |           pp512 |   428.27 ± 6.89 |    277.67 ± 0.20 |    268.62 ± 0.12 |    300.69 ± 0.20 |  298.92 ± 0.06 |    295.09 ± 0.13 |    293.63 ± 1.13 |   300.13 ± 0.28 |  295.73 ± 0.13 |
| llama 8B BF16   |   pp512 @ d2048 |   348.87 ± 0.13 |    225.60 ± 0.24 |    215.64 ± 0.07 |                  |  242.19 ± 0.44 |                  |    224.33 ± 0.39 |   253.65 ± 0.55 |  244.22 ± 0.16 |
| llama 8B BF16   |   pp512 @ d4096 |   289.80 ± 1.57 |    185.74 ± 0.33 |    177.57 ± 0.82 |    204.15 ± 0.20 |  206.08 ± 0.13 |    196.40 ± 0.32 |    184.88 ± 0.83 |   220.96 ± 0.46 |  208.83 ± 0.56 |
| llama 8B BF16   |   pp512 @ d8192 |   231.30 ± 1.11 |    141.81 ± 0.51 |    133.66 ± 0.06 |    125.32 ± 0.21 |  127.95 ± 0.10 |    124.05 ± 0.75 |    136.39 ± 0.25 |   171.49 ± 0.14 |  156.20 ± 0.01 |
| llama 8B BF16   |  pp512 @ d16384 |   161.54 ± 0.89 |     97.82 ± 0.40 |     86.70 ± 0.21 |     75.05 ± 0.08 |   78.68 ± 0.01 |     76.93 ± 0.47 |     90.11 ± 0.82 |   116.39 ± 0.19 |  105.02 ± 0.13 |
| llama 8B BF16   |  pp512 @ d32768 |    98.66 ± 0.80 |     53.82 ± 0.10 |     48.78 ± 0.00 |                  |   43.92 ± 0.11 |                  |     50.61 ± 0.16 |    72.62 ± 0.21 |   62.83 ± 0.24 |

For now on this my zen5 with llama 8B the faster I have is:
```
GGML_FA_TILE_Q 300
GGML_FA_TILE_KV 64
GEMM_RM 6
GEMM_RN 4
```

If I can find some time I like to "hack" your FA/gemm for BF16 support....

Do you think (or @ggerganov ) it is possible to store the KV-cache packed? so we don"t need to do full repack in FA ?

ggerganov:
@Djip007 The primary limitation for FA modifications is the code to remain easy to understand and test.

> Do you think (or @ggerganov ) it is possible to store the KV-cache packed? so we don"t need to do full repack in FA ?

Not sure, haven't thought about this. My feeling is even if it is possible, it wouldn't be worth the extra complexity.

Djip007:
> My feeling is even if it is possible, it wouldn't be worth the extra complexity.

I completely agree with that, and I saw your reservations about the testing part.

I haven't thought about it very long, so it might not be possible:

We could extend the extra buffers to handle KV in addition to weight cases.
This would allow us to keep the simple case in the basic CPU backend and add optimized versions of the FA, as was done for GEMM.
Well, we might also need to extend the tests to work in these cases.

It might not be a good idea to discuss on this pull request. Should I start a new discussion on this topic?

And it may not be possible or very complicated :sunglasses:

PR #19426, opened 2026-02-08
fix vulkan ggml_acc only works in 3d but not 4d

ymcki:

Discovered ggml_acc for vulkan only works in 3d not 4d while working on 
https://github.com/ggml-org/llama.cpp/pull/18792

So fixed it and also added test case to test-backend-ops


tests/test-backend-ops.cpp
```diff
@@ -5852,6 +5852,63 @@ struct test_acc : public test_case {
...
+        : type(type), block_size(block_size), n_blocks(n_blocks), ne2(ne2), ne3(ne3) {}
+
+    ggml_tensor * build_graph(ggml_context * ctx) override {
+        const int64_t chunk_size = block_size * n_blocks;
+
+        // Base tensor initialized to zero using ggml_clamp
+        ggml_tensor * a = ggml_new_tensor_4d(ctx, type, chunk_size, chunk_size, ne2, ne3);
+        ggml_set_param(a);
+        ggml_set_name(a, "a");
+        ggml_tensor * acc = ggml_clamp(ctx, a, 0.0f, 0.0f);
```
ggerganov: This clamp can be removed

ymcki: removed

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
```diff
@@ -13,17 +13,19 @@ void main() {
...
     const uint src1_i = idx - offset;
-    const uint oz = src1_i / p.nb02;
-    const uint oy = (src1_i - (oz * p.nb02)) / p.nb01;
-    const uint ox = src1_i % p.nb01;
+    const uint oz = src1_i / p.nb03;
+    const uint remy = src1_i - oz * p.nb03;
+    const uint oy = remy / p.nb02;
+    const uint remx = remy - oy * p.nb02;
+    const uint ox = remx / p.nb01;
+    const uint ow = remx % p.nb01;
```
jeffbolznv: Generally in GLSL code, you'd expect 'w' to be the fourth coordinate (xyzw). I suggest renaming to i0/i1/i2/i3 to avoid confusion.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
```diff
@@ -13,17 +13,19 @@ void main() {
 
     const uint offset = p.param3;
     const uint src1_i = idx - offset;
-    const uint oz = src1_i / p.nb02;
-    const uint oy = (src1_i - (oz * p.nb02)) / p.nb01;
-    const uint ox = src1_i % p.nb01;
+    const uint oz = src1_i / p.nb03;
```
jeffbolznv: Dividing by the 'b' strides doesn't seem right, but it's possible I'm missing some subtlety of the acc op. Shouldn't it divide by ne00-ne03 and then use the strides just for the addressing calculation?

I also noticed that ggml_vk_acc is not using the nb03 from the parameters, that seems like it needs to be updated.

tests/test-backend-ops.cpp
```diff
@@ -5852,6 +5852,62 @@ struct test_acc : public test_case {
     }
 };
 
+// GGML_OP_ACC - block accumulation test
+struct test_acc_block: public test_case {
```
jeffbolznv: I don't really see why a new test class is needed, rather than extending test_acc. And it seems like what's really needed is more variation on the nb parameters, which this class isn't doing.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
jeffbolznv: I see from the CPU reference that the params are used for the shape of src and dst, so this is sort of right. But it still needs the real nb3, and the b addressing should use the b strides.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
ymcki: So do you mean we change (ow,ox,oy,oz) to (i0,i1,i2,i3) instead of (ox,oy,oz,ow)?

If vulkan code is mostly coded by one person, then doing the former will require changing many files. So probably the person who is now in charge of vulkan code should make this change.

If the latter is ok with you, I can do it in the next commit.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
ymcki: Do you mean ggml_vk_acc is also wrong in the same vein? But when is ggml_vk_acc going to be called?

tests/test-backend-ops.cpp
ymcki: I find that by setting the 3rd and 4th dim of test_acc, I can reproduce the failure as expected. Then we can remove the new test_acc_block class. Is that what you want?

```
// GGML_OP_ACC
struct test_acc : public test_case {
    const ggml_type type;
    const std::array<int64_t, 4> ne_a;
    const std::array<int64_t, 4> ne_b;

    std::string vars() override {
        return VARS_TO_STR3(type, ne_a, ne_b);
    }

    test_acc(ggml_type type = GGML_TYPE_F32,
            std::array<int64_t, 4> ne_a = {256, 17, 2, 3},
            std::array<int64_t, 4> ne_b = {256, 16, 2, 3})
        : type(type), ne_a(ne_a), ne_b(ne_b) {}

    ggml_tensor * build_graph(ggml_context * ctx) override {
        ggml_tensor * a = ggml_new_tensor(ctx, type, 4, ne_a.data());
        ggml_set_param(a);
        ggml_set_name(a, "a");

        ggml_tensor * b = ggml_new_tensor(ctx, type, 4, ne_b.data());
        ggml_set_param(b);
        ggml_set_name(b, "b");

        ggml_tensor * out = ggml_acc(ctx, a, b, a->nb[1], a->nb[2], a->nb[3], b->nb[1]);
        ggml_set_name(out, "out");

        return out;
    }
};
```

jeffbolznv: Yes, but we should also test b being a view, and parameters to ggml_acc that don't match a's stride. I think it'll find more bugs.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
jeffbolznv: Yes, the 'o' naming is uncommon. Generally it should be i0-i3 for the destination, i00-i03 for src0, i10-i13 for src1, and so on.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
jeffbolznv: ggml_vk_acc is what invokes this shader. I don't think nb3 should be unused, and I think if you add the tests I suggest this would show up as a bug.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
ymcki: Interestingly, only this acc.comp uses ox,oy,oz, so I will change to i0,..,i3 then.

tests/test-backend-ops.cpp
ymcki: I am getting new failures after extending the test_acc class based on your suggestion.

```
[ACC] ERR = 0.803344802 > 0.000000100   ACC(type=f32,ne_a=[256,17,2,3],ne_b=[256,16,2,3],b_is_view=1): FAIL
[ACC] ERR = 0.524563173 > 0.000000100   ACC(type=f32,ne_a=[256,17,2,3],ne_b=[128,16,2,3],b_is_view=1): FAIL
```

```
// GGML_OP_ACC
struct test_acc : public test_case {
    const ggml_type type;
    const std::array<int64_t, 4> ne_a;
    const std::array<int64_t, 4> ne_b;
    const bool b_is_view;

    std::string vars() override {
        return VARS_TO_STR4(type, ne_a, ne_b, b_is_view);
    }

    test_acc(ggml_type type = GGML_TYPE_F32,
            std::array<int64_t, 4> ne_a = {256, 17, 2, 3},
            std::array<int64_t, 4> ne_b = {256, 16, 2, 3},
            bool b_is_view = false)
        : type(type), ne_a(ne_a), ne_b(ne_b), b_is_view(b_is_view) {}

    ggml_tensor * build_graph(ggml_context * ctx) override {
        ggml_tensor * a = ggml_new_tensor(ctx, type, 4, ne_a.data());
        ggml_set_param(a);
        ggml_set_name(a, "a");

        ggml_tensor * b;
        if (b_is_view) {
            // Create a larger tensor and take a view at a non-zero offset.
            // This tests that the backend correctly handles b's data offset
            std::array<int64_t, 4> ne_b_pad = {ne_b[0], ne_b[1] + 1, ne_b[2], ne_b[3]};
            ggml_tensor * b_pad = ggml_new_tensor(ctx, type, 4, ne_b_pad.data());
            ggml_set_param(b_pad);
            ggml_set_name(b_pad, "b_pad");
            // View that skips the first row, so b has a non-zero byte offset
            b = ggml_view_4d(ctx, b_pad,
                ne_b[0], ne_b[1], ne_b[2], ne_b[3],
                b_pad->nb[1], b_pad->nb[2], b_pad->nb[3],
                b_pad->nb[1]);
        } else {
            b = ggml_new_tensor(ctx, type, 4, ne_b.data());
            ggml_set_param(b);
        }
        ggml_set_name(b, "b");

        // When ne_b[0] < ne_a[0], a->nb[1] != b->nb[1], so the stride
        // parameters to ggml_acc don't match b's natural stride.
        ggml_tensor * out = ggml_acc(ctx, a, b, a->nb[1], a->nb[2], a->nb[3], 0);
        ggml_set_name(out, "out");

        return out;
    }
};

```

Does this new class look ok to you? Does that mean now I should fix ggml_vk_acc?

jeffbolznv: I think you should pad b's dim 2 and 3 as well. Otherwise, yeah, it looks ok. You'll need to fix ggml_vk_acc and also make sure the correct strides are used in the shader.

ymcki: Extended to 2 and 3. 2 is a failure but 3 is ok. Is this expected?

```
[ACC] ERR = 0.801934959 > 0.000000100   ACC(type=f32,ne_a=[256,17,2,3],ne_b=[256,16,2,3],stride_dim=1): FAIL
[ACC] ERR = 0.436916859 > 0.000000100   ACC(type=f32,ne_a=[256,17,2,3],ne_b=[128,16,2,3],stride_dim=2): FAIL
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[64,16,2,3],stride_dim=3): OK
```

```
// GGML_OP_ACC
struct test_acc2 : public test_case {
    const ggml_type type;
    const std::array<int64_t, 4> ne_a;
    const std::array<int64_t, 4> ne_b;
    const int64_t stride_dim;

    std::string vars() override {
        return VARS_TO_STR4(type, ne_a, ne_b, stride_dim);
    }

    test_acc2(ggml_type type = GGML_TYPE_F32,
            std::array<int64_t, 4> ne_a = {256, 17, 2, 3},
            std::array<int64_t, 4> ne_b = {256, 16, 2, 3},
            uint64_t stride_dim = -1)
        : type(type), ne_a(ne_a), ne_b(ne_b), stride_dim(stride_dim) {}

    ggml_tensor * build_graph(ggml_context * ctx) override {
        ggml_tensor * a = ggml_new_tensor(ctx, type, 4, ne_a.data());
        ggml_set_param(a);
        ggml_set_name(a, "a");

        ggml_tensor * b;
        if (stride_dim == 1 || stride_dim == 2 || stride_dim == 3) {
            // Create a larger tensor and take a view at a non-zero offset.
            // This tests that the backend correctly handles b's data offset
            std::array<int64_t, 4> ne_b_pad = {ne_b[0], ne_b[1], ne_b[2], ne_b[3]};
            ne_b_pad[stride_dim] += 1;
            ggml_tensor * b_pad = ggml_new_tensor(ctx, type, 4, ne_b_pad.data());
            ggml_set_param(b_pad);
            ggml_set_name(b_pad, "b_pad");
            // View that skips the first row, so b has a non-zero byte offset
            b = ggml_view_4d(ctx, b_pad,
                ne_b[0], ne_b[1], ne_b[2], ne_b[3],
                b_pad->nb[1], b_pad->nb[2], b_pad->nb[3],
                b_pad->nb[1]);
        } else {
            b = ggml_new_tensor(ctx, type, 4, ne_b.data());
            ggml_set_param(b);
        }
        ggml_set_name(b, "b");

        // When ne_b[0] < ne_a[0], a->nb[1] != b->nb[1], so the stride
        // parameters to ggml_acc don't match b's natural stride.
        ggml_tensor * out = ggml_acc(ctx, a, b, a->nb[1], a->nb[2], a->nb[3], 0);
        ggml_set_name(out, "out");

        return out;
    }
};
```

ymcki:
The latest commit seems to fix the potential problems pointed out by Jeff

ggerganov:
The metal failures will be resolved with #19427 after merging this PR.

ggerganov:
Added TODO for the CUDA implementation to support this

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
```diff
@@ -13,17 +13,19 @@ void main() {
...
+    const uint i1 = rem1 / p.nb01;
+    const uint i0 = rem1 % p.nb01;
 
     uint i00, i01, i02, i03;
     get_indices(idx, i00, i01, i02, i03);
 
-    if (ox < p.ne10 && oy < p.ne11 && oz < p.ne12) {
-        data_d[get_doffset() + dst_idx(i00, i01, i02, i03)] = D_TYPE(FLOAT_TYPE(data_a[get_aoffset() + src0_idx(i00, i01, i02, i03)]) + FLOAT_TYPE(data_b[get_boffset() + ox + oy * p.ne10 + oz * p.ne10 * p.ne11]));
+    if (i0 < p.ne10 && i1 < p.ne11 && i2 < p.ne12 && i3 < p.ne13) {
+        data_d[get_doffset() + dst_idx(i00, i01, i02, i03)] = D_TYPE(FLOAT_TYPE(data_a[get_aoffset() + src0_idx(i00, i01, i02, i03)]) + FLOAT_TYPE(data_b[get_boffset() + i0 + i1 * p.nb11 + i2 * p.nb12 + i3 * p.nb13]));
```
jeffbolznv: You could use `src1_idx` here and it would also handle permuted tensors (by applying nb10). Would be nice to test permuted tensors, too.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
```diff
@@ -13,17 +13,19 @@ void main() {
...
-    const uint ox = src1_i % p.nb01;
+    const uint i3 = src1_i / p.nb03;
+    const uint rem2 = src1_i - i3 * p.nb03;
+    const uint i2 = rem2 / p.nb02;
+    const uint rem1 = rem2 - i2 * p.nb02;
+    const uint i1 = rem1 / p.nb01;
+    const uint i0 = rem1 % p.nb01;
 
     uint i00, i01, i02, i03;
     get_indices(idx, i00, i01, i02, i03);
```
jeffbolznv: Looking at the CPU reference, seems like there should only be one set of indices and then you can apply `offset/sizeof(float)` to the final index for src0 and dst.

ggml/src/ggml-vulkan/vulkan-shaders/acc.comp
```diff
@@ -13,17 +13,19 @@ void main() {
...
 
-    if (ox < p.ne10 && oy < p.ne11 && oz < p.ne12) {
-        data_d[get_doffset() + dst_idx(i00, i01, i02, i03)] = D_TYPE(FLOAT_TYPE(data_a[get_aoffset() + src0_idx(i00, i01, i02, i03)]) + FLOAT_TYPE(data_b[get_boffset() + ox + oy * p.ne10 + oz * p.ne10 * p.ne11]));
+    if (i0 < p.ne10 && i1 < p.ne11 && i2 < p.ne12 && i3 < p.ne13) {
```
jeffbolznv, line 25: I think this bounds check may be unnecessary - the check against p.ne should cover it.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -9801,16 +9801,16 @@ static void ggml_vk_acc(ggml_backend_vk_context * ctx, vk_context& subctx, const
...
-    int nb2 = dst->op_params[1] / 4; // 4 bytes of float32
-    // int nb3 = dst->op_params[2] / 4; // 4 bytes of float32 - unused
-    int offset = dst->op_params[3] / 4; // offset in bytes
+    int nb1 = dst->op_params[0] / src0_type_size; // 4 bytes of float32
+    int nb2 = dst->op_params[1] / src0_type_size; // 4 bytes of float32
+    int nb3 = dst->op_params[2] / src0_type_size; // 4 bytes of float32
+    int offset = dst->op_params[3] / src0_type_size; // offset in bytes
 
     ggml_vk_op_f32<vk_op_binary_push_constants>(ctx, subctx, src0, src1, nullptr, nullptr, dst, GGML_OP_ACC, {
         (uint32_t)ggml_nelements(src0),
```
jeffbolznv, line 9810: I think this should be src1?

ymcki:
Latest commit should address Jeff's suggestions except for the boundary check. Without the boundary check, the code simply doesn't pass the tests.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
jeffbolznv: My suggestion was to change line 9810, not 9807. I think this change is needed to make it possible to remove the bound check.

ymcki: I just tried changing 9810 but not 9807 and remove boundary check, it still fails.

Changing 9810 not 9807 and keep boundary check, it also fails.

ymcki:
Committed a version without the check that passes my tests and measured the performance by adding my test cases to test_cases_perf

```
  ACC(type=f32,ne_a=[256,17,1,1],ne_b=[256,16,1,1],stride_dim=-1):             73710 runs -    14.36 us/run -       50 kB/run -    3.32 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[256,16,2,3],stride_dim=-1):             81900 runs -    12.65 us/run -      300 kB/run -   22.61 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[128,16,2,3],stride_dim=-1):             81900 runs -    12.72 us/run -      252 kB/run -   18.89 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[256,16,2,3],stride_dim=1):              81890 runs -    12.64 us/run -      305 kB/run -   23.01 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[128,16,2,3],stride_dim=2):              81890 runs -    12.67 us/run -      268 kB/run -   20.18 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[64,16,2,3],stride_dim=3):               81890 runs -    12.75 us/run -      228 kB/run -   17.05 GB/s
```

However, it seems to be significantly slower than previous code:

```
  ACC(type=f32,ne_a=[256,17,1,1],ne_b=[256,16,1,1],stride_dim=-1):            114660 runs -     9.21 us/run -       50 kB/run -    5.18 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[256,16,2,3],stride_dim=-1):            122850 runs -     8.44 us/run -      300 kB/run -   33.89 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[128,16,2,3],stride_dim=-1):            122850 runs -     8.53 us/run -      252 kB/run -   28.19 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[256,16,2,3],stride_dim=1):             122835 runs -     8.54 us/run -      305 kB/run -   34.05 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[128,16,2,3],stride_dim=2):             122835 runs -     8.56 us/run -      268 kB/run -   29.88 GB/s
  ACC(type=f32,ne_a=[256,17,2,3],ne_b=[64,16,2,3],stride_dim=3):              122835 runs -     8.40 us/run -      228 kB/run -   25.88 GB/s
```

I think it is better to revert to the previous boundary check version unless Jeff knows how to speed it up.

jeffbolznv:
Sorry, yes, it's better to do it in a single shader and have the branch. I somehow missed that there was an else case and that it was an "inside vs outside" check rather than just an out of bounds check.

ymcki:
Back to boundary check version but keep the test cases added in perf.

jeffbolznv (approved):

ggerganov (approved):

0cc4m (approved):
LGTM

PR #19170, opened 2026-01-29
Add Kimi-K2.5 support

AesSedai:
Adding support for https://huggingface.co/moonshotai/Kimi-K2.5

Since this model includes compressed-tensors (INT4 for the conditional experts), I moved the `dequant_model` to the `prepare_tensors` call at @compilade's suggestion. The model conversion fails otherwise because the `quantization_config` is nested under the `text_config` in the config.json.

Additionally, this model adds some new keys for the vision tower, prefixed as `vt_`, and the preprocessor_config.json has the expected fields nested in the `media_proc_cfg` key.

This PR does not include the "hacked" Q4_0 changes by @jukofyork, referred to in [this comment](https://github.com/ggml-org/llama.cpp/pull/17064#issuecomment-3521098057).

I have added a first pass at vision support, heavily aided by LLM assistance. I entirely expect @ngxson to tear it to shreds or call me a dummy and show me an easier way to add that vision support :)

PR #19429, opened 2026-02-08
cuda : extend GGML_OP_PAD to work with non-cont src0

ggerganov:
- Extend CUDA support
- Remove redundant assert in CPU implementation
- Add permuted PAD tests


JohannesGaessler (approved):

JohannesGaessler:
Does this fix the issue described in https://github.com/ggml-org/llama.cpp/pull/19170#discussion_r2779956434 ?

ggerganov:
> Does this fix the issue described in https://github.com/ggml-org/llama.cpp/pull/19170#discussion_r2779956434 ?

No, this is needed to optimize out some of the copies in the Qwen3 Next graphs.

PR #19125, opened 2026-01-27
Unified Delta Net

ymcki:
*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

This should work with Kimi Linear.

Please test it with Qwen3Next

PR #16780, opened 2025-10-26
[model] add support for qwen3vl series

JJJYmmm:
This PR adds support for the Qwen3-VL series, including both the dense and MoE variants.
The original implementation was contributed by @yairpatch and @Thireus (see #16207). @LETS-BEE also helped address issues such as weights loading.

In this PR, I’ve fixed several algorithmic implementation details (e.g., deepstack), added support for MRoPE-Interleave, and performed final code cleanup.

PR #19435, opened 2026-02-08
[Model] Qwen3.5 dense and MoE support (no vision)

pwilkin:
I've gotten a bit tired of Llama.cpp missing all the zero-day releases, so this time I decided to make (or, more precisely, instructed Opus 4.6 to make, based on reference implementations and my guidelines for model adaptation) a conversion based on the Transformers PR ( https://github.com/huggingface/transformers/pull/43830/changes ). It's mostly based on Qwen3Next, but it's rebased on the common-delta-net PR ( #19125 ).

Here are the mock models I generated to test it: https://huggingface.co/ilintar/qwen35_testing/tree/main

Here are the conversion results from `causal-verify-logits`:

Model	NMSE	NMSE (dB)	Result
Dense	8.94e-06	-50.49 dB	Excellent
MoE	9.36e-05	-40.29 dB	Excellent


ggerganov:
> instructed Opus 4.6 to make, based on reference implementations and my guidelines for model adaptation

A bit off-topic, but I'm curious to try the same task with a local model. Maybe GLM 4.7 Flash + OpenCode. Do you have the guidelines that you used somewhere shareable?

pwilkin:
> A bit off-topic, but I'm curious to try the same task with a local model. Maybe GLM 4.7 Flash + OpenCode. Do you have the guidelines that you used somewhere shareable?

I'm using my "adding model architectures" tutorial for this actually (https://github.com/ggml-org/llama.cpp/discussions/16770) + got an extra rules section in my agent MD reminding about the tensor layout:

```
## Tensor format

A very important caveat about tensor format in the GGML library: the GGML library uses a different format internally than most Python implementations,
including PyTorch or Transformers. Notably, in the GGML library:

* tensors are restricted to 4 dimensions - whenever you want to use more dimensions you have to pack them
* the semantic order of dimensions is reversed from PyTorch/Transformers - the *last two* dimensions are used for `[tokens_per_batch, batches]` - even though the physical layout is the same

So, for example, a `[1, 5, 4096, 1]` tensor in PyTorch will probably become a `[1, 4096, 5, 1]` tensor in GGML. This is especially important
when converting certain implementations from a reference implementation written in Python, because the tensor dimension ordering will be different,
but they can be semantically the same.
```

The prompt I used this time:
```console
"In the transformers directory, I have included a new version of Transformers that includes support for the new Qwen3.5 series of models (MoE and dense). @transformers/src/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py @transformers/src/transformers/models/qwen3_5/modeling_qwen3_5.py 

I have also created a script and generated mock models for testing, the script is in @transformers/generate_qwen_models.py and has been already run.

Based on the implementation, which seems to be based strongly on the Qwen3 Next architecture, currently handled in Llama.cpp in @reference/qwen3next.cpp and in @src/models/delta.cpp, please create the implementation and conversion code for the new architectures. There are scripts in @examples/model-conversion/Makefile for testing conversions of new models. 

See @.roo/rules/adding_new_models.md for tips on how to add new model support. Make sure to reuse as much existing code as possible. For now, only add text support, without the multimodal capabilities."
```

EDIT: Key thing was also interrupting the model when it starts to do something stupid, for example modifying permutation rules in delta_net because it added incorrect extra permutations in conversion. Even with Opus 4.6 I had to do it like 3 or 4 times to stop it from going in the wrong direction.

Oh yeah, I also made some agents that I feel work well with Lllama.cpp for OpenCode (I've used them in fact for some documentation work), so I might share those too:

[deep-code-modifier.md](https://github.com/user-attachments/files/25164201/deep-code-modifier.md)
[code-architecture-analyzer.md](https://github.com/user-attachments/files/25164202/code-architecture-analyzer.md)

CISC:
> Oh yeah, I also made some agents that I feel work well with Lllama.cpp for OpenCode (I've used them in fact for some documentation work), so I might share those too:

/me sits back and waits for the automated PR submissions whenever a new model pops up...

ggerganov:
> > Oh yeah, I also made some agents that I feel work well with Lllama.cpp for OpenCode (I've used them in fact for some documentation work), so I might share those too:
> 
> /me sits back and waits for the automated PR submissions whenever a new model pops up...

github actions workflow, runs on self-hosted mac mini, accepts vllm/transformers PR as input

convert_hf_to_gguf.py
```diff
@@ -4344,6 +4344,73 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+            ba_combined = torch.cat([b_tensor, a_tensor], dim=0)
+            yield (self.format_tensor_name(gguf.MODEL_TENSOR.SSM_BETA_ALPHA, bid, ".weight"), ba_combined)
+            return
+
+        # Handle packed expert tensors (Qwen3.5 MoE uses nn.Linear convention)
+        # HF stores [n_expert, out_features, in_features] which already matches GGML
+        # expectations, so we skip Qwen2MoeModel's permute(0,2,1) and handle directly.
+        if name.endswith("mlp.experts.down_proj") or name.endswith("mlp.experts.down_proj.weight"):
+            mapped = f"{name}.weight" if not name.endswith(".weight") else name
+            # HF: [n_expert, n_embd, n_ff_exp] → GGML: {n_ff_exp, n_embd, n_expert} ✓
```
```suggestion
            # HF: [n_expert, n_embd, n_ff_exp] -> GGML: {n_ff_exp, n_embd, n_expert}
```
Emojis next? :)

convert_hf_to_gguf.py
```diff
@@ -4344,6 +4344,73 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+            yield from super(Qwen2MoeModel, self).modify_tensors(data_torch, mapped, bid)
+            return
+
+        if name.endswith("mlp.experts.gate_up_proj") or name.endswith("mlp.experts.gate_up_proj.weight"):
+            # HF: [n_expert, 2*n_ff_exp, n_embd] → split dim=1 → [n_expert, n_ff_exp, n_embd]
+            # GGML expects {n_embd, n_ff_exp, n_expert} which is PyTorch [n_expert, n_ff_exp, n_embd] ✓
```
```suggestion
            # HF: [n_expert, 2*n_ff_exp, n_embd] -> split dim=1 -> [n_expert, n_ff_exp, n_embd]
            # GGML expects {n_embd, n_ff_exp, n_expert} which is PyTorch [n_expert, n_ff_exp, n_embd]
```

convert_hf_to_gguf.py
```diff
@@ -4344,6 +4344,73 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+            n_ff = data_torch.shape[1] // 2
+            gate = data_torch[:, :n_ff, :].contiguous()
+            up = data_torch[:, n_ff:, :].contiguous()
+            base_name = name.removesuffix(".weight").removesuffix(".gate_up_proj")
+            mapped_gate = f"{base_name}.gate_proj.weight"
+            mapped_up = f"{base_name}.up_proj.weight"
+            yield from super(Qwen2MoeModel, self).modify_tensors(gate, mapped_gate, bid)
+            yield from super(Qwen2MoeModel, self).modify_tensors(up, mapped_up, bid)
```
```suggestion
            yield from Qwen2MoeModel.modify_tensors(self, gate, mapped_gate, bid)
            yield from Qwen2MoeModel.modify_tensors(self, up, mapped_up, bid)
```

convert_hf_to_gguf.py
```diff
@@ -4344,6 +4344,73 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+            return
+
+        # in_proj_qkv → ATTN_QKV and in_proj_z → ATTN_GATE are handled by
+        # tensor_mapping.py, so we just delegate to Qwen2MoeModel (grandparent)
+        # to skip Qwen3NextModel's in_proj_qkvz splitting logic
+        yield from super(Qwen3NextModel, self).modify_tensors(data_torch, name, bid)
```
```suggestion
        # in_proj_qkv → ATTN_QKV and in_proj_z → ATTN_GATE are handled by
        # tensor_mapping.py, so we just delegate to Qwen2MoeModel (grandparent)
        # to skip Qwen3NextModel's in_proj_qkvz splitting logic
        yield from Qwen2MoeModel.modify_tensors(self, data_torch, name, bid)
```
Which is it?

gguf-py/gguf/tensor_mapping.py
```diff
@@ -228,6 +228,7 @@ class TensorNameMap:
             "transformer_encoder.{bid}.qkv",                                       # neobert
             "layers.{bid}.attn.Wqkv",                                              # modern-bert
             "model.layers.{bid}.self_attn.language_expert_query_key_value",        # cogvlm
+            "model.layers.{bid}.linear_attn.in_proj_qkv",                            # qwen3.5
```
CISC:
```suggestion
            "model.layers.{bid}.linear_attn.in_proj_qkv",                          # qwen3.5
```

gguf-py/gguf/tensor_mapping.py
```diff
@@ -358,8 +359,9 @@ class TensorNameMap:
         ),
 
         MODEL_TENSOR.ATTN_GATE: (
-            "model.layers.{bid}.self_attn.gate_proj", # afmoe
-            "model.layers.{bid}.self_attn.g_proj",    # step3.5 head-wise attention gate
+            "model.layers.{bid}.self_attn.gate_proj",  # afmoe
+            "model.layers.{bid}.self_attn.g_proj",     # step3.5 head-wise attention gate
```
CISC:
```suggestion
            "model.layers.{bid}.self_attn.gate_proj",   # afmoe
            "model.layers.{bid}.self_attn.g_proj",      # step3.5 head-wise attention gate
```

src/models/qwen3-5.cpp
```diff
@@ -0,0 +1,418 @@
+#include "ggml.h"
```
CISC:
```suggestion
```

src/models/qwen3-5.cpp
```diff
@@ -0,0 +1,418 @@
...
+    auto * inp = build_inp_mem_hybrid();
+
+    ggml_tensor * inp_pos     = build_inp_pos();
+    ggml_tensor * inp_out_ids = build_inp_out_ids();
+
+    ggml_tensor * causal_mask =
+        ggml_tri(ctx0, ggml_fill_inplace(ctx0, ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, CHUNK_SIZE, CHUNK_SIZE), 1.0f),
+                    GGML_TRI_TYPE_LOWER);
+
+    ggml_tensor * identity = ggml_diag(ctx0, ggml_fill_inplace(ctx0, ggml_new_tensor_1d(ctx0, GGML_TYPE_F32, CHUNK_SIZE), 1.0f));
```
CISC:
```suggestion
        ggml_tri(ctx0, ggml_fill(ctx0, ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, CHUNK_SIZE, CHUNK_SIZE), 1.0f),
                    GGML_TRI_TYPE_LOWER);

    ggml_tensor * identity = ggml_diag(ctx0, ggml_fill(ctx0, ggml_new_tensor_1d(ctx0, GGML_TYPE_F32, CHUNK_SIZE), 1.0f));
```
The inplace should be automatic now.

src/models/qwen3-5.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+    ggml_tensor * Qcur = ggml_view_4d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens, 1,
+                                             Qcur_full->nb[1], Qcur_full->nb[2], Qcur_full->nb[3], 0);
+    ggml_tensor * gate =
+        ggml_view_4d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens, 1,
+                     Qcur_full->nb[1], Qcur_full->nb[2], Qcur_full->nb[3], n_embd_head * ggml_element_size(Qcur_full));
+    cb(Qcur, "Qcur", il);
+    cb(gate, "gate", il);
+
+    Qcur = ggml_cont_3d(ctx0, Qcur, n_embd_head, n_head, n_tokens);
```
CISC: The cont is unnecessary, and why not just create a 3d view in the first place?

src/models/qwen3-5.cpp
```diff
@@ -0,0 +1,418 @@
...
+    cb(Kcur, "Kcur", il);
+
+    ggml_tensor * Vcur = build_lora_mm(model.layers[il].wv, cur);
+    cb(Vcur, "Vcur", il);
+
+    Kcur = ggml_reshape_3d(ctx0, Kcur, n_embd_head, n_head_kv, n_tokens);
+    Kcur = build_norm(Kcur, model.layers[il].attn_k_norm, nullptr, LLM_NORM_RMS, il);
+    cb(Kcur, "Kcur_normed", il);
+
+    gate = ggml_cont_2d(ctx0, gate, n_embd_head * n_head, n_tokens);
```
CISC, line 130: Same.

am17an:
I think it maybe now makes sense to have a dedicated operator for delta_net to eliminate all those cpy_scalars

src/models/qwen3-5.cpp
```diff
@@ -0,0 +1,418 @@
...
+        cb(z, "z", il);
+
+        ggml_tensor * query_flat = ggml_cont_3d(ctx0, query, head_k_dim * num_k_heads, n_seq_tokens, n_seqs);
+        cb(query_flat, "query_flat", il);
+
+        ggml_tensor * key_flat = ggml_cont_3d(ctx0, key, head_k_dim * num_k_heads, n_seq_tokens, n_seqs);
+        cb(key_flat, "key_flat", il);
+
+        ggml_tensor * value_flat = ggml_cont_3d(ctx0, value, head_v_dim * num_v_heads, n_seq_tokens, n_seqs);
+        cb(value_flat, "value_flat", il);
```
CISC: Pretty sure the `cont` is unnecessary for `concat`.

src/models/qwen3-5moe.cpp
```diff
@@ -0,0 +1,458 @@
...
+    auto * inp = build_inp_mem_hybrid();
+
+    ggml_tensor * inp_pos     = build_inp_pos();
+    ggml_tensor * inp_out_ids = build_inp_out_ids();
+
+    ggml_tensor * causal_mask =
+        ggml_tri(ctx0, ggml_fill_inplace(ctx0, ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, CHUNK_SIZE, CHUNK_SIZE), 1.0f),
+                    GGML_TRI_TYPE_LOWER);
+
+    ggml_tensor * identity = ggml_diag(ctx0, ggml_fill_inplace(ctx0, ggml_new_tensor_1d(ctx0, GGML_TYPE_F32, CHUNK_SIZE), 1.0f));
```
CISC:
```suggestion
        ggml_tri(ctx0, ggml_fill(ctx0, ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, CHUNK_SIZE, CHUNK_SIZE), 1.0f),
                    GGML_TRI_TYPE_LOWER);

    ggml_tensor * identity = ggml_diag(ctx0, ggml_fill(ctx0, ggml_new_tensor_1d(ctx0, GGML_TYPE_F32, CHUNK_SIZE), 1.0f));
```

src/models/qwen3-5moe.cpp
```diff
@@ -0,0 +1,458 @@
...
+
+    ggml_tensor * Qcur = ggml_view_4d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens, 1,
+                                             Qcur_full->nb[1], Qcur_full->nb[2], Qcur_full->nb[3], 0);
+    ggml_tensor * gate =
+        ggml_view_4d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens, 1,
+                     Qcur_full->nb[1], Qcur_full->nb[2], Qcur_full->nb[3], n_embd_head * ggml_element_size(Qcur_full));
+    cb(Qcur, "Qcur", il);
+    cb(gate, "gate", il);
+
+    Qcur = ggml_cont_3d(ctx0, Qcur, n_embd_head, n_head, n_tokens);
```
CISC: rinse

src/models/qwen3-5moe.cpp
```diff
@@ -0,0 +1,458 @@
...
+    cb(Kcur, "Kcur", il);
+
+    ggml_tensor * Vcur = build_lora_mm(model.layers[il].wv, cur);
+    cb(Vcur, "Vcur", il);
+
+    Kcur = ggml_reshape_3d(ctx0, Kcur, n_embd_head, n_head_kv, n_tokens);
+    Kcur = build_norm(Kcur, model.layers[il].attn_k_norm, nullptr, LLM_NORM_RMS, il);
+    cb(Kcur, "Kcur_normed", il);
+
+    gate = ggml_cont_2d(ctx0, gate, n_embd_head * n_head, n_tokens);
```
CISC: repeat :)

src/models/qwen3-5moe.cpp
```diff
@@ -0,0 +1,458 @@
...
+        cb(z, "z", il);
+
+        ggml_tensor * query_flat = ggml_cont_3d(ctx0, query, head_k_dim * num_k_heads, n_seq_tokens, n_seqs);
+        cb(query_flat, "query_flat", il);
+
+        ggml_tensor * key_flat = ggml_cont_3d(ctx0, key, head_k_dim * num_k_heads, n_seq_tokens, n_seqs);
+        cb(key_flat, "key_flat", il);
+
+        ggml_tensor * value_flat = ggml_cont_3d(ctx0, value, head_v_dim * num_v_heads, n_seq_tokens, n_seqs);
+        cb(value_flat, "value_flat", il);
```
CISC: etc

src/models/qwen3-5moe.cpp
```diff
@@ -0,0 +1,458 @@
+#include "ggml.h"
```
CISC:
```suggestion
```

convert_hf_to_gguf.py
pwilkin: Damn, I forgot to tell him to add emojis!

convert_hf_to_gguf.py
pwilkin: He's lying, we want Next for all the other mods and the qkvz modification is gated by tensor name,  Qwen 3.5 has separate qkv and z tensors so it doesn't trigger. I'll change the comment to not be misleading.

pwilkin:
@CISC actually I just made the MoE class inherit from the normal one since most of the code is duplicated :)

pwilkin:
@CISC BTW reportedly `super(class, self).method(...)` is the preferred way to call superclasses under multiple inheritance over `superclass.method(self, ...)`

CISC:
> @CISC BTW reportedly `super(class, self).method(...)` is the preferred way to call superclasses under multiple inheritance over `superclass.method(self, ...)`

No, absolutely not, it does not do what you think it does.

Edit: As a thought experiment; `super()` is shorthand for `super(MyClass, self)`.

pwilkin:
> No, absolutely not, it does not do what you think it does.

WDYM? I though `super(AClass, self).method` means basically "call method for superclass of (self interpreted as instance of AClass)", isn't that what it does?

CISC:
Not quite (in that case `super()` would call its own method).

src/models/qwen3-5.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+    ggml_tensor * Qcur = ggml_view_4d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens, 1,
+                                             Qcur_full->nb[1], Qcur_full->nb[2], Qcur_full->nb[3], 0);
+    ggml_tensor * gate =
+        ggml_view_4d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens, 1,
+                     Qcur_full->nb[1], Qcur_full->nb[2], Qcur_full->nb[3], n_embd_head * ggml_element_size(Qcur_full));
+    cb(Qcur, "Qcur", il);
+    cb(gate, "gate", il);
+
+    Qcur = ggml_view_3d(ctx0, Qcur, n_embd_head, n_head, n_tokens, Qcur->nb[1], Qcur->nb[2], 0);
```
CISC: Not quite what I meant, the `ggml_view_4d` is redundant, same with `gate`.

pwilkin:
> Not quite (in that case `super()` would call its own method).

But why, if it calls the method of the "superclass of AClass"?

The way I understand it: 
* under single inheritance, if A is subclass of B, then `super(A, self).method(...)` is equivalent to `B.method(self, ...)`
* under multiple inheritance, if A is subclass of B and C, then `super(A, self).method(...)` means calling both B.method and C.method if they exist

Am I misunderstanding something?

src/models/qwen3-5.cpp
pwilkin: Ah, okay, see it now :)

CISC:
> Am I misunderstanding something?

I'll repeat the example. :)

`super().method` -> `super(MyClass, self).method`, ie calling the method of the parent of `MyClass`.
`super(ParentClass, self).method`, calls which method? :)

pwilkin:
> `super().method` -> `super(MyClass, self).method`, ie calling the method of the parent of `MyClass`. `super(ParentClass, self).method`, calls which method? :)

But I wasn't talking about calling `super(ParentClass, self).method`, I was talking about calling super(MyClass, self).method`.

CISC:
> > `super().method` -> `super(MyClass, self).method`, ie calling the method of the parent of `MyClass`. `super(ParentClass, self).method`, calls which method? :)
> 
> But I wasn't talking about calling `super(ParentClass, self).method`, I was talking about calling super(MyClass, self).method`.

Not what you're doing here:
https://github.com/ggml-org/llama.cpp/blob/1df85b7361a57650eece846bce363040841a23be/convert_hf_to_gguf.py#L4388-L4389
~Edit: I understand now, you mean to call `ModelBase.modify_tensors`? Then do so explicitly, you have no idea of knowing what `Qwen2MoeModel` inherits from at any given time.~

And this is old (not recommended) style, should just be `super()`:
https://github.com/ggml-org/llama.cpp/blob/1df85b7361a57650eece846bce363040841a23be/convert_hf_to_gguf.py#L4394

transformers
CISC, line 1: accidentally committed

pwilkin:
> Edit: I understand now, you mean to call `ModelBase.modify_tensors`? Then do so explicitly, you have no idea of knowing what `Qwen2MoeModel` inherits from at any given time.

Okay, that's what I wanted to know :) (as in, whether we prefer explicit superclass calls as in "we know what it is and we want to call it" to "we want to call the superclass")

CISC:
> > Edit: I understand now, you mean to call `ModelBase.modify_tensors`? Then do so explicitly, you have no idea of knowing what `Qwen2MoeModel` inherits from at any given time.
> 
> Okay, that's what I wanted to know :) (as in, whether we prefer explicit superclass calls as in "we know what it is and we want to call it" to "we want to call the superclass")

Actually, took a closer look, and you definitely want to call `Qwen2MoeModel` method as these are experts.

Edit: So, if you want to do the inheritance thing you would `super(Qwen3NextModel, self)`, but I prefer being explicit to leave out the guesswork (which may get swept out from under your feet).

pwilkin:
Aight, I made the gate and Qcur explicit views from the WQ matmul, the gate unfortunately requires a cont due to interleaved layout but still chunked a few ops off.

convert_hf_to_gguf.py
```diff
@@ -4385,13 +4385,13 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
-            yield from super(Qwen2MoeModel, self).modify_tensors(gate, mapped_gate, bid)
-            yield from super(Qwen2MoeModel, self).modify_tensors(up, mapped_up, bid)
+            yield from Qwen2MoeModel.modify_tensors(self, gate, mapped_gate, bid)
+            yield from Qwen2MoeModel.modify_tensors(self, up, mapped_up, bid)
             return
         else:
             # Qwen3Next uses .qkvz tensor, so we use the super to get the other functionalities
             # (norm correction, A_log to A etc.) for free
-            yield from super(Qwen3_5Model, self).modify_tensors(data_torch, name, bid)
+            yield from Qwen3NextModel.modify_tensors(self, data_torch, name, bid)
```
```suggestion
            yield from super().modify_tensors(data_torch, name, bid)
```
Here we do want inheritance since we do not seek to bypass anything.

src/models/qwen3-5.cpp
```diff
@@ -126,7 +118,11 @@ ggml_tensor * llm_build_qwen3_5::build_layer_attn(
     Kcur = build_norm(Kcur, model.layers[il].attn_k_norm, nullptr, LLM_NORM_RMS, il);
     cb(Kcur, "Kcur_normed", il);
 
-    gate = ggml_view_2d(ctx0, gate, n_embd_head * n_head, n_tokens, gate->nb[1], 0);
+    ggml_tensor * gate = ggml_view_3d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens,
+        ggml_element_size(Qcur_full) * n_embd_head * 2,
+        ggml_element_size(Qcur_full) * n_embd_head * 2 * n_head,
+        ggml_element_size(Qcur_full) * n_embd_head);
+    gate = ggml_cont_2d(ctx0, gate, n_embd_head * n_head, n_tokens);
```
CISC, line 130: Fairly sure `sigmoid` does not require `cont`.

CISC: Oh, right, it requires `ggml_is_contiguous_1`.

src/models/qwen3next.cpp
CISC: Mind taking care of the include here too?

CISC (approved):
GGUF wen? :)

src/models/qwen3next.cpp
CISC: Snuck itself into `kimi-linear` as well.

pwilkin:
So fun fact - packed expert conversion for Qwen2Moe was apparently wrong, which is why all these shenanigans were added. It was likely never caught because no models actually used the packed experts. I moved the fix over to Qwen2Moe instead and now just mostly call super()

ngxson:
Tagging @JJJYmmm too, we are not sure if you're working a llama.cpp implementation of the model. Would appreciate if you can test this implementation.

~~I'll follow-up with the audio support in the next days.~~ nevermind, it is vision-only

src/llama-model.cpp
```diff
@@ -8603,6 +8755,8 @@ llama_rope_type llama_model_rope_type(const llama_model * model) {
         case LLM_ARCH_PANGU_EMBED:
         case LLM_ARCH_AFMOE:
         case LLM_ARCH_QWEN3NEXT:
+        case LLM_ARCH_QWEN3_5:
+        case LLM_ARCH_QWEN3_5_MOE:
```
ngxson, line 8759: It seems to me that it uses the same rope style as qwen3VL and not normal qwen3, so IMROPE seems to be a better option: https://github.com/bozheng-hit/transformers/blob/be63a26077e10243d3fd8eb8cc074ce314feb04f/src/transformers/models/qwen3_5/modular_qwen3_5.py#L299-L302

But honestly IMORPE kernel is very slow on some backend due to the modulo operation. I doubt that the conversion script will need to be updated to permute the QK into normal MROPE (not interleaved) to get the best performance.

In either way, it seems like `mrope_section` is missing, so user will very likely need to reconvert the model once the vision support is out.

pwilkin: Argh, omni models.

You're right, obviously.

JJJYmmm:
@ngxsonp, thanks for tagging me! I’ve implemented qwen 3.5 before and was planning to submit the PR today…😂

ngxson:
@JJJYmmm ah sorry 😅 I wanted to ask you via the transformers PR, but I was offline most of the day.

we are still working with the vision support though. Would appreciate if you can help. Currently, I want to permute the weight to use MROPE instead of IMROPE, as the MROPE kernel seems to be more performant across all backend.

btw just out of curiosity, you mentioned on your profile that you are ex-intern. Wondering if you were promoted to full-time?

JJJYmmm:
@pwilkin, thank you for your day-0 support effort! I’ll be opening a follow-up PR to ensure the correctness of this one and support vision. 🙏cc @ngxson

JJJYmmm:
> @JJJYmmm ah sorry 😅 I wanted to ask you via the transformers PR, but I was offline most of the day.
> 
> we are still working with the vision support though. Would appreciate if you can help. Currently, I want to permute the weight to use MROPE instead of IMROPE, as the MROPE kernel seems to be more performant across all backend.
> 
> btw just out of curiosity, you mentioned on your profile that you are ex-intern. Wondering if you were promoted to full-time?

@ngxson Got it! I’ll try to convert it to the original mrope. And sorry for the misleading profile🥹 I’m back in school, but for some reason, I’m still helping with Qwen’s open-source efforts.

ngxson:
@JJJYmmm hah no worries. I appreciate your works on recent models from Qwen. The community definitely love it.

Wish you the best for your studies! And looking forward to your next open source contributions 💪

rankaiyx:
I suggest contacting the Qwen team and collaborating with them on testing.  When the model is released, the official Qwen team could release the GGUF version.

JJJYmmm:
@rankaiyx Yes, I'm currently handling this part, including gguf generation and release. 🫡

rankaiyx:
> @rankaiyx Yes, I'm currently handling this part, including gguf generation and release. 🫡

That’s great to hear! 👍 Excited to see it drop as soon as possible.

ggerganov:
@pwilkin There was no reason to merge this PR so early. It breaks Qwen3 Next for some reason - high PPL and same bug as in #19305 . And moreover the transformers PR is not merged yet.

CISC:
> So fun fact - packed expert conversion for Qwen2Moe was apparently wrong, which is why all these shenanigans were added. It was likely never caught because no models actually used the packed experts. I moved the fix over to Qwen2Moe instead and now just mostly call super()

Are you sure? There are several models inheriting from `Qwen2MoeModel`, I'm fairly sure this was added for one of them!

Didn't mean for you to merge minus-day btw. :P

pwilkin:
Sorry, got a bit too excited... Gonna fix the Next bug ASAP.

pwilkin:
> > So fun fact - packed expert conversion for Qwen2Moe was apparently wrong, which is why all these shenanigans were added. It was likely never caught because no models actually used the packed experts. I moved the fix over to Qwen2Moe instead and now just mostly call super()
> 
> Are you sure? There are several models inheriting from `Qwen2MoeModel`, I'm fairly sure this was added for one of them!

Quite sure - the MoE code didn't change at all between Qwen2 and Qwen3.5, 3.5 actually uses the packed tensors and you have to split on the second dimension, not the last one.

If any other arch uses this conversion scheme, it should go in that arch's dedicated code instead.

CISC:
> If any other arch uses this conversion scheme, it should go in that arch's dedicated code instead.

Sure, but point is some arch **may** be using it, I'll have a look...

JJJYmmm:
> > > So fun fact - packed expert conversion for Qwen2Moe was apparently wrong, which is why all these shenanigans were added. It was likely never caught because no models actually used the packed experts. I moved the fix over to Qwen2Moe instead and now just mostly call super()
> > 
> > 
> > Are you sure? There are several models inheriting from `Qwen2MoeModel`, I'm fairly sure this was added for one of them!
> 
> Quite sure - the MoE code didn't change at all between Qwen2 and Qwen3.5, 3.5 actually uses the packed tensors and you have to split on the second dimension, not the last one.
> 
> If any other arch uses this conversion scheme, it should go in that arch's dedicated code instead.

For some historical reason, qwen3vlmoe and gptoss has the transposed packed expert weights. https://github.com/huggingface/transformers/pull/43307

CISC:
> For some historical reason, qwen3vlmoe and gptoss has the transposed packed expert weights. [huggingface/transformers#43307](https://github.com/huggingface/transformers/pull/43307)

@pwilkin Added in #16780, you fix? :)

pwilkin:
@ggerganov Are you sure about this broken Next BTW? I rebased it on the fixed delta branch, so should be fine. Just checked perplexity and it seems OK, tested 100k deep query on Next-Coder IQ2_S and seems to work well too:

```console
> /read ggml/src/ggml-quants.c

Loaded text from 'ggml/src/ggml-quants.c'

> Can you please explain the TQ quants?

Okay, let's break down the **TQ (Ternary Quantization)** methods implemented in the code, specifically `tq1_0` and `tq2_0`, which are inspired by models like **BitNet b1.58** and **TriLMs**.

The core idea of TQ is to represent weights using **ternary values (-1, 0, +1)**, but encoded in a way that is efficient for computation, unlike standard quantization methods that use powers of 2 (like Q4, Q8).
(...)
```

pwilkin:
> @pwilkin Added in #16780, you fix? :)

I break, I fix :)

ggerganov:
Yes, both Metal and CUDA produce higher PPL compared to before the change.

> Just checked perplexity and it seems OK

Is it the same as before the PR?

pwilkin:
@ggerganov I checked the perplexity and it looks fine, can you please verify? I rebased this on the fixed version of the delta_net branch, so it should be correct.

> Yes, both Metal and CUDA produce higher PPL compared to before the change.
> 
> > Just checked perplexity and it seems OK
> 
> Is it the same as before the PR?

Seems so, but I will test again.

ggerganov:
It looks fine, but the problem is that it is different and significantly higher.

```bash
# new
Final estimate: PPL = 5.7395 +/- 0.35952

# old (e06088da0fa86aa444409f38dff274904931c507)
Final estimate: PPL = 5.1777 +/- 0.31137
```

It should not change. Also, try to perform the test from https://github.com/ggml-org/llama.cpp/issues/19305 - it fails.

pwilkin:
@ggerganov Verifying rn.

ggerganov:
@pwilkin If you need more time to locate and fix the problem, maybe it would be better to revert the PR for now and take the time to support this properly. No need to rush it. WDYT?

pwilkin:
@ggerganov give me half an hour, if I don't find it then we can do that instead, OK?

ggerganov:
I just don't have the confidence that these changes are good. The branch that you rebased on was closed https://github.com/ggml-org/llama.cpp/pull/19125 and it is not clear at all that it is working correctly. IMO the cleanest thing is to go back.

pwilkin:
@ggerganov Okay, you're right, I'll prepare a new one after you revert.

ggerganov:
Ok, I'll open the revert now.

mirek190:
> > A bit off-topic, but I'm curious to try the same task with a local model. Maybe GLM 4.7 Flash + OpenCode. Do you have the guidelines that you used somewhere shareable?
> 
> I'm using my "adding model architectures" tutorial for this actually (#16770) + got an extra rules section in my agent MD reminding about the tensor layout:
> 
> ```
> ## Tensor format
> 
> A very important caveat about tensor format in the GGML library: the GGML library uses a different format internally than most Python implementations,
> including PyTorch or Transformers. Notably, in the GGML library:
> 
> * tensors are restricted to 4 dimensions - whenever you want to use more dimensions you have to pack them
> * the semantic order of dimensions is reversed from PyTorch/Transformers - the *last two* dimensions are used for `[tokens_per_batch, batches]` - even though the physical layout is the same
> 
> So, for example, a `[1, 5, 4096, 1]` tensor in PyTorch will probably become a `[1, 4096, 5, 1]` tensor in GGML. This is especially important
> when converting certain implementations from a reference implementation written in Python, because the tensor dimension ordering will be different,
> but they can be semantically the same.
> ```
> 
> The prompt I used this time:
> 
> ```
> "In the transformers directory, I have included a new version of Transformers that includes support for the new Qwen3.5 series of models (MoE and dense). @transformers/src/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py @transformers/src/transformers/models/qwen3_5/modeling_qwen3_5.py 
> 
> I have also created a script and generated mock models for testing, the script is in @transformers/generate_qwen_models.py and has been already run.
> 
> Based on the implementation, which seems to be based strongly on the Qwen3 Next architecture, currently handled in Llama.cpp in @reference/qwen3next.cpp and in @src/models/delta.cpp, please create the implementation and conversion code for the new architectures. There are scripts in @examples/model-conversion/Makefile for testing conversions of new models. 
> 
> See @.roo/rules/adding_new_models.md for tips on how to add new model support. Make sure to reuse as much existing code as possible. For now, only add text support, without the multimodal capabilities."
> ```
> 
> EDIT: Key thing was also interrupting the model when it starts to do something stupid, for example modifying permutation rules in delta_net because it added incorrect extra permutations in conversion. Even with Opus 4.6 I had to do it like 3 or 4 times to stop it from going in the wrong direction.
> 
> Oh yeah, I also made some agents that I feel work well with Lllama.cpp for OpenCode (I've used them in fact for some documentation work), so I might share those too:
> 
> [deep-code-modifier.md](https://github.com/user-attachments/files/25164201/deep-code-modifier.md) [code-architecture-analyzer.md](https://github.com/user-attachments/files/25164202/code-architecture-analyzer.md)

Have you tried to use GPT codex 5.3 xhigh  with codex-cli as is better for a complex coding?

pwilkin:
@mirek190 I think they're comparable, but I don't have an active Codex sub atm, just Claude.

mirek190:
According to Matt Maher

https://www.youtube.com/watch?v=hwvyew2iXpU&t=937s

On real world usage his tests he got 86% with opus 4.6 high and 95% with codex 5.3 xhigh and claims codex is smarter on complex code.
That's something I also noticed. Seems codex is better for complex tasks than opus 4.6 and is absurd cheap because for 20 USD you can work whole week now. 


I just saying....

PR #19440, opened 2026-02-08
Use Vulkan SDK Constants in ggml-vulkan.cpp

inforithmics:
I Used AI (Jetbrains Junie and Anthropic Claude) for finding the strings and replacing them with Vulkan SDK Constants.
I manually reviewed it, so that not an Extension Name change did happen.

I discovered this while looking at the code, while Debugging a Vulkan issue.

I have read the contributers guidlines. It compiles locally on Windows.

I did run tests:

ctest -C Debug
The following tests FAILED:
         18 - test-jinja-py (Failed)                            python
         22 - test-regex-partial (Exit code 0xc0000409)         main (Memory Access Exception)

ctest -C Release
The following tests FAILED:
         18 - test-jinja-py (Failed)                            python

But I think the failures are not releated


0cc4m:
These extension name constants are not gonna change and the names are actually longer than their values, so I don't really see a reason to use them. How does this help you with debugging an issue?

inforithmics:
In Debugging this doesn‘t help. But in reading the Code. Because it is a constant I don‘t have to check the writing because the compiler takes care of that.
 But in the End It is a question of Code Style. I like to use constants because I have the tendency to make small writing errors in string constants which I often don‘t see on reading them.
For example maintnance instead of maintenance…..
So mainly a habit of coding style.

0cc4m:
I agree for constants that get used more often, but in this case it would just make them longer and they are not actually used that much, so I'd rather not change it.

PR #19441, opened 2026-02-08
mtmd: qwen3 audio support (qwen3-omni and qwen3-asr)

ngxson:
Status:
- [x] qwen3-omni-moe working (vision + audio input)
- [x] qwen3-asr working


samshipengs:
Any updates on this?

yimlin:
any updates?

michoecho:
I wrote a working Qwen3-ASR support for my own use at https://github.com/michoecho/llama.cpp/commits/qwen3_asr_support. (I successfully used it to transcribe some lectures in Chinese). I don't know if it's good enough for upstreaming, because I wasn't thinking about qwen3-omni at all. (I have no idea what "deepstack" is). But you could use it as a working base if you are getting wrong results.

At a glance, what mainly seems to be missing from this PR is:
- To make my changes work properly, I had to fix a preexisting bug in whisper preprocessing which was causing the last audio chunk to be lost during: https://github.com/michoecho/llama.cpp/commit/63b3c1ec0cb1f73f4cb3a7056ae7356b413452f2#diff-a027f93a5e0a3fe643975f0ae176db52a3330a9422857b4f6fd9bfbac134c863R384. (I haven't reported it because I'm not 100% sure it's a bug — maybe I'm not seeing something — but I'm 99% sure).
- The `ggml_permute` seems to have channels and frames swapped around.
- Qwen3-ASR uses `<|audio_start|>` and `<|audio_end|>` instead of `<|audio_bos|>` and `<|audio_eos|>`.
- The audio encoder expects windowed (/chunked) attention, with window size between 1s and 8s. If you run the encoder with full attention on a 30s chunk, you will get bogus results. I didn't want to implement windowed attention (because I would have to learn how to do that), so I just solved it in the preprocessing layer by splitting audio into 8s chunks instead of 30s chunks.
- As a comment in this PR acknowledges, the reference implementation runs the conv2d layers on chunks of length 100. I followed the reference implementation. I don't know either if this chunking is necessary. I didn't test the non-chunked variant.
- The default chat template of Qwen3-ASR doesn't work with llama.cpp. (It expects the audio to be passed via some special params). (My fork doesn't care about chat templates at all either, because my application constructs the prompt directly anyway. But if you want to implement a chat template, the prompt expected by the model isn't anything fancy, it's basically just `chatml` with audio used as the user message).

By the way, note that Qwen3-ForcedAligner (the timestamp predictor model) has the same architecture as Qwen3-ASR, so if you implement support for the latter, you almost get support for the former too. "Almost" because the ForcedAligner is a non-autoregressive classification model. (You put in the encoded audio and the transcribed text with some `<timestamp>` tokens mixed in, then you run a single prediction on it, and the logits on the `<timestamp>` tokens will describe the timestamp at those points in the text). I'm not sure how to integrate something like that with llama.cpp's abstractions. For my private use case (generating subtitles for the Chinese lectures) I added "support" for Qwen3-ForcedAligner too, but it's too hacky to post.

convert_hf_to_gguf.py
```diff
@@ -4303,8 +4295,33 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
             if "audio_bos_eos_token" in name:
                 # this tensor is left unused in transformers code
                 # https://github.com/huggingface/transformers/blob/6e3063422c4b1c014aa60c32b9254fd2902f0f28/src/transformers/models/qwen2_5_omni/modular_qwen2_5_omni.py#L1809
-                return
-        yield from super().modify_tensors(data_torch, name, bid)
+                return []
+            return [(self.map_tensor_name(name), data_torch)]
+
+        return [] # skip other tensors
```
```suggestion
                return
            yield from super().modify_tensors(data_torch, name, bid)

        return  # skip other tensors
```

convert_hf_to_gguf.py
```diff
@@ -4303,8 +4295,33 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+        self.gguf_writer.add_clip_projector_type(gguf.VisionProjectorType.QWEN25O)
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        if "visual." in name:
+            yield from Qwen2VLVisionModel.modify_tensors(self, data_torch, name, bid)
+        elif "audio_tower." in name:
+            yield from Qwen25AudioModel.modify_tensors(self, data_torch, name, bid)
+        return [] # skip other tensors
```
```suggestion
        return  # skip other tensors
```

convert_hf_to_gguf.py
```diff
@@ -4917,11 +4939,64 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
             return
 
         if name.startswith("visual."):
-            yield from super().modify_tensors(data_torch, name, bid)
-            return
+            yield (self.map_tensor_name(name), data_torch)
+        return [] # skip other tensors
```
```suggestion
            yield from super().modify_tensors(data_torch, name, bid)
        return  # skip other tensors
```

convert_hf_to_gguf.py
```diff
@@ -4955,9 +5030,10 @@ class Qwen3VLTextModel(Qwen3Model):
 
     def set_gguf_parameters(self):
         super().set_gguf_parameters()
-
-        # Handle MRoPE (Multi-axis Rotary Position Embedding) for Qwen3-VL
-        vision_config = self.hparams.get("vision_config", {})
+        if "thinker_config" in self.hparams:
+            vision_config = self.hparams["thinker_config"].get("vision_config", {})
+        else:
+            vision_config = self.hparams.get("vision_config", {})
```
CISC, line 5111: Instead of handling this everywhere, can't we just merge in all sub-configs in `thinker_config` here:
https://github.com/ggml-org/llama.cpp/blob/eefcfeed7abcafc74901c46721e6c3324b85c144/convert_hf_to_gguf.py#L974-L976

convert_hf_to_gguf.py
```diff
@@ -4969,20 +5045,66 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
-        self.gguf_writer.add_num_deepstack_layers(deepstack_layer_num)
+        self.gguf_writer.add_num_deepstack_layers(0)
 
     def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
         # Skip vision tensors - they go in the mmproj file
-        if name.startswith("model.visual."):
-            return
+        if "visual." in name or "audio_tower." in name \
+                or "talker." in name or "code2wav." in name:
+            return []
```
```suggestion
            return
```

convert_hf_to_gguf.py
ngxson: nice catch, yeah this code was written before the `yield from` refactor 😅

convert_hf_to_gguf.py
ngxson: hmm that can be quite dangerous because the sub config may have conflict keys with the `thinker_config`

I think it's fine to keep this as-is (a bit lazy to re-test this). plus, we only have one single place in the whole file that does this.

for ref, normally a text model never have to read the vision config, but this is the specific case for qwen3 to support "deep stack". from qwen3.5, they removed the deep stack

convert_hf_to_gguf.py
```diff
@@ -5054,21 +5130,34 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+    def set_vocab(self):
+        super().set_vocab()
+        # correct BOS/EOS tokens
+        with open(self.dir_model / "tokenizer_config.json", "r", encoding="utf-8") as f:
+            tokenizer_config = json.load(f)
+            added_tokens = tokenizer_config.get("added_tokens_decoder", {})
+            for token_id, data in added_tokens.items():
+                if data.get("content") == "<|im_end|>":
+                    self.gguf_writer.add_bos_token_id(int(token_id))
+                    self.gguf_writer.add_eos_token_id(int(token_id))
```
```suggestion
                    self.gguf_writer.add_eos_token_id(int(token_id))
                    break
```

convert_hf_to_gguf.py
```diff
@@ -5054,21 +5130,34 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         yield from super().modify_tensors(data_torch, name, bid)
 
 
-@ModelBase.register("Qwen3VLMoeForConditionalGeneration")
+@ModelBase.register("Qwen3VLMoeForConditionalGeneration", "Qwen3OmniMoeForConditionalGeneration")
```
CISC: `Qwen3OmniMoeForConditionalGeneration` is also registered to `Qwen3OmniMoeTextModel` below (both `TextModel`s)?

convert_hf_to_gguf.py
```diff
@@ -5101,6 +5190,59 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+    def set_vocab(self):
+        super().set_vocab()
+        # correct BOS/EOS tokens
+        with open(self.dir_model / "tokenizer_config.json", "r", encoding="utf-8") as f:
+            tokenizer_config = json.load(f)
+            added_tokens = tokenizer_config.get("added_tokens_decoder", {})
+            for token_id, data in added_tokens.items():
+                if data.get("content") == "<|im_end|>":
+                    self.gguf_writer.add_bos_token_id(int(token_id))
+                    self.gguf_writer.add_eos_token_id(int(token_id))
```
CISC, line 5193:
```suggestion
                    self.gguf_writer.add_eos_token_id(int(token_id))
                    break
```

convert_hf_to_gguf.py
```diff
@@ -5101,6 +5190,59 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+        # fix chat template, use correct chatml format
+        self.gguf_writer.add_chat_template("{% for message in messages %}{{'<|im_start|>' + message['role'] + '\\n' + message['content'] + '<|im_end|>' + '\\n'}}{% endfor %}{% if add_generation_prompt %}{{ '<|im_start|>assistant\\n' }}{% endif %}")
+        # correct BOS/EOS tokens
+        with open(self.dir_model / "tokenizer_config.json", "r", encoding="utf-8") as f:
+            tokenizer_config = json.load(f)
+            added_tokens = tokenizer_config.get("added_tokens_decoder", {})
+            for token_id, data in added_tokens.items():
+                if data.get("content") == "<|im_end|>":
+                    self.gguf_writer.add_bos_token_id(int(token_id))
+                    self.gguf_writer.add_eos_token_id(int(token_id))
```
CISC, line 5229:
```suggestion
                    self.gguf_writer.add_eos_token_id(int(token_id))
                    break
```

CISC:
Uhoh, I think \r\n strikes again...

ngxson:
hmm seems like it mostly happens when you add a new line in the suggestion, right?

convert_hf_to_gguf.py
ngxson: yeah that was a bad merge, fixed in https://github.com/ggml-org/llama.cpp/pull/19441/commits/53d149b563c0d50ff3df1b4702a488c902b31d48

CISC:
> hmm seems like it mostly happens when you add a new line in the suggestion, right?

Yes, but only when you (and one other person so far) commit them. :)

CISC (approved):

ngxson:
hmm, the conversion is broken somehow, the omni mmproj now only contains audio tensors, not vision tensors

digging into this...

ngxson:
OK it got a bit messy with multi inheritance, but I fixed in the last commit

CC @pwilkin if you can give the 2nd approval, thanks!

ngxson:
pinging @ggml-org/maintainers if someone can give an approval, thanks!

ServeurpersoCom (approved):

bennmann:
no audio output for the nice voices?

pwilkin:
@ngxson sorry, somehow missed this one.

erazortt:
> no audio output for the nice voices?

Yes, that would also be of interest for me. Is there audio output planned?

pablopla:
Can you show example how to use it? I'm getting short transcription in low quality when transcribing a 1 minute audio file.
The qwen-asr [online demo](https://huggingface.co/spaces/Qwen/Qwen3-ASR) works fine.
Should I use a prompt and set specific parameters?
Should I send the request to "/v1/audio/transcriptions" or ""/v1/chat/completions"

PR #19451, opened 2026-02-09
Remove redundant traversal of oaicompat_msg_diffs from server_task_result_cmpl_final::to_json_*_stream

openingnow:
This PR removes `for (const auto & diff : oaicompat_msg_diffs) { }` block from 
- `server_task_result_cmpl_final::to_json_oaicompat_chat_stream()`
- `server_task_result_cmpl_final::to_json_anthropic_stream()`

(It is not present in `server_task_result_cmpl_final::to_json_oaicompat_resp_stream()`)

#### Rationale:
`server_task_result_cmpl_final::oaicompat_msg_diffs` is empty for streaming responses.
https://github.com/ggml-org/llama.cpp/blob/1e8924fd65ad349d1d838412a2172292618f3bbf/tools/server/README-dev.md?plain=1#L84-L88

https://github.com/ggml-org/llama.cpp/blob/1e8924fd65ad349d1d838412a2172292618f3bbf/tools/server/server-context.cpp#L2774-L2782

https://github.com/ggml-org/llama.cpp/blob/1e8924fd65ad349d1d838412a2172292618f3bbf/tools/server/server-context.cpp#L1438-L1452

https://github.com/ggml-org/llama.cpp/blob/1e8924fd65ad349d1d838412a2172292618f3bbf/tools/server/server-task.h#L377-L384

https://github.com/ggml-org/llama.cpp/blob/1e8924fd65ad349d1d838412a2172292618f3bbf/tools/server/server-task.cpp#L159-L176

The only difference is `is_partial` which does nothing when parsing identical output.

### DIFF
--- a/tools/server/server-task.cpp
+++ b/tools/server/server-task.cpp
@@ -767,22 +767,6 @@ json server_task_result_cmpl_final::to_json_oaicompat_chat_stream() {
     }
 
     json deltas = json::array();
-    for (const auto & diff : oaicompat_msg_diffs) {
-        deltas.push_back({
-            {"choices", json::array({
-                json {
-                    {"finish_reason", nullptr},
-                    {"index", 0},
-                    {"delta", common_chat_msg_diff_to_json_oaicompat(diff)},
-                },
-            })},
-            {"created", t},
-            {"id", oaicompat_cmpl_id},
-            {"model", oaicompat_model},
-            {"system_fingerprint", build_info},
-            {"object", "chat.completion.chunk"},
-        });
-    }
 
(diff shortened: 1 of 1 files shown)


ngxson:
Hmm that sounds strange. I'm not the original author of this code, so don't fully understand how it works.

IIRC this is to handle partial tool calling response. Can you confirm if it has no impacts to this case?

openingnow:
Figured out that `oaicompat_msg_diffs` will not empty if there is trucated tool call. I'll prepare a patch for `to_json_oaicompat_resp_stream`

PR #19445, opened 2026-02-09
Add special case for Qwen3VLMoe

pwilkin:
Fix special conversion case for Qwen3VLMoe models introduced in #16780

PR #19453, opened 2026-02-09
revert : "[Model] Qwen3.5 dense and MoE support (no vision) (#19435)"

ggerganov:
cont #19435

Taking a step back to implement support for Qwen3.5 properly.


pwilkin (approved):

CISC (approved):
I suggest updating and merging #19445 independently after this is merged.

PR #19454, opened 2026-02-09
mtmd : Implement tiling for LFM2.5-VL models

tdakhran:
HF [implementation ](https://github.com/huggingface/transformers/blob/v5.1.0/src/transformers/models/lfm2_vl/image_processing_lfm2_vl_fast.py)of LFM Vision models splits the image into tiles for the highest quality output (e.g. for OCR).

PR reuses some logic from `llava_uhd` and adds LFM unique logic.
Tiling triggers when the image width or height exceeds 512.
It is slow because each tile generates 256 image tokens. For fast inference, it's recommended to downsample the image so that its width and height are less than or equal to 512.

Change default value for `image_max_tokens` for LFM2.5-VL to `256` to match HF implementation

Example for image 
v5

with command
```
./llama-mtmd-cli -hf LiquidAI/LFM2.5-VL-1.6B-GGUF --image v5.png -p "What is the biggest breaking change?"
```

master:
>The biggest breaking change is in the transform command. This command starts an animi UI interact with a model. It is used to start a Chat Conversion powered by transform.

PR:
> The biggest breaking change is in the transformers chat command. This command starts a terminal UI to interact with a chat model. It used to also be able to start a Chat Completion server powered by transformers and chat with it.

HF: 
> The biggest breaking change is in the transformers chat command. This command starts a terminal UI to interact with a chat model. It used to also be able to start a Chat Completion server powered by transformers and chat with it. In this revamped version, this feature has been removed in favor of transformers serve. The goal of splitting transformers chat and transformers serve is to define clear boundaries between client and server code. It helps with maintenance but also makes the commands less bloated. The new signature of transformers chat is: transformers chat.

### DIFF
--- a/tools/mtmd/clip.cpp
+++ b/tools/mtmd/clip.cpp
@@ -10,6 +10,7 @@
 #include "ggml-backend.h"
 #include "gguf.h"
 
+#include <algorithm>
 #include <cassert>
 #include <cmath>
 #include <cstdlib>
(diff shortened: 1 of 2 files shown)


ngxson (approved):
Nice improvement, thanks! Merging when CI passes

PR #19455, opened 2026-02-09
python: Use NumPy 2.0+

akx:
Old versions of NumPy are increasingly difficult to install on newer Python and OS versions. This changes the minimum version of NumPy to 2.0 (with a very broad range of allowed versions; basically "any 2.x series").

The end result of this is that

```
uv pip install -r ./requirements.txt --index-strategy unsafe-best-match
```

just works on a Python 3.13 interpreter, where it would begin to build Numpy 1.26.4 from source on `master`. NumPy 2.0 has wheels down to Python 3.9 (which in itself is EOL), so not much should change on older Pythons. 

I eyeballed (and asked Claude Code to also double-check) for possible breaking NumPy API changes, and didn't find any.

Most of the size of this PR is just the result of `uvx --python=3.11 poetry@1.7.1 lock` (and whether or not the top-level `pyproject.toml` and `poetry.lock` are required or up to date is another question).


CISC:
> I eyeballed (and asked Claude Code to also double-check) for possible breaking NumPy API changes, and didn't find any.

Thanks, should be ok as most of us have been using NumPy 2.x for quite a while anyway.

However, as we technically still support 1.x it might make more sense to change restriction to just `<3.0` (including in all the requirements files).

> Most of the size of this PR is just the result of `uvx --python=3.11 poetry@1.7.1 lock` (and whether or not the top-level `pyproject.toml` and `poetry.lock` are required or up to date is another question).

TBH not entirely sure.

CISC:
Hmmm, seems `numpy.typing` and `pyright` don't get along...

PR #19457, opened 2026-02-09
Server: log when converting requests to chat completions format

openingnow:
Emit info/debug when oai responses / anthropic requests are converted


ngxson (commented):
IMO it will be more useful to log the request path instead. I'll do that in another PR.

For now, I think it's better to change **all** log lines to `SRV_DBG` if you need it

openingnow:
Now using debug for both lines

ngxson (approved):

Issue #19269, opened 2026-02-02
Compile bug: ROCm 7.2 + rocwmma


MikeLP:
### Git commit

6fdddb498780dbda2a14f8b49b92d25601e14764

### Operating systems

Linux

### GGML backends

HIP

### Problem description & steps to reproduce

I'm trying to compile llama.cpp with ROCm 7.2 + rocwmma support on a fresh install of Ubuntu 24.04, but I get a compilation error.

It looks like the issue is with rocwmma support. I understand that llama.cpp may not support it yet, so I decided it's a good idea to create a ticket for it.

btw. No problems for 6.4 and 7.1 (warning about rocwwma bug)

### Compile command

```shell
HIPCXX="$(hipconfig -l)/clang" \      
HIP_PATH="$(hipconfig -R)" \
cmake -S . -B build \
  -DGGML_HIP=ON \
  -DGGML_HIP_ROCWMMA_FATTN=ON \
  -DGGML_NATIVE=ON \
  -DBUILD_SHARED_LIBS=ON \
  -DLLAMA_BUILD_EXAMPLES=OFF \
  -DLLAMA_BUILD_TESTS=OFF \
  -DLLAMA_BUILD_TOOLS=ON \
  -DLLAMA_BUILD_SERVER=ON \
  -DGGML_RPC=ON \
  -DCMAKE_BUILD_TYPE=Release  \
&& cmake --build build --config Release -- -j 48
```

### Relevant log output

```shell
[ 40%] Building HIP object ggml/src/ggml-hip/CMakeFiles/ggml-hip.dir/__/ggml-cuda/template-instances/fattn-vec-instance-q4_0-q8_0.cu.o
In file included from /home/iyanello/Projects/ML/llama.cpp/ggml/src/ggml-cuda/fattn-wmma-f16.cu:18:
In file included from /opt/rocm/include/rocwmma/rocwmma.hpp:29:
In file included from /opt/rocm/include/rocwmma/internal/accessors.hpp:29:
In file included from /opt/rocm/include/rocwmma/internal/accessors_impl.hpp:35:
In file included from /opt/rocm/include/rocwmma/internal/mma_config.hpp:32:
In file included from /opt/rocm/include/rocwmma/internal/mfma.hpp:32:
/opt/rocm/include/rocwmma/internal/mfma_impl.hpp:2549:37: error: ambiguous partial specializations of 'amdgcn_mfma<__half, __half, __half, 16, 16, 16>'
[...]

PR #19461, opened 2026-02-09
Add a workaround for compilation with ROCWMMA_FATTN and gfx9

superm1:
There is an upstream problem [1] with AMD's LLVM 22 fork and rocWMMA 2.2.0 causing compilation issues on devices without native fp16 support (CDNA devices).

The specialized types aren't resolved properly:
```c
/opt/rocm/include/rocwmma/internal/mfma_impl.hpp:2549:37: error: ambiguous partial specializations of 'amdgcn_mfma<__half, __half, __half, 16, 16, 16>'
 2549 |             using ARegsT = typename Impl::ARegsT;
```

Add a workaround to explicitly declare the types and cast when compiling with HIP and ROCWMMA_FATTN [2].  When this is actually fixed upstream some guards can be used to detect and wrap the version that has the fix to only apply when necessary.

Link: https://github.com/ROCm/rocm-libraries/issues/4398 [1]
Link: https://github.com/ggml-org/llama.cpp/issues/19269 [2]

CC @IMbackK

### DIFF
--- a/ggml/src/ggml-cuda/fattn-wmma-f16.cu
+++ b/ggml/src/ggml-cuda/fattn-wmma-f16.cu
@@ -63,11 +63,19 @@ static __global__ void flash_attn_ext_f16(
     constexpr int frag_m = ncols == 8 ? 32 : 16;
     constexpr int frag_n = ncols == 8 ?  8 : 16;
     static_assert(D % frag_m == 0, "If ncols == 8 then D % frag_m must be 0.");
+#if defined(GGML_USE_HIP)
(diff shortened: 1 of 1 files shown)


IMbackK (approved):
Looks good, while i dont really like it, this is the best solution available to us.

JohannesGaessler (approved):
FYI in terms of my current priorities I'll take a crack at better AMD WMMA/MFMA support in the MMA kernel once I'm done with tensor parallelism. So hopefully rocWMMA can soon be removed as a dependency anyways.

superm1:
It looks like CI passed, can this be merged so my other ones can do test builds now?

PR #19456, opened 2026-02-09
[Model] Qwen3.5 support w/o vision, WIP

pwilkin:
Followup to https://github.com/ggml-org/llama.cpp/pull/19435

Rebased on top of #18792 together with @ngxson 's fix for chunking.

PR #19468, opened 2026-02-09
[MODEL] support qwen3.5 series 

JJJYmmm:
This PR adds model support for the upcoming Qwen3.5 models, including both dense and MoE variants. It has been verified with preview checkpoints from the Qwen Team in both vision and pure text modes (with or without mmproj file).

As mentioned [here](https://github.com/ggml-org/llama.cpp/pull/19435#issuecomment-3868687151), this PR was planned to be a follow-up PR after https://github.com/ggml-org/llama.cpp/pull/19435 or https://github.com/ggml-org/llama.cpp/pull/19456 by @pwilkin. However, it seems there is still a lot to do without the real checkpoints (such as vision support, vocab, weight mapping, and forward logic), so I'm sharing my implementation here for reference. 🫡

Some differences compared to https://github.com/ggml-org/llama.cpp/pull/19435 or https://github.com/ggml-org/llama.cpp/pull/19456:
- Qwen3.5 uses a different pre-tokenizer
- Qwen3.5 uses partial IMRoPE
- Qwen3.5 has fixed the qkvzba order—thanks to @ngxson in https://github.com/ggml-org/llama.cpp/pull/18683. I saw it last month and we decided to simplify it in the official weights! Now they are split into qkv, z, b, and a.
- Qwen3.5 has dense/MoE variants, allowing us to simplify the forward logic (no need to check the mlp type of the current layer).

@pwilkin Sorry for the work overlap, and I really appreciate your work on Qwen3Next! It has been a huge stepping stone for this PR. 

Reference HF implementation: https://github.com/huggingface/transformers/pull/43830


pwilkin:
@JJJYmmm no worries, glad if it helped even a little :)

convert_hf_to_gguf_update.py
```diff
@@ -148,6 +148,7 @@ class TOKENIZER_TYPE(IntEnum):
     {"name": "youtu",            "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/tencent/Youtu-LLM-2B", },
     {"name": "solar-open",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/upstage/Solar-Open-100B", },
     {"name": "exaone-moe",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B", },
+    {"name": "qwen35",         "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/Qwen/Qwen3.5-9B-Instruct", }
```
ngxson: wow, 35B moe and 9B dense, that's a very reasonable sizes, I like it! 😄

src/llama-model.cpp
```diff
@@ -125,6 +125,7 @@ const char * llm_type_name(llm_type type) {
         case LLM_TYPE_21B_A3B:       return "21B.A3B";
         case LLM_TYPE_30B_A3B:       return "30B.A3B";
         case LLM_TYPE_31B_A3_5B:     return "31B.A3.5B";
+        case LLM_TYPE_35B_A3B:     return "35B.A3B";
```
ngxson:
```suggestion
        case LLM_TYPE_35B_A3B:       return "35B.A3B";
```

src/llama-model.cpp
```diff
@@ -2412,6 +2413,56 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+                // Load linear attention (gated delta net) parameters
+                ml.get_key(LLM_KV_SSM_CONV_KERNEL,    hparams.ssm_d_conv);
+                ml.get_key(LLM_KV_SSM_INNER_SIZE,     hparams.ssm_d_inner);
+                ml.get_key(LLM_KV_SSM_STATE_SIZE,     hparams.ssm_d_state);
+                ml.get_key(LLM_KV_SSM_TIME_STEP_RANK, hparams.ssm_dt_rank);
+                ml.get_key(LLM_KV_SSM_GROUP_COUNT,    hparams.ssm_n_group);
+
+                // Mark recurrent layers (linear attention layers)
+                for (uint32_t i = 0; i < hparams.n_layer; ++i) {
+                    hparams.recurrent_layer_arr[i] = ((i + 1) % 4 != 0); // TODO: extract the magic 4 from "full_attention_interval"
```
ngxson: I think we can add a GGUF metadata for `recurrent_layer_arr`

ngxson (commented):
I haven't reviewed the actual implementation, but I assume it's mostly qwen3 next + imrope (please correct if I'm wrong)

Don't hesitate to ping if you need helps to add the vision support!

convert_hf_to_gguf_update.py
ngxson:
```suggestion
    {"name": "qwen35",           "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/Qwen/Qwen3.5-9B-Instruct", }
```

JJJYmmm:
> @JJJYmmm no worries, glad if it helped even a little :)

It definitely helped a lot 🙏

src/llama-model.cpp
JJJYmmm: got it. I'll update it tomorrow. (going to bed now 🫡

gguf-py/gguf/tensor_mapping.py
```diff
@@ -1872,6 +1879,12 @@ class TensorNameMap:
                 "model.layers.{bid}.post_attention_layernorm",
             ),
         },
+        MODEL_ARCH.QWEN35: {
+            MODEL_TENSOR.FFN_NORM: (),
+        },
+        MODEL_ARCH.QWEN35MOE: {
+            MODEL_TENSOR.FFN_NORM: (),
+        },
```
CISC: Hmmm, why?

gguf-py/gguf/constants.py
```diff
@@ -1818,6 +1824,62 @@ class MODEL_TENSOR(IntEnum):
...
+        MODEL_TENSOR.ATTN_Q,
+        MODEL_TENSOR.ATTN_Q_NORM,
+        MODEL_TENSOR.ATTN_K,
+        MODEL_TENSOR.ATTN_K_NORM,
+        MODEL_TENSOR.ATTN_V,
+        MODEL_TENSOR.ATTN_OUT,
+        MODEL_TENSOR.ATTN_POST_NORM,
+        MODEL_TENSOR.ATTN_GATE,
+        MODEL_TENSOR.ATTN_QKV,
+        MODEL_TENSOR.FFN_NORM,
```
CISC:
```suggestion
```

src/llama-model.cpp
```diff
@@ -7101,6 +7150,130 @@ bool llama_model::load_tensors(llama_model_loader & ml) {
...
+                        layer.ffn_gate_inp  = create_tensor(tn(LLM_TENSOR_FFN_GATE_INP,  "weight", i), { n_embd, n_expert }, 0);
+                        layer.ffn_gate_exps = create_tensor(tn(LLM_TENSOR_FFN_GATE_EXPS, "weight", i), { n_embd, n_ff_exp, n_expert }, 0);
+                        layer.ffn_down_exps = create_tensor(tn(LLM_TENSOR_FFN_DOWN_EXPS, "weight", i), { n_ff_exp, n_embd, n_expert }, 0);
+                        layer.ffn_up_exps   = create_tensor(tn(LLM_TENSOR_FFN_UP_EXPS,   "weight", i), { n_embd, n_ff_exp, n_expert }, 0);
+
+                        // Shared experts
+                        layer.ffn_gate_inp_shexp = create_tensor(tn(LLM_TENSOR_FFN_GATE_INP_SHEXP, "weight", i), { n_embd }, 0);
+                        layer.ffn_gate_shexp     = create_tensor(tn(LLM_TENSOR_FFN_GATE_SHEXP,     "weight", i), { n_embd, hparams.n_ff_shexp }, 0);
+                        layer.ffn_up_shexp       = create_tensor(tn(LLM_TENSOR_FFN_UP_SHEXP,       "weight", i), { n_embd, hparams.n_ff_shexp }, 0);
+                        layer.ffn_down_shexp     = create_tensor(tn(LLM_TENSOR_FFN_DOWN_SHEXP,     "weight", i), { hparams.n_ff_shexp, n_embd }, 0);
```
CISC: `hparams.n_ff_shexp` is optionally loaded from metadata, but you have no fallback.

src/models/qwen35.cpp
```diff
@@ -0,0 +1,762 @@
...
+    auto * inp = build_inp_mem_hybrid();
+
+    ggml_tensor * inp_pos     = build_inp_pos();
+    ggml_tensor * inp_out_ids = build_inp_out_ids();
+
+    ggml_tensor * causal_mask =
+        ggml_tri(ctx0, ggml_fill_inplace(ctx0, ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, CHUNK_SIZE, CHUNK_SIZE), 1.0f),
+                    GGML_TRI_TYPE_LOWER);
+
+    ggml_tensor * identity = ggml_diag(ctx0, ggml_fill_inplace(ctx0, ggml_new_tensor_1d(ctx0, GGML_TYPE_F32, CHUNK_SIZE), 1.0f));
```
CISC:
```suggestion
        ggml_tri(ctx0, ggml_fill(ctx0, ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, CHUNK_SIZE, CHUNK_SIZE), 1.0f),
                    GGML_TRI_TYPE_LOWER);

    ggml_tensor * identity = ggml_diag(ctx0, ggml_fill(ctx0, ggml_new_tensor_1d(ctx0, GGML_TYPE_F32, CHUNK_SIZE), 1.0f));
```
Mentioned this in the other PR, but inplace should be automatic for `fill` now. :)

src/models/qwen35.cpp
```diff
@@ -0,0 +1,762 @@
...
+    ggml_tensor * Qcur = ggml_view_4d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens, 1,
+                                             Qcur_full->nb[1], Qcur_full->nb[2], Qcur_full->nb[3], 0);
+    ggml_tensor * gate =
+        ggml_view_4d(ctx0, Qcur_full, n_embd_head, n_head, n_tokens, 1,
+                     Qcur_full->nb[1], Qcur_full->nb[2], Qcur_full->nb[3], n_embd_head * ggml_element_size(Qcur_full));
+    cb(Qcur, "Qcur", il);
+    cb(gate, "gate", il);
+
+    // Now reshape Qcur to [n_embd_head, n_head, n_tokens] for multi-head attention
+    Qcur = ggml_cont_3d(ctx0, Qcur, n_embd_head, n_head, n_tokens);
```
CISC: Maybe review changes in the other PR, these can be simplified.

pwilkin: It's this commit for reference: https://github.com/ggml-org/llama.cpp/commit/8b0110921c879cf9e5ec748650af91a3b4bfc1c4

convert_hf_to_gguf.py
```diff
@@ -4538,6 +4545,53 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
         yield from super().modify_tensors(data_torch, name, bid)
 
 
+@ModelBase.register("Qwen3_5ForConditionalGeneration")
```
```suggestion
@ModelBase.register("Qwen3_5ForCausalLM")
```

convert_hf_to_gguf.py
```diff
@@ -4538,6 +4545,53 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        # Skip vision tensors - they go in the mmproj file
+        if name.startswith("model.visual."):
+            return
+
+        yield from super().modify_tensors(data_torch, name, bid)
+
+
+@ModelBase.register("Qwen3_5MoeForConditionalGeneration")
+class Qwen3_5MoeTextModel(Qwen3_5TextModel):
```
```suggestion
@ModelBase.register("Qwen3_5MoeForCausalLM")
class Qwen3_5MoeTextModel(Qwen3NextModel):
```

convert_hf_to_gguf.py
JJJYmmm: Qwen3.5 combines the text/vl ability, so I use the ConditionalGeneration class following the previous vl models.

src/models/qwen35.cpp
JJJYmmm: thanks!

gguf-py/gguf/tensor_mapping.py
JJJYmmm: oops, I accidentally added a redundant FFN_NORM and skipped the mapping. 🥹 Removing it now!

src/models/qwen35.cpp
JJJYmmm: done! 🫡

src/llama-model.cpp
JJJYmmm: qwen3.5 always provides n_ff_shexp, but add a fallback to n_ff following qwen2moe.🫡

src/llama-model.cpp
JJJYmmm: done! add FULL_ATTENTION_INTERVAL(=4) for qwen3next and 3.5🫡

ggerganov (approved):
The `libllama` changes are good. I will follow up with a PR to deduplicate and optimize the delta net stuff across the Qwen family.

convert_hf_to_gguf.py
```diff
@@ -4538,6 +4546,53 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
```
```suggestion
class Qwen3_5MoeTextModel(Qwen3NextModel):
```

src/models/qwen35.cpp
```diff
@@ -0,0 +1,754 @@
...
+        ggml_tensor * k_repeated =
+            ggml_repeat_4d(ctx0, k_reshaped, head_k_dim, repeat_factor, num_k_heads * n_seq_tokens * n_seqs, 1);
+
+        // Reshape back to merge the head and repeat dimensions
+        // From [head_dim, num_k_heads, repeat_factor, n_seq_tokens * n_seqs]
+        // Back to [head_dim, num_k_heads * repeat_factor, n_seq_tokens, n_seqs]
+        q_conv = ggml_reshape_4d(ctx0, q_repeated, head_k_dim, num_k_heads * repeat_factor, n_seq_tokens, n_seqs);
+        k_conv = ggml_reshape_4d(ctx0, k_repeated, head_k_dim, num_k_heads * repeat_factor, n_seq_tokens, n_seqs);
+    }
+
```
ggerganov, line 681: Btw, this is an issue inherited from the Qwen3 Next implementation. Basically we are doing redundant repeats here because the V weights are not arranged correctly during conversion. Ideally this whole section should not be needed and we should only rely on broadcasts.

Basically, we want to broadcast Q and K into V. With the current order of the V heads, we need an interleaved broadcast. However the ggml binary ops perform tiled broadcast. Hence the need for this rearrangement.

convert_hf_to_gguf.py
```diff
@@ -4538,6 +4546,53 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+@ModelBase.register("Qwen3_5ForConditionalGeneration")
+class Qwen3_5TextModel(Qwen3NextModel):
+    model_arch = gguf.MODEL_ARCH.QWEN35
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        # Skip vision tensors - they go in the mmproj file
+        if name.startswith("model.visual."):
+            return
+
+        yield from super().modify_tensors(data_torch, name, bid)
```
```suggestion
```
Move to `Qwen3NextModel`.

convert_hf_to_gguf.py
```diff
@@ -4538,6 +4546,53 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+            up = data_torch[..., split_dim:, :].contiguous()
+            base_name = name.removesuffix(".weight")
+            base = base_name.rsplit('.', 1)[0]
+            mapped_gate = f"{base}.gate_proj.weight"
+            mapped_up = f"{base}.up_proj.weight"
+            yield (self.map_tensor_name(mapped_gate), gate)
+            yield (self.map_tensor_name(mapped_up), up)
+            return
+        
+        yield from super().modify_tensors(data_torch, name, bid)
```
```suggestion
```
No longer necessary, just merged #19445

src/llama-arch.cpp
```diff
@@ -195,6 +197,7 @@ static const std::map<llm_kv, const char *> LLM_KV_NAMES = {
     { LLM_KV_EMBEDDING_SCALE,                   "%s.embedding_scale"                   },
     { LLM_KV_TOKEN_SHIFT_COUNT,                 "%s.token_shift_count"                 },
     { LLM_KV_INTERLEAVE_MOE_LAYER_STEP,         "%s.interleave_moe_layer_step"         },
+    { LLM_KV_FULL_ATTENTION_INTERVAL,            "%s.full_attention_interval"            },
```
CISC:
```suggestion
    { LLM_KV_FULL_ATTENTION_INTERVAL,           "%s.full_attention_interval"           },
```

CISC (approved):

CISC:
@JJJYmmm https://github.com/ggml-org/llama.cpp/actions/runs/21854651667/job/63068757680?pr=19468

src/models/qwen35.cpp
ggerganov: If it is too difficult to figure out how to fix this for the new Qwen3.5 models, we'll have to do it in a follow-up PR. But it's going to be a breaking change. Or maybe we can try to gate it with some flag.

convert_hf_to_gguf.py
JJJYmmm: done🫡

convert_hf_to_gguf.py
CISC: You forgot to add the `model.visual.` check to `Qwen3NextModel`.

JJJYmmm: I found it already in Qwen2Moe's `modify_tensors`.
https://github.com/ggml-org/llama.cpp/blob/262364e31d1da43596fe84244fba44e94a0de64e/convert_hf_to_gguf.py#L4140

CISC: Oh, indeed, I'm just blind. :)

src/models/qwen35.cpp
JJJYmmm: I would try to fix it in the qwen3.5 gguf conversion since it is not released.

CISC: Since it's a new arch we should not need a flag as such then.

JJJYmmm: just reorder the related v weights to avoid repeat_interleave🫡 https://github.com/ggml-org/llama.cpp/pull/19468/changes/4248c938001a8b208f46e3e7f4293bb8399b3540

ggerganov: Nice. Just make sure results are still as expected after this change.

From here, I think we will be able to remove the explicit repeats - this is something for next PR. Thanks a lot!

JJJYmmm: Yes, I tested it with the new generated weights, and it works just as before. 🫡

ngxson (approved):
The changes outside of cgraph looks good to me (sorry don't have too much time to look deeper, but I believe the review from @ggerganov and @CISC should already be enough)

Looking forward to the vision support PR then!

JJJYmmm:
> The changes outside of cgraph looks good to me (sorry don't have too much time to look deeper, but I believe the review from @ggerganov and @CISC should already be enough)
> 
> Looking forward to the vision support PR then!

@ngxson This pr includes the vision part and I test it too! (qwen3.5 uses the same vit as qwen3vl and deepstack was removed temporally)🤗

ngxson:
@JJJYmmm ah yeah right, thanks for the clarification. Yes I already acknowledge that the deep stack is removed (confirmed by `transformers` implementation)

Look good to me then, thanks again!

ngxson:
It seems to me that all review comments are resolved, and @JJJYmmm also confirmed that the model is still working fine with last commit. So, I'll let @ggerganov to decide when to merge this.

PR #19470, opened 2026-02-09
Offload embeddings for LFM2.

pzdunows:
Allow offloading of additional tensor.

Before:
```
  Device 0: AMD Radeon RX 7900 XTX, gfx1100 (0x1100), VMM: no, Wave Size: 32
| lfm2 1.2B Q8_0                 |   1.16 GiB |     1.17 B | ROCm       | 100 |           pp512 |    18271.27 ± 142.69 |
| lfm2 1.2B Q8_0                 |   1.16 GiB |     1.17 B | ROCm       | 100 |           tg128 |        346.34 ± 3.38 |
```
After:
```
  Device 0: AMD Radeon RX 7900 XTX, gfx1100 (0x1100), VMM: no, Wave Size: 32
| lfm2 1.2B Q8_0                 |   1.16 GiB |     1.17 B | ROCm       | 100 |           pp512 |    18657.22 ± 316.48 |
| lfm2 1.2B Q8_0                 |   1.16 GiB |     1.17 B | ROCm       | 100 |           tg128 |        357.86 ± 6.60 
```

It's around 3% improvement.


CISC:
To be clear, the only reason you're seeing the benefit you do is because the model has tied embeddings (many other models do) and that they are fairly large.

The proposed change is no good, not entirely sure if there is an acceptable way to do this, but if there was it would have to check for this specific scenario.

PR #19475, opened 2026-02-10
[model] add support for ERNIE 4.5 VL MoE

isLinXu:
This PR adds model support for ERNIE 4.5 family models from Baidu, including Dense (ernie4_5), MoE (ernie4_5-moe), and Vision-Language MoE (ernie4_5-vl-moe) variants. It has been verified with vision and pure text modes.

- Key architectural features
**Dual MoE (Text + Vision Experts)**: The core innovation of ERNIE 4.5 VL MoE is its dual expert system within the same LLM backbone. MoE layers maintain two separate sets of experts — one for text tokens and one for vision tokens — dynamically routed based on input modality. Vision experts use a significantly smaller FFN intermediate size (default 512) compared to text experts, reflecting a compact representation design for visual features.

**Interleaved Dense/MoE layers**: Controlled by n_layer_dense_lead and n_moe_layer_step, the first few layers are Dense, and MoE layers are interleaved at a configurable step interval. A shared expert (SwiGLU FFN) is added on top of MoE output for both modalities.

**ERNIE3D RoPE**: A new RoPE type (GGML_ROPE_TYPE_ERNIE3D = 72) designed for multimodal use, with an interleaved 3D frequency layout encoding height/width/temporal dimensions (sections [22, 22, 20, 0]), distinct from the contiguous segmentation used by standard M-RoPE.

**Vision Encoder**: Standard ViT with 2D M-RoPE (no learned positional embeddings), using SwiGLU FFN in each transformer layer.

**Vision Projector**: A spatial + temporal resampler pipeline:

2×2 spatial patch merging (4× token reduction)
Spatial linear path (Linear → GELU → Linear → LayerNorm)
Temporal path (optional, for video frames; single images use self-concatenation)
Final MLP + RMS Norm projection to LLM embedding space


src/llama-arch.h
```diff
@@ -1,602 +1,561 @@
...
+#include "llama-arch.h"
+#include "llama-graph.h"
+#include "llama-hparams.h"
+#include "llama-memory.h"
+#include "llama-vocab.h"
+
+#include <map>
+#include <memory>
 #include <string>
-#include <set>
```
pwilkin, line 6: Please cleanup the llama-arch.h changes - remove all the whitespace modifications.

pwilkin (changes requested):

ggml/include/ggml.h
```diff
@@ -252,6 +252,7 @@
 #define GGML_ROPE_TYPE_MROPE  8
 #define GGML_ROPE_TYPE_VISION 24
 #define GGML_ROPE_TYPE_IMROPE 40 // binary: 101000
+#define GGML_ROPE_TYPE_ERNIE3D 72 // binary: 1001000, ERNIE-VL 3D RoPE (NORMAL rotation + interleaved h/w freq)
```
ngxson, line 255: the ROPE_TYPE system is quite fragile now and I think we should always reflect twice before adding a new mode.

It seems like interleaved h/w freq is already supported by Pixtral model, please verify one more time if you can reuse the code from Pixtral instead of adding a new rope kernel here.

ngxson (changes requested):

src/llama-arch.h
isLinXu: fixed

tools/mtmd/clip-model.h
```diff
@@ -359,6 +364,23 @@ struct clip_model {
...
+    ggml_tensor * mm_spatial_norm_b = nullptr;
+    ggml_tensor * mm_temp_0_w       = nullptr;
+    ggml_tensor * mm_temp_0_b       = nullptr;
+    ggml_tensor * mm_temp_2_w       = nullptr;
+    ggml_tensor * mm_temp_2_b       = nullptr;
+    ggml_tensor * mm_temp_norm_w    = nullptr;
+    ggml_tensor * mm_temp_norm_b    = nullptr;
+    ggml_tensor * mm_mlp_w          = nullptr;
+    ggml_tensor * mm_mlp_b          = nullptr;
+    ggml_tensor * mm_after_norm_w   = nullptr;
```
ngxson: I don't think adding these tensors are needed.

Spatial patch merge is nothing new, we already support many models using the same strategy, please reuse the existing tensor naming and code infrastructure

tools/mtmd/models/ernie45vlmoe.cpp
```diff
@@ -0,0 +1,219 @@
...
+    // Loop over encoder layers
+    for (int il = 0; il < n_layer; il++) {
+        const auto &  layer = model.layers[il];
+        ggml_tensor * cur   = inpL;
+
+        // Layernorm 1
+        cur = build_norm(cur, layer.ln_1_w, layer.ln_1_b, NORM_TYPE_NORMAL, eps, il);
+        cb(cur, "ln1", il);
+
+        // Self-attention
```
ngxson: It seems like the whole loop can be replaced by `build_vit()`

ggml/include/ggml.h
isLinXu: Thanks for the heads-up. I completely agree that we should be cautious with the ROPE_TYPE system. I’ll re-examine the Pixtral implementation to see if we can reuse its interleaved frequency logic instead of adding a new kernel.

isLinXu: Thanks for the feedback. I’ve conducted a detailed mathematical comparison between Pixtral’s `build_rope_2d` and the ERNIE implementation. It turns out they are **mathematically incompatible**, and direct reuse would result in incorrect positional embeddings.

Below is the technical breakdown:

| **Feature**            | **Pixtral build_rope_2d**                    | **ERNIE (Vision / LLM)**                        |
| ---------------------- | -------------------------------------------- | ----------------------------------------------- |
| **Rotation Mode**      | **NORMAL** (Adjacent pairs)                  | **NEOX** (Half-dimension offset)                |
| **Freq. Allocation**   | **2-way Interleaved** (via `freq_scale_odd`) | **Sectional** (2D) / **3-way Interleaved** (3D) |
| **Theta Accumulation** | Continuous across the head                   | **Independent reset** per section               |
| **Dimensionality**     | 2D (h, w) only                               | **3D (t, h, w)**                                |
| **Implementation**     | Dual `rope_ext` + `concat`                   | `ggml_rope_multi` with mrope 4-slot             |

**Key Technical Differences:**

1. **Mathematical Incompatibility**: Pixtral uses **NORMAL** rotation, whereas ERNIE follows the **NEOX** convention (commonly used in Vision Transformers). Since the pairing of dimensions differs, swapping them would break the model's spatial understanding.
2. **Frequency Mapping**: Pixtral achieves interleaved frequencies by applying a `freq_scale` to one-half of the dimensions. ERNIE uses `sections [20, 20, 0, 0]` to strictly block frequencies, where each section starts its theta accumulation independently from $base^0$.

**Regarding the complexity of the `ROPE_TYPE` system:**

- **Vision Side**: We are actually using the **existing** `GGML_ROPE_TYPE_VISION`. No new mode is introduced here.
- **LLM Side**: The new `GGML_ROPE_TYPE_ERNIE3D` is a strict requirement to support the **Temporal (t) dimension**. Current 2D implementations (like Pixtral) cannot handle this 3D mapping.

**Conclusion:**

To maintain mathematical correctness and support 3D RoPE, we cannot reuse the Pixtral logic. The new `ERNIE3D` type is the minimum necessary change to support these specific requirements. I will ensure the implementation is as modular as possible to keep the system maintainable.

tools/mtmd/clip-model.h
isLinXu: fixed

ggml/include/ggml.h
ngxson: If the difference is just the normal and neox style, you can also permute the Q and K tensor upon converting to GGUF.

Kimi 2.5 also do exactly this, you can copy the conversion code from https://github.com/ggml-org/llama.cpp/pull/19170

ngxson: Also just a friendly reminder: We don't allow replying to human maintainer with AI-generated response. Please write the response with your own writing,to prove that you fully understand your code

tools/mtmd/models/ernie45vlmoe.cpp
isLinXu: done

tools/mtmd/models/ernie45vlmoe.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    // GELU
+    spatial_out = ggml_gelu(ctx0, spatial_out);
+    cb(spatial_out, "spatial_gelu", -1);
+
+    // Second linear
+    ggml_tensor * spatial_2_w = ggml_cont(ctx0, ggml_transpose(ctx0, model.mm_2_w));
+    spatial_out = ggml_mul_mat(ctx0, spatial_2_w, spatial_out);
+    spatial_out = ggml_add(ctx0, spatial_out, model.mm_2_b);
+    cb(spatial_out, "spatial_linear_2", -1);
```
ngxson: this can be reduced to build_ffn

tools/mtmd/models/ernie45vlmoe.cpp
```diff
@@ -0,0 +1,120 @@
...
+    embeddings = build_patch_merge_permute(embeddings, spatial_conv_size);
+    cb(embeddings, "spatial_reshape", -1);
+
+    // Spatial linear path: Linear -> GELU -> Linear -> LayerNorm
+    // Note: weights were transposed (.t()) during GGUF conversion, so we must
+    // undo that with ggml_transpose before ggml_mul_mat
+    ggml_tensor * spatial_out = embeddings;
+
+    // First linear
+    ggml_tensor * spatial_0_w = ggml_cont(ctx0, ggml_transpose(ctx0, model.mm_0_w));
```
ngxson: any transposes to the weight must be done upon conversion

tools/mtmd/models/ernie45vlmoe.cpp
```diff
@@ -0,0 +1,120 @@
...
+    ggml_tensor * resampler_out = spatial_out;
+
+    // Temporal processing for single images (t=1):
+    // Following ERNIE-VL original: when t=1, slice_offsets and slice_offsets2 both point to the same frame
+    // So we concat(x, x, dim=-1) which in GGML's [hidden, seq] layout is dim=0
+    // This doubles the hidden dimension: [5120, 400] -> [10240, 400]
+    resampler_out = ggml_concat(ctx0, resampler_out, resampler_out, 0);
+
+    // Temporal linear path: Linear -> GELU -> Linear -> LayerNorm
+    // Weights were transposed (.t()) during GGUF conversion, undo with ggml_transpose
```
ngxson: hmm, why transpose it when conversion, then having to transpose it again here?

unless my math is broken somehow, `transpose(transpose(A))` is identical to just use `A` without any transposes

tools/mtmd/models/ernie45vlmoe.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    // GELU
+    resampler_out = ggml_gelu(ctx0, resampler_out);
+    cb(resampler_out, "temporal_gelu", -1);
+
+    // Second temporal linear
+    ggml_tensor * temp_2_w = ggml_cont(ctx0, ggml_transpose(ctx0, model.mm_3_w));
+    resampler_out = ggml_mul_mat(ctx0, temp_2_w, resampler_out);
+    resampler_out = ggml_add(ctx0, resampler_out, model.mm_3_b);
+    cb(resampler_out, "temporal_linear_2", -1);
```
ngxson: again, just build_ffn

src/models/ernie4-5-vl-moe.cpp
```diff
@@ -0,0 +1,185 @@
...
+            cb(cur, "ffn_out", il);
+        } else {
+            // MoE branch
+            cur = build_norm(ffn_inp, model.layers[il].ffn_norm, NULL, LLM_NORM_RMS, il);
+            cb(cur, "ffn_norm", il);
+
+            ggml_tensor * moe_out = nullptr;
+
+            // Use vision experts for vision tokens, text experts for text tokens
+            if (ubatch.embd) {
```
ngxson, line 115: This may make the graph to be non-static, thus reduce the overall performance.

Instead, you should follow the same optimization with `ggml_build_forward_select`, see the example in `llm_graph_context::build_inp_embd`

src/llama-model.cpp
```diff
@@ -2195,6 +2196,34 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+                ml.get_key(LLM_KV_VISION_EXPERT_FEED_FORWARD_LENGTH, hparams.n_ff_v_exp, false);
+                ml.get_key(LLM_KV_EXPERT_SHARED_FEED_FORWARD_LENGTH, hparams.n_ff_shexp, false);
+                ml.get_key(LLM_KV_INTERLEAVE_MOE_LAYER_STEP,         hparams.n_moe_layer_step);
+                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,         hparams.n_layer_dense_lead);
+                if (!ml.get_key_or_arr(LLM_KV_ROPE_DIMENSION_SECTIONS, hparams.rope_sections, 4, false)) {
+                    hparams.rope_sections[0] = 22;
+                    hparams.rope_sections[1] = 22;
+                    hparams.rope_sections[2] = 20;
+                    hparams.rope_sections[3] = 0;
+                }
```
ngxson, line 2212: since we are not trying to be backward-compatible here, I think it's better not to hard code any default values. this metadata `LLM_KV_ROPE_DIMENSION_SECTIONS` must be a requirement

src/llama-model.cpp
```diff
@@ -2195,6 +2196,34 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
+
+                LLAMA_LOG_INFO("%s: ERNIE-VL rope_sections=[%d,%d,%d,%d]\n", __func__,
```
ngxson, line 2214: logging must not be done here

src/llama-model.cpp
```diff
@@ -8617,6 +8865,8 @@ llama_rope_type llama_model_rope_type(const llama_model * model) {
             return model->hparams.use_mrope() ? LLAMA_ROPE_TYPE_MROPE : LLAMA_ROPE_TYPE_NORM;
         case LLM_ARCH_GLM4_MOE:
             return model->hparams.use_mrope() ? LLAMA_ROPE_TYPE_MROPE : LLAMA_ROPE_TYPE_NEOX;
+        case LLM_ARCH_ERNIE4_5_VL_MOE:
+            return LLAMA_ROPE_TYPE_ERNIE3D;
```
ngxson, line 8869: I wonder how this differ from qwen's IMROPE (the interleaved mrope version added in Qwen 3)

In any cases, I'm pretty sure that we don't need to add any new rope modes. Any more complex mode can be implemented in 2 ways (or combination of 2):
- Permute Q and K tensor upon conversion
- Frequency per-dimension can be controlled by the custom `freq_factors`, i.e. the `c` tensor in `ggml_rope_ext`

CSWYF3634076: @ngxson @isLinXu 
Qwen2VL is ttt...hhh...www..
Qwen3VL is thwthw...thw..
Ernie4.5VL is hwhwhw...ttt...

ngxson: permute Q and K to Qwen2 ordering, then use `freq_factors` to correct the `theta_base`

I don't think we can allow adding yet another rope mode here, this part of the code is already fragile enough

isLinXu:
- Fix generator function issues (return -> yield)
- Remove ERNIE-specific clip_hparams fields
- Reuse existing n_merge field instead of spatial_conv_size
- Use standard tensor naming (TN_LLAVA_PROJ)
- Use build_ffn() and build_norm() functions
- No new RoPE type added (uses GGML_ROPE_TYPE_VISION)"
@ngxson @pwilkin

segmond:
Can we get some action on this.  @isLinXu Do you have any gguf models you can upload?  I can help test.

isLinXu:
> Can we get some action on this. @isLinXu Do you have any gguf models you can upload? I can help test.
https://huggingface.co/ERNIE-Community/ERNIE-4.5-VL-28B-A3B-Thinking-F16-gguf，thinks

segmond:
I tried to build it

/home/seg/llama.box/ernie/llama.ernie-vl4.5/ggml/src/ggml-cuda/rope.cu(691): error: identifier "is_ernie3d" is undefined

1 error detected in the compilation of "/home/seg/llama.box/ernie/llama.ernie-vl4.5/ggml/src/ggml-cuda/rope.cu".
gmake[2]: *** [ggml/src/ggml-cuda/CMakeFiles/ggml-cuda.dir/build.make:707: ggml/src/ggml-cuda/CMakeFiles/ggml-cuda.dir/rope.cu.o] Error 2
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:1904: ggml/src/ggml-cuda/CMakeFiles/ggml-cuda.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2

seg@seg-epyc:~/llama.box/ernie/llama.ernie-vl4.5$ cat ~/bin/buildllama.sh 
#!/bin/bash
cmake -B build -DGGML_CUDA=ON -DGGML_RPC=ON -DGGML_CUDA_FA_ALL_QUANTS=ON -DBUILD_SHARED_LIBS=ON -DLLAMA_CURL=OFF -DLLAMA_OPENSSL=OFF
cmake --build build --config Release -j 8

src/llama-model.cpp
ngxson: > Ernie4.5VL is hwhwhw...ttt...

what you can do is to split the tensor in 2 parts, first part for hwhw... and second part for ttt....
- apply mrope on first part
- apply normal rope (or neox rope) on later part

refer to this docs for more: https://github.com/ggml-org/llama.cpp/blob/master/docs/development/HOWTO-add-model.md#working-with-ggml_rope_ext

ngxson:
> * No new RoPE type added

You still have a new rope mode added for text model. It should be removed from this PR.

Unless my comment https://github.com/ggml-org/llama.cpp/pull/19475#discussion_r3145575717 is resolved, I don't think we can proceed with this PR.

isLinXu:
> > * No new RoPE type added
> 
> You still have a new rope mode added for text model. It should be removed from this PR.
> 
> Unless my comment [#19475 (comment)](https://github.com/ggml-org/llama.cpp/pull/19475#discussion_r3145575717) is resolved, I don't think we can proceed with this PR.
> 
> 
> > 未* 添加新的RoPE类型
> 
> 您仍然为文本模型添加了新的rope模式。它应该从这个PR中删除。
> 
> 除非
> 我的评论[#19475（评论）](https://github.com/ggml-org/llama.cpp/pull/19475#discussion_r3145575717)得到解决，否则我认为我们无法继续进行此PR。

I'll do another pass on this PR. There might be some workspace contamination causing confusion — re-submitting after a clean review sounds like a reasonable path forward.

ngxson:
Yes it's ok to submit a new PR. I'm closing this one for now. Thanks for your understanding

PR #19479, opened 2026-02-10
model : fix wavtokenizer embedding notions

ggerganov:
alt #19462

The old `hparams.n_embd_features` was redundant since it was introduced before the `n_embd_inp`, `n_embd` and `n_embd_out` convention. Adapt the WavTokenizer to the new logic.


danbev (approved):
Verified locally that `llama-tts` works as expected  👍

PR #16991, opened 2025-11-04
CUDA: add stream-based concurrency

am17an:
Possibly supersede #16813. 

This PR adds support to run concurrent CUDA streams on single GPU setups. 
At the moment this only targets the Q, K, V branch. I feel this is the "correct" approach in case the Q, K, V tensors are of different types/not in the same place in memory. The downside is that this approach doesn't come for free and there's some complexity involved, but I'm not an expert at the ggml graph and I feel it could be simplified. 

Currently this is hidden by an env variable flag. To run you can use `GGML_CUDA_GRAPH_OPT=1`

TG Performance gain is more than the previous PR (1-9% gain depending on the model/GPU), probably because we parallelize MUL_MAT + NORM + ROPE rather than just MUL_MAT. At the moment we leave some performance on the table where we don't fuse operations in the parallel streams themselves (e.g. MUL_MAT + BIAS, RMS_NORM + MUL etc.), I couldn't find a simple enough way to enable fusion there.

Performance details:

<details> 

5090:

| Model                   | Test   |   t/s mxfp4 |   t/s 0fa5630c0 |   Speedup |
|:------------------------|:-------|------------:|----------------:|----------:|
| gpt-oss 20B MXFP4 MoE   | pp512  |    10554.76 |        10572.61 |      1.00 |
| gpt-oss 20B MXFP4 MoE   | tg128  |      359.67 |          382.52 |      1.06 |
| llama 8B Q8_0           | pp512  |    13948.06 |        13903.74 |      1.00 |
| llama 8B Q8_0           | tg128  |      166.44 |          174.08 |      1.05 |
| qwen3 8B F16            | pp512  |    14120.38 |        14170.43 |      1.00 |
| qwen3 8B F16            | tg128  |       97.21 |          100.39 |      1.03 |
| qwen3moe 30B.A3B Q4_K_M | pp512  |     7361.89 |         7367.90 |      1.00 |
| qwen3moe 30B.A3B Q4_K_M | tg128  |      306.84 |          334.19 |      1.09 |

4090:

| Model                   | Test   |   t/s mxfp4 |   t/s 0fa5630c0 |   Speedup |
|:------------------------|:-------|------------:|----------------:|----------:|
| gpt-oss 20B MXFP4 MoE   | pp512  |     9551.59 |         9549.55 |      1.00 |
| gpt-oss 20B MXFP4 MoE   | tg128  |      263.77 |          271.84 |      1.03 |
[...]

PR #19481, opened 2026-02-10
llama: fuse QKV weights into single matmul for LLaMA models (ROCm/HIP)

JoursBleu:
## Summary

Fuse Q/K/V weight matrices into a single concatenated QKV tensor at model load time for LLaMA architectures on ROCm/HIP, reducing 3 `mul_mat` kernel launches to 1 per transformer layer during inference.

## Motivation

On NVIDIA GPUs, [PR #16991](https://github.com/ggml-org/llama.cpp/pull/16991) uses CUDA stream concurrency to overlap Q/K/V matmuls via CUDA Graphs with multi-stream event synchronization. However, on AMD GPUs (ROCm/HIP), HIP Graph replay with multi-stream event synchronization incurs **14× overhead** on `hipGraphLaunch` (0.178ms → 2.498ms), making stream-based concurrency counterproductive.

QKV weight fusion provides an alternative approach for ROCm: instead of parallelizing 3 kernel launches, it eliminates 2 of them entirely by concatenating the weight matrices and using a single larger `mul_mat`.

### `src/llama-model.cpp`
- After tensor loading, if `GGML_USE_HIP` is defined and the architecture is `LLM_ARCH_LLAMA` or `LLM_ARCH_LLAMA_EMBED`:
  - Concatenates `wq`, `wk`, `wv` row-wise into a single `wqkv` tensor (`[n_embd, n_out_q + n_out_k + n_out_v]`)
  - Copies the fused tensor to device memory via a new `ggml_backend_buffer`
  - Stores the result in the existing `layer.wqkv` field

### `src/models/llama.cpp`
- Guarded by `#if defined(GGML_USE_HIP)`:
  - If `layer.wqkv` exists, performs a single `ggml_mul_mat` followed by `ggml_view_3d` to split Q/K/V outputs
  - Otherwise falls through to the original separate `build_lora_mm(wq/wk/wv)` path
- The original path is always compiled and serves as the fallback on all platforms

## Scope

- **ROCm/HIP only**: All new code is wrapped in `#if defined(GGML_USE_HIP)`. Zero impact on CUDA, Metal, CPU, or other backends.
- **Architecture**: Currently limited to LLaMA / LLaMA-Embed models.
- **No new struct fields**: Reuses the existing `wqkv` pointer in `llama_layer`.

## Benchmark

**Hardware**: AMD Radeon AI PRO R9700 (gfx1201, RDNA4), ROCm 6.3  
**Model**: Llama 2 7B Q4_0, single GPU, `tg128`, HIP Graphs enabled

| Configuration | tok/s | vs Master |
|---|---|---|
| Master baseline | 95.00 | — |
| **QKV Fusion (this PR)** | **100.85** | **+6.2%** |

### DIFF
--- a/src/llama-model.cpp
+++ b/src/llama-model.cpp
@@ -7390,6 +7390,95 @@ bool llama_model::load_tensors(llama_model_loader & ml) {
         }
     }
 
+#if defined(GGML_USE_HIP)
+    // QKV weight fusion: concat wq/wk/wv into wqkv for supported architectures
+    // This reduces 3 kernel launches to 1 per layer during inference
+    if (arch == LLM_ARCH_LLAMA || arch == LLM_ARCH_LLAMA_EMBED) {
+        const uint32_t n_layer_count = hparams.n_layer;
+
+        LLAMA_LOG_INFO("%s: fusing QKV weights for %u layers\n", __func__, n_layer_count);
+
+        for (uint32_t il = 0; il < n_layer_count; ++il) {
+            auto & layer = layers[il];
+
+            // skip if already fused or if any weight is missing
+            if (layer.wqkv || !layer.wq || !layer.wk || !layer.wv) {
+                continue;
+            }
+
+            // skip if types don't match (safety check)
+            if (layer.wq->type != layer.wk->type || layer.wq->type != layer.wv->type) {
+                LLAMA_LOG_WARN("%s: layer %u: QKV type mismatch, skipping fusion\n", __func__, il);
+                continue;
+            }
+
+            const int64_t n_embd     = layer.wq->ne[0]; // inner dimension (K)
+            const int64_t n_out_q    = layer.wq->ne[1]; // Q output rows
+            const int64_t n_out_k    = layer.wk->ne[1]; // K output rows
+            const int64_t n_out_v    = layer.wv->ne[1]; // V output rows
+            const int64_t n_out_qkv  = n_out_q + n_out_k + n_out_v;
+            const ggml_type wtype    = layer.wq->type;
+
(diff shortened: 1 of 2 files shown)


JoursBleu:
Hi @CISC, could you share the reason for closing this PR? I'd like to understand if there are specific concerns so I can address them or take a different direction. Thanks!

CISC:
> Hi @CISC, could you share the reason for closing this PR? I'd like to understand if there are specific concerns so I can address them or take a different direction. Thanks!

As such your previous PR made more sense (we already have plans in that direction for all models, though may not be at load), but this PR gated for HIP only makes no sense.

JoursBleu:
> > Hi @CISC, could you share the reason for closing this PR? I'd like to understand if there are specific concerns so I can address them or take a different direction. Thanks!
> 
> As such your previous PR made more sense (we already have plans in that direction for all models, though may not be at load), but this PR gated for HIP only makes no sense.

Hi @CISC, thanks for the explanation!

The reason I proposed the HIP-specific approach was that #16991 (stream-based concurrency) superseded #16813 (fused QKV), which led me to believe the project preferred stream concurrency over weight fusion for CUDA. However, HIP Graph replay with multi-stream event synchronization has a 14× overhead on `hipGraphLaunch`, making stream concurrency counterproductive on AMD GPUs. So I gated the fusion for HIP only as an alternative.

Given your feedback, should I reopen #19477 (the non-gated version) instead?

CISC:
> Given your feedback, should I reopen #19477 (the non-gated version) instead?

I don't think that is necessary at this point, as mentioned an alternative approach is currently planned, read the thread in #19139

PR #19495, opened 2026-02-10
jinja: add missing tojson filter for undefined type

jkkj:
jinja: add missing tojson filter for undefined type                                                                                                                                                                                                                                                                                                                                                                                                                               

value_undefined_t::get_builtins() was the only type missing the
tojson filter, causing "unknown filter 'tojson' for type Undefined"
when templates reference undefined variables (e.g. tools | tojson
with no tools defined).

### DIFF
--- a/common/jinja/value.cpp
+++ b/common/jinja/value.cpp
@@ -1096,6 +1096,7 @@ const func_builtins & value_undefined_t::get_builtins() const {
         {"strip", empty_value_fn<value_string>},
         {"sum", empty_value_fn<value_int>},
         {"title", empty_value_fn<value_string>},
+        {"tojson", tojson},
         {"truncate", empty_value_fn<value_string>},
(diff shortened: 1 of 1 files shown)


pwilkin:
Please add a test for it to `test-jinja.cpp` and test with `test-jinja -py` to verify whether the behavior is consistent with the Python reference implementation.

CISC:
I removed the `tojson` filter, `undefined` does not support it in `jinja2`.

CISC:
> value_undefined_t::get_builtins() was the only type missing the tojson filter, causing "unknown filter 'tojson' for type Undefined" when templates reference undefined variables (e.g. tools | tojson with no tools defined).

Any template relying on this is broken, however it is also likely that there is some other issue that caused the variable to be `undefined` so please link any template in question.

PR #19503, opened 2026-02-11
grammar : fix grammar trigger crash when token extends beyond trigger pattern

### LINKED ISSUE #19353: Eval bug: crash in llama_grammar_accept_token
### Name and Version

```
./build/bin/llama-cli --version
version: 7946 (7a4f97d19)
built with Clang 21.1.7 for Linux x86_64
```

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

Ryzen 5950X + NVIDIA RTX 3090

### Models

Qwen-Coder-Next-80B-A3B (unsloth UD-Q6-XL)

```
./build/bin/llama-server --jinja --model ~/llama/Qwen3-Coder-Next-GGUF/UD-Q6_K_XL/Qwen3-Coder
-Next-UD-Q6_K_XL-00001-of-00002.gguf --threads -1 -fa 1 -ctv q8_0 -ctk q8_0 --host 0.0.0.0 --temp 1.0 --top-k 40 --top-p 0.95 --min-p 0.01 -c 131072 --no-direct-io -fitt 384
```

### First Bad Commit

_No response_

### Relevant log output

```
slot print_timing: id  3 | task 0 |
prompt eval time =   57251.21 ms / 10634 tokens (    5.38 ms per token,   185.74 tokens per second)
       eval time =    4467.24 ms /    33 tokens (  135.37 ms per token,     7.39 tokens per second)
      total time =   61718.45 ms / 10667 tokens
slot      release: id  3 | task 0 | stop processing: n_tokens = 10666, truncated = 0
srv  log_server_r: done request: POST /v1/chat/completions 192.168.2.218 200
srv  params_from_: Chat format: Qwen3 Coder
slot get_availabl: id  3 | task -1 | selected slot by LCP similarity, sim_best = 0.914 (> 0.100 thold), f_keep = 1.000
slot launch_slot_: id  3 | task -1 | sampler chain: logits -> ?penalties -> ?dry -> ?top-n-sigma -> top-k -> ?typical -> ?to
p-p -> min-p -> ?xtc -> ?temp-ext -> dist
slot launch_slot_: id  3 | task 40 | processing task, is_child = 0
slot update_slots: id  3 | task 40 | new prompt, n_ctx_slot = 131072, n_keep = 0, task.n_tokens = 11675
slot update_slots: id  3 | task 40 | n_tokens = 10666, memory_seq_rm [10666, end)
slot update_slots: id  3 | task 40 | prompt processing progress, n_tokens = 11611, batch.n_tokens = 945, progress = 0.994518
slot update_slots: id  3 | task 40 | n_tokens = 11611, memory_seq_rm [11611, end)
slot update_slots: id  3 | task 40 | prompt processing progress, n_tokens = 11675, batch.n_tokens = 64, progress = 1.000000
slot update_slots: id  3 | task 40 | prompt done, n_tokens = 11675, batch.n_tokens = 64
slot init_sampler: id  3 | task 40 | init sampler, took 1.06 ms, tokens: text = 11675, total = 11675
slot update_slots: id  3 | task 40 | created context checkpoint 2 of 8 (pos_min = 11610, pos_max = 11610, size = 75.376 MiB)
```

```
0x0000737214b10813 in __GI___wait4 (pid=630236, stat_loc=0x0, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
warning: 30     ../sysdeps/unix/sysv/linux/wait4.c: No such file or directory
#0  0x0000737214b10813 in __GI___wait4 (pid=630236, stat_loc=0x0, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.
c:30
30      in ../sysdeps/unix/sysv/linux/wait4.c
#1  0x00007372179a14f6 in ggml_print_backtrace () from /home/morbo/git/llama.cpp/build/bin/libggml-base.so.0
#2  0x00007372179b6de6 in ggml_uncaught_exception() () from /home/morbo/git/llama.cpp/build/bin/libggml-base.so.0
#3  0x0000737214ebb0da in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x0000737214ea5a55 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x0000737214ebb391 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x0000737217cb5571 in llama_grammar_accept_token(llama_grammar&, int, std::__cxx11::basic_string<char, std::char_traits<
char>, std::allocator<char> > const&) () from /home/morbo/git/llama.cpp/build/bin/libllama.so.0
#7  0x0000737217cb4b0f in llama_grammar_accept_impl(llama_grammar&, int) () from /home/morbo/git/llama.cpp/build/bin/libllam
a.so.0
#8  0x00005e31b24060b9 in common_sampler_accept(common_sampler*, int, bool) ()
#9  0x00005e31b21b0a39 in server_context_impl::update_slots() ()
#10 0x00005e31b22645cc in server_queue::start_loop(long) ()
#11 0x00005e31b20e7498 in main ()
```

EliasOenal:
This fixes a crash when the model emits a tool call that a) completes the grammar trigger pattern, but also b) contains invalid extra text in the same token. An example would be: we're at `<function` in the buffer and match the prefix of the trigger pattern `<function=`. If then the next token comes in as `=list`, the `=` matches and completes the trigger, while the `list` part turns it into an invalid tool call, within the same token. (For the example we assume this tool is not available) I believe some models have per-tool-name triggers for their tool calls and are not susceptible to the bug, but the Qwen3 models and a few others are.

* This PR fixes the main issue in `src/llama-grammar.cpp`, by gracefully trying and failing calls to hallucinated tools. The resulting output will most likely be an invalid tool call to the client, but that is what the model generates.

* As a complementary improvement it adds basic exception handling to `tools/server/server-context.cpp`. This was missing and in turn resulted in llama-server crashing.

Resolves #19353 and resolves #19304.

Minimal reproducer:
```
#!/usr/bin/env python3
"""Reproducer: grammar crash when a token completes a trigger AND contains
text beyond it (e.g. token "=list" completing trigger "<function=").
Requires a running llama-server with a Qwen3 model (e.g. Qwen3-4B)."""
import requests, sys, time

URL = "http://127.0.0.1:8080"
PAYLOAD = {
    "prompt": """<|im_start|>system
You must use the tool to answer.
<tools><function><name>list</name><description>List files</description>
<parameters><parameter><name>path</name><type>string</type></parameter>
<required>["path"]</required></parameters></function></tools>
Reply format: <tool_call>\n<function=name>\n<parameter=p>v</parameter>\n</function>\n</tool_call><|im_end|>
<|im_start|>user
List /tmp<|im_end|>
<|im_start|>assistant
""",
    # grammar only allows specific tool names — "list" is intentionally missing
    "grammar": 'root ::= "<tool_call>\\n<function=" ("bash"|"search") ">" [^<]* "</function>\\n</tool_call>"',
    "grammar_lazy": True,
    "grammar_triggers": [{"type": 2, "value": r"<tool_call>\n<function="}],
    "n_predict": 256, "temperature": 0.7,
}

for i in range(1, 6):
    try:
        r = requests.post(f"{URL}/completion", json=PAYLOAD, timeout=60)
        print(f"{i}: {r.json().get('content', '')[:100]}")
    except requests.exceptions.ConnectionError:
        print(f"{i}: SERVER CRASHED"); sys.exit(1)
print("No crash after 5 attempts.")
```
[repro_min.py](https://github.com/user-attachments/files/25226109/repro_min.py)

### DIFF
--- a/src/llama-grammar.cpp
+++ b/src/llama-grammar.cpp
@@ -968,7 +968,7 @@ llama_grammar_stacks & llama_grammar_get_stacks(struct llama_grammar * grammar) 
 }
 
 static void llama_grammar_accept_chr(
-        struct llama_grammar       & grammar,
+        const llama_grammar_rules  & rules,
         const llama_grammar_stack  & stack,
               uint32_t               chr,
               llama_grammar_stacks & new_stacks) {
@@ -989,7 +989,7 @@ static void llama_grammar_accept_chr(
         if (!llama_grammar_is_end_of_sequence(match.second)) {
             new_stack.push_back(match.second);
         }
-        llama_grammar_advance_stack(grammar.rules, new_stack, new_stacks);
+        llama_grammar_advance_stack(rules, new_stack, new_stacks);
     }
 }
 
@@ -998,7 +998,7 @@ void llama_grammar_accept(struct llama_grammar * grammar, uint32_t chr) {
     stacks_new.reserve(grammar->stacks.size());
 
     for (const auto & stack : grammar->stacks) {
-        llama_grammar_accept_chr(*grammar, stack, chr, stacks_new);
+        llama_grammar_accept_chr(grammar->rules, stack, chr, stacks_new);
     }
 
     grammar->stacks = std::move(stacks_new);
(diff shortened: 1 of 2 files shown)


dstolpmann:
This resolves #19304 for me. Thank you!

tamascode:
@EliasOenal 

The PR fixes the crash — llama-server no longer terminates when the grammar issue occurs.

However, the underlying grammar failure still happens intermittently during tool-calling with **Qwen3 Coder Next**. The request now fails gracefully with:

```
Grammar error: Unexpected empty grammar stack after accepting piece: =read
```

Observed behavior:

* prompt processes normally
* generation starts
* model emits a tool-call piece (`=read`)
* grammar stack becomes empty
* request returns an error
* server stays alive and releases the slot correctly

So this appears to be:

* **crash fixed**
* **grammar/tool-call mismatch still present**

Environment:

* llama-server (`/v1/chat/completions`)
* Qwen3 Coder chat format
* tool calling / grammar-constrained decoding enabled
* large prompt (~43k tokens)

---

EliasOenal:
@tamascode I was looking into hallucinated invalid tool calls as well, but I didn't want to introduce major changes to the llama.cpp codebase for my first PR. Thus I focused on fixing the crash, which should be a straight improvement.

To me it seems to be a deliberate decision to only activate the "lazy grammar"-path after the tool name has been completed. And the issue is that those Qwen3 models tend to occasionally hallucinate invalid tools they call. At least for me, with my fix applied and OpenCode, the failed tool call informs the model, which in turn usually gets it right on the second try.

I believe it would be possible to trigger the grammar earlier, to force models to only emit calls to valid tools, but that doesn't seem to be the design goal of "lazy grammar". I am happy to further look into this, if I could get some guidance on would be the best fit for the project.

@ngxson I know #18675 may also address the current llama-server crash, but it seems like a larger undertaking with a potentially longer timeline. Given that Qwen3 Coder Next is a vastly popular model, and many people are facing crashes: do you think it makes sense to merge this targeted fix to make the model work, or would you prefer to wait for the autoparser to get merged instead?

aldehir:
The pattern should trigger before a tool name is generated, to ensure the grammar constrains model output to valid tool calls. The fix here is too invasive.

I rather roll out a separate custom Qwen 3 Coder Next parser, with the proper trigger rules, until the autoparser PR is merged. Could also be as simple as changing the pattern to look for `<function` instead of `<function=`, if `=` is not part of the token.

EliasOenal:
Triggering on `<function` would only reliably fix this, if `function` never tokenizes with any appendages. And I'm not convinced this is guaranteed. The current crash is a denial of service issue. Any user can take the server down by manipulating the model to emit the right token sequence (see reproducer). And it seemed to me like there were additional code paths that may throw as well. llama-server just wasn't handling the exceptions at all.

PR #19509, opened 2026-02-11
vulkan: split mul_mat into multiple dispatches to avoid overflow

jeffbolznv:
The batch dimensions can be greater than the max workgroup count limit, in which case we need to split into multiple dispatches and pass the base index through a push constant.

Fall back for the less common p021 and nc variants.

Fixes #19471.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -6768,11 +6768,11 @@ static void ggml_vk_matmul(
         uint32_t padded_n) {
         VK_LOG_DEBUG("ggml_vk_matmul(a: (" << a.buffer->buffer << ", " << a.offset << ", " << a.size << "), b: (" << b.buffer->buffer << ", " << b.offset << ", " << b.size << "), d: (" << d.buffer->buffer << ", " << d.offset << ", " << d.size << "), split_k: (" << (split_k_buffer.buffer != nullptr ? split_k_buffer.buffer->buffer : VK_NULL_HANDLE) << ", " << split_k_buffer.offset << ", " << split_k_buffer.size << "), m: " << m << ", n: " << n << ", k: " << k << ", stride_a: " << stride_a << ", stride_b: " << stride_b << ", stride_d: " << stride_d << ", batch_stride_a: " << batch_stride_a << ", batch_stride_b: " << batch_stride_b << ", batch_stride_d: " << batch_stride_d << ", split_k: " << split_k << ", batch: " << batch << ", ne02: " << ne02 << ", ne12: " << ne12 << ", broadcast2: " << broadcast2 << ", broadcast3: " << broadcast3 << ", padded_n: " << padded_n << ")");
     if (split_k == 1) {
+        ggml_pipeline_request_descriptor_sets(ctx, pipeline, CEIL_DIV(batch, ctx->device->properties.limits.maxComputeWorkGroupCount[2]));
+
         uint32_t base_work_group_z = 0;
         while (base_work_group_z < batch) {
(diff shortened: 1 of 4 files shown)


jeffbolznv:
The new tests are failing on multiple backends, I'll move them to a separate PR so this isn't blocked.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -6781,9 +6791,16 @@ static void ggml_vk_matmul(
...
-    ggml_vk_dispatch_pipeline(ctx, subctx, pipeline, { a, b, split_k_buffer }, pc1, { (CEIL_DIV(m, pipeline->wg_denoms[0]) * pipeline->wg_denoms[0]) * split_k, n, batch });
+    uint32_t base_work_group_z = 0;
+    while (base_work_group_z < batch) {
+        uint32_t groups_z = std::min(batch - base_work_group_z, ctx->device->properties.limits.maxComputeWorkGroupCount[2]);
+
+        ggml_pipeline_request_descriptor_sets(ctx, pipeline, 1);
+
+        const vk_mat_mat_push_constants pc1 = { m, n, k, stride_a, stride_b, stride_d, batch_stride_a, batch_stride_b, batch_stride_d, base_work_group_z, batch, k_split, ne02, ne12, broadcast2, broadcast3, padded_n };
+        // Make sure enough workgroups get assigned for split k to work
+        ggml_vk_dispatch_pipeline(ctx, subctx, pipeline, { a, b, split_k_buffer }, pc1, { (CEIL_DIV(m, pipeline->wg_denoms[0]) * pipeline->wg_denoms[0]) * split_k, n, groups_z });
```
0cc4m, line 6802: You forgot to increase `base_work_group_z` here.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -6766,8 +6768,16 @@ static void ggml_vk_matmul(
...
     if (split_k == 1) {
-        const vk_mat_mat_push_constants pc = { m, n, k, stride_a, stride_b, stride_d, batch_stride_a, batch_stride_b, batch_stride_d, k, ne02, ne12, broadcast2, broadcast3, padded_n };
-        ggml_vk_dispatch_pipeline(ctx, subctx, pipeline, { a, b, d }, pc, { m, n, batch });
+        uint32_t base_work_group_z = 0;
+        while (base_work_group_z < batch) {
+            uint32_t groups_z = std::min(batch - base_work_group_z, ctx->device->properties.limits.maxComputeWorkGroupCount[2]);
+
+            ggml_pipeline_request_descriptor_sets(ctx, pipeline, 1);
```
0cc4m: Why request the descriptor sets in the loop and not before? It's not gonna retrigger pipeline compile of course, but will ping the descriptor pools more than necessary.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
jeffbolznv: fixed, thanks

ggml/src/ggml-vulkan/ggml-vulkan.cpp
jeffbolznv: ok, moved it out of the loop for now. Eventually, I'd like to not have to explicitly call this anywhere.

0cc4m: It should be possible to automatically request one when grabbing the pipeline and to allocate the sets on demand before dispatch.

jeffbolznv: Yeah, the main catch right now is that some shaders won't get their wg_denoms initialized until after this call.

0cc4m (approved):

PR #19472, opened 2026-02-10
common : fix filename validation for subfolders

kaetemi:
In commit https://github.com/ggml-org/llama.cpp/commit/13628d8bdb48d2852712a6998d4f5a9e880059e8 an option `allow_subdirs` was added to `fs_validate_filename`. The original function was designed for filenames-only (blocking subdirs), and some of the checks (trailing and leading whitespaces on Windows, etc.) are not validated correctly when this flag is enabled.

This PR fixes the validation when subfolders are allowed and adds a test case. (The test case is AI generated.)

Additionally, new path traversal equivalents for Windows are added to the list of blocked Unicode characters, see https://blog.orange.tw/posts/2025-01-worstfit-unveiling-hidden-transformers-in-windows-ansi/ and https://worst.fit/mapping/#ANSI%3A0x5C%2C0x2F%2C0x2E%2C0x3A%20

PR #19517, opened 2026-02-11
common : replace deprecated codecvt using parse_utf8_codepoint

angt:


kaetemi:
Can you check this against the extended testcases for filename validation in https://github.com/ggml-org/llama.cpp/pull/19472 ?

angt:
> Can you check this against the extended testcases for filename validation in #19472 ?

Yes, there’s no issue. It will only cause a conflict when merging.

ggerganov (approved):
Wait for the CI

PR #19529, opened 2026-02-11
Add Jinja support for "indent" string filter

pwilkin:
Only pure `| indent` supported for now, without the `first` flag since that would require massive changes to the Jinja engine.

### DIFF
--- a/common/jinja/value.cpp
+++ b/common/jinja/value.cpp
@@ -741,7 +741,7 @@ const func_builtins & value_string_t::get_builtins() const {
                 if (!indented.empty()) {
                     indented.push_back('\n');
                 }
-                if ((indented.empty() && first) || !line.empty() || blank) {
+                if ((indented.empty() ? first : (!line.empty() || blank))) {
                     indented += indent;
                 }
                 indented += line;
--- a/common/jinja/value.cpp
+++ b/common/jinja/value.cpp
@@ -737,25 +737,19 @@ const func_builtins & value_string_t::get_builtins() const {
             std::string input = val_input->as_string().str();
             std::istringstream iss = std::istringstream(input);
             std::string line;
-            bool first_line = true;
             while (std::getline(iss, line)) {
-                bool do_indent = true;
-                if (first_line) {
-                    first_line = false;
-                    do_indent &= first;
-                } else {
-                    indented.append("\n");
+                if (!indented.empty()) {
+                    indented.push_back('\n');
                 }
-                do_indent &= blank || !line.empty();
-                if (do_indent) {
-                   indented.append(indent);
+                if ((indented.empty() && first) || !line.empty() || blank) {
+                    indented += indent;
                 }
-                indented.append(line);
+                indented += line;
             }
             if (!input.empty() && input.back() == '\n') {
-                indented.append("\n");
+                indented.push_back('\n');
                 if (blank) {
-                    indented.append(indent);
+                    indented += indent;
                 }
             }
 
--- a/common/jinja/value.cpp
+++ b/common/jinja/value.cpp
@@ -4,6 +4,7 @@
 // for converting from JSON to jinja values
 #include <nlohmann/json.hpp>
 
+#include <sstream>
 #include <string>
 #include <cctype>
 #include <vector>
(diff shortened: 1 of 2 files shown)


CISC:
> Only pure `| indent` supported for now, without the `first` flag since that would require massive changes to the Jinja engine.

Hmmm, I think you are confused, `first` simply refers to the first line of the input.

common/jinja/value.cpp
```diff
@@ -715,8 +715,26 @@ const func_builtins & value_string_t::get_builtins() const {
...
+
+            int count = arg0->as_int();
+            if (count <= 0) {
+                throw raised_exception("indent must be a positive number");
+            }
+            std::string indented;
+            for (int i = 0; i < count; i++) {
+                indented.append(" ");
+            }
+            indented.append(input->as_string().str());
```
CISC:
```suggestion
            args.ensure_count(1, 4);

            value val_input  = args.get_pos(0);
            value val_width  = args.get_kwarg_or_pos("width", 1);
            const bool first = args.get_kwarg_or_pos("first", 2)->as_bool(); // undefined == false
            const bool blank = args.get_kwarg_or_pos("blank", 3)->as_bool(); // undefined == false

            if (!is_val<value_string>(val_input)) {
                throw raised_exception("indent() first argument must be a string");
            }

            std::string indent;
            if (is_val<value_int>(val_width)) {
                indent.assign(val_width->as_int(), ' ');
            } else if (is_val<value_string>(val_width)) {
                indent = val_width->as_string().str();
            } else {
                indent = "    ";
            }
            std::string indented;
```
..now you just need to iterate through the input newlines and indent with `indent` string, first line if `first` is true and blank lines if `blank` is true.

pwilkin:
> Hmmm, I think you are confused, `first` simply refers to the first line of the input.

Oh, LOL, I actually thought for some reason that `first` referred to the first line of the *output*. Never mind then :)

tests/test-jinja.cpp
```diff
@@ -691,6 +691,12 @@ static void test_filters(testing & t) {
         "{\n  \"a\": 1,\n  \"b\": [\n    1,\n    2\n  ]\n}"
     );
 
+    test_template(t, "indent",
+        "{{ data|indent(4)}}",
+        { "data", "foo" },
+        "    foo"
+    );
```
CISC, line 734:
```suggestion

    test_template(t, "indent",
        "{{ data|indent }}",
        { "data", "foo\n\nbar" },
        "foo\n\n    bar"
    );

    test_template(t, "indent first",
        "{{ data|indent(2, first=true) }}",
        { "data", "\n\nbar" },
        "  \n\n  bar"
    );

    test_template(t, "indent blank",
        "{{ data|indent('  ', blank=true) }}",
        { "data", "foo\n\nbar" },
        "foo\n  \n  bar"
    );
```

pwilkin:
@CISC okay, removed "partial" :)

tests/test-jinja.cpp
```diff
@@ -691,6 +691,36 @@ static void test_filters(testing & t) {
...
+        "{{ data|indent(width=5,blank=true,first=true) }}",
+        {{ "data", "foo\n\nbar" }},
+        "     foo\n     \n     bar"
+    );
+
+    test_template(t, "indent with default width",
+        "{{ data|indent() }}",
+        {{ "data", "foo\nbar" }},
+        "foo\n    bar"
+    );
```
CISC, line 716:
```suggestion

    test_template(t, "indent with default width",
        "{{ data|indent() }}",
        {{ "data", "foo\nbar" }},
        "foo\n    bar"
    );

    test_template(t, "indent with no newline",
        "{{ data|indent }}",
        {{ "data", "foo" }},
        "foo"
    );

    test_template(t, "indent with trailing newline",
        "{{ data|indent(blank=true) }}",
        {{ "data", "foo\n" }},
        "foo\n    "
    );
```

common/jinja/value.cpp
```diff
@@ -715,8 +716,44 @@ const func_builtins & value_string_t::get_builtins() const {
...
+                    do_indent &= first;
+                } else {
+                    indented.append("\n");
+                }
+                do_indent &= blank || !line.empty();
+                if (do_indent) {
+                   indented.append(indent);
+                }
+                indented.append(line);
+            }
```
CISC, line 748: Added two more tests as I think this has issues due to `getline` eating newlines without telling you.

Also, you can simply prefill `indented` outside the loop if `first` is true, should simplify the block.

CISC:
@pwilkin gentle ping

common/jinja/value.cpp
```diff
@@ -715,8 +716,52 @@ const func_builtins & value_string_t::get_builtins() const {
...
+            if (!input.empty() && input.back() == '\n') {
+                indented.append("\n");
+                if (blank) {
+                    indented.append(indent);
+                }
+            }
```
CISC:
```suggestion
            while (std::getline(iss, line)) {
                if (!indented.empty()) {
                    indented.push_back('\n');
                }
                if ((indented.empty() && first) || !line.empty() || blank) {
                    indented += indent;
                }
                indented += line;
            }
            if (!input.empty() && input.back() == '\n') {
                indented.push_back('\n');
                if (blank) {
                    indented += indent;
                }
            }
```
This could probably be improved further (we could f.ex. work with the `jinja::string` directly), but will do for now.

CISC (approved):

common/jinja/value.cpp
CISC: Bah, messed that up slightly. :)

pwilkin: Yeah, I just decided on "don't fix it if it ain't broken" for this one :)

PR #19530, opened 2026-02-12
ggml webgpu: shader library organization

reeselevine:
We've been converting many of the existing WGSL shaders into a format that allows for efficient just-in-time compilation of variants used in specific model graphs, as well as sets them up for better performance tuning down the road. This PR makes a pretty large organizational change, moving the shader preprocessing, compilation, and caching into a new `ggml_webgpu_shader_lib` structure. As part of this, the existing matrix multiplication shaders were also converted in to the JIT compilation format (using the wgsl preprocessor), along with get_rows and scale.

This new shader library class also opens up the opportunity for tons of interesting specialization in the WebGPU backend. For example, if you have a shader specialized for a particular GPU vendor/architecture in WGSL, it should be pretty easy to hook it into the logic for choosing the right shader/pipeline.

It's always nice to have a PR that removes more lines of code than it adds too :)


SharmaRithik:
The changes look good to me! The new shader library structure and JIT variant setup make the design cleaner and more extensible, and the refactor looks solid overall.

ggerganov (approved):

PR #19375, opened 2026-02-05
models : optimizing qwen3next graph

ggerganov:
Rewording the ggml compute graph to avoid too many unnecessary copies.

M2 Ultra:

| Model                    | Test   |   t/s b7946 |   t/s gg/qwen3-next-opt |   Speedup |
|:-------------------------|:-------|------------:|------------------------:|----------:|
| qwen3next 80B.A3B Q4_0   | pp1    |       37.92 |                   51.99 |      1.37 |
| qwen3next 80B.A3B Q4_0   | pp2    |       43.20 |                   57.55 |      1.33 |
| qwen3next 80B.A3B Q4_0   | pp3    |       60.03 |                   78.88 |      1.31 |
| qwen3next 80B.A3B Q4_0   | pp4    |       77.27 |                  100.71 |      1.30 |
| qwen3next 80B.A3B Q4_0   | pp5    |       89.65 |                  119.23 |      1.33 |
| qwen3next 80B.A3B Q4_0   | pp6    |      109.17 |                  140.88 |      1.29 |
| qwen3next 80B.A3B Q4_0   | pp7    |      122.24 |                  156.66 |      1.28 |
| qwen3next 80B.A3B Q4_0   | pp8    |      137.75 |                  176.36 |      1.28 |
| qwen3next 80B.A3B Q4_0   | pp512  |      930.70 |                 1125.73 |      1.21 |
| qwen3next 80B.A3B Q4_0   | pp2048 |     1049.91 |                 1352.31 |      1.29 |
| qwen3next 80B.A3B Q4_0   | tg32   |       38.02 |                   50.39 |      1.33 |
| qwen3next 80B.A3B Q4_K_M | pp1    |       34.00 |                   46.47 |      1.37 |
| qwen3next 80B.A3B Q4_K_M | pp2    |       40.94 |                   52.84 |      1.29 |
| qwen3next 80B.A3B Q4_K_M | pp3    |       56.43 |                   72.38 |      1.28 |
| qwen3next 80B.A3B Q4_K_M | pp4    |       68.36 |                   87.54 |      1.28 |
| qwen3next 80B.A3B Q4_K_M | pp5    |       81.44 |                  104.15 |      1.28 |
| qwen3next 80B.A3B Q4_K_M | pp6    |       96.56 |                  121.09 |      1.25 |
| qwen3next 80B.A3B Q4_K_M | pp7    |      105.85 |                  131.26 |      1.24 |
[...]

PR #19531, opened 2026-02-12
Kimi Linear fix conv state update

ymcki:

The current implementation has incorrect conv state update such that it has state corruption when running parallel in llama-server. This is fixed in this PR.

```
./build/bin/llama-server -c 16384 --parallel 8 --mmap -m ~/Kimi-Linear-48B-A3B-Instruct-GGUF/Kimi-Linear-48B-A3B-Instruct-jp-imatrix.IQ3_M.gguf -ngl 100
```

This PR also includes the block implementation that speeds up 20% pp and VRAM saving.


CISC:
> This PR also includes the block implementation that speeds up 20% pp and VRAM saving.

Please take this out again and make this a purely bugfix PR.

The block implementation can be done in a separate PR, however it's worth noting that there are several incoming improvements to this in #19375 that can be applied here also.

ymcki:
Now only fix conv state update.

CISC (approved):

PR #19539, opened 2026-02-12
make `ggml_is_view` as API

foldl:

This PR makes `ggml_is_view` into an API.

## Motivation

Some backends complain ([example](https://github.com/ggml-org/llama.cpp/blob/b1ff83bbb08bd946f79b89bc7adbd7f08ae61303/ggml/src/ggml-vulkan/ggml-vulkan.cpp#L1907-L1911)) when an input is a view with offsets. It helps to have an `is_view` API to make these backends happy.

### DIFF
--- a/ggml/src/ggml-impl.h
+++ b/ggml/src/ggml-impl.h
@@ -98,7 +98,7 @@ static bool ggml_op_is_empty(enum ggml_op op) {
     }
 }
 
-static inline bool ggml_aux_is_view(const struct ggml_tensor * t) {
+static inline bool ggml_impl_is_view(const struct ggml_tensor * t) {
     return t->view_src != NULL;
 }
 
--- a/ggml/src/ggml-impl.h
+++ b/ggml/src/ggml-impl.h
@@ -98,6 +98,10 @@ static bool ggml_op_is_empty(enum ggml_op op) {
     }
 }
 
+static inline bool ggml_aux_is_view(const struct ggml_tensor * t) {
+    return t->view_src != NULL;
+}
+
 static inline float ggml_compute_softplus_f32(float input) {
     return (input > 20.0f) ? input : logf(1 + expf(input));
 }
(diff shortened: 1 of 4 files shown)


ggerganov:
So there is a concern that some of these helper functions should remain in the implementation so that they can be inlined efficiently in the internal ggml code - some of them can be called extensively in hot loops. `ggml_is_view()` specifically does not seem to fall in this category, but I think it would be a good time to introduce a pattern to fix this more generally.

I think we can do the following:

- Add `ggml_aux_is_view()` to public API (`ggml.h`) with implementation in `ggml.c`
- Move `ggml_is_view()` implementation to `ggml-impl.h`
- Make `ggml_aux_is_view()` implementation simply call `ggml_is_view()` (i.e. it will be inlined)

This way, internal ggml code will continue to use the inlined `ggml_is_view()` which is now located in `ggml-impl.h`. User code can call `ggml_aux_is_view()` which has the overhead of a function call.

With this new pattern established, we can continue to migrate the rest of the auxilary calls, such as `ggml_is_transposed()`, `ggml_is_contiguous()` etc. WDYT?

Edit: `_aux_` -> `_impl_` and `ggml_impl_*` -> `ggml-impl.h`. See comments below.

foldl:
Yes, this is the best solution, inline for performance and an auxiliary for API.

The only shortcoming is that API names become longer with `_aux_...`. So I suggest to interchange the use of `ggml_is_view()` and `ggml_aux_is_view()`. This also paves the way for `ggml_is_transposed()`, etc: API can be kept unchanged and `ggml_aux_is_transposed()` can be used internally for performance.

ggerganov:
> The only shortcoming is that API names become longer with _aux_.... So I suggest to interchange the use of ggml_is_view() and ggml_aux_is_view(). This also paves the way for ggml_is_transposed(), etc: API can be kept unchanged and ggml_aux_is_transposed() can be used internally for performance.

Yes, it won't be a breaking change this way. Though it would require a massive change in the internal ggml implementation to switch to the aux calls. But at the same time, it can be done incrementally. So your suggestion might be the better option indeed - let me give this a bit of more thought.

ggml/src/ggml-impl.h
```diff
@@ -98,6 +98,10 @@ static bool ggml_op_is_empty(enum ggml_op op) {
     }
 }
 
+static inline bool ggml_aux_is_view(const struct ggml_tensor * t) {
```
ggerganov: Let's use `ggml_impl_` prefix:

```suggestion
static inline bool ggml_impl_is_view(const struct ggml_tensor * t) {
```

ggerganov (approved):

PR #19260, opened 2026-02-02
fix: use physical cores for --threads auto-detect (#19110)

ingyukoh:
Replace std::thread::hardware_concurrency() with cpu_get_num_math() when --threads is set to -1 or 0 (auto-detect mode).

hardware_concurrency() returns logical cores (includes hyperthreads), causing thread oversubscription and performance degradation:
- 100% CPU usage instead of optimal ~50%
- 3.6x slower (2.5 tok/s vs 9 tok/s reported)

cpu_get_num_math() returns physical cores and also handles Intel hybrid CPUs by skipping efficiency cores for math workloads.

Fixes #19110

*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

PR #19542, opened 2026-02-12
arg : use cpu_get_num_math() for --threads fallback instead of hardware_concurrency()

yurekami:
## Summary

- Replace `std::thread::hardware_concurrency()` with `cpu_get_num_math()` in 4 thread-count fallback locations in `common/arg.cpp`
- Remove unused `#include <thread>` (was only needed for `hardware_concurrency`)

## Problem

When `--threads -1` (or `--threads-batch`, `--threads-draft`, `--threads-batch-draft`) is passed, the argument parser falls back to `std::thread::hardware_concurrency()` which returns the **logical** core count (including hyper-threads). On hyper-threaded CPUs, this spawns too many threads, causing context switching overhead and significant performance degradation (reported: 9 tok/s → 2.5 tok/s at 100% CPU utilization).

## Fix

Use `cpu_get_num_math()` instead, which returns the number of **physical/performance** cores. This function already exists in `common/common.cpp` and correctly handles:
- Linux: reads `/sys/devices/system/cpu/*/topology/thread_siblings` + hybrid CPU detection
- macOS: uses `sysctl hw.perflevel0.physicalcpu` / `hw.physicalcpu`
- Windows: uses `GetLogicalProcessorInformationEx(RelationProcessorCore)`

This is consistent with:
- `postprocess_cpu_params()` which already uses `cpu_get_num_math()` for the default `n_threads = -1` case
- `llama-bench` which already uses `cpu_get_num_math()` for its default thread count

## Test plan

- [ ] Verify `--threads -1` reports physical core count (not logical) in server log output
- [ ] Verify no regression on non-HT systems (physical == logical)
- [ ] Verify `--threads N` (explicit positive value) still works unchanged

Closes #19110

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -28,7 +28,6 @@
 #include <regex>
 #include <set>
 #include <string>
-#include <thread> // for hardware_concurrency
 #include <vector>
 
 #ifndef __EMSCRIPTEN__
@@ -977,7 +976,7 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         [](common_params & params, int value) {
             params.cpuparams.n_threads = value;
             if (params.cpuparams.n_threads <= 0) {
-                params.cpuparams.n_threads = std::thread::hardware_concurrency();
+                params.cpuparams.n_threads = cpu_get_num_math();
             }
         }
     ).set_env("LLAMA_ARG_THREADS"));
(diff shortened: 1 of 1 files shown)


MaxKruse:
https://github.com/ggml-org/llama.cpp/pull/19260

yurekami:
Closing as duplicate of #19260 which addresses the same issue with the same approach. Thanks @MaxKruse for pointing it out!

PR #19558, opened 2026-02-12
[CMake] Enable test-chat out of tree build

jplehr:
The test-chat binary relies on model files that it tries to find. However, when configuring the build directory to be parallel to the source tree those heuristics fail.

This sets the working directory for the test executable to be the source-tree which resolves this issue.
I validated locally with a build parallel to the source tree and nested inside the source tree.

I used an AI tool to help me find the problem and generate the change.
To the best of my understanding this is a reasonable fix.

Edit: clarified which binary.


jplehr:
Ping.

(I did not find any particular "ping policy" and would like to apologize if this ping is considered rude here.)

danbev (approved):

PR #19521, opened 2026-02-11
CUDA: Enable cuda graphs for qwen3 next-style architectures

ORippler:
This is achieved via:
1. Adding regex for pico-batched adds that happen even when ubatch is 1
2. Bumping the consecutive updates threshold by 1.

While we already know 1. as tech-debt in cuda backend from other models, 2. was new to me. I did try to investigate what happens with the frequent graph updates and hybrid model, but could not figure it out fully:

1. We seem to build different shapes of `conv_states_updated-0` two times at ˜128 token intervals from 0 -> 24576 -> 0. This seems to come from `llm_graph_input_rs`?. My understanding is that this `conv_states_updated-0` somewhat reflect a per-sequence mapping of our conv states, whereas we will only have a single sequence in llama-bench's token-generation phase (so it should always be 0 in that case? At least in the reused ggml_cgraph-case it ends-up being 0). This change in ggml_cgraph topology is however seemingly captured by `llm_graph_result::can_reuse`.
2. `node_143` (this is somewhere in the first `build_layer_attn_linear` of qwen3next) seems to have additional data added to it despite the graph being denoted as being reusable by `llm_graph_result::can_reuse`. Is this potentially a bug? I was under the impression that graph topology is not allowed to change when its reused in `llama_context::process_ubatch`.

<details>

<summary>Relevant sections from log output and diff patch for above two points</summary>

```
./build-x64-linux-gcc-reldbg/bin/llama-bench -m /mnt/share/gguf/Qwen/Qwen3-Coder-Next-GGUF/Qwen3-Coder-Next-Q4_K_M/Qwen3-Coder-Next-Q4_K_M-00001-of-00004.gguf -mmp 0 -dio 1 -fa 1 -p 0 2>&1 | tee run.txt

Middle of llama-bench
can_reuse: can reuse graph = 1
can_reuse: can reuse graph = 1
can_reuse: can reuse graph = 1
can_reuse: can reuse graph = 1
can_reuse: can reuse graph = 1
can_reuse: can reuse graph = 1
can_reuse: can reuse graph = 0
node->ne[0] mismatch: 24576 vs 0 for node->name: conv_states_updated-0 (reshaped) (view)
node 6 properties do not match, node name: conv_states_updated-0 (reshaped) (view)
can_reuse: can reuse graph = 0
node->ne[0] mismatch: 0 vs 24576 for node->name: conv_states_updated-0 (reshaped) (view)
node 6 properties do not match, node name: conv_states_updated-0 (reshaped) (view)
[...]

PR #19566, opened 2026-02-12
CUDA: Do not mutate cgraph for fused ADDs

ORippler:
1. We should try to minimize in-place changes to the incoming ggml_cgraph where possible (those should happen in a backends' `graph_optimize` function)
2. Modifying in-place leads to an additional, unnecessary graph capture step as we store the properties before modifying the graph in-place in the cuda-backend: We hit `ggml_cuda_graph_node_set_properties` via `ggml_cuda_graph_update_required` before entering `ggml_cuda_graph_evaluate_and_capture`. 

Isolated from #19521

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -3641,6 +3641,10 @@ static void ggml_cuda_graph_evaluate_and_capture(ggml_backend_cuda_context * cud
 
                         if (n_fuse > 1) {
                             ggml_tensor fused_add_node;
+                            // Let's ensure compilation fails should ggml_tensor at one point no-longer be trivially copyable.
+                            static_assert(
+                                std::is_trivially_copyable_v<ggml_tensor>,
+                                "ggml_tensor must be trivially copyable for memcpy usage in fused-add node-creation");
                             memcpy(&fused_add_node, node, sizeof(ggml_tensor));
                             for (int j = 0; j < n_fuse - 1; ++j) {
                                 fused_add_node.src[j + 2] = cgraph->nodes[i + j + 1]->src[1];
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -3640,11 +3640,13 @@ static void ggml_cuda_graph_evaluate_and_capture(ggml_backend_cuda_context * cud
                         n_fuse++;
 
                         if (n_fuse > 1) {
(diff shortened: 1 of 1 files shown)


ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3640,11 +3640,13 @@ static void ggml_cuda_graph_evaluate_and_capture(ggml_backend_cuda_context * cud
                         n_fuse++;
 
                         if (n_fuse > 1) {
+                            ggml_tensor fused_add_node;
+                            memcpy(&fused_add_node, node, sizeof(ggml_tensor));
```
am17an, line 3644: I'm not sure if this memcpy is safe, as ggml_tensor is not trivially copyable

ORippler: > I'm not sure if this memcpy is safe, as ggml_tensor is not trivially copyable

`std::is_trivially_copyable_v<ggml_tensor>` resolves to true on my side. Would you mind explaining more what you mean by "is not trivially copyable"? We duplicate all arrays that contain pointers into `fused_add_node`, and update those duplicated array-of-pointers only.

I added a static_assert in 17717f3fc6653196e018823b858edc673c0c913e for future-proofing the memcpy.

am17an: Sorry, my bad. I got confused with `ggml_cuda_graph_node_properties` which is already statically asserted. `ggml_tensor` being a C struct would always be trivially copyable. 
https://github.com/ggml-org/llama.cpp/blob/15d83e0c87530b1166acec0550f6320aa0fa2461/ggml/src/ggml-cuda/common.cuh#L1135-L1136

am17an (approved):

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3640,11 +3640,17 @@ static void ggml_cuda_graph_evaluate_and_capture(ggml_backend_cuda_context * cud
                         n_fuse++;
 
                         if (n_fuse > 1) {
+                            ggml_tensor fused_add_node;
+                            // Let's ensure compilation fails should ggml_tensor at one point no-longer be trivially copyable.
+                            static_assert(
+                                std::is_trivially_copyable_v<ggml_tensor>,
+                                "ggml_tensor must be trivially copyable for memcpy usage in fused-add node-creation");
+                            memcpy(&fused_add_node, node, sizeof(ggml_tensor));
```
```suggestion
                            memcpy(&fused_add_node, node, sizeof(ggml_tensor));
```

PR #19567, opened 2026-02-12
metal : fix incorrect output on AMD discrete GPUs

### LINKED ISSUE #19563: bug: Metal backend produces garbage output on AMD discrete GPUs (Radeon Pro 5300M)
## Bug Description

The Metal backend produces incorrect (garbage) output on AMD discrete GPUs while benchmarks show plausible performance numbers. The same model works correctly with the Vulkan backend (via MoltenVK) on the same GPU, and with the Metal backend when using CPU-only (`-ngl 0`).

## Environment

- **GPU**: AMD Radeon Pro 5300M (4 GB VRAM, discrete)
- **Secondary GPU**: Intel UHD Graphics 630 (integrated)
- **macOS**: 26.2 (Build 25C56)
- **llama.cpp**: commit 4d688f9 (current master)
- **Metal family**: MTLGPUFamilyMetal3 (5001), MTLGPUFamilyCommon3 (3003)
- **Metal props**: `simdgroup reduction = true`, `simdgroup matrix mul. = false`, `has unified memory = false`, `use shared buffers = false`

## Reproduction

```bash
# Build with Metal
cmake -B build -DGGML_METAL=ON -DGGML_METAL_EMBED_LIBRARY=ON
cmake --build build -j -t llama-completion

# Metal GPU — BROKEN (garbage output)
./build/bin/llama-completion \
  -m Qwen2.5-1.5B-Instruct-Q4_K_M.gguf \
  -ngl 99 -n 32 -s 42 -no-cnv \
  -p "Explain gravity in one sentence:"
# Output: "Explain gravity in one sentence:@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@"

# CPU only — WORKS
./build/bin/llama-completion \
  -m Qwen2.5-1.5B-Instruct-Q4_K_M.gguf \
  -ngl 0 -n 32 -s 42 -no-cnv \
  -p "Explain gravity in one sentence:"
# Output: "Gravity is the force that attracts two masses towards each other..."
```

## Observations

- Affects both Q4_K_M and F16 model formats — not quant-type specific
- `llama-bench` reports plausible performance numbers (pp128: ~100 t/s, tg32: ~63 t/s) but the computation is incorrect
- Vulkan backend (via MoltenVK → Metal → same GPU) produces correct output at ~68 t/s tg32
- The GPU is discrete with no unified memory (`has_unified_memory = false`, `use_shared_buffers = false`)
- `simdgroup matrix mul. = false` (no Apple7 GPU family), so flash attention falls back to CPU and matrix multiply uses the legacy `kernel_mul_mv` path

## Possible Causes

1. **Buffer management**: Discrete AMD GPUs use private buffers with explicit CPU↔GPU copies (unlike Apple Silicon's shared/unified memory). Data may not be correctly transferred.
2. **KV cache handling**: `kv_unified = false` in the logs suggests a different KV cache path for non-unified memory GPUs.
3. **Kernel correctness**: The legacy `mul_mv` kernels may have assumptions that hold on Apple Silicon but break on AMD (e.g., memory ordering, simdgroup behavior).

## Additional Context

This was discovered while benchmarking Metal vs Vulkan performance on AMD. The benchmarks appeared reasonable, masking the fact that Metal was producing incorrect results. This affects any AMD discrete GPU on macOS using the Metal backend.

ai-janitor:
## Summary

Fixes Metal backend producing garbage output on AMD discrete GPUs (e.g. Radeon Pro 5300M). Two bugs fixed:

1. **Concurrent compute encoders break on non-UMA GPUs** — root cause of garbage output
2. **`NSMakeRange` buffer overwrite in `memset_tensor`** — secondary bug on private buffer path

Closes #19563

## Root Cause

The Metal backend enables concurrent compute command encoders by default (`MTLDispatchTypeConcurrent`), relying on `memoryBarrierWithScope:MTLBarrierScopeBuffers` to synchronize between dependent dispatches. This works on Apple Silicon but **produces incorrect results on AMD discrete GPUs** — the barriers don't properly synchronize memory on these devices.

### How we found it

Bisected by GPU layer count:
- `ngl=0` (CPU): correct
- `ngl=1`: correct  
- `ngl=2`: **garbage starts** — more layers = more concurrent dispatches = more stale reads
- `ngl=99, n=1`: first token plausible (PP completes before read-back)
- `ngl=99, n=2+`: all garbage (`@@@@@`)

This pattern pointed to an intra-command-buffer synchronization issue rather than buffer allocation, KV cache, or kernel correctness. Tested with `GGML_METAL_CONCURRENCY_DISABLE=1` — output became correct immediately.

### The fix

Gate `use_concurrency` on `has_unified_memory` (same pattern used for `use_shared_buffers`). This cleanly separates Apple Silicon (UMA, concurrency works) from AMD discrete (non-UMA, concurrency broken). The `GGML_METAL_CONCURRENCY_DISABLE` env var still works as an override.

```objc
// before
res->use_concurrency = getenv("GGML_METAL_CONCURRENCY_DISABLE") == nil;

// after
res->use_concurrency = props_dev->has_unified_memory && getenv("GGML_METAL_CONCURRENCY_DISABLE") == nil;
```

### Secondary fix: `memset_tensor` NSMakeRange

`NSMakeRange(location, length)` was called as `NSMakeRange(bid_dst.offs, bid_dst.offs + size)` — second arg should be `size`, not `offset + size`. Overwrites past intended range when tensor is at non-zero offset in a private buffer. Only affects the private buffer path (non-UMA GPUs).

## Test Results

Tested on AMD Radeon Pro 5300M, macOS 26.2, Qwen2.5-1.5B-Instruct Q4_K_M:

| Test | Before | After |
|------|--------|-------|
| ngl=99, "capital of France" | `@@@@@` garbage | "Paris" ✓ |
| ngl=99, haiku (50 tokens) | garbage | coherent output ✓ |
| ngl=2 (failure boundary) | garbage | "Paris" ✓ |
| Bench pp512/tg128 | 99/63 t/s | 99/59 t/s (no meaningful regression) |

No impact on Apple Silicon — `has_unified_memory = true` so behavior is unchanged.

### DIFF
--- a/ggml/src/ggml-metal/ggml-metal-context.m
+++ b/ggml/src/ggml-metal/ggml-metal-context.m
@@ -130,7 +130,8 @@ ggml_metal_t ggml_metal_init(ggml_metal_device_t dev) {
     res->d_queue = dispatch_queue_create("ggml-metal", DISPATCH_QUEUE_CONCURRENT);
 
     res->use_fusion      = getenv("GGML_METAL_FUSION_DISABLE") == nil;
-    res->use_concurrency = getenv("GGML_METAL_CONCURRENCY_DISABLE") == nil;
+    // non-UMA GPUs (AMD discrete) produce incorrect results with concurrent encoders
+    res->use_concurrency = props_dev->has_unified_memory && getenv("GGML_METAL_CONCURRENCY_DISABLE") == nil;
 
     {
         const char * val = getenv("GGML_METAL_GRAPH_DEBUG");
--- a/ggml/src/ggml-metal/ggml-metal-device.m
+++ b/ggml/src/ggml-metal/ggml-metal-device.m
@@ -1595,7 +1595,7 @@ void ggml_metal_buffer_memset_tensor(ggml_metal_buffer_t buf, struct ggml_tensor
             id<MTLBlitCommandEncoder> encoder = [cmd_buf blitCommandEncoder];
 
             [encoder fillBuffer:bid_dst.metal
-                          range:NSMakeRange(bid_dst.offs, bid_dst.offs + size)
+                          range:NSMakeRange(bid_dst.offs, size)
                           value:value];
 
             [encoder endEncoding];


am17an:
@ai-janitor please read the contributing guidelines https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md and please stop submitting PRs which are not going to be accepted.

PR #19582, opened 2026-02-13
vulkan: restore -inf check in FA shaders

jeffbolznv:
For #19523.

I verified the performance is restored with llama-batched-bench.


0cc4m (approved):
LGTM, you can merge or wait for confirmation that this fixes it in the issue.

PR #18655, opened 2026-01-07
webui: Agentic Loop + MCP Client with support for Tools, Resources and Prompts

allozaur:
## New features

- Adding System Message to conversation or injecting it to an existing one
- CORS Proxy on llama-server backend side (run `llama-server` command witg `--webui-mcp-proxy` flag to enable it)
- MCP
    - Servers Selector
    - Settings with Server cards showing capabilities, instructions and other information
    - Tool Calls
        - Agentic Loop
            - Logic
            - UI with processing stats
    - Prompts
        - Detection logic in „Add” dropdown
        - Prompt Picker
        - Prompt Args Form
        - Prompt Attachments in Chat Form and Chat Messages
    - Resources
        - Browser with search & filetree view
        - Resource Attachments & Preview dialog
        - Use Resource Templates to attach resources
- Show raw output switch under the assistant message
- Favicon utility
- Key-Value form component (used for MCP Server headers in add new/edit mode)

## UI Improvements

- Created TruncatedText component
- Created Image CORS error fallback UI
- Created DropdownMenuSearchable component
- Created HorizontalScrollCarousel
- Created CollapsibleContentBlock component + refactored Reasoning Content && Tool Calls
- Code block improved UI + unfinished code block handling + syntax highlighting improvements
- Max-heght for components + autoscroll of overflowing content during streaming
- New statistics UI + added new data for Tool Calls and Agentic summary
- Better time formatting for Chat Message Statistics

## Architecture refactors/improvements:

- Autoscroll hook
- Renamed all service files to have `.service.ts` format
- Common types definitions
- Abort Signal utility + refactor
- API Fetch utility
- Cache TTL
- Components folder naming & restructuring
- Context API for editing messages and message actions
- Enums & constants cleanup
- Markdown Rendering improvements
- Chat Form components
- Resolving images in Markdown Content
- Chat Attachments API improvements
- Removed Model Change validation logic
- Removed File Upload validation logic
- New formatters
- Storybook upgrade
- New prop to define initial section when opening Chat Settings

> [!IMPORTANT]
> This PR includes MCP-only changes, but it builds on couple of PRs improving architecture and UI foundations in the codebase: #19689, #19685, #19596, #19586, #19571, #19556, #19551, #19541 and #20066

### Adding a new MCP Server and using it within an Agentic Loop

https://github.com/user-attachments/assets/56a3016b-a444-413b-ab53-7fe175d73583

### Using MCP Prompts

[...]

PR #19586, opened 2026-02-13
webui: UI and routing fixes

allozaur:
This is one of the PRs to unload code from #18655 that is not directly MCP-related


ggerganov (approved):
Btw, not sure when this regression occurred, but the UI does not allow to start a new chat when one is already running:

https://github.com/user-attachments/assets/1f1f8dcf-4328-4113-94d5-b853f601d065

allozaur:
> Btw, not sure when this regression occurred, but the UI does not allow to start a new chat when one is already running:
> Screen.Recording.2026-02-13.at.13.13.46.mov

Yep, I am aware and it's fixed in the upcoming PR :)

PR #19588, opened 2026-02-13
scripts : use official split.py for cpp-httplib

angt:
Using the official script is safer and ensures the generated code aligns with the library's standards.

### DIFF
--- a/scripts/sync_vendor.py
+++ b/scripts/sync_vendor.py
@@ -1,6 +1,9 @@
 #!/usr/bin/env python3
 
 import urllib.request
+import os
+import sys
+import subprocess
 
 HTTPLIB_VERSION = "f80864ca031932351abef49b74097c67f14719c6"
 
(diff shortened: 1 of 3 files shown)


scripts/sync_vendor.py
```diff
@@ -24,19 +28,15 @@
...
-            fc.write(implementation.replace('\ninline ', '\n'))
-            fc.write('} // namespace httplib\n')
+print("Splitting httplib.h...")
+try:
+    subprocess.check_call([
+        sys.executable, "split.py",
+        "--extension", "cpp",
+        "--out", "vendor/cpp-httplib"
+    ])
+    os.remove("split.py")
```
danbev, line 42: Perhaps  this could be moved to a finally block to always be cleaned up:
```python:
finally:
    os.remove("split.py")
    os.remove("httplib.h")
```

danbev (approved):

scripts/sync_vendor.py
```diff
@@ -14,7 +17,8 @@
     # "https://github.com/mackron/miniaudio/raw/refs/tags/0.11.23/miniaudio.h": "vendor/miniaudio/miniaudio.h",
     "https://github.com/mackron/miniaudio/raw/669ed3e844524fcd883231b13095baee9f6de304/miniaudio.h": "vendor/miniaudio/miniaudio.h",
 
-    f"https://raw.githubusercontent.com/yhirose/cpp-httplib/{HTTPLIB_VERSION}/httplib.h": "vendor/cpp-httplib/httplib.h",
+    f"https://raw.githubusercontent.com/yhirose/cpp-httplib/{HTTPLIB_VERSION}/httplib.h": "httplib.h",
+    f"https://raw.githubusercontent.com/yhirose/cpp-httplib/{HTTPLIB_VERSION}/split.py":  "split.py",
```
ngxson, line 21: maybe safer to download it temporary to one of the directory in `.gitignore`, example: `/tmp/httplib.h`, `/tmp/split.py`

angt: I prefer to keep it simple and avoid handling directory changes (cd) from the script. The files are deleted in the finally block anyway. The only "risk" I see is if someone runs `git commit -A` at that moment.

ngxson (approved):

Issue #19585, opened 2026-02-13
Eval bug: Llama2 outputs an infinite loop.


blime4:
### Name and Version

cmake -B build -DGGML_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES="86" &&   cmake --build build --config Release -j 14 && CUDA_VISIBLE_DEVICES=0 ./build/bin/llama-cli -m ../llama-2-7b.Q4_0.gguf --temp 0.0 --top-k 1 --top_p 1 --repeat-penalty 1 -s 42 -p "tell me something about you？" --no-warmup -fa on -ngl 99

ggml_cuda_init: found 1 CUDA devices:
  Device 0: NVIDIA A40, compute capability 8.6, VMM: yes

build      : b8027-25224c802
model      : llama-2-7b.Q4_0.gguf
modalities : text

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

NVIDIA A40, compute capability 8.6, VMM: yes

### Models

llama-2-7b.Q4_0.gguf

### Problem description & steps to reproduce

how to reproduce it: cmake -B build -DGGML_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES="86" &&   cmake --build build --config Release -j 14 && CUDA_VISIBLE_DEVICES=0 ./build/bin/llama-cli -m ../llama-2-7b.Q4_0.gguf --temp 0.0 --top-k 1 --top_p 1 --repeat-penalty 1 -s 42 -p "tell me something about you？" --no-warmup -fa on -ngl 99

narrow down:
in the preview version of llama.cpp. qwen2.5-1.5b-instruct-q4_0.gguf also have the same problem, fixed by #19115 。 but all the version llama-2-7b.Q4_0.gguf have this output infinite loop bug. 
FA dispatch：
llama2 --> VEC kernel --> maybe here have some error.
qwen2 --> MMA_F16 (GQA)

### Relevant log output

<details>
<summary>Logs</summary>

```console
cmake -B build -DGGML_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES="86" &&   cmake --build build --config Release -j 14 && CUDA_VISIBLE_DEVICES=0 ./build/bin/llama-cli -m ../llama-2-7b.Q4_0.gguf --temp 0.0 --top-k 1 --top_p 1 --repeat-penalty 1 -s 42 -p "tell me something about you？" --no-warmup -fa on -ngl 99

[...]

PR #19592, opened 2026-02-13
fix: use llama2 template for llama architecture models without chat template

ssam18:
This fixes the infinite loop bug reported in #19585. When Llama-2 models don't have an embedded chat template in their GGUF metadata, they were incorrectly defaulting to ChatML format (with `<|im_start|>` and `<|im_end|>` tokens). The fix detects the model architecture and uses the appropriate Llama-2 template format instead. Tested locally with llama-2-7b.Q4_0.gguf and confirmed the issue is resolved.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -583,6 +583,18 @@ std::string common_chat_format_example(const struct common_chat_templates * tmpl
     "  {{- '<|im_start|>assistant\n' -}}\n" \
     "{%- endif -%}"
 
+#define LLAMA2_TEMPLATE_SRC \
+    "{{- bos_token }}{%- for message in messages -%}" \
+    "    {%- if message['role'] == 'system' -%}" \
+    "        {{- '[INST] <<SYS>>\\n' + message['content'] + '\\n<</SYS>>\\n\\n' -}}" \
+    "    {%- elif message['role'] == 'user' -%}" \
+    "        {{- '[INST] ' + message['content'] + ' [/INST]' -}}" \
+    "    {%- elif message['role'] == 'assistant' -%}" \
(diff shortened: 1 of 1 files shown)


pwilkin (changes requested):
Looks good, but please appropriate tests to `test-chat` as well.

common/chat.cpp
```diff
@@ -634,7 +646,18 @@ common_chat_templates_ptr common_chat_templates_init(
         if (!template_tool_use_src.empty()) {
             default_template_src = template_tool_use_src;
         } else {
-            default_template_src = CHATML_TEMPLATE_SRC;
+            char arch_buf[128];
+            if (model && llama_model_meta_val_str(model, "general.architecture", arch_buf, sizeof(arch_buf)) >= 0) {
+                std::string arch(arch_buf);
+                if (arch == "llama") {
```
ngxson, line 652: I won't approve this change. Many models using llama arch but not all of them use llama2 chat template. Hermes is a good example.

Also, this is a very hacky solution. If model doesn't have the correct template, it's probably already broken

ngxson (changes requested):

common/chat.cpp
ssam18: You're absolutely right! 
Architecture alone is not enough. Would you prefer: 
(A) checking if ChatML tokens exist in vocabulary before using ChatML as default 
(B) not defaulting to any template and warning users to use --chat-template, or 
(C) a different approach? 

The issue is real, though and models without templates currently get ChatML, which causes infinite loops when \`<|im_start|>\`/\`<|im_end|>\` tokens don't exist in their vocabulary.

PR #19593, opened 2026-02-13
docs : explicit about banning accounts that violates policy

ngxson:
Alternative to https://github.com/ggml-org/llama.cpp/pull/18735

Pushing this PR as a suggestion, but I think one of the main reason why people keep pushing AI-generated PR **even when the agent warned** them, is because they see no consequences of doing so.

Even when banning accounts is technically hard (as maintainers don't technically have the right to do so), I think explicitly telling this will have a psychological impact.


CISC:
Fine by me, but as you wrote may be hard to enforce, only @ggerganov can I think?

ngxson:
Only ggml-org members can access the moderator settings I think (github UX is not very convenient in this case, so kinda waste time to actually enforce)

am17an:
I think there is no easy way here, I saw this https://github.com/mitchellh/vouch a few days ago. Perhaps we can wait and see how this turns out for the ghostty project.

ngxson:
@am17an I just saw this too, the idea is nice (same point as I brought up about the "consequences"), but I think their intent is to actually enforce something. My proposal here doesn't enforce anything, just to make it clear so that people know the potential consequence of their action (pure psychological)

JohannesGaessler:
I'm with banning people if they go against policy but I'm not sure if we need to explicitly spell this out. I think one will be hard-pressed to find a project where repeated policy violations don't get you eventually banned.

ngxson:
@JohannesGaessler I agree that this doesn't need to be written explicitly, but under one condition that we do actually ban accounts that repeatedly send AI-generated PRs

CISC:
> @JohannesGaessler I agree that this doesn't need to be written explicitly, but under one condition that we do actually ban accounts that repeatedly send AI-generated PRs

If there are no actual consequences the words are meaningless anyway.

Anyway, @danbev is a ggml-org member, I guess that means we have at least two people with the power in Diego's absence?

Green-Sky:
:)

ggerganov (approved):
I think maintainers should warn users who do not follow the AI policy + close PRs/issues. After a few failed attempts (let's say arbitrary 3), maintainers can request a ban of the account. We can discuss and execute ban requests internally within the organization?

CISC (approved):

Green-Sky:
@aviallon :P

PR #19594, opened 2026-02-13
Add a build target to generate ROCm artifacts using ROCm 7.11

superm1:
This builds the following targets:
 * gfx906
 * gfx908
 * gfx90a
 * gfx942
 * gfx950
 * gfx1100
 * gfx1101
 * gfx1102
 * gfx1150
 * gfx1151
 * gfx1152
 * gfx1200
 * gfx1201

For size reasons this artifact does NOT contain the ROCm stack, the user will be required to install that separately from AMD or their distro to use the binaries in the artifact.

This is an ALTERNATIVE to using 7.2 (https://github.com/ggml-org/llama.cpp/pull/19433).  Let me explain.

AMD has made a change to the [build system](https://github.com/rocm/therock) and rather than doing monolithic builds now generates per-architecture builds.  This allows supporting more ISAs.  You can read more about it in the [release notes for 7.11](https://rocm.docs.amd.com/en/7.11.0-preview/about/release-notes.html) and various [blogs](https://rocm.blogs.amd.com/software-tools-optimization/therock/README.html) and documentation linked to them.

Because of this users don't need to install a very large stack that includes ALL of CDNA, RDNA3, RDNA4 ISAs etc for a single GPU ISA.

Because of this transition happening with AMD ROCm builds, I think it makes sense to NOT distribute ROCm "inside" a llama.cpp artifact.  The requirement should be a user installs ROCm first, and then can use the artifact.

1. Install ROCm per [AMD's documentation](https://rocm.docs.amd.com/en/7.11.0-preview/install/rocm.html?fam=ryzen&gpu=max-pro-395&os=ubuntu&os-version=24.04&i=pkgman) (or in some cases distros have it in native repos too).  This can be done via `apt`, `pip` or a tarball.
2. If using tarball set up LD_LIBRARY_PATH so that all libraries can be found.  This is part of AMD's documentation.
3. Download artifact
4. Run llama.cpp

You'll notice that I downloaded the GFX1151 tarball in the CI target.  To prove this approach works this is what I did.
1. Build the artifact using GFX1151 tarball (you can see CI run here: https://github.com/superm1/llama.cpp/actions/runs/21990274735)
2. Downloaded and extracted TheRock tarball for GFX1150 (very intentionally NOT GFX1151).
3. Set up `LD_LIBRARY_PATH` to cover the place I extracted the tarball (in my case `LD_LIBRARY_PATH=~/src/therock-install/lib`
4. Downloaded and extracted github artifact
5. Added path of github artifact to `LD_LIBRARY_PATH` as well
6. Ran `llama-server` and verified it works.

---
I _feel_ that this approach is scalable for AMD's future ROCm release strategy.  When there is a 7.12 release, a new matrix entry can be made for building ROCm 7.12 llama.cpp artifacts the same way.  Same thing with 7.13, etc.

CC @slojosic-amd and @IMbackK for comments as well

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -516,6 +516,102 @@ jobs:
           path: llama-bin-win-sycl-x64.zip
           name: llama-bin-win-sycl-x64.zip
 
+  ubuntu-22-rocm:
+    runs-on: ubuntu-22.04
+
+    strategy:
+      matrix:
+        include:
+          - ROCM_VERSION: "7.11.0"
+            gpu_targets: "gfx906;gfx908;gfx90a;gfx942;gfx950;gfx1100;gfx1101;gfx1102;gfx1150;gfx1151;gfx1200;gfx1201"
+            build: x64
+
+    steps:
+      - name: Clone
+        id: checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: ccache
+        uses: ggml-org/ccache-action@v1.2.16
+        with:
+          key: ubuntu-rocm-cmake-${{ matrix.ROCM_VERSION }}-${{ matrix.build }}
+          evict-old-files: 1d
+
+      - name: Dependencies
+        id: depends
+        run: |
+          sudo apt install -y build-essential git cmake wget
+
+      - name: Setup Legacy ROCm
+        if: matrix.ROCM_VERSION == '7.2'
+        id: legacy_env
+        run: |
+          sudo mkdir --parents --mode=0755 /etc/apt/keyrings
+          wget https://repo.radeon.com/rocm/rocm.gpg.key -O - | \
+            gpg --dearmor | sudo tee /etc/apt/keyrings/rocm.gpg > /dev/null
+
+          sudo tee /etc/apt/sources.list.d/rocm.list << EOF
+          deb [arch=amd64 signed-by=/etc/apt/keyrings/rocm.gpg] https://repo.radeon.com/rocm/apt/${{ matrix.ROCM_VERSION }} jammy main
+          EOF
+
+          sudo tee /etc/apt/preferences.d/rocm-pin-600 << EOF
+          Package: *
+          Pin: release o=repo.radeon.com
+          Pin-Priority: 600
+          EOF
+
+          sudo apt update
+          sudo apt-get install -y libssl-dev rocm-hip-sdk
+
+      - name: Setup TheRock
+        if: matrix.ROCM_VERSION != '7.2'
+        id: therock_env
+        run: |
+          wget https://repo.amd.com/rocm/tarball/therock-dist-linux-gfx1151-${{ matrix.ROCM_VERSION }}.tar.gz
+          mkdir install
+          tar -xf *.tar.gz -C install
+          export ROCM_PATH=$(pwd)/install
+          echo ROCM_PATH=$ROCM_PATH >> $GITHUB_ENV
+          echo PATH=$PATH:$ROCM_PATH/bin >> $GITHUB_ENV
+          echo LD_LIBRARY_PATH=$ROCM_PATH/lib:$ROCM_PATH/llvm/lib:$ROCM_PATH/lib/rocprofiler-systems >> $GITHUB_ENV
+
+      - name: Build with native CMake HIP support
+        id: cmake_build
+        run: |
+          cmake -B build -S . \
+            -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" \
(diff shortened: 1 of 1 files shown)


IMbackK:
The problem with this strategy is that it assumes that you have only one type of gpu in your system.
Having multiple different gpus in one system is decidedly not uncommon in the llamacpp community - making this strategy non-viable at the moment.

In the future we could adjust the hip backend to generate several ggml backends for different architecture sets, which would solve most issues with split builds, since you can use multiple ggml backends at the same time.

IMbackK:
for now we need to focus on providing as few binaries as possible - to support as many heterogeneous systems as possible.

superm1:
> The problem with this strategy is that it assumes that you have only one type of gpu in your system.
Having multiple different gpus in one system is decidedly not uncommon in the llamacpp community - making this strategy non-viable at the moment.

When you say one type of GPU, do you mean "HIP" vs "CUDA"?  Or do you mean one AMD ISA and another AMD ISA?

Because at least for multiple AMD ISAs you can install multiple tarballs just fine.  Most of the binaries in the TheRock tarball are identical between ISAs, you just end up with extra low level tensile libraries for your ISA's specific architecture.  Those are the big things anyway (~400MB per ISA or so).

> for now we need to focus on providing as few binaries as possible - to support as many heterogeneous systems as possible.

Yeah this is why I am explicitly building for all the AMD GPU targets that have TheRock binaries published, and why I did a test to prove that building with one ISA's build environment I can install libraries for another and it works.

IMbackK:
You seem to have a fundamental misunderstanding on how these rocm objects are created.
So the therock builds are built by specifying all architectures at compile time and the splitting the just the tarballs along the files that the build systems of the rocm subprojects create that are architecture specific, sutch as the tensile library files.

This is why simply merging those tarballs again works, the common binaries already contain gpu code for all isas.

What you are doing here is building llamacpp for each architecture in a matrix build, this means each binary artifact created by this workflow contains gpu code for only a single isa.
What gpus the rocm installation supports dosent matter in this case, that llamacpp build will only ever be able to run on that one single isa as i contains all of our custom kernels only for that one isa.

superm1:
>  So the therock builds are built by specifying all architectures at compile time and the splitting the just the tarballs along the files that the build systems of the rocm subprojects create that are architecture specific, sutch as the tensile library files.

Right.

> What you are doing here is building llamacpp for each architecture in a matrix build, this means each binary artifact created by this workflow contains gpu code for only a single isa.

Actually it's a "single matrix" build that does every single ISA supported.  It only generates a single build, and single llamacpp artifact.  The purpose of organizing it the way I did is so that when ROCm 7.12 comes out we could add a new matrix configuration that says "7.12 builds with these GPU ISAs (in case it's additively different).

Tthe ROCM_VERSION shouldn't be in the env for the job, it should be in the matrix.  I've made a change to make this point clearer.

>  What gpus the rocm installation supports dosent matter in this case, that llamacpp build will only ever be able to run on that one single isa as i contains all of our custom kernels only for that one isa.

If I had done 15 builds in a matrix; yes I would agree with you.

IMbackK:
Right the single element matrix with matrix.gpu_targets threw me off.

Some thoughts:

i think we should stick with the official builds of rocm for now, not the therock builds.

I would lean towards not bundling rocm, if someone wants everything bundled we have the docker for that purpose after all, however rocm dosent really have a stable abi (i mean the runtime here, even ignoring the math libraries). I mean they claim they do, but it breaks somewhere essentially every release. Inevitably there would be friction here where things break when we upgrade the version we build against - or when a new rocm version releases and people upgrade before we change the release.
We can mitigate this a bit by providing releases built against multiple versions.

superm1:
> i think we should stick with the official builds of rocm for now, not the therock builds.

Does it need to be either-or?  How about we do both; IE merge both of the PRs?  This can let people more easily benchmark differences with the stable series and the preview series.

>  We can mitigate this a bit by providing releases built against multiple versions.

To avoid explosion of too many artifact options  we can potentially treat the 7.11 one as an experimental one that will go away when 7.12 comes out.  If 7.13 is still a preview release, 7.12 goes away.

superm1:
Here's artifacts from a run with both 7.2 and 7.11 artifacts:

https://github.com/superm1/llama.cpp/actions/runs/22011336432

You'll notice that I've adjusted these pull requests to basically be identical, just one uses 7.2 and one uses 7.11.
When I did my CI build I used it as a matrix that did both.

If you do agree to do two artifacts it's just 3 lines to add the matrix for the other to either PR.

superm1:
Closing this PR per guidance in https://github.com/ggml-org/llama.cpp/pull/19433 that we will only do 7.2 for now.  When ready to shift to TheRock builds then it's just two lines to add targets accordingly.

Issue #19513, opened 2026-02-11
Eval bug: Qwen3-Coder-Next generates prematurely EOS instead of tool call(/continued response)


bfroemel:
### Name and Version

./llama-cli --version
ggml_vulkan: No devices found.
load_backend: loaded Vulkan backend from /app/libggml-vulkan.so
load_backend: loaded CPU backend from /app/libggml-cpu-zen4.so
version: 7993 (6d0ee8d4d)
built with GNU 15.2.0 for Linux x86_64

### Operating systems

Linux

### GGML backends

Vulkan

### Hardware

AMD Strix Halo, 128 GB

### Models

Qwen/Qwen3-Coder-Next-GGUF Q8_0
unsloth/Qwen3-Coder-Next-GGUF UD_Q5_K_XL, UD_Q8_K_XL

### Problem description & steps to reproduce

With codex on Qwen3-Coder-Next using the responses API I often see something like this after about 10-20 successful tool calls (EOS token not part of the output, it just directly follows the colon, hence the model generation must stop):

```
Let me check that other file:<|im_end|>
```

And the agent turn ends prematurely (because it encountered the EOS token and no tool call was generated). If I tell the model to continue, it does so for a while, but after 1 to 10 tool calls the same issue pops up (until the task is finally done; could take a lot of "please continue!" messages).

For messages where the tool call is not generated the colon token appears to always be this one (and is directly followed by the EOS token):

```
next token:    25 ':'
```

For messages where the tool call is generated a different colon token is used (directly before token 151657 '<tool_call>'):

```
next token:  1447 ':

'
```

I am using the recommended sampler settings (`--temp 1.0 --min-p 0.01 --top-p 0.95 --top-k 40`), unsloth's UD-Q8_K_XL (and also Qwen's Q8_0); I haven't noticed any other quality issues. The model is supposed to be very good at tool calls; as it was fine-tuned on all major formats, not just Qwen3-XML (see https://github.com/QwenLM/Qwen3-Coder/blob/main/qwen3_coder_next_tech_report.pdf?spm=a2ty_o06.30285417.0.0.3bdec921p0xaBK&file=qwen3_coder_next_tech_report.pdf Section 4.2.2 + Table 12). At least one other person has experienced the same behavior, while confirming that other harnesses seemingly could give better results.

(Note that I have seen successful/uninterrupted agent turns with 100+ tool calls, but they are rare!

**Reproducing the issue**
[...]

PR #19595, opened 2026-02-13
mtmd : chat : Fix extra \n between text and media marker

tdakhran:
Thanks to @tugot17 for detecting and reporting the issue.

For vision models (e.g. LFM2.5-VL-1.6B and Qwen/Qwen3-VL-4B-Instruct) `llama-mtmd-cli` produces identical output to HF implementation.

However `llama-server` doesn't. I traced it down to extra newline inserted after `<__media__>`.

This happens in `to_json_oaicompat`, that treats media markers as text and joins all parts with `\n` separator.

PR introduces new type `media_marker` and uses it for media markers. Extra logic is added to prevent insertion of newlines before and after media markers.

With this change number of input tokens is identical to HF implementation and as a result the output is also identical.

I explored other ways to address the issue
* remove completely `\n` between text parts in `to_json_oaicompat`
* merge text messages in server-common.cpp before sending them to `to_json_oaicompat`

Please propose alternative ways of fixing this issue.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -84,7 +84,7 @@ json common_chat_msg::to_json_oaicompat(bool concat_typed_text) const {
                     text += '\n';
                 }
 
-                text           += part.text;
+                text += part.text;
             }
             jmsg["content"] = text;
         } else {
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -68,14 +68,22 @@ json common_chat_msg::to_json_oaicompat(bool concat_typed_text) const {
             bool last_was_media_marker = false;
             // join parts with newline, do not add newline before or after media markers
             for (const auto & part : content_parts) {
-                if (part.type != "text" && part.type != "media_marker") {
+                bool add_new_line = true;
+                if (part.type == "text") {
+                    add_new_line = !last_was_media_marker && !text.empty();
+                    last_was_media_marker = false;
+                } else if (part.type == "media_marker") {
+                    add_new_line = false;
+                    last_was_media_marker = true;
+                } else {
                     LOG_WRN("Ignoring content part type: %s\n", part.type.c_str());
                     continue;
                 }
-                if (part.type != "media_marker" && !last_was_media_marker && !text.empty()) {
+
+                if (add_new_line) {
                     text += '\n';
                 }
-                last_was_media_marker  = (part.type == "media_marker");
+
                 text           += part.text;
             }
             jmsg["content"] = text;
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -65,15 +65,18 @@ json common_chat_msg::to_json_oaicompat(bool concat_typed_text) const {
     } else if (!content_parts.empty()) {
         if (concat_typed_text) {
             std::string text;
+            bool last_was_media_marker = false;
+            // join parts with newline, do not add newline before or after media markers
             for (const auto & part : content_parts) {
-                if (part.type != "text") {
+                if (part.type != "text" && part.type != "media_marker") {
                     LOG_WRN("Ignoring content part type: %s\n", part.type.c_str());
                     continue;
                 }
-                if (!text.empty()) {
+                if (part.type != "media_marker" && !last_was_media_marker && !text.empty()) {
                     text += '\n';
                 }
-                text += part.text;
+                last_was_media_marker  = (part.type == "media_marker");
+                text           += part.text;
             }
             jmsg["content"] = text;
         } else {
@@ -319,7 +322,7 @@ std::vector<common_chat_msg> common_chat_msgs_parse_oaicompat(const json & messa
                             throw std::invalid_argument("Missing content part type: " + part.dump());
                         }
                         const auto & type = part.at("type");
-                        if (type != "text") {
+                        if (type != "text" && type != "media_marker") {
                             throw std::invalid_argument("Unsupported content part type: " + type.dump());
                         }
                         common_chat_msg_content_part msg_part;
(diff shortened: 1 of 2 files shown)


tools/server/server-common.cpp
```diff
@@ -916,8 +916,7 @@ json oaicompat_chat_params_parse(
                 json image_url = json_value(p, "image_url", json::object());
                 handle_media(out_files, image_url, opt.media_path);
 
-                // replace this chunk with a marker
-                p["type"] = "text";
+                p["type"] = "media_marker";
                 p["text"] = mtmd_default_marker();
```
ngxson, line 920: I think a simple fix like this should be enough:

```suggestion
                p["text"] = "\n" + mtmd_default_marker() + "\n";
```

tdakhran: @ngxson , not sure I understand, I try to remove `\n`, not add them

tdakhran: with existing code, we end up with `<__media__>\nDescribe the image.` instead of expected `<__media__>Describe the image.`.

ngxson: ah yeah sorry I misread. I think the `\n` between image and text should be expected by most models. I assume having `\n` or not doesn't change much the output?

ngxson: hmm, nevermind, you said that `mtmd-cli` doesn't add the newline either (not sure it was removed since when). But if that's the behavior of HF then it's ok to fix

common/chat.cpp
```diff
@@ -65,15 +65,18 @@ json common_chat_msg::to_json_oaicompat(bool concat_typed_text) const {
...
+            bool last_was_media_marker = false;
+            // join parts with newline, do not add newline before or after media markers
             for (const auto & part : content_parts) {
-                if (part.type != "text") {
+                if (part.type != "text" && part.type != "media_marker") {
                     LOG_WRN("Ignoring content part type: %s\n", part.type.c_str());
                     continue;
                 }
-                if (!text.empty()) {
+                if (part.type != "media_marker" && !last_was_media_marker && !text.empty()) {
```
ngxson: for clarity, this should be refactored info:

```cpp
if (part.type == "text") {
  ...
} else if (part.type == "media_marker") {
  ...
} else {
  LOG_WRN("Ignoring content part type: %s\n", part.type.c_str());
}
```

tdakhran: I'll wait CI run to complete in case some tests fail and then refactor.

tools/server/server-common.cpp
tdakhran: at least `LFM2.5-VL-1.6B` and `Qwen/Qwen3-VL-4B-Instruct` output differ from HF, and this breaks our internal correctness tests

pwilkin: Actually, you can look to https://github.com/ggml-org/llama.cpp/issues/19513 for a very interesting cautionary tale on how modifying newline outputs can influence model behavior. I think it's worth fixing TBH.

common/chat.cpp
tdakhran: @ngxson , shall I duplicate the body then in each of them? I'm struggling to understand how to split it because it's an && condition.

ngxson: the body is small, it's fine to duplicate it. we just need to be explicit about what are possible values of `type`

tdakhran: refactored in https://github.com/ggml-org/llama.cpp/pull/19595/commits/b642c339fe1a4c9baedda6207fdc2fb78e75385d

tdakhran:
Addressed existing comments. Please let me know if anything else is required to proceed with the merge.

common/chat.cpp
```diff
@@ -65,15 +65,26 @@ json common_chat_msg::to_json_oaicompat(bool concat_typed_text) const {
...
                     continue;
                 }
-                if (!text.empty()) {
+
+                if (add_new_line) {
                     text += '\n';
                 }
-                text += part.text;
+
+                text           += part.text;
```
pwilkin:
```suggestion
                text += part.text;
```

common/chat.cpp
```diff
@@ -65,15 +65,26 @@ json common_chat_msg::to_json_oaicompat(bool concat_typed_text) const {
...
+                    last_was_media_marker = false;
+                } else if (part.type == "media_marker") {
+                    add_new_line = false;
+                    last_was_media_marker = true;
+                } else {
                     LOG_WRN("Ignoring content part type: %s\n", part.type.c_str());
                     continue;
                 }
-                if (!text.empty()) {
+
```
pwilkin, line 82:
```suggestion
```

pwilkin (approved):
Please just fix the whitespace stuff.

common/chat.cpp
tdakhran: imo this is more readable with line separating one `if` from another `if`

tdakhran: here how it looks without
image

common/chat.cpp
tdakhran: missed it during refactor, thanks

common/chat.cpp
pwilkin: Aight.

tdakhran:
I'm looking into server-windows CI failure, might be related.

tdakhran:
Let's see now, updated legacy chat template handling.

tdakhran:
CI passed, good to go.

tdakhran:
@pwilkin, please let me know if anything is needed to proceed with the merge.

ngxson (approved):

PR #19597, opened 2026-02-13
models : deduplicate delta-net graphs for Qwen family

ggerganov:
cont #19375 

- Add `llm_build_delta_net_base` for common delta net builds. Currently used only by `qwen3next`
- Rename `llm_graph_context_mamba` -> `llm_build_mamba_base`
- Includes clean-up

Next PRs:

- [x] Use `llm_build_delta_net_base` for `qwen35` and `qwen35moe` (after release)
- [x] Adapt and use `llm_build_delta_net_base` for `kimi_linear` (cc @ymcki)


pwilkin:
@ggerganov out of curiosity - what's the reason for moving the identity / mask constructions from once-per-graph to once-per-layer?

ggerganov:
> @ggerganov out of curiosity - what's the reason for moving the identity / mask constructions from once-per-graph to once-per-layer?

Mainly wanted to get rid of the `ggml_new_tensor()` calls. In this case, it does not cause any issues, but it can become problematic if the same tensor created with `ggml_tensor_new()` would be used for different ops that would be performed on different backends (for example due to offloading or missing support). As I mentioned, it's always better to find alternatives that don't involve creating new tensors.

Btw, creating a mask and then multiplying does not have advantage compare to just masking the data (with `ggml_tri()`), so it seems simpler too.

ymcki:
Is this the new home for the unified delta_net effort? My job is to modify build_delta_net_chunking to use "if (is_kda) {} else {}" for kda and gdn?

ggerganov:
Yes, we'll merge this PR soon and then the new `llm_build_delta_net_base` should be adapted to work with both GDA and KDA. Ideally, the KDA would have to add as few extra ops as possible - likely utilizing broadcasts when possible. The KDA support should be a standalone change, without any additional optimizations in order to keep things simple.

ymcki:
@ggreganov, I am done with adding kimi linear to delta-net-base.cpp.

https://github.com/ymcki/llama.cpp/tree/dn

The original delta net path is the same except that I added o = ggml_cont(ctx0, o); after ggml_permute at the end.

This is to avoid the assertion error:
/home/user/gguf/gg/llama.cpp/ggml/src/ggml.c:3573: GGML_ASSERT(ggml_is_contiguous(a)) failed
raised by
ggml_tensor * attn_out_final = ggml_reshape_3d(ctx0, output, head_dim, n_head,  n_seq_tokens * n_seqs);

Do I open another PR? Or you will do that yourself?

ggerganov:
Please open a PR. I will attempt to simplify the implementation - hoping we can unify this and avoid branching on the `g` dimensions. Not sure if possible though - we'll see.

PR #19606, opened 2026-02-13
convert : store ffn_gate_inp_shexp as F32

CISC:
This tensor has been inadvertently stored as BF16 in several models, and since it's 1D it will never be quantized.

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -570,6 +570,7 @@ class ModelBase:
                         self.match_model_tensor_name(new_name, key, bid)
                         for key in (
                             gguf.MODEL_TENSOR.FFN_GATE_INP,
+                            gguf.MODEL_TENSOR.FFN_GATE_INP_SHEXP,
                             gguf.MODEL_TENSOR.POS_EMBD,
                             gguf.MODEL_TENSOR.TOKEN_TYPES,
                             gguf.MODEL_TENSOR.SSM_CONV1D,


ggerganov (approved):

ggerganov:
@CISC High-jacking this thread to report a minor regression that I observe. I need the following patch to be able to convert https://huggingface.co/Qwen/Qwen3-30B-A3B-Base:

```diff
diff --git a/convert_hf_to_gguf.py b/convert_hf_to_gguf.py
index 825080b58..1e5209cdc 100755
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -4161,7 +4161,7 @@ class Qwen2MoeModel(TextModel):
             return
 
         if name.find("experts") != -1:
-            n_experts = self.hparams["num_experts"]
+            n_experts = self.find_hparam(["num_local_experts", "num_experts"])
             assert bid is not None
 
             if self._experts is None:
```

CISC:
> @CISC High-jacking this thread to report a minor regression that I observe. I need the following patch to be able to convert https://huggingface.co/Qwen/Qwen3-30B-A3B-Base:

Hmmm, ok, not sure why that would have regressed, I'll look into it.

ggerganov:
Last I tested in mid-Jan, it was converting successfully with this commit: 2bbe4c2cf. Not sure if a problem in the convert script, or in something in the Python dependencies has changed.

CISC:
> Last I tested in mid-Jan, it was converting successfully with this commit: [2bbe4c2](https://github.com/ggml-org/llama.cpp/commit/2bbe4c2cf8298114e3908e285125b9d0d1c5bb42). Not sure if a problem in the convert script, or in something in the Python dependencies has changed.

Are you sure? I see the non-base model uses `num_experts`, I think this just changed in `transformers` by the time they released the base model.

CISC:
> > Last I tested in mid-Jan, it was converting successfully with this commit: [2bbe4c2](https://github.com/ggml-org/llama.cpp/commit/2bbe4c2cf8298114e3908e285125b9d0d1c5bb42). Not sure if a problem in the convert script, or in something in the Python dependencies has changed.
> 
> Are you sure? I see the non-base model uses `num_experts`, I think this just changed in `transformers` by the time they released the base model.

Edit: Ah, I see the problem, it's `transformers` itself that changed, it's `num_experts` in the config, but `AutoConfig` returns `num_local_experts`.

PR #19608, opened 2026-02-13
llama : add group feature to split-mode to minimize GPU spread for running a model

dan-and:
Merge request for #19607 

(Pre-)Checks for this PR: 
- I have searched for similar requests and PRs. It is unique. 
- I ran the full ci/run.sh  Runner (for CUDA) successfully. 
- No perplexity or performance degradation is introduced by the PR. (It even makes llama.cpp faster with multiple GPU setups) 

I would appreciate it if, @ggerganov could take a look at my PR as, based on the CODEOWNERS document, I touched mostly Georgi's code. Of course I am happy with any maintainer who takes the time to check my code and gives me feedback on it. 

Basically the main changed code is in src/llama.cpp : 
What I do: 
- I use a rough calculation of the model file size and multiply it by 1.5 ( MEMORY_ESTIMATE_RATIO ) to reach a size that is in the safe region to fit into VRAM. 
- Sort all available GPUs by their current available VRAM
- Select the minimum subset of GPUs (largest free RAM first) to fit the required memory estimation. 

Limitations of my tests: 
- I do not have any ROCM based GPUs, so I would really appreciate some additional tests and feedback from people with several GPUs that are not CUDA-related. 
- While I tried to make the code as universal as possible, I only tested it with Linux operating systems. Feedback from users with other operating systems is appreciated.


bchtrue:
useful! we need this

dan-and:
won't implement

PR #19609, opened 2026-02-13
ggml: ggml-cpu: force-no-lto-for-cpu-feats

talhaHavadar:
When LTO enabled in build environments it forces all builds to have LTO in place. But feature detection logic is fragile, and causing Illegal instruction errors with lto. This disables LTO for the feature detection code to prevent cross-module optimization from inlining architecture-specific instructions into the score function. Without this, LTO can cause SIGILL when loading backends on older CPUs (e.g., loading power10 backend on power9 crashes before feature check runs).

Please also see https://salsa.debian.org/deeplearning-team/ggml/-/merge_requests/6 for more information about the issue we saw on ppc64el builds with LTO enabled in ubuntu.

### DIFF
--- a/ggml/src/ggml-cpu/CMakeLists.txt
+++ b/ggml/src/ggml-cpu/CMakeLists.txt
@@ -9,6 +9,11 @@ function(ggml_add_cpu_backend_features cpu_name arch)
     target_compile_definitions(${GGML_CPU_FEATS_NAME} PRIVATE ${ARGN})
     target_compile_definitions(${GGML_CPU_FEATS_NAME} PRIVATE GGML_BACKEND_DL GGML_BACKEND_BUILD GGML_BACKEND_SHARED)
     set_target_properties(${GGML_CPU_FEATS_NAME} PROPERTIES POSITION_INDEPENDENT_CODE ON)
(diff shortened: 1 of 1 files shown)


taronaeo:
@shalinib-ibm Maybe you could take a look at this? They have a discussion going on at https://salsa.debian.org/deeplearning-team/ggml/-/merge_requests/6 already.

ckastner:
While I'm not entirely sure what the root cause for the `SIGILL` is, the fix at least seems consistent with the strategy for the CPU feature detection stuff: build that part with as few features as possible, so that it can run everywhere.

ggerganov (approved):
Waiting for CI before merge.

shalinib-ibm:
> @shalinib-ibm Maybe you could take a look at this? They have a discussion going on at https://salsa.debian.org/deeplearning-team/ggml/-/merge_requests/6 already.

Hi all, thanks for flagging this. Since LTO is disabled only for the CPU feature-detection code, this change has no impact on performance-critical paths. We verified this by running llama-bench with and without the change and observed no performance regressions.

taronaeo (approved):

PR #19612, opened 2026-02-14
fix: GLM 4.5 streaming tool-call parsing + grammar error handling

Gunther-Schulz:
**Disclosure: All code and this PR description were written by AI (Cursor).**

---

## Summary

This PR fixes GLM 4.5 (and related XML tool-call) streaming behaviour so that:
1. The server no longer hangs when `tool_choice=auto` and the model never outputs the grammar trigger (parse-only path, no grammar for AUTO).
2. The "Failed to parse up to error: ... attempting to parse an empty input" log spam is removed when streaming partial tool-call arguments.
3. Plain-text `arg_value` content (e.g. `subagent_type=explore`) is parsed correctly instead of throwing partial forever, so Task/tool calls complete.
4. Grammar/sampler runtime errors (e.g. "Unexpected empty grammar stack") return HTTP 500 and release the slot instead of aborting the server.

### 1. GLM 4.5 parse-only for AUTO (existing branch commit)
- **common/chat.cpp (GLM 4.5 init):** Use grammar only when `tool_choice == required`; for `auto`, do not set grammar/trigger so tool calls are detected by parsing decoded text (vLLM-style).
- **Tests:** Relax test-chat assert to allow empty grammar when the test message has tool_calls.
- Addresses server hang when the model never outputs the trigger (e.g. #19068).

### 2. Avoid "parse empty input" log (json-partial + XML parser)
- **common/json-partial.cpp:** When the SAX parser reports an error at position 0 (empty substring to re-parse), return `false` immediately without calling `json::parse` or logging "Failed to parse up to error". This avoids noisy logs for all formats (GLM 4.5, GPT-OSS, Granite, Hermes, etc.) when streaming partial or invalid JSON.
- **common/chat-parser-xml-toolcall.cpp:** Only call `try_consume_json()` when the remainder after `<arg_value>` looks like the start of a JSON value (`"`, `{`, `[`, digit, `-`, or prefix of `true`/`false`/`null`). Otherwise treat as incomplete/plain text and skip the JSON parser (vLLM-style), avoiding SAX error-at-position-0 for e.g. `exp` before `explore`.

### 3. Plain-text arg_value fallthrough + server grammar error handling
- **common/chat-parser-xml-toolcall.cpp:** When the remainder does not look like JSON (e.g. `explore`), do not throw; skip `try_consume_json()` and fall through to the existing plain-text path (`try_find_val_end()`). Fixes Task tool (e.g. `subagent_type=explore`) never completing.
- **tools/server/server-context.cpp:** Wrap `common_sampler_accept()` and `common_sampler_sample_and_accept_n()` in try/catch for `std::runtime_error`. On catch (e.g. "Unexpected empty grammar stack" from llama-grammar.cpp), log, send HTTP 500 with "Grammar constraint violation", release the slot, and continue so the server stays up.

## Testing

- `test-json-partial` and `test-chat-parser` pass.
- Manual: GLM 4.5 with `tool_choice=auto`, streaming tool calls (e.g. Task with `subagent_type=explore`) complete without log spam; server does not abort on grammar errors.

## Related

- vLLM behaviour: GLM-4 parser only parses non-string arg values when `</arg_value>` is seen; we align with that.
- Fixes / improves experience around streaming XML tool calls and grammar constraint violations in the server.

### DIFF
--- a/common/chat-parser-xml-toolcall.cpp
+++ b/common/chat-parser-xml-toolcall.cpp
@@ -498,7 +498,9 @@ inline bool parse_xml_tool_calls(common_chat_msg_parser & builder, const struct 
             }
             auto val_start = builder.pos();
 
-            // vLLM-style: only try to parse value when there is content; empty/whitespace or non-JSON start = incomplete (avoids "parse empty input" log)
+            // vLLM-style: only try to parse value when there is content; empty/whitespace = incomplete (avoids "parse empty input" log).
+            // When remainder does not look like JSON start, skip try_consume_json and fall through to plain-text path (e.g. "explore").
+            bool looks_like_json = true;
             {
                 const auto & inp = builder.input();
                 const size_t rem_len = (val_start < inp.size()) ? (inp.size() - val_start) : 0;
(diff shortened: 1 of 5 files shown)


pwilkin:
Obsoleted by #18675

PR #19526, opened 2026-02-11
add `--dry-run` option to `llama-quantize`

ddh0:
This PR adds a new `--dry-run` option to `llama-quantize`. This option calculates the size of each tensor in the target type without actually performing quantization, and prints the final quantization size in the same way that `llama-quantize` does currently.

Example command:

```sh
llama-quantize --dry-run gemma-3-4b-it-q8_0.gguf Q4_K
```

Example output:

```
main: build = 8015 (07f882bbb)
main: built with AppleClang 17.0.0.17000603 for Darwin arm64
main: calculating quantization size for '/Users/dylan/Documents/AI/gguf/gemma-3-4b-it-q8_0.gguf' as Q4_K
llama_model_loader: loaded meta data with 41 key-value pairs and 444 tensors from /Users/dylan/Documents/AI/gguf/gemma-3-4b-it-q8_0.gguf (version GGUF V3 (latest))
llama_model_loader: Dumping metadata keys/values. Note: KV overrides do not apply in this output.
llama_model_loader: - kv   0:                       general.architecture str              = gemma3
llama_model_loader: - kv   1:                               general.type str              = model
llama_model_loader: - kv   2:                               general.name str              = gemma-3-4b-it
llama_model_loader: - kv   3:                           general.finetune str              = it
llama_model_loader: - kv   4:                           general.basename str              = gemma-3
llama_model_loader: - kv   5:                         general.size_label str              = 4B
llama_model_loader: - kv   6:                            general.license str              = gemma
llama_model_loader: - kv   7:                   general.base_model.count u32              = 1
llama_model_loader: - kv   8:                  general.base_model.0.name str              = Gemma 3 4b Pt
llama_model_loader: - kv   9:          general.base_model.0.organization str              = Google
llama_model_loader: - kv  10:              general.base_model.0.repo_url str              = https://huggingface.co/google/gemma-3...
llama_model_loader: - kv  11:                               general.tags arr[str,1]       = ["image-text-to-text"]
llama_model_loader: - kv  12:                      gemma3.context_length u32              = 131072
llama_model_loader: - kv  13:                    gemma3.embedding_length u32              = 2560
llama_model_loader: - kv  14:                         gemma3.block_count u32              = 34
llama_model_loader: - kv  15:                 gemma3.feed_forward_length u32              = 10240
llama_model_loader: - kv  16:                gemma3.attention.head_count u32              = 8
[...]

PR #19616, opened 2026-02-14
quantize : refactor llama-quant.cpp

ddh0:


ddh0:
Things got messy since this PR included code from #19526, which is now in master, but the two got out of sync. Moving to a new PR soon.

ddh0:
Moved to #19770.

PR #19620, opened 2026-02-14
Fix gpt-oss tool calling: pass tool args and tool responses as json

matteoserva:
gpt-oss template expects json objects for tool arguments and tool responses.
llama.cpp passes strings, causing them to be escaped twice.

fixes https://github.com/ggml-org/llama.cpp/issues/19520
supersedes https://github.com/ggml-org/llama.cpp/pull/19553

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -2029,9 +2029,25 @@ static common_chat_params common_chat_params_init_gpt_oss(const common_chat_temp
         auto has_reasoning_content = msg.contains("reasoning_content") && msg.at("reasoning_content").is_string();
         auto has_tool_calls = msg.contains("tool_calls") && msg.at("tool_calls").is_array();
 
-        if (has_reasoning_content && has_tool_calls) {
+
+        if (msg.at("role") == "tool")
+        {
+                auto adjusted_message = msg;
+                std::string response = msg.value("content",std::string());
+                adjusted_message["content"] = json::parse(response);
+                adjusted_messages.push_back(adjusted_message);
+        } else if (has_tool_calls) {
             auto adjusted_message = msg;
-            adjusted_message["thinking"] = msg.at("reasoning_content");
+            if(has_reasoning_content)
+            {
+                adjusted_message["thinking"] = msg.at("reasoning_content");
+            }
+            for ( auto& tool_call : adjusted_message.at("tool_calls"))
+            {
+                auto& function = tool_call.at("function");
+                std::string args = function.value("arguments",std::string());
+                function["arguments"] = json::parse(args);
+            }
             adjusted_messages.push_back(adjusted_message);
         } else {
             adjusted_messages.push_back(msg);


common/chat.cpp
```diff
@@ -2029,9 +2029,25 @@ static common_chat_params common_chat_params_init_gpt_oss(const common_chat_temp
         auto has_reasoning_content = msg.contains("reasoning_content") && msg.at("reasoning_content").is_string();
         auto has_tool_calls = msg.contains("tool_calls") && msg.at("tool_calls").is_array();
 
-        if (has_reasoning_content && has_tool_calls) {
+
+        if (msg.at("role") == "tool")
+        {
+                auto adjusted_message = msg;
+                std::string response = msg.value("content",std::string());
+                adjusted_message["content"] = json::parse(response);
```
aldehir, line 2037: There is no guarantee clients return back valid JSON as the tool response. Perhaps fallback to string when parsing fails.

aldehir (changes requested):
Need a better way to handle tool content. This will break clients that return plain strings, which is perfectly valid.

Also, have you tested this against the Unsloth template? They patched this particular issue, and it appears they handle both cases but needs to be verified.

Additionally, please adjust the code style to match the rest of the file.

matteoserva:
You were right. The updated model fixed everything. Thanks very much

bfroemel:
> You were right. The updated model fixed everything. Thanks very much

Unfortunately, the template fix wasn't picked up by everyone, e.g., most importantly upstream (openai), and even https://huggingface.co/ggml-org/gpt-oss-120b-GGUF hasn't. So IMO the proposed code fix (with suggested changes by @aldehir ) could make sense and shouldn't interfere with the unsloth template fix.

btw: I couldn't really understand why some reported on reddit bad gpt-oss performance and always attributed it to some kind of openai bashing and/or bots. Now I see that issues like that are indistinguishable from bad model performance for average users. It would be great if we had some automated way to verify whether the model sees valid prompts; maybe in some future we have a small set of end2end testcases per model where API requests are verifiable against given prompts.

pwilkin:
@bfroemel that's exactly how the autoparser project started :) I wanted to have llama.cpp as a reliable local agentic coding backend and to my frustration I realized that like 5 theoretically supported models that I tested in a row broke down within the first few tool call requests. Then I started digging and I realized that a lot of what is commonly attribute to "bad model quality" is actually bad parser behavior.

aldehir:
I agree, this should be fixed for all templates. I'll try to bring this PR to the finish line when I find time, unless @matteoserva wants to. An easy workaround is to use the Unsloth template for now.

matteoserva:
> I agree, this should be fixed for all templates. I'll try to bring this PR to the finish line when I find time, unless @matteoserva wants to. An easy workaround is to use the Unsloth template for now.

Yes, please do. :)

PR #19622, opened 2026-02-14
[server] save generated text for the /slots endpoint (for LLAMA_SERVER_SLOTS_DEBUG=1)

matteoserva:
when the environment variable `LLAMA_SERVER_SLOTS_DEBUG=1` is set, the `/slots` endpoint allows reading the last prompt and the last generated text.

The slot reset also deletes the generated text, so it always looks empty when the slot is idle.

This PR adds a variable to allow keeping the data until a new generation is started.

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -426,7 +426,7 @@ struct server_slot {
 
             if (!only_metrics) {
                 res["prompt"] = ptask->tokens.detokenize(ctx, true);
-                res["generated"] = generated_text.empty()?debug_generated_text:generated_text;
+                res["generated"] = generated_text.empty() ? debug_generated_text : generated_text;
             }
         }
 
@@ -1445,7 +1445,7 @@ private:
         res->index           = slot.task->index;
 
         // keep copy of last generated text for debugging purposes
-        if (slots_debug > 0) {
+        if (slots_debug) {
             slot.debug_generated_text = slot.generated_text;
         }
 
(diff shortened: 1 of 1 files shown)


tools/server/server-context.cpp
```diff
@@ -1442,6 +1443,10 @@ struct server_context_impl {
         res->id_slot = slot.id;
 
         res->index           = slot.task->index;
+
+        // keep copy of last generated text for debugging purposes
+        slot.debug_generated_text = slot.generated_text;
```
ngxson: It's preferable to only do this if slots debug is enabled. Leaving this code in prod environment can cause performance problem if the `generated_text` is large

ngxson (changes requested):

tools/server/server-context.cpp
matteoserva: Solved

tools/server/server-context.cpp
```diff
@@ -1442,6 +1443,12 @@ struct server_context_impl {
         res->id_slot = slot.id;
 
         res->index           = slot.task->index;
+
+        // keep copy of last generated text for debugging purposes
+        if (slots_debug > 0) {
```
ngxson:
```suggestion
        if (slots_debug) {
```

tools/server/server-context.cpp
```diff
@@ -425,7 +426,7 @@ struct server_slot {
 
             if (!only_metrics) {
                 res["prompt"] = ptask->tokens.detokenize(ctx, true);
-                res["generated"] = generated_text;
+                res["generated"] = generated_text.empty()?debug_generated_text:generated_text;
```
ngxson:
```suggestion
                res["generated"] = generated_text.empty() ? debug_generated_text : generated_text;
```

ngxson (approved):

PR #19623, opened 2026-02-14
build : remove LLAMA_HTTPLIB option

angt:
This option was introduced as a workaround because cpp-httplib could not build on visionOS. Since it has been fixed and now compiles on all platforms, we can remove it and simplify many things.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,6 @@ option(LLAMA_TOOLS_INSTALL  "llama: install tools"        ${LLAMA_TOOLS_INSTALL_
 option(LLAMA_TESTS_INSTALL  "llama: install tests"        ON)
 
 # 3rd party libs
-option(LLAMA_HTTPLIB    "llama: httplib for downloading functionality" ON)
 option(LLAMA_OPENSSL    "llama: use openssl to support HTTPS" ON)
 option(LLAMA_LLGUIDANCE "llama-common: include LLGuidance library for structured output in common utils" OFF)
 
@@ -197,9 +196,7 @@ add_subdirectory(src)
 
 if (LLAMA_BUILD_COMMON)
     add_subdirectory(common)
-    if (LLAMA_HTTPLIB)
-        add_subdirectory(vendor/cpp-httplib)
-    endif()
+    add_subdirectory(vendor/cpp-httplib)
 endif()
 
 if (LLAMA_BUILD_COMMON AND LLAMA_BUILD_TESTS AND NOT CMAKE_JS_VERSION)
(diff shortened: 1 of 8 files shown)


danbev (approved):

ngxson (approved):

LostRuins:
FYI This breaks compilation on Win64devkit. httplib does not compile correctly there.

Example of errors:

```
undefined reference to `__imp_setsockopt'
undefined reference to `__imp_freeaddrinfo'
undefined reference to `__imp_WSACleanup'
undefined reference to `WSAPoll'
undefined reference to `__imp_shutdown'
```

I would strongly suggest not forcing httplib to be used on all builds. It is possible that there are other environments that it does not work either.

ngxson:
@LostRuins do you also build libcommon and llama-server?

It seems to me that httplib will always be built after this PR. We should only build httplib if any parts of the project depends on it.

CMakeLists.txt
```diff
@@ -112,7 +112,6 @@ option(LLAMA_TOOLS_INSTALL  "llama: install tools"        ${LLAMA_TOOLS_INSTALL_
 option(LLAMA_TESTS_INSTALL  "llama: install tests"        ON)
 
 # 3rd party libs
-option(LLAMA_HTTPLIB    "llama: httplib for downloading functionality" ON)
```
ngxson, line 115: probably better to add a new option: `LLAMA_COMMON_DOWNLOAD` to enable / disable download functionality, cc @angt

LostRuins:
@ngxson it's an unfortunate long dependency chain that requires httplib

I just want to build completion.exe -> so I build build completion.cpp -> which requires arg.cpp (for the launch params) -> which requires download.cpp (for definitions of `common_download_file_single` -> which now requires httplib.cpp (previously is ok because I can disable `LLAMA_USE_HTTPLIB`)

So, how can I build completion.exe without httplib?

angt:
Hi @LostRuins, thanks for reporting the issue.

This seems to be specific to the Win64devkit build configuration, as cpp-httplib compiles and works on Windows. 
I believe it is a very simple issue to fix, and might be related to this code:
```
if (WIN32 AND NOT MSVC)
    target_link_libraries(${TARGET} PRIVATE ws2_32)
endif()
```

angt:
Made a PR, let see if the CI is happy

CMakeLists.txt
angt: Yes, or maybe just not including `download` in `common` by default?

LostRuins:
I still think it would be better for the httplib library to be togglable. Some other platforms may not support it, or people might simply not want to link it - this forces bundling downloading and networking capabilities when they aren't really needed in many cases.

ngxson:
+1 for bringing back the ability to disable download functionalities at compile time.

@angt this is required because it doesn't make sense on certain use cases, for example wasm/emscripten or building it as for FFI binding. While it's technically compiles, it will become dead code

LostRuins:
i agree with ngxson

angt:
Yes, of course, but still we need to fix the bug.
Then we can isolate the download code in a better way to be used only when it's required.

PR #19631, opened 2026-02-14
Unload models based on VRAM usage

CooperCorona:
This is to track my progress on https://github.com/ggml-org/llama.cpp/discussions/19425. When finished, I'll create a pull request against llama.cpp directly.

llama.cpp currently has a flag to unload models after a fixed number has been reached. I'm trying to implement a strategy that unloads models when there's not enough VRAM to fit the new model. Otherwise, you can load as many models as you want.

### DIFF
--- a/include/llama.h
+++ b/include/llama.h
@@ -61,6 +61,7 @@ extern "C" {
     struct llama_model;
     struct llama_context;
     struct llama_sampler;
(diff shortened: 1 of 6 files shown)


CooperCorona:
whoops, meant to make this against my **own** fork until it is actually ready. Sorry for the churn.

PR #19283, opened 2026-02-03
Support Step3.5-Flash

forforever73:
This PR adds support for the Step3.5-Flash model architecture.

github:
- https://github.com/stepfun-ai/Step-3.5-Flash/tree/main

huggingface:
- mtp3_bf16: https://huggingface.co/stepfun-ai/Step-3.5-Flash
- mtp3_fp8: https://huggingface.co/stepfun-ai/Step-3.5-Flash-FP8
- int4(Q4_K_S.gguf): https://huggingface.co/stepfun-ai/Step-3.5-Flash-Int4

PR #19635, opened 2026-02-15
common : fix Step-3.5-Flash format detection and thinking support

jesseposner:
## Summary

Step-3.5-Flash (196B MoE) uses the same XML tool call output format as Qwen3-Coder and Nemotron 3 Nano (\`<tool_call><function=...><parameter=...>\`), but its template lacks the bare \`<function>\` and plural \`<parameters>\` markers in the tool enumeration section. The previous detection logic required all five XML markers, so Step-3.5-Flash fell through to Hermes 2 Pro, which doesn't call \`func_args_not_string()\`. Tool arguments stayed as JSON strings and templates using \`arguments|items\` crashed.

Reported by multiple users in #19283:
- [Leaked tool tokens with Codex](https://github.com/ggml-org/llama.cpp/pull/19283#issuecomment-3839920985) (@tarruda)
- [\`arguments|items\` crash with opencode](https://github.com/ggml-org/llama.cpp/pull/19283#issuecomment-3865972678) (@jacekpoplawski)
- [Root cause analysis: arguments is a string, not a dict](https://github.com/ggml-org/llama.cpp/pull/19283#issuecomment-3868260203) (@gelim)
- [Step-3.5-Flash authors: "Tool calling is still missing some support"](https://github.com/ggml-org/llama.cpp/pull/19283#issuecomment-3840185627) (@forforever73)

### Approach

Per @aldehir's [suggestion](https://github.com/ggml-org/llama.cpp/pull/19635#issuecomment-3905326359), Step-3.5-Flash is routed to the Nemotron v3 PEG parser rather than the simpler Qwen3-Coder GBNF grammar. The Nemotron v3 handler already provides \`thinking_forced_open\` handling, streaming support, and schema-aware parameter parsing.

The \`<think>\` tag in the template source distinguishes which parser to use: Step-3.5-Flash unconditionally emits \`<think>\`, Nemotron conditionally emits it (based on \`enable_thinking\`), and Qwen3-Coder has no \`<think>\` at all. Since both Step-3.5-Flash and Nemotron need thinking support, the presence of \`<think>\` routes to the PEG parser path.

### Changes
- Relax detection to require only the 3 output format markers (\`<tool_call>\`, \`<function=\`, \`<parameter=\`), dropping the tool enumeration markers (\`<function>\`, \`<parameters>\`) that Qwen3-Coder and Nemotron have but Step-3.5-Flash lacks
- Route models with \`<think>\` in template source to \`common_chat_params_init_nemotron_v3\`; models without \`<think>\` continue to \`common_chat_params_init_qwen3_coder_xml\`
- Add [unmodified HuggingFace chat template](https://huggingface.co/stepfun-ai/Step-3.5-Flash/blob/main/chat_template.jinja)

### Testing
- Format detection: verified \`COMMON_CHAT_FORMAT_PEG_CONSTRUCTED\` with \`thinking_forced_open=true\`
- 9 PEG parser tests covering: basic messages with/without thinking, tool calls with/without thinking, parallel tool calls, code string parameters, optional \`</parameter>\` closing tags, JSON schema response format
- All tests follow Nemotron v3 patterns, adapted for Step-3.5-Flash's unconditional \`<think>\` (all inputs include \`</think>\` delimiter since \`enable_thinking\` defaults to true)
- \`test-chat\` passes
- Validated against live server (IQ3_XXS on M4 Max) via \`/v1/chat/completions\`:
  - Tool calls: \`reasoning_content\` properly separated, \`tool_calls\` parsed with correct arguments, \`finish_reason: "tool_calls"\`
  - Streaming: tool call argument deltas arrive incrementally (PEG parser)
  - Code string params: multi-line Python code round-trips correctly through JSON
  - Plain messages: thinking/content separation works

### Related
- Builds on #19283 (initial Step-3.5-Flash support, merged)

### AI Disclosure
Claude was used for codebase exploration, pattern identification, and drafting. All changes follow established patterns from the existing Nemotron v3 format handler. Fully tested locally.

### DIFF
--- a/common/chat-parser-xml-toolcall.cpp
+++ b/common/chat-parser-xml-toolcall.cpp
@@ -279,7 +279,6 @@ void build_grammar_xml_tool_call(common_chat_params & data, const json & tools, 
             auto call_end = builder.add_rule("root-call-end", form.last_tool_end ? gbnf_format_literal(*form.last_tool_end) : gbnf_format_literal(form.tool_end));
             auto tool_call_multiple_with_end = builder.add_rule("root-tool-call-multiple-with-end", tool_call_once + " " + tool_call_more + "* " + call_end);
             builder.add_rule("root",
-                std::string(data.thinking_forced_open ? "( \"</think>\" space )? " : "") +
                 (form.scope_start.empty() ? "" : gbnf_format_literal(form.scope_start) + " ") +
                 tool_call_multiple_with_end  + "?" +
                 (form.scope_end.empty() ? "" : " " + gbnf_format_literal(form.scope_end))
--- a/common/chat-parser-xml-toolcall.cpp
+++ b/common/chat-parser-xml-toolcall.cpp
@@ -279,6 +279,7 @@ void build_grammar_xml_tool_call(common_chat_params & data, const json & tools, 
             auto call_end = builder.add_rule("root-call-end", form.last_tool_end ? gbnf_format_literal(*form.last_tool_end) : gbnf_format_literal(form.tool_end));
             auto tool_call_multiple_with_end = builder.add_rule("root-tool-call-multiple-with-end", tool_call_once + " " + tool_call_more + "* " + call_end);
             builder.add_rule("root",
+                std::string(data.thinking_forced_open ? "( \"</think>\" space )? " : "") +
                 (form.scope_start.empty() ? "" : gbnf_format_literal(form.scope_start) + " ") +
                 tool_call_multiple_with_end  + "?" +
                 (form.scope_end.empty() ? "" : " " + gbnf_format_literal(form.scope_end))
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -3141,16 +3141,15 @@ static common_chat_params common_chat_templates_apply_jinja(
 
     // Qwen3-Coder XML format detection (must come before Hermes 2 Pro)
     // Detect via XML markers: <tool_call>, <function=...>, and <parameter=...> blocks.
-    // Also matches Step-3.5-Flash which uses the same output format.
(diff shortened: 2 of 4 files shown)


tarruda:
Just tested this with codex and pi, seems to be working perfectly.

pwilkin (changes requested):
Please add tool calling and reasoning tests to the testcase.

aldehir:
I think you can just use the Nemotron parser here. It's the same format: qwen3-coder + thinking.

The difference lies in how it enumerates tools, which is why the original chat detection doesn't work.

tarruda:
While using this with pi CLI agent, I saw the following artifact in the UI:

image

aldehir (approved):
@tarruda if you have a chance, can you do a quick test to see if tool calling works as intended with the latest changes.

tarruda:
I did a quick test and couldn't see any special tokens in the frontend. One thing I noticed in the llama-server output was this message:

```
Template supports tool calls but does not natively describe tools. The fallback behaviour used may produce bad results, inspect prompt w/ --verbose & consider overriding the template.
```

aldehir:
I am more interested if it tool calls. That warning is problematic, might be a bug in the jinja code.

pwilkin:
That message is a known bug in the Jinja code, I've got it fixed on the autoparser branch (but @ngxson says he wants to fix it properly) - the thing is, if the template just dumps tools by using `tools | tojson`, the usage detection mechanism doesn't trigger for `tools[0].function.name`, which underlies this capability check.

It should not affect parsing though.

tarruda:
> I am more interested if it tool calls. That warning is problematic, might be a bug in the jinja code.

Tool calling seems fine on latest version

aldehir:
@pwilkin

pwilkin (approved):

PR #19642, opened 2026-02-15
ggml : avoid UB in gemm ukernel + tests

ggerganov:
cont #19422

- Reword the GEMM ukernel to not trip the compiler's aggressive loop optimization warnings. It's better to avoid the global pragma as it might be useful for other static analysis
- Add `test-backend-ops` with BS=75 to exercise the new tiled SIMD implementation


am17an (approved):
Thanks for fixing! AFAIK the compiler warning was only tripping up the AMX backend, not sure why. But this PR is anyways more readable

ggerganov:
Yes these warnings maybe depend on the gcc version.

PR #18958, opened 2026-01-20
CUDA: use mmvq for mul-mat-id for small batch sizes

am17an:
Currently for batch_sizes > 1, we immediately move to mmq which is suboptimal for small batch sizes. Bring performance of batched bench in line (previously there was a dip at n_tokens = 2)

Micro-benchmark for test-backend-ops

| Backend   | GGML op    | Op parameters                                                   |   TFLOPS master |   TFLOPS mmid-vec |   Speedup |
|:----------|:-----------|:----------------------------------------------------------------|----------------:|------------------:|----------:|
| CUDA0     | MUL_MAT_ID | type_a=q4_0,type_b=f32,n_mats=128,n_used=8,b=0,m=768,n=1,k=2048 |            4.61 |              4.62 |      1.00 |
| CUDA0     | MUL_MAT_ID | type_a=q4_0,type_b=f32,n_mats=128,n_used=8,b=0,m=768,n=4,k=2048 |            2.34 |              6.13 |      2.62 |
| CUDA0     | MUL_MAT_ID | type_a=q4_0,type_b=f32,n_mats=128,n_used=8,b=0,m=768,n=8,k=2048 |            4.27 |              6.83 |      1.60 |
| CUDA0     | MUL_MAT_ID | type_a=q4_0,type_b=f32,n_mats=32,n_used=4,b=0,m=1792,n=1,k=2048 |            5.49 |              5.49 |      1.00 |
| CUDA0     | MUL_MAT_ID | type_a=q4_0,type_b=f32,n_mats=32,n_used=4,b=0,m=1792,n=4,k=2048 |            3.37 |              6.37 |      1.89 |
| CUDA0     | MUL_MAT_ID | type_a=q4_0,type_b=f32,n_mats=32,n_used=4,b=0,m=1792,n=8,k=2048 |            6.57 |              7.23 |      1.10 |

Issue #18864, opened 2026-01-15
Feature Request: generalize MMQ CUDA kernel for floating-point data


JohannesGaessler:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

Generalize the kernel in `mmq.cuh`:

* Optionally use FP16 activations instead of q8_1.
* Implement function for loading FP16 data from `src0`.
* Implement function for FP16, FP16 -> FP32 matrix multiplication.
* Move and rename some parts of the code to different headers to organize it better, this should be done in a follow-up PR.

Same implementation can be done for BF16 and FP32, for FP8 and NVFP4 we should first standardize how to treat them in ggml and add CPU support.

This feature should be tackled by experienced CUDA programmers only.

### Motivation

I wrote a matrix multiplication CUDA kernel for quantized data that currently resides `mmq.cuh`. I wrote it because it allows for a fused matrix multiplication using custom data types since a library like cuBLAS only supports standard data types like FP16. I don't think it would be worthwhile to try and compete with cuBLAS for such data types when it comes to conventional GEMM. However, I at some point added optimizations for MoE models which I think will result in better performance than cuBLAS.

bssrdf:
There could be potential overflow risk (actually happening for some [Qwen Image gguf models](https://github.com/leejet/stable-diffusion.cpp/pull/851#discussion_r2900951475)  when quantizing activations into q8_1_mmq due to using `half2` for `ds4`. Should it be addressed?

JohannesGaessler:
I don't remember whether or not those values are actually used as FP16 or if they are always upcast to FP32 anyways. If it's the latter I would say to just use BF16.

bssrdf:
> I don't remember whether or not those values are actually used as FP16 or if they are always upcast to FP32 anyways. If it's the latter I would say to just use BF16.

Thanks, @JohannesGaessler.
The values could have already been overflowed here
https://github.com/ggml-org/llama.cpp/blob/451ef08432d1f7d3d6071d4006cbbeda21dcfbec/ggml/src/ggml-cuda/quantize.cu#L267
[...]

PR #19645, opened 2026-02-15
cuda : enable CUDA graphs for MMID 1 <= BS <= 4

ggerganov:
cont #19644 
cont #18958
cont #19521 

Enable CUDA graphs for ggml graphs with `GGML_OP_MUL_MAT_ID` at `1 < BS <= 4`. Improves the performance for parallel generation of up to 4 sequences. This is useful for example when running multiple local agents in parallel (https://github.com/ggml-org/llama.cpp/discussions/19564#discussioncomment-15808835)

Also, simplify the CUDA graph exception logic. Unless I am missing something, we no longer need to restrict `GGML_OP_ADD` for `BS > 1` because we perform an exhaustive node properties check to decide if the graph needs an update.

Next PRs:

- [ ] Try to extend support for `BS > 4 && BS <= 8`, or more?

DGX Spark:

| Model                  |   Microbatch size | Test   |   t/s pr/19644 |   t/s pr |   Speedup |
|:-----------------------|------------------:|:-------|------------------------------------:|-----------------------------------:|----------:|
| deepseek2 30B.A3B Q8_0 |                 1 | pp512  |                               52.58 |                              52.23 |      0.99 |
| deepseek2 30B.A3B Q8_0 |                 2 | pp512  |                               68.96 |                              78.45 |      1.14 |
| deepseek2 30B.A3B Q8_0 |                 3 | pp512  |                               83.68 |                              92.96 |      1.11 |
| deepseek2 30B.A3B Q8_0 |                 4 | pp512  |                               95.28 |                             104.23 |      1.09 |
| gpt-oss 120B MXFP4 MoE |                 1 | pp512  |                               70.73 |                              70.47 |      1.00 |
| gpt-oss 120B MXFP4 MoE |                 2 | pp512  |                               86.85 |                              96.82 |      1.11 |
| gpt-oss 120B MXFP4 MoE |                 3 | pp512  |                              104.29 |                             114.42 |      1.10 |
| gpt-oss 120B MXFP4 MoE |                 4 | pp512  |                              115.61 |                             124.86 |      1.08 |
| gpt-oss 20B MXFP4 MoE  |                 1 | pp512  |                              106.60 |                             106.66 |      1.00 |
| gpt-oss 20B MXFP4 MoE  |                 2 | pp512  |                              136.89 |                             153.91 |      1.12 |
| gpt-oss 20B MXFP4 MoE  |                 3 | pp512  |                              165.17 |                             186.38 |      1.13 |
| gpt-oss 20B MXFP4 MoE  |                 4 | pp512  |                              191.08 |                             207.84 |      1.09 |
| qwen3moe 30B.A3B Q8_0  |                 1 | pp512  |                               67.24 |                              66.92 |      1.00 |
| qwen3moe 30B.A3B Q8_0  |                 2 | pp512  |                               82.34 |                              91.74 |      1.11 |
| qwen3moe 30B.A3B Q8_0  |                 3 | pp512  |                               98.97 |                             108.10 |      1.09 |
| qwen3moe 30B.A3B Q8_0  |                 4 | pp512  |                              111.61 |                             120.74 |      1.08 |
| qwen3next 80B.A3B Q4_0 |                 1 | pp512  |                               65.15 |                              64.85 |      1.00 |
| qwen3next 80B.A3B Q4_0 |                 2 | pp512  |                               57.28 |                              75.91 |      1.33 |
| qwen3next 80B.A3B Q4_0 |                 3 | pp512  |                               78.10 |                             102.16 |      1.31 |
| qwen3next 80B.A3B Q4_0 |                 4 | pp512  |                               95.98 |                             121.15 |      1.26 |
| qwen3next 80B.A3B Q8_0 |                 1 | pp512  |                               43.10 |                              43.23 |      1.00 |
| qwen3next 80B.A3B Q8_0 |                 2 | pp512  |                               43.33 |                              53.78 |      1.24 |
| qwen3next 80B.A3B Q8_0 |                 3 | pp512  |                               56.90 |                              69.23 |      1.22 |
| qwen3next 80B.A3B Q8_0 |                 4 | pp512  |                               68.84 |                              81.43 |      1.18 |

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2278,6 +2278,7 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor * 
 
     const int cc = ggml_cuda_info().devices[ggml_cuda_get_device()].cc;
 
+    // [TAG_MUL_MAT_ID_CUDA_GRAPHS]
     if (src1->type == GGML_TYPE_F32 && dst->type == GGML_TYPE_F32) {
         static_assert(MMVQ_MAX_BATCH_SIZE == MMVF_MAX_BATCH_SIZE);
         if (ne2 <= MMVQ_MAX_BATCH_SIZE) {
@@ -2304,6 +2305,8 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor * 
             return;
         }
     }
+
+    // note: this path should not be reached when recording CUDA graphs, because it requires stream synchronization
 
     cudaStream_t stream = ctx.stream();
 
@@ -2865,15 +2868,6 @@ static bool ggml_cuda_graph_check_compability(ggml_cgraph * cgraph) {
     bool use_cuda_graph = true;
     // Loop over nodes in GGML graph to obtain info needed for CUDA graph
 
-    const std::string gemma3n_per_layer_proj_src0_name = "inp_per_layer_selected";
-    const std::string gemma3n_per_layer_proj_src1_name = "per_layer_proj";
-    const std::string ffn_moe_gate_bias_prefix = "ffn_moe_gate_biased";
-    const std::string ffn_moe_up_bias_prefix = "ffn_moe_up_biased";
-    const std::string ffn_moe_down_bias_prefix = "ffn_moe_down_biased";
-    const std::string nemotron_h_block_out_prefix = "nemotron_h_block_out";
-    const std::string mamba2_y_add_d_prefix = "mamba2_y_add_d";
-    const std::string delta_net_prefix = "dnet_add";
-
     for (int i = 0; i < cgraph->n_nodes; i++) {
         ggml_tensor * node = cgraph->nodes[i];
 
@@ -2888,31 +2882,14 @@ static bool ggml_cuda_graph_check_compability(ggml_cgraph * cgraph) {
 #endif
         }
 
-        if (node->op == GGML_OP_MUL_MAT_ID && node->ne[2] != 1) {
-            use_cuda_graph = false; // This node type is not supported by CUDA graph capture
-#ifndef NDEBUG
-            GGML_LOG_DEBUG("%s: disabling CUDA graphs due to unsupported node type\n", __func__);
-#endif
-        }
-
-        if (node->op == GGML_OP_ADD &&
-            node->src[1] && node->src[1]->ne[1] > 1 &&
-            (node->src[0] ? node->src[0]->name != gemma3n_per_layer_proj_src0_name : true) &&
-            (node->src[1] ? node->src[1]->name != gemma3n_per_layer_proj_src1_name : true) &&
-            strncmp(node->name, ffn_moe_gate_bias_prefix.c_str(), ffn_moe_gate_bias_prefix.size()) != 0 &&
-            strncmp(node->name, ffn_moe_up_bias_prefix.c_str(), ffn_moe_up_bias_prefix.size()) != 0 &&
-            strncmp(node->name, ffn_moe_down_bias_prefix.c_str(), ffn_moe_down_bias_prefix.size()) != 0 &&
-            strncmp(node->name, nemotron_h_block_out_prefix.c_str(), nemotron_h_block_out_prefix.size()) != 0 &&
-            strncmp(node->name, mamba2_y_add_d_prefix.c_str(), mamba2_y_add_d_prefix.size()) != 0 &&
-            strncmp(node->name, delta_net_prefix.c_str(), delta_net_prefix.size()) != 0) {
-            // disable CUDA graphs for batch size > 1 for now while excluding the matrix-matrix addition as part of Gemma3n's `project_per_layer_input` operation
-            // by means of matching node names. See
-            // https://github.com/ggml-org/llama.cpp/blob/f9a31eea06a859e34cecb88b4d020c7f03d86cc4/src/llama-model.cpp#L10199-L10241 and
-            // https://github.com/huggingface/transformers/blob/bda75b4011239d065de84aa3e744b67ebfa7b245/src/transformers/models/gemma3n/modeling_gemma3n.py#L1773,
-            // Generally, changes in batch size or context size can cause changes to the grid size of some kernels.
+        // [TAG_MUL_MAT_ID_CUDA_GRAPHS]
+        if (node->op == GGML_OP_MUL_MAT_ID && (!ggml_is_quantized(node->src[0]->type) || node->ne[2] > 4)) {
+            // under these conditions, the mul_mat_id operation will need to synchronize the stream, so we cannot use CUDA graphs
+            // TODO: figure out a way to enable for larger batch sizes, without hurting performance
+            // ref: https://github.com/ggml-org/llama.cpp/pull/18958
             use_cuda_graph = false;
 #ifndef NDEBUG
-            GGML_LOG_DEBUG("%s: disabling CUDA graphs due to batch size > 1 [%s] [%ld %ld %ld %ld]\n", __func__, node->name, node->ne[0], node->ne[1], node->ne[2], node->ne[3]);
+            GGML_LOG_DEBUG("%s: disabling CUDA graphs due to unsupported node type\n", __func__);
 #endif
         }
 
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2283,7 +2283,7 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor * 
         static_assert(MMVQ_MAX_BATCH_SIZE == MMVF_MAX_BATCH_SIZE);
         if (ne2 <= MMVQ_MAX_BATCH_SIZE) {
             if (ggml_is_quantized(src0->type)) {
-                if (ne2 <= 4) {
+                if (ne2 <= MMVQ_MMID_MAX_BATCH_SIZE) {
                     ggml_cuda_mul_mat_vec_q(ctx, src0, src1, ids, dst);
                     return;
                 }
@@ -2307,9 +2307,7 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor * 
     }
 
     // note: this path should not be reached when recording CUDA graphs, because it requires stream synchronization
-    cudaStreamCaptureStatus capture_status;
-    CUDA_CHECK(cudaStreamIsCapturing(stream, &capture_status));
-    GGML_ASSERT(capture_status == cudaStreamCaptureStatusNone);
+    // TODO: add asserts to verify this. should work with CUDA, HIP, etc.
     cudaStream_t stream = ctx.stream();
 
     GGML_ASSERT(nb12 % nb11 == 0);
@@ -2885,7 +2883,7 @@ static bool ggml_cuda_graph_check_compability(ggml_cgraph * cgraph) {
         }
 
         // [TAG_MUL_MAT_ID_CUDA_GRAPHS]
-        if (node->op == GGML_OP_MUL_MAT_ID && (!ggml_is_quantized(node->src[0]->type) || node->ne[2] > 4)) {
+        if (node->op == GGML_OP_MUL_MAT_ID && (!ggml_is_quantized(node->src[0]->type) || node->ne[2] > MMVQ_MMID_MAX_BATCH_SIZE)) {
             // under these conditions, the mul_mat_id operation will need to synchronize the stream, so we cannot use CUDA graphs
             // TODO: figure out a way to enable for larger batch sizes, without hurting performance
             // ref: https://github.com/ggml-org/llama.cpp/pull/18958
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2307,7 +2307,9 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor * 
     }
 
     // note: this path should not be reached when recording CUDA graphs, because it requires stream synchronization
-
+    cudaStreamCaptureStatus capture_status;
+    CUDA_CHECK(cudaStreamIsCapturing(stream, &capture_status));
+    GGML_ASSERT(capture_status == cudaStreamCaptureStatusNone);
     cudaStream_t stream = ctx.stream();
 
     GGML_ASSERT(nb12 % nb11 == 0);
--- a/ggml/src/ggml-cuda/mmvq.cuh
+++ b/ggml/src/ggml-cuda/mmvq.cuh
@@ -1,6 +1,7 @@
 #include "common.cuh"
 
 #define MMVQ_MAX_BATCH_SIZE 8 // Max. batch size for which to use MMVQ kernels.
+#define MMVQ_MMID_MAX_BATCH_SIZE 4 // Max. batch size for which to use MMVQ kernels for MUL_MAT_ID
 
 void ggml_cuda_mul_mat_vec_q(ggml_backend_cuda_context & ctx,
     const ggml_tensor * src0, const ggml_tensor * src1, const ggml_tensor * ids, ggml_tensor * dst, const ggml_cuda_mm_fusion_args_host * fusion = nullptr);


ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2888,31 +2882,14 @@ static bool ggml_cuda_graph_check_compability(ggml_cgraph * cgraph) {
...
-            strncmp(node->name, nemotron_h_block_out_prefix.c_str(), nemotron_h_block_out_prefix.size()) != 0 &&
-            strncmp(node->name, mamba2_y_add_d_prefix.c_str(), mamba2_y_add_d_prefix.size()) != 0 &&
-            strncmp(node->name, delta_net_prefix.c_str(), delta_net_prefix.size()) != 0) {
-            // disable CUDA graphs for batch size > 1 for now while excluding the matrix-matrix addition as part of Gemma3n's `project_per_layer_input` operation
-            // by means of matching node names. See
-            // https://github.com/ggml-org/llama.cpp/blob/f9a31eea06a859e34cecb88b4d020c7f03d86cc4/src/llama-model.cpp#L10199-L10241 and
-            // https://github.com/huggingface/transformers/blob/bda75b4011239d065de84aa3e744b67ebfa7b245/src/transformers/models/gemma3n/modeling_gemma3n.py#L1773,
-            // Generally, changes in batch size or context size can cause changes to the grid size of some kernels.
+        // [TAG_MUL_MAT_ID_CUDA_GRAPHS]
+        if (node->op == GGML_OP_MUL_MAT_ID && (!ggml_is_quantized(node->src[0]->type) || node->ne[2] > 4)) {
```
am17an: Perhaps we rename this 4 to `MMVQ_MMID_MAX_BATCH_SIZE` in case we end up optimizing for BS > 4

am17an (commented):
Speedup on 4090:

| Model                 |   Microbatch size | Test   |   t/s 8bc255a3a |   t/s gg/cuda-graphs-enable-bs-gt1 |   Speedup |
|:----------------------|------------------:|:-------|----------------:|-----------------------------------:|----------:|
| gpt-oss 20B MXFP4 MoE |                 1 | pp512  |          283.83 |                             284.09 |      1.00 |
| gpt-oss 20B MXFP4 MoE |                 2 | pp512  |          401.81 |                             443.68 |      1.10 |
| gpt-oss 20B MXFP4 MoE |                 3 | pp512  |          517.59 |                             561.15 |      1.08 |
| gpt-oss 20B MXFP4 MoE |                 4 | pp512  |          595.86 |                             635.65 |      1.07 |

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2305,6 +2306,8 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor *
         }
     }
 
+    // note: this path should not be reached when recording CUDA graphs, because it requires stream synchronization
+
```
ORippler:
```suggestion
    // note: this path should not be reached when recording CUDA graphs, because it requires stream synchronization
    cudaStreamCaptureStatus capture_status;
    CUDA_CHECK(cudaStreamIsCapturing(stream, &capture_status));
    GGML_ASSERT(capture_status == cudaStreamCaptureStatusNone);
```

Please add a check if this path should not be reached during capture

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2888,31 +2882,14 @@ static bool ggml_cuda_graph_check_compability(ggml_cgraph * cgraph) {
...
-            // disable CUDA graphs for batch size > 1 for now while excluding the matrix-matrix addition as part of Gemma3n's `project_per_layer_input` operation
-            // by means of matching node names. See
-            // https://github.com/ggml-org/llama.cpp/blob/f9a31eea06a859e34cecb88b4d020c7f03d86cc4/src/llama-model.cpp#L10199-L10241 and
-            // https://github.com/huggingface/transformers/blob/bda75b4011239d065de84aa3e744b67ebfa7b245/src/transformers/models/gemma3n/modeling_gemma3n.py#L1773,
-            // Generally, changes in batch size or context size can cause changes to the grid size of some kernels.
+        // [TAG_MUL_MAT_ID_CUDA_GRAPHS]
+        if (node->op == GGML_OP_MUL_MAT_ID && (!ggml_is_quantized(node->src[0]->type) || node->ne[2] > 4)) {
+            // under these conditions, the mul_mat_id operation will need to synchronize the stream, so we cannot use CUDA graphs
+            // TODO: figure out a way to enable for larger batch sizes, without hurting performance
+            // ref: https://github.com/ggml-org/llama.cpp/pull/18958
```
ORippler, line 2889: I see two ways for this:
1. Refactor the logic expressed on host side into a CUDA kernel
2. Use `cudaLaunchHostFunc` and manage lifetime of CPU objects inside `ggml_cuda_graph`

ORippler:
> Also, simplify the CUDA graph exception logic. Unless I am missing something, we no longer need to restrict GGML_OP_ADD for BS > 1 because we perform an exhaustive node properties check to decide if the graph needs an update.

That's correct.

ggml/src/ggml-cuda/ggml-cuda.cu
ggerganov: Btw, note here that we currently require quantized `src0`. This means that with BF16 MoE models for example, the CUDA graphs will not be enabled.

am17an: ~#18958 enables this for mmvf also, so bf16 should also work~

am17an: Ah sorry that is that just for AMD

JohannesGaessler (approved):

ggml/src/ggml-cuda/ggml-cuda.cu
ggerganov: @JohannesGaessler Can you confirm that adding the `MMVQ_MMID_MAX_BATCH_SIZE` constant is OK for now?

ggml/src/ggml-cuda/ggml-cuda.cu
ggerganov: This is more difficult to do to be compatible with HIP, so leaving just a TODO for now.

ggml/src/ggml-cuda/ggml-cuda.cu
JohannesGaessler: The fundamental problem here is that we are using cuBLAS for GEMM and are orchestrating the matrix multiplications per expert via CPU logic. It would in principle be possible to pad the matrix multiplications and use batched/strided GEMM but that will waste a lot of compute. I'm not sure whether it's possible to orchestrate cuBLAS GEMM from device code, launching CUDA kernels from within `cudaLaunchHostFunc` also seems dubious. For the custom ggml kernels I made it so that during the quantization kernel the data is re-arranged to be contiguous per expert and the kernel is skipping any output tiles that would be 100% padding. In principle this could be generalized to floating-point data, see https://github.com/ggml-org/llama.cpp/issues/18864 .

ggml/src/ggml-cuda/ggml-cuda.cu
JohannesGaessler: Adding a named constant here would be fine with me.

ggerganov:
Merging on green

ggml/src/ggml-cuda/ggml-cuda.cu
ORippler: > This is more difficult to do to be compatible with HIP, so leaving just a TODO for now.
For transpilation, one is often fine with defining aliases in vendors/*.h, as HIP/MUSA tend to closely follow CUDA's API (such as to enable transpilation):

https://rocm.docs.amd.com/projects/HIP/en/latest/reference/hip_runtime_api/modules/graph_management.html#_CPPv420hipStreamIsCapturing11hipStream_tP22hipStreamCaptureStatus

aliasing cudaStreamIsCapturing to hipStreamIsCapturing, hipStreamCaptureStatus to cudaStreamCaptureStatus and hipStreamCaptureStatusNone to cudaStreamCaptureStatusNone should do the trick. Not sure about MUSA, as I have not found docs for it anywhere yet

LostRuins:
Hello @ggerganov, this PR specifically seems to cause a new memory leak faced by @henk717 when used with stable-diffusion.cpp. I know it sounds strange but bear with me.

- Memory leak causes the allocation of 1-2GB of additional RAM every time an image is generated, eventually resulting in OOM after many images. 
- The memory leak increase per-step of the diffusion, so it's something to do with inference time allocations.
- All image generation models are affected, but text generation is unaffected
- The leak **ONLY happens on the CUDA backend**. Vulkan and CPU do not observe such a leak
- The leak **only seems to happen when CUDA graphs is enabled** ie. `GGML_CUDA_USE_GRAPHS` otherwise there is no leak.
- The leak seems to happen on a 3090 (@henk717 has a dual RTX 3090 setup) but not on my own RTX4090 (is fine)
- Reverting **this** PR causes the memory leak to disappear. This was determined by bisecting until this commit was found.

I can create a full issue too if you want but I figured since this is a very small commit maybe you might already know the fix.

Also tagging stable-diffusion.cpp stakeholders although I don't think it's caused there since their code is unchanged - @leejet @stduhpf

ORippler:
> Hello @ggerganov, this PR specifically seems to cause a new memory leak faced by @henk717 when used with stable-diffusion.cpp. I know it sounds strange but bear with me.

> Memory leak causes the allocation of 1-2GB of additional RAM every time an image is generated, eventually resulting in OOM after many images.
The memory leak increase per-step of the diffusion, so it's something to do with inference time allocations.
All image generation models are affected, but text generation is unaffected
The leak ONLY happens on the CUDA backend. Vulkan and CPU do not observe such a leak
The leak only seems to happen when CUDA graphs is enabled ie. GGML_CUDA_USE_GRAPHS otherwise there is no leak.
The leak seems to happen on a 3090 (@henk717 has a dual RTX 3090 setup) but not on my own RTX4090 (is fine)
Reverting this PR causes the memory leak to disappear. This was determined by bisecting until this commit was found.
I can create a full issue too if you want but I figured since this is a very small commit maybe you might already know the fix.

> Also tagging stable-diffusion.cpp stakeholders although I don't think it's caused there since their code is unchanged - @leejet @stduhpf

Seems related to https://github.com/ggml-org/llama.cpp/discussions/19708. Does stable-diffusion.cpp re-create a new ggml-cgraph object with different node-key for every step of the pipeline?

Also, does #19754 fix your issue?

henk717:
Small clarification, the amount of memory increase will depend on the image model. It can be as much as 4GB per image on Qwen Image Edit. Its very noticable when it happens. I monitor this in the regular ram, not the vram although that seemed to go up as well. While I run dual 3090's to the application only one of the 3090's is visible.

LostRuins:
Sent you a test build with https://github.com/ggml-org/llama.cpp/pull/19754 merged and this PR reinstated, let me know if it fixes the memory leak

henk717:
This did not resolve the issue.

LostRuins:
maybe gg knows whats wrong.

ggerganov:
Could you provide the most basic steps to repro?

LostRuins:
> Could you provide the most basic steps to repro?

Umm sorry, I'd like to if possible, unfortunately... I don't know how to reproduce it easily here in this repo... as most of the necessary code is not inside the llama.cpp repo. 

Henk717 was using it from within KoboldCpp. The issue only happens when using image generation code taken from https://github.com/leejet/stable-diffusion.cpp, however **sd.cpp itself** has yet to sync to the new ggml version (which need this commit), so it won't be reproducible there yet, additionally they also do not use CUDA graphs enabled there which is needed to trigger this issue, so it will also need to be enabled.

I could direct you to my own fork at https://github.com/LostRuins/koboldcpp however I must admit my code quality is not very good, my build process is jank, and might be troublesome to work with, plus I cannot trigger this bug myself... 

@henk717 do you have any better idea how we can help gg troubleshoot this here? 

The circumstances of the leak caused by this PR are as summarized in https://github.com/ggml-org/llama.cpp/pull/19645#issuecomment-3935496049

henk717:
It happens during regular KoboldCpp usage on every image I generate. System memory usage is extremely noticeable as a measurement as it will rapidly increase on an effected build. For me its consistent on Windows with driver version 581.29 on my local desktop PC with a 3090 GPU. On a 3090 linux cloud rental I was unable to measure this issue (I assume it wasn't present but I am less familiar with the linux tools). Lostruins was uneffected on his 4090 laptop.

The tricky part is that it manifests inside of stablediffusion.cpp because KoboldCpp uses the newer ggml from Llamacpp. At the moment the upstream sdcpp project is still on a 3 week old ggml version and will not be impacted yet.

One way to test it on KoboldCpp is by using https://huggingface.co/MaxedOut/ComfyUI-Starter-Packs/resolve/main/SDXL/checkpoints/sd_xl_base_1.0.safetensors?download=true (Any model should produce it, this is a relatively fast one that I know for certain is easy to observe) as the image model. This can be done in the Image Model tab by selecting it as an Image Gen Model. No further files are required for this model. The commandline equivalent would be --imgmodel 

Once the UI launches accept that you wish to be taken to sdui or use the /sdui link for easy image generation. Type any prompt (In my tests I use the word Kobold) and you will now see the system process's memory increase by several hundred megabytes every image you generate.

If you prefer working with the upstream sdcpp project you may be able to link it to a newer ggml version to introduce the same issue.

KoboldCpp could be compiled with cmake for the windows cuda build (Linux is make or koboldcpp.sh, but on Linux I haven't been able to reproduce the issue so far possibly due to my lack of familiarity with the linux memory management tools).  

@LostRuins would it be difficult for you to provide a stablediffusioncpp build (not koboldcpp) with the new ggml applied?

am17an:
Maybe add this check back again and see if it resolves the issue 

```cpp
index 7e6d33035..f841d8ac9 100644
@@ -2893,6 +2893,10 @@ static bool ggml_cuda_graph_check_compability(ggml_cgraph * cgraph) {
 #endif
         }
 
+        if (node->op == GGML_OP_ADD && node->src[1] && node->src[1]->ne[1] > 1) {
+            use_cuda_graph = false;
+        }
+
         if (!use_cuda_graph) {
             break;
         }
```

ggerganov:
Yes, let's start with @am17an's suggestion first to try to narrow down the possible root causes. @henk717 Let us know when you give it a try.

henk717:
Bullseye, @LostRuins made a test for me with the fix applied and on this test build I do not have the issue.
Update: As a sanity check I opened the older test build where it happened, it still happens there. Went back to this test and it doesn't happen. So I also confirmed I can still reproduce the issue without the fix, eliminates the potential placebo factor.

am17an:
So I think we could be adding a large of number of CUDA graphs which are never reclaimed. Namely this code is being called several times. It's probably because each time graph_compute is being called with a new pointer during nrows > 1. 

https://github.com/ggml-org/llama.cpp/blob/9051663d5d67bdab765e9bc9afee41ed34f09260/ggml/src/ggml-cuda/common.cuh#L1334-L1341

You can confirm this by using `valgrind --tool=massif ./binary` and then `ms_print massif.out.<pid>`.

henk717:
Its happening on my Windows system, so if that instruction is for me i'd need to know a windows equivalent.

am17an:
IDK on windows, from a quick search it looks like drmemory is useful.

henk717:
This is not going to be an easy task for me, the tool you suggested didn't work. I found another one which reduces the generation speeds immensely so it takes a long time, the end result was without any names so I don't think any of us will be able to decypher it.

Between @lostruins integrating fixes and me testing this would probably take significantly more time compared to queuing up some idea's. Normally when I am testing he queue's up multiple test builds so I can test them all in one go this will likely be faster than the hours it will take for a single test, assuming we get more useful output to begin with as I don't know if the tool I found can handle debug symbols (The exe I have isn't a debug symbol build but one from our ci).

My suggestion is to continue on the more speculative path, ideally in one go. Then he can queue it up for me in his timezone and I can test it in mine if that is needed.

am17an:
Okay, you can try this patch to confirm 
```cpp
diff --git a/ggml/src/ggml-cuda/common.cuh b/ggml/src/ggml-cuda/common.cuh
index 36d8a3aaa..d394fb475 100644
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1335,6 +1335,9 @@ struct ggml_backend_cuda_context {
         auto it = cuda_graphs.find(first_node_ptr);
         if (it == cuda_graphs.end()) {
             cuda_graphs[first_node_ptr] = std::make_unique<ggml_cuda_graph>();
+            if (cuda_graphs.size() % 100 == 0) {
+                GGML_LOG_INFO("cuda graphs current size: %zu\n", cuda_graphs.size());
+            }
             return cuda_graphs[first_node_ptr].get();
         }
         return it->second.get();
```

This should keep on increasing forever till OOM and you should see the log of the size of the map.

henk717:
The patch was added : https://github.com/LostRuins/koboldcpp/commit/c2e06d8324e12d9b2621c62c8538fd90bf3ae79f
On this build I could reproduce the leak but I did not see what i'd expect as this particular log does not appear for me.
I think the if statements aren't being hit.

Instead it looks like this:
generating image: 1/1 - seed 4394834
  |==>                                               | 1/20 - 2.26it/s
ggml_backend_cuda_graph_compute: CUDA graph warmup complete
  |=====>                                            | 2/20 - 3.20it/s
ggml_backend_cuda_graph_compute: CUDA graph warmup complete
  |=======>                                          | 3/20 - 3.83it/s
ggml_backend_cuda_graph_compute: CUDA graph warmup complete

ggml_backend_cuda_graph_compute: CUDA graph warmup complete
  |=================>                                | 7/20 - 5.10it/s
ggml_backend_cuda_graph_compute: CUDA graph warmup complete
  |======================>                           | 9/20 - 5.39it/s
ggml_backend_cuda_graph_compute: CUDA graph warmup complete
  |==========================================>       | 17/20 - 6.14it/s
ggml_backend_cuda_graph_compute: CUDA graph warmup complete
  |===============================================>  | 19/20 - 6.20it/s
ggml_backend_cuda_graph_compute: CUDA graph warmup complete
  |==================================================| 20/20 - 6.21it/s

henk717:
Update: I noticed it was kinda limited on sdxl so I went back to testing qwen, a few images in I saw the new debug output.

LostRuins:
But what did it output (the values)

henk717:
The check was an equals check, so I only got it when it hit 100.

am17an:
It's a modulo check. So you should get the print at 100, 200, 300 etc.

henk717:
By the time it hit 100 I already had multiple gigabytes of additional ram.

PR #19651, opened 2026-02-16
convert_hf_to_gguf: add JoyAI-LLM-Flash tokenizer hash mapping to deepseek-v3

dranger003:
- adding hash for `jdopensource/JoyAI-LLM-Flash` mapping to existing `deepseek-v3`
- `DeepseekV3ForCausalLM` architecture already supported
- moved `GLM-4.7-Flash` entry together with the other `glm` entries
- conversion and quantization successful and output generation looks great

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -1049,6 +1049,9 @@ class TextModel(ModelBase):
         if chkhsh == "9ca2dd618e8afaf09731a7cf6e2105b373ba6a1821559f258b272fe83e6eb902":
             # ref: https://huggingface.co/zai-org/GLM-4.5-Air
             res = "glm4"
+        if chkhsh == "cdf5f35325780597efd76153d4d1c16778f766173908894c04afc20108536267":
+            # ref: https://huggingface.co/zai-org/GLM-4.7-Flash
+            res = "glm4"
         if chkhsh == "1431a23e583c97432bc230bff598d103ddb5a1f89960c8f1d1051aaa944d0b35":
             # ref: https://huggingface.co/sapienzanlp/Minerva-7B-base-v1.0
             res = "minerva-7b"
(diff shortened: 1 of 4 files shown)


CISC:
Thanks, but `convert_hf_to_gguf_update.py` needs to be run to update `convert_hf_to_gguf.py`, which is why CI failed. :)

Anyway, I think I prefer that you create a new pre-tokenizer name for this in case there are any subtle differences discovered later on; move it out of `pre_computed_hashes` and add it in `llama-vocab.cpp` together with `DEEPSEEK3_LLM` and `HUNYAN_DENSE`.

dranger003:
Ah, of course I changed the order after running the script, thanks.

convert_hf_to_gguf.py
```diff
@@ -1265,6 +1265,9 @@ def get_vocab_base_pre(self, tokenizer) -> str:
         if chkhsh == "d30d75d9059f1aa2c19359de71047b3ae408c70875e8a3ccf8c5fba56c9d8af4":
             # ref: https://huggingface.co/Qwen/Qwen3.5-9B-Instruct
             res = "qwen35"
+        if chkhsh == "b4b8ca1f9769494fbd956ebc4c249de6131fb277a4a3345a7a92c7dd7a55808d":
+            # ref: https://huggingface.co/jdopensource/JoyAI-LLM-Flash
+            res = "joyai_llm"
```
```suggestion
            res = "joyai-llm"
```

convert_hf_to_gguf_update.py
```diff
@@ -148,7 +148,8 @@ class TOKENIZER_TYPE(IntEnum):
     {"name": "youtu",            "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/tencent/Youtu-LLM-2B", },
     {"name": "solar-open",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/upstage/Solar-Open-100B", },
     {"name": "exaone-moe",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B", },
-    {"name": "qwen35",           "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/Qwen/Qwen3.5-9B-Instruct", }
+    {"name": "qwen35",           "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/Qwen/Qwen3.5-9B-Instruct", },
+    {"name": "joyai_llm",        "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/jdopensource/JoyAI-LLM-Flash", },
```
CISC:
```suggestion
    {"name": "joyai-llm",        "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/jdopensource/JoyAI-LLM-Flash", },
```

CISC (commented):
You also need to add a section like this:
https://github.com/ggml-org/llama.cpp/blob/7af1cce091223b49a42f67275617aef97e7cff77/src/llama-vocab.cpp#L2039-L2042

src/llama-vocab.h
```diff
@@ -55,6 +55,7 @@ enum llama_vocab_pre_type {
     LLAMA_VOCAB_PRE_TYPE_YOUTU           = 44,
     LLAMA_VOCAB_PRE_TYPE_EXAONE_MOE      = 45,
     LLAMA_VOCAB_PRE_TYPE_QWEN35          = 46,
+    LLAMA_VOCAB_PRE_TYPE_JOYAI_LLM       = 47,
```
CISC: Just a heads up, this is conflicting with another pending PR and may need to be rebased and updated before merge.

CISC (approved):

PR #19655, opened 2026-02-16
server : implement verbose-prompt option; fix #19653

SlavikCA:


SlavikCA:
Similar to https://github.com/ggml-org/llama.cpp/pull/19576
But this one is for server.

danbev (approved):

ngxson:
I don't think it's a good idea to expose this option to server. There are security concerns. We already had an env var for debugging slot content, you should use that instead

PR #19666, opened 2026-02-16
build : link ws2_32 as PUBLIC on Windows

angt:
Related to https://github.com/ggml-org/llama.cpp/pull/19623


danbev (approved):
Verified using clang on windows 11.

PR #19670, opened 2026-02-16
Allow partial success of seq_rm for hybrid memory

Nekotekina:
This enables cheap snapshots by combining saving recurrent memory with cheap seq_rm rollbacks. When doing rollback, seq_rm is issued separately from llama_state_seq_set_data_ext which restores recurrent state, thus normal cache state is also restored.


Nekotekina:
Noticed another PR handling this issue in a more complete manner.

Nekotekina:
@ggerganov please review, I need this for my project and also I want to report a regression caused by recent commits:
```
state_read_meta: invalid seq_id, 1 is out of range [0, 1)
state_seq_set_data: error loading state: failed to restore kv cache
```
It used to work before but I have my commit on top of the codebase.

Nekotekina:
Seems not necessary anymore.

PR #19681, opened 2026-02-17
model-conversion : make printing of config values optional

danbev:
This commit updates run-org-model.py to make the printing of model configuration values optional.

The motivation for this change is that not all models have these configuration values defined and those that do not will error when running this script. With these changes we only print the values if they exist.

We could optionally just remove them but it can be useful to see these values when running the original model.

### DIFF
--- a/examples/model-conversion/scripts/causal/run-org-model.py
+++ b/examples/model-conversion/scripts/causal/run-org-model.py
@@ -42,11 +42,15 @@ def load_model_and_tokenizer(model_path, device="auto"):
         config = config.text_config
         multimodal = True
 
-    print("Vocab size:       ", config.vocab_size)
-    print("Hidden size:      ", config.hidden_size)
-    print("Number of layers: ", config.num_hidden_layers)
-    print("BOS token id:     ", config.bos_token_id)
-    print("EOS token id:     ", config.eos_token_id)
+    def print_if_exists(label, obj, attr, default="N/A"):
+        if (val := getattr(obj, attr, None)) is not None:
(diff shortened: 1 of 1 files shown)


examples/model-conversion/scripts/causal/run-org-model.py
```diff
@@ -42,11 +42,15 @@ def load_model_and_tokenizer(model_path, device="auto"):
...
         multimodal = True
 
-    print("Vocab size:       ", config.vocab_size)
-    print("Hidden size:      ", config.hidden_size)
-    print("Number of layers: ", config.num_hidden_layers)
-    print("BOS token id:     ", config.bos_token_id)
-    print("EOS token id:     ", config.eos_token_id)
+    def print_if_exists(label, obj, attr, default="N/A"):
+        if (val := getattr(obj, attr, None)) is not None:
+            print(f"{label}", val)
```
CISC:
```suggestion
        val = getattr(obj, attr) if hasattr(obj, attr) else default
        print(f"{label}", val)
```
Sometimes f.ex. `bos_token_id` is `null`, in that case it's useful to print that.

Also, you weren't using `default`, change the code (and remove `default`) if the intention really was to not display anything when not defined.

CISC (approved):

examples/model-conversion/scripts/causal/run-org-model.py
danbev: > Sometimes f.ex. `bos_token_id` is `null`, in that case it's useful to print that.

Good point, I'll update this. Thanks

PR #19686, opened 2026-02-17
CUDA: fix kernel selection logic for tile FA

### LINKED ISSUE #19652: Misc. bug: GGML FlashAttn Crash on NVIDIA Card
### Name and Version

GGML latest commit d6754f3d0e6d0acd

### Operating systems

Windows

Test code

### Command line

```shell

```

My graphics card is Geforce 1060, Pascal architecture
Run the unit test `test-backend-ops`
It would emit "Fatal Error" on the following case
``` cpp
test_cases.emplace_back(new test_flash_attn_ext(
        576, 512, 1, { 4, 1 }, 512, 32, true, true, 0.000000, 0.000000, GGML_PREC_F32, GGML_TYPE_F16));
```
Passed commit, it prints not support.
``` bash
commit fcc2a5c0cfd81ee0517ee42f1acdc371ec92d598 (HEAD)
Author: Georgi Gerganov <ggerganov@gmail.com>
Date:   Sat Oct 11 18:01:19 2025 +0300

    sync : llama.cpp
```
The output is
``` bash
  FLASH_ATTN_EXT(hsk=576,hsv=512,nh=1,nr23=[4,1],kv=512,nb=32,mask=1,sinks=1,max_bias=0.000000,logit_softcap=0.000000,prec=f32,type_KV=f16,permute=[0,1,2,3]): not supported [CUDA0]
```
Crash commit
``` bash
commit 0b56d427d54efe1797180de4f6db4b2d62e09ae7 (HEAD)
Author: Johannes Gäßler <johannesg@5d6.de>
Date:   Sat Oct 11 20:54:32 2025 +0200

    CUDA: faster tile FA, add oob checks, more HSs (llama/16492)
```

Crash call stack similar as this
``` bash
GGML_ABORT("fatal error");
launch_fattn_tile_switch_ncols2<576, 512, false>(...)
ggml_cuda_flash_attn_ext_tile_case<576, 512>(...)
ggml_cuda_flash_attn_ext(...)
```

The root cuase happens at
``` cpp
    const int gqa_limit = nvidia && gqa_ratio <= 4 ? 16 : INT_MAX;
    const bool use_gqa_opt = mask && max_bias == 0.0f && Q->ne[1] <= gqa_limit && K->ne[1] % FATTN_KQ_STRIDE == 0;
```
In the case, `Q->ne[1] <= gqa_limit` doesn't meet, so `use_gqa_opt` is false, so it would emit "Fatal Error"
Remove the requirement,  it would output
``` bash
  FLASH_ATTN_EXT(hsk=576,hsv=512,nh=1,nr23=[4,1],kv=512,nb=32,mask=1,sinks=1,max_bias=0.000000,logit_softcap=0.000000,prec=f32,type_KV=f16,permute=[0,1,2,3]): OK
```

I am not sure whether it is right.

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

JohannesGaessler:
Fixes https://github.com/ggml-org/llama.cpp/issues/19652 .

The problem is that the kernel selection logic is slightly wrong so the host code runs into an abort.

### DIFF
--- a/ggml/src/ggml-cuda/fattn-tile.cuh
+++ b/ggml/src/ggml-cuda/fattn-tile.cuh
@@ -1187,7 +1187,7 @@ static void launch_fattn_tile_switch_ncols2(ggml_backend_cuda_context & ctx, ggm
     const int gqa_ratio = Q->ne[2] / K->ne[2];
 
     const bool nvidia = GGML_CUDA_CC_IS_NVIDIA(ggml_cuda_info().devices[ggml_cuda_get_device()].cc);
-    const int gqa_limit = nvidia && gqa_ratio <= 4 ? 16 : INT_MAX;
+    const int gqa_limit = nvidia && gqa_ratio <= 4 && DV <= 256 ? 16 : INT_MAX;
     const bool use_gqa_opt = mask && max_bias == 0.0f && Q->ne[1] <= gqa_limit && K->ne[1] % FATTN_KQ_STRIDE == 0;
 
     if constexpr (DV == 512) {
--- a/ggml/src/ggml-cuda/fattn-tile.cuh
+++ b/ggml/src/ggml-cuda/fattn-tile.cuh
@@ -1186,6 +1186,8 @@ static void launch_fattn_tile_switch_ncols2(ggml_backend_cuda_context & ctx, ggm
     GGML_ASSERT(Q->ne[2] % K->ne[2] == 0);
     const int gqa_ratio = Q->ne[2] / K->ne[2];
 
+    // On NVIDIA (Pascal and older) the GQA optimizations seem to be detrimental in some cases.
+    // However, for DKQ == 576, DV == 512 only the kernel variant with GQA optimizations is implemented.
     const bool nvidia = GGML_CUDA_CC_IS_NVIDIA(ggml_cuda_info().devices[ggml_cuda_get_device()].cc);
     const int gqa_limit = nvidia && gqa_ratio <= 4 && DV <= 256 ? 16 : INT_MAX;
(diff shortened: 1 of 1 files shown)


pwilkin:
@JohannesGaessler CI seems to dislike something:

```
  MUL_MAT(type_a=q5_1,type_b=f32,m=16,n=1,k=256,bs=[1,1],nr=[1,1],per=[0,1,2,3],k_v=0,o=1)
  Backend CUDA0: FAIL
```

JohannesGaessler:
```console
2026-02-17T14:06:08.1940247Z [MUL_MAT] ERR = 0.000589984 > 0.000500000   MUL_MAT(type_a=q5_1,type_b=f32,m=16,n=1,k=256,bs=[1,1],nr=[1,1],per=[0,1,2,3],k_v=0,o=1): [1;31mFAIL[0m
```

The failing test is a matrix multiplication which is 100% unaffected by the changes in this PR. Since we are using random seeds in the tests it can happen from time to time that the small differences between the CUDA and CPU backends randomly exceed the tolerance we set, particularly if the input matrices are small and the differences are less likely to average out.

ggerganov (approved):
Adding a comment/reference of the DV limit would be useful.

pwilkin:
> The failing test is a matrix multiplication which is 100% unaffected by the changes in this PR. Since we are using random seeds in the tests it can happen from time to time that the small differences between the CUDA and CPU backends randomly exceed the tolerance we set, particularly if the input matrices are small and the differences are less likely to average out.

Aight, just throwing out an idea: maybe it would be a good idea to vary the permitted NMSE divergence according to tensor size, to avoid false positives like this?

PR #19687, opened 2026-02-17
model : Add tokenizer from LFM2.5-Audio-1.5B

tdakhran:
[LFM2.5-Audio-1.5B](https://huggingface.co/LiquidAI/LFM2.5-Audio-1.5B) introduced a lightweight audio tokenizer.
It is based on the LFM2 architecture and serves as an embedding model with different input `n_embd` and output `n_embd_out`.

To be used in https://github.com/ggml-org/llama.cpp/pull/18641.

To convert use

```shell
python3 convert_hf_to_gguf.py /path/to/LFM2.5-Audio-1.5B/audio_detokenizer
```

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -10898,7 +10898,7 @@ class LFM25AudioTokenizer(LFM2Model):
     def set_gguf_parameters(self):
         super().set_gguf_parameters()
         self.gguf_writer.add_sliding_window(self.hparams["sliding_window"])
-        self.gguf_writer.add_embedding_length_out(self.hparams.get("output_size"))
+        self.gguf_writer.add_embedding_length_out(self.hparams["output_size"])
 
     def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
         if name == "istft.window" or name.startswith("emb.emb"):
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -10891,6 +10891,25 @@ class LFM2AudioModel(ConformerAudioModel):
         yield from super().modify_tensors(data_torch, name, bid)
 
 
+@ModelBase.register("Lfm25AudioTokenizer")
+class LFM25AudioTokenizer(LFM2Model):
+    model_arch = gguf.MODEL_ARCH.LFM2
+
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+        self.gguf_writer.add_sliding_window(self.hparams["sliding_window"])
+        self.gguf_writer.add_embedding_length_out(self.hparams.get("output_size"))
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        if name == "istft.window" or name.startswith("emb.emb"):
+            return []
+
+        if name.startswith("lin"):
+            name = name.replace("lin", "dense_2_out")
+
+        yield from super().modify_tensors(data_torch, name, bid)
+
+
 @ModelBase.register("SmallThinkerForCausalLM")
 class SmallThinkerModel(TextModel):
     model_arch = gguf.MODEL_ARCH.SMALLTHINKER
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -10895,6 +10895,9 @@ class LFM2AudioModel(ConformerAudioModel):
 class LFM25AudioTokenizer(LFM2Model):
     model_arch = gguf.MODEL_ARCH.LFM2
 
+    def set_vocab(self):
+        self._set_vocab_none()
+
     def set_gguf_parameters(self):
         super().set_gguf_parameters()
         self.gguf_writer.add_sliding_window(self.hparams["sliding_window"])
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -10902,7 +10902,7 @@ class LFM25AudioTokenizer(LFM2Model):
 
     def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
         if name == "istft.window" or name.startswith("emb.emb"):
-            return []
+            return
 
         if name.startswith("lin"):
             name = name.replace("lin", "dense_2_out")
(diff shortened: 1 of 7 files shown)


convert_hf_to_gguf.py
```diff
@@ -10891,6 +10891,25 @@ def modify_tensors(self, data_torch, name, bid):
...
+    model_arch = gguf.MODEL_ARCH.LFM2
+
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+        self.gguf_writer.add_sliding_window(self.hparams["sliding_window"])
+        self.gguf_writer.add_embedding_length_out(self.hparams.get("output_size"))
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        if name == "istft.window" or name.startswith("emb.emb"):
+            return []
```
```suggestion
            return
```

CISC (approved):

CISC:
I see `layer_types` content differs in the config, should not this change be included?
https://github.com/ggml-org/llama.cpp/pull/18641/files#diff-ec77d8003b92ff283179456d36b8b56abf635e7b1232e70daf16676e8920ccf1L10299-R10299

tdakhran:
> I see `layer_types` content differs in the config, should not this change be included? https://github.com/ggml-org/llama.cpp/pull/18641/files#diff-ec77d8003b92ff283179456d36b8b56abf635e7b1232e70daf16676e8920ccf1L10299-R10299

You are absolutely right! I had multiple checkpoints and tested with wrong one... will fix ASAP

tdakhran:
Thank you @CISC , pushed the fix, and now it works correctly with the upstream HF checkpoint
```shell
INFO:hf-to-gguf:gguf: key-value head count = [0, 0, 8, 0, 8, 0, 8, 0]
```

tdakhran:
Tokenizer in this checkpoint is for conversion compatibility only. If there is a way to convert without a tokenizer, I can remove it.

tdakhran:
Will test it a bit more.

CISC:
> Tokenizer in this checkpoint is for conversion compatibility only. If there is a way to convert without a tokenizer, I can remove it.

I'm not sure I follow. Truthfully I haven't looked into the other PR, so I'm sure I'm missing some details. :)

tdakhran:
> > Tokenizer in this checkpoint is for conversion compatibility only. If there is a way to convert without a tokenizer, I can remove it.
> 
> I'm not sure I follow. Truthfully I haven't looked into the other PR, so I'm sure I'm missing some details. :)

The `audio tokenizer` is a model that takes embeddings as inputs and produces embeddings as output. I mean that it doesn't need `vocab` and `tokenizer`, because it doesn't have "tokens".

 
@CISC , thanks for your patience, but I didn't push the whole set of runtime changes required, somehow thought they are already merged.

Pushed them in this commit https://github.com/ggml-org/llama.cpp/pull/19687/commits/e3ae9359702f48843547fae66da8d43eabdb8eba, please re-review.

CISC:
> The `audio tokenizer` is a model that takes embeddings as inputs and produces embeddings as output. I mean that it doesn't need `vocab` and `tokenizer`, because it doesn't have "tokens".

I see, may require you to make a few changes to allow such a thing to load, the conversion part should be simple though. Maybe @ggerganov has some ideas?

src/models/lfm2.cpp
```diff
@@ -1,18 +1,27 @@
 #include "models.h"
 
+//
```
CISC:
```suggestion
```

src/models/lfm2.cpp
```diff
@@ -1,18 +1,27 @@
...
+llm_build_lfm2<iswa>::llm_build_lfm2(const llama_model & model, const llm_graph_params & params) :
     llm_graph_context(params),
     model(model) {
+    inp_hybrid_type * inp_hybrid = nullptr;
+    if constexpr (iswa) {
+        inp_hybrid = build_inp_mem_hybrid_iswa();
+    } else {
+        inp_hybrid = build_inp_mem_hybrid();
+    }
+
```
CISC, line 145: Move this after `ggml_build_forward_expand`.

tdakhran: addressed in https://github.com/ggml-org/llama.cpp/pull/19687/commits/6c41320d75436a87ff996c66b23ff797adda4ae6

src/models/lfm2.cpp
tdakhran: addressed in https://github.com/ggml-org/llama.cpp/pull/19687/commits/6c41320d75436a87ff996c66b23ff797adda4ae6

tdakhran:
looks like this did the job 
```python
    def set_vocab(self):
        self._set_vocab_none()
```

CISC:
> looks like this did the job
> 
> ```python
>     def set_vocab(self):
>         self._set_vocab_none()
> ```

Oh, right, forgot that was already a thing. :D

src/models/lfm2.cpp
```diff
@@ -73,10 +80,10 @@ ggml_tensor * llm_build_lfm2::build_dense_feed_forward(ggml_tensor * cur, int il
         NULL, LLM_FFN_SILU, LLM_FFN_PAR, il);
 }
 
-ggml_tensor * llm_build_lfm2::build_attn_block(ggml_tensor *             cur,
-                                               ggml_tensor *             inp_pos,
-                                               llm_graph_input_attn_kv * inp_attn,
-                                               int                       il) const {
+template <bool iswa> ggml_tensor * llm_build_lfm2<iswa>::build_attn_block(ggml_tensor *   cur,
```
CISC: Can we perhaps move all these out into a base class in a separate file so we don't build duplicates?

tdakhran: I went the lazy way and just moved them to inline lambdas in https://github.com/ggml-org/llama.cpp/pull/19687/commits/0b6e909188679efb35926ab6aaf377759258d5e7. Please let me know if this works. also rebased.

CISC: Hmmm, ok, I missed that they both used the conditional types so I guess it's unavoidable.

tdakhran:
@CISC , please let me know if anything is needed to proceed with the merge.

CISC:
> @CISC , please let me know if anything is needed to proceed with the merge.

Sorry, should be good, just juggling a lot of PRs. :)

PR #19689, opened 2026-02-17
Pre-MCP UI and architecture cleanup

allozaur:
> [!NOTE]
> This is one of the PRs to unload code from https://github.com/ggml-org/llama.cpp/pull/18655 that is not directly MCP-related

- Refactors chat input flow: ChatForm is made more generic, while orchestration (send/focus/draft behavior) moves into ChatScreenForm
- Reworks message editing to use shared contexts for chat actions and edit state
- Cleans up Service/store structure
- Tightens streaming and generation handling (abort handling, partial save behavior, processing-state tracking, branching/regeneration paths)
- Normalizes reasoning/tool-call handling around tags and structured fields
- Improves model metadata handling with TTL-based caching in modelsStore
- Cleans up settings/type/constants (new settings keys/constants, enum/type normalization, deprecated legacy fields)
- Updates dependencies


ggerganov (approved):

allozaur:
@ggerganov I've extracted all of the remaining code from #18655 that is not directly related to the MCP implementation, but is a result of architecture/UI improvements made along the way. After we have this one merged we are left only with actual MCP implementation code to review and test.

I've tested this code and it seems to be working properly, but I'd appreciate also a few tests from you as well :)

tools/server/webui/src/lib/components/app/chat/ChatForm/ChatForm.svelte
```diff
@@ -110,46 +121,109 @@
...
+	 */
+
+	/**
+	 *
+	 *
+	 * EVENT HANDLERS - File Management
+	 *
+	 *
+	 */
+
```
allozaur:
```suggestion
	/**
	 *
	 *
	 * EVENT HANDLERS - File Management
	 *
	 *
	 */

```

tools/server/webui/src/lib/utils/headers.ts
allozaur, line 1: this one is not needed in this PR

ggerganov (approved):

PR #19694, opened 2026-02-17
server : fix V-L embedding model support

oliveagle:
## Summary

Fix server crashes when using V-L (Vision-Language) embedding models like Qwen3-VL-Embedding.

## Changes

- **common/speculative.cpp**: Skip speculative decoding check for embedding models (they don't have output logits needed for speculative decoding)
- **src/llama-context.cpp**: Add tensor bounds validation for V-L architectures where embedding tensor dimensions may differ

## Problem

When running V-L embedding models with "--embedding --pooling mean", the server would crash:
1. During warmup in "common_speculative_is_compat" - because embedding models lack output logits
2. During inference in "decode" - because V-L models have different tensor sizes

## Testing

Tested with:
- Qwen3-VL-Embedding-2B-Q8_0.gguf
- Qwen3-VL-Embedding-8B-Q8_0.gguf
- Server starts successfully and responds to /v1/embeddings requests

## AI Usage Disclosure

This fix was developed with AI assistance for identifying the root cause and implementing the fix. The changes were manually reviewed and tested.

### DIFF
--- a/src/llama-context.cpp
+++ b/src/llama-context.cpp
@@ -1664,33 +1664,19 @@ int llama_context::decode(const llama_batch & batch_inp) {
                         auto & embd_seq_out = embd_seq;
 
                         // For V-L models, the embedding output tensor may have different dimensions
-                        // Use tensor's actual size to determine correct embedding dimension
-                        const size_t tensor_size = ggml_nbytes(t_embd);
-                        const uint32_t n_embd_tensor = tensor_size / (ubatch.n_seqs_unq > 0 ? ubatch.n_seqs_unq : 1) / sizeof(float);
-                        const uint32_t n_embd_to_use = (n_embd_tensor > 0 && n_embd_tensor < n_embd) ? n_embd_tensor : n_embd;
+                        // The embedding dimension is determined by the tensor shape (ne[0]), not by model hparams
+                        const uint32_t n_embd_tensor = t_embd->ne[0];
+
+                        // Use the tensor's embedding dimension if valid, otherwise fall back to model dimension
+                        const uint32_t n_embd_to_use = n_embd_tensor > 0 ? n_embd_tensor : n_embd;
 
                         for (uint32_t s = 0; s < ubatch.n_seqs_unq; ++s) {
                             const llama_seq_id seq_id  = ubatch.seq_id_unq[s];
                             const int32_t      seq_idx = ubatch.seq_idx[seq_id];
 
                             embd_seq_out[seq_id].resize(n_embd_to_use);
-                            const size_t src_offset = (size_t)n_embd_to_use * seq_idx * sizeof(float);
-                            const size_t copy_size = (size_t)n_embd_to_use * sizeof(float);
-                            // Validate bounds
-                            if (src_offset + copy_size <= tensor_size) {
-                                ggml_backend_tensor_get_async(backend_embd, t_embd, embd_seq_out[seq_id].data(), src_offset, copy_size);
-                            } else {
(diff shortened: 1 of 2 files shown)


src/llama-context.cpp
```diff
@@ -1705,12 +1705,34 @@ int llama_context::decode(const llama_batch & batch_inp) {
...
-                            embd_seq_out[seq_id].resize(n_embd);
-                            ggml_backend_tensor_get_async(backend_embd, t_embd, embd_seq_out[seq_id].data(), (n_embd*seq_idx)*sizeof(float), n_embd*sizeof(float));
+                            embd_seq_out[seq_id].resize(n_embd_to_use);
+                            const size_t src_offset = (size_t)n_embd_to_use * seq_idx * sizeof(float);
+                            const size_t copy_size = (size_t)n_embd_to_use * sizeof(float);
+                            // Validate bounds
+                            if (src_offset + copy_size <= tensor_size) {
+                                ggml_backend_tensor_get_async(backend_embd, t_embd, embd_seq_out[seq_id].data(), src_offset, copy_size);
+                            } else {
+                                LLAMA_LOG_ERROR("%s: tensor bounds check failed: offset=%zu + size=%zu > tensor_size=%zu, using fallback\n",
```
ngxson: This "fallback" logic looks fragile, I bet it will fail in some cases.

Instead, any info about embedding dimension must be determine on ubatch creation

ngxson (changes requested):

oliveagle:
@gxson Thanks for the review! I've simplified the embedding dimension handling as suggested:

1. **Removed fragile fallback logic** - The complex bounds checking and fallback paths have been removed
2. **Use tensor shape directly** - Now using `t_embd->ne[0]` to get the actual embedding dimension from tensor shape
3. **Cleaner code** - Reduced from ~21 lines to ~7 lines, much more maintainable

The new approach simply reads the embedding dimension directly from the tensor's first dimension (`ne[0]`), which is set correctly during graph computation. This avoids the need for the complex fallback logic that was flagged as fragile.

Please let me know if this looks better now.

oliveagle:
@Rebased to latest master (237958db3).

- Resolved any potential merge conflicts
- Code remains functionally identical, just rebased on top of 38 new commits
- Compilation verified successfully

Ready for another round of review if needed.

oliveagle:
## Test Report - V-L Embedding Fix Verification

### Test Environment
- **Hardware**: AMD Strix Halo (Ryzen AI) + NVIDIA V100
- **Models Tested**:
  - Qwen3-VL-Embedding-2B-Q8_0
  - Qwen3-VL-Embedding-8B-Q8_0

### Test Results

#### Before Fix
- Server crashed during warmup in `common_speculative_is_compat`
- Error: embedding models lack output logits needed for speculative decoding compatibility check
- Server also crashed during inference in `decode` due to tensor size mismatches

#### After Fix
- ✅ Server starts successfully with `--embedding --pooling mean` flags
- ✅ Health endpoint responds correctly
- ✅ /v1/embeddings endpoint generates embeddings
- ✅ No crashes during warmup or inference

### Code Review

The fix is minimal and targeted:

**1. common/speculative.cpp**
- Skips speculative decoding compatibility check for embedding models
- Correctly identifies embedding models via pooling type

**2. src/llama-context.cpp**  
- Handles variable tensor sizes in V-L architectures
- Uses tensor dimensions from `t_embd->ne[0]` instead of assuming model hparams
- Adds fallback logic for edge cases

### Request

This fix unblocks LLaDA2 support (#19740) which depends on V-L embedding handling. Could you please review and merge?

Happy to provide additional test data if needed.

PR #19698, opened 2026-02-17
Add  Mistral Voxtral Mini 4B Realtime 2602 4B streaming ASR support

Acceldium:
Add full support for the Mistral Voxtral Mini 4B Realtime 2602 model, including:

- Causal audio encoder with sliding window attention (window=750), causal conv1d stem, RoPE, SwiGLU FFN, RMSNorm, and 4x stacking
- Voxtral-specific mel spectrogram preprocessor (Slaney scale, fixed-max normalization, streaming padding)
- Dual-stream inference protocol combining audio and text embeddings
- Adaptive RMSNorm (precomputed ada_scale) in the text decoder
- Streaming CLI tool (llama-voxtral-stream) with real-time microphone input via miniaudio, chunked processing, and ring buffer
- GGUF conversion support for Mistral-format Voxtral weights (two-file output: text model + mmproj encoder)
- New tensor types: FFN_ADA_NORM_UP, FFN_ADA_NORM_DOWN
- Tied embeddings handling for the text decoder


convert_hf_to_gguf.py
```diff
@@ -2538,8 +2538,12 @@ def __init__(self, *args, **kwargs):
...
         if self.hf_arch == "VLlama3ForCausalLM":
             self.hparams["num_attention_heads"] = self.hparams.get("num_attention_heads", 32)
-        hparams = ModelBase.load_hparams(self.dir_model, is_mistral_format=False)
-        self.origin_hf_arch = hparams.get('architectures', [None])[0]
+        try:
+            hparams = ModelBase.load_hparams(self.dir_model, is_mistral_format=False)
+            self.origin_hf_arch = hparams.get('architectures', [None])[0]
+        except Exception:
+            # Mistral-format models may not have config.json
+            self.origin_hf_arch = None
```
CISC: Undo, you are required to use `--mistral-format` in this case.

CISC (commented):
Please read our [contributing guideline](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) carefully.

This PR is likely to get rejected:
https://github.com/ggml-org/llama.cpp/blob/63206230a1bc10cf8b350a521adc9b3fb3e82e2b/tools/mtmd/mtmd.h#L31-L32

ngxson:
I appreciate the efforts, but the issue is that this PR duplicates a lot of code while also adding a new model-specific binary which is exactly what `mtmd` set out to prevent.

A more generic, model-agnostic approach is necessary. Otherwise, I think we can keep this PR as a demo.

convert_hf_to_gguf.py
Acceldium: Good point , it was revised.

Acceldium:
> A more generic, model-agnostic approach is necessary. Otherwise, I think we can keep this PR as a demo.

The voxtral-stream-cli was merged with mtmd-cli. It was done initiallly to validate numerical parity.

The rest uses the llama.cpp APIs.

Some additional data in here:
https://huggingface.co/acceldium/Voxtral-Mini-4B-Realtime-2602_GGUF

I'm getting 110-114 t/s on a RTX5090 (24GB)

convert_hf_to_gguf.py
CISC: Not quite right though, it has `params.json`, so you just have to pass `is_mistral_format` correctly.

Acceldium: It looks correct to me.

When you pass --mistral-format to convert_hf_to_gguf.py, the script reads params.json instead of config.json. Without that flag, it tries to load config.json (which doesn't exist for Mistral-format models) and fails.

VoxtralRealtimeMmprojModel has is_mistral_format = True 
MistralModel has is_mistral_format = True
LlamaModel.__init__ checks self.is_mistral_format  - if True, it sets origin_hf_arch = None and skips loading config.json

ngxson:
> The voxtral-stream-cli was merged with mtmd-cli.

Look at the change:

image

Beside, as specified in contributing guide:

image

I personally refuse to review this PR, it has too many added features without careful planning.

didlawowo:
Hi, I've been looking into PR #19698 and I'd like to contribute a
  cleaner implementation, split into incremental PRs as suggested by
  @ngxson.

  Phase 1 — Core model support only (~800 lines):
  - GGUF conversion for VoxtralRealtimeForConditionalGeneration
  - Causal audio encoder in mtmd (voxtral-realtime-enc.cpp)
  - Decoder ada_rms_norm_t_cond support (precomputed at conversion time)
  - Audio preprocessor with Voxtral-specific mel params
  - No new binary, no mic capture, no streaming — just llama-server -m
  model.gguf --mmproj mmproj.gguf

  Phase 2 — Dual-stream inference protocol (separate PR, after
  discussion)

  Phase 3 — Mic capture / real-time streaming (separate PR)

  Before I start, I'd like to confirm this approach works for the
  maintainers. @ngxson would you be open to reviewing a scoped Phase 1
  PR?

ngxson:
This PR contains many model-specific code paths, which are considered as anti-pattern in libmtmd design. Please refer to https://github.com/ggml-org/llama.cpp/issues/20914 for a formal discussion before proceeding. For now, I'm closing the current PR.

PR #19711, opened 2026-02-18
llama: allow llama_params_fit to account host memory information for context size reduction

taronaeo:
fixes: #19646

This PR enables the `llama_params_fit` system to default to host memory when no dedicated backend devices are available. Prior to this PR, context size for CPU-only build defaults to `131072` which uses way too much resources on normal systems and would fail on startup with default flags if memory is unable to be allocated.

With this PR, context sizes remain at a default of `131072` and the context size reduction logic will now take into account the host memory information if no devices are available. As stated in https://github.com/ggml-org/llama.cpp/pull/19711#issuecomment-3926219519, we can only assume that the host free memory is the same as the total available memory on the system.

also fixes: #18733 #19714

### DIFF
--- a/src/llama.cpp
+++ b/src/llama.cpp
@@ -88,48 +88,72 @@ static std::vector<llama_device_memory_data> llama_get_device_memory_data(
         throw std::runtime_error("failed to create llama_context from model");
     }
 
-    std::vector<llama_device_memory_data> ret(model->devices.size());
+    // if no devices are found, we default to the CPU backend
+    ggml_backend_dev_t cpu_dev = nullptr;
+    if (model->devices.empty()) {
+        cpu_dev = ggml_backend_dev_by_type(GGML_BACKEND_DEVICE_TYPE_CPU);
+        if (cpu_dev == nullptr) {
+            throw std::runtime_error(format("%s: no CPU backend found", __func__));
+        }
+    }
+
+    // automatically assume 1 device since the CPU backend is always available
+    size_t n_dev = model->devices.empty() ? 1 : model->devices.size();
+    std::vector<llama_device_memory_data> ret(n_dev);
 
     std::map<ggml_backend_buffer_type_t, llama_memory_breakdown_data> memory_breakdown = ctx->memory_breakdown();
 
     for (const auto & [buft, mb] : memory_breakdown) {
-        if (ggml_backend_buft_is_host(buft)) {
-            continue;
-        }
+        if (cpu_dev) {
+            ret[0].mb.model   += mb.model;
+            ret[0].mb.context += mb.context;
+            ret[0].mb.compute += mb.compute;
+        } else {
+            if (ggml_backend_buft_is_host(buft)) {
+                continue;
+            }
 
-        ggml_backend_dev_t dev = ggml_backend_buft_get_device(buft);
-        if (!dev) {
-            continue;
-        }
-        for (size_t i = 0; i < ret.size(); i++) {
-            if (model->devices[i] == dev) {
-                ret[i].mb.model   += mb.model;
-                ret[i].mb.context += mb.context;
-                ret[i].mb.compute += mb.compute;
-                break;
+            ggml_backend_dev_t dev = ggml_backend_buft_get_device(buft);
+            if (!dev) {
+                continue;
+            }
+            for (size_t i = 0; i < ret.size(); i++) {
+                if (model->devices[i] == dev) {
+                    ret[i].mb.model   += mb.model;
+                    ret[i].mb.context += mb.context;
+                    ret[i].mb.compute += mb.compute;
+                    break;
+                }
             }
         }
     }
+
     for (size_t i = 0; i < ret.size(); i++) {
         size_t free;
         size_t total;
-        ggml_backend_dev_memory(model->devices[i], &free, &total);
 
-        // devices can return 0 bytes for free and total memory if they do not
-        // have any to report. in this case, we will use the host memory as a fallback
-        // fixes: https://github.com/ggml-org/llama.cpp/issues/18577
-        if (free == 0 && total == 0) {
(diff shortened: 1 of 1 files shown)


CISC:
IIRC the default used to be `4096`.

taronaeo:
> IIRC the default used to be `4096`.

Oops, updated the PR to `4096` :)

JohannesGaessler:
I think this should be fixed differently, a context size of 4096 will be a better default in some cases but a worse one in others. We don't have an implementation for determining the "free" memory for system RAM but we can relatively easily determine the total memory. So I would suggest that the context reduction logic considers all system RAM as free and reduces the context size accordingly. The default margin left is 1 GiB which should be safe on devices with very low memory where other processes can be expected to be relatively light.

taronaeo:
I've reworked the logic to default back to the host memory information if no devices can be found. The context size reduction logic should now use the host memory information to resize accordingly.

Tested on x86 Ubuntu Linux with 32 GiB of memory and CPU-only build:

```console
$ build/bin/llama-completion -hf ibm-granite/granite-3.3-8b-instruct-GGUF:Q4_K_M -no-cnv --temp 0 --seed 42 -p "Hello world" -ctk f32 -ctv f32 -n 50

...truncated...
build: 8125 (e877ad8bd) with GNU 14.2.0 for Linux x86_64
main: llama backend init
main: load the model and apply lora adapter, if any
llama_params_fit_impl: projected to use 46608 MiB of device memory vs. 31984 MiB of free device memory
llama_params_fit_impl: cannot meet free memory target of 1024 MiB, need to reduce device memory by 15647 MiB
llama_params_fit_impl: context size reduced from 131072 to 81920 -> need 15673 MiB less memory in total
llama_params_fit_impl: entire model can be fit by reducing context
llama_params_fit: fitting params to free memory took 0.17 seconds
```

Tested on s390x Ubuntu Linux with 2 GiB of memory and CPU-only build:

```console
$ build/bin/llama-completion -m /opt/hf_models/granite-3.3-2b-instruct-be.Q4_K_M.gguf --temp 0 --seed 42 -p "Hello world"

build: 8137 (807ecd922) with GNU 14.2.0 for Linux s390x
main: llama backend init             
main: load the model and apply lora adapter, if any
llama_params_fit_impl: projected to use 12520 MiB of device memory vs. 3983 MiB of free device memory
llama_params_fit_impl: cannot meet free memory target of 1024 MiB, need to reduce device memory by 9561 MiB
llama_params_fit_impl: context size reduced from 131072 to 15360 -> need 9579 MiB less memory in total
llama_params_fit_impl: entire model can be fit by reducing context
llama_params_fit: fitting params to free memory took 0.20 seconds
```

Tested on x86 Ubuntu Linux with 32 GiB and CUDA build:

```console
$ build/bin/llama-completion -hf ibm-granite/granite-3.3-8b-instruct-GGUF:Q4_K_M -no-cnv --temp 0 --seed 42 -p "Hello world" -ctk f32 -ctv f32 -n 50

  Device 0: NVIDIA GeForce RTX 2060, compute capability 7.5, VMM: yes
...truncated...
build: 8125 (e877ad8bd) with GNU 14.2.0 for Linux x86_64
main: llama backend init
main: load the model and apply lora adapter, if any
llama_params_fit_impl: projected to use 46600 MiB of device memory vs. 5586 MiB of free device memory
llama_params_fit_impl: cannot meet free memory target of 1024 MiB, need to reduce device memory by 42037 MiB
llama_params_fit_impl: context size reduced from 131072 to 4096 -> need 40489 MiB less memory in total
llama_params_fit_impl: filling dense layers back-to-front:
llama_params_fit_impl:   - CUDA0 (NVIDIA GeForce RTX 2060): 30 layers,   4481 MiB used,   1104 MiB free
llama_params_fit: fitting params to free memory took 0.66 seconds
llama_model_load_from_file_impl: using device CUDA0 (NVIDIA GeForce RTX 2060) (0000:09:00.0) - 5586 MiB free
```

@JohannesGaessler PTAL again :)

JohannesGaessler:
You should fix this differently. The memory breakdown is per buffer type rather than per device because there are multiple buffer types that can be in host memory. However, there is a check in `llama_get_device_memory_data` to filter out all host buffers. Remove that check and then in `llama_params_fit` check whether only host buffer types are returned. If no, filter out all host buffer types. If yes, adjust only the context size based on the maximum memory. In principle, if we ever get an agreed-upon definition of "free" system memory we can then adjust this code to adjust the context size based on that.

taronaeo:
> However, there is a check in llama_get_device_memory_data to filter out all host buffers. Remove that check and then in llama_params_fit check whether only host buffer types are returned

I'm not sure I follow. First hurdle, removing the check to filter out all host buffers would do nothing since the CPU buffer doesn't actually have a device pointer. See: 

```gdb
(gdb) p *dev
Cannot access memory at address 0x0
(gdb) p *buft
$1 = {iface = {get_name = 0x7ffff7569329 <ggml_backend_cpu_buffer_type_get_name(ggml_backend_buffer_type_t)>, 
    alloc_buffer = 0x7ffff756933e <ggml_backend_cpu_buffer_type_alloc_buffer(ggml_backend_buffer_type_t, size_t)>, 
    get_alignment = 0x7ffff7569427 <ggml_backend_cpu_buffer_type_get_alignment(ggml_backend_buffer_type_t)>, get_max_size = 0x0, get_alloc_size = 0x0, 
    is_host = 0x7ffff756943a <ggml_backend_cpu_buffer_type_is_host(ggml_backend_buffer_type_t)>}, device = 0x0, context = 0x0}
```

Furthermore, `ret.size()` will return 0 for CPU-only builds since 

https://github.com/ggml-org/llama.cpp/blob/05728db18eea59de81ee3a7699739daaf015206b/src/llama.cpp#L969-L978

doesn't account the CPU device into the devices list. All of the calculations will not iterate since there are "0 devices" to loop.

I'm pretty sure I'm not understanding this system properly/entirely so feel free to correct me :)

taronaeo:
Closing in favour of #21568

PR #19725, opened 2026-02-19
llama: add BertForTokenClassification support

giladgd:
This PR adds token classification support for `BertForTokenClassification` and `ModernBertForTokenClassification`, which allows me to use [this model](https://huggingface.co/dslim/bert-base-NER) and also [this one](https://huggingface.co/OpenMed/OpenMed-NER-GenomeDetect-ModernMed-395M).
I added a new pooling type for token-level classification since it makes `llama_get_embeddings_ith` return an array with a length of `llama_model_n_cls_out(model)` (instead of `llama_model_n_embd_out(model)`).

---

AI disclosure: I used Codex for assistance but wrote most of the code myself.


giladgd:
@CISC I saw you reviewed and added support for classification models in the past, so it'd be great if you could review this PR

PR #19738, opened 2026-02-19
ggml-cpu: add repack for mxfp4

am17an:
This is just a faithful copy of the `iq4_nl` quant to mxfp4 with just the scale loading changed. Tested on AVX2 only, would appreciate tests on ARM and AVX512. Perplexity is already high for gpt-oss-20b but I see it is the same between master and this branch

| Model                 | Test   |   t/s master |   t/s mxfp4-repack-cpu |   Speedup |
|:----------------------|:-------|-------------:|-----------------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512  |       160.35 |                 197.33 |      1.23 |
| gpt-oss 20B MXFP4 MoE | pp1024 |       160.54 |                 196.24 |      1.22 |
| gpt-oss 20B MXFP4 MoE | pp2048 |       157.45 |                 188.59 |      1.20 |
| gpt-oss 20B MXFP4 MoE | pp4096 |       150.34 |                 181.45 |      1.21 |
| gpt-oss 20B MXFP4 MoE | tg32   |        18.79 |                  21.39 |      1.14 |

### DIFF
--- a/ggml/src/ggml-cpu/arch/x86/repack.cpp
+++ b/ggml/src/ggml-cpu/arch/x86/repack.cpp
@@ -764,6 +764,7 @@ static void gemm_q4_b32_8x8_q8_0_lut_avx(int n, float * GGML_RESTRICT s, size_t 
                         std::is_same_v<block_tx8, block_iq4_nlx8>) {
                     col_scale_f32 = GGML_F32Cx8x2_LOAD(b_ptr_0[b].d, b_ptr_1[b].d);
                 } else if constexpr (std::is_same_v<block_tx8, block_mxfp4x8>) {
+                    //TODO: simd-ify
                     col_scale_f32 = _mm512_set_ps(
                         GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_1[b].e[7]),
                         GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_1[b].e[6]),
@@ -974,6 +975,7 @@ static void gemm_q4_b32_8x8_q8_0_lut_avx(int n, float * GGML_RESTRICT s, size_t 
                         std::is_same_v<block_tx8, block_iq4_nlx8>) {
                     col_scale_f32 = GGML_F32Cx8x2_LOAD(b_ptr_0[b].d, b_ptr_1[b].d);
                 } else if constexpr (std::is_same_v<block_tx8, block_mxfp4x8>) {
+                    //TODO: simd-ify
                     col_scale_f32 = _mm512_set_ps(
                         GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_1[b].e[7]),
                         GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_1[b].e[6]),
--- a/ggml/src/ggml-cpu/arch/x86/repack.cpp
+++ b/ggml/src/ggml-cpu/arch/x86/repack.cpp
@@ -522,7 +522,8 @@ template<typename block_tx8>
 static void gemv_q4_b32_8x8_q8_0_lut_avx(int n, float * GGML_RESTRICT s, size_t bs, const void * GGML_RESTRICT vx, const void * GGML_RESTRICT vy, int nr, int nc, __m256i signextendlut) {
     static_assert(
             std::is_same_v<block_tx8, block_q4_0x8> ||
-            std::is_same_v<block_tx8, block_iq4_nlx8>,
+            std::is_same_v<block_tx8, block_iq4_nlx8> ||
+            std::is_same_v<block_tx8, block_mxfp4x8>,
             "Unsupported block type");
 
     const int qk = QK8_0;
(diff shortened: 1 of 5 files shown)


am17an:
@ggerganov ping for this

ggml/src/ggml-cpu/arch/x86/repack.cpp
```diff
@@ -749,6 +763,24 @@ static void gemm_q4_b32_8x8_q8_0_lut_avx(int n, float * GGML_RESTRICT s, size_t
...
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_1[b].e[1]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_1[b].e[0]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_0[b].e[7]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_0[b].e[6]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_0[b].e[5]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_0[b].e[4]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_0[b].e[3]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_0[b].e[2]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_0[b].e[1]),
+                        GGML_CPU_E8M0_TO_FP32_HALF(b_ptr_0[b].e[0]));
```
ggerganov, line 784: This is likely not optimal. Add TODOs to SIMD-ify in the future.

ggerganov (approved):
Btw, note that although repacking helps for CPU-only systems, it actually hurts the performance for mixed CPU/GPU systems. It's always better to keep the data unpacked and offload it via the "op_offload" mechanism to the GPU for large operations.

So in this case for example, someone running `gpt-oss` with `-ncmoe` will actually have performance degraded due to this change.

I am thinking that probably we have to make repacking to be disabled by default and CPU-only users to opt-in to use it.

am17an:
@ggerganov my use-case was actually to speed up mixed inference, I see that it is beneficial at least for gpt-oss. I'm not sure what happens at 0 context, maybe the repack time is backed into the llama-bench time?

`GGML_CUDA=ON ./scripts/compare-commits.sh 2bf318fd2f690f12ba0ee87ac63157f5b9300886 mxfp4-cpu-repack llama-bench -m /opt/models/gpt_oss-20b-mxfp4.gguf -d 0,1024,2048,4096 -ncmoe 99 -n 32 -p
512 -fa 1`

| Model                 | Test        |   t/s 2bf318fd2 |   t/s mxfp4-cpu-repack |   Speedup |
|:----------------------|:------------|----------------:|-----------------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512       |         1623.97 |                1256.91 |      0.77 |
| gpt-oss 20B MXFP4 MoE | pp512@d1024 |         1820.60 |                1817.50 |      1.00 |
| gpt-oss 20B MXFP4 MoE | pp512@d2048 |         1786.90 |                1783.59 |      1.00 |
| gpt-oss 20B MXFP4 MoE | pp512@d4096 |         1789.31 |                1787.44 |      1.00 |
| gpt-oss 20B MXFP4 MoE | tg32        |           18.36 |                  20.25 |      1.10 |
| gpt-oss 20B MXFP4 MoE | tg32@d1024  |           18.78 |                  20.63 |      1.10 |
| gpt-oss 20B MXFP4 MoE | tg32@d2048  |           18.08 |                  20.14 |      1.11 |
| gpt-oss 20B MXFP4 MoE | tg32@d4096  |           18.80 |                  20.09 |      1.07 |

am17an:
I think that was just noise in my system, ncmoe is harder to accurately measure of because of system noise. So overall I think there is benefit to CPU users, and neutral for CPU/GPU mixed

| Model                 | Test        |   t/s 2bf318fd2 |   t/s mxfp4-cpu-repack |   Speedup |
|:----------------------|:------------|----------------:|-----------------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512       |         1752.19 |                1751.58 |      1.00 |
| gpt-oss 20B MXFP4 MoE | pp512@d1024 |         1809.67 |                1819.17 |      1.01 |
| gpt-oss 20B MXFP4 MoE | pp512@d2048 |         1767.31 |                1787.82 |      1.01 |
| gpt-oss 20B MXFP4 MoE | pp512@d4096 |         1783.82 |                1786.42 |      1.00 |
| gpt-oss 20B MXFP4 MoE | tg32        |           25.21 |                  25.23 |      1.00 |
| gpt-oss 20B MXFP4 MoE | tg32@d1024  |           24.80 |                  24.75 |      1.00 |
| gpt-oss 20B MXFP4 MoE | tg32@d2048  |           25.68 |                  25.42 |      0.99 |
| gpt-oss 20B MXFP4 MoE | tg32@d4096  |           25.80 |                  24.49 |      0.95 |

ggerganov:
With CUDA and `-ncmoe`, the repack is never actually triggered since the `CUDA_Host` buffer type is preferred over the CPU extra buffer type. But for example on Mac it's not the case and the repack takes place, leading to worse performance compared to not repacking + op_offload:

| Model                 | Test   |   t/s b8106 |   t/s mxfp4-cpu-repack |   Speedup |
|:----------------------|:-------|------------:|-----------------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512  |      428.24 |                 274.68 |      0.64 |
| gpt-oss 20B MXFP4 MoE | pp2048 |      435.32 |                 275.53 |      0.63 |
| gpt-oss 20B MXFP4 MoE | tg32   |       53.43 |                  56.99 |      1.07 |

I expect similar behaviour with Vulkan and other backends that do not have a host buffer type.

am17an:
Oh I see, then I think adding `--repack-cpu-weights` flag makes sense? Or should we automatically detect this?

ggerganov:
We already have `--repack`/`--no-repack`. I just think the default value should be disabled. Currently it is enabled by default:

https://github.com/ggml-org/llama.cpp/blob/a96a1120b45c011ef6ed38bcb3fc455209ea5b9e/src/llama-model.cpp#L8853

am17an:
Okay I see, not sure if it makes sense to disable explicitly if no flag is provided when doing hybrid inference

ggerganov:
> Okay I see, not sure if it makes sense to disable explicitly if no flag is provided when doing hybrid inference

I think we can just print a hint to use `--repack` in the logs when we determine that there are no devices registered apart from the CPU. Want to avoid defaults that depend on the hardware/OS/etc.

am17an:
Is this diff good? I can create a separate PR for it 

```cpp
diff --git a/common/common.h b/common/common.h
index 1fa172865..20472f3c8 100644
--- a/common/common.h
+++ b/common/common.h
@@ -484,7 +484,7 @@ struct common_params {
     bool warmup            = true;  // warmup run
     bool check_tensors     = false; // validate tensor data
     bool no_op_offload     = false; // globally disable offload host tensor operations to device
-    bool no_extra_bufts    = false; // disable extra buffer types (used for weight repacking)
+    bool no_extra_bufts    = true;  // disable extra buffer types (used for weight repacking)
     bool no_host           = false; // bypass host buffer allowing extra buffers to be used

     bool single_turn       = false; // single turn chat conversation
diff --git a/src/llama.cpp b/src/llama.cpp
index 6da90d6f1..ce3cde772 100644
--- a/src/llama.cpp
+++ b/src/llama.cpp
@@ -1003,6 +1003,10 @@ static struct llama_model * llama_model_load_from_file_impl(
                 props.memory_free/1024/1024);
     }

+    if (model->devices.empty() && !params.use_extra_bufts) {
+        LLAMA_LOG_WARN("%s: no non-CPU devices found, consider using --repack for better performance\n", __func__);
+    }
+
     const int status = llama_model_load(path_model, splits, *model, params);
     GGML_ASSERT(status <= 0);
     if (status < 0) {
```

ggerganov:
Yes, but also:

- change `use_extra_bufts` to false in `llama_model_default_params()`
- the log message in `llama.cpp` should not refer to the argument `--repack` because it is not a `libllama` parameter. rather refer to the `use_extra_bufts` flag

PR #19749, opened 2026-02-20
clarify operation of gguf-split

alosslessdev:
so that you don't have to find out by trial and error


ngxson (approved):
I think adding some examples for the usage will make it much easier to understand

PR #19752, opened 2026-02-20
llama-server: Support `--verbose-prompt`

a4lg:
llama-server has the option `--verbose-prompt` but was ineffective.

This commit recovers the logger for full prompt input (gated behind the non-default option `--verbose-prompt`) which was originally commented out.

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -2967,8 +2967,9 @@ std::unique_ptr<server_res_generator> server_routes::handle_completions_impl(
         std::vector<server_task> tasks;
 
         const auto & prompt = data.at("prompt");
-        // TODO: this log can become very long, put it behind a flag or think about a more compact format
-        //SRV_DBG("Prompt: %s\n", prompt.is_string() ? prompt.get<std::string>().c_str() : prompt.dump(2).c_str());
+        if (params.verbose_prompt) {
+            SRV_INF("prompt: '%s'\n", prompt.is_string() ? prompt.get<std::string>().c_str() : prompt.dump(2).c_str());
+        }
 
         // process prompt
         std::vector<server_tokens> inputs;


ngxson:
Use slots debug instead

a4lg:
Okay, undocumented slot debugging feature...?

# How to use it?

## Invoking with Slots Debug enabled

```sh
LLAMA_SERVER_SLOTS_DEBUG=1 llama-server --port 11434 [ARGS...]
```

## How to inspect the results?

After the process finishes, inspect `slots`.  Here is an example:

```console
# curl --silent 'http://localhost:11434/slots' | jq
[
  ...,
  {
    "id": 7,
    "n_ctx": 2048,
    "speculative": false,
    "is_processing": false,
    "id_task": 0,
    ...,
    "prompt": "<|im_start|>system\nYou are Qwen, created by Alibaba Cloud. You are a helpful assistant.<|im_end|>\n<|im_start|>user\nHello!<|im_end|>\n<|im_start|>assistant\n",
    "generated": "Hello! How can I assist you today?"
  }
]
```

You may need to specify the model name on the router mode.

```sh
curl --silent 'http://localhost:11434/slots?model=Qwen%2FQwen2.5-Coder-1.5B-Instruct' | jq
```

# Alternatives

@ngxson
I'm half convinced.  Although undocumented, it's a good feature which I have missed.

Then, I came up with one question.  What's the point leaving `--verbose-prompt` which is completely ineffective on `llama-server` in the first place?  If this is a valid question, I'll submit a PR to remove `--verbose-prompt` from `llama-server` instead.

What do you think?

ngxson:
@a4lg It was undocumented because it was mean to be used for testing only. But feel free to add it to a new section at the end of server docs.

> If this is a valid question, I'll submit a PR to remove `--verbose-prompt` from `llama-server` instead.

Yes, please, that would be helpful. You can push it with the same PR I suggested above

PR #19762, opened 2026-02-20
Proposal: Guidelines for quantization schemes

pwilkin:
Since we've had a bit of those popping up lately, I was thinking of adding an explicit requirement for anyone who wants new quants considered to *at least* provide the GGUF, PPL data, KLD data and pure CPU performance data. What do you think?


JohannesGaessler:
My opinion is that as of right now we don't have good tooling to determine whether we should be adding more quantization formats in the first place. To my knowledge no one has of yet conclusively investigated how abstract metrics like PPL or KLD correlate with the actual capabilities of a model. Quite frankly, unless the PR author in question is known to be an expert I don't think we should be adding more quantization formats.

When it comes to the tooling at least I hope to improve this in the coming months.

pwilkin:
> My opinion is that as of right now we don't have good tooling to determine whether we should be adding more quantization formats in the first place. To my knowledge no one has of yet conclusively investigated how abstract metrics like PPL or KLD correlate with the actual capabilities of a model. Quite frankly, unless the PR author in question is known to be an expert I don't think we should be adding more quantization formats.
> 
> When it comes to the tooling at least I hope to improve this in the coming months.

I think that's a fair point, that's why I added "minimal" in there to suggest that those are necessary conditions, not sufficient conditions.

Your point about the tooling is very good BTW. I feel like it might be worth extending `llama-perplexity` which already has a lot of tests people don't use (hellaswag, winogrande) with maybe proper HuggingFace testing dataset support (now that GGML.ai is a partner :) ). I feel like having benchmark scores for quants would go a long way towards establishing their viability.

am17an:
Also just as an aside, adding new quant formats increase the size of the binary in CUDA and also the total compilation time. Just something to keep in mind.

JohannesGaessler:
I don't care as much about the compilation time and binary size but more about the effort for proper implementation and maintenance. As it is a new quantization format would require coordinated efforts in all ggml backends and in the case of e.g. CUDA the extension of multiple kernels even within the same backend. So I want to be reasonably sure that these efforts are actually worth the opportunity cost.

CONTRIBUTING.md
```diff
@@ -30,15 +30,20 @@ Before submitting your PR:
...
+  - If you modified a `ggml` operator or added a new one, add the corresponding test cases to `test-backend-ops`
 - Create separate PRs for each feature or fix:
-    - Avoid combining unrelated changes in a single PR
-    - For intricate features, consider opening a feature request first to discuss and align expectations
-    - When adding support for a new model or feature, focus on **CPU support only** in the initial PR unless you have a good reason not to. Add support for other backends like CUDA in follow-up PRs
-- Consider allowing write access to your branch for faster reviews, as reviewers can push commits directly
+  - Avoid combining unrelated changes in a single PR
+  - For intricate features, consider opening a feature request first to discuss and align expectations
+  - When adding support for a new model or feature, focus on **CPU support only** in the initial PR unless you have a good reason not to. Add support for other backends like CUDA in follow-up PRs
+  - Since it's very easy to propose new quantization schemes and very hard to verify them without concrete data, the *minimum* requirement for a new quantization scheme to be considered is:
```
JohannesGaessler:
```suggestion
  - In particular, adding new data types (extension of the `ggml_type` enum) carries with it a disproportionate maintenance burden. As such, to add a new quantization type you will need to meet the following *additional* criteria *at minimum*:
```

CONTRIBUTING.md
```diff
@@ -30,15 +30,20 @@ Before submitting your PR:
...
+    - providing a model GGUF quantized in the new scheme
+    - providing perplexity comparisons to Q8 as well as to surrounding quants
+    - providing KLD divergence data calculated from the Q8 version for both the new quant as well as the surrounding quants
+    - performance data for the new quantization scheme as compared to surrounding quantization schemes on pure CPU
```
JohannesGaessler: Why are we asking for comparisons against Q8 rather than FP16/BF16 here?

pwilkin: Good question, I guess I'm just too used to doing stuff on Q8_0 thanks to my poor comp. Will change.

CONTRIBUTING.md
```diff
@@ -30,15 +30,20 @@ Before submitting your PR:
...
 - Create separate PRs for each feature or fix:
-    - Avoid combining unrelated changes in a single PR
-    - For intricate features, consider opening a feature request first to discuss and align expectations
-    - When adding support for a new model or feature, focus on **CPU support only** in the initial PR unless you have a good reason not to. Add support for other backends like CUDA in follow-up PRs
-- Consider allowing write access to your branch for faster reviews, as reviewers can push commits directly
+  - Avoid combining unrelated changes in a single PR
+  - For intricate features, consider opening a feature request first to discuss and align expectations
+  - When adding support for a new model or feature, focus on **CPU support only** in the initial PR unless you have a good reason not to. Add support for other backends like CUDA in follow-up PRs
+  - In particular, adding new data types (extension of the `ggml_type` enum) carries with it a disproportionate maintenance burden. As such, to add a new quantization type you will need to meet the following *additional* criteria *at minimum*:
+    - providing a model GGUF quantized in the new scheme
```
JohannesGaessler:
```suggestion
    - convert a small model to GGUF using the new type and upload it to HuggingFace
```

CONTRIBUTING.md
```diff
@@ -30,15 +30,20 @@ Before submitting your PR:
...
+    - providing perplexity comparisons to FP16/BF16 (whichever is the native precision) as well as to surrounding quants
```
JohannesGaessler:
```suggestion
    - provide [perplexity](https://github.com/ggml-org/llama.cpp/tree/master/tools/perplexity) comparisons to FP16/BF16 (whichever is the native precision) as well as to types of similar size
```

CONTRIBUTING.md
```diff
@@ -30,15 +30,20 @@ Before submitting your PR:
...
+    - providing KLD divergence data calculated from the FP16/BF16 (whichever is the native precision) version for both the new quant as well as the surrounding quants
```
JohannesGaessler:
```suggestion
    - provide KL divergence data calculated vs. the FP16/BF16 (whichever is the native precision) version for both the new type as well as types of similar size
```

CONTRIBUTING.md
```diff
@@ -30,15 +30,20 @@ Before submitting your PR:
...
+    - performance data for the new quantization scheme as compared to surrounding quantization schemes on pure CPU
```
JohannesGaessler:
```suggestion
    - provide [performance data](https://github.com/ggml-org/llama.cpp/tree/master/tools/llama-bench) for the new type in comparison to types of similar size on pure CPU
```

JohannesGaessler (approved):

PR #19763, opened 2026-02-20
Server: allow immediate reuse of http port

jpm-canonical:
Add the SO_REUSEADDR flag to the http port, similar to what is done on the GRPC port [here](https://github.com/ggml-org/llama.cpp/blob/b908baf1825b1a89afef87b09e22c32af2ca6548/ggml/src/ggml-rpc/ggml-rpc.cpp#L311-L315).

Related:
* Resolves #19758

AI disclosure: Copilot (Claude Sonnet 4.5) was used to triage and fix the issue.

### DIFF
--- a/tools/server/server-http.cpp
+++ b/tools/server/server-http.cpp
@@ -315,10 +315,12 @@ bool server_http_context::start() {
     } else {
         LOG_INF("%s: binding port with default address family\n", __func__);
 
-        // Enable address reuse to allow immediate restart of server
+        // Set linger time to 0 to force close the socket immediately when the server stops
         srv->set_socket_options([](socket_t sock) {
-            const char flag = 1;
-            setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &flag, sizeof(flag));
+            linger sl{};
+            sl.l_onoff  = 1;
+            sl.l_linger = 0;
+            setsockopt(sock, SOL_SOCKET, SO_LINGER, reinterpret_cast<const char *>(&sl), sizeof(sl));
         });
 
         // bind HTTP listen port
--- a/tools/server/server-http.cpp
+++ b/tools/server/server-http.cpp
@@ -115,7 +115,16 @@ bool server_http_context::init(const common_params & params) {
     srv->set_read_timeout (params.timeout_read);
     srv->set_write_timeout(params.timeout_write);
     srv->set_socket_options([reuse_port = params.reuse_port](socket_t sock) {
+        // Set linger to 0 to avoid TIME_WAIT state after server exits
+        linger sl{};
+        sl.l_onoff  = 1;
+        sl.l_linger = 0;
+        setsockopt(sock, SOL_SOCKET, SO_LINGER, reinterpret_cast<const char *>(&sl), sizeof(sl));
+
+        // Allow binding to port during TIME_WAIT state after server exited
         httplib::set_socket_opt(sock, SOL_SOCKET, SO_REUSEADDR, 1);
+
+        // If flag is set, allow multiple processes to bind to the same port for load balancing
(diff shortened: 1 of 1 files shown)


ngxson:
This seems to break all windows build configurations

jpm-canonical:
Windows builds failed because the value for the socket flag is an int, and it can't be interpreted as a const char. The variable is now changed to a const char. Building succeeds on Linux, as well as using Mingw for Windows.

Green-Sky:
> When SO_REUSEADDR is set on windows, it allows another process to "steal" the address, while in use.

see eg https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse#using-so_exclusiveaddruse

jpm-canonical:
@Green-Sky thanks a lot for the comment and link to the docs. I agree that using `SO_REUSEADDR` has a possible negative side effect. The initial implementation followed what is already being done [in the GRPC server code](https://github.com/ggml-org/llama.cpp/blob/b908baf1825b1a89afef87b09e22c32af2ca6548/ggml/src/ggml-rpc/ggml-rpc.cpp#L311-L315).

I investigated this further, and using `SO_LINGER` with a timeout of 0 seconds also solves the problem mentioned in #19758. This does not introduce the same port hijacking possibility as `SO_REUSEADDR` does.

ngxson:
I restarted the failed CI, seems to be unrelated to this change. will merge this once everything is green.

@Green-Sky feel free to take a quick look

ngxson:
The same workflow failed twice for the same reason, there should be a problem with this PR: https://github.com/ggml-org/llama.cpp/actions/runs/22354279984/job/64980570539?pr=19763

jpm-canonical:
@ngxson indeed. Looking at that test it does use streaming, so most likely touching the same code path. We will need to investigate better what is going on.

jpm-canonical:
I rebased on to the master branch yesterday, and the test that failed before is passing - both locally and in CI:
https://github.com/ggml-org/llama.cpp/actions/runs/24463225944/job/71483177633?pr=19763#step:6:205

There are some other CI test failures, but those seems unrelated to this PR.

ggerganov (approved):

ServeurpersoCom:
Doesn't this conflict with #21056 ?
set_socket_options overwrites the previous callback in cpp-httplib, so calling it again in start() drops the SO_REUSEADDR / SO_REUSEPORT setup from init() and silently disables --reuse-port.

jpm-canonical:
@ServeurpersoCom I believe you are correct. When I made this change that flag and code block did not exist. Let me refactor this PR to properly handle the flags.

jpm-canonical:
Looking into this now, we see #21056 introduced the `SO_REUSEADDR` socket option ([see this line](https://github.com/ggml-org/llama.cpp/pull/21056/changes#diff-5260e294d6ba344ce5e7a6408ad8daa545457ed1ccbdbae8b3f12e4421a8dd73R120)). Earlier in this PR we discussed it and decided that is not a good approach, and rather use the `SO_LINGER` option.

Because `SO_REUSEADDR` is currently [being set on the master branch](https://github.com/ggml-org/llama.cpp/blob/1e796eb41fb51950ada45811a303e57a5f4ea974/tools/server/server-http.cpp#L118), and that also solves the underlying issue related to immediate reuse of the socket after llama-server exits, this PR becomes unnecessary. I'll update this branch to add the `SO_LINGER` in the correct place, but then I will close this PR if I can prove the original issue is solved.

jpm-canonical:
I tested with both this branch as well as b8808 from the master branch. The original problem is solved, because #21056 intriduced `SO_REUSEADDR`.

PR #19768, opened 2026-02-20
vulkan: check for memory overlap before doing fusion

jeffbolznv:
This fixes a class of potential fusion bugs where the destination could overwrite a source tensor while other elements of the same op still need those source values. Add some logic to compare the memory ranges and disable fusion if the bad case is detected. Some operations contribute to the destination in an elementwise fashion and can do a more relaxed check where exact overlap is allowed.

In practice, I see this disabling TOPK_MOE fusion in some models (gpt-oss, qwen3) when there's more than one row, and this does appear to be a latent bug.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -13824,6 +13824,18 @@ static uint32_t ggml_vk_fuse_multi_add(ggml_backend_vk_context * ctx, const stru
     return num_adds;
 }
 
+static int32_t find_first_set(uint32_t x) {
+    int32_t ret = 0;
+    if (!x) {
+        return -1;
+    }
+    while (!(x & 1)) {
+        x >>= 1;
+        ret++;
+    }
+    return ret;
+}
+
 static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cgraph * cgraph) {
     VK_LOG_DEBUG("ggml_backend_vk_graph_compute(" << cgraph->n_nodes << " nodes)");
     ggml_backend_vk_context * ctx = (ggml_backend_vk_context *)backend->context;
@@ -14056,7 +14068,7 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
             ggml_tensor *output_nodes[2] {};
             output_nodes[0] = cgraph->nodes[i + ctx->num_additional_fused_ops];
             if (bits) {
-                int output_idx = (uint32_t)ceilf(log2f(float(bits)));
+                int output_idx = find_first_set(bits);
                 GGML_ASSERT(bits == (1u << output_idx));
                 output_nodes[1] = cgraph->nodes[i + output_idx];
             }
@@ -14089,6 +14101,7 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
                                 for (int n = 0; n < k; ++n) {
                                     if (cgraph->nodes[i + n] == src) {
                                         found = true;
+                                        break;
                                     }
                                 }
                                 if (!found) {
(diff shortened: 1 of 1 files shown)


am17an:
This would probably also exist in CUDA as well if I understand correctly. Is there a test case we can add to reproduce this?

jeffbolznv:
I'm not sure how to add a backend test for this, since it depends on the allocator behavior. Maybe there's a way to allocate a tensor and then have the operations go into views of that tensor and I can control the layout? Any ideas?

am17an:
That might not be easy to do but in any case can this check be a part of `ggml_can_fuse_subgraph`?

jeffbolznv:
I don't think this would be easy to fold into can_fuse. The memory range check would be different for each backend, and the analysis of what's ok to overlap ended up being kind of manual (some special cases).

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -14003,18 +14033,80 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
...
 
+        // Check whether fusion would overwrite src operands while they're still in use.
+        // If so, disable fusion.
+        if (ctx->num_additional_fused_ops) {
+            // There are up to two output nodes - topk_moe has two.
+            uint32_t bits = ctx->fused_ops_write_mask & ~(1 << ctx->num_additional_fused_ops);
+            ggml_tensor *output_nodes[2] {};
+            output_nodes[0] = cgraph->nodes[i + ctx->num_additional_fused_ops];
+            if (bits) {
+                int output_idx = (uint32_t)ceilf(log2f(float(bits)));
```
0cc4m: Wouldn't it be safer to just count the bit position here? In C++20 we get `countr_zero`, but we can't use that here. We can binary-search the bit or use a [De Bruijn Sequence](https://en.wikipedia.org/wiki/De_Bruijn_sequence#Finding_least-_or_most-significant_set_bit_in_a_word). All of this takes more lines of code though. If you prefer to keep the float log2f way, that's okay with me.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -14003,18 +14033,80 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
...
+                        for (uint32_t s = 0; s < GGML_MAX_SRC; ++s) {
+                            ggml_tensor *src = cgraph->nodes[i + k]->src[s];
+                            if (!src || src->op == GGML_OP_NONE) {
+                                continue;
+                            }
+                            if (ggml_vk_tensors_overlap(src, dst, op_srcs_fused_elementwise[k])) {
+                                bool found = false;
+                                for (int n = 0; n < k; ++n) {
+                                    if (cgraph->nodes[i + n] == src) {
+                                        found = true;
```
0cc4m, line 14103: Why not break early here?

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -13710,12 +13710,11 @@ static bool ggml_vk_can_fuse_rope_set_rows(ggml_backend_vk_context * ctx, const
     return true;
 }
 
-// Check whether the tensors overlap in memory but are not equal.
+// Check whether the tensors overlap in memory.
 // Fusions can potenitally overwrite src tensors in ways that are not prevented
```
0cc4m:
```suggestion
// Fusions can potentially overwrite src tensors in ways that are not prevented
```

ggml/src/ggml-vulkan/ggml-vulkan.cpp
jeffbolznv: I added a find_first_set function.

0cc4m (approved):

PR #19771, opened 2026-02-20
Document custom default webui preferences in server readme

woof-dog:
The llama.cpp server `--webui-config`/`--webui-config-file` settings are excellent, but knowing how to use them currently requires you to read the webui code or local storage in your browser.

This PR adds a brief example and a link to available preferences to the server readme.

### DIFF
--- a/tools/server/README.md
+++ b/tools/server/README.md
@@ -1748,6 +1748,16 @@ Apart from error types supported by OAI, we also have custom types that are spec
 }
 ```
 
+### Custom default Web UI preferences
+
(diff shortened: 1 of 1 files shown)


ServeurpersoCom (approved):
LGTM. Note: not all settings from settings-config.ts are syncable yet (see SYNCABLE_PARAMETERS in parameter-sync.service.ts), but we plan to make them all syncable soon, so the doc is fine as-is.

ServeurpersoCom:
It's still awaiting merge, and we need to finish this associated feature that clarifies the source of truth for the WebUI defaults

allozaur (approved):

PR #19772, opened 2026-02-20
Clean up per-thread parameter buffer pool and job submission logic

nikhilJain17:
After splitting per-thread state and execution, this is the final cleanup diff.

We allow the buffer pool to grow in case of multiple kernels in a command requiring more buffers, remove the inflight_threads logic, and replace it with num_kernels to decide when to submit a batch of commands.


reeselevine (commented):
Thanks, this looks good, and I think the heuristic for max param buffer pool size is reasonable.

reeselevine (approved):

PR #19773, opened 2026-02-21
server : merge contiguous Responses input items into a single assistant message

### LINKED ISSUE #19513: Eval bug: Qwen3-Coder-Next generates prematurely EOS instead of tool call(/continued response)

aldehir:
The Responses API endpoint constructs separate chat completion messages for each input item, except for reasoning. This causes problems with many chat templates that expect content, reasoning, and tool calls to appear in a single assistant message.

This PR merges contiguous assistant inputs into a single message before passing them to templates.

This also preserves reasoning content that isn't coupled with a tool call. A few models, such as Ministral 3, support interleaved reasoning within regular messages. Models that don't typically handle pruning the reasoning in their own templates.

ref: https://github.com/ggml-org/llama.cpp/pull/19765#issuecomment-3937342435
fixes #19513

cc @bfroemel

### DIFF
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -1195,7 +1195,7 @@ json convert_responses_to_chatcmpl(const json & response_body) {
                 item.at("type") == "message"
             ) {
                 // #responses_create-input-input_item_list-item-output_message
-                std::vector<json> chatcmpl_content;
+                auto chatcmpl_content = json::array();
 
                 for (const auto & output_text : item.at("content")) {
                     const std::string type = json_value(output_text, "type", std::string());
@@ -1218,7 +1218,7 @@ json convert_responses_to_chatcmpl(const json & response_body) {
                         prev_msg["content"] = json::array();
                     }
                     auto & prev_content = prev_msg["content"];
-                    prev_content.insert(prev_content.end(), chatcmpl_content);
+                    prev_content.insert(prev_content.end(), chatcmpl_content.begin(), chatcmpl_content.end());
                 } else {
                     item.erase("status");
                     item.erase("type");
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -1104,12 +1104,9 @@ json convert_responses_to_chatcmpl(const json & response_body) {
             return j.contains(key) && j.at(key).is_string();
         };
 
-        // Return true if we should merge with the last message
-        auto should_merge = [&chatcmpl_messages]() -> bool {
-            return !chatcmpl_messages.empty() && chatcmpl_messages.back().value("role", "") == "assistant";
-        };
-
         for (json item : input_value) {
+            bool merge_prev = !chatcmpl_messages.empty() && chatcmpl_messages.back().value("role", "") == "assistant";
+
             if (exists_and_is_string(item, "content")) {
                 // #responses_create-input-input_item_list-input_message-content-text_input
                 // Only "Input message" contains item["content"]::string
@@ -1215,9 +1212,13 @@ json convert_responses_to_chatcmpl(const json & response_body) {
                     });
                 }
 
-                if (should_merge()) {
+                if (merge_prev) {
                     auto & prev_msg = chatcmpl_messages.back();
-                    prev_msg["content"] = chatcmpl_content;
+                    if (!exists_and_is_array(prev_msg, "content")) {
+                        prev_msg["content"] = json::array();
+                    }
+                    auto & prev_content = prev_msg["content"];
+                    prev_content.insert(prev_content.end(), chatcmpl_content);
                 } else {
                     item.erase("status");
                     item.erase("type");
@@ -1240,7 +1241,7 @@ json convert_responses_to_chatcmpl(const json & response_body) {
                     {"type", "function"},
                 };
 
-                if (should_merge()) {
+                if (merge_prev) {
                     auto & prev_msg = chatcmpl_messages.back();
                     if (!exists_and_is_array(prev_msg, "tool_calls")) {
                         prev_msg["tool_calls"] = json::array();
@@ -1295,7 +1296,7 @@ json convert_responses_to_chatcmpl(const json & response_body) {
                     throw std::invalid_argument("item['content']['text'] is not a string");
                 }
 
-                if (should_merge()) {
+                if (merge_prev) {
                     auto & prev_msg = chatcmpl_messages.back();
                     prev_msg["reasoning_content"] = item.at("content")[0].at("text");
                 } else {
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -1249,14 +1249,7 @@ json convert_responses_to_chatcmpl(const json & response_body) {
                 } else {
                     chatcmpl_messages.push_back(json {
                         {"role",       "assistant"},
-                        {"tool_calls", json {
-                            {"function", json {
-                                {"arguments", item.at("arguments")},
-                                {"name",      item.at("name")},
-                            }},
-                            {"id",   item.at("call_id")},
-                            {"type", "function"},
-                        }}
+                        {"tool_calls", json::array({tool_call})}
                     });
                 }
             } else if (exists_and_is_string(item, "call_id") &&
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -1104,6 +1104,11 @@ json convert_responses_to_chatcmpl(const json & response_body) {
             return j.contains(key) && j.at(key).is_string();
         };
 
+        // Return true if we should merge with the last message
+        auto should_merge = [&chatcmpl_messages]() -> bool {
+            return !chatcmpl_messages.empty() && chatcmpl_messages.back().value("role", "") == "assistant";
+        };
+
         for (json item : input_value) {
             if (exists_and_is_string(item, "content")) {
                 // #responses_create-input-input_item_list-input_message-content-text_input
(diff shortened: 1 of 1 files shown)


tools/server/server-common.cpp
```diff
@@ -1210,35 +1215,43 @@ json convert_responses_to_chatcmpl(const json & response_body) {
                     });
                 }
 
-                item.erase("status");
-                item.erase("type");
-                item["content"] = chatcmpl_content;
-                chatcmpl_messages.push_back(item);
+                if (should_merge()) {
+                    auto & prev_msg = chatcmpl_messages.back();
+                    prev_msg["content"] = chatcmpl_content;
```
pwilkin: `=` or `+=`? Any idea if we can have successive blocks of pure text? (or text/reasoning without tool calls, for that matter)

aldehir: I can add it just in case, but I think it's unlikely to have multiple. `content` is a JSON array, constructed from an array of content objects from a single input item.

bfroemel:
Tested with codex & responses API:
- gpt-oss-120b: noticed no change of behavior; prompt is rendered exactly the same for the case of an assistant `output_text` element preceding a `function_call` element in the input array
- qwen3-coder-next, ud_q4_k_xl: no premature end of agent turns; fixes #19513, prompt renders with tool preamble messages ending with two newlines and the desired merging:

For example:
```
<|im_start|>assistant\nNow let me look at how the `needs_follow_up` is determined:\n\n<tool_call>\n<function=exec_command>\n<parameter=cmd>\ngrep -n \"needs_follow_up\" /home/b/work/codex-new/codex-rs/core/src/codex.rs | head -30\n</parameter>\n<parameter=justification>\nFind needs_follow_up handling\n</parameter>\n</function>\n</tool_call><|im_end|>
```


Only a minor discrepancy remains with respect to openrouter; the generated two newlines of tool call preambles are not part of the `output_text` content element in the response, e.g., on openrouter you really get back something like that:

```
"output":[{"role":"assistant","type":"message","status":"incomplete","content":[{"type":"output_text","text":"Now let me look at `handle_output_item_done` which processes output items and determines if follow-up is needed:\n\n","annotations":[],"logprobs":[]}],"id":"msg_tmp_vg3tku6apmm"},{"type":"function_call","call_id":"call_f16c662f06654160a2f0af2f","name":"exec_command","arguments":"{\"cmd\": \"grep -n \\\"handle_output_item_done\\\" /home/b/work/codex-new/codex-rs/core/src/codex.rs\", \"justification\": \"Find handle_output_item_done\"}","id":"call_f16c662f06654160a2f0af2f","status":"completed"}]
```

Thanks for the diligence/pushing back!! Too soon to tell, but model quality might have improved/feels a bit better than the non-merging fix where we just not trimmed assistant content (/preserved the generated newlines).


btw: just noticed that parallel tool calls also seem to work (https://github.com/ggml-org/llama.cpp/pull/19765 ); very nice! :)

```
 curl http://<llama.cpp host>/v1/responses -d '{
    "model": "current",
    "instructions": "Issue multiple tool calls in a single turn if possible to reduce latency and improve user experience!",
    "input": [
      {
        "type": "message",
        "role": "user",
        "content": [ {"type": "input_text",
        "text": "what\'s the weather in Paris and London?"}]
      }
    ],
    "tools": [
      {
        "type": "function",
        "name": "get_weather",
        "description": "Returns the weather of location.",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string"
            }
          },
          "required": ["location"],
          "additionalProperties": false
        }
      }
    ],
    "stream": false,
    "parallel_tool_calls": true
  }
  '
{"completed_at":1771668166,"created_at":1771668166,"id":"resp_HJZ7SNdrqK12w6i4TUJIqHkWDoe0BlgO","model":"qwen3-coder-next","object":"response","output":[{"type":"function_call","status":"completed","arguments":"{\"location\":\"Paris\"}","call_id":"fc_6ZZb88GAiND0rfZflEv52G8yqAHESV48","name":"get_weather"},{"type":"function_call","status":"completed","arguments":"{\"location\":\"London\"}","call_id":"fc_uzAq3J1ikJcxlu1YG21AcELop1Z4JNGd","name":"get_weather"}],"status":"completed","usage":{"input_tokens":291,"output_tokens":44,"total_tokens":335}}
```

Mushoz:
@bfroemel did you try to see if reasoning content is correctly working and being retained for non-gpt-oss models with interleaved thinking?

bfroemel:
@Mushoz no; only tested what I currently use. I kind of settled with codex for now and didn't look a lot at other agents in the past couple of months.

Do we have other non-gpt-oss models that work (well) with a responses API client like codex? Do we even have other mature responses API coding agents by now? :)

~When looking at the code this PR might fix another subtle issue ( https://github.com/ggml-org/llama.cpp/pull/19773/changes#diff-562aee47dd99ff76cbd0c2a3e9b98d30cdf4d0111c0f09f345130e1f096d7ef4L1299 ) as it won't remove any reasoning content anymore. btw: regarding gpt-oss models, the template discards reasoning content only, if there is a final message (one without tool calls) generated in a assistant turn (otherwise the interleaved reasoning is kept).~ /edit: removal of that reasoning content wasn't relevant for gpt-oss, because the prompt ignores messages that only has reasoning content without tool calls. 

Overall, I believe the responsibility remains with the client to include reasoning content in the request, if required by the model, otherwise the client can leave it out. llama.cpp won't remove it (anymore with this PR now regardless whether the reasoning content is followed by tool calls or not).

Uhm and just to make sure: This PR introduces no changes for the chat completions API - there everything stays the same.

pwilkin (approved):

aldehir:
GPT-OSS will render the same because their template format puts every element into a separate "message". It is a 1-to-1 mapping to Responses input items, which makes sense that OpenAI designed it that way.

The difference from OpenRouter should now be cosmetic, depending on if the client trims the content or not.

I appreciate you taking the time to test this out! Glad we got to the root of the issue.

ggerganov (approved):

PR #19783, opened 2026-02-21
cli : provide model with text filename

CISC:
Provide model with text filename when using `/read`, helps it contextualize the contents, use the `FIM_FILE_SEP` token if it exists.

Tested with various models with and without `FIM_SEP`, all acknowledging file name and type inferred from extension and path.

Also adds `FIM_PAD` and `FIM_REPO` tokens to server context meta for future use.

### DIFF
--- a/tools/cli/cli.cpp
+++ b/tools/cli/cli.cpp
@@ -380,6 +380,15 @@ int main(int argc, char ** argv) {
                 console::error("file does not exist or cannot be opened: '%s'\n", fname.c_str());
                 continue;
             }
+            if (inf.fim_sep_token != LLAMA_TOKEN_NULL) {
+                cur_msg += common_token_to_piece(ctx_cli.ctx_server.get_llama_context(), inf.fim_sep_token, true);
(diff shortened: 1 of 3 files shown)


danbev (approved):

ngxson (approved):
This is OK as a quick fix, but I think we should move all FIM models to using jinja template (we should invent one that can fit all FIM models)

If you look at server code, there is already a function to apply FIM "template"

CISC:
> This is OK as a quick fix, but I think we should move all FIM models to using jinja template (we should invent one that can fit all FIM models)

Normal FIM is simple enough, although it's impossible to tell if a model prefers SPM or PSM. Maybe this is something we should endeavor to define in metadata, or even template form? Would be nice if we could make a push towards this together with Hugging Face.

> If you look at server code, there is already a function to apply FIM "template"

Do you mean the FIM completion or something else?

This isn't strictly FIM though, it's just that the models that have this token more readily recognize files separated with this as they are trained with it for full repo dumps.

PR #19785, opened 2026-02-21
jinja: correct stats for tojson and string filters

ngxson:
Target fix https://github.com/ggml-org/llama.cpp/pull/18675

@pwilkin please give this a try (see the added test case for more info)


pwilkin:
@ngxson Cherry picked in #18675 and removed the workaround, can confirm it works correctly on all templates that in fact have tool definitions (found that Apertus had an incorrect macro for rendering tools, after correcting to OpenAI-compatible format it works as well).

ngxson:
@pwilkin hmm ok, if this works for you, please approve this PR so I can merge & proceed to review yours

pwilkin (approved):

PR #19790, opened 2026-02-21
vulkan: fix data race in mul_mat_id shader

jeffbolznv:
I've been working on automated data race detection (see https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/11717), and it found a data race in the mul_mat_id shaders. All invocations in a subgroup were storing the same value to shared memory, but this is still technically a data race. Just store on the first invocation.


0cc4m (approved):

0cc4m:
That's a good tool to have, great work.

0cc4m (approved):

PR #19793, opened 2026-02-22
vulkan: fix coopmat1 without bf16 support

jeffbolznv:
This should fix the CI failure on lavapipe. lavapipe added coopmat1 support recently, but does not have bf16 support, so it falls back to the scalar path. This fallback didn't have quite the same tile size logic for subgroupsize=8 as when going through the scalar path directly.


0cc4m (approved):
Thanks for the fix. The addition of coopmat to llvmpipe raises the question whether we want to use it in the CI. I think it makes more sense for it to go through the scalar path, since it's the "baseline" CI run.

PR #19797, opened 2026-02-22
common : add more aliases for sampler CLI params

ddh0:
Adds two CLI argument aliases for sampler parameters:
- `--top-n-sigma` (for existing `--top-nsigma`)
- `--temperature` (for existing `--temp`)


common/arg.cpp
```diff
@@ -1611,7 +1611,7 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_sparam());
     add_opt(common_arg(
-        {"--top-nsigma"}, "N",
+        {"--top-n-sigma", "--top-nsigma"}, "N",
```
danbev: The shortest option need to come first. This is check by CI and is the reason for the most of the failures there:
```console
test-arg-parser: shorter argument should come before longer one: --top-n-sigma, --top-nsigma
```
```suggestion
        {"--top-nsigma", "--top-n-sigma"}, "N",
```

common/arg.cpp
```diff
@@ -1634,7 +1634,7 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_sparam());
     add_opt(common_arg(
-        {"--typical"}, "N",
+        {"--typical-p", "--typical"}, "N",
```
danbev:
```suggestion
        {"--typical", "--typical-p"}, "N",
```

danbev (approved):

ddh0:
Whoops, thanks!

Issue #19795, opened 2026-02-22
Eval bug: Trailing quotation marks dropped from chat completion response on Kimi K2.5 and Minimax M2.5


DocShotgun:
### Name and Version

llama-server
version: 8124 (35715657c)
built with GNU 14.2.0 for Linux x86_64

### Operating systems

Linux

### GGML backends

CUDA, CPU

### Hardware

Dual 5th gen Xeon with 768gb DDR5 + RTX Pro 6000

### Models

Kimi K2.5 (q8_0 + q4_0 mix)
Minimax M2.5 (q8_0 + q4_k + q4_k + q5_k mix)

Of note, I could **NOT** reproduce this on Qwen 3.5 (q8_0 + q4_k + q4_k + q5_k mix)

Perhaps there is some issue common to the chat parsers for Kimi K2.5 and Minimax M2.5 but not Qwen 3.5?

### Problem description & steps to reproduce

Alrighty, this is a weird one.

I noticed that when using Kimi K2.5 via llama-server's chat completions, the model always omits the final `"` if the response ends on a `"`. However, if I manually format the prompt (by rendering the jinja template) and send the request as a raw request to `/v1/completions`, this doesn't happen. @ddh0 also encountered this issue on Minimax M2.5.

To reproduce, hit the endpoint with a curl request:
```
curl http://0.0.0.0:8080/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer AINT_GOT_NO_API_KEY" \
  -d '{
    "messages": [
      {
        "role": "user",
        "content": "Write the word test wrapped in quotes."
      }
    ],
    "top_k": 1,
    "chat_template_kwargs": {"thinking": false}
  }'
```

and then here we get:

```
[...]

PR #19801, opened 2026-02-22
fix(chat-parser): only trim partial content in XML tool-call parser + add regression test

julio75012:
### Description

This PR fixes a subtle trimming bug in the XML tool-call parser that caused the final `>` character to be lost when the model outputs **pure custom XML** (e.g. strict prompt-engineering formats like `<prompt_engineering_answer_*>` tags with `--ignore-eos --n-predict -1`).

**Root cause**  
In `parse_msg_with_xml_tool_calls`, the code was doing:

```cpp
if (builder.pos() == builder.input().size() && all_space(content)) {
```

This condition was triggering even for **full, non-partial** responses, incorrectly trimming the final character when the output ended exactly with `>` and no tool-call structure was matched.

**Changes**
- Added `&& builder.is_partial()` guard to both trimming sites in `chat-parser-xml-toolcall.cpp` (only trim in streaming/partial mode).
- Added regression test `test_seed_oss_closing_tag()` to ensure strings ending with `>` are preserved.

**Reproduction (for reviewers)**

```bash
# 1. Build Debug
cd llama.cpp
rm -rf build
cmake -B build
cmake --build build -j$(nproc)

# 2. Start server
./build/bin/llama-server \
  -m /path/to/Qwen3-Coder-30B-A3B-Instruct-UD-Q8_K_XL.gguf \
  --host 0.0.0.0 --port 8080 -ngl 999 \
  --api-key sk-foobar \
  --ctx-size 262143 --parallel 2 --batch-size 1024 \
  --verbose --ignore-eos --n-predict -1
```

In another terminal, with [payload.json](https://github.com/user-attachments/files/25467261/payload.json)

```bash

curl -X POST http://localhost:8080/v1/chat/completions \
  -H "Authorization: Bearer sk-loesho7CofengohL" \
  -d @payload.json --silent --output response.json

# Check the end of the response
jq -r '.choices[0].message.content' response.json | tail -c 100
```

**Before this PR** → ends with `pt_engineering_answer_comments` (missing `>`)

**After this PR** → ends with `</prompt_engineering_answer_comments>` (full tag preserved)

**Verification (copy-paste proof)**

On `master` (bug present):
```bash
git checkout master
git checkout fix/chat-parser-dangling-input -- tests/test-chat-parser.cpp
cd build && make -j$(nproc) test-chat-parser
./bin/test-chat-parser
```
→ Test fails (`Expected: 1 Actual: 0`)

On this branch (fixed):
```bash
git checkout fix/chat-parser-dangling-input
cd build && make -j$(nproc) test-chat-parser
./bin/test-chat-parser
```
→ `All tests passed!`

**Testing**
- Verified with the exact user case (pure custom XML + `--ignore-eos --n-predict -1`).
- Full test suite passes.
- No behaviour change for normal chat/tool/reasoning or streaming responses.

This is a minimal, targeted fix for a real edge case affecting custom XML output.

First-time contributor — happy to address any feedback.

### DIFF
--- a/common/chat-parser-xml-toolcall.cpp
+++ b/common/chat-parser-xml-toolcall.cpp
@@ -737,7 +737,7 @@ inline void parse_msg_with_xml_tool_calls(common_chat_msg_parser & builder, cons
                     reasoning_content = content.substr(0, pos);
                     content.erase(0, pos + end_think.size());
                 }
-                if (builder.pos() == builder.input().size() && all_space(content)) {
+                if (builder.pos() == builder.input().size() && all_space(content) && builder.is_partial()) {
                     rstrip(reasoning_content);
                     trim_potential_partial_word(reasoning_content);
                     rstrip(reasoning_content);
@@ -802,8 +802,8 @@ inline void parse_msg_with_xml_tool_calls(common_chat_msg_parser & builder, cons
             }
         }
 
-        // remove potential partial suffix
-        if (builder.pos() == builder.input().size()) {
+        // remove potential partial suffix (only for partial/streaming parsing)
+        if (builder.pos() == builder.input().size() && builder.is_partial()) {
             if (unclosed_reasoning_content.empty()) {
                 rstrip(content);
                 trim_potential_partial_word(content);
--- a/common/chat-parser.cpp
+++ b/common/chat-parser.cpp
@@ -129,14 +129,14 @@ static void parse_json_tool_calls(
     }
 }
 
-common_chat_msg_parser::common_chat_msg_parser(const std::string & input, bool is_partial, const common_chat_parser_params & syntax)
+common_chat_msg_parser::common_chat_msg_parser(std::string input, bool is_partial, const common_chat_parser_params & syntax)
     : input_(input), is_partial_(is_partial), syntax_(syntax)
 {
     result_.role = "assistant";
 
     while (true) {
         std::string id = std::to_string(std::rand());
-        if (input.find(id) == std::string::npos) {
+        if (input_.find(id) == std::string::npos) {
             healing_marker_ = id;
             break;
         }
--- a/common/chat-parser.cpp
+++ b/common/chat-parser.cpp
@@ -129,14 +129,14 @@ static void parse_json_tool_calls(
     }
 }
 
-common_chat_msg_parser::common_chat_msg_parser(std::string input, bool is_partial, const common_chat_parser_params & syntax)
+common_chat_msg_parser::common_chat_msg_parser(const std::string & input, bool is_partial, const common_chat_parser_params & syntax)
     : input_(input), is_partial_(is_partial), syntax_(syntax)
 {
     result_.role = "assistant";
 
     while (true) {
         std::string id = std::to_string(std::rand());
-        if (input_.find(id) == std::string::npos) {
+        if (input.find(id) == std::string::npos) {
             healing_marker_ = id;
             break;
         }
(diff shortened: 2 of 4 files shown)


aldehir:
The Qwen3-Coder parser on master no longer uses this XML parser. Are you saying the final `>` is omitted with the master branch?

julio75012:
normally - yes

but let me run the test of this branch on the master branch to put this behavior in evidence simply. hold on

julio75012:
> The Qwen3-Coder parser on master no longer uses this XML parser. Are you saying the final `>` is omitted with the master branch?

I just adjusted the test for an easy proof (master: the test does not work, fix-branch: the test works)

### Proof for Reviewers (Before vs After)

I verified the fix with a clean rebuild and the regression test.

**On `master` (bug present):**

```bash
git checkout master
git checkout fix/chat-parser-dangling-input -- tests/test-chat-parser.cpp
cd build && make -j$(nproc) test-chat-parser
./bin/test-chat-parser
```

**Result on master:**
```
Expected: 1
Actual: 0
...
terminate called after throwing an instance of 'std::runtime_error'
  what():  Test failed
Aborted (core dumped)
```

**On the fix branch (bug fixed):**

```bash
git checkout fix/chat-parser-dangling-input
cd build && make -j$(nproc) test-chat-parser
./bin/test-chat-parser
```

**Result on fix branch:**
```
All tests passed!
```

The same reproduction with the real user payload (`payload.json` + curl to `/v1/chat/completions`) now returns the **complete** XML with the final `>` preserved.

julio75012:
I reverted some unecessary changes, the PR is now even simpler... @aldehir

julio75012:
ah yes indeed master now works on my case, I close the PR @aldehir

aldehir:
@julio75012 I managed to reproduce it on Seed-OSS, but not MiniMax. We recently changed the Qwen3-Coder parser, so that should be fixed. There are still other models that use the XML parser though, and it seems you're onto something. Perhaps could explain #19795.

I'll dig in a bit further. Thank you for bringing this up.

julio75012:
clearly, last week, the last ">" was always missing with my qwen3-coder which is not the case anymore, so there might have been a push in between.

The minimax test is not anymore relevant but happy if the Seed-OSS test or my changes contributes to the explain and fix #19795 . feel free to copy-past it in any other PR if that helps @aldehir

julio75012:
@aldehir : adding logs to print "input" and then "msg" in common/chat-parser.cpp in this function around line 1600 helped me to dig the bug 

```
common_chat_msg common_chat_parse(const std::string & input, bool is_partial, const common_chat_parser_params & syntax) {

```

it looks like this [change](https://github.com/ggml-org/llama.cpp/commit/94b0200a01a753eff5897dab9311f51a7bc1c62f) is solving the problem (for qwen3-coder)

julio75012:
also @aldehir , last but not least, before this [commit](https://github.com/ggml-org/llama.cpp/commit/94b0200a01a753eff5897dab9311f51a7bc1c62f):

 the answer of a simple curl request to qwen3-coder such as :
 ```
 user@localhost:~/Documents/debug-lamma$ curl -X POST http://localhost:8080/v1/chat/completions   -H "Content-Type: application/json"   -H "Authorization: Bearer sk-xxx"   -d @payload.json --silent --output response.json
```

was suppressing the ">" in 
```
user@localhost:~/debug$ jq -r '.' response.json
{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "role": "assistant",
        "content": 
 ```
 
 but not in 
 ```
   "id": "chatcmpl-93951RcbV6BElSxqvL9Glwpl2EiXhyfF",
  "__verbose": {
    "index": 0,
    "content": "
 ```
 
 when launching llama.cpp server with `--verbose`

PR #19378, opened 2026-02-05
ggml: backend-agnostic tensor parallelism (experimental)

JohannesGaessler:
This PR adds initial support for tensor parallelism, enabled via specifying `--split-mode tensor`. This should be considered as an experimental feature that is not yet production ready. In principle the implementation is backend-agnostic, in practice as of right now only the CUDA backend has received the necessary extensions and performance optimizations to make the performance better than `--split-mode layer` (in some cases).

The preexisting `--split-mode row` could already parallelize some matrix multiplications in the CUDA backend but this required a synchronization after every single operation. As a consequence the overhead is so large that it is only really worthwhile for old and slow GPUs like P40s where adding a bit of latency between operations makes relatively little difference to the overall runtime. The new implementation works by adding a new "meta" backend that internally wraps multiple conventional ggml backends. When given a compute graph the meta backend then automatically infers how the data is split based on the ggml compute graph and only schedules a synchronization at the necessary points. And the external interface for a meta backend is the same as for any other ggml backend. So in practice the meta backend allows ggml to use multiple GPUs in the same way as a single GPU. Importantly all of this is done at the ggml backend level and there are no hard dependencies for any extensions beyond what already exists on master (but without extensions the performance may be so bad that there is no point).

### Current Support Status

What currently works:

* Multiple CUDA GPUs work.
* For good performance, make sure that [NCCL](https://developer.nvidia.com/nccl) is installed.
* Performance should become comparatively better for deeper contexts where there is more work for each individual GPU.
* The "ROCm" backend works since it is just the CUDA code translated via HIP. On the hardware combinations that I have (RX 6800 + MI50 or RX 9060 XT + MI100) the performance is bad vs. the `-sm layer` baseline though.
* Any number of GPUs is supported (limited to an arbitrary maximum of 16 GPUs for practical reasons, ping me if you have a use case that needs a higher limit).
* The preexisting argument `--tensor-split` can be used to distribute arbitrary fractions of the data across GPUs.
* In my testing `--n-gpu-layers` was working correctly in combination with `-sm tensor` though I suspect that there are some scenarios where this combination is broken.
* Most models should work correctly with `--split-mode tensor`, for more details see below.

What currently doesn't work:

* There is currently an issue with a VRAM leak due to CUDA graphs being continually created but never destroyed (negligible for `-sm layer` but not for `-sm tensor`).
* Vulkan technically works at short contexts but the performance is bad, at long contexts there are also stability issues.
[...]

PR #19796, opened 2026-02-22
Add model metadata loading from huggingface for use with tests requiring real model data

bartowski1182:
This is based on the work from huggingface here:

https://github.com/huggingface/huggingface.js/tree/main/packages/gguf

Idea is to partially load GGUF models from huggingface, just enough to get the metadata

The intention is to use this data with realistic unit tests for llama-quant.cpp, but it can be used for anyone needing real model data

To build:

```
cmake --build build --target test-gguf-model-data
```

To run the included test:

```
./build/bin/test-gguf-model-data
=== test-gguf-model-data ===
gguf_fetch: downloading 2097152 bytes from Qwen3-0.6B-Q8_0.gguf
gguf_fetch: downloading 4194304 bytes from Qwen3-0.6B-Q8_0.gguf
gguf_fetch: downloading 8388608 bytes from Qwen3-0.6B-Q8_0.gguf
gguf_fetch: cache write OK (/home/colin/.cache/llama.cpp/gguf-headers//ggml-org_Qwen3-0.6B-GGUF--Qwen3-0.6B-Q8_0.gguf.partial, 8388608 bytes)
Architecture: qwen3
n_embd:       1024
n_ff:         3072
n_vocab:      151936
n_layer:      28
n_head:       16
n_head_kv:    8
n_expert:     0
n_embd_head_k:128
n_embd_head_v:128
tensors:      311
gguf_fetch: loaded from cache: /home/colin/.cache/llama.cpp/gguf-headers//ggml-org_Qwen3-0.6B-GGUF--Qwen3-0.6B-Q8_0.gguf.partial
gguf_fetch: downloading 2097152 bytes from GLM-4.6V-Q8_0-00001-of-00003.gguf
gguf_fetch: downloading 4194304 bytes from GLM-4.6V-Q8_0-00001-of-00003.gguf
gguf_fetch: downloading 8388608 bytes from GLM-4.6V-Q8_0-00001-of-00003.gguf
gguf_fetch: downloading 16777216 bytes from GLM-4.6V-Q8_0-00001-of-00003.gguf
gguf_fetch: cache write OK (/home/colin/.cache/llama.cpp/gguf-headers//ggml-org_GLM-4.6V-GGUF--GLM-4.6V-Q8_0-00001-of-00003.gguf.partial, 16777216 bytes)
gguf_fetch: split model with 3 shards, fetching remaining 2...
gguf_fetch: downloading 2097152 bytes from GLM-4.6V-Q8_0-00002-of-00003.gguf
[...]

PR #19802, opened 2026-02-22
llama: end-to-end tests

JohannesGaessler:
Edit: sorry, I misclicked and accidentally opened the PR without a description.

This PR aims to implement end-to-end tests for model inference comparable to `test-backend-ops` except by creating toy models with random data. This would allow us to assert that the llama.cpp code runs without crashing and that the results between different backends are consistent. For me the main use case will be quality control in the context of https://github.com/ggml-org/llama.cpp/pull/19378 to more easily test whether the meta backend works correctly for the model archs we support. We can also check whether a roundtrip with `llama_model_save_to_file` and `llama_model_load_from_file` works correctly. long-term, if the training code works reliably for all model archs we can also train small toy models to overfit on short texts and assert that they can correctly predict it afterwards.

The test code works as follows: the llama API is extended with a new function `llama_model_init` that accepts a GGUF context with model metadata as well as a user-defined function to set the data of an individual tensor. Internally this then hijacks `llama_model_loader` to initialize the model weights with the passed function rather than by loading the weights from disk. The new code in `test-llama-archs.cpp` is currently just a PoC - I intend to iterate over the values in `llama-arch.h` for the final tests - long-term I think it may make sense to expose enums such as `llm_arch` to facilitate easier model creation via the llama API.

@ggerganov @CISC if this PoC does not meet your requirements, please tell me now when I haven't yet invested much time.


src/llama-model-loader.cpp
```diff
@@ -521,15 +526,17 @@ llama_model_loader::llama_model_loader(
...
     struct ggml_context * ctx = NULL;
     struct gguf_init_params params = {
         /*.no_alloc = */ true,
         /*.ctx      = */ &ctx,
     };
 
-    meta.reset(gguf_init_from_file(fname.c_str(), params));
-    if (!meta) {
+    metadata_ptr.reset(gguf_init_from_file(fname.c_str(), params));
+    metadata = metadata_ptr.get();
```
ggerganov: The naming here is problematic - you actually want to do:

```suggestion
    this->metadata = metadata_ptr.get();
```

As it is, it sets the input argument instead.

include/llama.h
```diff
@@ -440,19 +441,28 @@ extern "C" {
...
 
+    typedef void (*llama_model_set_tensor_data_t)(struct ggml_tensor * tensor, void * userdata);
+
+    // Create a new model from GGUF metadata as well as a function to set the tensor data
+    LLAMA_API struct llama_model * llama_model_init(
+                    struct gguf_context * metadata,
+          llama_model_set_tensor_data_t   set_tensor_data,    // function to initialize tensor data with
+                                   void * set_tensor_data_ud, // userdata for function
+              struct llama_model_params   params);
+
```
ggerganov, line 454: Likely should be called `llama_model_init_from_user`

src/llama-model.cpp
```diff
@@ -2824,6 +2825,34 @@ bool llama_model::load_tensors(llama_model_loader & ml) {
...
+                    default:
+                        GGML_ABORT("fatl error");
+                }
+                ggml_backend_buffer_type_t buft = buft_list->at(0).second;
+                ggml_context * ctx = ctx_for_buft(buft);
+                ggml_tensor * ret = ggml_new_tensor(ctx, GGML_TYPE_F32, ne.size(), std::data(ne));
+                std::string name = tn;
+                ggml_set_name(ret, name.c_str());
+                return ret;
+            }
```
ggerganov: Ideally, this logic should not be needed. The data loading should be completely abstracted by `llama_model_loader` and the logic in `llama_model` should not need to handle different ways of tensor data input explicitly.

ggerganov (commented):
As a public API this is OK.

I think the internal implementation should be improved. I assume it's just a PoC for now, so that is OK.

My main recommendation is to:

- Encapsulate better the tensor loading logic in the `llama_model_loader` class
- Abstract different ways of loading tensor data and avoid leaking any of that logic into `llama_model`

If you have doubts how to do that, I can make a pass on this part of the code - LMK.

CISC:
It wasn't super clear, but is the thinking that we then create a set of tensors based on hardcoded tensor parameters for that model?

Is it perhaps possible to utilize the model loading stage to get these parameters OTF (or in a generation pass)?

JohannesGaessler:
I moved the code where changes would need to be made from `llama-model.cpp` into `llama-model-loader.cpp`. Thinking about how to test for correctness, what I think we should do is test for changes in the outputs instead. We can extend `test-llama-archs` to optionally write the generated model + the generated outputs to disk. We can then in another pass load the model from disk and compare the outputs vs. the outputs produced with a previous software version. With the minimal parameters I pushed the size of an individual model at FP32 precision seems to be ~30 kB.

>It wasn't super clear, but is the thinking that we then create a set of tensors based on hardcoded tensor parameters for that model? Is it perhaps possible to utilize the model loading stage to get these parameters OTF (or in a generation pass)?

I'm not sure I understand your question. My goal is to produce tiny models that can be evaluated with minimal resources for testing specific language model architectures. I'm currently hardcoding the metadata like `n_embd` that would normally be found in a GGUF file in `test-llama-archs.cpp` but it could easily be made configurable. It needs to be fully specified before creating the `llama_model` though. I've been thinking it would make sense to look at `test-backend-ops.cpp` and to factor out functionality that can be re-used into something like `test-common.h`.

CISC:
> > It wasn't super clear, but is the thinking that we then create a set of tensors based on hardcoded tensor parameters for that model? Is it perhaps possible to utilize the model loading stage to get these parameters OTF (or in a generation pass)?
> 
> I'm not sure I understand your question. My goal is to produce tiny models that can be evaluated with minimal resources for testing specific language model architectures. I'm currently hardcoding the metadata like `n_embd` that would normally be found in a GGUF file in `test-llama-archs.cpp` but it could easily be made configurable. It needs to be fully specified before creating the `llama_model` though. I've been thinking it would make sense to look at `test-backend-ops.cpp` and to factor out functionality that can be re-used into something like `test-common.h`.

Right, so what I was thinking is that this can be retrieved from the tensor loading stage of an arch.

JohannesGaessler:
If a model is already on disk or with https://github.com/ggml-org/llama.cpp/pull/19796 it could be done but for my usecase I think it wouldn't make sense. However, if we want to test for changes to the outputs of real models I would just write the new code in such a way that the following things can be done separately:

* Create and save a dummy model.
* Load a model from disk, calulcate outputs, write outputs to disk.
* Load a model and outputs from disk, calculate outputs, compare calculated outputs vs. disk.

JohannesGaessler:
I pushed a version that has rudimentary support for generating models as well as testing whether results have changed. The CLI usage looks something like this:

```bash
./build/bin/test-llama-archs gen-model --model model.gguf
./build/bin/test-llama-archs gen-results --model model.gguf --results results.gguf
./build/bin/test-llama-archs test-vs-disk --model model.gguf --results results.gguf

export mn=llama_3-8b && export q=q4_0
./build/bin/test-llama-archs gen-results --model models/opt/${mn}-${q}.gguf --results results.gguf
./build/bin/test-llama-archs test-vs-disk --model models/opt/${mn}-${q}.gguf --results results.gguf
```

I'm thinking it would make sense to write a script for automating `git bisect` using this test.

tests/test-llama-archs.cpp
```diff
@@ -0,0 +1,289 @@
...
+        ggml_set_name(&t, name.c_str());
+        gguf_add_tensor(metadata.get(), &t);
+    };
+
+    add_tensor("token_embd.weight", n_embd, n_vocab);
+
+    std::random_device rd;
+    std::mt19937 gen(rd());
+
+    llama_model_params params = llama_model_default_params();
```
ggerganov: On the Mac, I need to disable both mmap and repack:

```suggestion
    llama_model_params params = llama_model_default_params();
    params.use_mmap = false;
    params.use_extra_bufts = false;
```

I wonder, do you not have to do the same on your end? At least I think you also need to set `use_mmap = false` explicitly, otherwise the `set_tensor_data()` ends up being called on empty buffers?

JohannesGaessler: I myself did not encounter this issue during testing.

ggerganov: I see - it does not happen with CUDA enabled because we have `buffer_from_host_ptr_supported`.

I'll do a bit of refactoring in `llama-model-loader.cpp` to make this a bit more robust and will open PR to this branch if that's OK.

JohannesGaessler: Keep in mind that the way I'm currently constructing the models is WIP anyways. In principle the code should work out-of-the-box if all model tensors were explicitly added to the GGUF context that the tests are passing. However, that would require duplicating all of the logic that we have already implemented for checking the expected tensor shapes and just be very tedious in general. So I'm instead changing the code to create a tensor based on the expected shape.

src/llama.cpp
```diff
@@ -1019,6 +1027,16 @@ static struct llama_model * llama_model_load_from_file_impl(
...
 
+struct llama_model * llama_model_init(
+        struct gguf_context * metadata,
+        llama_model_set_tensor_data_t set_tensor_data,
+        void * set_tensor_data_ud,
+        struct llama_model_params params) {
+    GGML_ASSERT(metadata != nullptr);
+    std::string path_model;
+    std::vector<std::string> splits = {};
+    return llama_model_load_from_file_impl(metadata, set_tensor_data, set_tensor_data_ud, path_model, splits, params);
```
ggerganov, line 1040: Ok, for now this change is enough as a stopgap:

```suggestion
    std::string path_model;
    std::vector<std::string> splits = {};
    params.use_mmap = false;
    params.use_extra_bufts = false;
    return llama_model_load_from_file_impl(metadata, set_tensor_data, set_tensor_data_ud, path_model, splits, params);
```

Will do a more thorough refactor later in a separate PR.

JohannesGaessler:
I implemented for the new code to re-use the buffer type selection logic from the preexisting code by creating a tensor with the expected dimensions on-the-fly. This then causes the embedding tensor to remain on CPU vs. my previous iteration where I just hard-coded all tensors to go to one buffer type. So then I ran into the same issue you did and I took over the fix you posted. Please let me know if you make further changes to the related code but for a model created in-memory I think setting things like mmap to false is correct either way though.

ggerganov:
I think we should keep what you have now until we merge. After that, we should consider 2 major refactors in `libllama`:

- Abstract the `llama_model_loader` implementation to avoid branching such as `if (files.empty()) { } else { }`. Instead, these should be 2 separate implementation of the loader - one that reads data from files and one that uses user-provided data
- Expose API to extract the weight tensors per arch so that we can construct the GGUF metadata in the user code as you suggested

For now, we can accept the hacky way in order to unblock work that depends on the ability to generate small models.

JohannesGaessler:
>Expose API to extract the weight tensors per arch so that we can construct the GGUF metadata in the user code as you suggested

To be clear, if in the user code such tensors were to be generated they would as of right now **not** be used as the actual weight tensors of the model. The model loader generates a bunch of empty tensors from the GGUF file, the code in `llama-model.cpp` then provides expected tensor shapes, compares them against what the model loader found, and creates its own tensors in its own ggml contexts. Do you mean that the user explicitly creates weight tensors that a model should then directly take over as its weight tensors?

JohannesGaessler:
@ggerganov @CISC do you have an opinion on whether the current "modes" of `test-llama-archs` (other than the one that would iterate over all architectures) should be made into standalone binaries? I've been thinking that in order to properly automate git bisect the code should just have all of the functionality that we give regular examples/tools in order to properly set things like `-ot`.

CISC:
I have no particular opinion, but there's no code-sharing between the modes, so splitting them up would make sense.

JohannesGaessler:
I pushed a first version for systematically testing our supported model architectures with in-memory models. The test with CUDA vs. CPU is taking ~1.5 seconds with ~50% of the models supported. I made some KV retrievals optional if the inference code supports it.

I think there is a bug in the MBT code regarding how a view is calculated, I'll need to get a real model to check how the tensor shapes are actually supposed to look like. For LLAMA_EMBED CPU and CUDA yield inconsistent results.

JohannesGaessler:
I've pushed a version that tests >90% of our supported models and that provides a new binary `llama-results` to `--check` whether results have changed. From my end I will write a simple script for automatic git bisect, go over my changes one more time, and write a proper changelog. After that I think this PR will be ready for review.

JohannesGaessler:
From my end I would now consider this PR ready for review.

* I extended the llama API with a function `llama_model_init_from_user` to create a model from a GGUF context.
* I added a new binary `llama-test-archs` that tests whether toy models for a given `LLM_ARCH` produce consistent results across ggml backend devices. I re-used and extended `llama_model_saver` to populate the GGUF contexts.
* I added a new binary `llama-results` that writes a `--model`'s logits for a `--prompt` to a given `--output` GGUF file. Existing results can be `--check`ed against a re-run. The script under `scripts/git-bisect.sh` can be used to determine when a model's outputs have changed.
* The inputs for llama models are allocated via the `ggml_cgraph` allocation. However, if any of the llama graph inputs are not used (e.g. because a model has a hard-coded period of layer types) `llama_decode` will crash when trying to set the input. Very tedious to debug because it's difficult to determine from the error message why this is happening. I suspect that if we keep it like this it will result in a lot of wasted contributor time going forward. Not sure what would be the best way to fix this, for now I've added a comment to make debugging easier.
* While implementing the tests I modified some of the model loading code. Mostly I just made some hardcoded arguments configurable (like the number and size of layers) or I made them optional for all architectures (when they previously were for only a subset of architectures) if there is a meaningful fallback. For Kimi Linear I changed the logic for how to determine whether a layer is MLA or KDA (consistently use `hparams::is_recurrent`). With the code on master it is possible to construct an inconsistent Kimi Linear GGUF file that segfaults. There should be no change for the actual model.
* `llama_model_saver` is currently broken in the context of `llama_model_save_to_file`. For this reason the corresponding functionality in `test-llama-archs` to save the generated models to disk is disabled. I intend to fix this going forward.
* As Georgi said, going foward replace the conditional statements using `files.empty()` with template methods and expose things like our architectures and KV to user code.

src/models/baichuan.cpp
```diff
@@ -1,3 +1,4 @@
+#include "llama-model.h"
```
ggerganov: The `llama-model.h` header is not needed

src/llama-context.cpp
```diff
@@ -1114,6 +1114,7 @@ llm_graph_result * llama_context::process_ubatch(const llama_ubatch & ubatch, ll
     {
         //const auto t_start_us = ggml_time_us();
 
+        // FIXME this call causes a crash if any model inputs were not used in the graph and were therefore not allocated
         res->set_inputs(&ubatch);
```
ggerganov, line 1118: > The inputs for llama models are allocated via the ggml_cgraph allocation. However, if any of the llama graph inputs are not used (e.g. because a model has a hard-coded period of layer types) llama_decode will crash when trying to set the input.

Can you provide a repro so I can take a look into this later?

src/llama-model-loader.h
```diff
@@ -58,9 +63,10 @@ struct llama_model_loader {
         }
     };
 
-    static const int TENSOR_NOT_REQUIRED = 1 << 0;
-    static const int TENSOR_DUPLICATED   = 1 << 1;
-    static const int TENSOR_SKIP         = 1 << 2;
+    static const int TENSOR_NOT_REQUIRED    = 1 << 0;
+    static const int TENSOR_DUPLICATED      = 1 << 1;
+    static const int TENSOR_SKIP            = 1 << 2;
+    static const int TENSOR_SKIP_IF_VIRTUAL = 1 << 3;
```
ggerganov, line 69: What is the logic for using `TENSOR_SKIP_IF_VIRTUAL`?

ggerganov (approved):

ggerganov:
> llama_model_saver is currently broken in the context of llama_model_save_to_file. For this reason the corresponding functionality in test-llama-archs to save the generated models to disk is disabled. I intend to fix this going forward.

What is the issue with the saving?

src/llama-model-loader.h
JohannesGaessler: In order to not load certain tensors, usually because they cause some sort of problem:

* `create_tensor_gate_up_exps`: Prior to https://github.com/ggml-org/llama.cpp/pull/20092 this was causing a crash, I intend to rebase and then remove it there.
* MPT: Crashes if Q norm is created, I'm not sure whether this is an issue with the MPT evaluation code.
* T5: That tensor is unused so I don't create it, I think it caused a crash.
* Kimi Linear: create GGUF in the most up-to-date format.

src/llama-context.cpp
JohannesGaessler:
```diff
diff --git a/tests/test-llama-archs.cpp b/tests/test-llama-archs.cpp
index 9d3a0ceea..4baf588c6 100644
--- a/tests/test-llama-archs.cpp
+++ b/tests/test-llama-archs.cpp
@@ -85,7 +85,7 @@ static gguf_context_ptr get_gguf_ctx(const llm_arch arch, const bool moe) {
     uint32_t n_ff    = 384;
     uint32_t n_layer = 2;
     if (arch == LLM_ARCH_LLAMA4) {
-        n_layer = 4; // hparams.n_no_rope_layer_step is hard-coded to 4
+        n_layer = 2; // hparams.n_no_rope_layer_step is hard-coded to 4
     } else if (arch == LLM_ARCH_GEMMA3N) {
         n_embd = 64;
         n_head = 1;
@@ -409,6 +409,9 @@ static int test_backends(const size_t seed, const ggml_log_level log_level) {
     printf("|%15s|%30s|%6s|%8s|%6s|\n", "Model arch.", "Device", "Config", "NMSE", "Status");
     printf("|---------------|------------------------------|------|--------|------|\n");
     for (const llm_arch & arch : llm_arch_all()) {
+        if (arch != LLM_ARCH_LLAMA4) {
+            continue;
+        }
         if (arch == LLM_ARCH_CLIP || arch == LLM_ARCH_GPTJ || arch == LLM_ARCH_UNKNOWN) {
             continue; // These models don't have usable implementations.
         }
```

JohannesGaessler:
>What is the issue with the saving?

`add_tensors_from_model` crashes because of duplicate tensors, this was already an issue back when I originally implemented it and never properly resolved. Given how little the code is actually in use and that no one bothered to update it when adding new KV I also have very little confidence in the other method `add_kv_from_model` working correctly.

JohannesGaessler:
The `macOS-latest-cmake-arm64` CI failure seems to be due to a segmentation fault. I cannot reproduce this segfault locally and I don't know what would be causing it. @ggerganov can you reproduce the issue or at least tell me how to disable these tests in the CI for now?

ggerganov:
I am able to reproduce - looking into it.

ggerganov:
It will take me a bit of time to fix this - I think it is a problem in the Metal backend when it uses a Paravirtual Metal device (i.e. what the Github runner uses).

For now, you can workaround using this patch:

```diff
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index 30365a361..c219cc237 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,7 +93,7 @@ jobs:
         id: cmake_test
         run: |
           cd build
-          ctest -L main --verbose --timeout 900
+          ctest -L main -E "test-llama-arch" --verbose --timeout 900
 
   macOS-latest-cmake-x64:
     runs-on: macos-15-intel
```

I'll try to fix this properly later in a follow-up PR.

src/models/plamo2.cpp
```diff
@@ -1,6 +1,7 @@
 #include "models.h"
 
 #include "llama-memory-recurrent.h"
+#include <cstdint>
```
CISC:
```suggestion
```

src/llama-model.cpp
```diff
@@ -1704,10 +1567,10 @@ void llama_model::load_hparams(llama_model_loader & ml) {
         case LLM_ARCH_DEEPSEEK:
             {
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS, hparams.f_norm_rms_eps);
-                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,   hparams.n_layer_dense_lead);
+                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,   hparams.n_layer_dense_lead, false);
                 ml.get_key(LLM_KV_EXPERT_FEED_FORWARD_LENGTH,  hparams.n_ff_exp);
                 ml.get_key(LLM_KV_EXPERT_SHARED_COUNT,         hparams.n_expert_shared);
-                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,        hparams.expert_weights_scale);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,        hparams.expert_weights_scale, false);
```
CISC: Hmmm, this is actually unused (conversion hardcodes this to 1.0):
https://github.com/ggml-org/llama.cpp/blob/d6308b9dd073dbc3368de2b34aee4117534980fd/src/models/deepseek.cpp#L103

src/llama-model.cpp
```diff
@@ -1823,7 +1686,7 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                 ml.get_key(LLM_KV_EXPERT_USED_COUNT,           hparams.n_expert_used);
                 ml.get_key(LLM_KV_EXPERT_SHARED_COUNT,         hparams.n_expert_shared);
                 ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,   hparams.n_layer_dense_lead, false);
-                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,        hparams.expert_weights_scale);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,        hparams.expert_weights_scale, false);
```
CISC, line 1688: Needs `scale_w` to be conditional:
https://github.com/ggml-org/llama.cpp/blob/d6308b9dd073dbc3368de2b34aee4117534980fd/src/models/glm4-moe.cpp#L131

src/llama-model.cpp
```diff
@@ -2202,15 +2069,15 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
             } break;
         case LLM_ARCH_BAILINGMOE:
             {
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS, hparams.f_norm_rms_eps);
-                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,   hparams.n_layer_dense_lead);
+                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,   hparams.n_layer_dense_lead, false);
                 ml.get_key(LLM_KV_EXPERT_FEED_FORWARD_LENGTH,  hparams.n_ff_exp);
                 ml.get_key(LLM_KV_EXPERT_SHARED_COUNT,         hparams.n_expert_shared);
-                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,        hparams.expert_weights_scale);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,        hparams.expert_weights_scale, false);
```
CISC: Guess I copied this from DeepSeek, also unused. :)
https://github.com/ggml-org/llama.cpp/blob/d6308b9dd073dbc3368de2b34aee4117534980fd/src/models/bailingmoe.cpp#L100

src/llama-model.cpp
```diff
@@ -2222,11 +2089,11 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
             {
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS,       hparams.f_norm_rms_eps);
-                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,         hparams.n_layer_dense_lead);
+                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,         hparams.n_layer_dense_lead, false);
                 ml.get_key(LLM_KV_EXPERT_FEED_FORWARD_LENGTH,        hparams.n_ff_exp);
-                ml.get_key(LLM_KV_EXPERT_SHARED_FEED_FORWARD_LENGTH, hparams.n_ff_shexp);
+                ml.get_key(LLM_KV_EXPERT_SHARED_FEED_FORWARD_LENGTH, hparams.n_ff_shexp, false);
                 ml.get_key(LLM_KV_EXPERT_SHARED_COUNT,               hparams.n_expert_shared);
-                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,              hparams.expert_weights_scale);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,              hparams.expert_weights_scale, false);
```
CISC, line 2094: Needs conditional:
https://github.com/ggml-org/llama.cpp/blob/d6308b9dd073dbc3368de2b34aee4117534980fd/src/models/bailingmoe2.cpp#L93

src/llama-model.cpp
```diff
@@ -2245,10 +2112,10 @@ void llama_model::load_hparams(llama_model_loader & ml) {
         case LLM_ARCH_DOTS1:
             {
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS, hparams.f_norm_rms_eps);
-                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,   hparams.n_layer_dense_lead);
+                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,   hparams.n_layer_dense_lead, false);
                 ml.get_key(LLM_KV_EXPERT_FEED_FORWARD_LENGTH,  hparams.n_ff_exp);
                 ml.get_key(LLM_KV_EXPERT_SHARED_COUNT,         hparams.n_expert_shared);
-                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,        hparams.expert_weights_scale);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,        hparams.expert_weights_scale, false);
```
CISC, line 2116: Needs conditional:
https://github.com/ggml-org/llama.cpp/blob/d6308b9dd073dbc3368de2b34aee4117534980fd/src/models/dots1.cpp#L94

src/llama-model.cpp
```diff
@@ -2632,8 +2503,8 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                 // MoE parameters - Kimi uses moe_intermediate_size = 1024
                 ml.get_key(LLM_KV_EXPERT_FEED_FORWARD_LENGTH,        hparams.n_ff_exp);
                 ml.get_key(LLM_KV_EXPERT_SHARED_COUNT,               hparams.n_expert_shared);
-                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,         hparams.n_layer_dense_lead);
-                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,              hparams.expert_weights_scale);
+                ml.get_key(LLM_KV_LEADING_DENSE_BLOCK_COUNT,         hparams.n_layer_dense_lead, false);
+                ml.get_key(LLM_KV_EXPERT_WEIGHTS_SCALE,              hparams.expert_weights_scale, false);
```
CISC, line 2505: Needs conditional:
https://github.com/ggml-org/llama.cpp/blob/d6308b9dd073dbc3368de2b34aee4117534980fd/src/models/kimi-linear.cpp#L344

CISC (approved):

JohannesGaessler:
Regarding the expert weight scale: is there a reason why we have separate arguments for whether or not to scale the weights and also for the scale itself? It seems to me like we could just do something like `if (w_scale != 0.0f && scale_w != 1.0f) {...}`.

CISC:
> Regarding the expert weight scale: is there a reason why we have separate arguments for whether or not to scale the weights and also for the scale itself? It seems to me like we could just do something like `if (w_scale != 0.0f && scale_w != 1.0f) {...}`.

I'm guessing it's legacy carry-over, don't see why it couldn't just be a check like that in `build_moe_ffn`. @ggerganov

JohannesGaessler:
Anyways, thanks for zapping me regarding the scales. For now I've pushed a version that uses the scales conditionally. While checking the usage I noticed though that `NEMOTRON_H` and `EXAONE_MOE` are already suffering from this same issue on master. Which I would see as an argument in favor of not explicitly specifying `bool scale_w` in `llm_graph_context::build_moe_ffn` since it increases the risk of inconsistencies.

ggerganov:
Yes, the `bool scale_w` seems redundant - probably some legacy leftover. Add TODO to remove it.

CISC:
> Yes, the `bool scale_w` seems redundant - probably some legacy leftover. Add TODO to remove it.

I'll make a PR after this is merged.

CISC:
@jeffbolznv @0cc4m We seem to have some semi-random failures on Vulkan:
https://github.com/ggml-org/llama.cpp/actions/runs/22841132052/job/66247374061#step:9:8995
https://github.com/ggml-org/llama.cpp/actions/runs/22841132052/job/66247374061#step:9:9042
https://github.com/ggml-org/llama.cpp/actions/runs/22844897991/job/66258833638#step:3:1204

JohannesGaessler:
Without knowing anything about the implementation details of Vulkan, I think most likely this is the result of numerical instability.

jeffbolznv:
The failure reproduces reliably for me with coopmat2 disabled, e.g. using `test-llama-archs.exe -a gpt-oss`. Happens with the scalar or coopmat1 path, and goes away if I make FA fall back to CPU. @0cc4m can you look into this?

0cc4m:
I'll look into it, yes.

0cc4m:
It passes all models reliably for me on AMD 8060S with or without coopmat, and on Intel Meteor Lake. It also passes on llvmpipe without coopmat.

jeffbolznv:
@0cc4m are you saying you can't reproduce it at all? Not even on RTX 3090?

0cc4m:
Yes, I've now also tested on the RTX 3090 as well, couldn't reproduce it without coopmat, or with coopmat1 or 2. I haven't seen a single failure yet, except llvmpipe with coopmat.

jeffbolznv:
OK, I'll debug it a bit.

Issue #17917, opened 2025-12-10
Misc. bug: Performance regression using ROCm on Strix Halo


eugr:
### Name and Version

build: 4df6e859e (7349)

### Operating systems

Linux

### Which llama.cpp modules do you know to be affected?

llama-server, llama-bench

### Command line

```shell
build/bin/llama-bench -m ~/.cache/llama.cpp/ggml-org_gpt-oss-120b-GGUF_gpt-oss-120b-mxfp4-00001-of-00003.gguf -fa 1 -d 0,4096,8192,16384,32768 -p 2048 -n 32 -ub 2048 -mmp 0
```

### Problem description & steps to reproduce

This commit (as found by using git bisect) has introduced significant performance regression in prompt processing on Strix Halo and ROCm 7.x: [668ed765742065f82c2899e101ee4384d6669f11] HIP: enable WMMA-MMQ INT kernels for RDNA 3 (#17576)

I tried to compile with both rocWMMA and without with the same result.

Here are llama-bench results. Note that they have a wider gap in bench numbers, but I didn't run full benches when doing git bisect, the first row was usually enough.

Command:

```bash
build/bin/llama-bench -m ~/.cache/llama.cpp/ggml-org_gpt-oss-120b-GGUF_gpt-oss-120b-mxfp4-00001-of-00003.gguf -fa 1 -d 0,4096,8192,16384,32768 -p 2048 -n 32 -ub 2048 -mmp 0
```

Performance before:

| model                  |      size |   params | backend |            test |           t/s |
| ---------------------- | --------: | -------: | ------- | --------------: | ------------: |
| gpt-oss 120B MXFP4 MoE | 59.02 GiB | 116.83 B | ROCm    |          pp2048 | 900.29 ± 2.04 |
| gpt-oss 120B MXFP4 MoE | 59.02 GiB | 116.83 B | ROCm    |            tg32 |  52.28 ± 0.01 |
| gpt-oss 120B MXFP4 MoE | 59.02 GiB | 116.83 B | ROCm    |  pp2048 @ d4096 | 758.10 ± 3.08 |
| gpt-oss 120B MXFP4 MoE | 59.02 GiB | 116.83 B | ROCm    |    tg32 @ d4096 |  48.68 ± 0.01 |
| gpt-oss 120B MXFP4 MoE | 59.02 GiB | 116.83 B | ROCm    |  pp2048 @ d8192 | 651.92 ± 1.93 |
[...]

PR #19806, opened 2026-02-22
CUDA: add CDNA3 MFMA support for flash attention MMA kernel

Jayluci4:
## Summary

Adds MI300X (gfx942) MFMA tensor core flash attention to `fattn-mma-f16.cuh`. MI300X now routes to `BEST_FATTN_KERNEL_MMA_F16` instead of the tile-based fallback.

- Uses `v_mfma_f32_16x16x16_f16` (FP16 inputs, FP32 accumulate) with wavefront64
- Supports head sizes 64, 80, 96, 112, 128 via MMA; others fall back to VEC
- Single-token generation stays on VEC to avoid tg regression
- Manual V transpose load to match MFMA A-register layout
- Fixes Q loading and combine stride granularity for non-power-of-2 head dimensions (e.g. hsk=80)

## Benchmarks (Qwen2.5-1.5B Q4_K_M, MI300X)

| Test | No FA | MFMA FA | Speedup |
|------|-------|---------|---------|
| pp512 | 25,212 t/s | 27,057 t/s | +7.3% |
| pp1024 | 24,367 t/s | 27,453 t/s | +12.7% |
| pp2048 | 22,218 t/s | 27,360 t/s | +23.1% |
| pp4096 | 18,723 t/s | 26,092 t/s | +39.4% |
| tg128 | 431 t/s | 389 t/s | -9.7% (FA overhead, VEC used for both) |

## Test plan

- [x] All 2480 FLASH_ATTN_EXT backend-ops tests pass (0 failures)
- [x] Covers hsk=40, 64, 72, 80, 96, 112, 128, 256, 576
- [x] llama-bench pp/tg on MI300X shows no regressions
- [ ] CI build verification

Ref: #17917

### DIFF
--- a/ggml/src/ggml-cuda/fattn-mma-f16.cuh
+++ b/ggml/src/ggml-cuda/fattn-mma-f16.cuh
@@ -5,6 +5,14 @@
 
 using namespace ggml_cuda_mma;
 
+// MFMA tiles are designed for 64-thread wavefronts (ne = I*J/64), so the effective
+// warp size for the fattn MMA kernel must be 64 on CDNA to match the hardware wavefront.
+#if defined(AMD_MFMA_AVAILABLE)
+#define FATTN_WARP_SIZE 64
+#else
+#define FATTN_WARP_SIZE WARP_SIZE
+#endif
+
 // Config options for the MMA kernel.
 // Should not affect results, only speed/register pressure/shared memory use.
 struct fattn_mma_config {
@@ -111,12 +119,53 @@ static constexpr __host__ __device__ fattn_mma_config ggml_cuda_fattn_mma_get_co
     return ggml_cuda_fattn_mma_get_config_ampere(DKQ, DV, ncols);
 }
 
+static constexpr __host__ __device__ fattn_mma_config ggml_cuda_fattn_mma_get_config_cdna(const int DKQ, const int DV, const int ncols) {
+    // Conservative configs for CDNA (MI100+): 64KB LDS, wavefront64, nstages=1 (no cp.async).
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 64,  64,  8, 128, 2, 128,  32,  32,  32, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 64,  64, 16, 128, 2,  64,  32,  32,  32, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 64,  64, 32, 128, 2,  64,  32,  32,  32, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 64,  64, 64, 256, 2,  64,  32,  32,  32, 1, true);
+
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 80,  80,  8, 128, 2, 128,  40,  40,  40, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 80,  80, 16, 128, 2,  64,  40,  40,  40, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 80,  80, 32, 128, 2,  64,  40,  40,  40, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 80,  80, 64, 256, 2,  64,  40,  40,  40, 1, true);
+
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 96,  96,  8, 128, 2, 128,  48,  48,  48, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 96,  96, 16, 128, 2,  64,  48,  48,  48, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 96,  96, 32, 128, 2,  64,  48,  48,  48, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE( 96,  96, 64, 256, 2,  64,  48,  48,  48, 1, true);
+
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(112, 112,  8, 128, 2, 128,  56,  56,  56, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(112, 112, 16, 128, 2,  64,  56,  56,  56, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(112, 112, 32, 128, 2,  64,  56,  56,  56, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(112, 112, 64, 256, 2,  64,  56,  56,  56, 1, true);
+
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(128, 128,  8, 128, 2, 128,  64,  64,  64, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(128, 128, 16, 128, 2,  64,  64,  64,  64, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(128, 128, 32, 128, 2,  64,  64,  64,  64, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(128, 128, 64, 256, 2,  64,  64,  64,  64, 1, true);
+
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(256, 256,  8,  64, 4,  64, 128, 128, 128, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(256, 256, 16,  64, 4,  32, 128, 128, 128, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(256, 256, 32, 128, 2,  32, 128, 128, 128, 1, true);
+    GGML_CUDA_FATTN_MMA_CONFIG_CASE(256, 256, 64, 256, 2,  32, 128, 128, 128, 1, true);
+
+    // Fallback for unsupported DKQ values (e.g. 576). Must return non-zero values to satisfy
+    // compile-time static_asserts even though the kernel guard prevents runtime execution.
+    // nthreads=256 gives nwarps=8 (FATTN_WARP_SIZE=32), nbatch_fa=128 satisfies np*16 divisibility.
+    return fattn_mma_config(256, 1, 128, 4, 4, 4, 1, false);
+}
+
 static __host__ fattn_mma_config ggml_cuda_fattn_mma_get_config(const int DKQ, const int DV, const int ncols, const int cc) {
     if (ampere_mma_available(cc)) {
         return ggml_cuda_fattn_mma_get_config_ampere(DKQ, DV, ncols);
     }
     if (turing_mma_available(cc)) {
         return ggml_cuda_fattn_mma_get_config_turing(DKQ, DV, ncols);
+    }
+    if (amd_mfma_available(cc)) {
+        return ggml_cuda_fattn_mma_get_config_cdna(DKQ, DV, ncols);
     }
     if (amd_wmma_available(cc)) {
         return ggml_cuda_fattn_mma_get_config_rdna(DKQ, DV, ncols);
@@ -130,6 +179,8 @@ static constexpr __device__ fattn_mma_config ggml_cuda_fattn_mma_get_config(cons
     return ggml_cuda_fattn_mma_get_config_ampere(DKQ, DV, ncols);
 #elif defined(TURING_MMA_AVAILABLE)
     return ggml_cuda_fattn_mma_get_config_turing(DKQ, DV, ncols);
+#elif defined(AMD_MFMA_AVAILABLE)
+    return ggml_cuda_fattn_mma_get_config_cdna(DKQ, DV, ncols);
 #elif defined(VOLTA_MMA_AVAILABLE)
     return ggml_cuda_fattn_mma_get_config_volta(DKQ, DV, ncols);
 #elif defined(AMD_WMMA_AVAILABLE)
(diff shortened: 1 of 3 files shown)


ggml/src/ggml-cuda/fattn-mma-f16.cuh
```diff
@@ -5,6 +5,14 @@
...
 using namespace ggml_cuda_mma;
 
+// MFMA tiles are designed for 64-thread wavefronts (ne = I*J/64), so the effective
+// warp size for the fattn MMA kernel must be 64 on CDNA to match the hardware wavefront.
+#if defined(AMD_MFMA_AVAILABLE)
+#define FATTN_WARP_SIZE 64
+#else
+#define FATTN_WARP_SIZE WARP_SIZE
+#endif
+
```
JohannesGaessler: Instead of this, do something like `constexpr int warp_size = ggml_cuda_get_physical_warp_size();` in the device code and use that to replace the hardcoded value of 32 in `WARP_SIZE`.

ggml/src/ggml-cuda/fattn.cu
```diff
@@ -440,6 +440,14 @@ static best_fattn_kernel ggml_cuda_get_best_fattn_kernel(const int device, const
         return BEST_FATTN_KERNEL_MMA_F16;
     }
 
+    // Use MFMA flash attention for CDNA (MI100+):
+    if (amd_mfma_available(cc) && Q->ne[0] != 40 && Q->ne[0] != 72 && Q->ne[0] != 256 && Q->ne[0] != 576) {
+        if (can_use_vector_kernel && Q->ne[1] == 1) {
```
JohannesGaessler: Due to GQA the arithmetic intensity of the mma kernel is `Q->ne[1]*gqa_ratio`, usually there is some fraction of the full mma tile above which tensor cores (or what one would call it for AMD) become worthwhile vs. the vector kernel. So I would suspect this kernel selection logic is suboptimal.

ggml/src/ggml-cuda/mma.cuh
```diff
@@ -964,6 +964,34 @@ namespace ggml_cuda_mma {
...
+
+        // Convert existing half2 accumulator to float for MFMA:
+        floatx4_t acc_f32;
+        {
+            const halfx4_t acc_h = reinterpret_cast<const halfx4_t&>(D.x[0]);
+#pragma unroll
+            for (int i = 0; i < 4; ++i) {
+                acc_f32[i] = (float)acc_h[i];
+            }
+        }
```
JohannesGaessler, line 980: Long-term it may make sense to check whether for AMD in particular there are better options in terms of data types (does not need to be in this PR).

ggml/src/ggml-cuda/fattn.cu
Jayluci4: good point. benchmarked MMA vs VEC across batch sizes on Qwen2.5-1.5B (gqa_ratio=6):

ne[1]=256 (eff=1536): VEC 3% faster
ne[1]=512 (eff=3072): MMA 7% faster

crossover is around eff_nq=2048. changed the dispatch to `Q->ne[1] * (gqa_opt_applies ? gqa_ratio : 1) < 2048` for the VEC threshold. pushed.

ggml/src/ggml-cuda/mma.cuh
Jayluci4: noted. will look into bf16 and other MFMA data type options in a follow-up.

ggml/src/ggml-cuda/fattn-mma-f16.cuh
Jayluci4: done. replaced with `constexpr int warp_size = ggml_cuda_get_physical_warp_size()` in each device function. pushed.

JohannesGaessler:
I did some quick performance testing:

| GPU   | Model         |   Microbatch size | Test         |   t/s e877ad8bd |   t/s c45d5d56a |   Speedup |
|:------|:--------------|------------------:|:-------------|----------------:|----------------:|----------:|
| MI100 | llama 1B Q4_0 |                 1 | pp512@d32768 |          275.12 |          208.77 |      0.76 |
| MI100 | llama 1B Q4_0 |                 2 | pp512@d32768 |          429.44 |          294.53 |      0.69 |
| MI100 | llama 1B Q4_0 |                 4 | pp512@d32768 |          625.13 |          442.42 |      0.71 |
| MI100 | llama 1B Q4_0 |                 8 | pp512@d32768 |          857.80 |          719.16 |      0.84 |
| MI100 | llama 1B Q4_0 |                16 | pp512@d32768 |         1261.78 |         1492.52 |      1.18 |
| MI100 | llama 1B Q4_0 |                32 | pp512@d32768 |         1621.56 |         2097.17 |      1.29 |
| MI100 | llama 1B Q4_0 |                64 | pp512@d32768 |         1997.44 |         2733.96 |      1.37 |
| MI100 | llama 1B Q4_0 |               128 | pp512@d32768 |         2080.41 |         2931.04 |      1.41 |
| MI100 | llama 1B Q4_0 |               256 | pp512@d32768 |         2263.18 |         3331.88 |      1.47 |
| MI100 | llama 1B Q4_0 |               512 | pp512@d32768 |         2338.09 |         3377.25 |      1.44 |
| MI100 | llama 8B Q4_0 |                 1 | pp512@d32768 |           82.13 |           60.48 |      0.74 |
| MI100 | llama 8B Q4_0 |                 2 | pp512@d32768 |          134.43 |           57.74 |      0.43 |
| MI100 | llama 8B Q4_0 |                 4 | pp512@d32768 |          159.32 |           81.40 |      0.51 |
| MI100 | llama 8B Q4_0 |                 8 | pp512@d32768 |          233.71 |          141.85 |      0.61 |
| MI100 | llama 8B Q4_0 |                16 | pp512@d32768 |          411.26 |          388.07 |      0.94 |
| MI100 | llama 8B Q4_0 |                32 | pp512@d32768 |          529.96 |          567.44 |      1.07 |
| MI100 | llama 8B Q4_0 |                64 | pp512@d32768 |          614.80 |          716.30 |      1.17 |
| MI100 | llama 8B Q4_0 |               128 | pp512@d32768 |          644.70 |          787.26 |      1.22 |
| MI100 | llama 8B Q4_0 |               256 | pp512@d32768 |          676.56 |          857.12 |      1.27 |
| MI100 | llama 8B Q4_0 |               512 | pp512@d32768 |          706.15 |          893.15 |      1.26 |

LLaMA 3.1 1b has a head size of 64, LLaMA 3 8b has a head size of 128, both have a GQA ratio of 4. Notably on master the MI100 is using the tile kernel rather than the vector kernel as I found that kernel to perform better. So for CDNA1 the logic should be to use the mma kernel for an arithmetic intensity >= 64 for a head size of 64 and for an arithmetic intensity >= 128 for head sizes >64 and <= 128. The vector kernel should *not* be used unless it's not possible to exploit GQA-specific optimizations.

I tested the performance like this:

```bash
export mn=llama_3.2-1b && export q=q4_0
./build/bin/llama-bench --model models/opt/${mn}-${q}.gguf -fa 1 -r 1 -d 32768 -n 0 -ub "512,1-256*2" -o sql|sqlite3 llama-bench.sqlite
python3 scripts/compare-llama-bench.py -s gpu_info,model_type,n_ubatch -i llama-bench.sqlite -b e877ad8bd9e2f1 -c c45d5d56a|tee bench.tx
```

I would suggest you do the equivalent for one model with head size 64 and one with head size 128 and set the kernel selection logic accordingly.

Jayluci4:
ran the same benchmarks on MI300X with your method (`-d 32768 -p 512` varying `-ub`). compared MMA vs tile baseline:

**llama 3.2 1B Q4_0 (hsk=64, gqa=4):**
| ub | eff | tile t/s | mma t/s | speedup |
|---|---|---|---|---|
| 8 | 32 | 1,951 | 1,193 | 0.61 |
| 16 | 64 | 2,923 | 2,614 | 0.89 |
| 32 | 128 | 4,156 | 4,629 | 1.11 |
| 64 | 256 | 5,328 | 6,693 | 1.26 |
| 128 | 512 | 6,027 | 8,522 | 1.41 |
| 512 | 2048 | 6,976 | 11,801 | 1.69 |

**qwen2.5 1.5B Q4_K_M (hsk=128, gqa=6):**
| ub | eff | tile t/s | mma t/s | speedup |
|---|---|---|---|---|
| 16 | 96 | 1,800 | 1,163 | 0.65 |
| 32 | 192 | 2,704 | 2,354 | 0.87 |
| 64 | 384 | 3,934 | 3,934 | 1.00 |
| 128 | 768 | 4,866 | 5,152 | 1.06 |
| 256 | 1536 | 6,190 | 7,321 | 1.18 |
| 512 | 3072 | 6,703 | 8,278 | 1.24 |

MI300X crossovers are higher than MI100 - hsk=64 at eff≈128, hsk=128 at eff≈384. tile kernel benefits more from MI300X's higher memory bandwidth.

updated dispatch: `eff_nq = ne[1] * (gqa_opt_applies ? gqa_ratio : 1)`, threshold 128 for hsk<=64, 512 for hsk>64. removed VEC fallback, small batches fall through to tile. pushed.

ggml/src/ggml-cuda/fattn.cu
Jayluci4: updated. benchmarked MMA vs tile on MI300X with varying microbatch at d32768. MI300X crossovers are higher than MI100 (tile kernel benefits more from the 5.3 TB/s bandwidth). using head-size-dependent thresholds now: eff >= 128 for hsk<=64, eff >= 512 for hsk>64. removed VEC fallback, small batches fall through to tile. details in the PR comment above.

JohannesGaessler:
I would recommend that for performance testing you use a model with a GQA ratio that is a power of 2. With a GQA ratio of 6 there are additional complications because the mma kernel (but not the tile kernel) is padding this to 8.

Mushoz:
Since this PR was referenced and discussed in an issue describing a regression with Strix Halo (RDNA3.5), does that mean you plan on implementing something similar for RDNA3(.5)?

Jayluci4:
hmm @JohannesGaessler  so i re benchmarked with Llama 3.2 3B Q4_0 (hsk=128, gqa=4):

| ub | eff | tile t/s | mma t/s | speedup |
|---|---|---|---|---|
| 8 | 32 | 808 | 603 | 0.75 |
| 16 | 64 | 1,394 | 1,306 | 0.94 |
| 32 | 128 | 2,079 | 2,159 | 1.04 |
| 64 | 256 | 2,636 | 2,975 | 1.13 |
| 128 | 512 | 2,926 | 3,675 | 1.26 |
| 512 | 2048 | 3,386 | 4,644 | 1.37 |

crossover at eff=128, same as hsk=64. the qwen gqa=6 padding to 8 was inflating the threshold. simplified to a single `eff_nq >= 128` for all head sizes. pushed.

ggml/src/ggml-cuda/fattn.cu
```diff
@@ -440,6 +440,18 @@ static best_fattn_kernel ggml_cuda_get_best_fattn_kernel(const int device, const
         return BEST_FATTN_KERNEL_MMA_F16;
     }
 
+    // Use MFMA flash attention for CDNA (MI100+):
+    if (amd_mfma_available(cc) && Q->ne[0] != 40 && Q->ne[0] != 72 && Q->ne[0] != 256 && Q->ne[0] != 576) {
+        const int64_t eff_nq = Q->ne[1] * (gqa_opt_applies ? gqa_ratio : 1);
+        // MMA vs tile crossover benchmarked on MI300X @ d32768:
+        //   hsk=64  (gqa=4): MMA wins at eff >= 128 (+11%)
+        //   hsk=128 (gqa=4): MMA wins at eff >= 128 (+4%)
+        if (eff_nq >= 128) {
```
JohannesGaessler:
```suggestion
        if (eff_nq >= (GGML_CUDA_CC_IS_CDNA1(cc) && Q->ne[0] == 64 ? 64 : 128)) {
```

JohannesGaessler (approved):

ggml/src/ggml-cuda/fattn.cu
IMbackK: its extremely likely, from experience, that the crossover is in the same place for cdna2 as cdna1, but i cant test this atm.

JohannesGaessler: The difference is only going to be a few % either way though so I think it doesn't matter much.

IMbackK:
I haven't found anywhere where this is slower than the wmma kernel, so with -DGGML_HIP_ROCWMMA_FATTN=On it probably should not be preferred over this, as it is atm in this pr.

zhanghuiyulo:
Not sure how to handle ldmatrix_trans well on CDNA3, maybe adding a transposed gmem -> smem in flash_attn_ext_f16_load_tile is better, I just use a trick to do in-register transpose by mma on RDNA4 as loading each half one by one is too slow.

ggml/src/ggml-cuda/fattn-mma-f16.cuh
```diff
@@ -1644,7 +1721,8 @@ void ggml_cuda_flash_attn_ext_mma_f16_case(ggml_backend_cuda_context & ctx, ggml
     const int  nstages        = ggml_cuda_fattn_mma_get_nstages       (DKQ, DV, ncols1, ncols2, cc);
 
     const int cols_per_warp = std::min(ncols, get_cols_per_warp(cc));
-    const int nwarps        = nthreads / WARP_SIZE;
+    const int warp_size_host = amd_mfma_available(cc) ? 64 : WARP_SIZE;
```
IMbackK: const int warp_size_host = ggml_cuda_info().devices[ctx.device].warp_size;

would be better.

JohannesGaessler: Thank you for catching this. Once this last thing has been added I will merge this PR.

JohannesGaessler:
>I haven't found anywhere where this is slower than the wmma kernel, so with -DGGML_HIP_ROCWMMA_FATTN=On it probably should not be preferred over this, as it is atm in this pr.

If a user explicitly opts into using rocWMMA it should be used, even if it is slower.

IMbackK:
> > I haven't found anywhere where this is slower than the wmma kernel, so with -DGGML_HIP_ROCWMMA_FATTN=On it probably should not be preferred over this, as it is atm in this pr.
> 
> If a user explicitly opts into using rocWMMA it should be used, even if it is slower.

The problem is that we build all of our release artifacts with rocWMMA enabled, and want to continue to do so as long as its requried for decent performance on RDNA3, but this now means the release artifacts are suboptimal on CDNA.

Mushoz:
> > > I haven't found anywhere where this is slower than the wmma kernel, so with -DGGML_HIP_ROCWMMA_FATTN=On it probably should not be preferred over this, as it is atm in this pr.
> > 
> > 
> > If a user explicitly opts into using rocWMMA it should be used, even if it is slower.
> 
> The problem is that we build all of our release artifacts with rocWMMA enabled, and want to continue to do so as long as its requried for decent performance on RDNA3, but this now means the release artifacts are suboptimal on CDNA.

Can a similar thing that was done for CDNA3 in this particular PR be done for RDNA3(.5)? Because then all artifacts could simply be built without rocWMMA.

IMbackK:
It can, but the matrix layout of RDNA3 is unique and different from rdna4 and cdna so not immidately.

JohannesGaessler:
>The problem is that we build all of our release artifacts with rocWMMA enabled, and want to continue to do so as long as its requried for decent performance on RDNA3, but this now means the release artifacts are suboptimal on CDNA.

I'm fine with either way.

ggml/src/ggml-cuda/fattn-mma-f16.cuh
Jayluci4: Will add soon

Jayluci4: pushed the fix. sorry for the long delay. i was AFK for 2 days

JohannesGaessler (approved):

PR #19807, opened 2026-02-22
server: add `--log-output` option

tarruda:
This options enables logging of the LLM output to stdout, which makes it more convenient to inspect what is being generated on the server side. I'm aware of `--verbose`, but it also mixes llm output with a lot of other stats.

Here's an example of it being used with lm-evaluation-harness:

image


ngxson:
IMO this flag is just too narrow use case, I don't quite like adding it. We already had too many rarely-used / poorly-documented arguments.

This is better to be implemented in a proxy sitting between llama-server and your application, you will have total control about what's being logged.

PR #19810, opened 2026-02-22
Update Windows ROCm build to 26.Q1

superm1:
This updates Windows to the latest stable release of ROCm just as Linux was just recently updated (https://github.com/ggml-org/llama.cpp/pull/19433).

Here is a [build in my fork](https://github.com/superm1/llama.cpp/actions/runs/22314451817/job/64555241597) demonstrating it passes.

CC @IMbackK , @slojosic-amd, @hammmmy

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -616,13 +616,13 @@ jobs:
     runs-on: windows-2022
 
     env:
-      HIPSDK_INSTALLER_VERSION: "25.Q3"
+      HIPSDK_INSTALLER_VERSION: "26.Q1"
 
     strategy:
       matrix:
         include:
           - name: "radeon"
-            gpu_targets: "gfx1151;gfx1200;gfx1201;gfx1100;gfx1101;gfx1102;gfx1030;gfx1031;gfx1032"
+            gpu_targets: "gfx1150;gfx1151;gfx1200;gfx1201;gfx1100;gfx1101;gfx1102;gfx1030;gfx1031;gfx1032"
 
     steps:
       - name: Clone
@@ -632,7 +632,7 @@ jobs:
       - name: Grab rocWMMA package
         id: grab_rocwmma
         run: |
-          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.0.1/pool/main/r/rocwmma-dev/rocwmma-dev_2.0.0.70001-42~24.04_amd64.deb"
+          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70200-43~24.04_amd64.deb"
           7z x rocwmma.deb
           7z x data.tar
 
@@ -655,7 +655,7 @@ jobs:
         run: |
           $ErrorActionPreference = "Stop"
           write-host "Downloading AMD HIP SDK Installer"
-          Invoke-WebRequest -Uri "https://download.amd.com/developer/eula/rocm-hub/AMD-Software-PRO-Edition-${{ env.HIPSDK_INSTALLER_VERSION }}-WinSvr2022-For-HIP.exe" -OutFile "${env:RUNNER_TEMP}\rocm-install.exe"
+          Invoke-WebRequest -Uri "https://download.amd.com/developer/eula/rocm-hub/AMD-Software-PRO-Edition-${{ env.HIPSDK_INSTALLER_VERSION }}-Win11-For-HIP.exe" -OutFile "${env:RUNNER_TEMP}\rocm-install.exe"
           write-host "Installing AMD HIP SDK"
           $proc = Start-Process "${env:RUNNER_TEMP}\rocm-install.exe" -ArgumentList '-install' -NoNewWindow -PassThru
           $completed = $proc.WaitForExit(600000)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -701,8 +701,8 @@ jobs:
           cmake --build build --target ggml-hip -j ${env:NUMBER_OF_PROCESSORS}
```
slojosic-amd, line 701: If I don't exclude `--target ggml-hip` I am getting only `ggml-base.dll` and `ggml-hip.dll` same as in your artifact: https://github.com/superm1/llama.cpp/actions/runs/22279363198/artifacts/5607583284
After excluding `--target ggml-hip` I'm getting all important llama.cpp dll and exe files

.github/workflows/release.yml
```diff
@@ -604,25 +604,25 @@ jobs:
...
     env:
-      HIPSDK_INSTALLER_VERSION: "25.Q3"
+      HIPSDK_INSTALLER_VERSION: "26.Q1"
 
     strategy:
       matrix:
         include:
           - name: "radeon"
-            gpu_targets: "gfx1151;gfx1200;gfx1201;gfx1100;gfx1101;gfx1102;gfx1030;gfx1031;gfx1032"
+            gpu_targets: "gfx1150;gfx1151;gfx1200;gfx1201;gfx1100;gfx1101;gfx1102;gfx1030;gfx1031;gfx1032"
```
slojosic-amd, line 625: Adding `gfx1150`  is OK since HIP SDK 7.1.1 has support for both Strix Point (gfx1150) and Strix Halo (gfx1151) as well

.github/workflows/release.yml
```diff
@@ -632,7 +632,7 @@ jobs:
       - name: Grab rocWMMA package
         id: grab_rocwmma
         run: |
-          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.0.1/pool/main/r/rocwmma-dev/rocwmma-dev_2.0.0.70001-42~24.04_amd64.deb"
+          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70200-43~24.04_amd64.deb"
```
slojosic-amd, line 635: This is also OK and it is mandatory to use rocWMMA from ROCm 7.2 because older rocWMMA (from ROCm 7.1.1) doesn't have support for gfx1150

.github/workflows/release.yml
slojosic-amd: Linux part has `-DGGML_CPU_ALL_VARIANTS=ON` instead of `-DGGML_CPU=OFF` so I think this flag should be updated

slojosic-amd (changes requested):

.github/workflows/release.yml
superm1: Thanks, rolled this in.

.github/workflows/release.yml
superm1: updated, thanks!

slojosic-amd:
@superm1 Actually, `--target ggml-hip` and `-DGGML_CPU=OFF` should stay as they were before your latest changes, because true packaging comes at the end https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/release.yml#L867 during release job when all other important targets and missing files will be added to the new zip package. For example, `llama-bin-win-hip-radeon-x64.zip` is just intermediate artifact which is than used at the end to generate final `llama-b8144-bin-win-hip-radeon-x64.zip` package with all important .dll and .exe files.

.github/workflows/release.yml
```diff
@@ -689,20 +689,20 @@ jobs:
           cmake -G "Unix Makefiles" -B build -S . `
             -DCMAKE_C_COMPILER="${env:HIP_PATH}\bin\clang.exe" `
             -DCMAKE_CXX_COMPILER="${env:HIP_PATH}\bin\clang++.exe" `
-            -DCMAKE_CXX_FLAGS="-I$($PWD.Path.Replace('\', '/'))/opt/rocm-7.0.1/include/ -Wno-ignored-attributes -Wno-nested-anon-types" `
+            -DCMAKE_CXX_FLAGS="-I$($PWD.Path.Replace('\', '/'))/opt/rocm-7.2.0/include/ -Wno-ignored-attributes -Wno-nested-anon-types" `
             -DCMAKE_BUILD_TYPE=Release `
             -DGGML_BACKEND_DL=ON `
             -DGGML_NATIVE=OFF `
-            -DGGML_CPU=OFF `
```
slojosic-amd, line 696: Please revert this change

.github/workflows/release.yml
```diff
@@ -689,20 +689,20 @@ jobs:
...
+            -DGGML_CPU_ALL_VARIANTS=ON `
             -DAMDGPU_TARGETS="${{ matrix.gpu_targets }}" `
             -DGGML_HIP_ROCWMMA_FATTN=ON `
             -DGGML_HIP=ON `
             -DLLAMA_BUILD_BORINGSSL=ON
-          cmake --build build --target ggml-hip -j ${env:NUMBER_OF_PROCESSORS}
```
slojosic-amd, line 701: Please revert this change

.github/workflows/release.yml
```diff
@@ -689,20 +689,20 @@ jobs:
...
             -DCMAKE_C_COMPILER="${env:HIP_PATH}\bin\clang.exe" `
             -DCMAKE_CXX_COMPILER="${env:HIP_PATH}\bin\clang++.exe" `
-            -DCMAKE_CXX_FLAGS="-I$($PWD.Path.Replace('\', '/'))/opt/rocm-7.0.1/include/ -Wno-ignored-attributes -Wno-nested-anon-types" `
+            -DCMAKE_CXX_FLAGS="-I$($PWD.Path.Replace('\', '/'))/opt/rocm-7.2.0/include/ -Wno-ignored-attributes -Wno-nested-anon-types" `
             -DCMAKE_BUILD_TYPE=Release `
             -DGGML_BACKEND_DL=ON `
             -DGGML_NATIVE=OFF `
-            -DGGML_CPU=OFF `
+            -DGGML_CPU_ALL_VARIANTS=ON `
             -DAMDGPU_TARGETS="${{ matrix.gpu_targets }}" `
```
slojosic-amd: `-DAMDGPU_TARGETS` should become `-DGPU_TARGETS`

slojosic-amd (changes requested):

superm1:
Thanks for digging into it @slojosic-amd .  I've updated per your guidance.  Here is the updated test run.

https://github.com/superm1/llama.cpp/actions/runs/22353759055/job/64687505235

slojosic-amd (approved):
LGTM

IMbackK (approved):
This is fine, and i am not aware of anything blocking this, but i dont have a windows rocm setup to try the artifacts on.

slojosic-amd:
@superm1 @IMbackK I tried https://github.com/superm1/llama.cpp/actions/runs/22314451817/artifacts/5620163961 on Windows on Strix Halo and llama.bench and llama-cli seems to work fine:

image
image

CISC (approved):

.github/workflows/release.yml
```diff
@@ -616,13 +616,13 @@ jobs:
...
     env:
-      HIPSDK_INSTALLER_VERSION: "25.Q3"
+      HIPSDK_INSTALLER_VERSION: "26.Q1"
 
     strategy:
       matrix:
         include:
           - name: "radeon"
-            gpu_targets: "gfx1151;gfx1200;gfx1201;gfx1100;gfx1101;gfx1102;gfx1030;gfx1031;gfx1032"
+            gpu_targets: "gfx1150;gfx1151;gfx1200;gfx1201;gfx1100;gfx1101;gfx1102;gfx1030;gfx1031;gfx1032"
```
CISC, line 625: Why do we have different targets on Windows?

IMbackK: The Windows kernel driver dose not support the full range of gpus, gfx9 for instance was never supported at all on windows via the clr kernel interface windows hip uses

CISC:
@superm1 Can you also update the regular build and `windows-setup-rocm` action to resolve the installer cache conflict?

slojosic-amd:
> @superm1 Can you also update the regular build and `windows-setup-rocm` action to resolve the installer cache conflict?

@CISC I am preparing PR for this

slojosic-amd:
> > @superm1 Can you also update the regular build and `windows-setup-rocm` action to resolve the installer cache conflict?
> 
> @CISC I am preparing PR for this

@CISC FYI: https://github.com/ggml-org/llama.cpp/pull/19891

PR #19826, opened 2026-02-23
Support EuroBERT architecture

maximilianwerk:
DISCLAIMER 1: This PR is created with heavy help from claude-4.5-opus-high via Cursor.

The reason I still open it:

1. It mostly adds a new architecture, which is straight-forward and only adds a new case to the required functions.
2. The change to `imatrix.cpp` is actually quite important for last-token pooling embedding models. Without this change, their imatrix calculation does not respect the actual token processing, which creates the embeddings.
3. It is needed for the latest [jina-embeddings-v5-text-nano](https://huggingface.co/jinaai/jina-embeddings-v5-text-nano), which ranks on place 11 on MTEB, while only having 239M parameters.

DISCLAIMER 2: I work in the Jina team.

This PR includes:
- EuroBERT model implementation for GGUF conversion
- C++ backend support for EuroBERT architecture
- Change to the imatrix calculation logic for models that use last token pooling.

We converted the 4 jina-embedding-v5-text-nano models and validated their quality after conversion. For this we run several MTEB benchmark tasks and only observed a very small degradation, which is typical for gguf converted models.

In case that is relevant, here are the resulting artifacts. We merged the adapter into the base model and then ran the gguf conversion.

https://huggingface.co/jinaai/jina-embeddings-v5-text-nano-retrieval-GGUF
https://huggingface.co/jinaai/jina-embeddings-v5-text-nano-text-matching-GGUF
https://huggingface.co/jinaai/jina-embeddings-v5-text-nano-classification-GGUF
https://huggingface.co/jinaai/jina-embeddings-v5-text-nano-clustering-GGUF


convert_hf_to_gguf.py
```diff
@@ -1084,6 +1084,9 @@ def get_vocab_base_pre(self, tokenizer) -> str:
         if chkhsh == "0ef9807a4087ebef797fc749390439009c3b9eda9ad1a097abbe738f486c01e5":
             # ref: https://huggingface.co/meta-llama/Meta-Llama-3-8B
             res = "llama-bpe"
+        if chkhsh == "a023e9fdc5a11f034d3ef515b92350e56fb2af1f66c6b6811a4444ea9bf8763d":
+            # ref: https://huggingface.co/jinaai/jina-embeddings-v5-text-0.2B (EuroBert)
+            res = "llama-bpe"
```
CISC: Besides the fact that this URL does not exist, this is the wrong way to add this, you need to add it to `convert_hf_to_gguf_update.py` (in `pre_computed_hashes` since it's a duplicate name) and run it (which then updates `convert_hf_to_gguf.py` for you). :)

convert_hf_to_gguf.py
```diff
@@ -5703,6 +5706,50 @@ def modify_tensors(self, data_torch, name, bid):
...
+        # RoPE parameters
+        rope_theta = self.hparams.get("rope_theta", 250000.0)
+        self.gguf_writer.add_rope_freq_base(rope_theta)
+        logger.info(f"gguf: rope theta = {rope_theta}")
+        self.gguf_writer.add_rope_scaling_type(gguf.RopeScalingType.NONE)
+
+        # RMSNorm epsilon
+        rms_eps = self.hparams.get("rms_norm_eps", 1e-5)
+        self.gguf_writer.add_layer_norm_rms_eps(rms_eps)
+        logger.info(f"gguf: rms norm epsilon = {rms_eps}")
```
```suggestion
        self.gguf_writer.add_rope_scaling_type(gguf.RopeScalingType.NONE)
```
These are handled in the base class.

convert_hf_to_gguf.py
```diff
@@ -5703,6 +5706,50 @@ def modify_tensors(self, data_torch, name, bid):
...
+
+        # Pooling type - last token pooling (EOS token aggregates sequence info)
+        self.gguf_writer.add_pooling_type(gguf.PoolingType.LAST)
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        # Skip MLM head layers
+        if name.startswith("lm_head"):
+            return []
```
```suggestion
        # Skip MLM head layers
        if name.startswith("lm_head"):
            return
```
This may be a mistake, and is actually used, similar to CLS?

convert_hf_to_gguf.py
```diff
@@ -5703,6 +5706,50 @@ def modify_tensors(self, data_torch, name, bid):
...
+
+        # Strip "model." prefix from tensor names
+        if name.startswith("model."):
+            name = name[6:]
+
+        return [(self.map_tensor_name(name), data_torch)]
```
```suggestion
        yield from super().modify_tensors(data_torch, name, bid)
```
Your fork is outdated, please rebase.

src/CMakeLists.txt
```diff
@@ -99,6 +99,7 @@ add_library(llama
             models/nemotron-h.cpp
             models/nemotron.cpp
             models/neo-bert.cpp
+            models/eurobert.cpp
```
CISC: Move alphabetically.

convert_hf_to_gguf.py
maximilianwerk: The `lm_head` is the MLM head for token prediction in a generative model and not related to the CLS pooling. As such it is not relevant - at least for the jina models. We only have embedding models with last token pooling and as such are not interested in the MLM prediction head at all.

Or am I missing something here?

CISC: For jina, sure (why are you including it if it's unused?), but for EuroBERT in general.

convert_hf_to_gguf.py
maximilianwerk: I decided to remove this completely, since 

1. we have 4 adapters, which are merged into the base Jina models and 
2. the Jina model is not the actual base model.

So any usage of this conversion anyhow is best done by having a local model_path for all 4 variants. If you disagree, tell me and I could add either the [retrieval variant](https://huggingface.co/jinaai/jina-embeddings-v5-text-nano-retrieval) or all 4 variants here.

convert_hf_to_gguf.py
maximilianwerk: Fair point. Then I decided to remove `"EuroBertForMaskedLM", "EuroBertForSequenceClassification"`, since they most probably anyhow would fail, given the pooling type. The base `EuroBertModel` does also not include a `lm_head` and as such the check is redundant.

maximilianwerk:
Thanks a lot for the super quick reaction and all your comments. I adjusted the code accordingly.

convert_hf_to_gguf.py
CISC: I need a working conversion for this PR to be acceptable, the URL needs to point to a retrievable model, it does not matter which as long as the pre-tokenizer hash is identical for them all (which I see no reason for them not to be?). I assume the hash differs from EuroBERT base?

convert_hf_to_gguf.py
```diff
@@ -6125,6 +6125,37 @@ def modify_tensors(self, data_torch, name, bid):
         yield from super().modify_tensors(data_torch, name, bid)
 
 
+@ModelBase.register("EuroBertModel")
+class EuroBertModel(TextModel):
+    """EuroBert - a bidirectional encoder with Llama-style architecture (RoPE, RMSNorm, SwiGLU)."""
```
```suggestion
```
This is not particularly useful and omitted from all other models for that reason.

convert_hf_to_gguf.py
```diff
@@ -6125,6 +6125,37 @@ def modify_tensors(self, data_torch, name, bid):
...
+    model_arch = gguf.MODEL_ARCH.EUROBERT
+
+    def set_vocab(self):
+        # EuroBert uses Llama-style BPE tokenizer
+        self._set_vocab_gpt2()
+        # Match HuggingFace tokenizer: no BOS, but add EOS for last-token pooling
+        self.gguf_writer.add_add_bos_token(False)
+        self.gguf_writer.add_add_eos_token(True)
```
```suggestion
```
I think this is unnecessary and added automatically by `SpecialVocab`s post-processor parser, please confirm.

convert_hf_to_gguf.py
```diff
@@ -6125,6 +6125,37 @@ def modify_tensors(self, data_torch, name, bid):
...
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+
+        # EuroBert is bidirectional (encoder)
+        self.gguf_writer.add_causal_attention(False)
+
+        self.gguf_writer.add_rope_scaling_type(gguf.RopeScalingType.NONE)
+
+        # Pooling type - last token pooling (EOS token aggregates sequence info)
+        self.gguf_writer.add_pooling_type(gguf.PoolingType.LAST)
```
```suggestion
```
This is also unnecessary as you include `1_Pooling` in your LoRA-merged variants. Forcing it to `LAST` is probably also incorrect for other EuroBERT models.

convert_hf_to_gguf.py
maximilianwerk: I struggled getting the correct outputs from the quantized model and these two lines are actually necessary. They are not part of the `tokenizer_config.json`. Neither for the Jina models nor for the original EuroBert Model.

convert_hf_to_gguf.py
maximilianwerk: While going through the calling path, `_try_set_pooling_type` is never called, so I added it - as it is added for other embedding models. Then the correct pooling is taken from the input model.

convert_hf_to_gguf.py
maximilianwerk: If you dislike it, since in other EuroBert models this could be different, I could also change the whole class to be specific to `jina-embeddings-v5-text-nano`.

CISC: I know they are not part of `tokenizer_config.json`, but they should be autodetected from `tokenizer.json`s post-processor by `SpecialVocab` (invoked by `_set_vocab_gpt2`), please test without, you should be able to see that the metadata is added anyway on load/dump.

convert_hf_to_gguf.py
maximilianwerk: Running `convert_hf_to_gguf_update.py` also did quite some change to `models/ggml-vocab-llama-spm.gguf.out` by setting a lot of values to `0`. Should I commit this?

convert_hf_to_gguf_update.py
```diff
@@ -107,6 +107,7 @@ class TOKENIZER_TYPE(IntEnum):
     {"name": "jina-v2-en",       "tokt": TOKENIZER_TYPE.WPM, "repo": "https://huggingface.co/jinaai/jina-embeddings-v2-base-en", }, # WPM!
     {"name": "jina-v2-es",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/jinaai/jina-embeddings-v2-base-es", },
     {"name": "jina-v2-de",       "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/jinaai/jina-embeddings-v2-base-de", },
+    {"name": "jina-v5-nano-ret", "tokt": TOKENIZER_TYPE.BPE, "repo": "https://huggingface.co/jinaai/jina-embeddings-v5-text-nano-retrieval", },
```
CISC: This has to be a name of a pre-tokenizer handled in `llama-vocab.cpp`, it's ok to create a new one, but then you also have to add support for it in the aforementioned file, otherwise you can also reuse an existing one if you add it to `pre-computed_hashes` instead.

CISC: If you choose to add support in `llama-vocab.cpp` I would suggest naming it `jina-v5-nano` instead, as it's not specific to retrieval?

convert_hf_to_gguf.py
CISC: No, these files are for you to verify tokenization, it should also have created `ggml-vocab-jina-v5-nano-ret.gguf` and associated files, if you run `test-tokenizer-0` with that GGUF it will compare tokenization to `transformers` tokenization.

convert_hf_to_gguf.py
maximilianwerk: I can safely remove the line for the `add_add_eos_token`. However, when removing `add_add_bos_token` and comparing embeddings of 5 sentences between the HF model variant and the GGUF converted version with and without these two lines.

Without `self.gguf_writer.add_add_bos_token(False)`:

```
Per-sentence cosine similarity (HF vs GGUF):
  0.935828  The quick brown fox jumps over the lazy dog.
  0.983022  Machine learning models can embed text into vector…
  0.933010  Paris is the capital of France.
  0.948612  Short.
  0.956491  A very very very very very very very very very ver…

  Mean: 0.951393
  Min:  0.933010
```

With the `self.gguf_writer.add_add_bos_token(False)`:

```
Per-sentence cosine similarity (HF vs GGUF):
  0.999897  The quick brown fox jumps over the lazy dog.
  0.999878  Machine learning models can embed text into vector…
  0.999927  Paris is the capital of France.
  0.999896  Short.
  0.999854  A very very very very very very very very very ver…

  Mean: 0.999890
  Min:  0.999854
```

The log output of `convert_hf_to_gguf.py` is identical in both cases.

My understanding (to be fair with the help of Claude): SpecialVocab infers `add_*_token` only when a SpecialToken is present at the template boundary. It doesn't infer False from absence. Since the template for Jina models has no leading BOS but llama.cpp defaults to adding BOS for Llama tokenizers, we must explicitly set `add_bos_token(False)`.

convert_hf_to_gguf_update.py
maximilianwerk: Done. Thanks for the guidance.

convert_hf_to_gguf.py
maximilianwerk: OK, thanks a lot for the guidance. The test worked successfully, even though I needed to adjust it for cmake to work. I hope this change is welcomed. Shall I commit `ggml-vocab-jina-v5-nano.gguf`? I understood your above comment, that this is not needed and thus I did not.

convert_hf_to_gguf.py
```diff
@@ -6125,6 +6128,34 @@ def modify_tensors(self, data_torch, name, bid):
...
 
+@ModelBase.register("EuroBertModel", "JinaEmbeddingsV5Model")
+class EuroBertModel(TextModel):
+    model_arch = gguf.MODEL_ARCH.EUROBERT
+
+    def set_vocab(self):
+        # EuroBert uses Llama-style BPE tokenizer
+        self._set_vocab_gpt2()
+        # Match HuggingFace tokenizer: no BOS
+        self.gguf_writer.add_add_bos_token(False)
```
```suggestion
        self.gguf_writer.add_add_bos_token(False)
        self._set_vocab_gpt2()
```

src/models/eurobert.cpp
```diff
@@ -0,0 +1,113 @@
+#include "models.h"
+
+// EuroBert: bidirectional encoder with Llama-style architecture
+// - RoPE positional embeddings
+// - RMSNorm (pre-norm)
+// - SwiGLU FFN with separate gate/up projections
+// - Separate Q/K/V projections (not fused)
```
CISC:
```suggestion
```

src/models/eurobert.cpp
```diff
@@ -0,0 +1,113 @@
...
+llm_build_eurobert::llm_build_eurobert(const llama_model & model, const llm_graph_params & params) : llm_graph_context(params) {
+    const int64_t n_embd_head = hparams.n_embd_head_v;
+
+    GGML_ASSERT(n_embd_head == hparams.n_embd_head_k);
+
+    ggml_tensor * cur;
+    ggml_tensor * inpL;
+    ggml_tensor * inp_pos = build_inp_pos();
+
+    // construct input embeddings (token only, no type embeddings, no position embeddings - using RoPE)
```
CISC:
```suggestion
```

src/models/eurobert.cpp
```diff
@@ -0,0 +1,113 @@
...
+    cb(inpL, "inp_embd", -1);
+
+    auto * inp_attn = build_attn_inp_no_cache();
+
+    ggml_tensor * inp_out_ids = build_inp_out_ids();
+
+    for (int il = 0; il < n_layer; ++il) {
+        ggml_tensor * cur = inpL;
+
+        // pre-norm (RMSNorm before attention)
```
CISC:
```suggestion
```

src/models/eurobert.cpp
```diff
@@ -0,0 +1,113 @@
...
+        cur = build_norm(inpL,
+                model.layers[il].attn_norm, NULL,
+                LLM_NORM_RMS, il);
+
+        {
+            ggml_tensor * Qcur;
+            ggml_tensor * Kcur;
+            ggml_tensor * Vcur;
+
+            // self-attention with separate Q/K/V projections (no bias)
```
CISC:
```suggestion
```

src/models/eurobert.cpp
```diff
@@ -0,0 +1,113 @@
...
+            Qcur = build_lora_mm(model.layers[il].wq, cur);
+            Kcur = build_lora_mm(model.layers[il].wk, cur);
+            Vcur = build_lora_mm(model.layers[il].wv, cur);
+
+            Qcur = ggml_reshape_3d(ctx0, Qcur, n_embd_head, n_head, n_tokens);
+            Kcur = ggml_reshape_3d(ctx0, Kcur, n_embd_head, n_head_kv, n_tokens);
+            Vcur = ggml_reshape_3d(ctx0, Vcur, n_embd_head, n_head_kv, n_tokens);
+
+            // RoPE
```
CISC:
```suggestion
```

src/models/eurobert.cpp
```diff
@@ -0,0 +1,113 @@
...
+            inpL = ggml_get_rows(ctx0, inpL, inp_out_ids);
+        }
+
+        // residual connection
+        cur = ggml_add(ctx0, cur, inpL);
+
+        ggml_tensor * ffn_inp = cur;
+        cb(ffn_inp, "ffn_inp", il);
+
+        // pre-norm (RMSNorm before FFN)
```
CISC:
```suggestion
```

src/models/eurobert.cpp
```diff
@@ -0,0 +1,113 @@
...
+        cur = build_norm(ffn_inp,
+                model.layers[il].ffn_norm, NULL,
+                LLM_NORM_RMS, il);
+        cb(cur, "ffn_norm", il);
+
+        // SwiGLU feed-forward network with separate gate and up projections
```
CISC:
```suggestion
```

src/models/eurobert.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+        // residual connection
+        cur = ggml_add(ctx0, cur, ffn_inp);
+
+        // input for next layer
+        inpL = cur;
+    }
+    cur = inpL;
+
+    // final output norm (RMSNorm)
```
CISC:
```suggestion
```

tests/test-tokenizer-0.sh
CISC, line 1: LOL, I don't think anyone has used this script in a long long while. :)

CISC (approved):

tests/test-tokenizer-0.sh
maximilianwerk: But it worked out-of-the-box. :D Hopefully also correctly.

convert_hf_to_gguf.py
CISC: Didn't mean the bash script, I don't think anyone is using that, the change is welcome though, thanks. :)

No need to commit the GGUF.

tests/test-tokenizer-0.sh
maximilianwerk: I actually checked the resulting files and they looked reasonable.

maximilianwerk:
When I ran the failing test locally, the tests that failed in CI passed, but another one was failing (`test-tokenizers-ggml-vocabs` - complaining about missing vocab files for quite some models, but not the newly added). Are these tests flaky and thus expected to fail here and there?

CISC:
> When I ran the failing test locally, the tests that failed in CI passed, but another one was failing (`test-tokenizers-ggml-vocabs` - complaining about missing vocab files for quite some models, but not the newly added). Are these tests flaky and thus expected to fail here and there?

If you're talking about `ggml-ci-x64-cpu-low-perf`, ignore it, it's just a random `ccache` problem, everything is ok, will merge now.

Regarding `test-tokenizers-ggml-vocabs` you might not have installed and initialized `git-lfs`? Then the files are not downloaded properly, either way, not important. :)

maximilianwerk:
Amazing. Thanks again for the very pleasant experience. You did a great job in reviewing! :heart:

PR #19829, opened 2026-02-23
webui: Add setting to have full height Code Blocks in Chat Messages

allozaur:
Close #19742


ggerganov (approved):

woof-dog:
@allozaur, thank you for providing this setting! It appears you forgot to update `parameter-sync.service.ts` - would you please also do this so the setting is configurable from the CLI? Thanks again.

allozaur:
> @allozaur, thank you for providing this setting! It appears you forgot to update `parameter-sync.service.ts` - would you please also do this so the setting is configurable from the CLI? Thanks again.

Ah, yes, of course. My bad, added it in #19835

PR #17956, opened 2025-12-12
server: add encoder-decoder model support (T5, BART, MADLAD)

Turee:
Add support for encoder-decoder models in llama-server, matching the behavior of llama-cli. This enables translation models like MADLAD and other T5-based models to work with the server.

Changes:
- Add has_encoder flag to detect encoder-decoder models at load time
- Implement llama_encode() call for encoder-decoder prompt processing
- Use decoder_start_token to initialize decoder after encoding
- Clear decoder KV cache before each new request (no prefix caching)
- Disable incompatible features for encoder-decoder models:
  - Context shift (encoder outputs are fixed)
  - Speculative decoding (not supported)
  - Warn about prompt caching differences
- Add edge case handling for empty text tokens

The encoder processes the full prompt, then the decoder generates output using cross-attention to the encoder's hidden states.

*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

Cursor /w Opus 4.5 was heavily utilized to develop this change.

PR #19832, opened 2026-02-23
Add Aya 101 multi-lingual translation support to llama.cpp

Acceldium:
Add support for [CohereForAI/aya-101](https://huggingface.co/CohereForAI/aya-101), a 13B parameter multilingual encoder-decoder model based on the T5-XXL architecture, supporting translation across 101 languages. 

The implementation reuses the existing `llm_build_t5_enc` / `llm_build_t5_dec` graph builders since AYA-101 is architecturally identical to Flan-T5-XXL.

This PR also fixes a bug in `llama-server` where encoder-decoder models were not handled in the `update_slots()` pipeline, causing garbage output when serving models like T5 or AYA through the completions API.

  - Extend the encoder output embedding cache check to include `LLM_ARCH_AYA` alongside `LLM_ARCH_T5`, so encoder embeddings are stored in `cross.v_embd` for cross-attention.
  - Extend the `output_reserve()` buffer allocation check to include `LLM_ARCH_AYA`, so `has_embd = true` is set for encoder-decoder models. Without this, `embd.data` is null and `llama_encode()` segfaults.

 Add encoder-decoder handling in `update_slots()`. 
 When `llama_model_has_encoder()` returns true:
  1. Build an explicit encoder batch with each prompt token assigned to `slot.id` as the sequence ID (critical for cross-attention lookup).
  2. Call `llama_encode()` to run the encoder and store embeddings.
  3. Determine the decoder start token via `llama_model_decoder_start_token()` (falls back to BOS).
  4. Replace the prompt tokens with a single decoder start token.
  5. Disable prompt caching for encoder-decoder models.

  **Note:** This fix benefits all encoder-decoder models served via `llama-server`, not just AYA. Previously, T5 models loaded in the server would also produce garbage output because the encoder pass was never executed.
  
  The key API function is `llama_model_has_encoder()` which returns `true` for AYA. Both `llama-completion` and `llama-server` check this to decide whether to run the two-pass encoder-decoder flow.

```bash
# Download AYA-101 from HuggingFace
huggingface-cli download CohereForAI/aya-101 --local-dir aya-101-hf

# Convert to GGUF (specify --model-name to select the "aya" converter)
python convert_hf_to_gguf.py aya-101-hf --outtype f16 --model-name aya

# Quantize (optional)
./build/bin/llama-quantize aya-101-f16.gguf aya-101-Q8_0.gguf Q8_0


## How to Run

### llama-completion (CLI)

```bash
# CPU only
./build/bin/llama-completion \
    -m aya-101-Q8_0.gguf \
    -p "Translate to French: Hello, how are you today?" \
    -n 50 --temp 0.7

# Output: Bonjour, comment allez-vous aujourd'hui ?

# With GPU offload (all 48 layers)
./build/bin/llama-completion \
    -m aya-101-Q8_0.gguf \
    -p "Translate to Spanish: The weather is beautiful." \
    -n 50 --temp 0.7 -ngl 48

# Output: El tiempo es hermoso.
```

### llama-server (HTTP API)

```bash
# Start server
./build/bin/llama-server \
    -m aya-101-Q8_0.gguf \
    -ngl 48 --host 0.0.0.0 --port 8080

# Translate via API
curl http://127.0.0.1:8080/v1/completions \
    -d '{
        "prompt": "Translate to French: Hello, how are you today?",
        "max_tokens": 50,
        "temperature": 0.7
    }'


ngxson:
@saurabhdash2512 Wondering if you can help reviewing this PR? Thanks!

Turee:
Hi! Thanks for adding Aya-101 support!

FYI I did similar fix to the server: https://github.com/ggml-org/llama.cpp/pull/17956 , I had some problems with concurrent requests I had to fix:

  - The encoder batch needs slot.id assigned as the seq_id, otherwise concurrent requests will
  corrupt each other's encoder KV state.
  - llama_memory_seq_rm() should be called before encoding to avoid stale decoder state leaking across
  requests in the same slot.

Turee:
Regarding my previous comment, I tested this on both implementations with concurrent requests and found the same issue. Here is the overview and hypothesis of the root cause: 


 ### Problem

  When `llama-server` processes multiple encoder-decoder requests in parallel (default `-np 4`), only the last-encoded request produces correct output. All other slots decode against the wrong encoder embeddings because `llama_encode()` writes
  to a shared cross-attention embedding buffer (`cross.v_embd`) that gets overwritten by each subsequent encode call within the same `update_slots` iteration.

  ### Affected

  All encoder-decoder models (T5, MADLAD, BART, AYA) served via `llama-server` with `-np > 1`. Both PR #17956 and PR #19832 exhibit the same behavior.

  ### Steps to reproduce

  ```bash
  # Start server with default parallel slots (np=4)
  ./build/bin/llama-server -m madlad-10b-q8_0.gguf --port 8090

  # Fire 4 concurrent translation requests
  for i in 1 2 3 4; do
    curl -s http://localhost:8090/v1/completions \
      -d "{\"prompt\": \"<2fi> Sentence $i: I love programming.\", \"max_tokens\": 50, \"temperature\": 0.1}" &
  done
  wait
  ```

  Result: Only 1 of 4 responses is a correct Finnish translation. The others produce garbage (wrong language, numbered lists, timestamps).

  Expected: All 4 responses should be correct Finnish translations.

  Workaround

  Force single slot with -np 1. Requests serialize through one slot and all produce correct output.

  Root cause

  cross.v_embd in llama_context is a single buffer, not per-sequence. Each llama_encode() overwrites the previous encoder output. When multiple slots encode in the same update_slots pass, only the last encoder output survives for the subsequent
  decode batch.

Turee:
Vibe coded fix for the above: 

https://github.com/ggml-org/llama.cpp/pull/17956/changes/54f926ea69c03829e04971257bad66e720f105b0

https://github.com/ggml-org/llama.cpp/pull/17956/changes/00206737bbe0baca6b243c3041a617af82fb515b

Acceldium:
I've added a more detailed explanation and proposal of an additional bug that was encountered. The ggml_can_repeat crash with -np 4 without --kv-unified is a pre-existing upstream issue in llama.cpp's multi-stream KV cache path for encoder-decoder models. It happens during graph reservation when kv_unified = false and n_seq_max > 1. 

This crash exists in upstream llama.cpp for all T5-family models -- it's not something we introduced or were asked to fix. The workaround is --kv-unified, which is the correct flag for encoder-decoder parallel serving anyway since these models don't benefit from per-stream KV isolation.

Vibe coded a proposal.

PR #19833, opened 2026-02-23
sampling : support multiple outputs per sequence

danbev:
This commit adds support for multiple outputs per sequence in the backend sampling implementation.

The main motivation for this change is to be able to support speculative decoding using backend samplers where multiple outputs for the same sequence would be needed.

----
Example usage with llama-server:
```console
$ llama-server -m models/Qwen2.5-VL-3B-Instruct-Q8_0.gguf  \
      --spec-type ngram-mod --spec-ngram-size-n 8 --draft-min 1 --draft-max 64 \
      -bs -v
```
----

### TODO:
- [x] Make the graph "static" for sequences with multiple outputs.
- [ ] Perhaps add a configuration option specifying that multiple outputs are not used for backend sampling to avoid the additional nodes in the compute graph.

### DIFF
--- a/common/reasoning-budget.cpp
+++ b/common/reasoning-budget.cpp
@@ -193,6 +193,7 @@ static struct llama_sampler_i common_reasoning_budget_i = {
     /* .backend_accept    = */ nullptr,
     /* .backend_apply     = */ nullptr,
     /* .backend_set_input = */ nullptr,
+    /* .backend_n_nodes   = */ nullptr,
 };
 
 static struct llama_sampler * common_reasoning_budget_clone(const struct llama_sampler * smpl) {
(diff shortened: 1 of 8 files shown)


ruixiang63:
I think this PR would be helpful for DFlash spec dec, as it can sample at most `block_size-1` draft tokens at the same time. Would be great to see this feature can be merged! cc @ggerganov

gaugarg-nv:
@ruixiang63 @ggerganov Just FYI, I am working on a PR that adds support for this.
I should have it ready by the end of this week.

ruixiang63:
> @ruixiang63 @ggerganov Just FYI, I am working on a PR that adds support for this. I should have it ready by the end of this week.

great!

danbev:
Closing in favor of #25532

PR #19834, opened 2026-02-23
Add gfx950 target for ROCm 7.2

superm1:
AMD's binary builds do include GFX950.

CC @IMbackK , @slojosic-amd, @hammmmy


superm1:
Here is test run in my fork: https://github.com/superm1/llama.cpp/actions/runs/22309686343/job/64538059208

IMbackK:
gfx950 is not supported.

it will match https://github.com/ggml-org/llama.cpp/blob/9051663d5d67bdab765e9bc9afee41ed34f09260/ggml/src/ggml-cuda/common.cuh#L90 but not https://github.com/ggml-org/llama.cpp/blob/9051663d5d67bdab765e9bc9afee41ed34f09260/ggml/src/ggml-cuda/vendors/hip.h#L198 which will cause the host and device code to get out of sync.

Unfortunately the hip compiler dose not provide a better solution to detect it on the device side at compile time other than to add every device family.

Someone with access to gfx950 hardware will have to add it, then test, and ideally also tune the parameters these macros are for first.

superm1:
@slojosic-amd can you help?

CISC (approved):
Since this is not built by any other CI I'd like to see a fresh successful run please.

PR #19836, opened 2026-02-23
Fix test-jinja on Windows.

Exile333:
When running test-jinja on Windows in python mode, there are several problems present. They are related to DOS-style line-breaks and UTF-8 handling on Windows.

This PR proposes following changes for test-jinja on Windows in python mode.
1. Expect line breaks to be in DOS style (`"\r\n"` instead of `"\n"`).
2. Skip test that checks jinja's `to_json(ensure_ascii=True)`. Windows by default uses ANSI codepages instead of UTF-8, thus when spawning a subprocess via `CreateProcessA`, Windows will treat command's arguments as in ANSI encoding. This test can still be run successfully on Windows if one turns on system option Use Unicode UTF-8 for worldwide language support, which is still in beta.

After these changes all Windows jinja tests are passed successfully.

### DIFF
--- a/tests/test-jinja.cpp
+++ b/tests/test-jinja.cpp
@@ -77,6 +77,15 @@ int main(int argc, char *argv[]) {
 }
 
 static void test_whitespace_control(testing & t) {
+#ifdef _WIN32
+    test_template(t, "trim_blocks removes newline after tag",
+        "{% if true %}\r\n"
+        "hello\r\n"
+        "{% endif %}\r\n",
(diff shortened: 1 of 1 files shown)


CISC:
Running the test in Python mode is a tool for maintainers, so it doesn't really matter that it doesn't work properly on Windows.

Either way, "fixing" the tests is the wrong way to go about it, if anything, fix the python invocation (input can be piped f.ex.).

Issue #19838, opened 2026-02-23
Feature Request: Context truncation to keep chat progressing


ltoniazzi:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

#### Truncate Message List [Spike]

Can we automatically truncate the chat template to keep the chat progressing even beyond the context limit?

A simple and relatively safe way (to keep consistent context) is to remove old messages, except for the (initial) system prompt.

This can be done by budgeting with `n_predict` how many tokens are left and trigger dropping old messages pairs/tuples (user msg + ai message + tool message etc), reducing the input size to a desired threshold.

Note that, with `n_predict = -1`,  we cannot budget clearly, so we need to trigger in a conservative manner.

##### Consideration

- This truncation should exploit (when active) the [--cache-reuse N](https://ggml.ai/f0.png) algorithm for supported models (see [this comment](https://github.com/ggml-org/llama.cpp/issues/17284#issuecomment-3943587648))
- What should be the role of `--context-shift` going forward. Should it be removed or should we make it's docs clearer?

#### Motivation

As discussed in Issue https://github.com/ggml-org/llama.cpp/issues/17284, `--context-shift`:
- only works by shifting when the generation crosses the `ctx_size` (but input prompt doesn't), 
- while the run fails when the input prompt already crosses the `ctx_size`.

Before, `--context-shift` would be triggered even in the latter case, allowing infinite conversations. This resulted in a breaking change with some llama.cpp users now experiencing a bug.

But the previous behaviour of `--context-shift` will not be reintroduced as it has several flaws, like:
  - Breaking chat template, 
  - Removing system prompt 
  - Invalidating KV caches that encode token position.

#### Possible Implementation

1. Focus on cli and server interactions
2. Create a common util to truncate the list of messages
3. Inject it separately to the cli and the relevant server routes 

See draft PR

JohnAlcatraz:
> * The table above does not include that you will also pay to recompute the KV-cache at step 3. In this examples it was 0.5s
> * Optimisations can be done to reduce the number of tokenisations needed.

[...]

PR #19841, opened 2026-02-23
server : add chat truncation to keep chat going

ltoniazzi:
### Context

Addressing Issue https://github.com/ggml-org/llama.cpp/issues/19838

Introduce an optional parameter so then:
1. Chat can continue indefinitely beyond context, 
2. Quality of the truncated prompt is not compromised too much in most cases.

#### Truncation

Define a **turn** as: a sequence of messages between two user messages, including only the left user message

- Remove old "turns" so that tokens align with a percentage `--chat-truncate-max-keep` of the full context (but preserve at least 1 turn). If
  - `n_predict>0`: trigger if it's possible to cross the context size (and target truncation as `min(budget, ctx-truncate*context)`
  - `n_predict=-1`: always trigger reset when crossing target percentage tokens (maybe need to add and upper pct threshold, as this will trigger KV-cache refresh must turns)

#### How to not recompute messages to remove every time? 

Issue:
- WebUI sends the full history to the server.
- Once threshold is crossed, server will always re-compute the turns to remove, wastefully, and removing turns unnecessarily.

Fix (quick + experimental + AI contributed)
- Added a mapping in `server_context_impl` for server to keep the hash the message history, with value equal to the turns to drop. So, when a conversation for which we already dropped turns comes in:
   - instantly drops to previously truncated turns, 
   - before computing the (potential) truncation of this turn.

- See [this comment](https://github.com/ggml-org/llama.cpp/issues/19838#issuecomment-4188459588) for more details.

#### Injection locations

Chat through `llama-server`: in `oaicompat_chat_params_parse` before the [first explicit message modification](https://github.com/ggml-org/llama.cpp/blob/c17dce4f5c725054e10cfafc883cbd24e8b6303b/tools/server/server-common.cpp#L1048).

### TODOs

- [x] What is the best injection strategy of the truncation logic. 
  - [x] Are the current locations good? 
  - [x] Should add truncation to other invocation paths beside cli/srever? Start server only
  - [ ] Add comprehensive test
- [x] Figure out logical groups of message sequences to remove (ex: user + ai + tool)
- [ ] How to not recompute truncation every turn
- [ ] Test consequence of reordering on downstream KV cache reordering 

<details>
<summary>AI Usage Disclamer</summary>

Draft PR
- Used AI to plan, discuss and generate part of the code, because this Draft PR is in a preliminary stage, trying to plan a direction for this feature with the maintainers.
- Still debugged all relevant lines

</details>

<details>
<summary>Note on llama-cli</summary>

Skipping `llama-cli` for this PR. 

It seems these two invocation approaches meet at  [server-context.cpp:`process_single_task`](https://github.com/ggml-org/llama.cpp/blob/b68a83e641b3ebe6465970b34e99f3f0e0a0b21a/tools/server/server-context.cpp#L1657) , but at that time in `server_task` the messages list is lost, and we only have tokens or full templated string (correct?)

So to select messages types, it's easier to catch the messages `common_chat_msg` upstream:
- `llama-cli`: inject truncation here before [cli.cpp:`common_chat_templates_apply`](https://github.com/ggml-org/llama.cpp/blob/b68a83e641b3ebe6465970b34e99f3f0e0a0b21a/tools/cli/cli.cpp#L194) in `format_chat`
- `llama-server`: inject truncation here a bit before [server-common.cpp:`common_chat_templates_apply`](https://github.com/ggml-org/llama.cpp/blob/b68a83e641b3ebe6465970b34e99f3f0e0a0b21a/tools/server/server-common.cpp#L1012) in `oaicompat_chat_params_parse`
</details>


ltoniazzi:
@ggerganov This is a basic draft for the truncation logic discussed. Let me know if you have in mind something different

pwilkin:
Throwing this in as it might be interesting and relevant: https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/compaction.md

Midaychi:
LM Studio's Truncate middle might be one potential strat?

common/chat.cpp
```diff
@@ -3353,3 +3353,64 @@ std::map<std::string, bool> common_chat_templates_get_caps(const common_chat_tem
...
+        if (first_non_sys < inputs.messages.size() &&
+            inputs.messages[first_non_sys].role == "assistant") {
+            inputs.messages.erase(inputs.messages.begin() + (ptrdiff_t)first_non_sys);
+        }
+
+        // Re-render and re-count
+        chat_result = common_chat_templates_apply(tmpls, inputs);
+        tokens      = common_tokenize(vocab, chat_result.prompt, true, true);
+        n_tokens    = (int32_t)tokens.size();
+    }
```
ggerganov: I think we should remove any type of message different than the `"system"`. The logic is simpler and I don't think it would cause issue. For example, this should be OK:

```bash
system
toolK          # <--- after truncation, new context starts with a tool message
assistantK
userK+1
...
assistantN
userN+1
```

common/arg.cpp
```diff
@@ -1310,6 +1310,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.ctx_shift = value;
         }
     ).set_examples({LLAMA_EXAMPLE_COMPLETION, LLAMA_EXAMPLE_CLI, LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_IMATRIX, LLAMA_EXAMPLE_PERPLEXITY}).set_env("LLAMA_ARG_CONTEXT_SHIFT"));
+    add_opt(common_arg(
+        {"--ctx-truncation"}, "F",
```
ggerganov: Should be called `--chat-truncation` to emphasize that this applies only to chat endpoint.

common/common.h
```diff
@@ -473,7 +473,9 @@ struct common_params {
     bool ctx_shift         = false; // context shift on infinite text generation
     bool swa_full          = false; // use full-size SWA cache (https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055)
     bool kv_unified        = false; // enable unified KV cache
-
+    
+    float ctx_truncation   = -1.0f; // fraction of n_ctx_slot to target after chat truncation (-1 = disabled)
```
ggerganov: Move along-side other server params below

tools/server/server-common.cpp
```diff
@@ -985,6 +989,15 @@ json oaicompat_chat_params_parse(
...
 
+    // Chat truncation: drop oldest non-system turn pairs until prompt fits in context
+    // TODO is this a good/consistent place for the truncation to happen?
+    if (ctx_truncation > 0.0f && vocab != nullptr && n_ctx_slot > 0) {
+        const int32_t n_predict_req = json_value(body, "max_tokens",
+                                        json_value(body, "n_predict", n_predict_default));
+        common_chat_truncate_messages(inputs, opt.tmpls.get(), vocab,
+                                      n_ctx_slot, n_predict_req, ctx_truncation);
+    }
+
```
ggerganov: Not 100% sure. Seems like the correct place, but would defer to @ngxson for comment.

tools/server/server-common.h
```diff
@@ -289,10 +289,17 @@ struct server_chat_params {
...
+// dropped from the message list until the rendered prompt fits within
+// ctx_truncation * n_ctx_slot tokens (triggered only when prompt > n_ctx_slot - n_predict_default).
 json oaicompat_chat_params_parse(
     json & body, /* openai api json semantics */
     const server_chat_params & opt,
-    std::vector<raw_buffer> & out_files);
+    std::vector<raw_buffer> & out_files,
+    const struct llama_vocab * vocab            = nullptr,
+    int32_t                   n_ctx_slot        = 0,
+    int32_t                   n_predict_default = -1,
```
ggerganov:
```suggestion
    int32_t                   n_predict = -1,
```

tools/server/server-common.h
```diff
@@ -289,10 +289,17 @@ struct server_chat_params {
...
+    float                     ctx_truncation    = -1.0f);
```
ggerganov: These new extra args are a bit too cumbersome for their purpose. Should probably wrap them in a helper `struct oaicompat_chat_truncation_params`.

ggerganov (commented):
@ngxson WDYT about this functionality? IMO looks OK and with a bit of polish it can be accepted.

common/chat.cpp
ltoniazzi: I think some jinja templates will raise if we break required sequences.

For example [this gemma template ](https://github.com/ggml-org/llama.cpp/blob/c830f99cfa79d7e627e48de32280838f97b41115/models/templates/google-gemma-2-2b-it.jinja#L1) raises if `user/assistant` alternation is broken.

As after we selected the desired messages, the code wants to run  `common_chat_templates_apply -> common_chat_templates_apply_jinja` to create the final templated string prompt.

Should we figure out a generally safe (but hardcoded) truncation strategy (like truncate only full sequences of user to user messages)? 

Or try extract it dynamically from the jinja templates?

CISC: AFAIK the only enforced sequence in many templates is the user/assistant alternation, meaning we must ensure that the first role after `system` is `user`. Any other roles interspersed should not matter as they have been accepted in previous turns, so no need to inspect template.

ngxson: IMO a safer logic would be to consider a conversation as a list of turns and not a list of messages. A turns starts from user message and ends before the next user message.

Example a conversation: system, user0, assistant0, user1, tool1, assistant1,...
- turn 0: user0, assistant0
- turn 1: user1, tool1, assistant1

If we need to purge the conversation, we start to purge turn0, then turn1, etc

Anything outside the turn must not be purged, so checking system is irrelevant as it will be outside of any turns anyway

ngxson: Note that some models inserts end-of-turn token only **after** assistant response, like GPT-OSS. and GLM even **doesn't** have end-of-turn token at all, so removing mid-turn will mess up the context (the template will be formatted successfully, but model performance is degraded)

tests/test-chat-truncation.cpp
ngxson, line 1: IMO we don't need a dedicated test for this, a simple test case in server test (pytest) should be enough

tools/cli/cli.cpp
```diff
@@ -190,6 +192,12 @@ struct cli_context {
         inputs.add_generation_prompt = true;
         inputs.enable_thinking       = chat_params.enable_thinking;
 
+        if (ctx_truncation > 0.0f) {
```
ngxson: it doesn't make sense to include this logic here, the CLI manage its own list of conversation, we can simply drop the message progressively

IMO better to touch the CLI in another PR, not in the same PR

ngxson (commented):
As mentioned in https://github.com/ggml-org/llama.cpp/pull/19841#discussion_r2847380962 , the logic can be simplified, and I think the code implementation can be simplified even further. We should only touch server and arg code in this PR. Adding a new function to `chat.h` seems a bit redundant IMO

tools/server/server-common.cpp
ngxson: The `n_predict` should be extracted to a getter, otherwise it's duplicated in other places in the code.

It's OK to put the `common_chat_truncate_messages` here, but as mentioned in my review, it's better to be a server-specific function, not a `common_chat_` API

tools/server/server-common.h
ngxson: We already had `const server_chat_params & opt` for chat-specific parameters

common/chat.h
```diff
@@ -250,3 +250,15 @@ nlohmann::ordered_json common_chat_msg_diff_to_json_oaicompat(const common_chat_
...
+// prompt fits within `fraction * n_ctx_slot` tokens, or only one non-system message remains.
+// Only triggered when the current prompt would overflow n_ctx_slot - n_predict tokens.
+// The system prompt and the most-recent user message are always preserved.
+void common_chat_truncate_messages(
+    common_chat_templates_inputs & inputs,
+    const common_chat_templates * tmpls,
+    const struct llama_vocab    * vocab,
+    int32_t                       n_ctx_slot,
+    int32_t                       n_predict,
+    float                         fraction);
```
ngxson: those 3 params can be collapsed into one:

```cpp
truncate_messages(
  common_chat_templates_inputs & inputs,
  const common_chat_templates * tmpls,
  const struct llama_vocab    * vocab, // only for doing tokenization
  size_t n_max // max number of tokens that the conversation can take, after truncation
)
```

the caller should decide `n_max` based on what it has

ltoniazzi: Thanks! I was thinking:

1. **Keep `fraction` and pass `n_ctx_slot`**
   Using `size_t n_max` means the caller needs to chose the `int` conscious of the total default `ctx_size` of the specific model, which can be prone to mistakes.  (Like setting something that is too low -> truncate too early and lose context, or too high -> never truncate or truncate too close to `ctx_size`, so then generation keeps interrupting.)

    While with `float fraction` we can achieve the same by cutting once we reach the threshold (we calculate from the extra param `n_ctx_slot`),  as in the current approach for `n_predict=-1`. 
Then the caller can have a "safe" param, say `0.8`, that works independently of the model and/or explicit ctx constraint.

    What do you think?

2. **Pass `n_predict`**
   I fell that  when `n_predict>0`, with budgeting (`n_ctx_slot - n_predict`) we can automatically preserve the most amount of context.  As we truncate only when actually the overflow can happen. 
While if we always truncate at the threshold `fraction/n_max`, then we will end up:
    - Likely truncating at every subsequent interaction -> breaking the KV cache each user message
    - Truncating too early, not exploit a potentially longer context conservation.

ngxson: In english, "truncate" means "if something is too long than N, then cut it and take the part shorter than N"

So I don't see why you need to pass inference-specific variables like n_predict / n_ctx_slot / fraction

Example: `truncate(string input, int max_characters)` --> truncate text such that `length(output) < max_characters`

Same idea for `truncate(conversation, int max_tokens)`

ltoniazzi: I think `n_ctx_slot` is needed if we allow the caller to pass a `fraction`/`chat-truncation` of the tot context to be truncated.
So the caller does not have to think in absolute `int` tokens (my point (1) above)

While `n_predict` is useful if we want to trigger truncation "optimally", meaning only when generation might saturate `n_ctx_slot` (only for `n_predict>0`). (my point (2) above)

ngxson: I don't see why we need to "compose" the truncate function into (1) calculation what's the optimal tokens to keep and (2) does the actual truncation.

Keeping them as 2 separated function make things much more intuitive to understand.

In terms of functional programming, one is higher-order function of the other, I don't see any reasons not to respect this principle of functional programming

ltoniazzi: Yes I agree 👍  

Sorry I think I misunderstood you, do you mean

a. Compute separately

  ```
  n_max = max_tokens_to_truncate(n_ctx_slot, fraction)
  ```
  
  and then pass it to `truncate_messages(..., n_max)`

b. Use `n_max` directly from user server parameters

?

ngxson: yes, that's what I mean: `truncate_messages(messages, calculate_max_tokens_to_keep(n_ctx_slot, fraction, n_predict))`

naming can be changed to make it blend in with other functions.

in terms of functional programming, this is just `f(g(x))`

tools/server/server-common.h
```diff
@@ -44,6 +44,18 @@ static T json_value(const json & body, const std::string & key, const T & defaul
...
 }
 
+// Resolves the requested number of tokens to predict from a request body.
+// "n_predict" takes priority over "max_tokens" (OAI alias); falls back to fallback_n_predict.
+static inline int32_t get_n_predict(const json & body, int32_t fallback_n_predict) {
+    return json_value(body, "n_predict", json_value(body, "max_tokens", fallback_n_predict));
+}
+
+// Get n_predict by prioritizing server value if body is missing or negative
+static inline int32_t get_n_predict_with_server_priority(const json & body, int32_t server_default) {
```
ltoniazzi, line 56: Not sure about this one, tried to mimic [`has_budget`](https://github.com/ggml-org/llama.cpp/blob/c17dce4f5c725054e10cfafc883cbd24e8b6303b/tools/server/server-context.cpp#L237).

tools/server/tests/unit/test_chat_truncate.py
```diff
@@ -0,0 +1,287 @@
+import pytest
```
ltoniazzi, line 2: These tests take 45s for me (locally).

Should I remove some? Minimise n_predict? Collect multiple tests in one to reuse the server?

Let me know

tools/server/server-common.h
```diff
@@ -369,3 +385,33 @@ server_tokens format_prompt_rerank(
         mtmd_context * mctx,
         const std::string & query,
         const std::string & doc);
+
+//
+// Chat truncation utils
+//
+
+// Calculate the target number of tokens to keep in the chat history as the floor of context size * the chat_truncate fraction.
```
ltoniazzi: I think these docstrings are a bit long compared to other ones in the repo. Should I shorten them?

tools/server/server-task.cpp
```diff
@@ -205,7 +205,7 @@ task_params server_task::params_from_json_cmpl(
     params.return_tokens    = json_value(data,       "return_tokens",      false);
     params.return_progress  = json_value(data,       "return_progress",    false);
     auto max_tokens         = json_value(data,       "max_tokens",         defaults.n_predict);
-    params.n_predict        = json_value(data,       "n_predict",          json_value(data, "max_completion_tokens", max_tokens));
+    params.n_predict        = get_n_predict(data, max_tokens);
```
ltoniazzi, line 208: @ngxson Not sure about this being the unified method that you requested

tools/server/tests/unit/test_chat_truncate.py
ngxson: This feature only affect the formatted input prompt, it doesn't sense to actually run the completion request.

Just n_predict=0 or 1 is enough to test it. n_ctx can be set to something very low, like 64 tokens only

common/arg.cpp
```diff
@@ -1310,6 +1310,17 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.ctx_shift = value;
         }
     ).set_examples({LLAMA_EXAMPLE_COMPLETION, LLAMA_EXAMPLE_CLI, LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_IMATRIX, LLAMA_EXAMPLE_PERPLEXITY}).set_env("LLAMA_ARG_CONTEXT_SHIFT"));
+    add_opt(common_arg(
+        {"--chat-truncate"}, "N",
```
ngxson: IMO it's quite confusing reading this help message.

Maybe just better to break this down into 2 flags:
- `--chat-truncate` / `--no-chat-truncate` to toggle the feature
- `--chat-truncate-max-keep F` keeping at maximum F*n_ctx_seq tokens after truncation, default to 0.5

tools/server/server-common.h
```diff
@@ -288,6 +300,10 @@ struct server_chat_params {
     bool allow_audio;
     bool enable_thinking = true;
     std::string media_path;
+    int32_t n_ctx_seq;
+    int32_t n_predict = -1;
+    float chat_truncate = -1.0f;
+    const llama_vocab * vocab = nullptr;
```
ngxson: do not add vocab here, it will break sleeping functionality

ngxson (commented):
I need to rethink about the overall structure, it seems to play not very nice with other functionalities like auto-sleep or multimodal

Also, context shifting is quite overlapped with this

aviallon:
I love this feature. So interesting.
PS: would it be possible to return something about what message (by index) were truncated in the `usage` field?

ltoniazzi:
@ngxson Update and a question.

#### Done
- I updated vocab to use the refreshed pointer after awaking
- Split the param `chat_truncate` into two

#### In progress
Fixing performance issue: 

Previous method: 
    - Tokenize `N` turns -> truncate 1 turn -> tokenize `N-1` turns -> truncate 1 turn -> ...
    - This is `O(N^2)` in turns and slowed down ~10s when truncating more than 70 turns on test models. 
    
Current fix: 
    - Estimate a batch of K turns to truncate in one shot by: tokenize each turn until threshold reached -> `O(N)`
    - This fix's token estimation is not exact as we tokenize messages alone and not in the full prompt, but error should be very small

### Question
For multimedia I am trying a similar approach, estimating turn by turn the number of tokens from text + img/audio, while dropping img/audio from the `out_files`. A first draft seems to work. What issues are you expecting with multimodal beside:
    - the more complicated token counting
    - keeping aligned the `out_files` indices with the `<__media__>` placeholders
    - count with `n_pos`
   ?

ngxson:
I'm not 100% sure about your idea of the estimation. A binary search approach where we truncate half the token at a time will have O(N log N) complexity while being more accurate. Though, it's better to tell if we test with real conversation data.

Re. multimodal, the problem is that multimodal tokenization also preprocesses the input which can be extremely slow. I will need to add a tokenization API which doesn't do that

ngxson:
> For multimedia I am trying a similar approach, estimating turn by turn the number of tokens from text + img/audio, while dropping img/audio from the `out_files`.

That won't work on models supporting dynamic resolution like qwen, there are always cases where user can enter a large image on one message and a small image on another message.

IMO best not to estimate anything, otherwise it can be too inaccurate in real-world scenarios. There is always a gap between what's possible in theory and what's possible in practice.

ltoniazzi:
> I'm not 100% sure about your idea of the estimation. A binary search approach where we truncate half the token at a time will have O(N log N) complexity while being more accurate. Though, it's better to tell if we test with real conversation data.

Sorry, here by "estimation" I meant real tokenization of a full turn.

```cpp
int n_turns_to_erase = 0;
int cumul_tokens = 0;

while cumul_tokens <= target_tokens_to_remove:
   cumul_tokens += common_tokenize(prompts_turns[n_turns_to_erase]).size(); // Real count for each turn
   n_turns_to_erase ++;

// Erase n_turns_to_erase in one go
...
```

I called it estimation as the sum `cumul_tokens` does not have to exactly match the number of tokens of the full prompt of the `n_turns_to_erase`. 
But should be very close to the real value.

---

> That won't work on models supporting dynamic resolution like qwen, there are always cases where user can enter a large image on one message and a small image on another message.

So also here I am computing the real turn tokens, so a large image will result in a large (almost exact) increase in `cumul_tokens`

---

> Re. multimodal, the problem is that multimodal tokenization also preprocesses the input which can be extremely slow. I will need to add a tokenization API which doesn't do that

Also with this `O(N)` strategy we would have to compute preprocessing only once, as then that token count be added to the respective message, right?

ngxson:
`common_tokenize` cannot process raw messages from API, it must go though chat template formatting.
- How do you format each turn separately?
- How do you deal with built-in template system prompt and other added elements inside the formatted template?

ltoniazzi:
> - How do you format each turn separately?
> - How do you deal with built-in template system prompt and other added elements inside the formatted template?

I am creating a new `common_chat_templates_inputs` for each specific turn -> apply template to it -> count tokens.  Something like:
```cpp
    const auto turns = chat_build_droppable_turns(inputs);  // Collects turns indices
    int32_t    accumulated = 0;
    size_t     n_drop      = 0;
    for (size_t t = 0; t < turns.size() && accumulated < deficit; ++t) {
        common_chat_templates_inputs inputs_turn;  // Collect turn's messages here
        for (size_t i = turns[t].msg_first; i < turns[t].msg_first + turns[t].msg_count; ++i)
            {
                inputs_turn.messages.push_back(inputs.messages[i]);
            }
        // Template full turn
        common_chat_params chat_params_turn = common_chat_templates_apply(tmpls, inputs_turn);
        accumulated += (int32_t)common_tokenize(vocab, chat_params_turn.prompt, true, true).size();
        ++n_drop;  // Count how many turns we will batch-drop
    }

    // Batch-drop the selected turns in a single vector operation.
    if (n_drop > 0) {
        inputs.messages.erase(
            inputs.messages.begin() + (ptrdiff_t)turns[0].msg_first,
            inputs.messages.begin() + (ptrdiff_t)(turns[n_drop - 1].msg_first + turns[n_drop - 1].msg_count));
    }
```

ngxson:
But you haven't replied to my second question above

ltoniazzi:
> How do you deal with built-in template system prompt and other added elements inside the formatted template?

I see! It's horrendous to manage extra tokens from system prompts defaults approaches for different templates. 

Agree bifurcation O(N*logN) is the best approach to try at this stage. I'll build a bit of a performance analysis for it.

Question: If we implement the bifurcation, are there other issues beside counting the `<__media__>` placeholder value correctly/efficiently?

ltoniazzi:
Got the estimates below with the expected behaviour. 

<details>
<summary>Gemma text only truncation time estimate</summary>

</br>


- Posting to `"/chat/completions"` with `-n=1`, `n_ctx=2048`

- Each turn is:
  ```python
  user_content = "[Uxx] Please explain topic x in detail." * 50
  assistant_content = "[Axx] Here is my explanation of topic x." * 50
  ```


- `ggml-org/gemma-3-1b-it-GGUF:Q4_K_M`



n_turns | prompt_tokens | prompt_tokens_trunc | turns_removed | time (s)
-- | -- | -- | -- | --
1 | 1 568 | 1 568 | 0 | 1.26
5 | 7 816 | 1 568 | 4 | 1.58
10 | 15 626 | 1 568 | 9 | 1.75
50 | 78 106 | 1 568 | 49 | 3.00
100 | 156 206 | 1 568 | 99 | 4.38
200 | 312 406 | 1 568 | 199 | 7.43

<br class="Apple-interchange-newline">


- `ggml-org/gemma-3-4b-it-GGUF:Q4_K_M`

n_turns | prompt_tokens | prompt_tokens_trunc | turns_removed | time (s)
-- | -- | -- | -- | --
1 | 1 568 | 1 568 | 0 | 4.89
5 | 7 816 | 1 568 | 4 | 5.41
10 | 15 626 | 1 568 | 9 | 5.62
50 | 78 106 | 1 568 | 49 | 6.95
100 | 156 206 | 1 568 | 99 | 8.40
200 | 312 406 | 1 568 | 199 | 11.74

<br class="Apple-interchange-newline">


</details>

Looks fine to me, I think I'll work on the media processing. 
Let me know if you want me to test text-only mode in more detail.

ngxson:
that sounds a bit wrong to me, both the 1B and 4B model uses exactly the same tokenizer, the results are expected to be exactly the same. your test results should be affected by something else.

maybe you should use the `/apply-template` endpoint instead

ngxson:
> Question: If we implement the bifurcation, are there other issues beside counting the `<__media__>` placeholder value correctly/efficiently?

I don't see any other issues for now. compared to the "naive" approach (removing one turn each time), it should make no difference to the end-result. in other words, it's similar to bubble sort vs quick sort - different algo, different speed, but same result

in later stage, we can optimize the multimodal handling by simply allow the `mtmd_tokenize` to skip the pre-processing. but for now, let's simplify things by re-using the same chat formatting / tokenization pipeline without any modifications

ltoniazzi:
> that sounds a bit wrong to me, both the 1B and 4B model uses exactly the same tokenizer, the results are expected to be exactly the same. your test results should be affected by something else.

Yes, their difference is due to the decoding step, both scale as expected in number of tokens (thanks for pointing to `/apply-template`)

ltoniazzi:
I've updated the code.

For text only the count should be exact. If we like this approach I can give it a final clean.


A couple questions/observations:

1. If the **user keeps sending the full messages**, even after truncation, we will truncate again, so the user pays this truncation latency at every turn wastefully. An option is to report back a parameter, as suggested by @aviallon, to tell the user to drop certain turns. Anyway, what should we do?

2. **For `ctx-shift`**, if `n_predict>0`, then `chat-truncate` will guarantee we never hit the context. While if `n_predict=-1`, then `ctx-shift` can allow to complete response even if ctx is crossed. I feel `ctx-shift` could be dropped, to make it less confusing as if a user wants to make sure they avoid a generation being suddenly stopped, then they can pass `n_predict>0`. The subsequent turn they will need truncation anyway. What do you think?


3. **For media** I am processing the media once, then performing an estimate. The issue is that `<__media__>` will be tokenized by each tokeniser differently, so we do not know if it will give 1, 2, ..., or 11 tokens.  While media re-placement will bring 2 special tokens around the media tokens.

     So the current [quick and dirty solution is to estimate the token error](https://github.com/ggml-org/llama.cpp/pull/19841/changes#diff-562aee47dd99ff76cbd0c2a3e9b98d30cdf4d0111c0f09f345130e1f096d7ef4R2218) of `<__media__>` - minus media special tokens, and set up a guard to make sure it's small (abs<= 5 now). Then if negative, we correct, to avoid undershooting our estimate (otherwise we risk not removing enough turns). Overshooting by a little bit per media should result in dropping at most 1 extra message then needed, so we can tolerate it.

     I agree a better solution is to allow the mdtm tokeniser to cache the image preprocessing, and to leave it to another PR

4. I feel the new methods now are complex enough to benefit from a couple of unit tests, rather than just e2e python tests?

Let me know

ngxson:
> 3\. **For media** I am processing the media once, then performing an estimate. The issue is that `<__media__>` will be tokenized by each tokeniser differently, so we do not know if it will give 1, 2, ..., or 11 tokens.  While media re-placement will bring 2 special tokens around the media tokens.

I don't know what you (or your AI) is talking about, but I believe this is well-documented inside mtmd library.

Don't trust the AI-generated code, verify the documentation yourself

ltoniazzi:
Apologies it felt I've been pushing AI slop. I am mortified. Any sloppy idea is my very own, I did my best to test/debug/explore/experiment before any push.
I asked 3 because I got stuck finding a clean way to count without `process_mtmd_prompt` while re-using apis from mdmt.

My confusion with 3 current strategy is:
- Run `process_mtmd_prompt` once to get the real total count + single images count (to reuse later)
- Then, to not run `process_mtmd_prompt` for each candidate truncation, do: 
   1. Tokenise the raw text messages  `common_chat_templates_apply -> common_tokenize` (this includes the `media_marker`s in the prompt `common_tokenize` receives), 
   2. Compensate for error due to `media_marker` tokenisation in step (i)
   3. Add single images count, 
  
    After step (i) do not see how to avoid the compensation of step (ii) because:
        - looks to me that the tokenisation of `common_tokenize` for `media_marker` is not handled in any special way, and if true, then the count depends on the tokeniser, and `<` of `<__media__>` might be merged with text tokens.

Another option is to split all media `media_marker`s and replay the logic of mdmt utils to perform (iii) and get the exact count. 
But I wanted to check with you as it might be overkill if it will be soon replaced by allowing `process_mtmd_prompt` to skip preprocessing images multiple times.

PR #16857, opened 2025-10-30
CUDA: add expert reduce kernel

am17an:
This PR adds a kernel to fuse the common MoE mul + (n_expert_used-1)*add operations into one. 1-2% TG speed-up depending on num_expert_used

Tested on a 4090

| Model                 | Test   |   t/s master |   t/s expert-reduce |   Speedup |
|:----------------------|:-------|-------------:|--------------------:|----------:|
| gpt-oss 20B MXFP4 MoE | tg32   |       198.21 |              200.64 |      1.01 |
| gpt-oss 20B MXFP4 MoE | tg64   |       195.64 |              197.59 |      1.01 |
| gpt-oss 20B MXFP4 MoE | tg128  |       193.52 |              194.77 |      1.01 |
| qwen3moe 30B.A3B Q4_0 | tg32   |       167.77 |              171.98 |      1.03 |
| qwen3moe 30B.A3B Q4_0 | tg64   |       161.53 |              165.00 |      1.02 |
| qwen3moe 30B.A3B Q4_0 | tg128  |       159.33 |              162.50 |      1.02 |

PR #19850, opened 2026-02-24
ggml-webgpu: Support non-contiguous `src0` and overlapping `src0/src1` in binary ops

yomaytk:
Hello. This PR improves the handling of binary operations in the WebGPU backend, adding support for patterns required by #16857 (MoE expert reduce).

The changes are as follows:
- The index is now calculated based on stride to support cases where `src0` is a non-contiguous tensor.
- This supports cases where `src0` and `src1` have overlapping byte ranges. To avoid WebGPU validation errors from aliased `storage` bindings, a single merged buffer region containing both `src0` and `src1` is passed to the shader.
- Added a test case for situations where `src0` and `src1` overlap.

Since https://github.com/ggml-org/llama.cpp/pull/16857 has been reverted for now, I made these changes in case the feature is reintroduced in the future.


yomaytk:
It seems that the EditorConfig check is failing. I will fix it later (already done).

reeselevine:
Thanks for getting around to this, and adding the backend tests to support it. Right now they actually will pass regardless since we give the `skip_validation` flag to Dawn native (there are still some issues with other operations and overlapping buffers, particularly FlashAttention k/v cache buffers). Did you test this with `skip_validation` removed to make sure the changes are actually working natively?

I also am not sure if anyone is really in charge of `test-backend-ops` now that @slaren is taking a break. My only comment here would be to ask if it makes sense to try to integrate these overlapping buffers into the existing `test_bin_bcast` struct, instead of duplicating some of the code for the new test case struct. If there's a reason this seems complicated/infeasible, let me know.

yomaytk:
Thank you for the review!

> Did you test this with skip_validation removed to make sure the changes are actually working natively?

Yes, I tested the four instructions `ADD, SUB, MUL, DIV` without `skip_validation` in my environment and confirmed that they pass. Also, when I ran the new test with the existing WebGPU binary ops implementation without `skip_validation`, I got the following error. So the added code seems to work.

```bash
ggml_webgpu: Device error! Reason: 2, Message: Writable storage buffer binding aliasing found between [BindGroup "ADD_f32"] set at bind group index 0, binding index 0, and [BindGroup "ADD_f32"] set at bind group index 0, binding index 1, with overlapping ranges (offset: 4096, size: 4352) and (offset: 4352, size: 4352) in [Buffer "tensor_buf352"].
 - While encoding [ComputePassEncoder (unlabeled)].DispatchWorkgroups(1, 1, 1).
 - While finishing [CommandEncoder (unlabeled)].
```

> I also am not sure if anyone is really in charge of test-backend-ops now that @slaren is taking a break. My only comment here would be to ask if it makes sense to try to integrate these overlapping buffers into the existing test_bin_bcast struct, instead of duplicating some of the code for the new test case struct. If there's a reason this seems complicated/infeasible, let me know.

I understand.
To test two overlapping tensors, I prepare two tensor views with the same shape whose buffers overlap. To do this, I created a larger base tensor and then created two views that represent parts of that base tensor, and the test then performed a binary operation on those views. On the other hand, `test_bin_bcast` only uses tensors with the same shape and does not use views.
Then, at first, I considered adding this to `test_bin_bcast`, but since the logic is a bit different, I thought it would be clearer to create a separated test. Also, I thought it might be good not to change the existing tests to avoid unexpected failure.

> there are still some issues with other operations and overlapping buffers, particularly FlashAttention k/v cache buffers

I see. I was a bit curious about FlashAttention being disabled in the Emscripten environment. I haven’t looked into the details yet, but are the overlapping k/v cache buffers the main reason for that?

reeselevine:
> Then, at first, I considered adding this to test_bin_bcast, but since the logic is a bit different, I thought it would be clearer to create a separated test. Also, I thought it might be good not to change the existing tests to avoid unexpected failure.

I have a preference towards still trying to include this in the existing test struct. From what I know, that's the path for other operators—new options for the operators are included as parameters in the test case generation, not new test types. Also, I think the CI should catch any issues with other backends. Can you try this out? If it really doesn't work, then we can stick with the current approach.

> I see. I was a bit curious about FlashAttention being disabled in the Emscripten environment. I haven’t looked into the details yet, but are the overlapping k/v cache buffers the main reason for that?

The reason for this is actually just that the current FlashAttention implementation requires subgroup matrices, which as far as I'm aware are still not standard/included in most browser's WebGPU implementations. Eventually, they hopefully will be, and/or it would be great if someone contributed a FlashAttention implementation not using them, so we could choose different versions of the shader depending on what the browser supports.

yomaytk:
I see. OK, I will modify the test to use the existing test case struct.

tests/test-backend-ops.cpp
```diff
@@ -3008,6 +3009,8 @@ struct test_bin_bcast : public test_case {
 
                 b[i] = ggml_new_tensor_4d(ctx, type, ne[p[0]], ne[p[1]], ne[p[2]], ne[p[3]]);
                 b[i] = ggml_permute(ctx, b[i], p[0], p[1], p[2], p[3]);
+            } else if (src_overlap) {
+                b[i] = ggml_view_4d(ctx, a, ne[0], ne[1], ne[2], 2 * (ne[3] / 3), a->nb[1], a->nb[2], a->nb[3], (ne[3] / 3) * a->nb[3]);
```
yomaytk, line 3013: `b[i]` covers the last 2/3 and `out` covers the first 2/3 of dim3, so they overlap in the middle third.

yomaytk:
I modified the test.

I confirmed that the changes in this PR pass the tests (`ADD, SUB, MUL, DIV`) without `skip_validation`. I also confirmed that the existing implementation fails without `skip_validation` with the same error as mentioned above. This fix requires fewer changes and looks better.

Could you please review it again?

reeselevine (commented):
great thanks, looks good now!

reeselevine (approved):

PR #19853, opened 2026-02-24
feat: async pipeline-parallel scheduler + parallel model loading (multi-GPU layer split)

mxxm-t:
This PR improves multi‑GPU layer-split execution by moving the pipeline scheduler toward event-driven, non-blocking coordination and by reducing CPU-side stalls during transfers and submission.

What changed
Async CPU→GPU transfers
Prefer asynchronous host→device transfers when supported, so scheduling can continue while the transfer is in flight.
Non-blocking pipeline scheduling
Reduce CPU-side waiting between split submissions by using streams/events for cross-split dependencies, allowing GPUs to self-synchronize while the CPU prepares the next microbatch.
Queue-throttle for stability
Add a lightweight throttle to prevent the CPU from getting too far ahead of GPU execution, helping keep submission latency stable under fully async pipelining.
Parallel multi-GPU model loading
Add --threads-load/-tl N to load tensors for multiple GPU contexts in parallel (with env-based controls to limit/disable).
ROCm/HIP: disable async upload path to avoid known hipStreamPerThread/event issues.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2353,6 +2353,22 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_env("LLAMA_ARG_SPLIT_MODE"));
     add_opt(common_arg(
+        {"-tl", "--threads-load"}, "N",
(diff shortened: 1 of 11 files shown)


mxxm-t:
Needs few changes before its ready

PR #19875, opened 2026-02-25
ggml/gguf : fix Windows regression due to `ftell` #19874 

naco-siren:
### Bug
As https://github.com/ggml-org/llama.cpp/issues/19874 explains, after commit https://github.com/ggml-org/llama.cpp/commit/418dea39cea85d3496c8b04a118c3b17f3940ad8, any GGUF model file larger than 2GB fails to load ___on Windows___ with: 
```
llama_params_fit: encountered an error while trying to fit params to free device memory: failed to load model
gguf_init_from_file_impl: failed to read magic
```

### Root cause
The newly added `nbytes_remain()` function in `gguf.cpp` (L321) uses `ftell()` and `fseek()`, which return `long`. 
However, on MSVC (both x86 and x64), `long` is always 32-bit. For files > 2GB, `ftell()` overflows and returns -1, causing this `nbytes_remain()` to return 0, which then fails the subsequent array/string size validation checks added in the same commit.
> This does not affect Linux/macOS where `long` is 64-bit on 64-bit systems though.

### Fix
Use `ftelli64` / `_fseeki64` on Windows, esp when the codebase is already using this pattern, e.g. `examples\convert-llama2c-to-ggml\convert-llama2c-to-ggml.cpp`  (L457, L467)

### DIFF
--- a/ggml/src/gguf.cpp
+++ b/ggml/src/gguf.cpp
@@ -319,22 +319,37 @@ struct gguf_reader {
 
     // remaining bytes in the file
     uint64_t nbytes_remain() const {
+#if defined(_WIN32)
+        const __int64 cur = _ftelli64(file);
+        if (cur < 0) {
+            return 0;
+        }
+        if (_fseeki64(file, 0, SEEK_END) != 0) {
+            _fseeki64(file, cur, SEEK_END);
+            return 0;
+        }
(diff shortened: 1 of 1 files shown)


naco-siren:
duplicated

PR #19849, opened 2026-02-24
server : support multi-modal context checkpoints

ggerganov:
alt #19747 
fix #19690
fix #19858

This continues the work from #19747 to enable multi-modal checkpointing in `llama-server`. Avoids changes to `libllama`. Fixes `n_past` ambiguity of "number of tokens" and "next position" by tracking the position separately with a new `pos_next` var.

PR #16391, opened 2025-10-02
server : host-memory prompt caching

ggerganov:
target #16440
rel https://github.com/ggml-org/llama.cpp/discussions/16117

Initial version of automatic memory offloading to host memory using an extended logic for minimizing the prompt reprocessing. The host-memory prompt cache acts as "extra slots" with which we can calculate prefix similarity and decide to hot-swap them into the `llama_context` if it would reduce the processing. The cache is stored in regular RAM.

The RAM size that is used for caching prompts has 2 limits:

- Max size in bytes (controlled with new `--cache-ram, -cram` CLI arg)
- Max number of cached tokens (by default, equal to `--context-size`)

The server logs provide detailed prompt cache information each time the cache is updated:

image

- A small QoL improvement is that `update_slots()` now also logs the old and new prompt for each task around `n_past` (up to 10 tokens) so we can have a better understanding what caused the particular choice of the `n_past` value for the new task.

- Setting `LLAMA_SERVER_SLOTS_DEBUG=1` env will make the `/slots` endpoint output a more detailed  output containing the prompt and the generated text of the current or last task. This is useful for debugging purposes.

Note: mtmd workarounds are starting to cause some headaches. For example `server_tokens` is not copyable which complicates the cache logic and makes the prompt caching feature incompatible with mtmd.

## Usage

```bash
# use 8192 MiB of host RAM for caching prompts
llama-server ... --cache-ram 8192

# use as much host RAM is available (i.e. no limit)
llama-server ... -cram -1

# disable prompt caching in RAM
llama-server ... -cram 0
```

## Server refactor

- Replace `server_slot` members with a single `server_task`
- Remove `server_slot.n_predict`
- Remove prompt truncation logic (obsolete and not useful anymore)
- `slot.task` is now `const ptr` to reflect that the task parameters should not change when it is passed to the slot
- Bump default context checkpoints from 3 to 8

## TODOs

- [x] Set memory limit for the host-memory cache from CLI
- [x] Clean-up implementation
- [x] Test with agentic workflows
- [x] Multi-slot tests
- [x] Fix progress report

PR #19877, opened 2026-02-25
server : enable multi-modal prompt caching

ggerganov:
target #19849
cont #16391

We can now clone `server_tokens` so re-enable host-memory prompt caching for multi-modal cases.

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -995,9 +995,6 @@ private:
             // don't update the cache if the slot's context is empty
             update_cache = update_cache && tokens.size() > 0;
 
-            // TODO: mtmd does not support prompt cache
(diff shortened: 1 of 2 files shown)


ngxson (approved):

andyceo:
I switched to gg/server-mtmd-prompt-cache, do git rebase master and build app.

Confirm now prompt caching works! Note that I run with: no-mmproj-offload = on and not try without it

VRAM usage is quite unstable for qwen3.5-27B, but this is not related to that issue. Thank you!

BVEsun:
Thank you very much for [ggerganov](https://github.com/ggerganov) effort to troubleshoot this problem

I am using windows 10 b8157 release cuda 12.4

After this PR, text prompt processing is being handled by the CPU instead of the GPU, which is causing a significant slowdown.

PR #19867, opened 2026-02-24
ggml : fix AMX and improve alignment checks

angt:
Before, the selector was too restrictive and some models fell back to
CPU_REPACK: 

```
perplexity: calculating perplexity over 2 chunks, n_ctx=2048, batch_size=2048, n_seq=1
perplexity: 2.82 seconds per pass - ETA 0.08 minutes
[1]17.3868,[2]22.2199,
Final estimate: PPL = 22.2199 +/- 1.59692

llama_perf_context_print:        load time =     511.56 ms
llama_perf_context_print: prompt eval time =    2529.70 ms /  4096 tokens (    0.62 ms per token,  1619.16 tokens per second)
llama_perf_context_print:        eval time =       0.00 ms /     1 runs   (    0.00 ms per token,      inf tokens per second)
llama_perf_context_print:       total time =    7630.80 ms /  4097 tokens
llama_perf_context_print:    graphs reused =          0
llama_memory_breakdown_print: | memory breakdown [MiB] | total   free    self   model   context   compute    unaccounted |
llama_memory_breakdown_print: |   - Host               |                  845 =   318 +     224 +     302                |
llama_memory_breakdown_print: |   - CPU_REPACK         |                  288 =   288 +       0 +       0                |
llama_memory_breakdown_print: |   - AMX                |                   31 =    31 +       0 +       0                |
```

Now they use AMX directly:

```
perplexity: calculating perplexity over 2 chunks, n_ctx=2048, batch_size=2048, n_seq=1
perplexity: 2.39 seconds per pass - ETA 0.07 minutes
[1]17.2005,[2]21.8220,
Final estimate: PPL = 21.8220 +/- 1.56485

llama_perf_context_print:        load time =     376.37 ms
llama_perf_context_print: prompt eval time =    2159.65 ms /  4096 tokens (    0.53 ms per token,  1896.60 tokens per second)
llama_perf_context_print:        eval time =       0.00 ms /     1 runs   (    0.00 ms per token,      inf tokens per second)
llama_perf_context_print:       total time =    5189.06 ms /  4097 tokens
llama_perf_context_print:    graphs reused =          0
llama_memory_breakdown_print: | memory breakdown [MiB] | total   free    self   model   context   compute    unaccounted |
llama_memory_breakdown_print: |   - Host               |                  845 =   318 +     224 +     302                |
llama_memory_breakdown_print: |   - AMX                |                  319 =   319 +       0 +       0                |
```

PR #19884, opened 2026-02-25
ggml : add fallback to CPU for unsupported ops in scheduler

angt:
I found this while fixing:
```
llama-server -v -hf unsloth/LFM2.5-VL-1.6B-GGUF:Q8_0
```
with this PR: https://github.com/ggml-org/llama.cpp/pull/19867

I guess it's a fallback of the fallback mechanism somehow and might not be the real root cause.

anyway, with that, I have `llama-server` running:

```
...
llama_context: enumerating backends
llama_context: backend_ptrs.size() = 1
sched_reserve: reserving ...
sched_reserve: max_nodes = 1192
sched_reserve: reserving full memory module
sched_reserve: worst-case: n_tokens = 512, n_seqs = 4, n_outputs = 4
graph_reserve: reserving a graph for ubatch with n_tokens =    1, n_seqs =  4, n_outputs =    4
graph_reserve: making n_tokens a multiple of n_seqs - n_tokens = 4, n_seqs = 4, n_outputs = 4
sched_reserve: Flash Attention was auto, set to enabled
graph_reserve: reserving a graph for ubatch with n_tokens =  512, n_seqs =  4, n_outputs =  512
graph_reserve: reserving a graph for ubatch with n_tokens =    4, n_seqs =  4, n_outputs =    4
graph_reserve: reserving a graph for ubatch with n_tokens =  512, n_seqs =  4, n_outputs =  512
sched_reserve:        CPU compute buffer size =   395.01 MiB
sched_reserve: graph nodes  = 549
sched_reserve: graph splits = 1
sched_reserve: reserve took 1.47 ms, sched copies = 1
...
```
and at the end:
```
que    start_loop: terminate
srv    operator(): operator(): cleaning up before exit...
llama_memory_breakdown_print: | memory breakdown [MiB] | total   free    self   model   context   compute    unaccounted |
llama_memory_breakdown_print: |   - Host               |                 3064 =  1169 +    1500 +     395                |
llama_memory_breakdown_print: |   - AMX                |                 1325 =  1325 +       0 +       0                |
~llama_context:        CPU compute buffer size is 395.0137 MiB, matches expectation of 395.0137 MiB
```

### DIFF
--- a/ggml/src/ggml-backend.cpp
+++ b/ggml/src/ggml-backend.cpp
@@ -1127,6 +1127,10 @@ void ggml_backend_sched_split_graph(ggml_backend_sched_t sched, struct ggml_cgra
             *cur_backend_id = tensor_backend_id(node->view_src);
             SET_CAUSE(node, "4.vsrc");
         }
+        if (*cur_backend_id == -1) {
+            *cur_backend_id = sched->n_backends - 1;
+            SET_CAUSE(node, "4.force");
+        }
         for (int j = 0; j < GGML_MAX_SRC; j++) {
             struct ggml_tensor * src = node->src[j];
             if (src == NULL) {


ggerganov:
Not sure this is the proper fix.

> graph_reserve: reserving a graph for ubatch with n_tokens =  512, n_seqs =  4, n_outputs =  512
sched_reserve:        CPU compute buffer size =   395.01 MiB
sched_reserve: graph nodes  = 549
> ...
> llama_memory_breakdown_print: |   - Host               |                 3064 =  1169 +    1500 +     395                |
llama_memory_breakdown_print: |   - AMX                |                 1325 =  1325 +       0 +       0                |

This logs indicate the llama.cpp logic incorrectly decides to create AMX buffers for some of the weights, even though the computation is not supported (hence 0 AMX compute buffers). Rather something in `llama-model.cpp` / `weight_buft_supported()` likely needs to be adjusted to not create the AMX buffers in the first place.

angt:
Looking into `weight_buft_supported` i think i'll need to pass more context info (like n_parallel or n_seqs) to correctly accept or refuse a backends like AMX (so in the llama_model struct).

Using `--parallel 1` works without any patch.

ggerganov:
> Looking into `weight_buft_supported` i think i'll need to pass more context info (like n_parallel or n_seqs) to correctly accept or refuse a backends like AMX (so in the llama_model struct).
> 
> Using `--parallel 1` works without any patch.

It should be possible to infer these from the tensor and src shapes I think?

ggerganov:
> This logs indicate the llama.cpp logic incorrectly decides to create AMX buffers for some of the weights, even though the computation is not supported (hence 0 AMX compute buffers). 

So I think this earlier statement is not correct. Seems like the compute buffer can remain 0 size and the backend to be still used. Atm, I am not sure what is the proper fix.

angt:
when it fails (parrallel > 1) we have (custom logs to debug):

```
weight_buft_supported: tensor=blk.0.shortconv.in_proj.weight, op=MUL_MAT, buft=AMX
weight_buft_supported: tensor=blk.0.shortconv.in_proj.weight, ne = [2048,6144,1,1]
operator(): tensor blk.0.shortconv.in_proj.weight, ne = [2048,6144,1,1] OK
operator(): tensor , ne = [2048,512,1,1] OK
```
so we check AMX with this code, `w->ne[2]=1` and `w->ne[3]=1`:
```
ggml_tensor * b = ggml_new_tensor_4d(ctx, GGML_TYPE_F32, w->ne[0], 512, w->ne[2], w->ne[3]);
op_tensor = ggml_mul_mat(ctx, w, b);
```
which is OK
but when doing the job for real with parallel>1 later:
```
operator(): tensor blk.0.shortconv.in_proj.weight, ne = [2048,6144,1,1] OK
operator(): tensor model.layers.{}.operator_norm-0 (reshaped), ne = [2048,1,4,1] NOT OK
/home/angt/hf/llama.cpp/ggml/src/ggml-backend.cpp:1163: GGML_ASSERT(*cur_backend_id != -1) failed
operator(): tensor blk.0.shortconv.out_proj.weight, ne = [2048,2048,1,1] OK
```

we now have `w->ne[2]=4` so we need to check for the reshaped tensors, taking into account `n_parallel` 🤔

ggerganov:
Got it. It's much simpler to extend AMX ops to not impose limit on dim 2 and 3.

angt:
I was worried but it turns out all the code was already in place to allow `n_parallel` to be used for backend selection.

angt:
> Got it. It's much simpler to extend AMX ops to not impose limit on dim 2 and 3.

Yes this is the best solution. I had assumed it would be harder and wanted to fix this issue.

But the `n_parallel` PR doesn’t seem that hacky and makes sense to me.

ggerganov:
`n_parallel` is context-specific parameter. It's not only simpler - it is also consistent with the logic of `libllama`. In principle for a single model, we can have multiple contexts, each with different `n_parallel`. So it's not OK to impose a restriction at model creation for `n_parallel`.

angt:
That was my initial thought, i was thinking of a max value of `n_parallel` maybe.

Ok, so we should not have a backend that use the `w->ne` in `supports_op()`.

ggerganov:
> Ok, so we should not have a backend that use the w->ne in supports_op().

For example, these usages of `ne[0]` are completely fine:

https://github.com/ggml-org/llama.cpp/blob/9051663d5d67bdab765e9bc9afee41ed34f09260/ggml/src/ggml-metal/ggml-metal-device.m#L1132-L1151

Conditioning on `ne[2]` and `ne[3]` for `GGML_OP_MUL_MAT` is problematic as these dimensions can be different based on context parameters.

Technically, the `weight_buft_supported()` logic should exercise all potential usages of the weight tensor. For example, we see that this is not enough:

https://github.com/ggml-org/llama.cpp/blob/9051663d5d67bdab765e9bc9afee41ed34f09260/src/llama-model.cpp#L203-L207

It would also have to check support when `b->ne[2]` and `b->ne[3]` is a multiple of `w->ne[2]` and `w->ne[3]`. We can improve this logic separately.

angt:
True, but i still dont know how we could support a backend like AMX that restrict `ne[2]` to 1 in the current design.

so far my options are:
- change `ggml-backend` to correctly report the issue or fallback to CPU (maybe we need to memcpy the tensor).
- use n_parallel or n_seqs in the `weight_buft_supported`.

angt:
I have a batched version for AMX, i'll push tomorrow after more testing so let's close this one

PR #19891, opened 2026-02-25
[HIP] Updates the ROCm/HIP toolchain versions used in CI pipelines

slojosic-amd:
## Summary
Updates the ROCm/HIP toolchain versions used in CI pipelines to bring them up to date.
This PR is an add-on for https://github.com/ggml-org/llama.cpp/pull/19433 and https://github.com/ggml-org/llama.cpp/pull/19810

#### Linux (`ubuntu-22-cmake-hip`)

Updated Docker container from `rocm/dev-ubuntu-22.04:6.1.2` → `rocm/dev-ubuntu-22.04:7.2`
Updated rocWMMA package to `2.2.0.70200` (from `1.7.0.60402`), fetched from the `rocm/apt/7.2` repo

#### Windows (`windows-cmake-hip` / `build-cache`)

Bumped `HIPSDK_INSTALLER_VERSION` from `25.Q3` → `26.Q1` (kept in sync between `build.yml` and `build-cache.yml`)
Updated the HIP SDK installer URL target from `WinSvr2022` → `Win11`

### Notes
Removed the `ROCM_VERSION` env variable — it was previously used for both the rocWMMA apt repo path and the `CMAKE_CXX_FLAGS` include path. However, these two uses require different formats: the apt repo uses `rocm/apt/7.2` while the include directory is `rocm-7.2.0` (with a trailing `.0`). Since a single variable can't satisfy both formats, the values are now hardcoded directly at each usage site.

### Motivation
Keeps the CI environment aligned with the latest AMD ROCm 7.2 / HIP SDK 26.Q1 release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -586,7 +586,7 @@ jobs:
 
   ubuntu-22-cmake-hip:
     runs-on: ubuntu-22.04
-    container: rocm/dev-ubuntu-22.04:7.2
+    container: rocm/dev-ubuntu-22.04:6.1.2
 
     steps:
       - name: Clone
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -586,7 +586,7 @@ jobs:
 
   ubuntu-22-cmake-hip:
     runs-on: ubuntu-22.04
-    container: rocm/dev-ubuntu-22.04:6.1.2
+    container: rocm/dev-ubuntu-22.04:7.2
 
     steps:
       - name: Clone
@@ -1175,10 +1175,8 @@ jobs:
     runs-on: windows-2022
 
     env:
-      # The ROCm version must correspond to the version used in the HIP SDK.
-      ROCM_VERSION: "6.4.2"
       # Make sure this is in sync with build-cache.yml
-      HIPSDK_INSTALLER_VERSION: "25.Q3"
+      HIPSDK_INSTALLER_VERSION: "26.Q1"
 
     steps:
       - name: Clone
@@ -1188,7 +1186,7 @@ jobs:
       - name: Grab rocWMMA package
         id: grab_rocwmma
         run: |
-          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/${{ env.ROCM_VERSION }}/pool/main/r/rocwmma-dev/rocwmma-dev_1.7.0.60402-120~24.04_amd64.deb"
+          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70200-43~24.04_amd64.deb"
           7z x rocwmma.deb
           7z x data.tar
 
@@ -1231,7 +1229,7 @@ jobs:
           cmake -G "Unix Makefiles" -B build -S . `
             -DCMAKE_C_COMPILER="${env:HIP_PATH}\bin\clang.exe" `
             -DCMAKE_CXX_COMPILER="${env:HIP_PATH}\bin\clang++.exe" `
-            -DCMAKE_CXX_FLAGS="-I$($PWD.Path.Replace('\', '/'))/opt/rocm-${{ env.ROCM_VERSION }}/include/" `
+            -DCMAKE_CXX_FLAGS="-I$($PWD.Path.Replace('\', '/'))/opt/rocm-7.2.0/include/" `
             -DCMAKE_BUILD_TYPE=Release `
             -DLLAMA_BUILD_BORINGSSL=ON `
             -DROCM_DIR="${env:HIP_PATH}" `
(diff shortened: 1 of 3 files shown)


CISC (approved):

.github/workflows/build.yml
```diff
@@ -586,7 +586,7 @@ jobs:
 
   ubuntu-22-cmake-hip:
     runs-on: ubuntu-22.04
-    container: rocm/dev-ubuntu-22.04:6.1.2
+    container: rocm/dev-ubuntu-22.04:7.2
```
IMbackK: This one needs to remain 6.1.2. 6.1 is the oldest version we support and building it here serves the purpose of preventing functions added to later versions ending up in llamacpp.

The argument could be made that we should build against rocm 7.2 as well, but the local builds by developers and windows build below pretty mutch cover this.

IMbackK (changes requested):

.github/workflows/build.yml
CISC: Good point.

CISC:
Nice, the ROCm installer still has this bug I see. :)
https://github.com/ggml-org/llama.cpp/actions/runs/22397277428/job/64834068878?pr=19891

slojosic-amd:
> Nice, the ROCm installer still has this bug I see. :) https://github.com/ggml-org/llama.cpp/actions/runs/22397277428/job/64834068878?pr=19891

@CISC Based on raw logs I noticed that CI machine is Windows Server 2022:
 
image

and new HIP SDK 26.Q1 installer is marked as Win11: https://www.amd.com/en/developer/resources/rocm-hub/eula/licenses.html?filename=AMD-Software-PRO-Edition-26.Q1-Win11-For-HIP.exe

Maybe this is the root cause for the ROCm installer problem...

However, on my forked repo the same 26.Q1 installer is working without any problem: https://github.com/slojosic-amd/llama.cpp/actions/runs/22441008001/job/64983482526
image

And Windows version is the same as for official llama.cpp CI job:

image

CISC:
@slojosic-amd Nah, it's always been like this, sometimes the installer just hangs, so we added a timeout.

slojosic-amd:
> @slojosic-amd Nah, it's always been like this, sometimes the installer just hangs, so we added a timeout.

Should we increase timeout?

CISC:
> > @slojosic-amd Nah, it's always been like this, sometimes the installer just hangs, so we added a timeout.
> 
> Should we increase timeout?

No, the timeout is fine, and since we are caching the installation it'll hardly ever be a problem.

hammer-83:
Maybe somebody else with RX 6xxx can confirm but on my RX 6950 the SDK 26.Q1 does not work. It's not a question of llama.cpp integration, simply installing the SDK and running `hipinfo.exe` shows that there is no card detected. And every build since this change was merged is not detecting rocm device in llama.cpp.

I have setup a local build that does detect a rocm device, but it uses [TheRock](https://github.com/ROCm/TheRock) nightly binaries. The last official release of rocm SDK to work with gfx1030 seems to be 6.4

slojosic-amd:
> Maybe somebody else with RX 6xxx can confirm but on my RX 6950 the SDK 26.Q1 does not work. It's not a question of llama.cpp integration, simply installing the SDK and running `hipinfo.exe` shows that there is no card detected. And every build since this change was merged is not detecting rocm device in llama.cpp.
> 
> I have setup a local build that does detect a rocm device, but it uses [TheRock](https://github.com/ROCm/TheRock) nightly binaries. The last official release of rocm SDK to work with gfx1030 seems to be 6.4

@hammer-83 which driver do you have installed? Could you please try with [https://drivers.amd.com/drivers/installer/25.30/whql/amd-software-adrenalin-edition-26.1.1-minimals…](https://drivers.amd.com/drivers/installer/25.30/whql/amd-software-adrenalin-edition-26.1.1-minimalsetup-260119_web.exe)

hammer-83:
@slojosic-amd Driver was the first thing I tried to eliminate. I installed different past and current versions of retail driver as well as experimental Radeon pro drivers that came around same time as 26.Q1 (had pytorch in the name, can find the link of necessary). I think the sdk even packages those drivers in the install.

I'm fairly certain the problem is not in the driver and, on a higher level, I think the solution of not using the mainline driver to be able to use rocm is not viable long-term, especially considering that the current driver works well with 6.4 SDK and with TheRock nightlies.

I spent a good week working on this issue to setup a local build environment and I came to the conclusion that 26.Q1 does not work, at least with gfx1030. So I would advise the maintainers to rollback this change and switch either back to 6.4 or to TheRock, although the latter is still in an experimental phase.

Looks like somebody else with gfx1030 is having the same issue:
https://github.com/ggml-org/llama.cpp/issues/19943

djismgaming:
Indeed. I'm on driver Adrenalin 26.1.1 and exhibit this behavior.

Edit: 

Note: I see the llama.cpp in LM Studio does work with ROCm. Their latest ROCm release notes mention: "llama.cpp release b8175 (commit https://github.com/ggml-org/llama.cpp/commit/8d3b962f47cdb05df8ab8801da38cb3c55d50337)", and it's working on RDNA2

PR #19896, opened 2026-02-25
test-backend-ops: allow loading tests from file and parsing model operators into file

0cc4m:
When working on backends, I often have the problem that a specific operator inside of a model fails. I have to track that down, add a test and then figure out a fix. This change is supposed to make that easier by allowing you to extract operators from a model file in a way that test-backend-ops can run them directly.

I first wanted to allow test-backend-ops to parse them directly, but since it's GGML-specific, I didn't find a good way to do that. Instead, I added the tool `llama-export-graph-ops` to load a model, parse the pp/tg cgraphs and put the operators into a JSON file, which test-backend-ops can load. Let me know if there's a better way to do this that I missed, I had to add llama_graph_reserve to the public API to avoid using internal API headers.

A "generic operator" test also didn't fit as neatly into test-backend-ops as I had hoped, I tried to put all the special error threshold and initialization functions into it in the least intrusive way. Let me know if there's a better way to handle this and the graph extraction that I didn't see.

I plan to expand `llama-export-graph-ops` to allow other sources for tests, for example HF metadata (#19796) could be useful to avoid downloading a model if a backend issue has been reported with it.

Model-specific test-backend-ops may also be useful to identify operators that perform worse than expected inside of models or to compare operator performance across backends.

<details>
<summary>Example for Qwen3 4B Q8_0:</summary>

```
> build/test-backend-ops --test-json qwen3_4b_q8_0.json
ggml_vulkan: 0 = Intel(R) Arc(tm) Graphics (MTL) (Intel open-source Mesa driver) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 65536 | int dot: 1 | matrix cores: none
Testing 2 devices

Backend 1/2: Vulkan0
  Device description: Intel(R) Arc(tm) Graphics (MTL)
  Device memory: 47790 MB (41065 MB free)

  ADD(name=ffn_inp-0,type=f32,ne=[2560,1,1,1],op_params=[],sources=f32[2560,1,1,1],f32[2560,1,1,1]): OK
  ADD(name=ffn_inp-0,type=f32,ne=[2560,512,1,1],op_params=[],sources=f32[2560,512,1,1],f32[2560,512,1,1]): OK
  MUL(name=Kcur_normed-0,type=f32,ne=[128,8,1,1],op_params=[],sources=f32[128,8,1,1],f32[128,1,1,1]): OK
  MUL(name=Kcur_normed-0,type=f32,ne=[128,8,512,1],op_params=[],sources=f32[128,8,512,1],f32[128,1,1,1]): OK
  MUL(name=Qcur_normed-0,type=f32,ne=[128,32,1,1],op_params=[],sources=f32[128,32,1,1],f32[128,1,1,1]): OK
  MUL(name=Qcur_normed-0,type=f32,ne=[128,32,512,1],op_params=[],sources=f32[128,32,512,1],f32[128,1,1,1]): OK
  MUL(name=attn_norm-0,type=f32,ne=[2560,1,1,1],op_params=[],sources=f32[2560,1,1,1],f32[2560,1,1,1]): OK
  MUL(name=attn_norm-0,type=f32,ne=[2560,512,1,1],op_params=[],sources=f32[2560,512,1,1],f32[2560,1,1,1]): OK
  RMS_NORM(name=norm-0,type=f32,ne=[128,8,1,1],op_params=[0:897988541],sources=f32[128,8,1,1]): OK
  RMS_NORM(name=norm-0,type=f32,ne=[128,8,512,1],op_params=[0:897988541],sources=f32[128,8,512,1]): OK
  RMS_NORM(name=norm-0,type=f32,ne=[128,32,1,1],op_params=[0:897988541],sources=f32[128,32,1,1]): OK
  RMS_NORM(name=norm-0,type=f32,ne=[128,32,512,1],op_params=[0:897988541],sources=f32[128,32,512,1]): OK
  RMS_NORM(name=norm-0,type=f32,ne=[2560,1,1,1],op_params=[0:897988541],sources=f32[2560,1,1,1]): OK
  RMS_NORM(name=norm-0,type=f32,ne=[2560,512,1,1],op_params=[0:897988541],sources=f32[2560,512,1,1]): OK
  MUL_MAT(name=Vcur-0,type=f32,ne=[1024,1,1,1],op_params=[],sources=q8_0[2560,1024,1,1],f32[2560,1,1,1]): OK
  MUL_MAT(name=Vcur-0,type=f32,ne=[1024,512,1,1],op_params=[],sources=q8_0[2560,1024,1,1],f32[2560,512,1,1]): OK
  MUL_MAT(name=node_28,type=f32,ne=[2560,1,1,1],op_params=[],sources=q8_0[4096,2560,1,1],f32[4096,1,1,1]): OK
  MUL_MAT(name=ffn_out-0,type=f32,ne=[2560,1,1,1],op_params=[],sources=q8_0[9728,2560,1,1],f32[9728,1,1,1]): OK
  MUL_MAT(name=node_28,type=f32,ne=[2560,512,1,1],op_params=[],sources=q8_0[4096,2560,1,1],f32[4096,512,1,1]): OK
  MUL_MAT(name=ffn_out-0,type=f32,ne=[2560,512,1,1],op_params=[],sources=q8_0[9728,2560,1,1],f32[9728,512,1,1]): OK
  MUL_MAT(name=Qcur-0,type=f32,ne=[4096,1,1,1],op_params=[],sources=q8_0[2560,4096,1,1],f32[2560,1,1,1]): OK
  MUL_MAT(name=Qcur-0,type=f32,ne=[4096,512,1,1],op_params=[],sources=q8_0[2560,4096,1,1],f32[2560,512,1,1]): OK
  MUL_MAT(name=ffn_gate-0,type=f32,ne=[9728,1,1,1],op_params=[],sources=q8_0[2560,9728,1,1],f32[2560,1,1,1]): OK
  MUL_MAT(name=ffn_gate-0,type=f32,ne=[9728,512,1,1],op_params=[],sources=q8_0[2560,9728,1,1],f32[2560,512,1,1]): OK
  MUL_MAT(name=result_output,type=f32,ne=[151936,1,1,1],op_params=[],sources=q8_0[2560,151936,1,1],f32[2560,1,1,1]): OK
  MUL_MAT(name=result_output,type=f32,ne=[151936,512,1,1],op_params=[],sources=q8_0[2560,151936,1,1],f32[2560,512,1,1]): OK
  CPY(name= (copy),type=f16,ne=[4096,1,1,1],op_params=[],sources=f32[4096,1,1,1],f16[4096,1,1,1]): OK
  CPY(name= (copy),type=f16,ne=[4096,512,1,1],op_params=[],sources=f32[4096,512,1,1],f16[4096,512,1,1]): OK
  GET_ROWS(name=node_1254,type=f32,ne=[2560,1,1,1],op_params=[],sources=f32[2560,1,1,1],i32[1,1,1,1]): OK
  GET_ROWS(name=embd,type=f32,ne=[2560,1,1,1],op_params=[],sources=q8_0[2560,151936,1,1],i32[1,1,1,1]): OK
  GET_ROWS(name=node_1254,type=f32,ne=[2560,512,1,1],op_params=[],sources=f32[2560,512,1,1],i32[512,1,1,1]): OK
  GET_ROWS(name=embd,type=f32,ne=[2560,512,1,1],op_params=[],sources=q8_0[2560,151936,1,1],i32[512,1,1,1]): OK
  SET_ROWS(name=cache_k_l0 (view),type=f16,ne=[1024,4096,1,1],op_params=[],sources=f32[1024,1,1,1],i64[1,1,1,1],f16[1024,4096,1,1]): OK
  SET_ROWS(name=cache_k_l0 (view),type=f16,ne=[1024,4096,1,1],op_params=[],sources=f32[1024,512,1,1],i64[512,1,1,1],f16[1024,4096,1,1]): OK
  ROPE(name=Kcur-0,type=f32,ne=[128,8,1,1],op_params=[1:128,2:2,4:262144,5:1251513984,6:1065353216,8:1065353216,9:1107296256,10:1065353216],sources=f32[128,8,1,1],i32[1,1,1,1]): OK
  ROPE(name=Kcur-0,type=f32,ne=[128,8,512,1],op_params=[1:128,2:2,4:262144,5:1251513984,6:1065353216,8:1065353216,9:1107296256,10:1065353216],sources=f32[128,8,512,1],i32[512,1,1,1]): OK
  ROPE(name=Qcur-0,type=f32,ne=[128,32,1,1],op_params=[1:128,2:2,4:262144,5:1251513984,6:1065353216,8:1065353216,9:1107296256,10:1065353216],sources=f32[128,32,1,1],i32[1,1,1,1]): OK
  ROPE(name=Qcur-0,type=f32,ne=[128,32,512,1],op_params=[1:128,2:2,4:262144,5:1251513984,6:1065353216,8:1065353216,9:1107296256,10:1065353216],sources=f32[128,32,512,1],i32[512,1,1,1]): OK
  FLASH_ATTN_EXT(name=__fattn__-0,type=f32,ne=[128,32,1,1],op_params=[0:1035273459,3:10],sources=f32[128,1,32,1]nb[4,16384,512,16384],f16[128,4096,8,1]nb[2,2048,256,8388608],f16[128,4096,8,1]nb[2,2048,256,8388608],f16[4096,1,1,1]): OK
  FLASH_ATTN_EXT(name=__fattn__-0,type=f32,ne=[128,32,512,1],op_params=[0:1035273459,3:10],sources=f32[128,512,32,1]nb[4,16384,512,8388608],f16[128,4096,8,1]nb[2,2048,256,8388608],f16[128,4096,8,1]nb[2,2048,256,8388608],f16[4096,512,1,1]): OK
  SWIGLU(name=ffn_swiglu-0,type=f32,ne=[9728,1,1,1],op_params=[0:2],sources=f32[9728,1,1,1],f32[9728,1,1,1]): OK
  SWIGLU(name=ffn_swiglu-0,type=f32,ne=[9728,512,1,1],op_params=[0:2],sources=f32[9728,512,1,1],f32[9728,512,1,1]): OK
  42/42 tests passed
  Backend Vulkan0: OK
Backend 2/2: CPU
  Skipping CPU backend
2/2 backends passed
OK
```

</details>

Claude Code was used to assist, but I wrote and tested the code.


0cc4m:
@ggerganov @CISC Could one of you take a look at this?

ggerganov:
I think using JSON here is not really necessary - would prefer to avoid it. Simple ad-hoc data read/write should be OK.

0cc4m:
You mean a simple binary format? Sure, I can do that.

ggerganov:
Yes, either simple binary, or even text is fine.

0cc4m:
Alright, switched to a simple text file format.

0cc4m:
@ggerganov Is it okay like this?

include/llama.h
```diff
@@ -617,6 +617,13 @@ extern "C" {
             const char * fname_out,
             const llama_model_quantize_params * params);
 
+    // Reserve a new compute graph. It is valid until the next call to llama_graph_reserve.
+    LLAMA_API struct ggml_cgraph * llama_graph_reserve(
+            struct llama_context * ctx,
+            uint32_t n_tokens,
+            uint32_t n_seqs,
+            uint32_t n_outputs);
+
```
ggerganov: I'm wondering if we should mark this function experimental/unstable in some way, so that it does not affect the official `libllama` ABI?

0cc4m: Is there a better way to make this function accessible? The model loading isn't actually required here, it was just the easiest way to grab the cgraph. Maybe some kind of model introspection callback?

ggerganov: Given that https://github.com/ggml-org/llama.cpp/pull/19802 also needs to access some internals of `libllama`, I think we can create a an internal header called `llama-ext.h` and move this new function there in `src/llama-ext.h`. I.e. it will not be exposed publicly and won't be part of the official API of the library. The `llama.cpp` tests can access it by crossing the C-style boundary and including it directly from `src`:

```c++
#include "../src/llama-ext.h"
```

The `export-graph-ops.cpp` would have to become a `test` binary rather than a `tool`.

0cc4m: Alright, I'll give that a shot.

0cc4m: Done, do you think the function definition should stay in llama-context or move into a new `llama-ext.cpp` file?

ggerganov: Either way is fine - up to you.

src/llama-ext.h
```diff
@@ -0,0 +1,15 @@
+#ifndef LLAMA_EXT_H
+#define LLAMA_EXT_H
```
ggerganov: `#pragma once` is more consistent with the codebase.

ggerganov (approved):
~Yes, this is better.~

~I'm still wondering if we should mark the new `llama_graph_reserve` as experimental/unstable in some way? With the upcoming llama.cpp packages (#20042) we should be mindful how we change the public API and minimize the changes. This function seems quite~ 

Sorry, some leftover partial comment from earlier - ignore.

include/llama.h
0cc4m: I think it's okay where it is.

0cc4m:
@ggerganov It's failing to build on Windows because llama-ext.h does not export the function with `__declspec(dllexport)`. What's the right fix for this, add `LLAMA_API` to the function in the ext-header, or would that get us back to the issue that the function shouldn't be part of the public API?

ggerganov:
Yes, add `LLAMA_API`. It not a big issue because a 3rd party using `libllama` would only see the declarations from `llama.h` and not from `llama-ext.h` (because it is private header).

vt-alt:
Does it supposed to be installed as `export-graph-ops`? Not as `llama-export-graph-ops` as `tests/test-backend-ops.cpp` mention it, and if this is testing helper so why install at all?

0cc4m:
I suppose it could be called `test-export-graph-ops`, but it's not a full test. It's being treated like all other tests, you don't have to install them.

vt-alt:
Thanks for clarifying. ~~I don't install it, it is installed.~~

vt-alt:
~~Perhaps, it's installed due to using `llama_build` unlike other tests using `llama_build_and_test`.~~

vt-alt:
Ah yeah I now looked deeper, we (for ALT Linux) delete test binaries with `test-*` pattern and `export-graph-ops` is not deleted, so `llama_build` is unrelated (sorry) and only missing `test-` prefix matters in our case.

Excuse me, llama.cpp build process is so compilated to comprehend everything from the first glance.

0cc4m:
Oh, I wasn't aware of that. @ggerganov Should I rename the binary?

PR #19916, opened 2026-02-26
ggml-cuda: add mem check for fusion

am17an:
Fixes #19659

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -3366,6 +3366,69 @@ static bool ggml_cuda_can_fuse(const struct ggml_cgraph *                cgraph,
     return false;
 }
 
+// returns whether the write (out) nodes overwrite the read nodes in operation
+static bool ggml_cuda_check_fusion_memory_ranges(ggml_cgraph * cgraph,
+                                                 int           node_idx,
+                                                 int           node_count,
+                                                 int *         out_nodes,
+                                                 int           out_count) {
+    auto nodes_overlap = [&](const ggml_tensor * a, const ggml_tensor * b) {
+        const int64_t a_start = (int64_t) a->data;
+        const int64_t a_end   = a_start + ggml_nbytes(a);
+
+        const int64_t b_start = (int64_t) b->data;
+        const int64_t b_end   = b_start + ggml_nbytes(b);
+
+        if ((b_start <= a_start && a_start < b_end) || (a_start <= b_start && b_start < a_end)) {
+            return true;
+        }
+
+        return false;
+    };
+
+    bool is_ok = true;
+    // for nrows=1, all fusion operation correctly read the src before writing dst or do it elementwise, so we should be ok
+    if (ggml_nrows(cgraph->nodes[node_idx]) == 1) {
+        return true;
+    }
+
+    for (int i = 0; i < out_count; ++i) {
+        const ggml_tensor * dst = cgraph->nodes[out_nodes[i]];
+
+        for (int j = node_idx; j < node_idx + node_count; ++j) {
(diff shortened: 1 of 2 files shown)


ggml/src/ggml-cuda/topk-moe.cu
```diff
@@ -119,6 +119,18 @@ __launch_bounds__(4 * WARP_SIZE, 1) __global__ void topk_moe_cuda(const float *
...
+    // NaN comparisons always return false, which would cause the same expert to be
+    // selected repeatedly. -FLT_MAX compares normally and is still excluded by the
+    // -INFINITY sentinel used after each selection round.
+    // More relevant for the cuBLAS path. See https://github.com/ggml-org/llama.cpp/issues/19659
+#pragma unroll
+    for (int i = 0; i < experts_per_thread; i++) {
+        if (__isnanf(wt[i])) {
+            wt[i] = -FLT_MAX;
+        }
+    }
```
ORippler, line 132: 1. If the issue is in llama.cpp and not cuBLAS, I feel we should use `fmaxf` as a NaN-safe comparator: https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__SINGLE.html#_CPPv45fmaxfff (I presume we are talking about `val_s > max_val_s` later on in this kernel?)
2. If the issue is in cuBLAS, I'd love more details so I can ask the cuBLAS team/take a look myself

am17an: 1. Yes but it's not just `val_s > max_val_s` it's `val_s > max_val_s || (val_s == max_val_s && expert < max_expert)` 
2. The linked issue has a repro. It's cuBLAS + Nemotron, so think it would be fun for you guys to look at :)

ORippler: > Yes but it's not just val_s > max_val_s it's val_s > max_val_s || (val_s == max_val_s && expert < max_expert)

Shouldn't we be fine with fmaxf, so long as `max_val` & `max_val_s` are initialized to -FLT_MAX instead of -INFINITY at the beginning of the selection-loop over `n_expert_used`? At least for the case where k non-NAN values exist inside the logits for a given row. But at this point we are just pulling your proposal into the loop itself 😄

am17an:
@JohannesGaessler can you review this PR? Apart from the NaN check it also fixes a latent bug

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3366,6 +3366,69 @@ static bool ggml_cuda_can_fuse(const struct ggml_cgraph *                cgraph,
...
+        const int64_t a_end   = a_start + ggml_nbytes(a);
+
+        const int64_t b_start = (int64_t) b->data;
+        const int64_t b_end   = b_start + ggml_nbytes(b);
+
+        if ((b_start <= a_start && a_start < b_end) || (a_start <= b_start && b_start < a_end)) {
+            return true;
+        }
+
+        return false;
```
JohannesGaessler, line 3386:
```suggestion
        return (b_start <= a_start && a_start < b_end) || (a_start <= b_start && b_start < a_end);
```

This would maybe be slightly simpler but either way is fine I think.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3366,6 +3366,69 @@ static bool ggml_cuda_can_fuse(const struct ggml_cgraph *                cgraph,
...
+    };
+
+    bool is_ok = true;
+    // for nrows=1, all fusion operation correctly read the src before writing dst or do it elementwise, so we should be ok
```
```suggestion
    // for nrows=1, all fusion operations correctly read the src before writing dst or do it elementwise, so we should be ok
```

JohannesGaessler (approved):

PR #19924, opened 2026-02-26
server : fix ctx checkpoint restore logic

### LINKED ISSUE #19901: Misc. bug: PR19849 causes inconsistent sequence positions
### Name and Version

$ llama-server --version
  Device 0: NVIDIA GeForce GTX 1080 Ti, compute capability 6.1, VMM: yes
version: 8151 (c747294b2)

EDIT: Oops. I didn't know I'd need the "--version" output as part of a bug report, so I'd already rebuilt my binaries to the latest master without the changes that cause the issue. PR19849 (commit d7d826b3c1fca0c1564a59d92bf6c2c40c8e69fb) and forward show the problem, the version shown above (from my cli output) does _NOT_ show the problem.

### Operating systems

Linux

llama-server

### Command line

```shell
llama-server \
        --port 8080 --jinja \
        --parallel 1 \
        -hf unsloth/Qwen3-Coder-Next-GGUF:UD-Q4_K_XL \
        --ctx-size 200000 \
        --batch-size 256 \
        --ubatch-size 128 \
        --temp 1.0 \
        --min-p 0.01 \
        --top-p 0.95 \
        --top-k 40 \
        -fitt 1
```

Happens regularly for me by connecting claude code to the server and merely prompting it more than one or two times.

### First Bad Commit

d7d826b3c1fca0c1564a59d92bf6c2c40c8e69fb

### Relevant log output

<details>
<summary>Logs</summary>

```console
[...]
init: the tokens of sequence 0 in the input batch have inconsistent sequence positions:
 - the last position stored in the memory module of the context (i.e. the KV cache) for sequence 0 is X = 43686
 - the tokens for sequence 0 in the input batch have a starting position of Y = 43688
 it is required that the sequence positions remain consecutive: Y = X + 1
decode: failed to initialize batch
llama_decode: failed to decode, ret = -1
[...]
```
</details>

ggerganov:
cont #19849
fix #19901

Prevent `n_past` from exceeding the available tokens in the checkpoint.

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -2363,7 +2363,7 @@ private:
                                             //printf("[DEBUG] `do_reset` was set to `true` after failing to restore a checkpoint");
                                         } else {
                                             pos_next = std::min(pos_next, std::max(it->pos_min + 1, it->pos_max));
-                                            n_past = slot.prompt.tokens.size_up_to_pos(pos_next);
+                                            n_past = std::min(slot.prompt.tokens.size_up_to_pos(pos_next), (size_t) it->n_tokens);
                                             SLT_WRN(slot, "restored context checkpoint (pos_min = %d, pos_max = %d, n_tokens = %" PRId64 ", size = %.3f MiB)\n", it->pos_min, it->pos_max, it->n_tokens, (float) checkpoint_size / 1024 / 1024);
                                         }
                                     }


ngxson (approved):

yshui:
this feels wrong? `pos_next` is one past the last token, but `size_up_to_pos` treats `max_pos` as inclusive.

ggerganov:
@yshui Can you clarify a bit more? I think the logic is correct, though indeed it's a bit convoluted and I could have miss something.

yshui:
in that case, i must've misunderstood something. please ignore me.

PR #17470, opened 2025-11-24
server: introduce API for serving / loading / unloading multiple models

ngxson:
Close https://github.com/ggml-org/llama.cpp/issues/16487
Close #16256
Close #17556

**For more detailes on WebUI changes**, please refer to this comment from @allozaur : https://github.com/ggml-org/llama.cpp/pull/17470#issuecomment-3596666360

This PR introduces the ability to use multiple models, unload/load them on the fly in `llama-server`

The API was designed to take advantage of OAI-compat `/v1/models` endpoint, as well as the `"model"` in body payload for POST requests like `/v1/chat/completions`. By default, if the model is not yet loaded, it will be loaded automatically on-demand.

This is the first version of the feature, aims to be **experimental**. Here is the list of capabilities:
- API for listing, loading, unloading models
- Routing request based on `"model"` field
- Limit maximum number of models to be loaded at the same time
- Allow loading models from a local directory
- (Advanced) allow specifying custom per-model config via API

Other features like downloading new models, delete cached models, real-time events, etc are planned for next iteration.

Example commands:

```sh
# start the server as router (using models in cache)
llama-server

# use GGUFs from a local directory - see directory structure in README.md
llama-server --models-dir ./my_models

# specify default arguments to be passed to models
llama-server -n 128 -ctx 8192 -ngl 4

# allow setting the arguments per-model via API (warning: only used in trusted network)
llama-server --models-allow-extra-args
```

For the full documentation, please refer to the ["Using multiple models" section of the new documentation](https://github.com/ngxson/llama.cpp/blob/xsn/server_model_management_v1_2/tools/server/README.md#using-multiple-models)

Note: waiting for further webui changes from @allozaur 

https://github.com/user-attachments/assets/bf99fff3-16a3-4d47-be21-6cf2d0c001bd

## Implementation

The feature was implemented using multi-process approach. The reason for this choice is to be more resilient in case the model crashes.

Most of the implementation is confined inside `tools/server/server-models.cpp`

There will be one main "router" server whose the job is to create other "child" processes that will actually run the inference.

This system was design and test against these unexpected cases:
- Child process suddenly exit due to error (for example, a `GGML_ASSERT`)
- Child process failed to load (for example, the system cannot launch the process)
- Router process suddenly exit due to error. In this case, child processes automatically stop themself.

These steps happen when user request the router to launch a model instance:
1. Check if the model already had a process; if yes, skip
2. Construct argv and envp to launch the child process; a random HTTP port is selected for each child process
3. Start the child process
[...]

PR #19926, opened 2026-02-26
server : support multiple model aliases via comma-separated --alias

### LINKED ISSUE #17860: server : multiple model aliases support
Currently, we support setting a single alias for each model. I think it would be very useful to extend this functionality and support multiple "secondary" aliases/ids for each model.

The main use case is to be able to generically configure 3rd party clients (for example `llama.vim`) without the client having to know about what models are specifically being provided by `llama-server`. Instead of specifying a specific model id in the client (i.e. the vim plugin in this case), I would like to be able to specify a model with a generic alias/id: `"fim"`. This way, we shift the responsibility from the client to the server to provide a FIM model.

In this example, we can indeed simply set the alias of the completion model in `llama-server` that we want to use as `"fim"` and it would work:

```bash
# works, but not great
llama-server -hf ggml-org/qwen3-coder --alias "fim" ...
```

But we would loose information about the identity of the model - i.e. it would be still useful to know that it is a `"Qwen3 Coder"` f.ex. With multiple aliases, we can identify the model both with `"Qwen3 Coder"` + `"fim"`. Moreover, some models such as `Qwen/Qwen3-Coder-30B-A3B-Instruct` can be used both for FIM and chat, so a single alias is not enough.

Additional feature request along these lines is for the clients to be able to provide multiple model ids for each request. On the server-side, the server will look at the ids in decreasing priority and route to the first model that matches the id from the request.

For example, a client would send a request:

```json
{
    "model": ["Qwen3 Coder", "fim"],
    ...
}
```

This request would indicate that if the server has a model with the alias `"Qwen3 Coder"` - use that. If not, then use whatever FIM model it has (i.e. one that has the `"fim"` alias/tag).

Open questions:

- How to call these extra identifiers: "aliases", "tags", "ids" ?
- How to specify them? Probably CLI arg that accepts array of strings.
- How to handle duplicate aliases/ids? "match any", "match first", "match recent", "match loaded", etc.
- Case sensitivity of the ids?
- Regex in the ids?

ServeurpersoCom:

### Description:
This PR adds support for multiple model aliases and informational tags, as discussed in #17860.

The `--alias` argument accepts comma-separated values parsed into a `std::set` (no duplicates). A `--tags` argument provides informational metadata not used for routing. In single-model mode, the first alias becomes the model name for backward compatibility. In router mode with INI presets, aliases and tags are parsed from the `alias` and `tags` options.

All aliases resolve equally, there is no fallback hierarchy. `has_model()` and `get_meta()` scan the flat `mapping` for both canonical names and aliases. Resolution to the canonical name happens inside `router_validate_model()` via `meta->name`.

Duplicate aliases across models are detected at startup and throw an error. Tags have no uniqueness constraint, multiple models can share the same tags.

### Changes:
* `common.h`: `model_alias` is a `std::set<std::string>`, `model_tags` added
* `arg.cpp`: `--alias` parses CSV into set, `--tags` argument added
* `server-models.h`: `aliases` and `tags` fields (`std::set`) in `server_model_meta`
* `server-models.cpp`: alias/tag parsing in `add_model()`, alias-aware `has_model()`/`get_meta()`, alias resolution in `router_validate_model()`, both exposed in `/v1/models`, startup log
* `server-context.h/cpp`: alias/tag support for single-model mode, backward compat for model name
* `server.cpp`: fallback uses `std::set::insert`

### Testing :

```

presets.ini :

[MoE-Qwen3-235B-A22B-Thinking-2507]
m = unsloth/Qwen3-235B-A22B-Thinking-2507-GGUF/Qwen3-235B-A22B-Thinking-2507-UD-Q2_K_XL-00001-of-00002.gguf
c = 98304
alias = qwen3,think


curl -s http://localhost:8080/v1/chat/completions -d '{
  "model": "MoE-Qwen3-235B-A22B-Thinking-2507",
  "messages": [{"role":"user","content":"ping"}]
}'

curl -s http://localhost:8080/v1/chat/completions -d '{
  "model": "qwen3",
  "messages": [{"role":"user","content":"ping"}]
}'


llama-server log :
...
srv   load_models:   * MoE-Qwen3-235B-A22B-Instruct-2507
srv   load_models:   * MoE-Qwen3-235B-A22B-Thinking-2507 (aliases: qwen3, think)
srv   load_models:   * MoE-Qwen3-30B-A3B-Instruct-2507
...


(root|/var/www/ia/models) curl -s http://localhost:8080/v1/models | jq '.data[] | {id, aliases}'
...
{
  "id": "MoE-Qwen3-235B-A22B-Thinking-2507",
  "aliases": [
    "qwen3",
    "think"
  ]
}
...

```
Closes #17860

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -814,7 +814,10 @@ private:
         }
         SRV_WRN("%s", "for more info see https://github.com/ggml-org/llama.cpp/pull/16391\n");
 
-        if (!params_base.model.name.empty()) {
+        if (!params_base.model_alias.empty()) {
+            // backward compat: use first alias as model name
+            model_name = *params_base.model_alias.begin();
+        } else if (!params_base.model.name.empty()) {
             model_name = params_base.model.name;
         } else {
             // fallback: derive model name from file name
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -580,6 +580,7 @@ private:
     float slot_prompt_similarity = 0.0f;
 
     std::string model_name; // name of the loaded model, to be used by API
+    std::vector<std::string> model_aliases; // additional names for the model
 
     bool sleeping = false;
 
@@ -813,8 +814,15 @@ private:
         SRV_WRN("%s", "for more info see https://github.com/ggml-org/llama.cpp/pull/16391\n");
 
         if (!params_base.model_alias.empty()) {
-            // user explicitly specified model name
-            model_name = params_base.model_alias;
+            // user explicitly specified model name (may include comma-separated aliases)
+            auto aliases = string_split<std::string>(params_base.model_alias, ',');
+            model_name = string_strip(aliases[0]);
+            for (size_t i = 1; i < aliases.size(); i++) {
+                auto alias = string_strip(aliases[i]);
+                if (!alias.empty()) {
+                    model_aliases.push_back(alias);
+                }
+            }
         } else if (!params_base.model.name.empty()) {
             // use model name in registry format (for models in cache)
             model_name = params_base.model.name;
@@ -2892,6 +2900,7 @@ server_context_meta server_context::get_meta() const {
     return server_context_meta {
         /* build_info             */ build_info,
         /* model_name             */ impl->model_name,
+        /* model_aliases          */ impl->model_aliases,
         /* model_path             */ impl->params_base.model.path,
         /* has_mtmd               */ impl->mctx != nullptr,
         /* has_inp_image          */ impl->chat_params.allow_image,
@@ -3688,6 +3697,7 @@ void server_routes::init_routes() {
             {"data", {
                 {
                     {"id",       meta->model_name},
+                    {"aliases",  meta->model_aliases},
                     {"object",   "model"},
                     {"created",  std::time(0)},
                     {"owned_by", "llamacpp"},
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -580,7 +580,8 @@ private:
     float slot_prompt_similarity = 0.0f;
 
     std::string model_name; // name of the loaded model, to be used by API
-    std::vector<std::string> model_aliases; // additional names for the model
+    std::set<std::string> model_aliases; // additional names for the model
+    std::set<std::string> model_tags;    // informational tags
 
     bool sleeping = false;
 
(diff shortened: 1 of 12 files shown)


tools/server/server-models.h
```diff
@@ -52,6 +52,7 @@ static std::string server_model_status_to_string(server_model_status status) {
 struct server_model_meta {
     common_preset preset;
     std::string name;
+    std::vector<std::string> aliases; // additional names that resolve to this model
```
ngxson: I think we should use `std::map` to be explicit that aliases cannot be duplicated

tools/server/server-context.cpp
```diff
@@ -813,8 +814,15 @@ struct server_context_impl {
...
-            model_name = params_base.model_alias;
+            // user explicitly specified model name (may include comma-separated aliases)
+            auto aliases = string_split<std::string>(params_base.model_alias, ',');
+            model_name = string_strip(aliases[0]);
+            for (size_t i = 1; i < aliases.size(); i++) {
+                auto alias = string_strip(aliases[i]);
+                if (!alias.empty()) {
+                    model_aliases.push_back(alias);
+                }
+            }
```
ngxson: I think this should better be handled in `arg.cpp`, same logic as other CSV params

`params_base.model_alias` can be changed to an `std::set`

tools/server/server-models.h
```diff
@@ -84,6 +85,7 @@ struct server_models {
     std::mutex mutex;
     std::condition_variable cv;
     std::map<std::string, instance_t> mapping;
+    std::map<std::string, std::string> name_index; // alias/name -> canonical name
```
ngxson: I would prefer keeping just one flat map `std::map<std::string, instance_t>` for simplicity.

Also, we should throw an error or a warning if user accidentally define 2 models with the same alias

tools/server/server-models.h
```diff
@@ -112,6 +114,9 @@ struct server_models {
     // check if a model instance exists (thread-safe)
     bool has_model(const std::string & name);
 
+    // resolve alias/name to canonical model name, returns empty string if not found (thread-safe)
+    std::string resolve_name(const std::string & name);
```
ngxson: no longer need this if everything is merged into one single `mapping`

tools/server/server-context.cpp
```diff
@@ -812,18 +814,17 @@ struct server_context_impl {
         }
         SRV_WRN("%s", "for more info see https://github.com/ggml-org/llama.cpp/pull/16391\n");
 
-        if (!params_base.model_alias.empty()) {
-            // user explicitly specified model name
-            model_name = params_base.model_alias;
```
ngxson, line 817: for backward compat, we will still need to set the model alias as model name (in case only one alias is set)

for simplicity, maybe just set it to the first one in the set for now, we don't actually care about order

ngxson (approved):

ngxson:
optionally, a test case would be useful too

ServeurpersoCom:
> optionally, a test case would be useful too

I added a single-model test for aliases and tags in test_basic.py: that's the mode I don't run in prod (I use router mode everywhere), so it made sense to cover it first. Router-mode test would need a temp INI preset, can add later if needed.

ngxson:
Yes that sounds good, thanks

ggerganov:
This is great, but it's still not completely what I was looking for. For example, I have this model defined:

```
0.00.019.212 I srv   load_models:   * qwen3-coder-30b-hf [tags: fim, infill]
```

I would like this request to match it:

```bash
curl \
   -XPOST "localhost:8013/infill" \
   -d '{"model": "infill", "input_prefix": "My favourite ", "input_suffix": " is called:", "n_predict": 4}'

# {"error":{"code":400,"message":"model 'infill' not found","type":"invalid_request_error"}}
```

ggerganov:
Nevermind, actually it is OK as it is. I'll work with the aliases instead. Thanks!

strawberrymelonpanda:
>> A --tags argument provides informational metadata not used for routing.

> Nevermind, actually it is OK as it is. I'll work with the aliases instead. Thanks!

Personally, I've still wanted what you were originally describing since the router feature was first introduced.  I'm not sure aliases, with no duplicates, can reproduce that functionality?  It comes close if you manually designate one. 

It's an improvement this way at least. 
 _Edit: to be clear, an ___appreciated___ improvement!_

https://github.com/ggml-org/llama.cpp/pull/17470#issuecomment-3652386180 
> If you start llama-server with an alias, i.e. --alias "local", it seems to be ignored.

Should fix this issue, more or less.

PR #17924, opened 2025-12-11
Gigachat 3 tool parser and tests

Mishusha:
This PR adds a tool parser for parsing the function calls made by the [Gigachat 3 models](https://huggingface.co/collections/ai-sage/gigachat3).
Models can generate only one function call using format:

`{text}<|message_sep|>\n\nfunction call<|role_sep|>\n{json}`
Where `<|message_sep|>\n\n` and `<|role_sep|>\n` are special tokens.

PR #19931, opened 2026-02-26
tool parser: add GigaChatV3/3.1 models support in PEG format

Mishusha:
I have recreated the PR of https://github.com/ggml-org/llama.cpp/pull/17924 for cleaner commits and no merge conflicts

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1350,6 +1350,77 @@ static common_chat_params common_chat_params_init_lfm2(const common_chat_templat
     return data;
 }
 
+static common_chat_params common_chat_params_init_gigachat_v3(
+        const common_chat_template & tmpl,
+        const autoparser::templates_params & inputs) {
+
+    common_chat_params data;
+
+    data.prompt            = common_chat_template_direct_apply(tmpl, inputs);
+    data.format            = COMMON_CHAT_FORMAT_PEG_NATIVE;
+    data.supports_thinking = false;
+    data.preserved_tokens  = {
+        "<|message_sep|>\n\n",
+        "<|role_sep|>\n",
+    };
+
+    auto has_tools         = inputs.tools.is_array() && !inputs.tools.empty();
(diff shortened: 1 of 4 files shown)


Mishusha:
@pwilkin @ggerganov 

Can you check the PR please? Thanks)

pwilkin:
@Mishusha sorry, wanted to get the autoparser in first. Could you rebase on latest master? Nothing should change except the positioning of the functions and the detection code in `chat.cpp`.

Mishusha:
@pwilkin Hello! I tried setting up an autoparser with the GigaChatv3 model.

The tests themselves pass, but the autoparser doesn't save the `<|message_sep|>\n\n` and `<|role_sep|>\n` tokens I need in preserved tokens separately, but rather their combined stripped prefix, `<|message_sep|>\n\nfunction call<|role_sep|>` from chat-template

Therefore, when I run my server without the -sp flag, toolcalls aren't automatically parsed because special tokens are discarded.

What's the best way to do this—set up my own custom parser for GigaChatv3? (Leave the one currently committed) Or do you recommend something else?

P.S. For gigachatv3.1, there's a single <|function_call|> token, and there are no such problems with the autoparser.

pwilkin:
@Mishusha Yeah, I checked it with the autoparser and determined that it needs the separate parser due to the role thing (similar to Functionary), that's why I told you to do that instead of just closing it as deprecated :) sorry, should've made myself clearer.

The parser can stay as it is since it's already PEG based, you just have to move the detection code to the right place in chat.cpp and that's it.

Mishusha:
@pwilkin Ok, thank you very much)
All the tests are passing now and everything looks correct on inference.
Please take a look)

pwilkin (approved):
Waiting for CI and then will merge.

Mishusha:
@pwilkin Some checks failed, but it seems that this is not related to this PR.

pwilkin:
Yup.

PR #16310, opened 2025-09-28
Enable Intel AMX acceleration while in CPU/GPU hybrid with new "--amx" toggle. 

Gadflyii:
This change adds a new toggle, "--no-host" that will allow the extra buft to remain functional when a GPU is present,  enabling AMX operations when in a CPU/GPU hybrid.  If the "--amx" toggle is not present, current behavior is maintained. 

- The toggle is functional in llama-bench, llama-cli, and llama-server. 
- Compatible with all --cpu-moe, --n-cpu-moe N, --cpu-moe-draft, and --n-cpu-moe-draft N  as implemented on Sep 27th, 2025.
- Compatible with all Sapphire Rapids, Emerald Rapids, and Granite rapids CPU's.
- If "--no-host" is accidentally enabled on non-Intel CPU's, or Intel CPU's without AMX, there is no change in behavior (Tested with AMD 9950X3D & Intel 14900k).
- Works in WSL or native Linux (Tested Ubuntu 24.04 LTS and Windows 11 + Ubuntu WSL). 

This change allows significant performance increases on the CPU offloaded layers / moe while in hybrid operations; especially in prompt eval, where with 100%-150%+ performance uplifts are common:

**Examples:**

**Base command:**

```
numactl -N 2,3 -m 2,3 ~/src/llama.cpp/build/bin/llama-cli -m /mnt/ssd2/AI/Qwen3_30B/Q4_0/Qwen3-30B-A3B-Thinking-2507-Q4_0.gguf  -t 64 -b 1024 -c 1024 -n 1024 --numa numactl -p "The quick brown fox jumps over the lazy dog many times. A curious cat watches carefully from the garden wall nearby. Birds sing softly in the morning air, while the sun rises gently above the hills. Children walk slowly to school carrying bright backpacks filled with books, pencils, and small notes. The teacher greets them warmly at the classroom door. Lessons begin with stories about science, history, art, and music. Ideas flow clearly and simply, creating a calm rhythm of learning. Friends share smiles, trade sandwiches, and laugh during the short break. The day continues peacefully until the afternoon bell finally rings." -no-cnv --n-gpu-layers 10
```

**No AMX (Current behavior):**

```
llama_perf_sampler_print:    sampling time =      91.57 ms /   927 runs   (    0.10 ms per token, 10123.18 tokens per second)
llama_perf_context_print:        load time =    1202.46 ms
[...]

PR #19932, opened 2026-02-26
llama: disable repack by default

am17an:
This PR disable repack by default. The reason to do this is because repack slows stuff down when doing hybrid inference. Users are expected to supply `--repack` argument when doing pure-CPU inference for speedups.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1482,7 +1482,7 @@ jobs:
       - name: Test
         id: ggml-ci
         run: |
-          LLAMA_ARG_THREADS=$(nproc) GG_BUILD_LOW_PERF=1 bash ./ci/run.sh ./tmp/results ./tmp/mnt
+          LLAMA_ARG_THREADS=$(nproc) LLAMA_ARG_REPACK=1 GG_BUILD_LOW_PERF=1 bash ./ci/run.sh ./tmp/results ./tmp/mnt
 
   ggml-ci-arm64-cpu-low-perf:
     runs-on: ubuntu-22.04-arm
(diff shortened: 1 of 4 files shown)


ggerganov:
Will need to fix this problem https://github.com/ggml-org/llama.cpp/actions/runs/22444527480/job/64995293723?pr=19932#step:5:1780 before merging. And will likely need to add `--repack` to the `ggml-ci` for CPU builds. But after that issue is resolved.

am17an:
It looks like it's overflowing fp16 range, since accumulation is being done in f16

ggerganov:
> It looks like it's overflowing fp16 range, since accumulation is being done in f16

Ah yes, I keep forgetting about this. OK, so we just need to set `LLAMA_ARG_REPACK=1` for all ggml-ci workflows that run on the CPU and should be OK.

.github/workflows/build.yml
```diff
@@ -2088,49 +2088,50 @@ jobs:
...
-           sudo apt-get update
-           sudo DEBIAN_FRONTEND=noninteractive NEEDRESTART_MODE=a \
-           apt-get install -y \
-            build-essential \
-            python3-venv \
-            gpg \
-            wget \
-            time \
-            git-lfs
+      runs-on: ah-ubuntu_22_04-c8g_8x
```
am17an: this indentation was not consistent with the rest of the file, that's why the diff

DocShotgun:
Doesn't repack already get disabled by default via the presence of host buffers during CPU+GPU inference (https://github.com/ggml-org/llama.cpp/pull/16310), or am I missing something?

am17an:
Only on CUDA see https://github.com/ggml-org/llama.cpp/pull/19738#issuecomment-3964843282

.github/workflows/build.yml
CISC: It's even less consistent now, both `kleidiai` jobs have an initial indentation of 4 instead of 2 like the rest of the file.

am17an:
Not sure why the kleidai builds are failing. Any ideas?

ggerganov:
Yes, they are not picking the repack argument from the env variable.

In any case, I read the discussion in #16310 as pointed by @DocShotgun and realized that my understanding about how the repack mechanism is activated was a bit incorrect. Basically, if the backend provides host buffers, then the repack will not be used which eliminates most of my concern that the repack will be used with hybrid CPU/GPU systems. The reason is that actually most backend (CUDA/Vulkan/CANN etc) have host buffers. The only exception is the Metal backend, but it's not a big issue since hybrid inference is rarely needed on Macs. I'll think how to improve this anyway.

So with this in mind, I now think it's ok to merge https://github.com/ggml-org/llama.cpp/pull/19738 and the changes in this PR are not needed.

PR #19938, opened 2026-02-26
scripts: ini_to_opencode.py

am17an:
Convert `ini` files to OpenCode config. 

Steps (assuming you have installed [OpenCode](https://github.com/anomalyco/opencode) already).
1. Create an `models.ini` file with the models of your choice 
e.g.
```
[qwen35-model]
model = models/Qwen3.5-35B-A3B-Q4_K_M.gguf
chat-template = qwen
jinja = true
n-cpu-moe = 10

[gpt-oss]
model = models/gpt-oss-20b-mxfp4.gguf
jinja = true 
```

2. Run `python scripts/ini_to_opencode.py models.ini > ~/.config/opencode/opencode.json` . Please note this will overwrite your existing `opencode.json`

3. Now `opencode` should have `llama.cpp (local)` as a provider, and you should see something like the image below

Screenshot 2026-02-27 015616

---- 

AI Disclosure: Recursively, this PR was written using OpenCode using Qwen3.5 on my machine.

### DIFF
--- a/scripts/ini_to_opencode.py
+++ b/scripts/ini_to_opencode.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+"""
+Convert llama.cpp .ini model presets to opencode.json provider config
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import re
+import sys
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def parse_ini(content: str) -> dict:
+    """Parse simple INI format with sections and key-value pairs"""
+    result = {}
+    current_section = None
+
+    for line in content.split("\n"):
+        line = line.strip()
+
+        # Skip empty lines and comments
+        if not line or line.startswith(";") or line.startswith("#"):
+            continue
+
+        # Section header
+        section_match = re.match(r"^\[([^\]]+)\]$", line)
+        if section_match:
+            current_section = section_match.group(1)
+            result[current_section] = {}
+            continue
+
+        # Key-value pair
+        if "=" in line and current_section:
+            key, _, value = line.partition("=")
+            key = key.strip()
+            value = value.strip()
+            result[current_section][key] = value
+
+    return result
+
+
+def ini_to_opencode(ini_file: str, output_file: str | None = None):
+    """Convert INI preset file to opencode.json"""
+
+    # Read and parse INI
+    ini_path = Path(ini_file)
+    ini_content = ini_path.read_text()
+    ini_data = parse_ini(ini_content)
+
+    # Build opencode.json structure
+    opencode_config = {
+        "$schema": "https://opencode.ai/config.json",
+        "provider": {
+            "llamacpp": {
+                "npm": "@ai-sdk/openai-compatible",
+                "name": "llama.cpp (local)",
+                "options": {"baseURL": "http://127.0.0.1:8080/v1", "apiKey": "no-key-needed"},
+                "models": {},
+            }
+        },
+    }
+
+    # Map INI presets to opencode models
+    # Skip global section [*] and version
+    for section_name, settings in ini_data.items():
+        if section_name == "[*]" or section_name == "version":
+            continue
+
+        model_id = section_name.lower().replace(" ", "-").replace(":", "-")
+        model_config = {"name": section_name}
+
+        # Map common INI options to opencode/llama.cpp server settings
+        if "model" in settings:
+            model_config["model"] = settings["model"]
+
+        if "n-gpu-layer" in settings or "n-gpu-layers" in settings:
+            gpu_layer = settings.get("n-gpu-layer") or settings.get("n-gpu-layers")
+            model_config["n_gpu_layers"] = gpu_layer
+
+        if "c" in settings:
+            model_config["context_size"] = settings["c"]
+
+        if "chat-template" in settings:
+            model_config["chat_template"] = settings["chat-template"]
+
+        if "jinja" in settings:
+            model_config["jinja"] = settings["jinja"].lower() == "true"
+
+        opencode_config["provider"]["llamacpp"]["models"][model_id] = model_config
+
+    # Write output
+    output_path = Path(output_file) if output_file else ini_path.with_suffix(".json")
+    output_path.write_text(json.dumps(opencode_config, indent=2) + "\n")
+
+    logger.info(json.dumps(opencode_config, indent=2))
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO, format="%(message)s")
+
+    if len(sys.argv) < 2:
+        logger.error("Usage: python3 ini_to_opencode.py <input.ini> [output.json]")
+        sys.exit(1)
+
+    ini_to_opencode(sys.argv[1], sys.argv[2] if len(sys.argv) > 2 else None)


ngxson:
I think we can add this directly to llama-cli or llama-server, will see what's the best UX, but having this shipping out-of-the-box would be nice

ITankForCAD:
I think this is a good idea. Furthermore, it aligns with the principles of [this discussion](https://github.com/ggml-org/llama.cpp/discussions/15313). Maybe scripts of this nature would fit better in the tools directory ?

jacekpoplawski:
I didn't run the script, but in my json I have:

```
    "llama.cpp linux": {
      "npm": "@ai-sdk/openai-compatible",
      "name": "llama.cpp",
      "options": {
        "baseURL": "http://192.168.0.165:8080/v1",
        "apiKey": "sk-local"
      },
      "models": {
        "local-model": {
          "name": "llama.cpp linux",
          "limit": {
            "context": 100000,
            "output": 20000
          }
        }
      }
    },
```

in the script source I see only "context_size", is this same setting?
also could we have IP somewhere?

fkroener:
This contribution doesn't appear to adhere to the guidelines re AI usage at all.

Was this actually tested in any serious manner?

Why the lower-casing? Why replace : with -? Why the jinja key in opencode.json? gpu layers? Ignoring the global section doesn't seem to work either, nor does ignoring ";". Numeric values are transformed to strings (e.g. context_size: "context_size": "32768").

I really like the idea though, and @ngxson is probably right that it might be better implemented in llama-{cli,server}, especially to easily extract context sizes if not specified by the model preset.

This pull request should be closed though.

am17an:
@ngxson integrating into llama-cli may not work. OpenCode is the harness which allows any model to work, local or cloud. We shouldn't restrict it to run via llama-cli, rather just figure out how to be a provider. How were you thinking the UX should be like?

ggerganov:
I've been using and thinking about the interaction of applications with model providers for some time, so here's my 2 cents. IMO the common wisdom in most of the existing local-AI-enabled software is suboptimal and misguided by trying to apply the same "cloud-provider" UX to the "local-provider" UX.

As a specific example, I don't see a good reason for explicitly listing specific model names in the opencode config. From opencode PoV, I would like to connect to some models and these models should support some context size and max output length. So that's what the opencode config should convey.

In that sense, the opencode config makes sense to be something like this:

```json
{
    "$schema": "https://opencode.ai/config.json",
    "provider": {
        "my-local-provider": {
            "npm": "@ai-sdk/openai-compatible",
            "options": {
                "baseURL": "http://localhost:2276/v1"
            },
            "models": {
                "opencode-local-32k": {
                    "limit": {
                        "context": 32768,
                        "output": 16384
                    }
                },
                "opencode-local-64k": {
                    "limit": {
                        "context": 65536,
                        "output": 32768
                    }
                },
                "opencode-local-128k": {
                    "limit": {
                        "context": 131072,
                        "output": 65536
                    }
                }
            }
        }
    }
}
```

Once you define this local provider with a set of abstract models, you are all set. You never have to change opencode config anymore. **You delegate the specific model selection to the model provider.**

So for example, using `llama-server` and the new functionality of multiple aliases (#17860), one can define these models to be served by the model provider (in this case `llama.cpp`):

```ini
[glm-4.7-flash-hf]
hf          = ggml-org/GLM-4.7-Flash-GGUF:Q8_0
alias       = webui-chat,opencode-local-128k,vscode-agent
ctx-size    = 131072
top-p       = 1.0
top-k       = 0
temp        = 0.2

[qwen3-coder-next-hf]
hf          = ggml-org/Qwen3-Coder-Next-GGUF
ctx-size    = 131072
top-p       = 1.0
top-k       = 40
temp        = 1.0
```

This way we have configured the model provider to serve GLM 4.7 Flash to opencode by specifying the `opencode-local-128k` alias.

Tomorrow, if we decide to use another model with opencode, we change the model provider to serve it instead by moving the opencode alias to the model we want to use:

```ini
[glm-4.7-flash-hf]
hf          = ggml-org/GLM-4.7-Flash-GGUF:Q8_0
alias       = webui-chat,vscode-agent            # <-- opencode alias removed
ctx-size    = 131072
top-p       = 1.0
top-k       = 0
temp        = 0.2

[qwen3-coder-next-hf]
hf          = ggml-org/Qwen3-Coder-Next-GGUF
alias       = opencode-local-128k                # <-- opencode alias added here
ctx-size    = 131072
top-p       = 1.0
top-k       = 40
temp        = 1.0
```

So without any changes to opencode, it is now using a different model.

This method allows to configure a given model to be used by different third-party apps by configuring the apps to "signal" what they are looking for in terms of functionality - not in terms of models. We concentrate this configuration logic in the model provider. We don't complicate the 3rd-party apps with "knowing" explicitly what models are there.

For cloud-based inference, we _have_ to tell opencode to use "Claude" or "GPT 5" or "Gemini", etc. This is because we do not control the model provider. For local-based inference, this is not the case. The user controls the provider.

am17an:
Thanks for your thoughts. The only doubt/confusion I foresee is some confusion regarding selecting models in OpenCode, from your proposal they would look like "opencode-local-32k", "opencode-local-64k" etc, which is fine for someone who has spent time setting up the .ini file. 

My 2c is because we have `--fit` automatically enabled, we can even abstract this away. My "just works" setup would be 
- point llama-server at my models dir
- some scripts which sync opencode configs vs the models (possibly combine 1 and 2 into some runner script)
- opencode has these models when I run it 
- opencode agent also has these models (in IDEs like vscode, zed etc)

ngxson:
> @ngxson integrating into llama-cli may not work. OpenCode is the harness which allows any model to work, local or cloud. We shouldn't restrict it to run via llama-cli, rather just figure out how to be a provider. How were you thinking the UX should be like?

I'm not talking about it functionally, my point is simply: these logic should better be implemented in `llama-cli` because it's easier to distribute.

The idea you introduced in this PR is the same as `--completion-bash`. In a broader sense, what it does can be categorized as "automatically write the config for other applications"

And if we want to go a step further, why don't we just submit a patch to opencode to directly use `llama-server`?

am17an:
Ah okay, I see what you mean. That would also be good 

> And if we want to go a step further, why don't we just submit a patch to opencode to directly use llama-server?

That was my initial idea, we can easily register has an openCode provider, but we also have to register all the models available, which is not possible for `llama-server` (need to register here: https://github.com/anomalyco/models.dev)

ngxson:
That's even more simple: llama-server can simply return the same data format as models.dev

The API base domain can be set here: https://github.com/anomalyco/opencode/blob/157920b2fbfbfb46dae8badb8328a53f6904e510/packages/opencode/src/provider/models.ts#L85

And it's controlled via env var `OPENCODE_MODELS_URL`: https://github.com/anomalyco/opencode/blob/157920b2fbfbfb46dae8badb8328a53f6904e510/packages/opencode/src/flag/flag.ts#L56

That make this not much different from claude code compat

fkroener:
Hey, I really wonder why my comment above got hidden and by whom, since I appear to be the only person to actually look at and test the code put up for review. This doesn't seem to be the right way to take feedback and criticism. 

> This contribution doesn't appear to adhere to the guidelines re AI usage at all.
> 
> Was this actually tested in any serious manner?
> 
> Why the lower-casing? Why replace : with -? Why the jinja key in opencode.json? gpu layers? Ignoring the global section doesn't seem to work either, nor does ignoring ";". Numeric values are transformed to strings (e.g. context_size: "context_size": "32768").
> 
> I really like the idea though, and @ngxson is probably right that it might be better implemented in llama-{cli,server}, especially to easily extract context sizes if not specified by the model preset.
> 
> This pull request should be closed though.

am17an:
@ngxson that's a good idea. But say I run out of context on a particular model and want to shift to the cloud, then this doesn't work

am17an:
@fkroener I did. Please stop, you're not helping this discussion.

strawberrymelonpanda:
I like the idea, my $0.02 as someone who uses OpenCode + Roocode and Llama.cpp together frequently:
My main pain point, especially after the multiple aliases PR that was just merged, is keeping the ctx size of the loaded model aligned to the client.

Up until now I've been using a tiny script to get the currently loaded model and tell OpenCode to use it.  But for every model, I still have to set up the context sizes from --fit.  It's hardly a big trial, but I think it's the last step in automating this away.

PR #19952, opened 2026-02-27
scripts : improve get-wikitext-2.sh

angt:
Switch to sh, add curl fallback, and avoid redundant downloads

### DIFF
--- a/scripts/get-wikitext-2.sh
+++ b/scripts/get-wikitext-2.sh
@@ -1,37 +1,38 @@
 #!/bin/sh
+# vim: set ts=4 sw=4 et:
 
 ZIP="wikitext-2-raw-v1.zip"
 FILE="wikitext-2-raw/wiki.test.raw"
 URL="https://huggingface.co/datasets/ggml-org/ci/resolve/main/$ZIP"
 
 die() {
-	printf "%s\n" "$@" >&2
-	exit 1
+    printf "%s\n" "$@" >&2
+    exit 1
 }
 
 have_cmd() {
-	for cmd; do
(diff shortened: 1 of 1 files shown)


scripts/get-wikitext-2.sh
```diff
@@ -1,11 +1,42 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
CISC, line 1: I get that not all systems have `bash`, but I'm fairly sure `/usr/bin/env` is pretty much guaranteed, while `/bin/sh` may not be.

angt: We use `/usr/bin/env` for `bash` (but also `python`, and almost everything) as they can be installed at many places, but for `sh`, we should always use `/bin/sh` as this path is standardized.

CISC: That's the thing though, I don't think it is as standard as `/usr/bin/env`.

angt: Also in POSIX, `/bin/sh` is guaranteed while even `/usr` isn't.

CISC: It's not part of POSIX, but in practise it's ubiquitous.

angt: Yes, while `/bin/sh` is required, that's why it is almost the only one with this shebang.

CISC: Ahwell, probably not a real issue either way.

ggerganov (approved):

PR #19956, opened 2026-02-27
ggml-vulkan: limit the number of command buffers to transfer data to 4

mklimenk:
This PR introduces a limit to the amount of command buffers to force reusing them instead of creating a new command buffer for each megabyte of data to transfer.

I stumbled upon this limitation while trying to launch GPT-OSS-120B MXFP4 via Vulkan backend, which failed during the model loading stage. 

Adding the proposed fix allows for a command buffer reuse instead of creating a new one for every transfer, which result in a lot of command buffers for larger models and might hit GPU hardware limitations. A number 4 is taken from the number of staging buffers for async uploads, but can be adjusted. No performance or accuracy implications were observed.


0cc4m:
We've been discussing possible solutions in #19420. This is one of them, yes.

mklimenk:
@0cc4m I checked PRs for duplicates, but didn't look into issues. Thank you for investigating it, closing this as a duplicate.
In case you need some additional info or help, feel free to ping me

PR #19958, opened 2026-02-27
llama : add native param2moe architecture support

iambhuvan:
Adds support for the **Param2-17B-A2.4B** mixture-of-experts model architecture (`param2moe`).

Param2 is a 17B parameter MoE model with 2.4B active parameters, trained on 22 Indian + English languages by BharatGen.

  ## Why a new architecture

  Param2 cannot reuse `glm4moe` or other existing MoE archs because:
  - Uses `ffn_norm` (pre-FFN RMSNorm), not `attn_post_norm` (no post-attention norm)
  - No multimodal / mrope support
  - No NextN / MTP layers (all 28 layers are standard transformer blocks)
  - Rope type: NEOX
  - Expert config: 64 experts, 6 active + 2 shared, sigmoid gating, scale = 2.5

  ## Files changed

  | File | Change |
  |------|--------|
  | `gguf-py/gguf/constants.py` | PARAM2MOE enum + tensor list |
  | `src/llama-arch.h` | `LLM_ARCH_PARAM2MOE` enum entry |
  | `src/llama-arch.cpp` | name map + tensor name list |
  | `src/llama-model.cpp` | param load, tensor load, dispatch, rope type |
  | `src/models/models.h` | struct declaration |
  | `src/models/param2-moe.cpp` | new graph builder (~115 lines) |
  | `src/CMakeLists.txt` | added `param2-moe.cpp` to build |
  | `convert_hf_to_gguf.py` | `model_arch = PARAM2MOE` |

  ## Performance (Q4_K_M, Apple M4 Pro, -ngl 99)

  | Test | Speed |
  |------|-------|
  | Prompt processing (pp512) | ~850 t/s |
  | Token generation (tg128) | ~100 t/s |

  ## GGUF

  Q4_K_M available at: https://huggingface.co/Bhuvan3389/Param2-17B-A2.4B-Thinking-GGUF


convert_hf_to_gguf.py
```diff
@@ -1280,6 +1280,9 @@ def get_vocab_base_pre(self, tokenizer) -> str:
         if chkhsh == "9b1be57e70d20d9501b2b3186e792d81181ae36ada3903c26f9fea418cf87206":
             # ref: https://huggingface.co/inclusionAI/Ling-mini-base-2.0
             res = "bailingmoe2"
+        if chkhsh == "f9c762bae4ba0efbb6e71279acfb1546f9711729551c3a81c5c0645116f988cd":
+            # ref: https://huggingface.co/bharatgenai/Param2-17B-A2.4B-Thinking
+            res = "bailingmoe2"
```
CISC, line 1285: Nope, this is not a `bailingmoe2` vocab, the pre-tokenizer regex is unique and requires its own.

(Also, this is not how you add a new `chkhsh`, see):
https://github.com/ggml-org/llama.cpp/blob/a5625daff3eb6ffcd0ff992008f427fff80af4fc/convert_hf_to_gguf.py#L1044-L1046

convert_hf_to_gguf.py
```diff
@@ -9039,6 +9042,100 @@ def prepare_tensors(self):
...
+            num_kv_heads = self.hparams["num_key_value_heads"]
+            head_dim = self.hparams.get("head_dim", 64)
+
+            total_heads = num_heads + 2 * num_kv_heads
+            hidden = data_torch.shape[-1]
+            qkv = data_torch.reshape(total_heads, head_dim, hidden)
+
+            q = qkv[:num_heads].reshape(-1, hidden)
+            k = qkv[num_heads:num_heads + num_kv_heads].reshape(-1, hidden)
+            v = qkv[num_heads + num_kv_heads:].reshape(-1, hidden)
```
CISC, line 9093: Please do not split QKV, leave it fused, there are plenty of examples on how to do this.

CISC (commented):
Ok, even though you clearly ignored our [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) I will leave you with this:

>   Param2 cannot reuse `glm4moe` or other existing MoE archs because:
>   - Uses `ffn_norm` (pre-FFN RMSNorm), not `attn_post_norm` (no post-attention norm)

Yet your implementation is identical, only using different tensor name (there is basically no difference between these two unless both are used).

>   - No multimodal / mrope support

Doesn't matter, the related code is conditional.

>   - No NextN / MTP layers (all 28 layers are standard transformer blocks)

Again, conditional.

>   - Rope type: NEOX

Same as GLM4MOE then.

>   - Expert config: 64 experts, 6 active + 2 shared, sigmoid gating, scale = 2.5

So, in total, as implemented in this PR, there is no difference between these two models.

However, if that is truly the case I cannot say without comparing the two, which I do not have the time to do.

If you want to proceed with this PR please do the work to convince me that they are indeed identical in implementation, or as the case may be, not.

iambhuvan:
@CISC You're right — thank you for the careful review. The forward pass of Param2-MoE is functionally identical to GLM4_MOE with mrope and NextN disabled, so a separate graph builder added no value.

I've refactored the PR based on your feedback:

**What changed:**
- Removed `src/models/param2-moe.cpp` entirely — no duplicate graph builder
- `LLM_ARCH_PARAM2MOE` now falls through to `llm_build_glm4_moe` in the dispatch switch
- Fixed tensor naming: the Param2 HF model uses `post_attention_layernorm` (same as GLM4_MOE), which I had incorrectly mapped to `ffn_norm` in the first version — corrected to `LLM_TENSOR_ATTN_POST_NORM` / `blk.%d.post_attention_norm`
- `src/models/models.h` — removed the redundant struct declaration

**What remains:**
- `LLM_ARCH_PARAM2MOE` enum entry and name mapping — allows GGUFs to carry `architecture: param2moe` instead of `glm4moe`
- PARAM2MOE-specific param load and tensor load cases in `llama-model.cpp`
- `MODEL_ARCH.PARAM2MOE` in `gguf-py/gguf/constants.py` with correct tensor list
- `Param2MoEModel` converter class in `convert_hf_to_gguf.py`

The PR is now 241 insertions, 0 deletions from upstream — no new graph code, just the architecture identifier and converter.

PR #19697, opened 2026-02-17
New option GGML_CUDA_FORCE_CUBLAS_COMPUTE_32F to use fp32 as compute type in cuBLAS

wallentri88:
It probably should help for numerical overflows. It's especially useful for cards like V100 which uses cuBLAS for prompt processing by default.

And it's still much faster than MMQ kernels.

`vanilla build, no special flags`:

ggml_cuda_init: found 1 CUDA devices:
  Device 0: Tesla V100-SXM2-32GB, compute capability 7.0, VMM: yes
| model                          |       size |     params | backend    | threads | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | -: | --------------: | -------------------: |
| gemma3 27B Q8_0                |  26.73 GiB |    27.01 B | CUDA,BLAS  |       7 |  1 |         pp10000 |        931.24 ± 0.57 |
| gemma3 27B Q8_0                |  26.73 GiB |    27.01 B | CUDA,BLAS  |       7 |  1 |           tg512 |         23.05 ± 0.00 |

`-DGGML_CUDA_FORCE_CUBLAS_COMPUTE_32F=ON`:

ggml_cuda_init: found 1 CUDA devices:
  Device 0: Tesla V100-SXM2-32GB, compute capability 7.0, VMM: yes
| model                          |       size |     params | backend    | threads | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | -: | --------------: | -------------------: |
| gemma3 27B Q8_0                |  26.73 GiB |    27.01 B | CUDA,BLAS  |       7 |  1 |         pp10000 |        916.00 ± 0.99 |
| gemma3 27B Q8_0                |  26.73 GiB |    27.01 B | CUDA,BLAS  |       7 |  1 |           tg512 |         23.03 ± 0.00 |

`-DGGML_CUDA_FORCE_MMQ=ON`:

ggml_cuda_init: found 1 CUDA devices:
  Device 0: Tesla V100-SXM2-32GB, compute capability 7.0, VMM: yes
| model                          |       size |     params | backend    | threads | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | -: | --------------: | -------------------: |
| gemma3 27B Q8_0                |  26.73 GiB |    27.01 B | CUDA,BLAS  |       7 |  1 |         pp10000 |        591.17 ± 2.14 |
[...]

PR #19959, opened 2026-02-27
Use fp32 in cuBLAS V100 to avoid overflows, env variables to override cuBLAS compute type

wallentri88:
See discussion in https://github.com/ggml-org/llama.cpp/pull/19697


wallentri88:
@JohannesGaessler can you review this one, please.
I also fixed (?) existing logic in `ggml_cuda_mul_mat_batched_cublas_impl` for AMD devices because it was weird: they change compute type to fp32, but not result type. So it's being computed in fp32, but could be saved in fp16 and in the end manually converted back in fp32 by `ggml_get_to_fp32_cuda` func. So now they are being computed and saved both in fp32 and no unneeded conversions happen.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1225,6 +1225,30 @@ static cudaError_t ggml_cuda_cpy_tensor_2d(
     }
 }
 
+static cublasComputeType_t ggml_cuda_cublas_get_force_compute_type()
+{
```
```suggestion
static cublasComputeType_t ggml_cuda_cublas_get_force_compute_type() {
```

JohannesGaessler (commented):
Avoid weird enum conversions and instead set the values of one boolean variable per compute type in the function you added. Also document the new environment variables as such rather than as a subsection of `GGML_CUDA_FORCE_CUBLAS`.

wallentri88:
I would like to avoid creating several static variables to prevent extra checks when accessing them. So I can do more explicit assert: `static_assert(CUBLAS_COMPUTE_32F != cublasComputeType_t(0) && CUBLAS_COMPUTE_16F != cublasComputeType_t(0));` or define a new enum class for this function.

>document the new environment variables as such

Can you please point me where is existing documentation of llama.cpp environment variables? I haven't found any.

docs/build.md
JohannesGaessler, line 267: This is where environment variables are currently documented, longer-term it would probably make sense to refactor the documentation.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1225,6 +1225,29 @@ static cudaError_t ggml_cuda_cpy_tensor_2d(
     }
 }
 
+static cublasComputeType_t ggml_cuda_cublas_get_force_compute_type() {
```
```suggestion
static cublasComputeType_t ggml_cuda_cublas_get_force_compute_type(bool & force_32f, bool force_16f) {
```

Retrieve the environment variables like this.

am17an:
BTW in the original issue, see https://github.com/ggml-org/llama.cpp/issues/19659#issuecomment-3992065156, it's not only the accumulator to f32 that is the issue, but also quantizing the activation to fp16

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1225,6 +1225,35 @@ static cudaError_t ggml_cuda_cpy_tensor_2d(
     }
 }
 
+struct cublas_force_compute_type
+{
```
```suggestion
struct cublas_force_compute_type {
```

JohannesGaessler (approved):

docs/build.md
```diff
@@ -254,6 +254,14 @@ The environment variable [`CUDA_SCALE_LAUNCH_QUEUES`](https://docs.nvidia.com/cu
 
 Consider setting `CUDA_SCALE_LAUNCH_QUEUES=4x`, which increases the CUDA command buffer to 4 times its default size. This optimization is particularly beneficial for **Multi-GPU setups with pipeline parallelism**, where it significantly improves prompt processing throughput by allowing more operations to be enqueued across GPUs.
 
+#### GGML_CUDA_FORCE_CUBLAS_COMPUTE_32F
+
+Use `GGML_CUDA_FORCE_CUBLAS_COMPUTE_32F` environment variable to use FP32 compute type on all GPUs in FP16 cuBLAS for preventing possible numerical overflows in exchange for slightly slower prompt processing.
+
+#### GGML_CUDA_FORCE_CUBLAS_COMPUTE_16F
+
+Use `GGML_CUDA_FORCE_CUBLAS_COMPUTE_16F` environment variable to force use FP16 compute type (instead of default FP32) in FP16 cuBLAS for V100, CDNA and RDNA4.
```
ORippler, line 263: Afaik the only Tensor Cores that do not have same throughput for FP16 vs. FP32 accumulation are those used in Geforce products, so those statements are inaccurate from an NVDA perspective (COMPUTE_F32 will not slow down RTX PRO or Datacenter products).

Consequentially, setting 16F on RTX PRO/Datacenter products doesn't make any sense and should not be done (in case it's not ignored inside cuBLAS)

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1906,10 +1941,24 @@ static void ggml_cuda_mul_mat_batched_cublas_impl(ggml_backend_cuda_context & ct
...
-    if (dst->op_params[0] == GGML_PREC_DEFAULT) {
+    const auto & force_compute_type = ggml_cuda_cublas_get_force_compute_type();
+    
+    int id = ggml_cuda_get_device();
+    const int cc = ggml_cuda_info().devices[id].cc;
+    static constexpr bool is_src0_type_f16 = src0_type == GGML_TYPE_F16;
+    
+    // bf16 and fp32 are already being computed in fp32 (ensure it using static_assert),
+    // so checking necessity of forced fp32 only for fp16 src0_type
+    static_assert(is_src0_type_f16 || traits::compute_type == CUBLAS_COMPUTE_32F);
```
ORippler, line 1969: See above, if one wants to make a distinction for NVGPUs it should be DC/RTX PRO --> always F32, and GeForce products can choose between F32 and F16 precision (and default to F32 for better numerical stability imo)

docs/build.md
wallentri88: >COMPUTE_F32 will not slow down RTX PRO or Datacenter product

Benches in the previous PR (linked in the first message of this PR) show that V100 has measurable differences larger than measurement error between two modes.
On 3090 Ti difference is much more noticeable, however:

  Device 0: NVIDIA GeForce RTX 3090 Ti, compute capability 8.6, VMM: yes
| model                          |       size |     params | backend    | threads | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | -------: | -: | --------------: | -------------------: |
| llama 13B Q8_0                 |  12.12 GiB |    12.25 B | CUDA,BLAS  |       7 |     2048 |  1 |          pp1000 |      4097.25 ± 81.69 |
| llama 13B Q8_0                 |  12.12 GiB |    12.25 B | CUDA,BLAS  |       7 |     2048 |  1 |          pp1000 |      2787.15 ± 22.93 |

First result with FP16 accumulation, second is FP32. So you are right about GeForce products, but using COMPUTE_16F on V100 also has some positive impact compared to COMPUTE_32F.

Thank you for the info, I'll update the docs accordingly, but taking into account [this comment](https://github.com/ggml-org/llama.cpp/pull/20078#issuecomment-3995888546) I would like to avoid to overcomplicate things here (given that FP16 accumulation was here for years). As far as I understand, nobody really uses cuBLAS path except AMD users and V100, this PR solves the accumulation problem for V100 and env variables were introduced just for testing or experiments.

(same for the comment below)

wallentri88:
@am17an 

  Device 0: Tesla V100-SXM2-32GB, compute capability 7.0, VMM: yes
| model                          |       size |     params | backend    | threads | n_ubatch | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | -------: | -: | --------------: | -------------------: |
| llama 13B Q8_0                 |  12.12 GiB |    12.25 B | CUDA,BLAS  |       7 |     2048 |  1 |          pp1000 |       2500.78 ± 3.48 |
| llama 13B Q8_0                 |  12.12 GiB |    12.25 B | CUDA,BLAS  |       7 |     2048 |  1 |          pp1000 |        567.90 ± 0.53 |
| llama 13B Q8_0                 |  12.12 GiB |    12.25 B | CUDA,BLAS  |       7 |     2048 |  1 |          pp1000 |       1417.77 ± 0.95 |

First result is current PR, no env vars. Second one is disabling FP16 cuBLAS path. Third one is forced mmq. I don't think it would be a good idea to disable FP16 cuBLAS path for this device by default.

JohannesGaessler:
Were your previous numbers correct? I find it weird that this change would make such a large difference for LLaMA 13b but no difference at all for Gemma 27b.

wallentri88:
These are different numbers. Benchmarks linked in the first post compared FP32 compute type vs FP16 compute type in **FP16** cuBLAS branch.

Benchmarks above compare:
1. FP32 compute type in FP16 cuBLAS branch
vs
2. FP32 cuBLAS branch (FP16 branch is **disabled**)
vs
3. Forced mmq

It means that adding FP32 compute type in FP16 cuBLAS branch has negligibly small overhead for V100, but disabling FP16 branch completely is kinda devastating for PP. At most I can add another env variable for it, but it would be kinda impractical because even MMQ is ~3x faster. So if someone still catching NaNs or garbage output on V100 after this PR is merged then it's probably no other way than to advice to compile with forced mmq (of course these are my general thoughts and I'm not CUDA expert or anything, maybe there's some way to prevent NaNs and still use FP16 cuBLAS branch)

wallentri88:
Also I should point that adding FP32 compute type for V100 is not useless (see https://github.com/ggml-org/llama.cpp/issues/19659#issuecomment-3919038896), it fixes problem with gpt oss. But seemingly it's not enough for Nemotron model for some reason.

am17an:
Yeah I guess for volta there is no tf32 available, so it will be super slow. Ideally it would crash if we found NaNs in the path and point the user to compile with `FORCE_MMQ`, but not sure if it is worth the complexity for pre-Ampere

wallentri88:
Um, is there something else needs to be done to make it merged? Or what my next steps should be. Sorry, I'm not really familiar with PR flow here

am17an:
Yes you need to fix the CI. https://github.com/ggml-org/llama.cpp/actions/runs/22687620944/job/66128523972?pr=19959

docs/build.md
ORippler: > First result with FP16 accumulation, second is FP32. So you are right about GeForce products, but using COMPUTE_16F on V100 also has some positive impact compared to COMPUTE_32F.

Most likely the required F32->F16 conversion that is induced by COMPUTE_F32 (if we only have F16 output dtypes, otherwise it would be a mix/depend on the workload)

JohannesGaessler:
Also I wanted to make sure to check V100 performance with default setting myself. Seems to be fine:

<details>
<summary>Performance</summary>

| GPU            | Model            |   Microbatch size | Test   |   t/s b8193 |   t/s ad27ec115 |   Speedup |
|:---------------|:-----------------|------------------:|:-------|------------:|----------------:|----------:|
| V100-PCIE-32GB | llama 8B BF16    |                 1 | pp512  |       57.01 |           56.97 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |                 2 | pp512  |      103.02 |          102.95 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |                 4 | pp512  |      125.41 |          125.30 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |                 8 | pp512  |      117.82 |          117.76 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |                16 | pp512  |       70.60 |           70.58 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |                32 | pp512  |      140.44 |          140.42 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |                64 | pp512  |      277.61 |          277.58 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |               128 | pp512  |      411.40 |          411.65 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |               256 | pp512  |      577.32 |          577.25 |      1.00 |
| V100-PCIE-32GB | llama 8B BF16    |               512 | pp512  |      615.38 |          613.42 |      1.00 |
| V100-PCIE-32GB | llama 8B F16     |                 1 | pp512  |       57.17 |           57.18 |      1.00 |
| V100-PCIE-32GB | llama 8B F16     |                 2 | pp512  |      107.78 |          107.79 |      1.00 |
| V100-PCIE-32GB | llama 8B F16     |                 4 | pp512  |      198.55 |          198.56 |      1.00 |
| V100-PCIE-32GB | llama 8B F16     |                 8 | pp512  |      366.63 |          366.80 |      1.00 |
| V100-PCIE-32GB | llama 8B F16     |                16 | pp512  |      653.18 |          653.65 |      1.00 |
| V100-PCIE-32GB | llama 8B F16     |                32 | pp512  |     1045.47 |          997.40 |      0.95 |
| V100-PCIE-32GB | llama 8B F16     |                64 | pp512  |     1348.60 |         1417.91 |      1.05 |
| V100-PCIE-32GB | llama 8B F16     |               128 | pp512  |     2433.17 |         2244.09 |      0.92 |
| V100-PCIE-32GB | llama 8B F16     |               256 | pp512  |     3695.84 |         3542.19 |      0.96 |
| V100-PCIE-32GB | llama 8B F16     |               512 | pp512  |     4360.14 |         4366.94 |      1.00 |
| V100-PCIE-32GB | llama 8B Q4_0    |                 1 | pp512  |      142.32 |          141.68 |      1.00 |
| V100-PCIE-32GB | llama 8B Q4_0    |                 2 | pp512  |      261.09 |          259.89 |      1.00 |
| V100-PCIE-32GB | llama 8B Q4_0    |                 4 | pp512  |      360.36 |          359.63 |      1.00 |
| V100-PCIE-32GB | llama 8B Q4_0    |                 8 | pp512  |      509.64 |          508.26 |      1.00 |
| V100-PCIE-32GB | llama 8B Q4_0    |                16 | pp512  |      730.04 |          729.25 |      1.00 |
| V100-PCIE-32GB | llama 8B Q4_0    |                32 | pp512  |     1112.09 |         1109.91 |      1.00 |
| V100-PCIE-32GB | llama 8B Q4_0    |                64 | pp512  |      621.71 |          635.99 |      1.02 |
| V100-PCIE-32GB | llama 8B Q4_0    |               128 | pp512  |     1202.12 |         1161.08 |      0.97 |
| V100-PCIE-32GB | llama 8B Q4_0    |               256 | pp512  |     2107.13 |         2056.85 |      0.98 |
| V100-PCIE-32GB | llama 8B Q4_0    |               512 | pp512  |     3051.43 |         3091.78 |      1.01 |
| V100-PCIE-32GB | llama 8B all F32 |                 1 | pp512  |       30.32 |           30.33 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |                 2 | pp512  |       59.47 |           59.48 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |                 4 | pp512  |      101.30 |          101.30 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |                 8 | pp512  |      165.24 |          165.28 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |                16 | pp512  |      292.88 |          292.92 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |                32 | pp512  |      574.94 |          574.86 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |                64 | pp512  |      680.83 |          680.91 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |               128 | pp512  |      864.38 |          864.40 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |               256 | pp512  |      918.57 |          918.38 |      1.00 |
| V100-PCIE-32GB | llama 8B all F32 |               512 | pp512  |      892.52 |          892.30 |      1.00 |

</details>

wallentri88:
fixed whitespaces

PR #19965, opened 2026-02-28
ggml webgpu: fix workgroup dispatch limit for large batch sizes

abhijitramesh:
WebGPU limits workgroup counts to 65535 per dimension. MUL_MAT operations with batch sizes exceeding this limit would fail or corrupt memory.

This PR implements 2D workgroup dispatch to handle arbitrary batch sizes:

- Adds `compute_2d_workgroups()` helper to split workgroups across X/Y dimensions when exceeding the 65535 limit                                                               
- Updates mul_mat shaders to reconstruct linear workgroup ID from 2D dispatch coordinates                                                                             
- Adds bounds checking in shaders to handle over-dispatched workgroups (inevitable with rectangular 2D grids)

### DIFF
--- a/ggml/src/ggml-webgpu/ggml-webgpu.cpp
+++ b/ggml/src/ggml-webgpu/ggml-webgpu.cpp
@@ -31,6 +31,11 @@
 #define ROUNDUP_POW2(x, pow2) (((x) + ((pow2) - 1)) & ~((pow2) - 1))
 #define CEIL_DIV(M, N)        (((M) + (N) - 1) / (N))
 
+static inline void compute_2d_workgroups(uint32_t total_wg, uint32_t max_per_dim, uint32_t & wg_x, uint32_t & wg_y) {
+    wg_x = std::min(total_wg, max_per_dim);
+    wg_y = CEIL_DIV(total_wg, max_per_dim);
+}
+
 #ifdef GGML_WEBGPU_DEBUG
 #    define WEBGPU_LOG_DEBUG(msg)  std::cout << msg << std::endl
 #    define WEBGPU_DEBUG_BUF_ELEMS 512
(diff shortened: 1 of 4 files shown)


ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -1112,18 +1117,17 @@ static webgpu_command ggml_webgpu_mul_mat(webgpu_context & ctx,
...
+    uint32_t       wg_y = 1;
+    const uint32_t max_wg_per_dim = ctx->global_ctx->capabilities.limits.maxComputeWorkgroupsPerDimension;
 
     if (use_fast && is_vec) {
         auto decisions = static_cast<ggml_webgpu_mul_mat_vec_shader_decisions *>(pipeline.context.get());
 
         uint32_t batches       = dst->ne[2] * dst->ne[3];
         uint32_t output_groups = CEIL_DIV(dst->ne[0], decisions->outputs_per_wg);
         uint32_t total_wg      = output_groups * batches;
-        // TODO: split large sizes into multiple batches to avoid way over-provisioning workgroups
```
reeselevine, line 1124: This TODO still applies I think? Just now to all paths, not just matrix-vector :P

reeselevine (commented):
Glad this turned out to be a simple fix! We should soon implement a way to avoid launching too many workgroups in these cases too, since Safari does seem to limit the allowed number more aggressively than Chrome.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
abhijitramesh: Added it back on top of the in-line function that is calculating the 2d workgroups - https://github.com/ggml-org/llama.cpp/pull/19965/changes#diff-b2c7d4c552934a20d3623603462787dc951617e22eaa503953b4b853f93b4a38R34

reeselevine (approved):
The failing WebGPU CI is unrelated to this change, it's an issue introduced in https://github.com/ggml-org/llama.cpp/pull/19772, which @nikhilJain17 is working on fixing separately.

PR #19966, opened 2026-02-28
cuda: fix ggml_cuda_cpy crash on partial GPU offload

yossiovadia:
The contiguous same-type fast path in `ggml_cuda_cpy` uses `cudaMemcpyDeviceToDevice`, which crashes when one or both pointers are host pointers.

This happens with partial GPU offload — when a model doesn't fully fit in VRAM, some KV cache tensors stay in host memory. On the second prompt the engine copies cached state back, hitting a tensor pair where at least one side is a host pointer. `cudaMemcpyDeviceToDevice` rejects that with `CUDA error: invalid argument`.

The fix swaps `cudaMemcpyDeviceToDevice` for `cudaMemcpyDefault`, which resolves pointer residency automatically. It's safe for the existing all-on-GPU path too — CUDA just does a page table lookup, negligible overhead.

**Crash log (before fix):**
```
offloading 38 repeating layers to GPU
offloading output layer to CPU

>>> Hello, how are you?
(success)

>>> Talk to me like you're The Dude from The Big Lebowski
CUDA error: invalid argument
  current device: 0, in function ggml_cuda_cpy at cpy.cu:438
  cudaMemcpyAsyncReserve(src1_ddc, src0_ddc, ggml_nbytes(src0), cudaMemcpyDeviceToDevice, main_stream)
```

**After fix:** multi-turn works, no CUDA errors.

**Setup:** RTX 4090 24GB, qwen3.5:35b-a3b (39/41 layers on GPU), Ollama v0.17.4, CUDA 12.6, driver 572.83.

Same crash reported across RTX 4070, RTX 3060, dual 1080 Ti — any config that partially offloads.

Related: ollama/ollama#14444, ollama/ollama#14497, ollama/ollama#14501, ollama/ollama#14433

**Note:** I tried reproducing this with `llama-cli` directly (built from HEAD, same model, partial offload) and couldn't trigger it. The crash only happens through Ollama, so the root cause may be in how Ollama dispatches the CPY op rather than in llama.cpp itself. That said, `cudaMemcpyDefault` costs nothing and makes the op robust against mixed-residency pointers regardless of the caller.

### DIFF
--- a/ggml/src/ggml-cuda/cpy.cu
+++ b/ggml/src/ggml-cuda/cpy.cu
@@ -412,7 +412,7 @@ void ggml_cuda_cpy(ggml_backend_cuda_context & ctx, const ggml_tensor * src0, gg
         } else
 #endif // GGML_USE_MUSA && GGML_MUSA_MUDNN_COPY
         {
-            CUDA_CHECK(cudaMemcpyAsync(src1_ddc, src0_ddc, ggml_nbytes(src0), cudaMemcpyDeviceToDevice, main_stream));
+            CUDA_CHECK(cudaMemcpyAsync(src1_ddc, src0_ddc, ggml_nbytes(src0), cudaMemcpyDefault, main_stream));
         }
     } else if (src0->type == GGML_TYPE_F32 && src1->type == GGML_TYPE_F32) {
         if (can_be_transposed) {


yossiovadia:
Hi @JohannesGaessler, @ggerganov, @CISC — sorry to ping directly, just wanted to make sure this gets on your radar since it affects a fair number of users (there are four duplicate Ollama issues for it).

This is a one-line fix for a crash in `ggml_cuda_cpy` that hits any model requiring partial GPU offload (confirmed across RTX 4090, RTX 4070, RTX 3060, and dual 1080 Ti setups). The root cause is `cudaMemcpyDeviceToDevice` being used when one of the pointers may be a host pointer from CPU-offloaded KV cache entries.

Happy to adjust anything if needed — thank you for maintaining this project!

CISC:
This seems reasonable, however `cudaMemcpyDeviceToDevice` is redefined for hip/musa, `cudaMemcpyDefault` is not (and probably needs to be). `cudaMemcpyDeviceToDevice` is also used numerous other places which probably should be verified for correctness.

Deferring this to @JohannesGaessler.

ggerganov:
Please provide a repro using llama.cpp tools.

JohannesGaessler:
As is clearly laid out in the llama.cpp [AI usage policy](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md#ai-usage-policy):

In any case, this is fundamentally the wrong fix. If the tensor data is not accessible in the `CPY` ggml op that is a ggml (CUDA) backend bug, not a bug of the op.

CISC:
> In any case, this is fundamentally the wrong fix. If the tensor data is not accessible in the `CPY` ggml op that is a ggml (CUDA) backend bug, not a bug of the op.

So it's likely Ollama is doing something strange...

yossiovadia:
Apologies for the AI-generated description — I've rewritten it. Won't happen again.

I also tried reproducing with llama-cli (built from HEAD, qwen3.5:35b-a3b with partial offload on RTX 4090, ~38-39/41 layers on GPU) — interactive mode, multiple prompts, save/restore state — and couldn't trigger the crash. The native tools handle partial offload correctly.

So you're right that the CPY op shouldn't normally see mixed-residency pointers. The crash only happens through Ollama, which suggests the root cause is in how Ollama dispatches the op. I'll look into that side.

That said, would you consider cudaMemcpyDefault here as a defensive measure? It costs nothing (CUDA resolves pointer type via page table lookup) and would prevent a crash if any caller passes a host pointer. But I understand if you'd rather keep the contract strict and fix the caller instead.

JohannesGaessler:
>That said, would you consider cudaMemcpyDefault here as a defensive measure?

Absolutely not, if the current code crashes that is indicative of a bug elsewhere in the code and this change would just be masking it.

ArthurusDent:
Just in case you haven't seen it yet, somebody came up with this solution but they didn't open a pull request: https://github.com/jatjat/ollama/commit/c50bbdf114808133e65e9e4f1d6287f13144c6d8
I didn't test their solution.

yossiovadia:
Thanks @JohannesGaessler — you were right that the fix belongs in the caller, not in the op. I traced it to `deltanet.go` in Ollama's `qwen3next` model implementation.

**Root cause:** `SetInplace` in the DeltaNet chunked attention loop creates `GGML_OP_SET` with a view of a buffer-less intermediate tensor. The scheduler can't determine the correct backend from the view (`view_src->buffer` is NULL), so expansion leaks GPU assignments from adjacent layers into CPU-layer ops during partial offload.

**Fix:** Replace `SetInplace` with `Concat`, which creates fresh tensors whose backend is correctly inferred from their inputs. One file, pure Go, no C/CUDA changes: [ollama/ollama#14536](https://github.com/ollama/ollama/pull/14536)

Happy to close this PR if you'd like, or leave it to your discretion.

yossiovadia:
Closing — the root cause was in Ollama's model code, not in llama.cpp. Fixed in [ollama/ollama#14541](https://github.com/ollama/ollama/pull/14541).

PR #19968, opened 2026-02-28
Fix logic for retrieving schema items in json_schema_to_grammar.py

RayXu14:
If `schema['items']` is `{}` and `prefixItems not in schema`, as `{}` is Falsy, the original code here will raise an error like `KeyError: 'prefixItems'`.

I think if `schema['items']` is `{}`, then `items` should just be `{}`

### DIFF
--- a/examples/json_schema_to_grammar.py
+++ b/examples/json_schema_to_grammar.py
@@ -633,7 +633,7 @@ class SchemaConverter:
             return self._add_rule(rule_name, self._build_object_rule(properties, required, hybrid_name, additional_properties=None))
 
         elif schema_type in (None, 'array') and ('items' in schema or 'prefixItems' in schema):
-            items = schema.get('items') or schema['prefixItems']
+            items = schema['items'] if 'items' in schema or schema['prefixItems']
             if isinstance(items, list):
                 return self._add_rule(
                     rule_name,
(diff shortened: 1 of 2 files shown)


examples/json_schema_to_grammar.py
```diff
@@ -633,7 +633,7 @@ def add_component(comp_schema, is_required):
             return self._add_rule(rule_name, self._build_object_rule(properties, required, hybrid_name, additional_properties=None))
 
         elif schema_type in (None, 'array') and ('items' in schema or 'prefixItems' in schema):
-            items = schema.get('items') or schema['prefixItems']
+            items = schema['items'] if 'items' in schema or schema['prefixItems']
```
CISC: It already makes that check in the line right above.

CISC:
Sorry, didn't mean to close it.

examples/json_schema_to_grammar.py
CISC:
```suggestion
            items = schema.get('items', schema.get('prefixItems'))
```
Since it makes this check above one of them must be true.

CISC (approved):

CISC:
Could you also add a test to `tests/test-json-schema-to-grammar.cpp`?

PR #19976, opened 2026-02-28
vulkan: improve partial offloading performance on AMD

0cc4m:
I saw a big difference between Vulkan and ROCm performance in partial offloads. I narrowed it down to transfer speeds for weight transfer from CPU to GPU with offloaded ops. One possible explanation is that using the dedicated transfer queue on AMD may be faster than using a compute queue, so I implemented using a transfer queue for async transfers as well and synchronizing transfers using a timeline semaphore. This does improve performance.

Then I checked and found that the dedicated transfer queue on AMD is not exposed by the Linux driver by default, so it's not actually being used. The difference comes from using a second queue (the graphics queue) for transfers, so I assume the issue was the compute queue being congested with other work.

This helps on AMD RDNA4, but not on GCN and not on Nvidia. I couldn't test Intel because the Linux driver only exposes a single queue.

I also changed the Vulkan backend offload_op function to work like the CUDA backend's, to cover some more cases. And I moved some redundant context code into a function to reduce code duplication.

image
image

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -14670,6 +14670,8 @@ static void ggml_backend_vk_event_record(ggml_backend_t backend, ggml_backend_ev
     ggml_backend_vk_context * ctx = (ggml_backend_vk_context *)backend->context;
     vk_event *vkev = (vk_event *)event->context;
 
+    ggml_vk_submit_transfer_ctx(ctx);
+
     vk_context compute_ctx = ggml_vk_get_compute_ctx(ctx);
 
     // the backend interface doesn't have an explicit reset, so reset it here
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -13519,8 +13519,7 @@ static bool ggml_backend_vk_cpy_tensor_async(ggml_backend_t backend_src, ggml_ba
     VK_LOG_DEBUG("ggml_backend_vk_cpy_tensor_async()");
     ggml_backend_vk_context * ctx = (ggml_backend_vk_context *)backend_dst->context;
 
-    if (dst->buffer->buft != ggml_backend_vk_get_default_buffer_type(backend_dst) &&
-        dst->buffer->buft != ggml_backend_vk_host_buffer_type()) {
+    if (dst->buffer->buft != ggml_backend_vk_get_default_buffer_type(backend_dst)) {
         return false;
     }
 
@@ -13529,6 +13528,11 @@ static bool ggml_backend_vk_cpy_tensor_async(ggml_backend_t backend_src, ggml_ba
 
     if (ggml_backend_buffer_is_vk(src->buffer)) {
         ggml_backend_vk_buffer_context * src_buf_ctx = (ggml_backend_vk_buffer_context *)src->buffer->context;
+
+        // Async copy only works within the same device
+        if (src_buf_ctx->dev_buffer->device != dst_buf->device) {
+            return false;
+        }
 
         vk_context compute_ctx = ggml_vk_get_compute_ctx(ctx);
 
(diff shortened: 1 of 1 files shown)


inforithmics:
Would this help with --cpu-moe (weight offloading to cpu) ?

0cc4m:
It should, yes. Any partial offload.

inforithmics:
I checked out the pull request and indeed it was faster too with ---n-cpu-moe (on an 890M Integrated GPU). 
Even the TG Performance went up from 13 to 15 tg/s with ---n-cpu-moe 20 and gpt-oss:20B.

characharm:
common_init_from_params: warming up the model with an empty run - please wait ... (--no-warmup to disable)
S:\LLM\Vulkan\llama.cpp\ggml\src\ggml-vulkan\ggml-vulkan.cpp:6816: GGML_ASSERT(src->device == dst->device) failed

disabling intel gpu - no crash

0cc4m:
@characharm Thanks for testing it, there was a check missing. Can you try again?

characharm:
> @characharm Thanks for testing it, there was a check missing. Can you try again?

Yeah, looks like it's not crashing anymore.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -13972,6 +14077,8 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
     bool first_node_in_batch = true; // true if next node will be first node in a batch
     int submit_node_idx = 0; // index to first node in a batch
 
+    ggml_vk_submit_transfer_ctx(ctx);
```
jeffbolznv, line 14080: does this case need to set submit_pending?

jeffbolznv (approved):

ggml/src/ggml-vulkan/ggml-vulkan.cpp
0cc4m: I don't think so, the transfer is submitted and it is only important that the next compute submission waits for the semaphore, not that it is immediately also submitted.

rhjdvsgsgks:
> Then I checked and found that the dedicated transfer queue on AMD is not exposed by the Linux driver by default, so it's not actually being used.

transfer queue is supported since mesa 26. but hidden under a flag [`RADV_PERFTEST=transfer_queue`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25594)


> I couldn't test Intel because the Linux driver only exposes a single queue.

it is also supported on intel. but only avaliable when using [xe drm driver (rather than i915) and gen > 12.5](https://gitlab.freedesktop.org/mesa/mesa/-/blob/d6eb6c58c7580ee68d500fda6f8c04bb76aeae67/src/intel/vulkan/anv_physical_device.c#L2511-2522)

0cc4m:
> > Then I checked and found that the dedicated transfer queue on AMD is not exposed by the Linux driver by default, so it's not actually being used.
> 
> transfer queue is supported since mesa 26. but hidden under a flag [`RADV_PERFTEST=transfer_queue`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25594)

I know, I've tested it. It's slightly faster than using the graphics queue. The majority of the improvement comes from any second queue.

> > I couldn't test Intel because the Linux driver only exposes a single queue.
> 
> it is also supported on intel. but only avaliable when using [xe drm driver (rather than i915) and gen > 12.5](https://gitlab.freedesktop.org/mesa/mesa/-/blob/d6eb6c58c7580ee68d500fda6f8c04bb76aeae67/src/intel/vulkan/anv_physical_device.c#L2511-2522)

That's good to know, thank you! I'll give it a try with Xe.

0cc4m:
I get mixed results with Xe, slightly improved pp, slightly worse tg. I'll leave it off there, for now.

acbits:
> > Then I checked and found that the dedicated transfer queue on AMD is not exposed by the Linux driver by default, so it's not actually being used.
> 
> transfer queue is supported since mesa 26. but hidden under a flag [`RADV_PERFTEST=transfer_queue`](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25594)
> 
> > I couldn't test Intel because the Linux driver only exposes a single queue.
> 
> it is also supported on intel. but only avaliable when using [xe drm driver (rather than i915) and gen > 12.5](https://gitlab.freedesktop.org/mesa/mesa/-/blob/d6eb6c58c7580ee68d500fda6f8c04bb76aeae67/src/intel/vulkan/anv_physical_device.c#L2511-2522)

So do we have to upgrade to Mesa 26 or is the RADV_PERFTEST flag is also available on older versions?

0cc4m:
It's available on older versions as well. I'm not sure why it was reported again for 26.0

HumerousGorgon:
This merged PR seems to be causing issues on the vulkan backend regarding gibberish output with -fa on on Intel GPUs.

tvall43:
> This merged PR seems to be causing issues on the vulkan backend regarding gibberish output with -fa on on Intel GPUs.

im also just getting gibberish on my setup with this. qwen3.5-35b on 2 radeon pro v340l's. reverting this fixed it, havent done any other debugging

0cc4m:
Please see if #20233 helped.

tvall43:
> Please see if #20233 helped.

seems like that fixed it! 
edit: uh nevermind. also needed this https://github.com/ggml-org/llama.cpp/issues/20097#issuecomment-4018909498 but i think its working now

Issue #19153, opened 2026-01-28
Feature Request: init with multiple lora but only using one lora at inference


Lairs:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

Max running lora setting for saving vram.

### Motivation

When I running llama-server.exe with multiple lora, for example:
math lora,
code lora,
chat lora, etc.
I only using one lora at a time, but the vram increased with each lora.
For example, running qwen3 model using 4.7GB vram, and my lora is 0.3 GB,
running llama-server.exe with one lora will be 5.0GB total,
running llama-server.exe with two lora will be 5.3GB total, and so on.

### Possible Implementation

Maybe I can have a setting for all lora (index, path, size, etc) 
and allocate vram for lora.
When inference, the request has lora index, so the server can load the target lora weight to the allocated vram then inference.
This idea from vllm serving lora adapter feature, but I did not check the code implement.

renne:
Not stale!

PR #19983, opened 2026-02-28
Re-enable manual LoRA adapter free

PopFlamingo:
This PR proposes re-enabling manual LoRA adapter free (the `llama_adapter_lora_free` function), that had been previously deprecated as part of #18490.

# Motivation
After reading the discussion on why `llama_adapter_lora_free` was deprecated and made a no-op, my understanding is that it was considered that this had no real use case, however I am currently working on a project where having the ability to unload adapters is very important due to memory constraints (mobile devices).

Without `llama_adapter_lora_free`, we have to fully unload and re-load the model, as well as lose any contexts (and their cached tokens) associated with it, just to free the memory associated with those LoRAs. I am not aware of any other way to achieve this.

# Summary of changes
The `llama_adapter_lora_free` function has been re-enabled and un-deprecated; calling `llama_adapter_lora_free` stays optional as the LoRA will still be freed if necessary when its parent model is released. Documentation comments have been updated to reflect this new behavior.

Concretely we make sure to remove the freed LoRA from the ownership list of its owning model to prevent double frees.

# Other related issues
Issue #19153 would benefit from this change as well, since I don't think the requested feature would be implementable without `llama_adapter_lora_free`.

### DIFF
--- a/include/llama.h
+++ b/include/llama.h
@@ -647,9 +647,8 @@ extern "C" {
     LLAMA_API int32_t llama_adapter_meta_val_str_by_index(const struct llama_adapter_lora * adapter, int32_t i, char * buf, size_t buf_size);
 
     // Manually free a LoRA adapter
-    // NOTE: loaded adapters will be free when the associated model is deleted
-    LLAMA_API DEPRECATED(void llama_adapter_lora_free(struct llama_adapter_lora * adapter),
-            "adapters are now freed together with the associated model");
+    // NOTE: loaded adapters that are not manually freed will be freed when the associated model is deleted
+    LLAMA_API void llama_adapter_lora_free(struct llama_adapter_lora * adapter);
 
     // Get the invocation tokens if the current lora is an alora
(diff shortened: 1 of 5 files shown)


include/llama.h
ggerganov: > // All adapters must be loaded before context creation

Is this comment still valid? If not, should remove it.

PopFlamingo: @ggerganov As far as I understand this comment can indeed be removed and potentially could have been since an earlier commit: based on a quick review of the commit and PR  (#18469) that introduced this comment, it seems to me this was necessary at the time because it would have made graph sizing calculations incorrect, but a later change in #18547 introduced the `sched_need_reserve` flag which re-updates the size when needed while taking into account the LoRAs (see `llama_context::graph_max_nodes` and how it loops over the model attached LoRAs).
So after this commit it seems like this has become stale / more of a defensive contract, and it might even prevent certain use cases such as loading aLoRAs on demand rather than from the very beginning. Removed it and also another related comment in `common.cpp`.

ggerganov (approved):

CISC (approved):

PR #19985, opened 2026-02-28
cli : add command and file auto-completion

CISC:
Adds auto-completion callback to `console` and implements command and file auto-completion in `cli`.

### DIFF
--- a/common/console.cpp
+++ b/common/console.cpp
@@ -80,6 +80,8 @@ namespace console {
     static termios      initial_state;
 #endif
 
+    static completion_callback completion_cb = nullptr;
+
     //
     // Init and cleanup
     //
@@ -493,7 +495,7 @@ namespace console {
     }
 
     static void set_line_contents(std::string new_line, std::string & line, std::vector<int> & widths, size_t & char_pos,
-                                  size_t & byte_pos) {
+                                  size_t & byte_pos, int cursor_byte_pos = -1) {
         move_to_line_start(char_pos, byte_pos, widths);
         clear_current_line(widths);
 
@@ -503,6 +505,7 @@ namespace console {
         char_pos = 0;
 
         size_t idx = 0;
+        int back_width = 0;
         while (idx < line.size()) {
             size_t advance = 0;
             char32_t cp = decode_utf8(line, idx, advance);
@@ -511,8 +514,15 @@ namespace console {
             if (real_width < 0) real_width = 0;
             widths.push_back(real_width);
             idx += advance;
-            ++char_pos;
-            byte_pos = idx;
+            if (cursor_byte_pos >= 0 && static_cast<size_t>(cursor_byte_pos) < idx) {
+                back_width += real_width;
+            } else {
+                ++char_pos;
+                byte_pos = idx;
+            }
+        }
+        if (cursor_byte_pos >= 0) {
+            move_cursor(-back_width);
         }
     }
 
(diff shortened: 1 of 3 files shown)


ngxson:
nice feature, @ServeurpersoCom could you please test this on windows? thanks!

ServeurpersoCom:
That's cool!

Feature request: It would be great if directory contents were automatically listed when the path reaches a separator (\ or /), similar to shell behavior.
https://github.com/user-attachments/assets/e5c1fc50-10e5-40c2-941a-9004ed0e5660

Bug: When moving the cursor back (left arrow) to a separator character and pressing Tab, the separator gets repeated across the entire line (e.g. \\\\\\\\\\\...).
https://github.com/user-attachments/assets/3861a0a1-6fc3-4b71-bc1c-b53163d8b718

CISC:
> Feature request: It would be great if directory contents were automatically listed when the path reaches a separator (\ or /), similar to shell behavior.

That would be the TODO.

> Bug: When moving the cursor back (left arrow) to a separator character and pressing Tab, the separator gets repeated across the entire line (e.g. \\\\\...).

Not a bug in this PR, this happens in master.

ServeurpersoCom:
Found this while testing, existing issue but your PR exposes it, here's a one-liner you can include:)

```
(root|~/llama.cpp.pascal) git diff
diff --git a/common/console.cpp b/common/console.cpp
index a770416ab..1fb278c14 100644
--- a/common/console.cpp
+++ b/common/console.cpp
@@ -995,7 +995,7 @@ namespace console {
                 }
             }

-            if (!line.empty() && (line.back() == '\\' || line.back() == '/')) {
+            if (!line.empty() && char_pos == widths.size() && (line.back() == '\\' || line.back() == '/')) {
                 replace_last(line.back());
                 is_special_char = true;
             }
(root|~/llama.cpp.pascal)
```

CISC:
> Found this while testing, existing issue but your PR exposes it, here's a one-liner you can include:)

Should be a separate PR, though I'm not sure what that code was supposed to do and if that's the correct fix or not?

ServeurpersoCom:
> > Found this while testing, existing issue but your PR exposes it, here's a one-liner you can include:)
> 
> Should be a separate PR, though I'm not sure what that code was supposed to do and if that's the correct fix or not?

Yes I can make short demo videos or even a PR, I tested it under Linux and Windows, it fixes the bug of the repeating character. Furthermore, I believe it was me who worked on the cursor movement with arrows keys.

CISC:
> > Should be a separate PR, though I'm not sure what that code was supposed to do and if that's the correct fix or not?
> 
> Yes I can make short demo videos or even a PR, I tested it under Linux and Windows, it fixes the bug of the repeating character. Furthermore, I believe it was me who worked on the cursor movement with arrows keys.

The reason I'm saying it should be a separate PR is so that we don't hide a bugfix in an unrelated PR.

Anyway, maybe you can explain what the intention was? :) I don't understand what it's trying to do with the slashes?

ServeurpersoCom:
> The reason I'm saying it should be a separate PR is so that we don't hide a bugfix in an unrelated PR.
I completely agree

> Anyway, maybe you can explain what the intention was? :) I don't understand what it's trying to do with the slashes?

Ah yes, it's multi-line like in bash! It works well! But I need to test it again with the PR; I completely forgot about that feature! 
To try, enter :

```
Hello \
It's another line \
... enter
```

https://github.com/ggml-org/llama.cpp/pull/1040

CISC:
> > Anyway, maybe you can explain what the intention was? :) I don't understand what it's trying to do with the slashes?
> 
> Ah yes, it's multi-line like in bash! It works well!

Must keep that in mind if/when adding another feature PR; reading whole directories (though maybe just wildcards is more useful).

CISC:
gentle ping for approval :)

ggerganov (approved):
Works on Mac

PR #19991, opened 2026-02-28
vulkan: On Windows use DXGI + PDH to detect the correct amount of Free GPU Memory

inforithmics:
I discovered that on Windows Vulkan reports to much Free Memory. 
For example on my 890M GPU it says that there are 17.3 GB Free but in reality there is 16.9 GB free. With this Code (taken from Ollama patch https://github.com/ollama/ollama/blob/main/llama/patches/0028-Add-memory-detection-using-DXGI-PDH.patch It uses DXGI + PDH to detect the free Memory).

This pull request is meant as discussion if this is acceptable in principle. If yes I would refine this pull request further.

### DIFF
--- a/ggml/src/CMakeLists.txt
+++ b/ggml/src/CMakeLists.txt
@@ -205,6 +205,7 @@ add_library(ggml-base
             ggml-threading.h
             ggml-quants.c
             ggml-quants.h
+            mem_dxgi_pdh.cpp
             gguf.cpp)
 
 set_target_properties(ggml-base PROPERTIES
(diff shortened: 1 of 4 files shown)


jeffbolznv:
I didn't look at this in great detail, but IMO basing the free memory on total dedicated memory would be wrong. The memory budget extension is intended to match the WDDM process residency budget (https://learn.microsoft.com/en-us/windows-hardware/drivers/display/process-residency-budgets) and we should avoid going beyond that. I'd also prefer not to add dependencies on these additional DLLs if we can avoid it.

inforithmics:
Closing this because it would add Windows specific Code/Dependency.

PR #19992, opened 2026-02-28
common : fix common_chat_peg_parse for incomplete utf-8 sequence tail

akreal:
Sometimes I experience an issue with `mistralai/Ministral-3-3B-Instruct-2512-GGUF:Q8_0` when it hits maximum generated tokens and leaves an incomplete UTF-8 sequence in the end of message (e.g. token IDs 1226 and 1156). This message then causes `common_chat_peg_parse` to fail with `Failed to parse input at pos 0` error.

Claude Sonnet 4.6 was used to analyze the problem and to generate the code. Everything was manually reviewed and tested.

### DIFF
--- a/common/chat-parser.cpp
+++ b/common/chat-parser.cpp
@@ -4,6 +4,7 @@
 #include "log.h"
 #include "peg-parser.h"
 #include "regex-partial.h"
+#include "unicode.h"
 
 #include <algorithm>
 #include <cctype>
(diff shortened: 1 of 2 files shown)


akreal:
After more experiments it appeared that invalid UTF-8 sequence can appear also in the middle of string, so now the fix handles both mid-string and end-string conditions. But I am not sure if this is the correct place to do this.

Issue #13250, opened 2025-05-02
When using the qwen2.5-vl model on AMD Ryzen APU under Windows, the error "failed to allocate Vulkan0 buffer of size 4342230552" may appear.


xeden3:
### Name and Version

C:\Users\xeden\Downloads\llama-b5255-bin-win-vulkan-x64>llama-cli --version
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = AMD Radeon(TM) 8060S Graphics (AMD proprietary driver) | uma: 1 | fp16: 1 | warp size: 64 | shared memory: 32768 | int dot: 1 | matrix cores: KHR_coopmat
version: 5255 (d24d5928)
built with MSVC 19.43.34808.0 for x64

### Operating systems

Windows

### GGML backends

Vulkan

### Hardware

CPU AMD Ryzen AI MAX 395 Memory 128G （CPU 64G GPU 64G）

### Models

Qwen2.5-VL-3B-Instruct-f16.gguf

### Problem description & steps to reproduce

Device
CPU AMD Ryzen AI MAX 395
Memory
128GB GPU 64mb, CPU 64mb
Operating system
win11
Used llama.cpp version
llama-b5255-bin-win-vulkan-x64

Since AMD does not support ROCM of Ryzen AI MAX 395, I used vulkan as the backends, and it is no problem to run most of the llm models, including deepseek.

### Relevant log output

```shell
C:\Users\xeden\Downloads\llama-b5255-bin-win-vulkan-x64>llama-mtmd-cli -m Qwen2.5-VL-3B-Instruct-f16.gguf --mmproj mmproj-Qwen2.5-VL-3B-Instruct-f16.gguf -p '描述图片内容.' --image demo.jpg
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = AMD Radeon(TM) 8060S Graphics (AMD proprietary driver) | uma: 1 | fp16: 1 | warp size: 64 | shared memory: 32768 | int dot: 1 | matrix cores: KHR_coopmat
build: 5255 (d24d5928) with MSVC 19.43.34808.0 for x64
llama_model_load_from_file_impl: using device Vulkan0 (AMD Radeon(TM) 8060S Graphics) - 65536 MiB free
llama_model_loader: loaded meta data with 27 key-value pairs and 434 tensors from Qwen2.5-VL-3B-Instruct-f16.gguf (version GGUF V3 (latest))
[...]

PR #20017, opened 2026-03-01
vulkan: add sparse OOM fallback for large UMA allocations and chunked staging fallback

neilopet:
Problem:
- On UMA under heap fragmentation, large contiguous device-local allocations can fail even when total free memory is sufficient.
- Large staging allocations can also fail during load/checkpoint transfers.

Changes:
- Keep existing contiguous-first allocator behavior.
- Add sparse binding as OOM fallback only for eligible large (`>= 512 MB`) device-local buffers on UMA with sparse-capable compute queue support.
- Preserve BDA correctness for sparse-backed buffers and keep sparse failure non-fatal in the fallback chain.
- Add try-large-first staging behavior with sticky 32 MB cap learning and chunk fallback, plus transfer-context reuse in chunk loops.
- Add focused regression tests in `tests/test-vulkan-sparse.cpp`.

Validation:
- `test-vulkan-sparse`: pass (deterministic branch subtests with `-DGGML_VULKAN_SPARSE_TESTING=ON`).
- `test-backend-ops`: baseline-matched RDNA f16 DIV failures only.
- `llama-bench` local UMA: no meaningful regression on measured models.
- `llama-bench` A/B on RunPod RTX 4090 (5 successful runs each):
  - Baseline: `pp512=6825.08`, `tg128=148.14`, `pp512+tg128=639.79`
  - Head: `pp512=6812.29`, `tg128=148.23`, `pp512+tg128=639.85`
  - Delta: `-0.19% / +0.06% / +0.01%`
- Models used:
    - Local UMA benchmarks: `Qwen3.5-27B-Q4_K_M.gguf`, `Qwen3-Coder-Next-Q4_K_M.gguf`
    - RunPod A/B benchmark: `Qwen2.5-7B-Instruct-Q4_K_M.gguf` (`-pg 512,128 -r 5`)
 
_Note: Subtests that require `GGML_VULKAN_SPARSE_TESTING` are deterministic path-reachability checks and are expected to skip when that build option is off._

Scope / consistency:
- No shader changes.
- Non-UMA behavior remains unchanged.
- Existing strict staging-buffer contract is preserved.

Related: #15054, #13250

AI Disclosure: AI tools were used for review and test planning.


jeffbolznv:
I don't understand why this change is needed. Normal vkdevicememory allocations don't need to be physically contiguous. There's a comment in the code about "3x perf" but the perf data in the description doesn't show any benefit, and doesn't describe what actual use cases are failing.

PR #18047, opened 2025-12-15
vulkan: Implement set_tensor_async and the event interfaces

jeffbolznv:
The goal is to enable the async loading code paths in llama_model_loader::load_all_data, originally from #7896. This works and the loads themselves are faster, but with host visible vidmem I think the cost of allocating/mapping vidmem moves and becomes more expensive, and I don't see a benefit by default. But with GGML_VK_DISABLE_HOST_VISIBLE_VIDMEM=1 I do see a significant improvement in model loading time.

It would be interesting to test on Linux how this interacts with #18012.

PR #16059, opened 2025-09-17
vulkan: optimize UMA buffer operations and fix driver hangs

giuseppe:
The previous implementation was blocking the GPU for extended periods, causing the i915 driver to reset the context due to the hangcheck protection.

[32628.443070] i915 0000:00:02.0: [drm] GPU HANG: ecode 12:1:85dffffb, in llama-server [194114]
[32628.443091] i915 0000:00:02.0: [drm] llama-server[194114] context reset due to GPU hang

*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

PR #20018, opened 2026-03-01
vulkan: add UMA zero-copy async transfers and fix event_record deferred memcpy handling

neilopet:
Problem:
- On UMA, host-visible mapped buffers were still using staging in async transfer paths.
- When the UMA zero-copy async path is used, deferred memcpy queues can be pending at event_record; without drain/preserve across context reset, copies can be dropped.

Changes:
- Add UMA zero-copy async fast paths for host-visible mapped buffers in `ggml_vk_buffer_write_2d_async` and `ggml_vk_buffer_read_2d_async`.
- In `ggml_backend_vk_event_record`, drain pending `in_memcpys` before submit and preserve/restore `out_memcpys` across compute-context reset.
- Add regression tests in `tests/test-vulkan-uma.cpp` for round-trip, multiple event-record sequences, empty-queue path, and out-memcpy persistence.

Validation:
- `test-vulkan-uma`: 4/4 pass.
- `test-backend-ops`: baseline-matched RDNA f16 DIV failures only.
- `llama-bench` local UMA:
  - `Qwen3.5-27B-Q4_K_M`: `pp512=152.75`, `tg128=11.23`
  - `Fine-tuned private Qwen3-Coder-Next-Q4_K_M`: `pp512=460.61`, `tg128=42.69`

Scope / consistency:
- This builds on #19976 transfer-context changes and only adds deferred memcpy lifecycle handling for UMA zero-copy paths.
- Preserves existing event lifecycle structure (`ctx_end -> submit -> reset/recreate`), adding only deferred-copy correctness handling.
- UMA zero-copy path remains gated by (`device->uma && HOST_VISIBLE`).
- No shader changes.
- No changes to non-UMA transfer behavior.

Related: #16059, #18302, #18047

AI Disclosure: AI tools were used for review and test planning.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -6647,6 +6647,21 @@ static bool ggml_vk_buffer_write_2d_async(vk_context subctx, vk_buffer& dst, siz
         subctx->s->buffer.copyBuffer(buf->buffer, dst->buffer, slices);
         return true;
     }
+
+    // UMA zero-copy: destination is directly mapped, skip staging buffer
+    if (dst->memory_property_flags & vk::MemoryPropertyFlagBits::eHostVisible &&
+        dst->device->uma) {
+        if (width == spitch) {
+            deferred_memcpy((uint8_t *)dst->ptr + offset, src, width * height, &subctx->in_memcpys);
+        } else {
+            for (size_t i = 0; i < height; i++) {
+                deferred_memcpy((uint8_t *)dst->ptr + offset + i * width,
+                                (const uint8_t *)src + i * spitch, width, &subctx->in_memcpys);
+            }
+        }
+        return true;
+    }
+
     VK_LOG_DEBUG("STAGING");
 
     if (!sync_staging) {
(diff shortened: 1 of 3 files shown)


ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -6647,6 +6647,21 @@ static bool ggml_vk_buffer_write_2d_async(vk_context subctx, vk_buffer& dst, siz
         subctx->s->buffer.copyBuffer(buf->buffer, dst->buffer, slices);
         return true;
     }
+
+    // UMA zero-copy: destination is directly mapped, skip staging buffer
```
jeffbolznv, line 6651: I don't think this is correct. When called from set_tensor_async, I think the operation needs to be ordered against other work submitted to the same backend. When called from set_tensor, we already have a CPU copy path so this would be redundant.

tests/test-vulkan-uma.cpp
```diff
@@ -0,0 +1,343 @@
+// Regression tests for UMA zero-copy async buffer transfers.
```
jeffbolznv, line 1: FWIW, I do think we need more thorough tests of the set/get paths and ordering guarantees of the backend interfaces. But if we have 300 lines of AI-generated tests per PR nobody will be able to maintain it.

jeffbolznv (changes requested):
A change like this needs more clear before/after perf results.

PR #20019, opened 2026-03-01
gguf: add big-endian magic "FUGG" for explicit endianness detection

### LINKED ISSUE #3957: GGUF endianness cannot be determined from GGUF itself

Scottcjn:
## Summary

Resolves #3957 — GGUF files currently use the same magic (`GGUF`) regardless of byte order, making endianness detection impossible from the header alone.

This PR adds `FUGG` (byte-reversed `GGUF`) as the magic number for big-endian GGUF files, providing explicit endianness detection.

### C++ (`ggml/include/gguf.h`, `ggml/src/gguf.cpp`)
- Add `GGUF_MAGIC_BE "FUGG"` constant
- **Reader**: accepts both `GGUF` (LE) and `FUGG` (BE) magic
  - `FUGG` on LE host → clear error with conversion command
  - `FUGG` on BE host → proceeds normally (native endian)
  - Legacy version-field heuristic preserved as fallback for old BE files
- **Writer**: emits correct magic based on host endianness
- Improved error messages with actionable conversion commands

### Python (`gguf-py/gguf/`)
- `constants.py`: Add `GGUF_MAGIC_BE = 0x47475546`
- `gguf_reader.py`: Detect endianness from magic (primary) with version-field fallback (legacy)
- `gguf_writer.py`: Write `FUGG` magic when target endianness is big-endian
- `scripts/gguf_convert_endian.py`: Update magic bytes during LE↔BE conversion

### Tests (`tests/test-gguf.cpp`)
- Bad magic test updated from `FUGG` → `XXXX` (since `FUGG` is now valid)

## Design Decisions

1. **Semantics**: `GGUF` = little-endian file, `FUGG` = big-endian file (always, regardless of host)
2. **Backward compatible**: Old BE files with `GGUF` magic still detected via version-field heuristic
3. **C++ reader doesn't byte-swap** (too large a change) — gives clear error pointing to `gguf_convert_endian.py`
4. **Python reader handles both** natively (already had cross-endian support)

## Test Results

Tested on real big-endian hardware: **Power Mac G5** (Darwin 9.8.0, PowerPC ppc64, big-endian) and **x86_64** (Ubuntu, little-endian).

### C standalone test (5 cases)

| Test | G5 (BE) | x86_64 (LE) |
|------|---------|-------------|
| Native-endian write/read | PASS | PASS |
| Cross-endian detection | PASS | PASS |
| Legacy BE file (GGUF magic + BE data) | PASS | PASS |
| Invalid magic rejection | PASS | PASS |
| Magic byte layout verification | PASS | PASS |

Key result: G5 natively writes `FUGG` magic (`46 55 47 47`), x86_64 writes `GGUF` (`47 47 55 46`).

### Cross-platform file transfer

Copied G5-generated big-endian GGUF file to x86_64 host. Hex dump confirms FUGG magic:
```
00000000: 4655 4747 0000 0003 0000 0000 0000 0000  FUGG............
00000010: 0000 0000 0000 0000                      ........
```

### Python gguf_reader (4 cases, on x86_64 LE host)

| Test | Result | Details |
|------|--------|---------|
| LE file (GGUF magic) | PASS | byte_order='I', version=3 |
| BE file (FUGG magic, from G5) | PASS | byte_order='S', version=3 |
| Legacy BE file (GGUF magic + BE data) | PASS | byte_order='S', version=3 (detected via version-field heuristic) |
| Invalid magic | PASS | Correctly raises ValueError |

### Bug found and fixed during testing

The initial implementation had a double byte-swap bug in `gguf_reader.py`: when FUGG magic set `byte_order='S', `_get()` already applied the swap via `newbyteorder()`. An `elif` branch was re-swapping, un-doing the correction (version 3 read as 50,331,648). Fixed in second commit.

## Test Hardware

- **Power Mac G5** (192.168.0.179): Darwin 9.8.0, PowerPC ppc64, big-endian, GCC 10.5.0
- **x86_64 workstation**: Ubuntu, little-endian, Python 3.x with numpy

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/ggml/include/gguf.h
+++ b/ggml/include/gguf.h
@@ -38,7 +38,8 @@
 #include <stdbool.h>
 #include <stdint.h>
 
-#define GGUF_MAGIC   "GGUF"
+#define GGUF_MAGIC    "GGUF"
+#define GGUF_MAGIC_BE "FUGG"
 #define GGUF_VERSION 3
 
 #define GGUF_KEY_GENERAL_ALIGNMENT "general.alignment"
--- a/ggml/src/gguf.cpp
+++ b/ggml/src/gguf.cpp
@@ -411,16 +411,32 @@ struct gguf_context * gguf_init_from_file_impl(FILE * file, struct gguf_init_par
             return nullptr;
         }
 
-        for (uint32_t i = 0; i < magic.size(); i++) {
-            if (magic[i] != GGUF_MAGIC[i]) {
-                char c0 = isprint(magic[0]) ? magic[0] : '?';
-                char c1 = isprint(magic[1]) ? magic[1] : '?';
-                char c2 = isprint(magic[2]) ? magic[2] : '?';
-                char c3 = isprint(magic[3]) ? magic[3] : '?';
-                GGML_LOG_ERROR("%s: invalid magic characters: '%c%c%c%c', expected 'GGUF'\n", __func__, c0, c1, c2, c3);
+        const bool is_le_magic = memcmp(magic.data(), GGUF_MAGIC,    4) == 0;
+        const bool is_be_magic = memcmp(magic.data(), GGUF_MAGIC_BE, 4) == 0;
+
+        if (!is_le_magic && !is_be_magic) {
+            char c0 = isprint(magic[0]) ? magic[0] : '?';
+            char c1 = isprint(magic[1]) ? magic[1] : '?';
+            char c2 = isprint(magic[2]) ? magic[2] : '?';
+            char c3 = isprint(magic[3]) ? magic[3] : '?';
+            GGML_LOG_ERROR("%s: invalid magic characters: '%c%c%c%c', expected 'GGUF' or 'FUGG'\n", __func__, c0, c1, c2, c3);
+            gguf_free(ctx);
+            return nullptr;
+        }
+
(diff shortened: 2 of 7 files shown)


taronaeo:
Please read the contributing guidelines as shown here. Likewise goes to your other opened PRs too: https://github.com/ggml-org/llama.cpp/blob/319146247e643695f94a558e8ae686277dd4f8da/CONTRIBUTING.md?plain=1#L9-L25

Also, Big-Endian (BE) GGUFs have been supported for awhile now with `llama-cli` throwing warning messages if heuristics determine that the byte-order is incorrect.

I'm not sure if this is still an "issue" to be fixed.

Scottcjn:
Thanks for the review @taronaeo. You're right that BE GGUFs are handled via heuristics today. The motivation was to make endianness detection explicit rather than heuristic-based (the current approach reads the magic as LE and infers BE if it doesn't match), but I understand if this isn't considered a priority.

I've read the contributing guidelines — I'll make sure any future PRs follow the process (CI, perplexity checks, etc). If the consensus is this isn't needed, I'm happy to close it.

PR #20048, opened 2026-03-02
ggml: add Q1_0 and Q1_0_g128 1-bit quantization support (CPU, Metal, …

khosravipasha:
## ggml: add Q1_0 and Q1_0_g128 1-bit quantization support (CPU, Metal, CUDA)

Adds two new 1-bit quantization types to ggml/llama.cpp:

- **Q1_0**: 1 bit per weight with block size 32 (6 bytes per block: 2-byte fp16 scale + 4 bytes of bit-packed weights). Each bit encodes the sign: `0 → -scale`, `1 → +scale`. Effective ~1.5 bpw including scale overhead.
- **Q1_0_g128**: Similar to `Q1_0` but with group size 128. Effective ~1.125 bpw instead.

Backend support: CPU (x86 SSE/AVX + ARM NEON), Metal, and CUDA (including MMQ kernels). Metal and CUDA kernels are fairly optimized. The CPU path is functional, but it may not be as optimized yet (especially on ARM NEON) since I am not too familiar best practices on CPUs like (simd, etc).

### Implementation approach

The kernel implementations follow Q4_0 as the reference, adapting it for 1-bit dequantization and dot-product logic. The reason is that Q4_0 is very similar to Q1_0 and for simplicity kept things similar.

### Why two types?

**Q1_0** uses block size 32, which is consistent with llama.cpp's existing naming convention (Q4_0, Q8_0, etc. all use block size 32). This keeps the code structure familiar and fits naturally into the existing quantization infrastructure.

However, block size 32 means the 2-byte fp16 scale is used for each group of 32 weights, resulting in ~1.5 bpw overhead. **Q1_0_g128** uses block size 128 to get ~1.125 bpw. The larger block size requires special handling in the kernels (especially on CUDA, where the MMQ tile sizes and thread mappings differ from the standard block-32 path).

The naming `Q1_0` seemed closest to llama.cpp conventions (can be called `Int1` from other naming conventions), happy to rename if the maintainers prefer something else. The `Q1_0_g128` is mostly same as `Q1_0` but with group size 128.

## Correctness Validation

- **Packed vs unpacked KL check (kernel correctness)**: We start with a `Q1_0` weights, then unpacked them to equivalent weights. The packed and unpacked weights are functionally equivalent just stored differently so their output should be the same.  We use KL to show this is the case and that the Q1_0 kernels are numerically correct (see Appendix: KL metrics (packed vs unpacked) for more details on setup).

## Performance Benchmarks

Expected performance characteristics of these formats:

- **Token generation** should benefit strongly from reduced memory traffic, since 1-bit formats significantly reduce weight bandwidth.
- **`Q1_0_g128`** generally improves effective bits-per-weight vs `Q1_0` by amortizing scale metadata over larger blocks.
- **Prompt processing on CUDA** is currently limited by a cuBLAS fallback path for `Q1_0`/`Q1_0_g128` in some cases (see note below), so prompt-side gains are currently not representative of the final intended path.

Note: prompt processing for Q1_0/Q1_0_g128 on CUDA currently falls back to cuBLAS dequantization instead of using the native MMQ path. The MMQ kernel code is included (see `mmq.cuh`: `load_tiles_q1_0`, `vec_dot_q1_0_q8_1_dp4a`, and the `mmq_type_traits` specialization) but is disabled in `mmq.cu` (`ggml_cuda_mul_mat_q_switch_type` and `ggml_cuda_should_use_mmq`) due to accuracy issues on some configurations. Community help debugging the MMQ path would be appreciated  (once enabled, pp performance should be on par with Q4_0. On Metal, prompt processing is on par with FP16.

## What we tested

- KL divergence of packed (Q1_0 / Q1_0_g128) vs unpacked (FP16) on Metal and CUDA (see Appendix)
- Performance benchmarks (`llama-bench`) on Metal and CUDA (see Appendix)
- Hardware used: Apple M4 Pro 48 GB (Metal), NVIDIA RTX 4090 (CUDA)

## Questions for maintainers

1. **CUDA MMQ path**: The MMQ kernel code for Q1_0/Q1_0_g128 is included but disabled due to accuracy issues. Should I remove it or leave it as-is for someone to debug later? Currently cuBLAS fallback is used for prompt processing on CUDA.
2. **Naming**: Is `Q1_0` / `Q1_0_g128` okay, or would you prefer a different convention?
3. **CPU kernel optimization**: How important is it for the CPU kernels to be fully optimized before merging? The Metal and CUDA kernels should be fairly optimized, but I'm less experienced with CPU-specific SIMD optimization.
4. **Anything else needed?** Happy to do more testing, benchmarks, or changes to help with merging.

### Performance

Benchmarks below are a sanity check for the new kernels, they confirm correctness and give a rough sense of throughput. These numbers represent an upper bound on speed since the model was quantized with `--pure` (all tensors packed to 1-bit), and flash attention was enabled (`-fa 1`). Results are from an **8B model** measured with `llama-bench`.

<details>
<summary>Apple M4 Pro (Metal)</summary>

| Quant     | pp512 (t/s) | tg128 (t/s) |
|-----------|-------------|-------------|
| FP16      | 459         | 16.6        |
| Q4_0      | 472         | 51.1        |
| Q1_0      | 486         | 88.0        |
| Q1_0_g128 | 490         | 88.4        |

</details>

<details>
<summary>NVIDIA RTX 4090 (CUDA)</summary>

| Quant     | pp512 (t/s) | tg128 (t/s) |
|-----------|-------------|-------------|
| FP16      | 10,951      | 60          |
| Q4_0      | 12,927      | 176         |
| Q1_0      | 7,856       | 381         |
| Q1_0_g128 | 7,828       | 401         |

Note: prompt processing currently uses cuBLAS fallback for `Q1_0` and `Q1_0_g128`. After fixing MMQ for these formats, `pp512` should move closer to `Q4_0` (roughly ~2x room for improvement from current numbers).

</details>

### KL metrics (packed vs unpacked)

Motivation for this section is to validate correctness of `Q1_0` kernels end-to-end. 

Simple setup:
- Start from `Q1_0` (`Q1_0_g128`) weights.
- Unpack those same weights to FP16.
- Run KL-divergence on both prompt-processing (qmm) and generation (qmv) paths across few prompts and wikitext.

For example, lets look at an individual linear layer with one group, if we start from `Q1_0` and unpack into `fp16` we get:
- `Q1_0` layer  => `scale = 0.5`, `bits = 0011....1`
- `fp16`  layer  => `-0.5, -0.5, +0.5, +0.5, .... 0.5`

These two layers are equivlaent and their output should be the same.

<details>
<summary>Q1_0 - Metal</summary>

Prompt processing (qmm path):
- Forward KL mean: **0.000024**
- Reverse KL mean: **0.000017**
- Mean top-1 agreement: **99.85%** (min **99.29%**)

Token generation (qmv path):
- Forward KL mean: **0.000067**
- Reverse KL mean: **0.000038**
- Mean top-1 agreement: **100.0%** (min **100%**)

</details>

<details>
<summary>Q1_0 - CUDA</summary>

`Q1_0` vs FP16:
- Mean KL: **0.000118**
- Top-1 agreement: **99.231%**
- RMS Δp: **0.346%**
- Correlation: **99.99%**

</details>

<details>
<summary>Q1_0_g128 - Metal</summary>

`Q1_0_g128` vs FP16:
- Mean KL: **0.000000**
- Top-1 agreement: **99.984%**
- RMS Δp: **0.015%**

</details>

<details>
<summary>Q1_0_g128 - CUDA</summary>

`Q1_0_g128` vs FP16:
- Mean KL: **0.000120**
- Top-1 agreement: **99.020%**
- RMS Δp: **0.344%**
- Correlation: **99.99%**

</details>

### Kernels implemented

<details>
<summary>Metal</summary>

- `kernel_mul_mv_q1_0_f32` / `kernel_mul_mv_q1_0_g128_f32` — matrix-vector (token generation)
- `kernel_mul_mv_ext_q1_0_f32_r1_{2,3,4,5}` / `kernel_mul_mv_ext_q1_0_g128_f32_r1_{2,3,4,5}` — small-batch mat-vec (batch size 2-8)
- `kernel_mul_mm_q1_0_f32` / `kernel_mul_mm_q1_0_g128_f32` — matrix-matrix (prompt processing)
- `dequantize_q1_0` / `dequantize_q1_0_g128` — dequantization helpers

</details>

<details>
<summary>CUDA</summary>

- `dequantize_q1_0` / `dequantize_q1_0_g128` (`dequantize.cuh`) — dequantization for cuBLAS fallback path
- `load_tiles_q1_0` / `load_tiles_q1_0_g128` (`mmq.cuh`) — MMQ tile loaders
- `vec_dot_q1_0_q8_1_dp4a` / `vec_dot_q1_0_g128_q8_1_dp4a` (`mmq.cuh`) — MMQ dp4a dot product kernels
- `mmq_type_traits` specializations for Q1_0 and Q1_0_g128 (`mmq.cuh`)
- MMQ kernel code is included but currently disabled in `mmq.cu` (`ggml_cuda_mul_mat_q_switch_type` and `ggml_cuda_should_use_mmq`) due to accuracy issues on some configurations, help in debugging this is appreciated (can probably boost prompt processing speed by 2x).

</details>

<details>
<summary>CPU</summary>

- Quantize/dequantize functions for Q1_0 and Q1_0_g128
- Dot product: x86 SSE/AVX, ARM NEON, and scalar fallback
- `vec_dot_q1_0_q8_1` / `vec_dot_q1_0_g128_q8_1` implementations

</details>

### AI usage disclosure

Per the project's [CONTRIBUTING.md](CONTRIBUTING.md) policy:

- The Q4_0 kernels were used as boilerplate. AI was partially used to help adapt each relevant function/kernel for Q1_0 and Q1_0_g128 (replacing 4-bit nibble extraction with 1-bit sign-based dequantization). The output was then manually reviewed, debugged, and modified to produce correct and performant kernels.
- The CPU x86 AVX2 dot-product kernels (`__AVX2__` path) were mostly AI-generated, then reviewed and tested for correctness.
- All AI-assisted code has been thoroughly tested (unit tests, KL divergence, benchmarks) and manually reviewed by the contributor.

### DIFF
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -427,7 +427,9 @@ extern "C" {
         // GGML_TYPE_IQ4_NL_4_8 = 37,
         // GGML_TYPE_IQ4_NL_8_8 = 38,
         GGML_TYPE_MXFP4   = 39, // MXFP4 (1 block)
-        GGML_TYPE_COUNT   = 40,
+        GGML_TYPE_Q1_0    = 40,
+        GGML_TYPE_Q1_0_g128 = 41,
+        GGML_TYPE_COUNT   = 42,
     };
 
     // precision
@@ -463,6 +465,8 @@ extern "C" {
         GGML_FTYPE_MOSTLY_IQ1_M   = 23, // except 1d tensors
         GGML_FTYPE_MOSTLY_BF16    = 24, // except 1d tensors
         GGML_FTYPE_MOSTLY_MXFP4   = 25, // except 1d tensors
+        GGML_FTYPE_MOSTLY_Q1_0    = 26, // except 1d tensors
+        GGML_FTYPE_MOSTLY_Q1_0_g128 = 27, // except 1d tensors
     };
 
     // available tensor operations:
--- a/ggml/src/ggml-common.h
+++ b/ggml/src/ggml-common.h
@@ -93,6 +93,13 @@ typedef sycl::half2 ggml_half2;
 // QR = QK / number of values before dequantization
 // QI = number of 32 bit integers before dequantization
 
+#define QI1_0 (QK1_0 / 32)  // Number of int32s needed for QK1_0 bits (QK1_0/32)
+#define QR1_0 1              // 1 bit per quantized element (matches the 1-bit nature of Q1_0)
+
+#define QI1_0_g128 (QK1_0_g128 / 32)  // Number of int32s needed for QK1_0_g128 bits (QK1_0_g128/32)
+#define QR1_0_g128 1              // 1 bit per quantized element (matches the 1-bit nature of Q1_0_g128)
+
+
 #define QI4_0 (QK4_0 / (4 * QR4_0))
 #define QR4_0 2
 
@@ -166,6 +173,20 @@ typedef sycl::half2 ggml_half2;
 #else // _MSC_VER
 #define GGML_EXTENSION __extension__
 #endif // _MSC_VER
+
+#define QK1_0 32  // MUST match QK8_0 for vec_dot computation! TODO see if we can do larger blocks later
+typedef struct {
+    ggml_half d;           // delta
+    uint8_t qs[QK1_0 / 8]; // bits / quants
+} block_q1_0;
+static_assert(sizeof(block_q1_0) == sizeof(ggml_half) + QK1_0 / 8, "wrong q1_0 block size/padding");
+
+#define QK1_0_g128 128
+typedef struct {
+    ggml_half d;               // delta
+    uint8_t qs[QK1_0_g128 / 8]; // bits / quants
+} block_q1_0_g128;
+static_assert(sizeof(block_q1_0_g128) == sizeof(ggml_half) + QK1_0_g128 / 8, "wrong q1_0_g128 block size/padding");
 
 #define QK4_0 32
 typedef struct {
--- a/ggml/src/ggml-cpu/arch/arm/quants.c
+++ b/ggml/src/ggml-cpu/arch/arm/quants.c
@@ -137,6 +137,146 @@ void quantize_row_q8_K(const float * GGML_RESTRICT x, void * GGML_RESTRICT y, in
 
 //===================================== Dot products =================================
 
+void ggml_vec_dot_q1_0_q8_0(int n, float * GGML_RESTRICT s, size_t bs, const void * GGML_RESTRICT vx, size_t bx, const void * GGML_RESTRICT vy, size_t by, int nrc) {    
+    // For nrc > 1, call generic multiple times
+    if (nrc == 1) {
+        ggml_vec_dot_q1_0_q8_0_generic(n, s, bs, vx, bx, vy, by, nrc);
+    } else {
+        // Handle multiple rows by calling generic for each
+        const int qk = QK8_0;
+        const int nb = n / qk;
+        const size_t x_size = nb * sizeof(block_q1_0);
+        const size_t y_size = nb * sizeof(block_q8_0);
+        
+        for (int i = 0; i < nrc; i++) {
+            ggml_vec_dot_q1_0_q8_0_generic(
+                n,
+                s + i,
+                bs,
+                (const char *)vx + i * x_size,
+                bx,
+                (const char *)vy + i * y_size,
+                by,
+                1
+            );
+        }
+    }
+}
+
+void ggml_vec_dot_q1_0_g128_q8_0(int n, float * GGML_RESTRICT s, size_t bs, const void * GGML_RESTRICT vx, size_t bx, const void * GGML_RESTRICT vy, size_t by, int nrc) {
+    const int qk = QK1_0_g128;  // 128
+    const int nb = n / qk;
+
+    assert(n % qk == 0);
+    assert(nrc == 1);
+    UNUSED(nrc);
+    UNUSED(bx);
+    UNUSED(by);
+    UNUSED(bs);
+
+    const block_q1_0_g128 * GGML_RESTRICT x = vx;
+    const block_q8_0 * GGML_RESTRICT y = vy;
+
+    float sumf = 0.0f;
+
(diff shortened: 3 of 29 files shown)


JohannesGaessler:
There is no way the posted quality metrics are correct. Since the entire PR description is clearly "AI" generated in blatant violation of our guidelines they must be hallucinations.

khosravipasha:
@JohannesGaessler, thanks for the quick feedback, I wrote the PR description myself, and the AI usage is just for polishing the draft. All the details are accurate. 

As for "KL metrics (packed vs unpacked)", its not hallucinated.  I should rephrase this to make it more clear.  The goal is to test correctness of new matmul kernels. We are not directly comparing with a fp16 model, its the other way, we are unpacking a model with Q1_0 weights into fp16 format then running inference in both Q1_0 and fp16 format.  

For example, lets look at an individual linear layer with one group, if we start from `Q1_0` and unpack into `fp16` we get:
- `Q1_0` layer  => `scale = 0.5`, `bits = 0011....1`
- `fp16`  layer  => `-0.5, -0.5, +0.5, +0.5, .... 0.5`

These two linear layers have equivalent weights but stored in different formats (and running coressponding matmul. kernels), their output should be the same, we know fp16 kernels are correct, so the the KL metrics section aims to show the Q1_0 matmuls are correct. That's why KL of almost 0 is expected (if not would show Q_0 kernels had issues)

As for whether this is useful without a model in this format that's a fair question. Happy to chat more if there are any concerns.

JohannesGaessler:
The speed alone is not enough, we would want to see that a model at an extreme quantization is still acceptable in terms of quality. Please use the [perplexity example](https://github.com/ggml-org/llama.cpp/tree/master/tools/perplexity), calculate the KL divergence of a small base model using the new type vs. FP16/BF16 on Wikitext-2, and compare this against existing datatypes of similar size.

It's fine if the CPU implementation is suboptimal, it's primarily needed as a fallback and as a reference implementation.

PR #20056, opened 2026-03-03
vendor : update cpp-httplib to 0.36.0

cabelo:


angt:
Hi @cabelo, thanks for your contribution!
Can you not change the `CMakeLists.txt` except if you find an issue with the current one?
Also, keep the split .h/.cpp. Running the `sync_vendor.py` on the same version should not change the current code.

cabelo:
> Hi @cabelo, thanks for your contribution! Can you not change the `CMakeLists.txt` except if you find an issue with the current one? Also, keep the split .h/.cpp. Running the `sync_vendor.py` on the same version should not change the current code.

@angt Thank You!

PR #19769, opened 2026-02-20
ggml : add NVFP4 quantization type support

richarddd:
I'm not super experienced with the ggml/gguf internals so feedback is very welcome. Note on AI usage: Claude Opus 4.6 was used for navigating the codebase, debugging, and writing parts of the code. All changes have been reviewed and tested manually. Open to reworking anything that doesn't meet the project's standards.

This adds support for NVIDIA's NVFP4 quantization format (FP4 E2M1 weights, UE4M3 per-block scale, 16 elements per block). This is the format produced by NVIDIA ModelOpt's NVFP4 algo. The main difference is the scale encoding (UE4M3 vs E8M0).

What's in here:

- New GGML_TYPE_NVFP4 type, block struct, UE4M3 conversion helpers, reference quantize/dequantize
- convert_hf_to_gguf.py detects NVFP4 ModelOpt models and repacks into the GGUF block format
- CPU backend: scalar dot product + ARM NEON
- gguf-py: type constant, quant/dequant, endian conversion
- Tests added to test-backend-ops and test-quantize-fns

Tested with models from https://huggingface.co/NVFP4 Apple M5 MacBook (CPU, NEON) Ran llama-bench and a basic server smoke test. Would appreciate help with that if someone has a good baseline to compare against.

Here is a [Qwen3-4B](https://huggingface.co/richarddavison/Qwen3-4B-NVFP4-GGUF) model to test with.

PR #20060, opened 2026-03-03
ggml : add NVFP4 quantization type support for metal

richarddd:
@ggerganov  on top of https://github.com/ggml-org/llama.cpp/pull/19769 to provide proof of implementation. Here are some numbers for `Polaris-4B-Preview-FP4` on M5 Pro

| | CPU | Metal | Speedup |
|--------|-----|-------|---------|
| Prompt (tok/s) | 13.7 | 292.5 | 21.4x |
| Generation (tok/s) | 9.1 | 46.2 | 5.1x |


michaelw9999:
I also have a full NVFP4 implementation been working on for a long time for CUDA but haven't shared it yet, lots of testing and V&V and polishing etc.  I'll have to post that soon and hopefully we can merge the best of both!

richarddd:
> I also have a full NVFP4 implementation been working on for a long time for CUDA but haven't shared it yet, lots of testing and V&V and polishing etc. I'll have to post that soon and hopefully we can merge the best of both!

Fantastic! Can you add the CUDA implementation on top of https://github.com/ggml-org/llama.cpp/pull/19769?

ggerganov:
Thanks. For now this is OK - we can improve this a bit after the base implementation is merged.

PR #20062, opened 2026-03-03
llama: parallel model loading across GPU contexts

mxxm-t:
Speeds up model load when multiple backend contexts are used (e.g. multi-GPU split) by loading contexts in parallel.

Add -pl/--parallel-load N (env: LLAMA_ARG_PARALLEL_LOAD) to cap concurrency (-1 = unlimited, 1 = sequential, default = 4)
Make model loader thread-safe for concurrent loads (per-thread file handles + guarded shared cleanup)


abc-nix:
Have you tried `--direct-io` for model loading? How does it compare?

mxxm-t:
> Have you tried `--direct-io` for model loading? How does it compare?

I’ve tried --direct-io/-dio, yes. This PR doesn’t change the loading logic or I/O path selection (mmap vs read, direct I/O vs buffered I/O). It only parallelizes the existing loading routine across multiple backend contexts (e.g. multiple GPUs) with a configurable concurrency limit (--parallel-load).

bchtrue:
For we it's currently top 5 feature that I'm waiting for and already tried this code from other PR (working for me!!! it's really fast!!!) Thank you @mxxm-t for so quick re-submitting!

**UPDATE:** for me this current PR also working after testing.
Amazing, thank you again @mxxm-t

mxxm-t:
> For we it's currently top 5 feature that I'm waiting for and already tried this code from other PR (working for me!!! it's really fast!!!) Thank you @mxxm-t for so quick re-submitting!
> 
> **UPDATE:** for me this current PR also working after testing. Amazing, thank you again @mxxm-t

Thank you for feedback.

bchtrue:
@mxxm-t This branch has conflicts that must be resolved conflicts :-(

bchtrue:
@ggerganov Please add this it was working PR with great speed improve

Mikec78660:
This works great! Thank you for this. I have an inference server with 6 24GB GPUs but they are limited to Gen1x1. So with this my effective load speed went from 250MB/s to 1.5GB/s.

bchtrue:
@CISC please check this really work

Mikec78660:
> @CISC please check this really work

It really works.

src/llama-model.cpp
```diff
@@ -7844,10 +7847,60 @@ bool llama_model::load_tensors(llama_model_loader & ml) {
     }
 
     // load tensor data
-    for (auto & [ctx, buf_map] : ctx_buf_maps) {
-        if (!ml.load_all_data(ctx, buf_map, use_mlock ? &pimpl->mlock_mmaps : NULL, params.progress_callback, params.progress_callback_user_data)) {
+    const char * limit_env = getenv("LLAMA_ARG_PARALLEL_LOAD");
+    const size_t default_limit = 4;
```
ggerganov: Shouldn't this be a function of the number of distinct GPU devices used in the current context?

ORippler:
>> Have you tried --direct-io for model loading? How does it compare?
> I’ve tried --direct-io/-dio, yes. This PR doesn’t change the loading logic or I/O path selection (mmap vs read, direct I/O vs buffered I/O). It only parallelizes the existing loading routine across multiple backend contexts (e.g. multiple GPUs) with a configurable concurrency limit (--parallel-load).

Wondering about potential implications on Speed/SRAM consumption for DirectIO/mmap; ~According to my crude understanding, we would wish do one load into shared staging buffer & scatter to all GPUs in the multi-GPU setting~ for the Directio path, as we manage to bring SSDs to full throughput already with a single worker (@JTischbein pls comment if you have the time).

Also wondering if/how this will require rework for #19378

Edit: scatter from shared staging buffer to each context would work only if each context holds all weights, which I'm not sure is the case for pipeline-parallel

src/llama-model.cpp
mxxm-t: I don't see any draw backs on setting it to number of GPUs used, 4 was just minimum that could saturate my NVMe drive. Also maybe the flag (-pl/--parallel-load) should be boolean type instead.

mxxm-t:
> > > Have you tried --direct-io for model loading? How does it compare?
> > > I’ve tried --direct-io/-dio, yes. This PR doesn’t change the loading logic or I/O path selection (mmap vs read, direct I/O vs buffered I/O). It only parallelizes the existing loading routine across multiple backend contexts (e.g. multiple GPUs) with a configurable concurrency limit (--parallel-load).
> 
> Wondering about potential implications on Speed/SRAM consumption for DirectIO/mmap; ~According to my crude understanding, we would wish do one load into shared staging buffer & scatter to all GPUs in the multi-GPU setting~ for the Directio path, as we manage to bring SSDs to full throughput already with a single worker (@JTischbein pls comment if you have the time).
> 
> Also wondering if/how this will require rework for #19378
> 
> Edit: scatter from shared staging buffer to each context would work only if each context holds all weights, which I'm not sure is the case for pipeline-parallel

From my testing it still gives a performance boost when used with direct IO and also works with mmap.
The more GPUs you have, the more you benefit from it - while one GPU is loading, the others can keep loading instead of sitting idle waiting for their turn.
Where it really matters is when you have multiple GPUs on a slow connection as mentioned above. That was the reason I built it in the first place. With 10x MI50s connected via PCIe 3x1, it cut my MinMax-2.0 q8 load time from 18+ minutes down to 4 minutes 50 seconds. Now I have them connected via PCI4x8 and I see 15-20% faster load time.

SPAM was reason why I added that configurable flag where you can set the number of threads for loading but in reality that does not have real impact.

About the staging buffer - there are two cases: with layer split, mmap kind of does that already but poorly in my testing (the FS and OS probably handle it better), and the second case is the one-to-many upload which is what tensor split does.

bchtrue:
Any version is ok with hardcoded param or not. Merge asap. With this it's like car vs bicycle

Mikec78660:
My vote, setting -pl or --parallel-load  should default to number of devices you have defined or all devices and --parallel-load N would obviously restrict it to N at a time. I mean if you have 3,4,10 gpus where it will matter you will figure it out.

mxxm-t:
Changed default to number of  GPUs used
Added flags to llama-bench
Resolved conflcts

mxxm-t:
@ggerganov any other concerns?

common/common.h
CISC, line 1: Bad merge.

ggerganov:
I can't commit atm to maintain these changes - at the very least I rarely use multi-GPU systems and it's difficult to verify and test for me.

The things that are not good about the change:

- Introduces `std::thread` (and related) to `llama.cpp`. This is not great because we will have to special-case for emscripten builds which don't support `std::thread`
- There are a lot of combination of loading flags (mmap, dio, parallel, etc.) that is difficult to reason if they work correctly or not without tests
- Calling `setenv` is not great

These are not dealbreakers, but I would look to avoid these in some way or at least make sure we have somebody to cater. Maybe the ideal version is to do a refactoring of the `llama_model_loader` and make the logic a bit easier to follow.

mxxm-t:
All valid points. I will close this for now.

stiff:
> * Introduces `std::thread` (and related) to `llama.cpp`. This is not great because we will have to special-case for emscripten builds which don't support `std::thread`

We can add compile time flag to bring std::thread only when explicitly enabled.

> * There are a lot of combination of loading flags (mmap, dio, parallel, etc.) that is difficult to reason if they work correctly or not without tests

Now that load-mode is cleaned without independent mmap/dio can just add --load-mode=parallel ?

PR #20064, opened 2026-03-03
cmake: fix ARM feature detection hang on platforms without SVE/SME

mbucko:
Add a compile-time pre-check before the runtime feature test in check_arm_feature. Before running a test binary with -mcpu=native+<tag>, first check if -mcpu=native (without the forced +tag) defines the compiler's __ARM_FEATURE_* macro.

If the compiler doesn't think the native CPU has the feature, skip the runtime test entirely. This prevents hangs on platforms where the compiler can compile the feature code (via forced +tag) but the binary traps at an OS level without delivering SIGILL to userspace — as happens on macOS/Apple Silicon with SVE instructions.

When the compiler does report native support, the runtime test still runs as before to verify actual hardware capability.

### DIFF
--- a/ggml/src/ggml-cpu/CMakeLists.txt
+++ b/ggml/src/ggml-cpu/CMakeLists.txt
@@ -148,12 +148,28 @@ function(ggml_add_cpu_backend_variant_impl tag_name)
                     message(STATUS "ARM detected flags: ${ARM_NATIVE_FLAG}")
                 endif()
 
+                include(CheckCXXSourceCompiles)
                 include(CheckCXXSourceRuns)
 
                 macro(check_arm_feature tag feature code)
                     set(CMAKE_REQUIRED_FLAGS_SAVE ${CMAKE_REQUIRED_FLAGS})
-                    set(CMAKE_REQUIRED_FLAGS "${ARM_NATIVE_FLAG}+${tag}")
-                    check_cxx_source_runs("${code}" GGML_MACHINE_SUPPORTS_${tag})
+                    # check if the feature is available natively before running the
+                    # test binary — avoids hangs when the runtime test traps (e.g. SVE on macOS)
+                    set(CMAKE_REQUIRED_FLAGS "${ARM_NATIVE_FLAG}")
+                    check_cxx_source_compiles(
+                        "
+                        #if !defined(__ARM_FEATURE_${feature})
+                        #  error \"${feature} not supported\"
+                        #endif
+                        int main() { return 0; }
+                        "
+                        GGML_COMPILER_SUPPORTS_${tag}
+                    )
+                    if (GGML_COMPILER_SUPPORTS_${tag})
+                        # feature is supported, override CMAKE_REQUIRED_FLAGS to include +tag
+                        set(CMAKE_REQUIRED_FLAGS "${ARM_NATIVE_FLAG}+${tag}")
+                        check_cxx_source_runs("${code}" GGML_MACHINE_SUPPORTS_${tag})
+                    endif()
                     if (GGML_MACHINE_SUPPORTS_${tag})
                         set(ARM_NATIVE_FLAG_FIX "${ARM_NATIVE_FLAG_FIX}+${tag}")
                     else()


mbucko:
@angt @slaren could you review?

angt:
Hi @mbucko!

I think we need to rethink all of this, but first, can you run the following code on your setup?
```
curl -O https://github.com/angt/featcode/releases/latest/download/aarch64-macos-featcode
chmod +x aarch64-macos-featcode
./aarch64-macos-featcode .
```
and tell me if it hangs. If not, can you give us the output?
Thanks a lot!

mbucko:
> Hi @mbucko!
> 
> I think we need to rethink all of this, but first, can you run the following code on your setup?
> 
> ```
> curl -O https://github.com/angt/featcode/releases/latest/download/aarch64-macos-featcode
> chmod +x aarch64-macos-featcode
> ./aarch64-macos-featcode .
> ```
> 
> and tell me if it hangs. If not, can you give us the output? Thanks a lot!

Doesn't hang. Running on M3 Max:
```
./aarch64-macos-featcode .                                                                                                                                                                                                                                                                  
fp16
dotprod
i8mm
```

angt:
Thanks! I don’t think this fixes the root cause tho. Could you share your cmake output?

In your code, setting `CMAKE_REQUIRED_FLAGS` to `${ARM_NATIVE_FLAG}` without appending `+${tag}` means you’re no longer explicitly enabling the feature for the test compilation. Instead, you’re only checking if `-mcpu=native` enables it. So the runtime check will only run if the compiler natively detects the feature, which defeats the purpose of the original function.

mbucko:
> Thanks! I don’t think this fixes the root cause tho. Could you share your cmake output?
> 
> In your code, setting `CMAKE_REQUIRED_FLAGS` to `${ARM_NATIVE_FLAG}` without appending `+${tag}` means you’re no longer explicitly enabling the feature for the test compilation. Instead, you’re only checking if `-mcpu=native` enables it. So the runtime check will only run if the compiler natively detects the feature, which defeats the purpose of the original function.

The issue is that currently the code hangs on my mac (M3 Max) when checking for SVE. M3 doesnt support SVE or SME (SME/SVE only arrived with M4).

I did some digging and here is what actually happens: the compiler emits SVE instructions (addvl) in the function prologue before any user code runs. Since M3 has no SVE support at all, the CPU hits an unknown instruction and the process gets trapped and never terminates. SIGILL never gets delivered, and a signal handler cant help because the trap happens before main() even executes. Since the process never terminates, cmake waits forever.

What would you suggest as a better approach to handle this?

Also, is your concern the case when the compiler doesnt detect a feature with mcpu=native but the hardware actually supports it?

angt:
Yes, I also have M3, and it doesn't hang. I was thinking of some security configuration maybe, but the `featcode` binary runs the SVE instruction too and disables the feature only if it catches the `SIGILL` so the exception is working on your setup.

I have the following:
```
-- The C compiler identification is AppleClang 17.0.0.17000604
-- The CXX compiler identification is AppleClang 17.0.0.17000604
```
and:
```
$ cmake --version
cmake version 4.2.0
```

And you can test directly:
```
gcc -mcpu=native+sve -xc - <<EOF && ./a.out
#include <arm_sve.h>
int main()  { svfloat32_t _a, _b; volatile svfloat32_t _c = svadd_f32_z(svptrue_b8(), _a, _b); return 0; }
EOF
```
you should get:
```
Illegal instruction: 4
```

> Also, is your concern the case when the compiler doesnt detect a feature with mcpu=native but the hardware actually supports it?

Yes the code was originally made for this use case.

mbucko:
> Yes, I also have M3, and it doesn't hang. I was thinking of some security configuration maybe, but the `featcode` binary runs the SVE instruction too and disables the feature only if it catches the `SIGILL` so the exception is working on your setup.
> 
> I have the following:
> 
> ```
> -- The C compiler identification is AppleClang 17.0.0.17000604
> -- The CXX compiler identification is AppleClang 17.0.0.17000604
> ```
> 
> and:
> 
> ```
> $ cmake --version
> cmake version 4.2.0
> ```
> 
> And you can test directly:
> 
> ```
> gcc -mcpu=native+sve -xc - <<EOF && ./a.out
> #include <arm_sve.h>
> int main()  { svfloat32_t _a, _b; volatile svfloat32_t _c = svadd_f32_z(svptrue_b8(), _a, _b); return 0; }
> EOF
> ```
> 
> you should get:
> 
> ```
> Illegal instruction: 4
> ```
> 
> > Also, is your concern the case when the compiler doesnt detect a feature with mcpu=native but the hardware actually supports it?
> 
> Yes the code was originally made for this use case.

Interesting, so I think could be a macOS 26 (Tahoe) issue. Im on:

ProductName: macOS
ProductVersion: 26.3
BuildVersion: 25D125

I tried your exact test with both system gcc (AppleClang) and Homebrew clang 20.1.7, both hang. The process never terminates and SIGILL is never delivered. I cant even kill it with Ctrl+C.

My compiler and cmake versions:
Homebrew clang version 20.1.7
cmake version 4.0.3

The process ends up in an uninterruptible state, even kill -9 cant terminate it.

mbucko:
> > Yes, I also have M3, and it doesn't hang. I was thinking of some security configuration maybe, but the `featcode` binary runs the SVE instruction too and disables the feature only if it catches the `SIGILL` so the exception is working on your setup.
> > I have the following:
> > ```
> > -- The C compiler identification is AppleClang 17.0.0.17000604
> > -- The CXX compiler identification is AppleClang 17.0.0.17000604
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and:
> > ```
> > $ cmake --version
> > cmake version 4.2.0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And you can test directly:
> > ```
> > gcc -mcpu=native+sve -xc - <<EOF && ./a.out
> > #include <arm_sve.h>
> > int main()  { svfloat32_t _a, _b; volatile svfloat32_t _c = svadd_f32_z(svptrue_b8(), _a, _b); return 0; }
> > EOF
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > you should get:
> > ```
> > Illegal instruction: 4
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > Also, is your concern the case when the compiler doesnt detect a feature with mcpu=native but the hardware actually supports it?
> > 
> > 
> > Yes the code was originally made for this use case.
> 
> Interesting, so I think could be a macOS 26 (Tahoe) issue. Im on:
> 
> ProductName: macOS ProductVersion: 26.3 BuildVersion: 25D125
> 
> I tried your exact test with both system gcc (AppleClang) and Homebrew clang 20.1.7, both hang. The process never terminates and SIGILL is never delivered. I cant even kill it with Ctrl+C.
> 
> My compiler and cmake versions: Homebrew clang version 20.1.7 cmake version 4.0.3
> 
> The process ends up in an uninterruptible state, even kill -9 cant terminate it.

I tried your exact test with both system gcc (AppleClang) and Homebrew clang 20.1.7, both hang. The process never terminates and I cant even kill it with Ctrl+C or kill -9. The process ends up in an uninterruptible state.

I did some more digging and heres what I found:

1. Disassembled the binary and confirmed the compiler emits `addvl` (SVE) in the function prologue before any user code runs

2. Used macOS `sample` tool on the hung process, all 882/882 samples are stuck on that `addvl` instruction at main+8

3. Launched a fresh binary under lldb and single stepped to the `addvl` instruction. lldb reports `EXC_BAD_INSTRUCTION (code=1, subcode=0x43f57bf)` so the Mach exception IS raised

4. Its not SVE specific, a generic undefined instruction (`.word 0x00000000`) also hangs the same way

5. Wrote a test program that registers a Mach exception port for EXC_BAD_INSTRUCTION on itself, spawns a listener thread, then triggers the undefined instruction. The exception arrives correctly at the port. So the Mach exception mechanism works fine.

6. But when no exception port is registered (which is the normal case), the process just freezes. The fallback from Mach exception to Unix signal (SIGILL) doesnt seem to work on my machine.

7. Diffed the open source XNU code between macOS 15 (xnu-11417.140.69) and macOS 26 (xnu-12377.81.4). The exception handler (sleh.c), exception triage (exception.c), signal conversion (ux_exception.c), and signal delivery (kern_sig.c, unix_signal.c) are all functionally identical between versions.

What macOS version are you on? That would help confirm if this is version specific.

As for a fix that doesnt change the detection behavior, what do you think about replacing `check_cxx_source_runs` with `try_compile` + `execute_process(TIMEOUT 5)`? The binary still gets compiled with +tag and executed, but if it hangs for whatever reason cmake kills it after 5 seconds and moves on. Something like:

```cmake
file(WRITE "${_src}" "${code}")
try_compile(GGML_COMPILES_${tag} ${CMAKE_BINARY_DIR}/CMakeTmp
    SOURCES "${_src}" COPY_FILE "${_bin}"
    CMAKE_FLAGS "-DCMAKE_CXX_FLAGS=${ARM_NATIVE_FLAG}+${tag}")
set(GGML_MACHINE_SUPPORTS_${tag} FALSE)
if (GGML_COMPILES_${tag})
    execute_process(COMMAND "${_bin}" RESULT_VARIABLE _r TIMEOUT 5)
    if (_r EQUAL 0)
        set(GGML_MACHINE_SUPPORTS_${tag} TRUE)
    endif()
endif()
```

angt:
Thanks a lot this is definitely weird 🤔
I have macOS 26.3 (25D125) too

The current intrinsics for SVE look good, but can you try with this branch:
```
git clone https://github.com/angt/llama.cpp -b ggml-simplify-sve-detection llama.cpp-test-sve
cd llama.cpp-test-sve
cmake -B build
```
I got the following:
```
-- Performing Test GGML_MACHINE_SUPPORTS_dotprod
-- Performing Test GGML_MACHINE_SUPPORTS_dotprod - Success
-- Performing Test GGML_MACHINE_SUPPORTS_i8mm
-- Performing Test GGML_MACHINE_SUPPORTS_i8mm - Success
-- Performing Test GGML_MACHINE_SUPPORTS_sve
-- Performing Test GGML_MACHINE_SUPPORTS_sve - Failed
-- Performing Test GGML_MACHINE_SUPPORTS_nosve
-- Performing Test GGML_MACHINE_SUPPORTS_nosve - Success
-- Performing Test GGML_MACHINE_SUPPORTS_sme
-- Performing Test GGML_MACHINE_SUPPORTS_sme - Failed
-- Performing Test GGML_MACHINE_SUPPORTS_nosme
-- Performing Test GGML_MACHINE_SUPPORTS_nosme - Success
-- Checking for ARM features using flags:
--   -U__ARM_FEATURE_SVE
--   -U__ARM_FEATURE_SME
--   -mcpu=native+dotprod+i8mm+nosve+nosme
-- Performing Test HAVE_DOTPROD
-- Performing Test HAVE_DOTPROD - Success
-- Performing Test HAVE_SVE
-- Performing Test HAVE_SVE - Failed
-- Performing Test HAVE_MATMUL_INT8
-- Performing Test HAVE_MATMUL_INT8 - Success
-- Performing Test HAVE_FMA
-- Performing Test HAVE_FMA - Success
-- Performing Test HAVE_FP16_VECTOR_ARITHMETIC
-- Performing Test HAVE_FP16_VECTOR_ARITHMETIC - Success
-- Performing Test HAVE_SME
-- Performing Test HAVE_SME - Failed
-- Adding CPU backend variant ggml-cpu: -U__ARM_FEATURE_SVE;-U__ARM_FEATURE_SME;-mcpu=native+dotprod+i8mm+nosve+nosme
```

mbucko:
> Thanks a lot this is definitely weird 🤔 I have macOS 26.3 (25D125) too
> 
> The current intrinsics for SVE look good, but can you try with this branch:
> 
> ```
> git clone https://github.com/angt/llama.cpp -b ggml-simplify-sve-detection llama.cpp-test-sve
> cd llama.cpp-test-sve
> cmake -B build
> ```
> 
> I got the following:
> 
> ```
> -- Performing Test GGML_MACHINE_SUPPORTS_dotprod
> -- Performing Test GGML_MACHINE_SUPPORTS_dotprod - Success
> -- Performing Test GGML_MACHINE_SUPPORTS_i8mm
> -- Performing Test GGML_MACHINE_SUPPORTS_i8mm - Success
> -- Performing Test GGML_MACHINE_SUPPORTS_sve
> -- Performing Test GGML_MACHINE_SUPPORTS_sve - Failed
> -- Performing Test GGML_MACHINE_SUPPORTS_nosve
> -- Performing Test GGML_MACHINE_SUPPORTS_nosve - Success
> -- Performing Test GGML_MACHINE_SUPPORTS_sme
> -- Performing Test GGML_MACHINE_SUPPORTS_sme - Failed
> -- Performing Test GGML_MACHINE_SUPPORTS_nosme
> -- Performing Test GGML_MACHINE_SUPPORTS_nosme - Success
> -- Checking for ARM features using flags:
> --   -U__ARM_FEATURE_SVE
> --   -U__ARM_FEATURE_SME
> --   -mcpu=native+dotprod+i8mm+nosve+nosme
> -- Performing Test HAVE_DOTPROD
> -- Performing Test HAVE_DOTPROD - Success
> -- Performing Test HAVE_SVE
> -- Performing Test HAVE_SVE - Failed
> -- Performing Test HAVE_MATMUL_INT8
> -- Performing Test HAVE_MATMUL_INT8 - Success
> -- Performing Test HAVE_FMA
> -- Performing Test HAVE_FMA - Success
> -- Performing Test HAVE_FP16_VECTOR_ARITHMETIC
> -- Performing Test HAVE_FP16_VECTOR_ARITHMETIC - Success
> -- Performing Test HAVE_SME
> -- Performing Test HAVE_SME - Failed
> -- Adding CPU backend variant ggml-cpu: -U__ARM_FEATURE_SVE;-U__ARM_FEATURE_SME;-mcpu=native+dotprod+i8mm+nosve+nosme
> ```

tried your branch, same problem on my machine. cmake hangs on the SVE test. I disassembled the binary and your changes do work correctly, `main()` has no SVE and `p()` just has a single `cntw` instruction. but on my machine it still freezes on `cntw` which is also an SVE instruction.

might just be something weird with my setup at this point.

PR #20067, opened 2026-03-03
cli: Don't clear system prompt when using '/clear'

roj234:
Add system prompt to messages when clearing chat history.

### DIFF
--- a/tools/cli/cli.cpp
+++ b/tools/cli/cli.cpp
@@ -262,12 +262,15 @@ int main(int argc, char ** argv) {
         modalities += ", audio";
     }
 
-    if (!params.system_prompt.empty()) {
-        ctx_cli.messages.push_back({
-            {"role",    "system"},
-            {"content", params.system_prompt}
-        });
-    }
+    auto add_system_prompt = [&]() {
+        if (!params.system_prompt.empty()) {
+            ctx_cli.messages.push_back({
+                {"role",    "system"},
+                {"content", params.system_prompt}
+            });
+        }
+    };
+    add_system_prompt();
 
     console::log("\n");
     console::log("%s\n", LLAMA_ASCII_LOGO);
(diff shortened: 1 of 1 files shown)


roj234:
Currently system prompt is specified by command line argument, so /clear remove it is pretty weird

tools/cli/cli.cpp
```diff
@@ -357,6 +357,14 @@ int main(int argc, char ** argv) {
             }
         } else if (string_starts_with(buffer, "/clear")) {
             ctx_cli.messages.clear();
+
```
danbev: Perhaps we can create a lambda for this and not have to duplicate this code:
```console
diff --git a/tools/cli/cli.cpp b/tools/cli/cli.cpp
index 334ba8492..2e979a641 100644
--- a/tools/cli/cli.cpp
+++ b/tools/cli/cli.cpp
@@ -262,12 +262,15 @@ int main(int argc, char ** argv) {
         modalities += ", audio";
     }
 
-    if (!params.system_prompt.empty()) {
-        ctx_cli.messages.push_back({
-            {"role",    "system"},
-            {"content", params.system_prompt}
-        });
-    }
+    const auto add_system_prompt = [&params, &ctx_cli] () {
+        if (!params.system_prompt.empty()) {
+            ctx_cli.messages.push_back({
+                {"role",    "system"},
+                {"content", params.system_prompt}
+            });
+        }
+    };
+    add_system_prompt();
 
     console::log("\n");
     console::log("%s\n", LLAMA_ASCII_LOGO);
@@ -358,12 +361,7 @@ int main(int argc, char ** argv) {
         } else if (string_starts_with(buffer, "/clear")) {
             ctx_cli.messages.clear();
 
-            if (!params.system_prompt.empty()) {
-                ctx_cli.messages.push_back({
-                    {"role",    "system"},
-                    {"content", params.system_prompt}
-                });
-            }
+            add_system_prompt();
 
             ctx_cli.input_files.clear();
             console::log("Chat history cleared.\n");
```

danbev (approved):

PR #20068, opened 2026-03-03
ggml-webgpu: Add the support of `GGML_OP_CONCAT`

yomaytk:
Hello. This PR adds `GGML_OP_CONCAT` support to the WebGPU backend. This op is used by models such as DeepSeek-V2. 
This change supports two types `F32`, `I32` to match the types covered by `test_concat` in `test-backend-ops`.


reeselevine (approved):
looks great, thanks!

PR #11356, opened 2025-01-22
Avoid fp32->fp16->fp32 conversion on cdna in ggml_cuda_op_mul_mat_cublas

IMbackK:
This further improves on https://github.com/ggerganov/llama.cpp/pull/10498 by removeing the fp32->fp16->fp32 conversion on cdna in ggml_cuda_op_mul_mat_cublas. Unlike what is stated in #10498 this actually dose improve performance, as the issue fixed by https://github.com/ggerganov/llama.cpp/pull/11244 was simply hiding the change. the issue fixed by https://github.com/ggerganov/llama.cpp/pull/11244 was also hideing a pessimisation in #10498 which this pr also reverts.

Master:

```
  Device 0: AMD Instinct MI100, compute capability 9.0, VMM: no
| model                          |       size |     params | backend    | ngl | n_batch |          test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | ------: | ------------: | -------------------: |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |      64 |          pp64 |        434.23 ± 0.43 |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |      64 |         pp128 |        433.79 ± 0.39 |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |      64 |         pp512 |        430.21 ± 0.28 |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |     512 |          pp64 |        433.94 ± 0.20 |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |     512 |         pp128 |        785.42 ± 0.49 |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |     512 |         pp512 |       1108.48 ± 0.63 |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |    1024 |          pp64 |        433.87 ± 0.41 |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |    1024 |         pp128 |        783.96 ± 1.65 |
| llama 8B Q4_K - Medium         |   4.58 GiB |     8.03 B | ROCm       |  99 |    1024 |         pp512 |       1106.24 ± 0.35 |
[...]

PR #20078, opened 2026-03-03
CUDA: Add BF16 path to CUBLAS and increase precision of FP16 path

ORippler:
BF16 has better numerical stability than FP16 and is thus the preferred path if available. Since we require the output to be FP32, we avoid loss of precision by skipping FP32 -> BF16 -> FP32 round-trip. This was also a bit more performant on my GPU:

```
➜  llama.cpp git:(master) ✗ ./build-x64-linux-gcc-reldbg/bin/llama-bench -m /mnt/share/gguf/Nemotron-Nano-3-30B-A3B-Q8_0.gguf -fa 1 -ub 2048 -b 8192 -d 8000 -dio 1 -p 8096 -n 0
  Device 0: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition, compute capability 12.0, VMM: yes
| model                          |       size |     params | backend    | ngl | n_batch | n_ubatch | fa | dio |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | ------: | -------: | -: | --: | --------------: | -------------------: |
| nemotron_h_moe 31B.A3.5B Q8_0  |  31.27 GiB |    31.58 B | CUDA       |  99 |    8192 |     2048 |  1 |   1 |  pp8096 @ d8000 |       6853.30 ± 6.78 |

build: https://github.com/ggml-org/llama.cpp/commit/ec88c3ceeaaa037dc8a80413c566ed6d71d9e85c (8186)

FP32 -> BF16 -> FP32 path
➜  llama.cpp git:(master) ✗ ./build-x64-linux-gcc-reldbg/bin/llama-bench -m /mnt/share/gguf/Nemotron-Nano-3-30B-A3B-Q8_0.gguf -fa 1 -ub 2048 -b 8192 -d 8000 -dio 1 -p 8096 -n 0
  Device 0: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition, compute capability 12.0, VMM: yes
| model                          |       size |     params | backend    | ngl | n_batch | n_ubatch | fa | dio |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | ------: | -------: | -: | --: | --------------: | -------------------: |
| nemotron_h_moe 31B.A3.5B Q8_0  |  31.27 GiB |    31.58 B | CUDA       |  99 |    8192 |     2048 |  1 |   1 |  pp8096 @ d8000 |      6467.15 ± 18.85 |

build: https://github.com/ggml-org/llama.cpp/commit/ec88c3ceeaaa037dc8a80413c566ed6d71d9e85c (8186)
```

The avoidance of FP16 -> FP32 upcast was also applied to the FP16-path for increased stability and perf (should affect only pre-Ampere GPUs). This was added already in #11356 for the ROCm path. This PR should fix numerical instabilities observed in #19659, though only for Ampere+ GPUs. I did not evaluate potential impact for pre-VOLTA GPUs that do not possess tensor-cores.

### DIFF
--- a/ggml/src/ggml-cuda/convert.cu
+++ b/ggml/src/ggml-cuda/convert.cu
@@ -81,6 +81,52 @@ static __global__ void dequantize_block_q8_0_f16(const void * __restrict__ vx, h
 #endif // __CUDA_ARCH__ >= GGML_CUDA_CC_PASCAL
 }
 
+template <bool need_check>
+static __global__ void dequantize_block_q8_0_bf16(const void * __restrict__ vx,
+                                                  nv_bfloat16 * __restrict__ y,
+                                                  const int64_t k) {
+#if __CUDA_ARCH__ >= GGML_CUDA_CC_AMPERE
+    constexpr int nint = CUDA_Q8_0_NE_ALIGN / sizeof(int) + WARP_SIZE;
+
+    const int64_t    i0 = CUDA_Q8_0_NE_ALIGN * blockIdx.x;
+    const int *      x0 = ((int *) vx) + blockIdx.x * nint;
+    __nv_bfloat162 * y2 = (__nv_bfloat162 *) (y + i0);
+
+    __shared__ int vals[nint];
+
+#    pragma unroll
+    for (int ix0 = 0; ix0 < nint; ix0 += WARP_SIZE) {
+        if (need_check &&
+            i0 * sizeof(block_q8_0) / QK8_0 + sizeof(int) * (ix0 + threadIdx.x) >= k * sizeof(block_q8_0) / QK8_0) {
+            break;
+        }
+
+        const int ix = ix0 + threadIdx.x;
+        vals[ix]     = x0[ix];
+    }
+
+    __syncthreads();
+
+#    pragma unroll
+    for (int iy = 0; iy < CUDA_Q8_0_NE_ALIGN; iy += 2 * WARP_SIZE) {
+        if (need_check && i0 + iy + 2 * threadIdx.x >= k) {
+            return;
+        }
+
+        const half * b0 =
+            ((const half *) vals) + (sizeof(block_q8_0) / sizeof(half)) * ((iy + 2 * threadIdx.x) / QK8_0);
+        const half        d     = *b0;
+        const nv_bfloat16 d_b16 = nv_bfloat16(d);
+        const char2       qs    = ((const char2 *) (b0 + 1))[threadIdx.x % (QK8_0 / 2)];
+
+        y2[iy / 2 + threadIdx.x] = __hmul2(make_bfloat162(qs.x, qs.y), __bfloat162bfloat162(d_b16));
+    }
+#else
+    GGML_UNUSED_VARS(vx, y, k);
+    NO_DEVICE_CODE;
+#endif  // __CUDA_ARCH__ >= GGML_CUDA_CC_AMPERE
+}
+
 template<typename dst_t>
 static __global__ void dequantize_block_q4_0(const void * __restrict__ vx, dst_t * __restrict__ yy, int nb32) {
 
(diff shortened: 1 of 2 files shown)


JohannesGaessler:
Did you also check the performance of these changes. Previously I decided against using BF16 because it was slow even on GPUs that should have BF16 tensor cores (though I can't rule out that I was just using cuBLAS wrong).

JohannesGaessler:
More generally FYI: the usage of cuBLAS is currently rather convoluted as an artifact of how we had written the code before we implemented ggml backends. Once we have `--split-mode tensor` I intend to remove `--split-mode row` and a lot of this legacy code.

ORippler:
> Did you also check the performance of these changes. Previously I decided against using BF16 because it was slow even on GPUs that should have BF16 tensor cores (though I can't rule out that I was just using cuBLAS wrong).

I checked it to be perf positive on B6000. In general, all Tensor Core generations that support BF16 (namely Ampere +) have equal throughput in BF16 & FP16, see [Table 3 of the BW whitepaper](https://images.nvidia.com/aem-dam/Solutions/geforce/blackwell/nvidia-rtx-blackwell-gpu-architecture.pdf) and [Table 10 of the Ampere whitepaper](https://www.nvidia.com/content/PDF/nvidia-ampere-ga-102-gpu-architecture-whitepaper-v2.1.pdf). Only Volta and Turing will take the FP16 path.

Note that I did not check for perf implications of switching from FP16 to FP32 accumulation. Unfortunately, I don't have easy-access to a Geforce GPU, and RTX PRO GPUs have same Tensor Core throughput for both FP16 and FP32 accumulates. However, for numerical stability I'd recommend to always accumulate in FP32 and pay the associated perf cost over potentially over/underflowing, especially as we will output in FP32 and thus will have to upcast from FP16 to FP32.

> More generally FYI: the usage of cuBLAS is currently rather convoluted as an artifact of how we had written the code before we implemented ggml backends. Once we have --split-mode tensor I intend to remove --split-mode row and a lot of this legacy code

Yeah when I skimmed the code and saw "last edited 22 months ago" I knew what I was getting myself into. I still think libs like cuBLAS/cuDNN could be beneficial to ggml. When profiling some stable-diffusion models I saw 2x perf for cuDNN's SDPA compared to our current FA-implementation (FP16 precision). Thoughts about deepening their integration where possible?

ORippler:
Maybe @IMbackK can have a look at the hippification failures. Seems like at a first glance
1. ROCm allows precision loss when casting F32 -> F16, but not F32 -> BF16
2. make_bfloat162 was added in 12.2, not sure which version we use and if it would need to be updated.

IMbackK:
@ORippler make_bfloat162 requres rocm 7.0, unfortunately upgrading our rocm requirement above rocm 6.3 would cause issues for users of gfx803, gfx900, gfx90c and gfx906 as this is the last version amd built for these devices.

The changes would have a large negative effect on CDNA as its best matmul instruction for bf16 is V_MFMA_F32_16X16X8BF16 which has half the rate of the fp16 equivalent V_MFMA_F32_16X16X16F16
These changes would have a small postive effect on CDNA3

These changes would also have a massive negative effect on RDNA3 as you removed the possibility of CUBLAS_COMPUTE_16F being used, for RDNA3 rocblas dose not (or at least did not, have not tested since rocm 7.0) implement WMMA kernels that can do fp32 accumulation, this means that using CUBLAS_COMPUTE_32F here will eliminate the use of RDNA3's matrix acceleration, hipblaslt includes sutch kernels and we could add logic to use hipblaslt on rdna3 - but i question the cost benefit here.

overall the consequences of this pr for the HIP backend would be dire as is.

ORippler:
> overall the consequences of this pr for the HIP backend would be dire as is.

Hmm. Would you be fine with keeping old paths for ROCm, despite their numerical risks? Guess one could also add some flexibilization/customziation similar to #19959 to the dense cuBLAS path here, and I'd recommend for NVGPUs to default to  BF16&FP32 accumulation where possible

IMbackK:
Well on rdna3, we need to use the current path with fp16&fp16 or add a whole other method to dispatch to hipblaslt instead. 
CDNA1 needs to use fp16&f32 since while it has matrix acceleration for bf16&fp32 its slower. For CDNA2+ and RDNA4 we can use BF16&FP32 as proposed.

We must maintain compatibility with ROCM 6.3 at least, this means providing our own make_bfloat162 implementation, not a big deal but we must do it.

IMbackK:
Using fp32 accumulation where this is possible with little performance impact like https://github.com/ggml-org/llama.cpp/pull/19959 is worth it. Im not sure if using bf16 over fp16 for A and B gains us that much, really. Regardless it needs to by applied selectively.

PR #20084, opened 2026-03-03
vulkan: Fix data races in coopmat1 mul_mat(_id)

jeffbolznv:
Add barriers between coopmat store and regular loads. We sort of got away with this because it was the same subgroup accessing the values, but it's still a race and may not work.

I added shared memory data race detection for coopmat1 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/11780) and this fixes the issues it found. No performance regressions on my system.

### DIFF
--- a/ggml/src/ggml-vulkan/vulkan-shaders/mul_mm.comp
+++ b/ggml/src/ggml-vulkan/vulkan-shaders/mul_mm.comp
@@ -377,6 +377,7 @@ void main() {
         [[unroll]] for (uint cm_col = 0; cm_col < cms_per_col; cm_col++) {
             coopMatStore(sums[cm_col * cms_per_row + cm_row], coopmat_stage, warp_i * TM * TN, TM, gl_CooperativeMatrixLayoutColumnMajor);
 
+            barrier();
             [[unroll]] for (uint col = 0; col < TN; col += storestride) {
                 const uint row_i = dc + cm_col * TN + col + store_c;
                 if (row_i >= _ne1) break;
@@ -387,6 +388,7 @@ void main() {
                     data_d[row_idx.y * p.batch_stride_d + row_idx.x * p.stride_d + dr + cm_row * TM + store_r] = D_TYPE(coopmat_stage[warp_i * TM * TN + (col + store_c) * TM + store_r]);
                 }
             }
+            barrier();
         }
     }
 #else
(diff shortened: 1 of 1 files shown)


0cc4m:
I do see small performance regressions from this change on AMD. Considering even without barriers it seems to work, since the accesses are limited within subgroups, we don't need full barriers here. A perfect place for `subgroupMemoryBarrierShared()`? The overall differences are too small for me to be sure, but it does seem to help.

jeffbolznv:
subgroupMemoryBarrierShared wouldn't be sufficient, but `controlBarrier(gl_ScopeSubgroup, gl_ScopeSubgroup, gl_StorageSemanticsShared, gl_SemanticsAcquireRelease)` (https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_memory_scope_semantics.txt#L443) ought to be.

I'm surprised this shows up as a real perf hit since this is just in the epilogue of the shader.

0cc4m:
Why isn't it enough? It's about shared memory that's only used by the subgroup.

jeffbolznv:
subgroupMemoryBarrierShared is only an OpMemoryBarrier. Synchronization requires a release on the storing thread, an acquire on the loading thread, and an "edge" between those that can be provided either by a control barrier or an atomic store and an atomic load that sees the stored value (this is in the "Synchronizes-With" section of the memory model appendix). When using a control barrier, the release and acquire can be performed by the same control barrier (similarly, the release and acquire can optionally be folded into the corresponding atomics). A subgroup control barrier ought to be relatively cheap compared to a workgroup control barrier.

0cc4m:
Yeah, then let's use the minimal `controlBarrier`. My tests aren't fully conclusive, but it seems to make a small difference.

jeffbolznv:
OK, changed to subgroup control barriers.

0cc4m (approved):

CISC:
This seems to be causing `MUL_MAT` failures:
https://github.com/ggml-org/llama.cpp/actions/runs/22820245245/job/66191745551#step:9:32598

0cc4m:
Damn, I missed that. But I don't see these failures locally, with llvmpipe. That's odd.

jeffbolznv:
Do you have a recent version of llvmpipe that supports coopmat1?

I think we should just revert this for now. I wonder if llvmpipe has some bug in its handling of this unusual barrier instruction.

0cc4m:
Ah yeah, on Arch I can trigger it with a newer llvmpipe version. But I don't think llvmpipe coopmat makes much sense anyways, so we can also just disable coopmat in the CI.

PR #20086, opened 2026-03-03
llama : add attention weights extraction API [EXPERIMENTAL]

QuentinFuxa:
### Attention weights extraction API

Adds a minimal API to extract per-head attention weights from selected (layer, head) pairs during decoding. Enables attention-based applications (alignment, interpretability, simultaneous translation) without modifying the model or graph topology.

### API

```c
// Enable attention extraction when creating context
llama_context_params.attn_weights = true;  // requires flash_attn disabled

// Select which (layer, head) pairs to track
void llama_set_attn_heads(ctx, layers, heads, n_pairs);

// After decode, get attention weights for the i-th output token
// Returns [n_pairs * n_ctx] floats
float * llama_get_attn_ith(ctx, i);

// Number of valid KV entries per pair
int32_t llama_get_attn_n_kv(ctx);
```

### Implementation

Internally installs a `cb_eval` callback that intercepts `kq_soft_max` tensors matching the requested (layer, head) pairs. The public API handles tensor name matching, stride-aware copying, and buffer management so users don't have to.

Zero graph splits because graph topology is unchanged (2 splits, same as baseline), 0% overhead when enabled , and chains correctly with user-provided `cb_eval` callbacks

### GQA support

Works out of the box with Grouped Query Attention. the `kq_soft_max` tensor already has full `[n_head, n_kv, n_tokens]` dimensions via GGML broadcasting. No code changes needed.

<details>
<summary>Verified models (attention sums = 1.0 for all)</summary>

| Model | Architecture | GQA ratio | Layers |
|-------|-------------|-----------|--------|
| Qwen3-4B | Qwen3 | 4:1 (32q/8kv) | 36 |
| Qwen3-8B | Qwen3 | 4:1 (32q/8kv) | 36 |
| Qwen3-14B | Qwen3 | 4:1 (32q/8kv) | 40 |
| HY-MT1.5-7B | HunYuanDense | 4:1 (32q/8kv) | 32 |
| EuroLLM-9B | Llama | 4:1 (32q/8kv) | 42 |
| TowerInstruct-7B | Llama | 1:1 (32q/32kv) | 32 |
| TranslateGemma-4B | Gemma2 | 4:1 (8q/1kv) | 34 |

</details>

### KV cache compatibility

Fully compatible with llama.cpp's KV cache management (`llama_memory_seq_rm()` + incremental `llama_decode()`). Extracted weights correctly reflect the updated KV state after partial cache invalidation. **3.7–4.6x speedup** vs context re-creation in streaming scenarios.

### Use case: Simultaneous Machine Translation

We use this for the [AlignAtt policy](https://openreview.net/forum?id=q8fTgw8e5E) (Polák et al., ICLR 2026), an attention-based read/write policy for simultaneous speech translation. The system detects "translation heads" whose attention patterns align with word alignment, then uses `llama_get_attn_ith()` at each generation step to decide when to READ more source vs. WRITE output.

<details>
<summary>Benchmark results (EN→ZH, FLORES, NVIDIA A40)</summary>

| Model | Top Head TS | COMET (bd=2) | Avg Lagging |
|-------|-----------|-------------|-------------|
| HY-MT1.5-7B (7B, Q8_0) | **0.731** | **0.862** | **3.13** |
| EuroLLM-9B (9B, Q8_0) | 0.704 | 0.828 | 2.82 |
| Qwen3-4B (4B, Q8_0) | ~0.3 | 0.844 | 3.93 |
| TranslateGemma-4B (4B, f16) | 0.388 | 0.847 | 3.91 |

- **Translation Score (TS)**: how well a head's attention matches word alignments
- **COMET**: translation quality (higher = better)
- **bd=2**: border distance parameter

Also tested on EN→DE, EN→IT, CS→EN.

</details>

### Files

**Core** (minimal diff):
- `include/llama.h`: 3 new API functions
- `src/llama-context.cpp`: attention extraction via cb_eval (~150 lines)
- `src/llama-context.h`: internal state for tracked heads
- `src/llama-cparams.h`: `attn_weights` flag
- `src/llama-graph.cpp`: minor: pass head info to callback

**Example** (`examples/attn-weights/`):
- C++ example with ASCII attention heatmap
- Python ctypes bindings + test suite + benchmark scripts

### Breaking changes

None. Opt-in via `attn_weights = true`. Zero overhead when not enabled.

### Usage

```bash
# C++ example
./build/bin/llama-attn-weights -m model.gguf -l -1 -hd 0 "The cat sat on the mat"

# Python tests
python3 examples/attn-weights/test_attn_weights.py model.gguf

# Full benchmark
python3 examples/attn-weights/eval_benchmark_translategemma.py model.gguf \
    --heads examples/attn-weights/translation_heads_en_zh.json -n 100
```


ngxson:
Such public API is pretty much redundant, we already had `cb_eval` that allow reading any intermediately values, not just attention scores

QuentinFuxa:
Yes I've tried `cb_eval` before, but that's more a debug tool, that syncs after every node in the graph to fire the callback, which kills performance. Not great when you need attention at every autoregressive step at inference time

My API only marks the specific layers you request as output tensors, the rest of the graph runs untouched. And you get a clean float * per output token instead of having to match tensor names with regex and figure out the 4D layout yourself.

For instance, on TranslateGemma we only monitor attention from around 10 attention heads at every single generation step. cb_eval would make that unusably slow.

But maybe, the usecase is too specific ? thus should not be in the public API ? Other use cases include interpretability/hallucination detection (visualizing which tokens the model attends to and flagging when attention is diffuse over the context), and RAG relevance scoring without reranking. These all need attention at inference time without the overhead of stopping the graph at every node. 

Happy to discuss alternatives if you think this shouldn't be in the public API though - and if not, is there is room somewhere else for that ?

ggerganov:
> Yes I've tried cb_eval before, but that's more a debug tool, that syncs after every node in the graph to fire the callback, which kills performance. Not great when you need attention at every autoregressive step at inference time

The callback can return false for nodes that you are not interested in:

https://github.com/ggml-org/llama.cpp/blob/c99909dd0b6589070e2a129fa172c40fa1675453/ggml/include/ggml-backend.h#L295-L304

graehl:
At least for MT I often want alignments - sure you can just run an encoder similarity matrix on input, output tokens and never peek at attention, but I think attention extracts are actually a pretty likely thing for people to peek at if it's made convenient (they have *some* kind of explanatory value). I think there have been enough research papers over the years looking at signatures of attention that the use case is worth supporting first-class.
Sure, you can modify the model to make it output such things explicitly. But frankly that's tedious.
Callbacks that allow peeking at values not marked as output do kill throughput. Maybe if just enabling a callback for just the particular layer doesn't kill throughput more than 10% (presumably there's something like a cuda sync. cost?) it could be argued that the additional code isn't worth it.
To be slightly more general, a facility for appending or mean/max pooling arbitrary not-normally-output tensors would be great. For example, you can detect some pathological outputs by observing that the wrong parts of the prompt (e.g. few shot prompts) have too much attention on them in total or during output of certain tokens.

QuentinFuxa:
Thanks everyone for the feedback
@ggerganov You were right for cb_eval with ask=true
I've reworked my implementation then: internally it now installs a cb_eval that only fires on the kq_soft_max tensors matching the requested layer +head pairs. Graph splits are back to baseline (2 vs almost 70 before), no more overhead measured

The API now "just" wraps the stuff needed to work with attention now (tensor name matching etc). Happy to keep it if you think the convenience is worth it, but also totally fine if you'd rather leave this to cb_eval users. For my usecase I have what I need!

thanks again for the super project

QuentinFuxa:
Updated the PR description to reflect the current state:

- Implementation reworked per @ggerganov's suggestion: now uses an internal `cb_eval` that only fires on `kq_soft_max` tensors for requested layers.
- GQA verified across 7 models (Qwen3, HY-MT, EuroLLM, TowerInstruct, TranslateGemma), works out of the box, attention sums = 1.0
- KV cache compatibility confirmed: 3.7–4.6x speedup with incremental decoding in streaming scenarios
- Benchmarks added (from my usecase): COMET + latency on 4 models, tested on EN→ZH/DE

WilliamTambellini:
@QuentinFuxa  have you rebased recently?

Issue #18497, opened 2025-12-30
Eval bug: cache-reuse not effective in qwen3-next


IMbackK:
### Name and Version

llama-server at commit 0f89d2ecf14270f45f43c442e90ae433fd82dab1

### Operating systems

Linux

### GGML backends

HIP

### Hardware

MI100s

### Models

Qwen3-Next-80B-A3B-Instruct-Q5_K_S.gguf

### Problem description & steps to reproduce

I am not sure if this is a bug or a model incompatibility with Qwen3-Next but using Qwen3-Next kv cache reuse seams to not work.

llama-server is started like this:
```
llama-server -ngl 99 -fa 1 -m Qwen3-Next-80B-A3B-Instruct-Q5_K_S.gguf --jinja -c 64000 -np 2 -sps 0.3 --cache-reuse 128 -cram 65536 --port 8092 -ub 2048 --swa-full
```

then a request with a context of 4096 tokens is sent, kv looks like this with each letter being 512 tokens long block:
```
AAAABBBB
```

Then a new request is sent with the same content except for an extra block of 512 tokens inserted in the center:
```
AAAACBBBB
```

Result with Qwen3-Next:

Context gets fully reporccesed:
```
slot update_slots: id  3 | task 173 | forcing full prompt re-processing due to lack of cache data (likely due to SWA or hybrid/recurrent memory, see https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055)
```

Result with Qwen3:

512 tokens get processed, the rest is filled from shfited kv cache.

I also tried with `--ctx-checkpoints 128` but this did not change the behavior, im not sure how this feature is supposed to work.

pwilkin:
The problem is the recurrent state, not the KV cache. You will get checkpoints after each processed prompt, so if a checkpoint matches, content will get reprocessed from there. But the recurrent state doesn't work like the KV cache. It only keeps a single state that gets updated with each token. Therefore, the only way this could work if we added arbitrary snapshot checkpoints *during* prompt processing, for example each N tokens.

othermod:
This seems to be the issue I'm experiencing too. I send 100k tokens of context (which stays the same) + and additional 3k tokens of instruction at the end (which changes on each run), and it always rebuilds the whole thing rather than the last 3k. I switched back to using another model for the task because its taking so long to do the prompt processing on each pass.

ggerganov:
[...]

PR #19970, opened 2026-02-28
server: batch checkpoints to support kvcache context truncation

aagit:
Currently, checkpoints are only created once the entire kvcache has been computed. This renders the kvcache an append-only cache.

This change creates checkpoints at every token batch. If the context is truncated, the kvcache remains useful like with models that do not require checkpoints.

Reproducer simulating a coding workflow frequently editing files in context:

```python
import requests
import time
import json

url = "http://localhost:8811/v1/chat/completions"

repeated_line = "This is a repeated line. "
payload_long = {
    "messages": [{"role": "user", "content": repeated_line * 10000}],
    "max_tokens": 1
}

payload_short = {
    "messages": [{"role": "user", "content": repeated_line * 9000}],
    "max_tokens": 1
}

start_time = time.time()
response_long = requests.post(url, json=payload_long)
time_long = time.time() - start_time
if response_long.status_code != 200:
    raise RuntimeError(f"Request failed with status {response_long.status_code}")

start_time = time.time()
response_short = requests.post(url, json=payload_short)
time_short = time.time() - start_time
if response_short.status_code != 200:
    raise RuntimeError(f"Request failed with status {response_short.status_code}")

print(f"kvcache preload: {time_long:.4f} seconds")
print(f"truncated kvcache lookup: {time_short:.4f} seconds")
```

Results with `llama.cpp --ctx-checkpoints 128`.

Devstral-Small-2-24B-Instruct-2512 (stock):
kvcache preload: 133.6303 seconds
truncated kvcache lookup: 0.1259 seconds

Qwen3.5-35B-A3B nothink (stock):
kvcache preload: 43.1127 seconds
truncated kvcache lookup: 37.4520 seconds

Qwen3.5-35B-A3B nothink (with batch checkpoints):
kvcache preload: 43.4365 seconds
truncated kvcache lookup: 0.9374 seconds

Performance boost for razor coding rg-edit [1] workflows with context LRU management and optional preloading [2]: ~39x with ~52k context.

[1] https://gitlab.com/aarcange/ripgrep-edit
[2] https://github.com/karthink/gptel/pull/1147

PR #17428, opened 2025-11-21
[Hybrid] Create checkpoints while processing the prompt

whoreson:
Currently, llama-server creates a single checkpoint after e.g. processing 600000 tokens with IBM Granite. Which will likely get erased real soon.

I have deemed this sad state of affairs insufferable, so I instructed my girl Gemini to solve the issue, and generate half of the requested checkpoints during PP.

Now it's actually usable.

Here is the result, do the needful saars.

```
slot update_slots: id  0 | task 487 | prompt processing progress, n_tokens = 4096, batch.n_tokens = 2048, progress = 0.235375
slot update_slots: id  0 | task 487 | n_tokens = 4096, memory_seq_rm [4096, end)
slot update_slots: id  0 | task 487 | created prompt processing context checkpoint 2 of 256 (pos_min = 4095, pos_max = 4095, size = 147.481 MiB)
slot update_slots: id  0 | task 487 | prompt processing progress, n_tokens = 6144, batch.n_tokens = 2048, progress = 0.353063
slot update_slots: id  0 | task 487 | n_tokens = 6144, memory_seq_rm [6144, end)
slot update_slots: id  0 | task 487 | created prompt processing context checkpoint 3 of 256 (pos_min = 6143, pos_max = 6143, size = 147.481 MiB)
slot update_slots: id  0 | task 487 | prompt processing progress, n_tokens = 8192, batch.n_tokens = 2048, progress = 0.470750
slot update_slots: id  0 | task 487 | n_tokens = 8192, memory_seq_rm [8192, end)
slot update_slots: id  0 | task 487 | created prompt processing context checkpoint 4 of 256 (pos_min = 8191, pos_max = 8191, size = 147.481 MiB)
slot update_slots: id  0 | task 487 | prompt processing progress, n_tokens = 10240, batch.n_tokens = 2048, progress = 0.588438
slot update_slots: id  0 | task 487 | n_tokens = 10240, memory_seq_rm [10240, end)
slot update_slots: id  0 | task 487 | created prompt processing context checkpoint 5 of 256 (pos_min = 10239, pos_max = 10239, size = 147.481 MiB)
slot update_slots: id  0 | task 487 | prompt processing progress, n_tokens = 12288, batch.n_tokens = 2048, progress = 0.706126
slot update_slots: id  0 | task 487 | n_tokens = 12288, memory_seq_rm [12288, end)
[...]

PR #13194, opened 2025-04-29
kv-cache : add SWA support

ggerganov:
### Overview

Add `class llama_kv_cache_unified_iswa` for interleaved SWA attention support.

The implementation internally utilizes 2 instances of the existing `llama_kv_cache_unified` - one for the non-SWA and one for the SWA layers of the model. To achieve that, the `llama_kv_cache_unified` implementation is updated to be able to cache a subset of the model's layers (instead of always caching all layers as it is on `master`). The 2 internal caches behave almost in exactly the same way with 2 main differences:

- The SWA cache is much smaller
- The SWA cache automatically "forgets/prunes" old tokens upon successful commit (i.e. successful batch decode)

The size of the SWA cache is computed as:

```
PAD(n_swa*n_seq_max + n_batch)
```

This way we can store the cache data for the last `n_swa` tokens for all sequences and we also have room to evaluate a new batch of tokens with size up to `n_batch`.

Note that advanced cache operations such as removing tokens or shifting their positions are not possible when using SWA cache, because token information becomes lost when the window slides. For such cases, we can "fallback" to the old implementation by expanding the SWA cache size to the full context and disabling the SWA token pruning. This of course would lead to more memory usage. See the `swa_full` flag for more info.

The new `llama_kv_cache_unified_iswa` can be used for non-SWA models with `n_swa = n_ctx_train`.

---

## Main changes

- Move KV cache store and view logic from `llama-graph` to `llama-kv-cache`
- Move KV cache mask creation logic from `llama-graph` to `llama-kv-cache`
- The inputs to `build_attn_mha()` are now not permuted
- The QKV self-attention code is now more harmonious:
  ```c++
    const llama_kv_cache_unified * kv_self = static_cast<const llama_kv_cache_unified *>(memory);

    // store to KV cache
    {
        ggml_build_forward_expand(gf, kv_self->cpy_k(ctx0, k_cur, il));
        ggml_build_forward_expand(gf, kv_self->cpy_v(ctx0, v_cur, il));
    }

    const auto & kq_mask = inp->get_kq_mask();

    ggml_tensor * q = q_cur;
    ggml_tensor * k = kv_self->get_k(ctx0, il);
    ggml_tensor * v = kv_self->get_v(ctx0, il);

    ggml_tensor * cur = build_attn_mha(gf, q, k, v, kq_b, kq_mask, v_mla, kq_scale);
    cb(cur, "kqv_out", il);
  ```
[...]

PR #20087, opened 2026-03-03
Hybrid model cache: add `--checkpoint-every-nb`

pwilkin:
Add an option to create checkpoints after processing every `n` batches during prompt processing.

Hopefully solves #19794 #19298 #18497 and similar.

Usage: `llama-server -m model.gguf --checkpoint-every-nb 3`: creates a checkpoint every 3 batches.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -1279,13 +1279,20 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_env("LLAMA_ARG_SWA_FULL"));
     add_opt(common_arg(
-        {"--ctx-checkpoints", "--swa-checkpoints"}, "N",
+        {"-ctxcp", "--ctx-checkpoints", "--swa-checkpoints"}, "N",
         string_format("max number of context checkpoints to create per slot (default: %d)"
             "[(more info)](https://github.com/ggml-org/llama.cpp/pull/15293)", params.n_ctx_checkpoints),
         [](common_params & params, int value) {
             params.n_ctx_checkpoints = value;
         }
     ).set_env("LLAMA_ARG_CTX_CHECKPOINTS").set_examples({LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_CLI}));
+    add_opt(common_arg(
+        {"-cpent", "--checkpoint-every-n-tokens"}, "N",
+        string_format("create a checkpoint every n tokens during prefill (processing), -1 to disable (default: %d)", params.checkpoint_every_nt),
+        [](common_params & params, int value) {
+            params.checkpoint_every_nt = value;
+        }
+    ).set_env("LLAMA_ARG_CHECKPOINT_EVERY_NT").set_examples({LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_CLI}));
     add_opt(common_arg(
         {"-cram", "--cache-ram"}, "N",
         string_format("set the maximum cache size in MiB (default: %d, -1 - no limit, 0 - disable)"
--- a/common/common.h
+++ b/common/common.h
@@ -516,14 +516,15 @@ struct common_params {
     std::string cls_sep    = "\t";  // separator of classification sequences
 
     // server params
-    int32_t port              = 8080;         // server listens on this network port
-    int32_t timeout_read      = 600;          // http read timeout in seconds
-    int32_t timeout_write     = timeout_read; // http write timeout in seconds
-    int32_t n_threads_http    = -1;           // number of threads to process HTTP requests (TODO: support threadpool)
-    int32_t n_cache_reuse     = 0;            // min chunk size to reuse from the cache via KV shifting
-    bool    cache_prompt      = true;         // whether to enable prompt caching
-    int32_t n_ctx_checkpoints = 8;            // max number of context checkpoints per slot
-    int32_t cache_ram_mib     = 8192;         // -1 = no limit, 0 - disable, 1 = 1 MiB, etc.
+    int32_t port                = 8080;          // server listens on this network port
+    int32_t timeout_read        = 600;           // http read timeout in seconds
+    int32_t timeout_write       = timeout_read;  // http write timeout in seconds
+    int32_t n_threads_http      = -1;    // number of threads to process HTTP requests (TODO: support threadpool)
+    int32_t n_cache_reuse       = 0;     // min chunk size to reuse from the cache via KV shifting
+    bool    cache_prompt        = true;  // whether to enable prompt caching
+    int32_t n_ctx_checkpoints   = 32;     // max number of context checkpoints per slot
+    int32_t checkpoint_every_nt = 8192;   // make a checkpoint every n tokens during prefill
+    int32_t cache_ram_mib       = 8192;  // -1 = no limit, 0 - disable, 1 = 1 MiB, etc.
 
     std::string hostname      = "127.0.0.1";
     std::string public_path   = "";                                                                         // NOLINT
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -12,6 +12,7 @@
 #include "mtmd.h"
 #include "mtmd-helper.h"
 
+#include <algorithm>
 #include <cstddef>
 #include <cinttypes>
 #include <memory>
@@ -2348,8 +2349,10 @@ private:
                                     const auto it = std::find_if(
                                         slot.prompt.checkpoints.rbegin(),
                                         slot.prompt.checkpoints.rend(),
-                                        [&](const auto & cur) {
+                                        [&, func_name = __func__](const auto & cur) {
                                             // guarantee that a checkpoint will result in at least one token being processed [TAG_PROMPT_LOGITS]
+                                            LOG_INF("slot %12.*s: id %2d | task %d | Checking checkpoint with [%d, %d] against %d...\n", 12,
+                                                func_name, (slot).id, ((slot).task ? (slot).task->id : -1), cur.pos_min, cur.pos_max, pos_min_thold);
                                             return cur.pos_min < pos_min_thold;
                                         }
                                     );
@@ -2533,47 +2536,65 @@ private:
                         slot.i_batch   = batch.n_tokens - 1;
 
                         slot.init_sampler();
-
-                        const auto pos_min = llama_memory_seq_pos_min(llama_get_memory(ctx), slot.id);
-                        const auto pos_max = llama_memory_seq_pos_max(llama_get_memory(ctx), slot.id);
-
-                        // no need for empty or small checkpoints
-                        do_checkpoint = do_checkpoint && (pos_min >= 0 && pos_max >= 64);
-
-                        // no need to create checkpoints that are too close together
-                        do_checkpoint = do_checkpoint && (slot.prompt.checkpoints.empty() || pos_max > slot.prompt.checkpoints.back().pos_max + 64);
-
-                        // note: we create the checkpoint before calling llama_decode(), so the current batch is not
-                        //       yet processed and therefore it is not part of the checkpoint.
-                        if (do_checkpoint) {
-                            while (slot.prompt.checkpoints.size() >= (size_t) params_base.n_ctx_checkpoints) {
-                                // make room for the new checkpoint, if needed
-                                const auto & cur = slot.prompt.checkpoints.front();
-
-                                SLT_WRN(slot, "erasing old context checkpoint (pos_min = %d, pos_max = %d, n_tokens = %" PRId64 ", size = %.3f MiB)\n",
-                                        cur.pos_min, cur.pos_max, cur.n_tokens, (float) cur.data.size() / 1024 / 1024);
-
-                                slot.prompt.checkpoints.erase(slot.prompt.checkpoints.begin());
-                            }
-
-                            const size_t checkpoint_size = llama_state_seq_get_size_ext(ctx, slot.id, LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY);
-
-                            auto & cur = slot.prompt.checkpoints.emplace_back(server_prompt_checkpoint{
-                                /*.pos_min  = */ pos_min,
-                                /*.pos_max  = */ pos_max,
-                                /*.n_tokens = */ slot.prompt.n_tokens() - batch.n_tokens,
-                                /*.data     = */ std::vector<uint8_t>(checkpoint_size),
-                            });
-
-                            llama_state_seq_get_data_ext(ctx, cur.data.data(), checkpoint_size, slot.id, LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY);
-
-                            SLT_WRN(slot, "created context checkpoint %d of %d (pos_min = %d, pos_max = %d, n_tokens = %" PRId64 ", size = %.3f MiB)\n",
-                                    (int) slot.prompt.checkpoints.size(), params_base.n_ctx_checkpoints, cur.pos_min, cur.pos_max, cur.n_tokens, (float) cur.data.size() / 1024 / 1024);
-                        }
-
                         SLT_INF(slot, "prompt processing done, n_tokens = %d, batch.n_tokens = %d\n", slot.prompt.n_tokens(), batch.n_tokens);
                     } else {
+                        // only do non-end checkpoints if the "checkpoint every n tokens" option is set
+                        do_checkpoint = do_checkpoint && params_base.checkpoint_every_nt > 0;
+                        if (do_checkpoint) {
+                            llama_pos last_checkpoint = 0;
+                            if (!slot.prompt.checkpoints.empty()) {
+                                last_checkpoint = slot.prompt.checkpoints.back().n_tokens;
+                            }
+                            do_checkpoint = do_checkpoint && slot.prompt.n_tokens() - batch.n_tokens - last_checkpoint >= params_base.checkpoint_every_nt;
+                            if (do_checkpoint) {
+                                SLT_INF(slot, "%d tokens since last checkpoint at %d, creating new checkpoint during processing at position %d\n", params_base.checkpoint_every_nt, last_checkpoint, slot.prompt.n_tokens());
+                            }
+                        }
                         SLT_INF(slot, "prompt processing progress, n_tokens = %d, batch.n_tokens = %d, progress = %f\n", slot.prompt.n_tokens(), batch.n_tokens, (float) slot.prompt.n_tokens() / slot.task->n_tokens());
+                    }
+
+                    const auto pos_min = llama_memory_seq_pos_min(llama_get_memory(ctx), slot.id);
+                    const auto pos_max = llama_memory_seq_pos_max(llama_get_memory(ctx), slot.id);
+
+                    // no need for empty or small checkpoints
+                    do_checkpoint = do_checkpoint && (pos_min >= 0 && pos_max >= 64);
+
+                    // no need to create checkpoints that are too close together
+                    do_checkpoint = do_checkpoint && (slot.prompt.checkpoints.empty() || pos_max > slot.prompt.checkpoints.back().pos_max + 64);
+
+                    // note: we create the checkpoint before calling llama_decode(), so the current batch is not
+                    //       yet processed and therefore it is not part of the checkpoint.
+                    if (do_checkpoint) {
+                        while (slot.prompt.checkpoints.size() >= (size_t) params_base.n_ctx_checkpoints) {
+                            // make room for the new checkpoint, if needed
+                            const auto & cur = slot.prompt.checkpoints.front();
+
+                            SLT_WRN(slot,
+                                    "erasing old context checkpoint (pos_min = %d, pos_max = %d, n_tokens = %" PRId64
+                                    ", size = %.3f MiB)\n",
+                                    cur.pos_min, cur.pos_max, cur.n_tokens, (float) cur.data.size() / 1024 / 1024);
+
+                            slot.prompt.checkpoints.erase(slot.prompt.checkpoints.begin());
+                        }
+
+                        const size_t checkpoint_size =
+                            llama_state_seq_get_size_ext(ctx, slot.id, LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY);
+
+                        auto & cur = slot.prompt.checkpoints.emplace_back(server_prompt_checkpoint{
+                            /*.pos_min  = */ pos_min,
+                            /*.pos_max  = */ pos_max,
+                            /*.n_tokens = */ slot.prompt.n_tokens() - batch.n_tokens,
+                            /*.data     = */ std::vector<uint8_t>(checkpoint_size),
+                        });
+
+                        llama_state_seq_get_data_ext(ctx, cur.data.data(), checkpoint_size, slot.id,
+                                                     LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY);
+
+                        SLT_WRN(slot,
+                                "created context checkpoint %d of %d (pos_min = %d, pos_max = %d, n_tokens = %" PRId64
+                                ", size = %.3f MiB)\n",
+                                (int) slot.prompt.checkpoints.size(), params_base.n_ctx_checkpoints, cur.pos_min,
+                                cur.pos_max, cur.n_tokens, (float) cur.data.size() / 1024 / 1024);
                     }
                 }
 


Dampfinchen:
Thank you for your great work, as always. I wanted to ask, does this fix prompt caching not working after(!) the context size has been exceeded? Or is that a problem that is impossible to solve for RNN models? For non RNN models, the solution is simple: The beginning of the prompt is truncated by the UI and then sent to the backend to keep the chat rolling, prompt caching still works. This however doesn't work with Qwen 3.5.

IMO this is a dealbreaker for this kind of architecture, as not everyone can run 250k context + (and even that fills up eventually) so either you start a new chat regularly or live with very long prompt processing each enquiry. 

I am a bit confused about the many PRs around prompt caching for hybrid models to be honest, as prompt caching for Qwen 3.5 has been working as expected for quite a while now, as long as, and that is the important part, you stay within your context window.

pwilkin:
@Dampfinchen to be honest, the very idea of the hybrid architecture is that you can run a very long context cheaply, so this eliminates the need for context truncation. Unfortunately, due to how recurrent states are constructed, there is no way to take a "partial range" of a recurrent model's cache - once any prefix is invalid, you have to reprocess. 

This solution is mostly intended for situations where agentic coders keep a *certain* prefix of the prompt, but not long enough for the default solution (n_tokens - 512) to work. The idea behind this is basically - build up snapshots incrementally, and if at any time any prefix is needed then one of the checkpoints should probably work.

However, to reiterate, there is no possible solution for a situation where the prefix itself changes. For example, there's a long-standing issue with Claude Code which inserts a custom header that gets added to the *beginning* of the prompt each time - this breaks any sort of checkpointing. Likewise, if you have the exact datetime in your prompt, this will force reprocessing every time. There is nothing that can be done in those cases - those have to be fixed on the client side.

jhemmond:
> there's a long-standing issue with Claude Code which inserts a custom header that gets added to the _beginning_ of the prompt each time - this breaks any sort of checkpointing.

Exactly this bug is how I ended up here. Just switched from Opencode to Claude Code this afternoon, and was wondering why Qwen3.5-122B is reprocessing every single prompt. Will this flag (--checkpoint-every-nb) be enabled by default or has to be set explicitly?


Thanks for the great work!

arcanemachine:
> Will this flag (--checkpoint-every-nb) be enabled by default or has to be set explicitly?

I built it locally, and it looks like it is disabled by default. Here is the relevant section from `llama-server --help`:

```
--checkpoint-every-nb N                 create a checkpoint every n batches during prefill (processing), -1 to
                                        disable (default: -1)                                                                
```

I am still having to rebuild the cache in OpenCode after sending a new message after the context window gets above ~50,000 tokens (tried with `--checkpoint-every-nb 3`, then also tried with that flag and `--swa-full`, still getting the cache issue):

```
srv  log_server_r: done request: POST /v1/chat/completions 192.168.1.100 200
slot print_timing: id  3 | task 3166 | 
prompt eval time =     534.69 ms /    96 tokens (    5.57 ms per token,   179.54 tokens per second)
       eval time =    2018.21 ms /    68 tokens (   29.68 ms per token,    33.69 tokens per second)
      total time =    2552.90 ms /   164 tokens
slot      release: id  3 | task 3166 | stop processing: n_tokens = 55002, truncated = 0
srv  params_from_: Chat format: peg-constructed
slot get_availabl: id  3 | task -1 | selected slot by LCP similarity, sim_best = 0.918 (> 0.100 thold), f_keep = 0.918
slot launch_slot_: id  3 | task -1 | sampler chain: logits -> ?penalties -> ?dry -> ?top-n-sigma -> top-k -> ?typical -> ?top-p -> min-p -> ?xtc -> ?temp-ext -> dist 
slot launch_slot_: id  3 | task 3235 | processing task, is_child = 0
slot update_slots: id  3 | task 3235 | new prompt, n_ctx_slot = 120064, n_keep = 0, task.n_tokens = 55049
slot update_slots: id  3 | task 3235 | n_past = 50518, slot.prompt.tokens.size() = 55002, seq_id = 3, pos_min = 55001, n_swa = 1
slot update_slots: id  3 | task 3235 | forcing full prompt re-processing due to lack of cache data (likely due to SWA or hybrid/recurrent memory, see https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055)
slot update_slots: id  3 | task 3235 | erased invalidated context checkpoint (pos_min = 53491, pos_max = 53491, n_tokens = 53492, n_swa = 1, size = 75.376 MiB)
slot update_slots: id  3 | task 3235 | erased invalidated context checkpoint (pos_min = 53599, pos_max = 53599, n_tokens = 53600, n_swa = 1, size = 75.376 MiB)
slot update_slots: id  3 | task 3235 | erased invalidated context checkpoint (pos_min = 53689, pos_max = 53689, n_tokens = 53690, n_swa = 1, size = 75.376 MiB)
slot update_slots: id  3 | task 3235 | erased invalidated context checkpoint (pos_min = 53898, pos_max = 53898, n_tokens = 53899, n_swa = 1, size = 75.376 MiB)
slot update_slots: id  3 | task 3235 | erased invalidated context checkpoint (pos_min = 53981, pos_max = 53981, n_tokens = 53982, n_swa = 1, size = 75.376 MiB)
slot update_slots: id  3 | task 3235 | erased invalidated context checkpoint (pos_min = 54383, pos_max = 54383, n_tokens = 54384, n_swa = 1, size = 75.376 MiB)
slot update_slots: id  3 | task 3235 | erased invalidated context checkpoint (pos_min = 54529, pos_max = 54529, n_tokens = 54530, n_swa = 1, size = 75.376 MiB)
slot update_slots: id  3 | task 3235 | erased invalidated context checkpoint (pos_min = 54838, pos_max = 54838, n_tokens = 54839, n_swa = 1, size = 75.376 MiB)
slot update_slots: id  3 | task 3235 | n_tokens = 0, memory_seq_rm [0, end)
slot update_slots: id  3 | task 3235 | prompt processing progress, n_tokens = 2048, batch.n_tokens = 2048, progress = 0.037203
slot update_slots: id  3 | task 3235 | n_tokens = 2048, memory_seq_rm [2048, end)
slot update_slots: id  3 | task 3235 | prompt processing progress, n_tokens = 4096, batch.n_tokens = 2048, progress = 0.074406
slot update_slots: id  3 | task 3235 | n_tokens = 4096, memory_seq_rm [4096, end)
slot update_slots: id  3 | task 3235 | prompt processing progress, n_tokens = 6144, batch.n_tokens = 2048, progress = 0.111610
slot update_slots: id  3 | task 3235 | n_tokens = 6144, memory_seq_rm [6144, end)
slot update_slots: id  3 | task 3235 | 3 batches since last checkpoint at 0, creating new checkpoint during processing at position 8192
slot update_slots: id  3 | task 3235 | prompt processing progress, n_tokens = 8192, batch.n_tokens = 2048, progress = 0.148813
slot update_slots: id  3 | task 3235 | created context checkpoint 1 of 8 (pos_min = 6143, pos_max = 6143, n_tokens = 6144, size = 75.376 MiB)
...
```

The strange thing is that tool calls don't trigger this issue (using OpenCode), only manual messages. Makes me wonder if this is some sort of issue with OpenCode, or maybe some PEBKAC issue on my part.

EDIT: I am successfully working around the issue with `--ctx-checkpoints 128`.

allkhor:
@jhemmond, take a look at #20003 — maybe this will help you.

ggerganov:
@pwilkin I don't think this flag is needed. Which case does it fix that is not already covered by the existing logic?

pwilkin:
@ggerganov From all the reports I gathered that there are cases where the agent reprocessing cuts of a larger part of the prompt while still keeping a reasonably large prefix - a notable example of such cases is when the agent doesn't pass back reasoning content to the model, so the last reasoning content gets cut off (and it's often more than 512 tokens).

pwilkin:
Another thing this helps with is keeping a checkpoint with all (or most) of the agent's fixed tools / instructions header, which can go up to 20k tokens in some cases.

whoreson:
> @pwilkin I don't think this flag is needed. Which case does it fix that is not already covered by the existing logic?

I have explained 4 months ago here: https://github.com/ggml-org/llama.cpp/pull/17428

ggerganov:
> a notable example of such cases is when the agent doesn't pass back reasoning content to the model, so the last reasoning content gets cut off (and it's often more than 512 tokens).

This change is not going to fix that - when the reasoning is removed, the prefix changes. So making checkpoints during reasoning will just increase memory usage without reducing computation.

pwilkin:
> This change is not going to fix that - when the reasoning is removed, the prefix changes. So making checkpoints during reasoning will just increase memory usage without reducing computation.

No, because in a typical agentic scenario, you're going to have something like this:

[AGENT INSTRUCTIONS][USER QUERY][AGENT REASONING][AGENT RESPONSE]

Now, if reasoning is removed, you'll have

[AGENT INSTRUCTIONS][USER QUERY][AGENT RESPONSE][USER REQUERY]

Although some part is removed, you're still left with the [AGENT INSTRUCTIONS][USER QUERY] part, which can be quite large, esp. if files have been attached.

tools/server/server-context.cpp
```diff
@@ -2533,47 +2533,65 @@ struct server_context_impl {
...
 
                         slot.init_sampler();
+                        SLT_INF(slot, "prompt processing done, n_tokens = %d, batch.n_tokens = %d\n", slot.prompt.n_tokens(), batch.n_tokens);
+                    } else {
+                        // only do non-end checkpoints if the "checkpoint every n batches" option is set
+                        do_checkpoint = do_checkpoint && params_base.checkpoint_every_nb > 0;
+                        if (do_checkpoint) {
+                            llama_pos last_checkpoint = 0;
+                            if (!slot.prompt.checkpoints.empty()) {
+                                last_checkpoint = slot.prompt.checkpoints.back().pos_max;
```
ggerganov: Should use `n_tokens` here instead of the `pos_max`:

```suggestion
                                last_checkpoint = slot.prompt.checkpoints.back().n_tokens;
```

common/arg.cpp
```diff
@@ -1286,6 +1286,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.n_ctx_checkpoints = value;
         }
     ).set_env("LLAMA_ARG_CTX_CHECKPOINTS").set_examples({LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_CLI}));
+    add_opt(common_arg(
+        {"--checkpoint-every-nb"}, "N",
```
ggerganov: Don't think there is advantage in coupling this argument to the batch size. Better to express this in terms of tokens. For example `--ctx-checkpoints-interval N`, where `N` is number of tokens. N = 4096 or 8192 could be a good default.

smilediver:
I have a case where the prompt looks like `[append_part][dynamic_part]`, where the model is called with the same `append_part` and different `dynamic_part` several times (different queries for the same data). After that, `append_part` is appended and the process is repeated.

ggerganov:
The `[AGENT INSTRUCTIONS][USER QUERY]` is checkpointed with the existing logic, so it's not a good example - it should still work even without this PR.

I think a better example where the current logic on `master` would fail is if you continue an existing session that looks like this for example:

```
system
user0
assistant0
user1
assistant1
...
userN
assistantN
userN+1
```

With a fresh `llama-server` we will create only one checkpoint right before `userN+1`. Now if we want to go back to an earlier message (e.g. `userN-1`) and branch from there, we won't have checkpoint data (because the `llama-server` has just started). I guess this is something that can happen often with OpenCode, so it's good to support it.

Ideally, the best logic is to create a checkpoint before each user message. But it might be more complicated to do that than the basic interval-based checkpointing proposed here.

@smilediver If you have control over the client, you can send `[append_part]` alone once at the start to force a checkpoint creation and then follow up with the different dynamic parts. This is something we utilize in the `llama.vim` plugin to reduce computation.

common/arg.cpp
pwilkin: Done, changed it to `--checkpoint-every-n-tokens` with short form `-cpent`.

pwilkin:
> The `[AGENT INSTRUCTIONS][USER QUERY]` is checkpointed with the existing logic, so it's not a good example - it should still work even without this PR.

It is if you start a new conversation, yes. But if you resume a session, as people often do with those agents, then it doesn't work anymore. There are a lot of real-life edge cases here which is I thought that this approach would be the simplest to capture most of them.

common/common.h
```diff
@@ -516,14 +516,15 @@ struct common_params {
...
-    int32_t n_ctx_checkpoints = 8;            // max number of context checkpoints per slot
-    int32_t cache_ram_mib     = 8192;         // -1 = no limit, 0 - disable, 1 = 1 MiB, etc.
+    int32_t port                = 8080;          // server listens on this network port
+    int32_t timeout_read        = 600;           // http read timeout in seconds
+    int32_t timeout_write       = timeout_read;  // http write timeout in seconds
+    int32_t n_threads_http      = -1;    // number of threads to process HTTP requests (TODO: support threadpool)
+    int32_t n_cache_reuse       = 0;     // min chunk size to reuse from the cache via KV shifting
+    bool    cache_prompt        = true;  // whether to enable prompt caching
+    int32_t n_ctx_checkpoints   = 8;     // max number of context checkpoints per slot
+    int32_t checkpoint_every_nt = -1;    // make a checkpoint every n batches during prefill
```
ggerganov:
```suggestion
    int32_t checkpoint_every_nt = -1;    // make a checkpoint every n tokens during prefill
```

ggerganov (commented):
I think we can bump the default number of checkpoints to 32 and enable this new functionality by default to make checkpoints every 8192 tokens.

LucaAmigoni:
> However, to reiterate, there is no possible solution for a situation where the prefix itself changes. For example, there's a long-standing issue with Claude Code which inserts a custom header that gets added to the _beginning_ of the prompt each time - this breaks any sort of checkpointing. Likewise, if you have the exact datetime in your prompt, this will force reprocessing every time. There is nothing that can be done in those cases - those have to be fixed on the client side.

You can set the env variable CLAUDE_CODE_ATTRIBUTION_HEADER to 0 to avoid the custom header in CC.

aagit:
FWIW  to keep it minimal I used the logic batch size instead of a new config option. https://github.com/ggml-org/llama.cpp/pull/19970 . Hardcoded to 8k isn't ideal.

aagit:
> However, to reiterate, there is no possible solution for a situation where the prefix itself changes. For example, there's a long-standing issue with Claude Code which inserts a custom header that gets added to the _beginning_ of the prompt each time - this breaks any sort of checkpointing. Likewise, if you have the exact datetime in your prompt, this will force reprocessing every time. There is nothing that can be done in those cases - those have to be fixed on the client side.

yes luckily gptel never alters the prompt and the prompt performs much better after the context anyway... verified both with rg-edit and synthmerge_bench for all models except gemini 2.5 with reasoning enabled. When you're editing files in context you're inevitably going to truncate context. gptel context LRU management PR [I posted](https://github.com/ggml-org/llama.cpp/pull/19970) makes sure the most frequently edited files are pushed to the end of context for this reason, it gives an extra noticeable boost.

pwilkin:
> You can set the env variable CLAUDE_CODE_ATTRIBUTION_HEADER to 0 to avoid the custom header in CC.

Ye I know that one :) but what I mean is that if you don't use that option, there is literally no way to avoid reprocessing.

pwilkin:
@ggerganov aight, I changed the defaults to 32 checkpoints and checkpointing every 8192 tokens.

ggerganov (approved):
> Ideally, the best logic is to create a checkpoint before each user message. But it might be more complicated to do that than the basic interval-based checkpointing proposed here.

If you are interested to improve this further, I think this can be implemented during chat formatting to emit a list of token positions based on the locations of the user messages. And then use this list to determine the checkpoint locations during prompt processing.

Dampfinchen:
I have just stumbled across this comment: https://www.reddit.com/r/LocalLLaMA/comments/1rlkptk/comment/o8ta804/

> `I noticed Qwen always reprocess it's lasts response, this is because the chat template is configured to not include previous thinking in the context. This is fine, it's a tradeoff and that's how Qwen was trained.
> 
> The problem is, the same happens with thinking disabled. I think the cause is that even with thinking disabled there are empty think tags in the template, which are also removed from the conversation, causing a context shift and forcing reprocessing
> 
> It would be nice to modify the template to keep empty tags In the conversation history`

Might or might not have some relevancy here.

pwilkin:
@ggerganov might try this, but that's largely nontrivial - when we parse the chat message, we first detokenize it, so at the point of parsing the parser doesn't know what the token positions are. I know @ngxson has been thinking of making the Jinja engine token-aware, but IIRC that's a work in progress as well.

JamePeng:
I suggest adding a timestamp to the cache. For cache management of the same prompt word, each time the best cache is restored, only the timestamp needs to be updated. In addition, cache blocks that have not been hit for a long time in the cache stack can be discarded first.

Dampfinchen:
> @Dampfinchen to be honest, the very idea of the hybrid architecture is that you can run a very long context cheaply, so this eliminates the need for context truncation. Unfortunately, due to how recurrent states are constructed, there is no way to take a "partial range" of a recurrent model's cache - once any prefix is invalid, you have to reprocess.
> 
> This solution is mostly intended for situations where agentic coders keep a _certain_ prefix of the prompt, but not long enough for the default solution (n_tokens - 512) to work. The idea behind this is basically - build up snapshots incrementally, and if at any time any prefix is needed then one of the checkpoints should probably work.
> 
> However, to reiterate, there is no possible solution for a situation where the prefix itself changes. For example, there's a long-standing issue with Claude Code which inserts a custom header that gets added to the _beginning_ of the prompt each time - this breaks any sort of checkpointing. Likewise, if you have the exact datetime in your prompt, this will force reprocessing every time. There is nothing that can be done in those cases - those have to be fixed on the client side.

Thank you for your explanation and your fantastic work. I have nothing to add except for disagreeing with "run very long context cheaply, so this eliminates the need for context truncation."

Even for longer context, you will run into this issue eventually. Regardless if you use 10K context or 300K, context will fill up eventually and then you run into this issue, with more context it just takes longer to happen. To make matters worse, agentic workflows and vision tasks easily fill up context in no time. With non hybrid models, you just keep cruising not having to worry about it. 

I'm really not a fan of this approach "just use more context" even for very efficient models like Qwen 3.5 more context still adds more memory and needs more compute which not everyone has especially in these times. To be honest, I think the concept is having a context that eventually fills up and will have to be reset is fundamentally flawed, imagine if the human brain worked like that. In 5 years I predict we will look at context in LLMs the same way we look at dial up internet now.

Anyways, thank you again for your great work. I realize there is nothing you can do about this. I hope Qwen will change the architecture in the future so atleast a rolling context is viable again. But more so, for the future I hope the whole concept of having a context for LLMs will be completely overhauled.

ggerganov:
> With non hybrid models, you just keep cruising not having to worry about it.

@Dampfinchen Just to make sure I understand, with non-recurrent models did you use the `--cache-reuse` functionality in order to avoid the reprocessing when the context becomes full?

IIIIIllllIIIIIlllll:
Sorry to bother you. I downloaded the latest code from the main branch and compiled it using ROCm, but it doesn't work with this parameter. After checking the code modified in this PR, I noticed that another parameter is used in the code.
Am I missing something?

```
llama-server -m /home/mark/Models/Q8/Qwen3.5-35B-A3B-Q8_0/Qwen3.5-35B-A3B-UD-Q8_K_XL.gguf 
--port 8088 
--mmproj /home/mark/Models/Q8/Qwen3.5-35B-A3B-Q8_0/mmproj-F32.gguf 
--ctx-size 262144 
--flash-attn on --no-mmap 
--temp 1 --top-p 0.95 --top-k 20 --min-p 0.0 --presence-penalty 1.5 --repeat-penalty 1 --frequency-penalty 0.0 
--batch-size 4096 --ubatch-size 4096 
--parallel 4 
--cache-ram -1 
--cache-type-k f16 --cache-type-v f16 
--threads -1 --seed -1 -dio --no-webui 
--reasoning-budget -1 
--swa-full 
--checkpoint-every-nb 3 
--chat-template-file /home/mark/App/llama.cpp/cache/Qwen3.5-35B-A3B-Q8_0.jinja 
--metrics 
--slot-save-path /home/mark/App/llama.cpp/cache --alias Qwen3.5-35B-A3B-Q8_0 
--timeout 36000 --host 0.0.0.0
```

aagit:
I just tested the new upstream commit with default settings, and as anticipated, my reproducer (available in my previous PR) shows a regression in truncated kvcache lookup time.

**Default 8k Settings:**
```text
kvcache preload: 32.5362 seconds
truncated kvcache lookup: 3.3391 seconds
kvcache preload: 32.5151 seconds
truncated kvcache lookup: 3.3146 seconds
```

**`--checkpoint-every-n-tokens 2048`** restores the performance to the level of my previous setup:
```text
kvcache preload: 33.0262 seconds
truncated kvcache lookup: 0.7092 seconds
```

### Why would I care about 2 extra seconds after truncation?

1.  **Context Ingestion:** I add source files (and relevant buffers) to the context using `gptel-add-file`/`gptel-add` via a per-project function (after clearing the context).
2.  **Refactoring:** I traverse a ~100k files project using `ripgrep-edit` with regex patterns. I have `llama.cpp` rewrite the `rg-edit` buffer under the control of a dynamically generated GBNF grammar with two algo picked control lines at the start and end of each snippnet.
3.  **Flush:** I close the ripgrep buffer, flushing the changes.

The razor conding workflow gives the full source file that is going to be rewritten in point 1, but by ripping through the source with ripgrep in point 2, input and output of the LLM (provided to the LLM after context) are as razor thin as possible, greatly reducing the generation wait time and improving accuracy of the output too. I never rewrite files in full, that's not workable with dozen of files to edit in one go all 10k lines (not tokens) long.

**The Issue:**
Step 3 inevitably truncates the context, so the extra 2 sec wait is added to most new requests.

### Suggestion to avoid an extra ~2 sec delay after kvcache truncation if your workflow is similar
*   `--checkpoint-every-n-tokens 2048`
*   `--ctx-checkpoints` (increase as needed, e.g., to 64 to cover the last 128k of context)

This is combined with context managed in LRU fashion, with which also improves accuracy further. The files that are most frequently modified by llama.cpp are refiled at the of context using a gptel-context-optimizer module hooking into file save/invalidate hooks, so the truncation is most frequently happening at the end of the kvcache.

Thank you for merging this configurable solution upstream so I can use these new models in addition of my devstral-2-small default pick!

Dampfinchen:
> > With non hybrid models, you just keep cruising not having to worry about it.
> 
> @Dampfinchen Just to make sure I understand, with non-recurrent models did you use the `--cache-reuse` functionality in order to avoid the reprocessing when the context becomes full?

Yeah. When the context of the prompt is larger than the maximum context size set with -c and the UI, the UI connected to llama.cpp server has to truncate older messages to keep a rolling chat window going. With non-reccurent models this is not an issue, as the cache reuse function works as expected. But with hybrid model this is not supported. This means that every time I send a message to the AI, it has to reprocess the entire prompt again and again which slows down things alot. 

> ``forcing full prompt re-processing due to lack of cache data (likely due to SWA or hybrid/recurrent memory, see https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055)`

But yeah nothing llama.cpp can do about that. Perhaps some UI trickery could work, like truncating a huge chunk of the prompt at once when it goes near the maximum context, then after that intitial processing you'd be safe from reprocessing until the ctx is almost equal to the max ctx again and the thing repeats. I think koboldcpp's smart context does that.

Petrox:
I just want to thank you for your efforts. (Hope this comment does not bother anyone after the closed PR. :) )

aldehir:
Never seen this before, might be a regression introduced here:

gpt-oss-120b
https://cdn.alde.dev/llama.cpp/videos/pp-loop.mp4

Seems like an off-by-one issue, `509/508 = 1.001969`.

ggerganov:
What is your server command?

aldehir:
> What is your server command?

Nothing fancy, although the auto-fit will offload to CPU since I cannot load the model entirely in VRAM. No other logs, the loop filled my scrollback buffer.
```
llama-server.exe -m E:\Models\gpt-oss-120b\gpt-oss-120b-F16.gguf -c 64000
```
ae87863d

I'll troubleshoot a bit more and see if I can reliably reproduce. Wanted to share in case there was something obvious to the smarter people in the room.

ggerganov:
Yes, the logic is a bit flaky there - it's most likely an off-by-one error.

With #20277 it will be easier to trace it down.

PR #20105, opened 2026-03-04
mtmd : rename mtmd_get_audio_bitrate to mtmd_get_audio_sample_rate

danbev:
This commit renames the the function `mtmd_get_audio_bitrate` to `mtmd_get_audio_sample_rate` to better reflect its purpose.

The motivation for this is that the function currently returns the audio sample rate, not the bitrate (sample_rate × bit_depth × channels), and that is how it is used in the code as well.

This is a breaking change, but I believe mtmd is still in experimental/development phase so it might be alright to simply rename.


elfarolab:
@danbev 

Hello Daniel,

audio bitrate and samplerate are two completely different things in audio. I agree with you about the naming but the specific bitrate could be also affected by the codec used when not raw audio.

ngxson (approved):

PR #20112, opened 2026-03-04
Add unit test coverage for llama_tensor_get_type

bartowski1182:
This is part of a larger goal of reworking or replacing the `llama_tensor_get_type` function

Before major work starts in that area, I want to capture the current existing behaviour thoroughly, so that any accidental changes are easy to spot, and any purposeful changes are easy to document

To that end, this PR introduces unit test coverage for the function itself

Using a pre-set list of models, and taking advantage of the new `gguf-model-data` utility, these tests pull real model metadata directly from huggingface, create mock models/tensors, and runs them through the `llama_tensor_get_type` function and document the schema into the `tests/snapshots/` directory as `model-name.schema`

I hope that the storage method I've decided to use won't be overly burdensome, I iterated a lot through various ways to store the existing tensor layouts, and I think this landed in an acceptable compromise area: not too many files, "only" 700 KB of storage

If this is not acceptable, I can go back to the drawing board

To capture current layouts of tensors, the script can be run with the `--generate` flag which will download the metadata and produce the schema files:

<details>
<summary>`--generate output`</summary>

```
./build/bin/test-quant-type-selection --generate
This will overwrite all snapshot files in:
  /home/colin/git_repos/forks/quant-refactor/tests/snapshots
Continue? [y/N] y
Fetching model metadata for Qwen3-0.6B from ggml-org/Qwen3-0.6B-GGUF...
gguf_fetch: loaded from cache: /home/colin/.cache/llama.cpp/gguf-headers//ggml-org_Qwen3-0.6B-GGUF--Qwen3-0.6B-Q8_0.gguf.partial
  wrote /home/colin/git_repos/forks/quant-refactor/tests/snapshots/qwen3-0.6b.schema
Fetching model metadata for GLM-4.6V from ggml-org/GLM-4.6V-GGUF...
[truncated]
  wrote /home/colin/git_repos/forks/quant-refactor/tests/snapshots/qwen3.5-397b-a17b.schema
Fetching model metadata for Qwen3.5-27B from bartowski/Qwen_Qwen3.5-27B-GGUF...
gguf_fetch: loaded from cache: /home/colin/.cache/llama.cpp/gguf-headers//bartowski_Qwen_Qwen3.5-27B-GGUF--Qwen_Qwen3.5-27B-Q8_0.gguf.partial
  wrote /home/colin/git_repos/forks/quant-refactor/tests/snapshots/qwen3.5-27b.schema
12 files written
```
</details>

Then you can run the test itself without any arguments:

<details>
<summary>`test output`</summary>

```
./build/bin/test-quant-type-selection
=== Qwen3-0.6B ===
Fetching model metadata for Qwen3-0.6B from ggml-org/Qwen3-0.6B-GGUF...
gguf_fetch: loaded from cache: /home/colin/.cache/llama.cpp/gguf-headers//ggml-org_Qwen3-0.6B-GGUF--Qwen3-0.6B-Q8_0.gguf.partial
  PASS  Qwen3-0.6B: 33/33 ftype sections passed (198 tensors)
[truncated]
=== Qwen3.5-27B ===
Fetching model metadata for Qwen3.5-27B from bartowski/Qwen_Qwen3.5-27B-GGUF...
gguf_fetch: loaded from cache: /home/colin/.cache/llama.cpp/gguf-headers//bartowski_Qwen_Qwen3.5-27B-GGUF--Qwen_Qwen3.5-27B-Q8_0.gguf.partial
  PASS  Qwen3.5-27B: 33/33 ftype sections passed (498 tensors)

12/12 models passed
```
</details>

There are a few changes to llama-quant.cpp, notably making the function itself non-static so I can access it, extracting `init_quantize_state_counters` and `llama_ftype_default_type` so I can use them without recreating them, and lastly I pulled in the change from @ddh0 in their PR #19770 that adds `tensor_allows_quantization` (except I made it non-static) since I needed a function like that, so once that PR is merged (and this is rebased on latest) that change will go away

I also moved a couple things like `quantize_state_impl` to the header file for similar reasons

Finally, I found an issue with the `gguf-model-data` `gguf_read_uint32_val` when the per-layer head counts are an array like with Step-3.5-Flash, so I've added a fix for that and a unit test to capture that behaviour


bartowski1182:
Regarding the +23,000 lines and 700kb from the .schema files, I wouldn't be opposed to instead suggesting maintainers run the script on mainline and then again on their changes, so then the files don't have to exist in the actual repo itself, would certainly declutter

tests/test-quant-type-selection.cpp
```diff
@@ -0,0 +1,510 @@
+#include "../src/llama-arch.h"
+#include "../src/llama-model.h"
+#include "../src/llama-quant.h"
```
bartowski1182: @ggerganov 

I know you mentioned to @ddh0 about these kinds of includes and to avoid by duplicating the structs

Figured I should check, is my case a special one because it's being done in tests?

If not, I'm willing to back-burner this and look into the changes to the API you wanted first

ggerganov: Tests are OK to include internal files. It's problematic for tools and examples because 3rd parties that use `libllama` will not be able to do that (include internal files) - they only work with the public API.

ggerganov:
How long does it currently take to generate the snapshots?

bartowski1182:
16 seconds if the models aren't cached, 3 seconds if they are


<details>
<summary>timed</summary>

```
time ./build/bin/test-quant-type-selection --generate
This will overwrite all snapshot files in:
  /home/colin/git_repos/forks/quant-refactor/tests/snapshots
Continue? [y/N] y

Fetching model metadata for Qwen3-0.6B from ggml-org/Qwen3-0.6B-GGUF...
gguf_fetch: downloading 2097152 bytes from Qwen3-0.6B-Q8_0.gguf
...
gguf_fetch: downloading 16777216 bytes from Qwen_Qwen3.5-27B-Q8_0.gguf
  wrote /home/colin/git_repos/forks/quant-refactor/tests/snapshots/qwen3.5-27b.schema
12 files written

real    0m16.441s
user    0m5.363s
sys     0m0.713s
```

```
time ./build/bin/test-quant-type-selection --generate
This will overwrite all snapshot files in:
  /home/colin/git_repos/forks/quant-refactor/tests/snapshots
Continue? [y/N] y

Fetching model metadata for Qwen3-0.6B from ggml-org/Qwen3-0.6B-GGUF...
gguf_fetch: loaded from cache: /home/colin/.cache/llama.cpp/gguf-headers//ggml-org_Qwen3-0.6B-GGUF--Qwen3-0.6B-Q8_0.gguf.partial
...
gguf_fetch: loaded from cache: /home/colin/.cache/llama.cpp/gguf-headers//bartowski_Qwen_Qwen3.5-27B-GGUF--Qwen_Qwen3.5-27B-Q8_0.gguf.partial
  wrote /home/colin/git_repos/forks/quant-refactor/tests/snapshots/qwen3.5-27b.schema
12 files written

real    0m3.010s
user    0m0.973s
sys     0m0.098s
```

</details>

src/llama-quant.cpp
```diff
@@ -1,25 +1,18 @@
...
 #include <cinttypes>
 #include <fstream>
 #include <mutex>
 #include <regex>
 #include <thread>
 #include <unordered_map>
 
-// result of parsing --tensor-type option
-// (changes to this struct must be reflected in tools/quantize/quantize.cpp)
-struct tensor_type_option {
```
bartowski1182, line 18: Moved to llama-quant.h

src/llama-quant.cpp
```diff
@@ -1,25 +1,18 @@
-#include "llama.h"
```
bartowski1182, line 1: Moved to llama-quant.h

src/llama-quant.cpp
```diff
@@ -1,25 +1,18 @@
...
+#include "llama-quant.h"
 #include "llama-impl.h"
 #include "llama-model.h"
 #include "llama-model-loader.h"
 
+#include <algorithm>
 #include <cmath>
 #include <cstring>
-#include <string>
```
bartowski1182, line 8: Moved to llama-quant.h

src/llama-quant.cpp
```diff
@@ -157,46 +150,6 @@ static bool category_is_attn_v(tensor_category cat) {
            cat == tensor_category::ATTENTION_KV_B;
 }
 
-//
-// quantization state
-//
-
-struct quantize_state_impl {
```
bartowski1182, line 163: Moved to llama-quant.h

src/llama-quant.cpp
```diff
@@ -286,7 +239,7 @@ static void llama_tensor_dequantize_impl(
 // do we allow this tensor to be quantized?
 //
 
-static bool tensor_allows_quantization(const llama_model_quantize_params * params, llm_arch arch, const ggml_tensor * tensor) {
+bool tensor_allows_quantization(const llama_model_quantize_params * params, llm_arch arch, const ggml_tensor * tensor) {
```
bartowski1182: Made public so I can use it in my tests

src/llama-quant.cpp
```diff
@@ -652,8 +604,15 @@ static ggml_type llama_tensor_get_type_impl(quantize_state_impl & qs, ggml_type
     return new_type;
 }
 
-// outer wrapper: determine the ggml_type that this tensor should be quantized to
-static ggml_type llama_tensor_get_type(quantize_state_impl & qs, const llama_model_quantize_params * params, const ggml_tensor * tensor, ggml_type default_type, const tensor_metadata & tm) {
+// public API: compute category from tensor name and delegate to _impl
+ggml_type llama_tensor_get_type(quantize_state_impl & qs, ggml_type new_type, const ggml_tensor * tensor, llama_ftype ftype) {
```
bartowski1182: Swapped to an internal/public API so that I can avoid the need for ggml_get_name and tensor_get_category

src/llama-quant.cpp
```diff
@@ -784,7 +743,7 @@ static bool tensor_requires_imatrix(const char * tensor_name, const ggml_type ds
 // given a file type, get the default tensor type
 //
 
-static ggml_type llama_ftype_get_default_type(llama_ftype ftype) {
+ggml_type llama_ftype_get_default_type(llama_ftype ftype) {
```
bartowski1182, line 792: Made public so it can be tested

src/llama-quant.cpp
```diff
@@ -823,16 +782,89 @@ static ggml_type llama_ftype_get_default_type(llama_ftype ftype) {
         case LLAMA_FTYPE_MOSTLY_IQ3_S:
         case LLAMA_FTYPE_MOSTLY_IQ3_M:   return GGML_TYPE_IQ3_S;
 
-        default: throw std::runtime_error(format("invalid output file type %d\n", ftype));
```
bartowski1182, line 830: This was causing annoyance when trying to iterate through all FTYPEs, 1 by 1, so removed the `throw` from here and added it in place below

I could just iterate in a way where I avoid the missing middle quants (Q4_0_4_4 etc) and revert this

src/llama-quant.cpp
```diff
@@ -823,16 +782,89 @@ static ggml_type llama_ftype_get_default_type(llama_ftype ftype) {
...
+const char * llama_ftype_to_name(llama_ftype ftype) {
+    for (const auto & e : ftype_name_table) {
+        if (e.ftype == ftype) {
+            return e.name;
+        }
+    }
+    return nullptr;
+}
+
+void init_quantize_state_counters(quantize_state_impl & qs, const std::vector<std::string> & tensor_names) {
```
bartowski1182: extracted the initialization of state counters so I can use them outside the main function

src/llama-quant.cpp
```diff
@@ -841,7 +873,10 @@ static void llama_model_quantize_impl(const std::string & fname_inp, const std::
         nthread = std::thread::hardware_concurrency();
     }
 
-    default_type = llama_ftype_get_default_type(ftype);
+    ggml_type default_type = llama_ftype_get_default_type(ftype);
+    if (default_type == GGML_TYPE_COUNT) {
```
bartowski1182, line 866: (this is where I added the throw back)

bartowski1182 (commented):
Adding comments to justify changes to `llama-quant.cpp`

src/llama-quant.cpp
```diff
@@ -823,16 +751,90 @@ static ggml_type llama_ftype_get_default_type(llama_ftype ftype) {
...
+    { "TQ1_0",      LLAMA_FTYPE_MOSTLY_TQ1_0 },
+    { "TQ2_0",      LLAMA_FTYPE_MOSTLY_TQ2_0 },
+    { "MXFP4_MOE",  LLAMA_FTYPE_MOSTLY_MXFP4_MOE },
+};
+
+llama_ftype llama_ftype_from_name(const char * name) {
+    for (const auto & e : ftype_name_table) {
+        if (strcmp(name, e.name) == 0) {
+            return e.ftype;
+        }
```
ddh0: Can this logic reasonably be deduplicated with respect to `static std::string llama_model_ftype_name` from `src/llama-model-loader.cpp`?

If not, can it be made identical?

src/llama-quant.cpp
```diff
@@ -823,16 +751,90 @@ static ggml_type llama_ftype_get_default_type(llama_ftype ftype) {
...
     }
+    return (llama_ftype)-1;
+}
+
+const char * llama_ftype_to_name(llama_ftype ftype) {
```
ddh0: (same as the above?)

src/llama-quant.cpp
```diff
@@ -961,6 +966,15 @@ static void llama_model_quantize_impl(const std::string & fname_inp, const std::
         });
     }
 
+    // compute tensor metadata once and cache it
+    std::vector<tensor_metadata> metadata(tensors.size());
+    for (size_t i = 0; i < tensors.size(); ++i) {
+        metadata[i].name = ggml_get_name(tensors[i]->tensor);
```
ddh0, line 990:
```suggestion
        metadata[i].name = tensors[i]->tensor->name;
```

Nitpick, not sure if it matters, but calling `ggml_get_name` is generally unnecessary where we can just do `tensor->name`.

src/llama-quant.cpp
bartowski1182: The problem with the one in `llama-model-loader.cpp` is it's not returning only the FTYPE

For instance LLAMA_FTYPE_MOSTLY_Q5_K_M returns "Q5_K - Medium"

and `LLAMA_FTYPE_MOSTLY_IQ2_XXS` returns "IQ2_XXS - 2.0625 bpw"

So I wouldn't want to make it identical either

There's also another similar table in `quantize.cpp` `QUANT_OPTIONS`, which includes the name, FTYPE, and a description

The only less-ideal part of that is that it includes aliases, like `Q4_K` and `Q4_K_M` both map to `LLAMA_FTYPE_MOSTLY_Q4_K_M`

So I'm not sure if there's an ideal solution here unfortunately. I could take it out of `llama-quant.cpp` and put it in my test code only, since I'm the only one currently who "needs" this table/conversion

src/llama-quant.cpp
bartowski1182: Interesting, this raises the question of why does `ggml_get_name` even exist if it just returns `tensor->name` ... but yeah I can make that change

src/llama-quant.h
```diff
@@ -1 +1,99 @@
 #pragma once
```
bartowski1182, line 1: @ggerganov probably worth grabbing your opinion here

I'm adding a LOT to this header file, including `<regex>`, `<string>`, `<vector>`, and `llama-arch.h`

Is this acceptable, or should I avoid making this a C++-header? The alternative of course would be to duplicate  `quantize_state_impl` in my test code and then change the header to only be exposing structs/functions that are needed

src/llama-quant.cpp
bartowski1182: in the meantime i've moved these to my test file

src/llama-quant.h
bartowski1182: @CISC in case you also have an opinion

CISC: Including regex here is unfortunate, I don't think you need to expose that though, make that part opaque and hidden inside `llama-quant.cpp`.

bartowski1182: Okay I think [a7132f6](https://github.com/ggml-org/llama.cpp/pull/20112/commits/a7132f6428eec8c861017ae64df207269e296a82) should accomplish what you suggested, no more regex in the header, using a unique pointer to a new struct `compiled_tensor_type_patterns` so that it can live in `llama-quant.cpp` exclusively

Relied on Claude's assistance for this but it looks like it follows similar practices from other sections of code that also use `std::unique_ptr` and compiling and unit tests both work still

CISC:
Rebase to fix CI.

Edit: Well, some of it at least.

CISC:
Looks like some kind of linkage error on Windows.

tests/test-quant-type-selection.cpp
```diff
@@ -0,0 +1,574 @@
...
+        t = ggml_new_tensor_4d(ctx.get(), GGML_TYPE_F32, ne0, ne1, ne2, ne3);
+    } else if (ne2 > 1) {
+        t = ggml_new_tensor_3d(ctx.get(), GGML_TYPE_F32, ne0, ne1, ne2);
+    } else {
+        t = ggml_new_tensor_2d(ctx.get(), GGML_TYPE_F32, ne0, ne1);
+    }
+    ggml_set_name(t, name.c_str());
+    return { std::move(ctx), t };
+}
+
```
ggerganov: Not sure why we want to create mock tensors with individual contexts. Just create a single ggml context with all tensors in it.

tests/test-quant-type-selection.cpp
```diff
@@ -0,0 +1,574 @@
...
+// Determine llm_type from metadata.
+// Only LLM_TYPE_70B matters -> probably can/should be dropped in the future
+static llm_type infer_llm_type(llm_arch arch, const gguf_remote_model & remote) {
+    if (arch == LLM_ARCH_LLAMA && remote.n_layer == 80 && remote.n_head != remote.n_head_kv) {
+        return LLM_TYPE_70B;
+    }
+    return LLM_TYPE_UNKNOWN;
+}
+
```
ggerganov: Try to avoid exposing `llm_arch` outside of the `libllama`. This will help to avoid exposing some internal headers, like `llama-arch.h`

src/llama-quant.h
```diff
@@ -1 +1,75 @@
...
+
+    // used to figure out if a model has tied embeddings (tok_embd shares weights with output)
+    bool has_tied_embeddings = true; // assume tied until we see output.weight
+
+    // tensor type override patterns (compiled once, used in llama_tensor_get_type)
+    std::unique_ptr<compiled_tensor_type_patterns> tensor_type_patterns;
+
+    llama_quant(const llama_model & model, const llama_model_quantize_params * params);
+    ~llama_quant();
+};
```
ggerganov: The idea is this to be the "qauntization context". Some time in the future, we can think about exposing it to the user through the public API.

For now, access it in a minimal way through `llama-ext.h`

ggerganov (commented):
The general idea is OK, but the implementation is not good. The AI tends to expose too many internal details, use inconsistent names and add extra logic that seems unnecessary. We'll need to fix that before merging.

I've added some TODOs in the code to get you started. Basically, try to minimize the internal things that are exposed outside of `libllama`.

For example, this test at the end of the day, generates a list of tensors names and tensor types. So what I would look to achieve is to have a single `llama-ext.h` function that gives us exactly this information and encapsulate all the logic in the `libllama` implementation. Currently, we "spill" too many details which complicates the "ext" interface and increases the technical debt.

bartowski1182:
Appreciate the feedback, this will be super helpful

Not sure I can blame it all on the AI unfortunately, some of that is going to be my lack of experience, but this gives me some important context going forward

In particular I think I'll need to purposefully study the `libllama` line, it wouldn't have occured to me exposing `llm-arch.h` here was a bad thing

ggerganov:
Yes, the ideal state is like this:

- The new test only includes `llama-ext.h`
- The `llama-ext.h` does not include any additional internal headers

The purpose of `llama-ext.h` is to be a stopgap solution for things that are not available through the public API yet. It must be very minimal because eventually we don't want to have it at all. So the less logic we add here, the easier it will be to remove it later.

bartowski1182:
Okay @ggerganov I think (hope) this should be in a much more acceptable state

Everything I need is exposed through `llama-ext.h`, no changes to `llama-quant.h`

back to being minimal changes to `llama-quant.cpp`, added all the helpers at the bottom

Could be argued that `llama_quant_compute_types` should be brought in to the main loop? But that feels more disruptive, can consider it if you think that would be appropriate

tests now only include public APIs (assuming I understand correctly that anything under `*/include/*` is "public") and `llama-ext.h`

Appreciate your pointers and your initial "deslop" commit :D was a very helpful jumping-off point to understand what your issues were with the implementation, and thankfully in hindsight it makes a lot of sense

bartowski1182:
Just a quick poke @ggerganov in case this one slipped through the cracks

ggerganov:
Thanks for the ping. From very quick look this seems much better.

Is this good for merge on your end?

bartowski1182:
Yup I'm ready! Thankfully noticed a funny ordering bug but it's fixed now

Do you want me to remove the generated schemas or leave them in the repo?

ggerganov (approved):

tests/test-quant-type-selection.cpp
```diff
@@ -0,0 +1,519 @@
...
+    { "IQ2_M",     LLAMA_FTYPE_MOSTLY_IQ2_M     },
+    { "IQ3_XXS",   LLAMA_FTYPE_MOSTLY_IQ3_XXS   },
+    { "IQ3_XS",    LLAMA_FTYPE_MOSTLY_IQ3_XS    },
+    { "IQ3_S",     LLAMA_FTYPE_MOSTLY_IQ3_S     },
+    { "IQ3_M",     LLAMA_FTYPE_MOSTLY_IQ3_M     },
+    { "IQ4_NL",    LLAMA_FTYPE_MOSTLY_IQ4_NL    },
+    { "IQ4_XS",    LLAMA_FTYPE_MOSTLY_IQ4_XS    },
+    { "TQ1_0",     LLAMA_FTYPE_MOSTLY_TQ1_0     },
+    { "TQ2_0",     LLAMA_FTYPE_MOSTLY_TQ2_0     },
+    { "MXFP4_MOE", LLAMA_FTYPE_MOSTLY_MXFP4_MOE },
```
CISC, line 58:
```suggestion
    { "MXFP4_MOE", LLAMA_FTYPE_MOSTLY_MXFP4_MOE },
    { "NVFP4",     LLAMA_FTYPE_MOSTLY_NVFP4     },
```

CISC (approved):

tests/test-quant-type-selection.cpp
bartowski1182: will regenerate with the NVFP4 target, then will be good to merge

bartowski1182: oh nvm right I don't pull from that list

tested again to be sure and everything still passes (as expected)

CISC: Do you need/want to make a NVFP4 GGUF?

CISC: https://huggingface.co/nvidia/models?search=nvfp4

bartowski1182: I just iterate all possible ftype targets, if it doesn't have a supported value in `llama_ftype_get_default_type` from `llama-quant.cpp` I just skip it, so at this time it won't be included

CISC (approved):

tests/gguf-model-data.h
```diff
@@ -40,7 +40,8 @@ struct gguf_remote_model {
 std::optional<gguf_remote_model> gguf_fetch_model_meta(
     const std::string & repo,
     const std::string & quant = "Q8_0",
-    const std::string & cache_dir = "");  // empty = default
+    const std::string & cache_dir = "",  // empty = default
+    bool verbose = true);
 
 gguf_context_ptr gguf_fetch_gguf_ctx(
```
CISC, line 46: Ooops, this needs `verbose` too, as it call `fetch_or_cached`.

bartowski1182: yeah marked as draft since I noticed I didn't fix all merge conflicts, will reopen in a sec :)

bartowski1182: okay fixed, built, tested both mine and 0cc4m's script and both are good 👍

CISC (approved):

PR #20117, opened 2026-03-04
llama-quant : left-align tensor names in output

ddh0:
Left-align the tensor names printed in the output of `llama-quantize`.

Before:

Screenshot 2026-03-04 at 3 53 03 PM

After:

Screenshot 2026-03-04 at 3 53 54 PM

---

Now the layer names `blk.%d` are all vertically aligned. Easier to read IMO.


ddh0:
Is this change welcome? In my opinion it's a small change that makes the quantization output a little cleaner and more readable, since your eyes can just follow the column of layer indices down as it runs.

ggerganov (approved):
Yes, don't hesitate to ping if a PR gets stale.

PR #19780, opened 2026-02-21
[WIP] ggml-hexagon: convert f32 to f16 - fa opt part4

chraac:
### Dot Product Function Improvements

* Replaced the previous `hvx_dot_f16_f16_aa_rx2` function with new, more parallelized `hvx_dot_f16_f16_aa_rx4` and `hvx_dot_f16_f16_aa_rx32` functions in `flash-attn-ops.c`, allowing computation of 4 and 32 dot products at a time, respectively. This increases throughput and simplifies the code by leveraging vectorization.
* Updated the main attention kernel (`flash_attn_ext_f16_thread`) to use the new `hvx_dot_f16_f16_aa_rx32` function, replacing the looped calls to the old function and removing the need for temporary arrays.

### Vector Reduction Utilities

* Added `hvx_vec_reduce_sum_f32x4` utility in `hvx-reduce.h` for both HVX architectures, enabling efficient reduction of four HVX vector results into a single vector. This supports the new parallel dot product functions. [[1]](diffhunk://#diff-a61b8b4ec9b687ceb6adecb4f2de734f398493514475aa35a2ed1697d58e8a78R49-R63) [[2]](diffhunk://#diff-a61b8b4ec9b687ceb6adecb4f2de734f398493514475aa35a2ed1697d58e8a78R90-R104)

## Performance

// TODO

PR #20118, opened 2026-03-04
hexagon: Flash Attention optimizations (dma, mpyacc, multi-row) and MatMul updates

max-krasnyansky:
Further updates on top of #19780 by @chraac 

- Improved DMA pipelining in FA
- Reduced FA block size from 128 to 64 to improve DMA prefetch (128 is too big for most models)
- Improved usage or vmpyacc intrinsics in dot products

Some quick perf numbers on S25+ (Gen4) with Llama3.2-3B-Q4_0 and FA on Hexagon

```
Before:
common_perf_print: prompt eval time =    2711.51 ms /   205 tokens (   13.23 ms per token,    75.60 tokens per second)
common_perf_print:        eval time =    3934.46 ms /    63 runs   (   62.45 ms per token,    16.01 tokens per second)

After:
common_perf_print: prompt eval time =    2538.64 ms /   205 tokens (   12.38 ms per token,    80.75 tokens per second)
common_perf_print:        eval time =    3586.00 ms /    63 runs   (   56.92 ms per token,    17.57 tokens per second)
```

> Original notes from #19780
> ### Dot Product Function Improvements
> Replaced the previous hvx_dot_f16_f16_aa_rx2 function with new, more parallelized hvx_dot_f16_f16_aa_rx4 and
> hvx_dot_f16_f16_aa_rx32 functions in flash-attn-ops.c, allowing computation of 4 and 32 dot products at a time,
> respectively. This increases throughput and simplifies the code by leveraging vectorization.
> Updated the main attention kernel (flash_attn_ext_f16_thread) to use the new hvx_dot_f16_f16_aa_rx32 function,
> replacing the looped calls to the old function and removing the need for temporary arrays.
> 
> ### Vector Reduction Utilities
> Added hvx_vec_reduce_sum_f32x4 utility in hvx-reduce.h for both HVX architectures, enabling efficient reduction 
> of four HVX vector results into a single vector. This supports the new parallel dot product functions.

### DIFF
--- a/ggml/src/ggml-hexagon/htp/hvx-base.h
+++ b/ggml/src/ggml-hexagon/htp/hvx-base.h
@@ -170,4 +170,27 @@ static inline HVX_Vector hvx_vec_i16_from_hf_rnd_sat(HVX_Vector vin) {
     return Q6_Vh_vround_VwVw_sat(vsf_1, vsf_0);
 }
 
+#if __HVX_ARCH__ < 79
+
+static inline HVX_VectorPair hvx_vec_mpyacc_f32_f16(HVX_VectorPair acc, HVX_Vector x, HVX_Vector y)
+{
+    HVX_Vector a0 = Q6_V_lo_W(acc);
+    HVX_Vector a1 = Q6_V_hi_W(acc);
+
+    HVX_VectorPair m = Q6_Wqf32_vmpy_VhfVhf(x, y);
+    a0 = Q6_Vsf_equals_Vqf32(Q6_Vqf32_vadd_Vqf32Vsf(Q6_V_lo_W(m), a0));
+    a1 = Q6_Vsf_equals_Vqf32(Q6_Vqf32_vadd_Vqf32Vsf(Q6_V_hi_W(m), a1));
+
+    return Q6_W_vcombine_VV(a0, a1);
+}
+
+#else
+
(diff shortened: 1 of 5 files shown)


lhez (approved):

ggml/src/ggml-hexagon/htp/hvx-base.h
```diff
@@ -38,7 +38,7 @@ static inline HVX_Vector hvx_vec_splat_f32(float v) {
     return Q6_V_vsplat_R(u.i);
 }
 
-static inline HVX_Vector hvx_vec_splat_f16(float v) {
+static inline HVX_Vector hvx_vec_splat_f16(_Float16 v) {
```
chraac, line 41: nit: better to keep the same type (`__fp16`) as union below.

max-krasnyansky: __fp16 can't be used as a function argument, would have to be a pointer.
_Float16 can. That's pretty much the only reason I used it.

chraac: From gcc [doc](https://gcc.gnu.org/onlinedocs/gcc/Half-Precision.html), there's a minor different between `__fp16` and `_Float16` in some arch, but both okay in this case

PR #20120, opened 2026-03-05
server : preserve anthropic thinking blocks in conversion

T0mSIlver:
Fix Anthropic `/v1/messages` conversion to preserve assistant `thinking` blocks as `reasoning_content` when converting to internal OpenAI-compatible chat messages.

Fixes #20090.

AI usage disclosure:
- AI was used in an assistive role for code review suggestions, small implementation adjustments, and command execution support.
- I manually reviewed the final patch and validation results.

### DIFF
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -1463,6 +1463,7 @@ json convert_anthropic_to_oai(const json & body) {
             json tool_calls = json::array();
             json converted_content = json::array();
             json tool_results = json::array();
+            std::string reasoning_content;
             bool has_tool_calls = false;
 
             for (const auto & block : content) {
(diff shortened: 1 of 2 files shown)


ngxson (changes requested):
please add a test case with `/apply-template` to prove that this works

T0mSIlver:
Added two test cases:

- test_anthropic_thinking_history_in_count_tokens : sends Anthropic-format messages with interleaved thinking + tool use through /v1/messages/count_tokens and verifies thinking blocks increase the token count (i.e. they're not silently dropped by convert_anthropic_to_oai).
- test_anthropic_thinking_history_in_template : uses /apply-template with the Qwen3 template to verify the converted reasoning_content renders inside <think> tags in the prompt.

Both use the Qwen3 template which natively handles reasoning_content.

florianbrede-ayet:
@T0mSIlver just wanted to open a PR for the same issue when I came across yours. I've been struggling with the same problems and qwen35 models. The core fix is basically identical to mine - i compiled it against ROCm and it solves the issue with claude code.

On a side note, it also fixes the autoparser branch where you'd previously get `500 {"error":{"code":500,"message":"Failed to parse input at pos 67: ","type":"server_error"}}` (or other low positions within malformed thinking blocks) after some turns.

ngxson (approved):

PR #20126, opened 2026-03-05
model : fix Qwen3.5 model type detection

EZForever:
Now that all different sizes of Qwen3.5 models have came out, we can check the model cards and correct the (cosmetic) model type detection logic. For example, Qwen3.5 does not have a 80BA3B variant, 35BA3B have 40 layers instead of 28, and both 0.8B and 2B have 24 layers.

This change should be purely cosmetic, fixes "?B" model names shown by `llama-bench`, etc.

### DIFF
--- a/src/llama-model.cpp
+++ b/src/llama-model.cpp
@@ -61,6 +61,7 @@ const char * llm_type_name(llm_type type) {
         case LLM_TYPE_0_3B:          return "0.3B";
         case LLM_TYPE_0_5B:          return "0.5B";
         case LLM_TYPE_0_6B:          return "0.6B";
+        case LLM_TYPE_0_8B:          return "0.8B";
         case LLM_TYPE_1B:            return "1B";
         case LLM_TYPE_1_2B:          return "1.2B";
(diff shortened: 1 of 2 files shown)


src/llama-model.cpp
```diff
@@ -2528,7 +2531,9 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                 }
 
                 switch (hparams.n_layer) {
-                    case 24: type = LLM_TYPE_2B; break;
+                    case 24: hparams.n_embd == 1024 ? LLM_TYPE_0_8B : LLM_TYPE_2B; break;
+                    case 32: hparams.n_embd == 2560 ? LLM_TYPE_4B : LLM_TYPE_9B; break;
```
CISC:
```suggestion
                    case 24: type = hparams.n_embd == 1024 ? LLM_TYPE_0_8B : LLM_TYPE_2B; break;
                    case 32: type = hparams.n_embd == 2560 ? LLM_TYPE_4B : LLM_TYPE_9B; break;
```

CISC (approved):

PR #20150, opened 2026-03-06
ggml-blas: report memory when calling ggml_backend_blas_device_get_memory.

pwhelan:
Implement memory information retrieval for ggml_backend_blas_device_get_memory. It uses sysinfo on linux, which I tested.

I copied the windows version from #18578. I have no desire to test it on windows and the implementation looks sane. I am also assuming the original author was more familiar with windows.

My reason for fixing this bug is to allow for light weight testing of code that requires querying the currently used memory of a device using BLAS on my laptop.

### DIFF
--- a/ggml/src/ggml-blas/ggml-blas.cpp
+++ b/ggml/src/ggml-blas/ggml-blas.cpp
@@ -6,6 +6,8 @@
 #include <vector>
 #include <cstring>
 
+#include <sys/sysinfo.h>
+
 #if defined(GGML_BLAS_USE_ACCELERATE)
 #   include <Accelerate/Accelerate.h>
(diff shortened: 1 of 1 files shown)


taronaeo:
Unfortunately its not a bug. The BLAS backend uses host memory. Setting `*free = 0; *total = 0;` simply lets the CPU backend report the memory information instead and no code duplication is required across backend implementations.

See: https://github.com/ggml-org/llama.cpp/blob/f7db3f37895ad3a98405afdb84b42ed462542b12/src/llama.cpp#L118-L127

taronaeo:
Closing as not a bug. Feel free to reopen if otherwise.

Issue #19213, opened 2026-01-30
Feature Request: CI/CD systems for llama.cpp+Vulkan


bbharti:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

There is lot of active development happening on llama.cpp+Vulkan. It seems like there is no CI/CD system for Intel platforms. We do have latest Intel platforms and can be used for this effort. How can we move forward?

### Motivation

llama.cpp+Vulkan is very important for many developers. Great to see community is actively improving it. However, we are finding that many times there is regression or failure on Intel platforms. It would be great to catch them as early as possible and provide a stable release to the users.

### Possible Implementation

We can provide access to Intel Hardware. 
@ggerganov, @jeffbolznv and @0cc4m please let me know how we can proceed.

virajwad:
We initially saw vulkan perf regression when going from **b7657 --> b7885** on one of our platforms

load_backend: loaded RPC backend from C:\Users\dungeon\Desktop\llamacpp\llama-b7657-bin-win-vulkan-x64\ggml-rpc.dll
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = Intel(R) Graphics <driver redacted> (Intel Corporation) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: KHR_coopmat
load_backend: loaded Vulkan backend from C:\Users\dungeon\Desktop\llamacpp\llama-b7657-bin-win-vulkan-x64\ggml-vulkan.dll
load_backend: loaded CPU backend from C:\Users\dungeon\Desktop\llamacpp\llama-b7657-bin-win-vulkan-x64\ggml-cpu-alderlake.dll
| model                          |       size |     params | backend    | ngl |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | --------------: | -------------------: |
| qwen3 8B Q4_K - Medium         |   4.68 GiB |     8.19 B | Vulkan     |  99 |           pp512 |        954.77 ± 7.24 |

[...]

PR #20154, opened 2026-03-06
ci: Setup self-hosted CI for Intel Linux Vulkan backend

rillomas:
We would like add a single self-hosted CI runner for Intel Linux Vulkan backend related to #19213.
This will be focused for Vulkan backend only at the moment, but when OpenVINO backend gets merged we will use it for both backends.

An example workflow execution on the branch can be viewed at the following:
 - https://github.com/rillomas/llama.cpp/actions/runs/22884633443/job/66394545350


rillomas:
~~Hi @ggerganov we are ready to add an instance for Intel Vulkan backend. Will you please send me a runner token? I am located in Japan so hopefully your TZ is close to mine (since the runner tokens only last for an hour)~~

Update: I was able to receive a token and confirm CI instance added

ggerganov:
Thanks, the workflow appears to run successfully: https://github.com/ggml-org/llama.cpp/actions/runs/22884579987/job/66394392012?pr=20154

@0cc4m @jeffbolznv FYI this is a coopmat runner

ggerganov (approved):

CISC (approved):

rillomas:
> Thanks, the workflow appears to run successfully: https://github.com/ggml-org/llama.cpp/actions/runs/22884579987/job/66394392012?pr=20154

FYI this specific workflow was run on another instance which was already added by the OpenVINO team (LNL). Vulkan workload will work on both instances but I haven't tested if OpenVINO workload works on the instance I added today (PTL)

ggerganov:
Thanks for pointing that out. If the OpenVINO workflows don't run on your runner, we will gate them with an extra tag.

0cc4m (approved):

0cc4m:
The runners don't seem to be working currently, please check what is going on.

CISC:
> The runners don't seem to be working currently, please check what is going on.

They are currently offline for maintenance, see https://github.com/ggml-org/llama.cpp/pull/29009#issuecomment-5708115690

PR #20157, opened 2026-03-06
ggml: update comments for backends which have no memory to report

taronaeo:
Ref: #20150

This PR updates the comments for backends which have no memory to report. By default, llama.cpp will default to host memory information if the backend reports `*free = 0; *total = 0;`.

This will help maintainers understand that reporting 0 bytes is not a bug, but rather, allow downstream GGML consumers to decide what they want to do in the event that the backend reports 0 bytes.


lhez (approved):

taronaeo:
`CI / ggml-ci-x64-cpu-high-perf (pull_request)` failure unrelated to this PR. Merging.

PR #20158, opened 2026-03-06
server: Introduce LLAMA_BUILD_WEBUI build flag to allow disabling the embedded web ui

kushagharahi:
Allows users to build a lighter binary for those who don't need the web ui embedded but still want a llama-server

Usage: Add `-DLLAMA_BUILD_WEBUI=OFF` to the list of build flags or set `useWebUi = false;` in package override for nix.

resolves #10986


CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ option(LLAMA_BUILD_TESTS    "llama: build tests"          ${LLAMA_STANDALONE})
 option(LLAMA_BUILD_TOOLS    "llama: build tools"          ${LLAMA_STANDALONE})
 option(LLAMA_BUILD_EXAMPLES "llama: build examples"       ${LLAMA_STANDALONE})
 option(LLAMA_BUILD_SERVER   "llama: build server example" ${LLAMA_STANDALONE})
+option(LLAMA_BUILD_WEBUI    "llama: build the embedded Web UI for server"  ${LLAMA_STANDALONE})
```
ngxson: it should be set to ON by default

kushagharahi: Updated!

ngxson (approved):
@ggerganov could you give the 2nd approval? Thanks!

ggerganov (approved):

PR #20168, opened 2026-03-06
model : add support for Phi4ForCausalLMV

dranger003:
Add support for [microsoft/Phi-4-reasoning-vision-15B](https://huggingface.co/microsoft/Phi-4-reasoning-vision-15B).
This is reusing the existing Phi-3 text path for the decoder and adds a mmproj/mtmd path for Phi SigLIP2 vision.

I uploaded some converted weights for testing https://huggingface.co/dranger003/Phi-4-reasoning-vision-15B-GGUF.
The model generates coherent text and proper image descriptions, including OCR (see below).

DISCLAIMER: GPT-5.4 was used to help with this PR.

image

image

EDIT: More information about this model here
https://www.microsoft.com/en-us/research/blog/phi-4-reasoning-vision-and-the-lessons-of-training-a-multimodal-reasoning-model/

> Several open-source multimodal language models have adapted their methodologies accordingly, e.g., Gemma3(opens in new tab) uses pan-and-scan and NVILA(opens in new tab) uses Dynamic S2. However, their trade-offs are difficult to understand across different datasets and hyperparameters. To this end, we conducted an ablation study of several techniques. We trained a smaller 5 billion parameter Phi-4 based proxy model on a dataset of 10 million image-text pairs, primarily composed of computer-use and GUI grounding data. We compared with Dynamic S2, which resizes images to a rectangular resolution that minimizes distortion while admitting a tiling by 384×384 squares; Multi-crop, which splits the image into potentially overlapping 384×384 squares and concatenates their encoded features on the token dimension; Multi-crop with S2, which broadens the receptive field by cropping into 1536×1536 squares before applying S2; and Dynamic resolution using the Naflex variant of SigLIP-2, a natively dynamic-resolution encoder with adjustable patch counts.

> Our primary finding is that dynamic resolution vision encoders perform the best and especially well on high-resolution data. It is particularly interesting to compare dynamic resolution with 2048 vs 3600 maximum tokens: the latter roughly corresponds to native HD 720p resolution and enjoys a substantial boost on high-resolution benchmarks, particularly ScreenSpot-Pro. Reinforcing the high-resolution trend, we find that multi-crop with S2 outperforms standard multi-crop despite using fewer visual tokens (i.e., fewer crops overall). The dynamic resolution technique produces the most tokens on average; due to their tiling subroutine, S2-based methods are constrained by the original image resolution and often only use about half the maximum tokens. From these experiments we choose the SigLIP-2 Naflex variant as our vision encoder.


tools/mtmd/models/phi4-siglip.cpp
ngxson, line 1: can this be merged to `siglip.cpp` ?

gguf-py/gguf/constants.py
```diff
@@ -3869,6 +3869,7 @@ class VisionProjectorType:
     GEMMA3 = "gemma3"
     GEMMA3NV = "gemma3nv"
     GEMMA3NA = "gemma3na"
+    PHI4_SIGLIP = "phi4_siglip"
```
ngxson: just call it `phi4` is enough


```suggestion
    PHI4 = "phi4"
```

tools/mtmd/models/phi4-siglip.cpp
dranger003: Looking at the diff I don't see how this would help. I mean it's not impossible but it wouldn't be clean since the current abstraction gap is rather a large one. I think the better outcome is to keep them separate?

gguf-py/gguf/constants.py
ngxson: @dranger003  I assume you are joking, but do you understand my suggestion means that you must update it in other places in the code?

tools/mtmd/models/phi4-siglip.cpp
ngxson: I removed this entire file and added 8 lines of code into `siglip.cpp`: https://github.com/ggml-org/llama.cpp/pull/20168/commits/87d77145d493f05fbca6958be40f3595e1274d29

Not sure if you refer this 8 lines of code as "abstraction gap is rather a large one"

gguf-py/gguf/constants.py
dranger003: Yes, I clicked accept to mark this as resolved. I just pushed the rename across the PR.

tools/mtmd/models/phi4-siglip.cpp
dranger003: Thanks, that's on me then... the code has been updated.

convert_hf_to_gguf.py
```diff
@@ -4905,6 +4905,55 @@ def set_gguf_parameters(self):
...
+            and bytelevel.get("type") == "ByteLevel"
+            and bytelevel.get("add_prefix_space") is False
+            and bytelevel.get("trim_offsets") is True
+            and bytelevel.get("use_regex") is False
+            and isinstance(decoder, dict)
+            and decoder.get("type") == "ByteLevel"
+            and decoder.get("add_prefix_space") is True
+            and decoder.get("trim_offsets") is True
+            and decoder.get("use_regex") is True
+        )
```
CISC: Why all this?

convert_hf_to_gguf.py
```diff
@@ -5077,6 +5077,141 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
         yield (self.format_tensor_name(gguf.MODEL_TENSOR.ROPE_FACTORS_SHORT), torch.tensor(short_factors, dtype=torch.float32))
 
 
+@ModelBase.register("Phi4ForCausalLMV")
+class Phi4VisionTextModel(Phi3MiniModel):
+    model_arch = gguf.MODEL_ARCH.PHI3
+
+    def set_vocab(self):
+        super().set_vocab()
+        self.gguf_writer.add_tokenizer_pre("phi4")
```
```suggestion
```
Ok, better, but why are you forcing it to `phi4`?

I'm guessing it matches some other vocab, but do the settings somehow differ, or is it purely cosmetic? If the latter I would suggest to just use the default.

tools/mtmd/clip.cpp
```diff
@@ -3340,6 +3417,8 @@ int clip_n_output_tokens_x(const struct clip_ctx * ctx, struct clip_image_f32 *
     const int n_total = clip_n_output_tokens(ctx, img);
     const auto & proj = ctx->proj_type();
     switch (proj) {
+        case PROJECTOR_TYPE_PHI4:
+            return img->nx / params.patch_size;
```
ngxson: remove this, it's only used by M-RoPE models

tools/mtmd/clip.cpp
```diff
@@ -3020,6 +3072,31 @@ bool clip_image_preprocess(struct clip_ctx * ctx, const clip_image_u8 * img, str
...
+                GGML_ASSERT(params.image_max_pixels % patch_area == 0);
+
+                const int min_num_patches = params.image_min_pixels / patch_area;
+                const int max_num_patches = params.image_max_pixels / patch_area;
+                int num_patches = std::max((original_size.height / patch_size) * (original_size.width / patch_size), 1);
+                if (num_patches < min_num_patches) {
+                    num_patches = min_num_patches;
+                } else if (num_patches > max_num_patches) {
+                    num_patches = max_num_patches;
+                }
```
ngxson: unless I misread something, this seems to be an overcomplicated version of "smart_resize", which is essentially the same logic as `PROJECTOR_TYPE_PIXTRAL`

the logic is simply: if the total area (or the number of pixels) is more than max_pixels, we resize such that the new size fits into max_pixels, while preserving ratio. The size must also be aligned to patch size.

if that's the case, please remove this whole thing and reuse the same code as `PROJECTOR_TYPE_PIXTRAL`

tools/mtmd/clip.cpp
```diff
@@ -3357,6 +3436,8 @@ int clip_n_output_tokens_y(const struct clip_ctx * ctx, struct clip_image_f32 *
     const auto & params = ctx->model.hparams;
     const auto & proj = ctx->proj_type();
     switch (proj) {
+        case PROJECTOR_TYPE_PHI4:
+            return img->ny / params.patch_size;
```
ngxson: remove this

convert_hf_to_gguf.py
```diff
@@ -5077,6 +5077,137 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+
+            new_name = name.replace("model.vision_tower.vision_tower.", "vision_tower.")
+
+            if ".vision_model.post_layernorm." in new_name:
+                return
+
+            layer_prefix = "vision_tower.vision_model.encoder.layers."
+            if new_name.startswith(layer_prefix):
+                layer_suffix = new_name[len(layer_prefix):]
+                layer_idx = layer_suffix.split(".", maxsplit=1)[0]
```
ngxson: use can just use `bid` (block ID)

ngxson (changes requested):
I believe this can be simplify further. Phi-4 siglip is not very breakthrough in term of architecture, it reuse a lot of things already existing in the code base, so there is no need to add too many new code paths for it.

tools/mtmd/clip.cpp
ngxson: ?

ngxson: Ok since you cannot provide me a response, I asked Claude, it confirms the code is redundant and also helped me to remove it.

Since you claimed that GPT-5.4 is used in writing this PR, I suppose it's nowhere near Claude then.

ngxson (approved):
I'm approving this PR because this model is too trivial to support.

I can be a bit harsh here, but I want to make it clear: I don't recommend contributions where the author cannot properly respond to trivial questions (proof: https://github.com/ggml-org/llama.cpp/pull/20168#discussion_r2896693520 and https://github.com/ggml-org/llama.cpp/pull/20168#discussion_r2898423361). This shows the author put too little effort into their own work.

Many other contributors are willing to spend time understanding the code, even when it's AI-generated, and we welcome that kind of contribution. What we don't encourage is the type of PR where more than half of the work ends up being done by the reviewers.

convert_hf_to_gguf.py
```diff
@@ -5077,6 +5077,136 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
         yield (self.format_tensor_name(gguf.MODEL_TENSOR.ROPE_FACTORS_SHORT), torch.tensor(short_factors, dtype=torch.float32))
 
 
+@ModelBase.register("Phi4ForCausalLMV")
+class Phi4VisionTextModel(Phi3MiniModel):
+    model_arch = gguf.MODEL_ARCH.PHI3
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        if name.startswith(("model.vision_tower.", "vision_tower.", "model.mm_projector.", "mm_projector.")):
+            return
```
CISC: Might as well just move this to `Phi3MiniModel` now.

convert_hf_to_gguf.py
```diff
@@ -5077,6 +5077,136 @@ def generate_extra_tensors(self) -> Iterable[tuple[str, Tensor]]:
...
+
+            if bid is not None and bid == self.vision_last_layer_idx:
+                return
+
+            if new_name.endswith("vision_model.embeddings.patch_embedding.weight"):
+                if data_torch.ndim != 2:
+                    raise ValueError(f"Unexpected Phi-4 patch embedding shape: {tuple(data_torch.shape)}")
+
+                if self.hparams_vision is None:
+                    raise ValueError("hparams_vision is not set")
```
```suggestion
                assert self.hparams_vision is not None
                if data_torch.ndim != 2:
                    raise ValueError(f"Unexpected Phi-4 patch embedding shape: {tuple(data_torch.shape)}")
```

CISC (approved):

tools/mtmd/clip.cpp
dranger003: Both functions have a different implementation as far as I can see. The phi path appears to be targetting a tile count while the pixtral approach is trying to fit the image and doesn't care about tile count and I thought this would lead to some edge cases. Understood if you're saying it doesn't matter or if my understanding is off.

dranger003:
> I'm approving this PR because this model is too trivial to support.
> 
> I can be a bit harsh here, but I want to make it clear: I don't recommend contributions where the author cannot properly respond to trivial questions (proof: [#20168 (comment)](https://github.com/ggml-org/llama.cpp/pull/20168#discussion_r2896693520) and [#20168 (comment)](https://github.com/ggml-org/llama.cpp/pull/20168#discussion_r2898423361)). This shows the author put too little effort into their own work.
> 
> Many other contributors are willing to spend time understanding the code, even when it's AI-generated, and we welcome that kind of contribution. What we don't encourage is the type of PR where more than half of the work ends up being done by the reviewers.

Thanks for your honesty @ngxson. I was genuinely trying to help and I understand this is likely wasting your time more than if I didn't contribute to the project using AI to help me.

xyehya:
Working now?

dranger003:
> Working now?

Yes, still working here.

PR #20172, opened 2026-03-06
tools: Implement --output flag on cli to save response

### LINKED ISSUE #19256: Misc. bug: Llama-cli saving output to file using ">" is not working anymore

dandm1:
Adjustments to the llama-cli to support the --output flag. When the flag is specified the output will be written to a file as well as to the screen. When specified in conjunction with --special the raw output will be written.

### DIFF
--- a/tools/cli/cli.cpp
+++ b/tools/cli/cli.cpp
@@ -127,10 +127,8 @@ struct cli_context {
             console::set_display(DISPLAY_TYPE_RESET);
         }
 
-        append_file_out(
-            "[Prompt]: " + messages.back()["content"].get<std::string>() + "\n\n",
-            chat_params.prompt
-        );
+        append_file_out("[Prompt]: " + messages.back()["content"].get<std::string>() + "\n\n", false);
+        append_file_out(chat_params.prompt, true);
 
         // wait for first result
         console::spinner::start();
@@ -162,24 +160,27 @@ struct cli_context {
                         if (is_thinking) {
                             console::log("\n[End thinking]\n\n");
                             console::set_display(DISPLAY_TYPE_RESET);
-                            append_file_out("\n\n", "</think>");
+                            append_file_out("\n\n", false);
+                            append_file_out("</think>", true);
                             is_thinking = false;
                         }
                         curr_content += diff.content_delta;
                         console::log("%s", diff.content_delta.c_str());
                         console::flush();
                         if (!content_started) {
-                            append_file_out("[Assistant]: ", "");
+                            append_file_out("[Assistant]: ", false);
                             content_started = true;
                         }
-                        append_file_out(diff.content_delta);
+                        append_file_out(diff.content_delta, false);
+                        append_file_out(diff.content_delta, true);
                     }
                     if (!diff.reasoning_content_delta.empty()) {
                         console::set_display(DISPLAY_TYPE_REASONING);
                         std::string reasoning_delta = diff.reasoning_content_delta;
                         if (!is_thinking) {
                             console::log("[Start thinking]\n");
-                            append_file_out("[Thinking]: ", "<think>");
+                            append_file_out("[Thinking]: ", false);
+                            append_file_out("<think>", true);
                             if (reasoning_delta == "<think>") {
                                 reasoning_delta = "";
                             }
@@ -187,14 +188,15 @@ struct cli_context {
                         is_thinking = true;
                         console::log("%s", reasoning_delta.c_str());
                         console::flush();
-                        append_file_out(reasoning_delta);
+                        append_file_out(reasoning_delta, false);
+                        append_file_out(reasoning_delta, true);
                     }
                 }
             }
             auto res_final = dynamic_cast<server_task_result_cmpl_final *>(result.get());
             if (res_final) {
                 out_timings = std::move(res_final->timings);
-                append_file_out("\n\n", "");
+                append_file_out("\n\n", false);
                 break;
             }
             result = rd.next(should_stop);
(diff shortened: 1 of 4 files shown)


tools/cli/cli.cpp
```diff
@@ -104,13 +107,26 @@ struct cli_context {
             console::set_display(DISPLAY_TYPE_RESET);
         }
 
+        // check if we are doing file output
+        bool file_streaming  = (file_out != nullptr && file_out->is_open());
+        if (file_streaming) {
+            if (defaults.special_characters) {
+                *file_out << chat_params.prompt;
```
ngxson: this pattern of checking `file_streaming` then dereferencing `*file_out` doesn't look clean

instead, abstract the `file_out` into a new function `append_file_out(const std::string & content)`, the function is a no-op if file is not set

ngxson (changes requested):

tools/cli/cli.cpp
dandm1: Yeah, good call. Have refactored with a append_file_out function, in which I also included logic to select between special tokens and plain text outputs, to further simplify.

tools/cli/cli.cpp
```diff
@@ -77,9 +81,10 @@ struct cli_context {
         verbose_prompt = params.verbose_prompt;
         reasoning_budget = params.reasoning_budget;
         reasoning_budget_message = params.reasoning_budget_message;
+        reasoning_format = params.reasoning_format;
     }
 
-    std::string generate_completion(result_timings & out_timings) {
+    std::string generate_completion(result_timings & out_timings, std::ofstream * file_to_use = nullptr) {
```
ngxson:
```suggestion
    std::string generate_completion(result_timings & out_timings) {
```

there is no need to use pointer this way. the file must be created as a member of `cli_context`

use std::optional to indicate if it's enabled ot not

ngxson (changes requested):

tools/cli/cli.cpp
dandm1: Made that change. I suspect file_out.has_value() is very slightly slower than a dedicated bool, but it's elegant.

tools/cli/cli.cpp
```diff
@@ -59,6 +59,8 @@ struct cli_context {
     bool verbose_prompt;
     int reasoning_budget = -1;
     std::string reasoning_budget_message;
+    common_reasoning_format reasoning_format;
+    std::optional<std::ofstream *> file_out = std::nullopt;
```
ngxson: `std::optional<std::ofstream>`, std optional already does the job as the pointer manager

ngxson:
sorry for the delay, @dandm1 can you rebase to latest master?

dandm1 (commented):
Will do

dandm1:
Squashed and rebased. @ngxson

tools/cli/cli.cpp
```diff
@@ -199,6 +221,18 @@ struct cli_context {
...
+            return;
+        }
+        if (defaults.special_characters && special_content.has_value()) {
+            file_out.value() << special_content.value();
+        } else {
+            file_out.value() << content;
+        }
+        file_out.value().flush();
+    }
+
```
ngxson: instead of passing the `special_content` as std::option, it is better to refactor this function like this:

```cpp
void append_file_out(const std::string & content, bool is_special)
```

and the caller need to call it twice:

```cpp
append_file_out("</think>", true);
append_file_out(content, false);
```

it makes the code easier to keep track

ngxson (changes requested):

tools/cli/cli.cpp
dandm1: OK @ngxson - updated on that basis. I don't much like the structure where I need two calls to write the same content to either output format - but I guess it's better that introducing a tri-state just for this or using std::optional<bool> .

tools/cli/cli.cpp
```diff
@@ -124,13 +127,17 @@ struct cli_context {
             console::set_display(DISPLAY_TYPE_RESET);
         }
 
+        append_file_out("[Prompt]: " + messages.back()["content"].get<std::string>() + "\n\n", false);
```
ngxson, line 130: that won't work correctly with multimodal input

ngxson (commented):
I'm not very comfortable with this implementation as it's not well-thought, better to do this with the upcoming CLI refactoring, cc @pwilkin

PR #20173, opened 2026-03-06
ggml webgpu: faster normal quant and some k-quant matrix operations, better shader parameter handling

reeselevine:
TLDR: The WebGPU implementation is pretty fast now for some quantization types, and hopefully it's more stable/works on a decent number of devices now. Try this code out in wllama here: https://reeselevine.github.io/wllama/. At least on my machine, it mostly seems to outperform WebLLM and OnnxRuntimeWeb (through transformers.js) on roughly equivalent models. And if there are any issues let's try to fix them!

* Adds faster matrix-matrix and matrix-vector multiplication for all normal (q4-q8) quantization types.
* Adds faster q6_k matrix-vector dequantization and decently faster matrix-matrix dequantization for all q_k types
* Pretty major improvement in shader parameter handling. I realized we can just use `queue.WriteBuffer` to avoid having an extra buffer to copy parameters from host to device, which removes blocking operations from the encode path.
* Also moved to using a single error buffer for `set_rows` and checking it at the end of each `graph_compute`, instead of having separate error buffers for each `set_rows` operation.
* Better separation of the GPU profiling timestamp buffers. This can be improved further but leaving it for now.
* Some other minor formatting cleanups.


ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_decls.tmpl
```diff
@@ -101,3 +101,675 @@ fn init_shmem_src0(thread_id: u32, batch_offset: u32, offset_m: u32, k_outer: u3
...
+        let hm = select(4.0, 0.0, (hmask_byte & m) != 0);
+        let qs_val = (q_byte >> shift) & 3u;
+
+        let q_val = (f16(qs_val) - f16(hm)) * dl;
+        shmem[elem_idx] = q_val;
+    }
+}
+
+#endif
+#ifdef INIT_SRC0_SHMEM_Q4_K
```
CISC: Some consistent separation between these would be nice, and possibly `// INIT_SRC0_SHMEM_Q3_K` type comments on the `#endif`s.

CISC (approved):

nikhilJain17 (approved):
The logic changes around future submissions and waiting lgtm!

PR #20177, opened 2026-03-06
Autoparser: True streaming

pwilkin:
In the final changes to the autoparser I modified the atomicity constraint to be pretty restrictive due to models such as GLM 4.7-Flash which put arguments directly after the function name, with no markers in between. Now I'm relaxing that constraint so people can look at their favorite assistant building the tool call live ;)

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -234,7 +234,7 @@ common_peg_parser analyze_tools::build_tool_parser_tag_json(parser_build_context
             call_id_section = p.optional(call_id.prefix + p.tool_id(p.until(call_id.suffix))) + call_id.suffix;
         }
 
-        auto func_parser = p.tool_open(p.atomic(function.name_prefix + p.tool_name(p.literal(name)) + function.name_suffix)) +
+        auto func_parser = p.tool_open(function.name_prefix + p.tool_name(p.literal(name)) + function.name_suffix) +
                            call_id_section + p.tool_args(p.schema(p.json(), "tool-" + name + "-schema", schema));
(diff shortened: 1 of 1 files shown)


aldehir:
These already wrap in `atomic`:
```
    common_peg_parser tool_open(const common_peg_parser & p) { return atomic(tag(TOOL_OPEN, p)); }
    common_peg_parser tool_close(const common_peg_parser & p) { return atomic(tag(TOOL_CLOSE, p)); }
    common_peg_parser tool_id(const common_peg_parser & p) { return atomic(tag(TOOL_ID, p)); }
    common_peg_parser tool_name(const common_peg_parser & p) { return atomic(tag(TOOL_NAME, p)); }
    common_peg_parser tool_arg_open(const common_peg_parser & p) { return atomic(tag(TOOL_ARG_OPEN, p)); }
    common_peg_parser tool_arg_close(const common_peg_parser & p) { return atomic(tag(TOOL_ARG_CLOSE, p)); }
    common_peg_parser tool_arg_name(const common_peg_parser & p) { return atomic(tag(TOOL_ARG_NAME, p)); }
```

aldehir (approved):
I trust you tested this on multiple models? Again, the consequence of a generalized approach: the impact surface is way larger.

pwilkin:
@aldehir yeah in most cases this wasn't even a problem to begin with, from real-life models the only one affected was GLM 4.7-Flash since it doesn't wrap its function name in anything. Pure JSON models are unaffected because they have function names atomically in `""`, so this just covers all the fringe cases.

bchtrue:
@pwilkin You did it, congratulations and thank you from all of us!

PR #20185, opened 2026-03-07
cuda : display total and free VRAM capacity during device initialization

tehsiuhuang:
- While running the benchark, realized that no VRAM info and had to check in another console
- Helps users verify hardware constraints directly from the log output even when running the benchmark
- Get total VRAM via cudaGetDeviceProperties and free memory via cudaMemGetInfo.
- Tested on Alienware m16 R2 (RTX 4070 Laptop GPU).

### Test Result

**llama-bench** & **test-backend-ops**
image

image

image

**ctest**
1 failure (# 14) not related to my change
image

image


am17an (approved):

tehsiuhuang:
@am17an Thanks for your review!!!

PR #20189, opened 2026-03-07
llama: add cb_pre_alloc callback for pre-allocation backend reassignment

aristath:
## Summary

Adds a `llama_pre_alloc_callback` that fires after graph construction but before memory allocation in `llama_decode`/`llama_encode`. This allows downstream consumers to call `ggml_backend_sched_set_tensor_backend()` to route specific ops to a different backend — without modifying llama.cpp internals.

**Use case:** When the model weights + KV cache exceed discrete GPU VRAM (common with long contexts, large models, or constrained hardware), the KV cache spills to system RAM. The dGPU must then read KV data over PCIe for attention ops, which becomes the bottleneck. On systems with an integrated GPU, the iGPU can read system RAM directly at ~40 GB/s vs ~27 GB/s over PCIe. This callback allows routing attention ops to the iGPU for faster KV cache access, without modifying llama.cpp core.

## Changes

- **`include/llama.h`** — New `llama_pre_alloc_callback` typedef + `cb_pre_alloc`/`cb_pre_alloc_user_data` fields on `llama_context_params`
- **`src/llama-cparams.h`** — Matching fields on `llama_cparams`
- **`src/llama-context.cpp`** — Wire params→cparams, invoke callback in `process_ubatch()` between `build_graph` and `alloc_graph`, add nullptr defaults
- **`tests/test-pre-alloc-callback.cpp`** — Verifies callback is invoked during decode and that `ggml_backend_sched_set_tensor_backend()` works from within it
- **`tests/CMakeLists.txt`** — Register test

## Design notes

- Follows the same pattern as the existing `cb_eval` callback
- The callback does **not** fire when a previous graph is reused (`can_reuse` path) — backend assignments from the last invocation persist, which is correct for static offload patterns
- The callback does **not** fire during `graph_reserve()` — the scheduler's built-in re-reserve in `alloc_splits` handles the first-decode buffer mismatch automatically
- Defaults to `nullptr` (no-op) — zero overhead when unused

## Test plan

- [x] Build passes with no warnings (`llama-cli`, `test-pre-alloc-callback`, `test-mtmd-c-api`)
- [x] C compilation unaffected (verified via `test-mtmd-c-api.c`)
- [x] Test verifies callback invocation and backend reassignment
- [x] Run `ctest -L main` for regression (requires CI)
- [x] Run `ctest -R test-pre-alloc-callback` with a model

### DIFF
--- a/include/llama.h
+++ b/include/llama.h
@@ -212,6 +212,12 @@ extern "C" {
 
     typedef bool (*llama_progress_callback)(float progress, void * user_data);
 
+    // called after graph build but before memory allocation in llama_decode/llama_encode
+    // use ggml_backend_sched_set_tensor_backend() to reassign ops to a different backend
+    // NOTE: not called when a previous graph is reused; assignments from the last invocation
+    //       persist. set LLAMA_GRAPH_REUSE_DISABLE=1 for per-decode control.
+    typedef void (*llama_pre_alloc_callback)(ggml_backend_sched_t sched, struct ggml_cgraph * gf, void * user_data);
+
     // Input data for llama_encode/llama_decode
     // A llama_batch object can contain input about one or many sequences
     // The provided arrays (i.e. token, embd, pos, etc.) must have size of n_tokens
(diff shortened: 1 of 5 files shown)


taronaeo:
https://github.com/ggml-org/llama.cpp/blob/c5a778891ba0ddbd4cbb507c823f970595b1adc2/CONTRIBUTING.md?plain=1#L9-L25

aristath:
AI was used in an assistive capacity due to a temporary a11y issue.
I have the code in my head, the fact that I used an AI tool to do the actual typing and push the commits does not mean this is an AI-generated contribution...

ggerganov:
> Use case: When the model weights + KV cache exceed discrete GPU VRAM (common with long contexts, large models, or constrained hardware), the KV cache spills to system RAM.

llama.cpp already provides many knobs to partially offload/distribute model layers and cache, so I doubt this API is needed. It would be useful if you describe a specific use case that you have and likely we can guide you how to use the existing options to make it work without changes to llama.cpp.

Issue #20198, opened 2026-03-07
Eval bug: llama-server tool_calls returns arguments as JSON object instead of string, breaking OpenAI compatibility


alvis233:
### Name and Version

build: 8233 (c5a778891) with GNU 11.4.0 for Linux x86_64

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

NVIDIA GeForce RTX 5090

### Models

- Qwen3-Coder-Next-UD-Q4_K_XL.gguf
https://huggingface.co/unsloth/Qwen3-Coder-Next-GGUF

- Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf
https://huggingface.co/unsloth/Qwen3.5-35B-A3B-GGUF

### Problem description & steps to reproduce

Dear mods, I am trying to run quantized model in llama.cpp through the [instructions](https://unsloth.ai/docs/models/qwen3-coder-next#run-qwen3-coder-next). However, following the recent Autoparser refactoring PR (#18675), `llama-server` returns the `arguments` field in `tool_calls` as a parsed JSON object rather than a JSON string. 

This breaks strict OpenAI API compatibility. According to the [OpenAI API Reference](https://platform.openai.com/docs/api-reference/chat/object), `tool_calls[].function.arguments` must be a string containing JSON, not a parsed object. Because of this change, the official `openai` Python SDK (which uses Pydantic for strong type checking) crashes with a `TypeError` when attempting to process tool calls.

**What I tried**
Adding the argument --jinja or --chat-template chatml both failed to adhere to OpenAI API compatibility, which made the *claw frameworks (openclaw, nanoclaw, zeroclaw, ironclaw) fail to work as expected.

**To Reproduce**
1. Start `llama-server` with a tool-capable model (e.g., Qwen3-Coder-Next) and Jinja template enabled:
   ```bash
   ./llama-server --model Qwen3-Coder-Next-UD-Q4_K_XL.gguf --jinja --port 8001
   ```
2. Send a `curl` request to the `/v1/chat/completions` endpoint with tools provided:
   ```bash
   curl -s http://127.0.0.1:8001/v1/chat/completions \
     -H "Content-Type: application/json" \
     -d '{
       "model": "Qwen3-Coder-Next",
       "messages": [{"role": "user", "content": "What is 1+1? Use the add tool."}],
[...]

PR #20202, opened 2026-03-07
Add --tool-args-object to serve tool call arguments as objects, revert to OpenAI compatible mode by default (return as strings)

pwilkin:
Currently, due to testing with numerous clients, the behavior of the API has been changed to output tool call arguments as JSON objects. However, this breaks compatibility with several clients that actually *do* expect OpenAI-compatible behavior. Therefore, I'm adding the option to return the OpenAI-compatible version via `--tool-args-compat` (or `-tac` for short).

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -3092,6 +3092,14 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_PREFILL_ASSISTANT"));
     add_opt(common_arg(
+        {"-tac", "--tool-args-compat"},
+        {"--no-tool-args-compat"},
+        "return tool call arguments as JSON strings instead of JSON objects (default: disabled)",
+        [](common_params & params, bool value) {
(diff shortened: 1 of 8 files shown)


pwilkin:
Fixes https://github.com/ggml-org/llama.cpp/issues/20198

aldehir:
This should only ever be a string. Why was it changed?

pwilkin:
@aldehir because a lot of tools actually expect this to be an object (we had a discussion on it some time ago).

aldehir:
It's not the spec. I've yet to experience a client that requires it be an object. We need to revert this entirely ASAP.

I was under the impression that we had to parse the string as JSON before feeding it to the template. Not introduce an API change. 

Who are the offending clients?

pwilkin:
Okay, let's do it this way - I'll make the string behavior the default and change the switch, that way it'll make the OAI-compat behavior the default.

pwilkin:
You're right, we shouldn't introduce backwards-incompatible changes.

pwilkin:
Aight, going to just go with the simple revert for now.

PR #20203, opened 2026-03-07
Fix compile bug

pwilkin:
Fixes #20187

### DIFF
--- a/common/chat-auto-parser-helpers.cpp
+++ b/common/chat-auto-parser-helpers.cpp
@@ -162,7 +162,7 @@ diff_split calculate_diff_split(const std::string & left, const std::string & ri
         right_fully_consumed = true;
     }
 
-    auto eat_segment = [](std::string str, segment & seg) -> std::string { return str.append(seg.value); };
+    auto eat_segment = [](std::string str, const segment & seg) -> std::string { return std::move(str) + seg.value; };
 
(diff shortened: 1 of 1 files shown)


common/chat-auto-parser-helpers.cpp
```diff
@@ -162,7 +162,7 @@ diff_split calculate_diff_split(const std::string & left, const std::string & ri
         right_fully_consumed = true;
     }
 
-    auto eat_segment = [](std::string & str, segment & seg) -> std::string { return str.append(seg.value); };
+    auto eat_segment = [](std::string str, segment & seg) -> std::string { return str.append(seg.value); };
```
CISC:
```suggestion
    auto eat_segment = [](std::string str, const segment & seg) -> std::string { return std::move(str) + seg.value; };
```

CISC (approved):
It's not a compile bug, you were modifying the input, which was the point, but that doesn't work with the C++20 implementation of `accumulate`.

common/chat-auto-parser-helpers.cpp
CISC: This is not as efficient pre-C++20 though, so you may want to avoid using `accumulate` altogether.

rwmjones:
I tested this and it fixes the build problem with GCC 16 (https://github.com/ggml-org/llama.cpp/issues/20187).  Thanks.

PR #20204, opened 2026-03-07
Attempt to recover from parsing failure on truncated input by returning last partial parse

pwilkin:
When model output ends abruptly, we might end in a scenario where not everything is output, for example, we might not have ended reasoning yet. Nevertheless, try to recover by returning the last partial output instead of throwing an error.

Fixes https://github.com/ggml-org/llama.cpp/issues/20193


aldehir:
I think we just produce an incomplete AST like we do during streaming. I'll add it to my PR.

PR #20208, opened 2026-03-07
Parse port numbers from MCP server URLs in CORS proxy

eapache:
Fixes #20205.

When specifying an MCP server on a custom port (e.g. `http://127.0.0.1:8123/mcp`), the CORS proxy was not parsing the port number, leading to a malformed host (with the trailing port still attached) and falling back to port 80/443.

This is my first PR here and it's been a while since I touched C/C++ so likely I've done something wrong with the formatting or data types or something :sweat_smile: 

The only AI used in making this PR was Copilot's autocomplete, which I trusted to pick the right stdlib methods for `stoi` and throwing a runtime error as my C++ is so rusty.

### DIFF
--- a/common/http.h
+++ b/common/http.h
@@ -7,6 +7,7 @@ struct common_http_url {
     std::string user;
     std::string password;
     std::string host;
+    int port;
     std::string path;
 };
 
@@ -47,6 +48,20 @@ static common_http_url common_http_parse_url(const std::string & url) {
         parts.host = rest;
         parts.path = "/";
     }
+
+    auto colon_pos = parts.host.find(':');
+
+    if (colon_pos != std::string::npos) {
+        parts.port = std::stoi(parts.host.substr(colon_pos + 1));
+        parts.host = parts.host.substr(0, colon_pos);
+    } else if (parts.scheme == "http") {
+        parts.port = 80;
+    } else if (parts.scheme == "https") {
+        parts.port = 443;
+    } else {
+        throw std::runtime_error("unsupported URL scheme: " + parts.scheme);
+    }
+
     return parts;
 }
 
--- a/common/http.h
+++ b/common/http.h
@@ -83,7 +83,7 @@ static std::pair<httplib::Client, common_http_url> common_http_client(const std:
     }
 #endif
 
-    httplib::Client cli(parts.scheme + "://" + parts.host);
+    httplib::Client cli(parts.scheme + "://" + parts.host + ":" + std::to_string(parts.port));
 
     if (!parts.user.empty()) {
         cli.set_basic_auth(parts.user, parts.password);
--- a/tools/server/server-cors-proxy.h
+++ b/tools/server/server-cors-proxy.h
@@ -30,7 +30,7 @@ static server_http_res_ptr proxy_request(const server_http_req & req, std::strin
         throw std::runtime_error("unsupported URL scheme in target URL: " + parsed_url.scheme);
     }
 
-    SRV_INF("proxying %s request to %s://%s%s\n", method.c_str(), parsed_url.scheme.c_str(), parsed_url.host.c_str(), parsed_url.path.c_str());
+    SRV_INF("proxying %s request to %s://%s:%i%s\n", method.c_str(), parsed_url.scheme.c_str(), parsed_url.host.c_str(), parsed_url.port, parsed_url.path.c_str());
 
     auto proxy = std::make_unique<server_http_proxy>(
             method,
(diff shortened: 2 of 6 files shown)


ngxson (commented):
Technically this won't support something like `https://myserver.com:8443`

eapache:
Why not? Because it doesn't recognize 8443 as a tls port?

ngxson:
There is a check `port == 443`

https://github.com/ggml-org/llama.cpp/blob/c5a778891ba0ddbd4cbb507c823f970595b1adc2/tools/server/server-models.cpp#L1080-L1101

eapache:
Ah ok I just need to pass the scheme in explicitly there. I will refresh the patch tomorrow morning when I'm back at my desk.

tools/server/server-models.cpp
```diff
@@ -783,6 +783,7 @@ server_http_res_ptr server_models::proxy_request(const server_http_req & req, co
     }
     auto proxy = std::make_unique<server_http_proxy>(
             method,
+            "http",
```
eapache, line 786: While adding scheme to be passed around we also now need a scheme here (for the model router). I don't see an existing scheme here, but AFAICT the child processes always just use http so I can hard-code it (model routing still works in a trivial local test with it hard-coded like this). Wanted to call this out in case there's a flag I'm not familiar with for sending the communications to the child inference processes via https or something.

tools/server/server-cors-proxy.h
```diff
@@ -30,12 +30,13 @@ static server_http_res_ptr proxy_request(const server_http_req & req, std::strin
         throw std::runtime_error("unsupported URL scheme in target URL: " + parsed_url.scheme);
     }
 
-    SRV_INF("proxying %s request to %s://%s%s\n", method.c_str(), parsed_url.scheme.c_str(), parsed_url.host.c_str(), parsed_url.path.c_str());
+    SRV_INF("proxying %s request to %s://%s:%i%s\n", method.c_str(), parsed_url.scheme.c_str(), parsed_url.host.c_str(), parsed_url.port, parsed_url.path.c_str());
```
eapache, line 33: I don't love that we now always log the port explicitly, but `parsed_url` no longer indicates whether the port was explicit or implicit.

common/http.h
```diff
@@ -68,7 +83,7 @@ static std::pair<httplib::Client, common_http_url> common_http_client(const std:
     }
 #endif
 
-    httplib::Client cli(parts.scheme + "://" + parts.host);
+    httplib::Client cli(parts.scheme + "://" + parts.host + ":" + std::to_string(parts.port));
```
eapache, line 86: I ran it through ChatGPT Codex and it pointed out that I'd broken http _downloads_ on non-standard ports because that had been relying on the port being smushed into the host still. Not quite sure how to test this, but it seems right.

common/http.h
```diff
@@ -47,6 +48,20 @@ static common_http_url common_http_parse_url(const std::string & url) {
         parts.host = rest;
         parts.path = "/";
     }
+
+    auto colon_pos = parts.host.find(':');
```
eapache, line 52: How do we feel about IPv6 addresses? They can have colons in their literals which will break here. But parsing those properly feels way outside the scope of a little internal helper method, and starts to feel like we should be pulling in a library which can do proper URL parsing...

eapache:
@ngxson I went down a bit of a rabbit-hole trying to support IPv6 literals and it turned into a substantial refactor which I didn't manage to get to the bottom of before giving up. In particular, it adds more logic anywhere we convert from parts to full-URL or back (because a host of `::1` needs to be printed and parsed as `[::1]` in the context of a URL).

I haven't tried but I'm pretty sure those are already broken for MCP in master, because we're not stripping the `[ ]` when passing the host to `server_http_proxy`, so I'd rather not consider that blocking this fix. And as I mentioned in my previous comment, this level of complexity is getting to the point where I'd consider just using a library to do the URL manipulation, it's not as bad as email addresses but there are still a lot of edge cases.

Putting all that aside, I've addressed the tls-inferred-from-port issue you pointed out, and I think otherwise this PR is a reasonable improvement over the current state of master.

ngxson (approved):
I think it's fine as a quick fix, we can refactor the parser to support ipv6 in a dedicated PR, but I think not many people will actually use it

kurnevsky:
There is one problem still left - server [sets Host header with host only](https://github.com/ggml-org/llama.cpp/blob/568aec82d2fc48341c54cae565768ac75072a31d/tools/server/server-models.cpp#L1184), but it should include port as well. There are some strict mcp servers that reject requests that have invalid Host header. Created MR: https://github.com/ggml-org/llama.cpp/pull/20843

PR #20214, opened 2026-03-07
WebUI: add tests for model id parser

bluemoehre:
Targets #20165 #20199 

I have added a set of typical model names to test the specifications extraction process. Please let me know if you think any additional special names should be included.

While working on this, I noticed that the current data structure for the technical details could be improved. We need to clarify the following points:

- Should the values `MXFP4`, `F16`, `BF16` be also assigned to the **`quantization`** field?
- If a model’s format is not `MXFP4` / `F16` / `BF16`, what value should be assigned to the **`format`** field?
- Should the properties `16E` and `MoE` each have an **`experts`** field?
- Should the categories `coder`, `instruct`, and `embed` have a **`purpose`** field?

---

Feel free to add any missing items or suggest alternative wording.

### DIFF
--- a/tools/server/webui/tests/unit/model-id-parser.test.ts
+++ b/tools/server/webui/tests/unit/model-id-parser.test.ts
@@ -61,6 +61,18 @@ describe('parseModelId', () => {
 			raw: 'openai/gpt-oss-20b:Q4_K_M',
 			tags: []
 		});
+
+		expect(parseModelId('Qwen/Qwen3-Coder-30B-A3B-Instruct-1M-BF16')).toStrictEqual({
+			activatedParams: 'A3B',
+			format: 'BF16',
+			modelName: 'Qwen3-Coder',
+			orgName: 'Qwen',
+			params: '30B',
+			// purpose: ['coder', 'instruct'],
+			quantization: null,
+			raw: 'Qwen/Qwen3-Coder-30B-A3B-Instruct-1M-BF16',
+			tags: ['Instruct', '1M']
+		});
 	});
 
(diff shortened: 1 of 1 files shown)


allozaur:
Hey @bluemoehre, thank you for starting this discussion, it's important that we standarize this sooner or later :) My answers to your questions would be:

>     * Should the values `MXFP4`, `F16`, `BF16` be also assigned to the **`quantization`** field?

Yes, ultimately from the user's perspective these fall into the same category.

>     * If a model’s format is not `MXFP4` / `F16` / `BF16`, what value should be assigned to the **`format`** field?

I think in cases like this it's always better to see what the world brings to us rather than trying to invent/forecast every edge case ahead. I. e. let's for now support all of the values that awe are aware of and are practically standarized and as it goes for the new/non-standard ones, I think we should deal with them ad hoc.

>     * Should the properties `16E` and `MoE` each have an **`experts`** field?

We could consider that, but not every model with MoE does actually have this mentioned in their name and this could be a bit misleading i think...

>     * Should the categories `coder`, `instruct`, and `embed` have a **`purpose`** field?

I think this might be a bit too extreme to have a separate field for these at this stage yet. For now i'd keep treating this as a part of model's name, but if there's gonna be a certain trend or a pattern in the future that gives us arguments good enough for creating a separate `purpose` field, then we'll do it :)

> Feel free to add any missing items or suggest alternative wording.

I think we're good for now. Lemme know what u think and let's move forward with this PR :)

allozaur:
Closind this PR as it's superseded by #20749

PR #20215, opened 2026-03-07
Map developer role to system

pwilkin:
This adds the compatibility solution to map the developer role to system role in non-GPT-OSS models, needed for some OpenAI tools.


common/chat.cpp
```diff
@@ -1276,6 +1276,48 @@ static common_chat_params common_chat_params_init_kimi_k2(const common_chat_temp
 
 namespace workaround {
 
+static void map_developer_role_to_system(json & messages) {
```
aldehir, line 1279: Just iterate and replace instances of messages with `role = "developer"`?
```cpp
for (auto & msg : messages) {
    if (msg.is_object() && msg.value("role", "") == "developer") {
        msg["role"] = "system";
    }
}
```
Why all the ceremony?

CISC: Yes, this is over-engineered, especially as gpt-oss does this:
https://github.com/ggml-org/llama.cpp/blob/022dc2425113d1c31290731ee9f2adb7bc6b7466/models/templates/openai-gpt-oss-120b.jinja#L233-L240

pwilkin: Because there can be only one system message and if there's more than one system message some templates will complain.

aldehir: Which templates, and do we really care about them?

ChatGPT's API supports multiple system messages. If we condense them into one, then an added system message at the end would affect the kv cache, would it not?

pwilkin: Mistral, for one.

```
{%- for message in loop_messages %}
    {%- if not (message.role == "tool" or (message.get('tool_calls'))) %}
        {%- if (message["role"] == "user") != (ns.index % 2 == 0) %}
            {{- raise_exception("After the optional system message, conversation roles must alternate user/assistant/user/assistant/...") }}
        {%- endif %}
        {%- set ns.index = ns.index + 1 %}
    {%- endif %}
{%- endfor %}
```

pwilkin: Qwen3.5 as well:

```
    {%- if message.role == "system" %}
        {%- if not loop.first %}
            {{- raise_exception('System message must be at the beginning.') }}
        {%- endif %}
```

CISC: Most templates only support system messages as the first message (gpt-oss included, as seen above), also I'm pretty sure mixing `developer` and `system` roles are not allowed (again; gpt-oss will ignore all later ones).

Quite a few templates will raise an error if you deviate from the [system]->user->assistant flow.

CISC: I guess you could add a check if the template allows multiple system messages, but I have a feeling even those models that do won't actually respond properly to it.

Another spanner in the works is that the intention of a late system message is probably to change course, so simply appending it to the first one is not likely to produce the desired result either.

pwilkin: But the idea here wasn't to allow a late system message - the idea was to merge the developer and system messages into one if both were present.

aldehir: They shouldn't, since developer is a replacement for system. At some point clients need to be held responsible instead of capitulating to them.

I think it's easier to change this if we do a basic 1-to-1 conversion for now and then address any issues that arise later. These issues don't sound exclusive to the developer role.

pwilkin: But developer *isn't* a replacement for system.

System is supposed to be the message provided by the deployer of the API server, developer is supposed to be the message provided by the application creator. But if you feel this strongly about it I'll just map it all to system.

CISC: If so, is system ever exposed, or is it kept internally? If our end never sees it it makes no difference.

aldehir: Using gpt-oss as the reference, the system role is not exposed to the user or dev, it's embedded in the chat template. You can override settings but only with chat template vars. It maps messages with "system" to "developer".

> The developer message is used to provide information about the instructions for the model (what is normally considered the “system prompt”) and available function tools

https://developers.openai.com/cookbook/articles/openai-harmony/

To me, this reads that developer is the old system role and the new system role is akin to the embedded instructions in the template.

It's not that I feel strongly about it, but that I don't want to provide an interpretation that users depend on only to find out it was incorrect (which I guess goes both ways).

pwilkin: Okay, I'll simplify it, realistically I don't think anyone will be passing a convo with both system and developer roles filled in (hopefully).

aldehir (approved):

gsgxnet:
According to ChatGPTs docs:  
https://developers.openai.com/api/reference/resources/chat/?utm_source=chatgpt.com#:~:text=ChatCompletionDeveloperMessageParam
All newer models should support the `developer` kind of message.  

The unsloth Qwen3.5 GGUF template I am using does not know a developer message.    
```jinja
...
%- for message in messages %}
    {%- set content = render_content(message.content, true)|trim %}
    {%- if message.role == "system" %}
        {%- if not loop.first %}
            {{- raise_exception('System message must be at the beginning.') }}
        {%- endif %}
    {%- elif message.role == "user" %}
   ...
    {%- else %}
        {{- raise_exception('Unexpected message role.') }}
    {%- endif %}
{%- endfor %}
...
```
May be because **unsloth** modified the template?  They state:  
- **Tool-calling improved** following our _chat template fixes_. **Fix is universal** and applies to **any** Qwen3.5 format and **any** uploader.

In general I followed their instructions: https://unsloth.ai/docs/models/qwen3.5#qwen3.5-122b-a10b  

For reference I cross checked with a **ggml.org** GGUF model, Qwen3-Coder.  
The template embedded in there handles `developer` as needed:  

```jinja
{#- Extract developer message #}
{%- if messages[0].role == "developer" or messages[0].role == "system" %}
    {%- set developer_message = messages[0].content %}
    {%- set loop_messages = messages[1:] %}
{%- else %}
    {%- set developer_message = "" %}
    {%- set loop_messages = messages %}
{%- endif %}

{#- Render developer message #}
{%- if developer_message or tools %}
    {{- "<|start|>developer<|message|>" }}
    {%- if developer_message %}
        {{- "# Instructions

" }}
        {{- developer_message }}
    {%- endif %}
    {%- if tools -%}
        {{- "

" }}
        {{- "# Tools

" }}
        {{- render_tool_namespace("functions", tools) }}
    {%- endif -%}
    {{- "<|end|>" }}
{%- endif %}
```

So I wonder is this a llama.cpp issue or a problem of models not packaged with the appropriate modern templates as e.g. OpenAI needs?  
I do not want to blame unsloth here, may be the template without developer role they use came directly from the original Qwen3.5 model files?  

To further complicate this, OpenAI states that when using the responses API (which codex prefers) there can `instructions` messages be used?  

Looking through all this kind of mess I do even more appreciate the great work all you maintainers do.   **THANKS!**

aldehir:
> For reference I cross checked with a ggml.org GGUF model, Qwen3-Coder.
The template embedded in there handles developer as needed:

This looks like the gpt-oss template. I checked the metadata for both ggml-org/qwen3-coder and ggml-org/qwen3-coder-next, none of those mention a developer role.

> So I wonder is this a llama.cpp issue or a problem of models not packaged with the appropriate modern templates as e.g. OpenAI needs?

Neither. OpenAI changed their API to add a developer role which supersedes the system role. Newer clients are adopting this, breaking support for older models.

> To further complicate this, OpenAI states that when using the responses API (which codex prefers) there can instructions messages be used?

This appears to be for their stateful variant of the Response API. llama.cpp only supports the stateless operation, AFAIK. It boils down to a message with `role = "developer" | "system"`.

isaac-mcfadyen:
> * following our _chat template fixes_

I mentioned this in a HuggingFace Discussion - I did a quick diff and it looks like their fixes were only related to validating that the tool call arguments are actually a dict. Nothing else was changed: https://huggingface.co/unsloth/Qwen3.5-35B-A3B-GGUF/discussions/10#69a2f3bc51e1b4bebaa2484d

aldehir:
@pwilkin let's make sure we follow the contributor's guideline on commit messages in merges.

gsgxnet:
> ...
> This looks like the gpt-oss template. I checked the metadata for both ggml-org/qwen3-coder and ggml-org/qwen3-coder-next, none of those mention a developer role.
> 
Yes my fault, was gpt-oss.  I mixed two template extractions.

> > So I wonder is this a llama.cpp issue or a problem of models not packaged with the appropriate modern templates as e.g. OpenAI needs?
> 
> Neither. OpenAI changed their API to add a developer role which supersedes the system role. Newer clients are adopting this, breaking support for older models.
> 
Understood.  It is very difficult if you have to follow a foreign standard which is not long term fixed.  


> > To further complicate this, OpenAI states that when using the responses API (which codex prefers) there can instructions messages be used?
> 
> This appears to be for their stateful variant of the Response API. llama.cpp only supports the stateless operation, AFAIK. It boils down to a message with `role = "developer" | "system"`.

I tried to better understand the stateless operation and the role of the llama.cpp caches.  ChatGPT 5.4 after a short research summarised:  
**llama.cpp server with Responses-style usage should be treated as effectively stateless at the API-contract level, with internal KV/slot caches used as a best-effort optimization rather than as authoritative conversation state.**

Got it!  
Final question (do not put much effort in answering) - would you estimate how much development effort it would take to implement the stateful version of the responses API within llama.cpp?  And how beneficial or even worse it would be, performance and memory consumption wise?

lostmsu:
codex + qwen3.5 with this PR still fail for me with

> Jinja Exception: System message must be at the beginning.

pwilkin:
@lostmsu that's a codex problem, unfortunately :/

isaac-mcfadyen:
@lostmsu You could (manually) modify the Qwen chat template to remove this restriction and then use `--chat-template-file` with the updated Jinja file, but Qwen was trained with the system prompt always at the start so it might cause degraded performance/hallucinations.

PR #20223, opened 2026-03-08
Fix structured outputs

pwilkin:
Fixes support for structured outputs in autoparser.

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -1,6 +1,7 @@
 #include "chat-auto-parser.h"
 #include "chat-peg-parser.h"
 #include "chat.h"
+#include "common.h"
 #include "json-schema-to-grammar.h"
 #include "nlohmann/json.hpp"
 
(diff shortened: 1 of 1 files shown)


pwilkin:
Fixes #20221

tarruda:
@pwilkin thanks for this.

I'm not familiar with the testing infrastructure, but could we add the example request as a regression test for llama-server structured output?

pwilkin:
Can try to add a server test for it, see how the tiny-stories model handles structured output ;)

common/chat-auto-parser-generator.cpp
```diff
@@ -104,8 +107,9 @@ common_peg_arena autoparser::build_parser(const templates_params & inputs) const
         bool has_response_format = inputs.json_schema.is_object() && !inputs.json_schema.empty();
 
         if (has_response_format) {
-            return ctx.reasoning_parser + p.space() +
-                   p.content(p.schema(p.json(), "response-format", inputs.json_schema)) + p.end();
+            return ctx.reasoning_parser + p.space() + p.optional(p.literal("```json") + p.space()) +
```
aldehir: Let's use order choice with the code fence wrapped version as the first and a non-code fences as the other. Place the response JSON in a separate rule. This way we enforce one or the other and not allow some weird mix of both.

common/chat-auto-parser-generator.cpp
```diff
@@ -104,8 +107,9 @@ common_peg_arena autoparser::build_parser(const templates_params & inputs) const
...
+                p.content(p.schema(p.json(), "response-format", inputs.json_schema)) +
+                p.optional(p.space() + p.literal("```")) + p.end();
```
aldehir:
```suggestion
            auto response_format = p.rule("response-format", p.content(p.schema(p.json(), "response-format-schema", inputs.json_schema)));
            return ctx.reasoning_parser + p.space() + p.choice({
                p.literal("```json") + p.space() + response_format + p.space() + p.literal("```"),
                response_format
            }) + p.end();
```

aldehir (approved):

common/chat-auto-parser-generator.cpp
```diff
@@ -104,8 +107,11 @@ common_peg_arena autoparser::build_parser(const templates_params & inputs) const
         bool has_response_format = inputs.json_schema.is_object() && !inputs.json_schema.empty();
 
         if (has_response_format) {
-            return ctx.reasoning_parser + p.space() +
-                   p.content(p.schema(p.json(), "response-format", inputs.json_schema)) + p.end();
+            auto response_format = p.rule("response-format", p.content(p.schema(p.json(), "response-format-schema", inputs.json_schema)));
+            return ctx.reasoning_parser + p.space() + p.choice({
+                p.literal("```json") + p.space() + response_format + p.space() + p.literal("```"),
```
tarruda, line 112: I might not be interpreting this correctly, but should the model be able to output fenced markdown blocks on a structured response output? The main guarantee of a structured JSON response is that it will only contain valid JSON without having to extract from any markdown blocks.

aldehir: The parser will extract the JSON within the code fences, you won't see them in the response (see definition wrapped in `p.content()`).

This is good for models whose training set contain a bunch of JSON examples in code fences, such as Gemma 3.

tarruda: I've always assumed that llama.cpp manipulated the model's token prediction in a way that ensured only tokens that kept the output valid according to the JSON schema, and that outputting a fenced tokens would not even be possible.

aldehir: Grammar-constrained decoding works for any context-free language. JSON is one, but so is JSON wrapped in fences. We need to support more than just JSON, especially for reasoning models that need to reason before generating a structured response.

Issue #18389, opened 2025-12-26
mtmd: plan to add video support


ngxson:
Ref discussions from https://github.com/ggml-org/llama.cpp/pull/16910 and https://github.com/ggml-org/llama.cpp/issues/17660

### API design

**Do we need to allow streaming frames?**

The main issue is that streaming is only beneficial for the encoding pass, not the decoding pass. We can probably only allow async frame feeding to avoid having to store the whole decoded video in memory.

A pseudo code looks like this:

```py
chunks = mtmd_tokenize(...)
while chunk in chunks:
    mtmd_tokenize_lazy(chunk, frame_data, frame_size) # will also call preprocessing
    mtmd_helper_eval_chunk_single(chunk, ...) # will do the encode+decode pass, then free the chunk
```

**Audio + image interleave**

TODO: investigate how Qwen-Omni is handling this

**What are other APIs that need to be added?**

TODO: maybe exposing FPS for models using fixed FPS?

### Video decoding

Linking against `libavcodec` sounds like a cleaner approach, but will lead to a bad UX in practice. Some systems may not have `libavcodec` and will be forced to use another build of `libmtmd.so/.dll`

A temporary solution for the first iteration would be to call `ffmpeg` command and check its existence. If yes, we launch it via `subprocess.h` and ask it to decode the video.

Second solution is to manually use `dlopen` and search for `libavcodec`, reject if it's not found. This will avoid`libmtmd.so/.dll` to be linked against `libavcodec` on building phrase.

### Internal design

As mentioned in the section earlier, we will need to implement a "lazy" image tokenizer.

For models doing "fused" frame like Qwen (via 3D conv), we need to allow multiple images to be processed in the same batch. `clip_image_f32` will need to be extended with a `nz` dimension.

awneesht:
Hi @ngxson, I'd like to contribute to this. I'm planning to start with Phase 1 — frame extraction via ffmpeg subprocess + wiring it into mtmd for SmolVLM2-Video (frame-by-frame, no 3D conv). I'll follow the design you outlined here. Will submit a focused PR for just the frame sampler first. Any guidance or blockers I should know about?"

igor-suhorukov:
[Similar PR from Ollama](https://github.com/ollama/ollama/pull/12962)

andrewmd5:
[...]

PR #20224, opened 2026-03-08
feat: add video support to mtmd

andrewmd5:
As discussed in https://github.com/ggml-org/llama.cpp/issues/18389#issuecomment-4016586057, this PR adds support for video to mtmd. Results below are from a ~5 minute video scene-by-scene summary from Qwen3.5

Image

### DIFF
--- a/tools/mtmd/clip.cpp
+++ b/tools/mtmd/clip.cpp
@@ -3596,9 +3596,17 @@ bool clip_image_batch_encode(clip_ctx * ctx, const int n_threads, const clip_ima
 
     // set input pixel values
     if (!imgs.is_audio) {
+        // detect number of channels from the buffer size
+        const int nx = imgs.entries[0]->nx;
+        const int ny = imgs.entries[0]->ny;
+        const int n  = nx * ny;
+        const size_t buf_size = imgs.entries[0]->buf.size();
+        const int n_channels = (int)(buf_size / n);
+        GGML_ASSERT(n_channels == 3 || n_channels == 6);
+
         size_t nelem = 0;
         for (const auto & img : imgs.entries) {
-            nelem += img->nx * img->ny * 3;
+            nelem += img->nx * img->ny * n_channels;
         }
         std::vector<float> inp_raw(nelem);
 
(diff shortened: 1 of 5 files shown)


andrewmd5:
I noticed upon further testing that #20284 seems to have degraded the performance of this implementation substantially; I'll collect more data and see whats going on.

libin049:
I tried this branch but encountered some issues. Did I miss any steps? Here are the commands I used:
```shell
./build/bin/llama-mtmd-cli -m ./Qwen2.5-Omni-3B.gguf --mmproj ./mmproj-Qwen2.5-Omni-3B.gguf \
  --image ./draw.mp4 --temp 0 -n 128 --flash-attn on -p "what is this?"
```
The draw.mp4 file is from:
https://qianwen-res.oss-cn-beijing.aliyuncs.com/Qwen2.5-Omni/draw.mp4?spm=5176.28103460.0.0.4d216308swiAIN&file=draw.mp4

Error output:
```
WARN: This is an experimental CLI for testing multimodal capability.
      For normal use cases, please use the standard llama-cli
mtmd_helper_bitmap_init_from_buf: failed to decode image bytes
```

tools/mtmd/mtmd.cpp
```diff
@@ -34,15 +36,17 @@ struct mtmd_bitmap {
 struct mtmd_image_tokens {
     uint32_t nx; // number of tokens in x direction
     uint32_t ny; // number of tokens in y direction
+    uint32_t nt = 1; // number of temporal positions (1 for images, > 1 for video)
     bool use_mrope_pos = false; // use M-RoPE position counting (the whole image is 1 temporal position)
-    uint32_t n_tokens() const { return nx * ny; }
+    uint32_t n_tokens() const { return nt * nx * ny; }
```
ngxson, line 41: this seems to be wrong for qwen. it merges 2 frame into one output, so the output token count should stays `nx * ny`

tools/mtmd/models/qwen3vl.cpp
```diff
@@ -13,16 +13,34 @@ ggml_cgraph * clip_graph_qwen3vl::build() {
...
+        ggml_tensor * inp_even = ggml_view_3d(ctx0, inp_raw, img.nx, img.ny, 3, nb1, nb2, 0);
+        ggml_tensor * inp_odd  = ggml_view_3d(ctx0, inp_raw, img.nx, img.ny, 3, nb1, nb2, nb2 * 3);
+        inp = ggml_add(ctx0,
+            ggml_conv_2d(ctx0, model.patch_embeddings_0, inp_even, patch_size, patch_size, 0, 0, 1, 1),
+            ggml_conv_2d(ctx0, model.patch_embeddings_1, inp_odd,  patch_size, patch_size, 0, 0, 1, 1));
+    } else {
+        inp = ggml_add(ctx0,
+            ggml_conv_2d(ctx0, model.patch_embeddings_0, inp_raw, patch_size, patch_size, 0, 0, 1, 1),
+            ggml_conv_2d(ctx0, model.patch_embeddings_1, inp_raw, patch_size, patch_size, 0, 0, 1, 1));
+    }
```
ngxson, line 37: if I read this correctly, that means number of output token stays unchanged whether we input single image, or 2 frames

tools/mtmd/clip.cpp
```diff
@@ -3612,21 +3620,21 @@ bool clip_image_batch_encode(clip_ctx * ctx, const int n_threads, const clip_ima
...
+            const int cur_n  = cur_nx * cur_ny;
+
+            float * batch_entry = inp_raw.data() + b * (n_channels * cur_n);
+            for (int y = 0; y < cur_ny; y++) {
+                for (int x = 0; x < cur_nx; x++) {
+                    size_t base_src = n_channels * (y * cur_nx + x);
+                    size_t base_dst =              y * cur_nx + x;
+                    for (int c = 0; c < n_channels; c++) {
+                        batch_entry[c * cur_n + base_dst] = imgs.entries[b]->buf[base_src + c];
                     }
```
ngxson, line 3638: probably better to enter multiple images via the batch dimension, rather than using 6 channels

ngxson (commented):
I'm opening some comments for discussion. I'm taking over this PR and will push commits directly here, so please do not push new commits to avoid conflicts.

tools/mtmd/mtmd-helper.cpp
```diff
@@ -174,6 +174,28 @@ struct decode_embd_batch {
...
+                    pos[i + batch.n_tokens * 3] = 0;
+                }
+            }
+        }
+        for (int i = 0; i < batch.n_tokens; i++) {
+            batch.n_seq_id[i] = 1;
+            batch.seq_id  [i] = seq_id_0.data();
+            batch.logits  [i] = false;
+        }
+    }
```
ngxson, line 197: note: to be replaced by #21851

ngxson:
hmm ok I cannot push to this PR because it's created from an org account

closing this and move to a new one

PR #20226, opened 2026-03-08
server : correct index on finish in OAI completion streams

decahedron1:
When calling `/v1/chat/completions` with `"stream": true` and an `n` value greater than 1, the stop message would always have its `index` set to `0` instead of the correct index, making it impossible to tell which stream actually finished.

### DIFF
--- a/tools/server/server-task.cpp
+++ b/tools/server/server-task.cpp
@@ -827,7 +827,7 @@ json server_task_result_cmpl_final::to_json_oaicompat_chat_stream() {
             {"choices", json::array({
                 json {
                     {"finish_reason", nullptr},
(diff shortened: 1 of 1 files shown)


ggerganov (approved):

ngxson (approved):

ggerganov:
@decahedron1 Thanks for the fix. If you would like to contribute more, adding a basic server test to https://github.com/ggml-org/llama.cpp/blob/master/tools/server/tests/unit/test_chat_completion.py would be appreciated. It can send a chat completion request with stream = true and n > 1 and assert that the n indices are observed in the responses.

PR #20230, opened 2026-03-08
ggml-webgpu: Add supports for `GGML_OP_REPEAT`

yomaytk:
This PR adds supports for `GGML_OP_REPEAT` to the WebGPU backend. The status of `REPEAT` for WebGPU in `docs/ops.md` is changed to "partially supported" because WebGPU doesn't seem to support `i16`.

Also, this PR includes formatting changes (clang-format) for the modified files. Since https://github.com/ggml-org/llama.cpp/pull/20173 touches the same parts, this PR might need to be merged after that one.


reeselevine:
Potentially could support `i16` by just treating the bytes being repeated as opaque, since they're just copied/repeated over to the `dst` buffer, but I don't think that's required. Curious, are there any particular models/usecases that need repeat? I haven't seen it come up in what I'm working on yet.

reeselevine:
And sounds good on merging after https://github.com/ggml-org/llama.cpp/pull/20173, I should merge that soon.

CISC:
> Curious, are there any particular models/usecases that need repeat? I haven't seen it come up in what I'm working on yet.

There are a couple, but most recently (and notably) Qwen 3.5.

yomaytk:
> Curious, are there any particular models/usecases that need repeat? I haven't seen it come up in what I'm working on yet.

I confirmed that DeepSeek-V2 also uses REPEAT, and DeepSeek-V3 may as well.

> Potentially could support i16 by just treating the bytes being repeated as opaque, since they're just copied/repeated over to the dst buffer

I see. As you mentioned, treating the repeated bytes as opaque would work. However, using `f16` is simpler here, and the CPU backend does the same. So I added `i16` support in the same way, and the `i16` test pass as well. https://github.com/ggml-org/llama.cpp/blob/c96f608d9861f7e8466bc1b6ac2ff4e3c6f96641/ggml/src/ggml-cpu/ops.cpp#L1787-L1792

reeselevine:
Now that https://github.com/ggml-org/llama.cpp/pull/20173 is merged just need to fix some conflicts then we should be good to merge this!

reeselevine (approved):

reeselevine:
Looks good to me! Do you know if there are other operations/changes necessary to get Qwen 3.5 running? I tried [this](https://huggingface.co/unsloth/Qwen3.5-2B-GGUF?show_file_info=Qwen3.5-2B-Q4_0.gguf) model off this branch and I'm getting a segfault in the WebGPU backend. I can try and debug unless you're already looking into/working on it.

yomaytk:
According to my experiments, the following operations appear to be not yet implemented for Qwen 3.5 in WebGPU backend.

`L2_NORM, SET, DIAG, TRI, SSM_CONV, FLASH_ATTN_EXT, SOLVE_TRI, GATED_DELTA_NET`

> I tried [this](https://huggingface.co/unsloth/Qwen3.5-2B-GGUF?show_file_info=Qwen3.5-2B-Q4_0.gguf) model off this branch and I'm getting a segfault in the WebGPU backend.

In my program (based on `examples/simple-chat`) too, the model fails with the following errors. 

スクリーンショット 2026-03-11 19 01 54

On the other hand, [unsloth/Qwen3.5-4B-Q4_0.gguf](https://huggingface.co/unsloth/Qwen3.5-4B-GGUF?show_file_info=Qwen3.5-4B-Q4_0.gguf) and [unsloth/Qwen3.5-9B-Q4_0.gguf](https://huggingface.co/unsloth/Qwen3.5-9B-GGUF?show_file_info=Qwen3.5-9B-Q4_0.gguf) seem to work well with my program.

> I can try and debug unless you're already looking into/working on it.

I haven't been working on Qwen 3.5 support myself, so it would be great if you could work on this!

Issue #19955, opened 2026-02-27
Eval bug: Vulkan llama-server crashes with vk::DeviceLostError randomly (long contexts)


el95149:
### Name and Version

./llama-server --version
version: 8173 (2e7e63852)

### Operating systems

Linux

### GGML backends

Vulkan

### Hardware

- Ryzen 7900X
- 64 GB DDR5@6000
- X870e board
- NVIDIA RTX 5080 16GB
- AMD Radeon AI Pro R9700

### Models

Has happened with at least the following models:
- Unsloth Qwen3-Coder-Next-Q4_K_M.gguf
- Unsloth Qwen3-Coder-Next-UD-Q6_K_XL
- Unsloth Qwen3.5-35B-A3B-Q8_0

### Problem description & steps to reproduce

I start the server with the following command, either using both cards or any single one (via GGML_VK_VISIBLE_DEVICES):

```bash
/llama-server \
  -m /home/<redacted>/.lmstudio/models/unsloth/Qwen3-Coder-Next-GGUF/Qwen3-Coder-Next-Q4_K_M.gguf \
  --port 1234 --host 0.0.0.0 -fa on --no-mmap \
  -b 2048 -ub 2048 -c 100000 --cache-ram 0 \
  --temp 1 --top-k 40 --top-p 0.95 --repeat-penalty 1 --min-p 0.01
```
After a few, random, thousands worth of token processing/generating (typically above ~40K in total), I get a momentary system freeze, and then the model crashes, with any client unable to proceed with inference. (see attached log).

### First Bad Commit

Was able to notice it after build [b8143](https://github.com/ggml-org/llama.cpp/releases/tag/b8143), since before that inference was too slow to make things really usable.

### Relevant log output

<details>
<summary>Logs</summary>

```console
[39895] srv  params_from_: Chat format: peg-constructed
[39895] slot get_availabl: id  2 | task -1 | selected slot by LRU, t_last = -1
[39895] slot launch_slot_: id  2 | task -1 | sampler chain: logits -> ?penalties -> ?dry -> ?top-n-sigma -> top-k -> ?typical -> top-p -> min-p -> ?xtc -> temp-ext -> dist
[39895] slot launch_slot_: id  2 | task 1791 | processing task, is_child = 0
[39895] slot update_slots: id  2 | task 1791 | new prompt, n_ctx_slot = 100096, n_keep = 0, task.n_tokens = 58741
[...]

PR #20233, opened 2026-03-08
vulkan: skip zero size tensors in backend copies

0cc4m:
I noticed validation errors in #19955
```
Validation Error: [ VUID-VkBufferCopy-size-01988 ] | MessageID = 0x636fcc49
vkCmdCopyBuffer(): pRegions[0].size is zero.
The Vulkan spec states: The size must be greater than 0 (https://docs.vulkan.org/spec/latest/chapters/copies.html#VUID-VkBufferCopy-size-01988)
Objects: 1
    [0] VkCommandBuffer 0x57d940f3cda0
```

I traced it to the backend async tensor copy function being called with zero-size tensors. I added checks to all backend interfaces to skip size 0 tensors.


HumerousGorgon:
With -ctk q8_0 -ctv q8_0, I still get gibberish. Removing both brings back function to the model on latest builds.

jeffbolznv (approved):

PR #20235, opened 2026-03-08
graph : remove redundant scale_w parameter

CISC:
Check `w_scale` before applying instead.


JohannesGaessler (approved):
In principle we could also change the default value from `0.0f` to `1.0f` since it is multiplicative but I have no real preference when it comes to that.

CISC:
> In principle we could also change the default value from `0.0f` to `1.0f` since it is multiplicative but I have no real preference when it comes to that.

True, but we'd still have to check for `0.0f`, so makes no real difference.

ggerganov (approved):

PR #20238, opened 2026-03-08
webui : add option to copy assistant response without thinking content

rankaiyx:
## Summary

Add a setting to copy assistant messages without thinking/reasoning content.

Close #19982

## Changes

- `settings-keys.ts`: add `COPY_WITHOUT_THINKING` key constant
- `settings-config.ts`: add default value (`false`) and description for the new setting
- `ChatSettings.svelte`: add checkbox in the General section
- `ChatMessage.svelte`: strip reasoning blocks before copying when the setting is enabled

## Behavior

When **"Copy response without thinking content"** is enabled in Settings → General, clicking the copy button on an assistant message will strip the `<<<reasoning_content_start>>>...<<<reasoning_content_end>>>` block and only copy the final response.

The implementation adheres to the **minimum modification principle**,  reuses the existing `AGENTIC_REGEX.REASONING_BLOCK`, `AGENTIC_REGEX.REASONING_OPEN`, and `TRIM_NEWLINES_REGEX` from `agentic.ts`. No new parsing logic is introduced.

The default value is `false`, so existing behavior is unchanged.

## Testing

Tested with a reasoning model (Qwen3.5-35B-A3B). With the setting off, the full message (including thinking) is copied. With the setting on, only the final response is copied.

---

This patch was developed with the assistance of Claude (Anthropic).

### DIFF
--- a/tools/server/webui/src/lib/components/app/chat/ChatMessages/ChatMessage.svelte
+++ b/tools/server/webui/src/lib/components/app/chat/ChatMessages/ChatMessage.svelte
@@ -6,6 +6,9 @@
 	import { conversationsStore } from '$lib/stores/conversations.svelte';
 	import { DatabaseService } from '$lib/services';
 	import { SYSTEM_MESSAGE_PLACEHOLDER } from '$lib/constants';
+	import { AGENTIC_REGEX, TRIM_NEWLINES_REGEX } from '$lib/constants/agentic';
+	import { copyToClipboard } from '$lib/utils/clipboard';
+	import { config } from '$lib/stores/settings.svelte';
 	import { MessageRole, AttachmentType } from '$lib/enums';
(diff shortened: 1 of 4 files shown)


allozaur (commented):
Hey! Nice, clean change.

Please just add:
1. The updated WebUI build output.

```bash
npm run build
git add server/tools/webui/public/index.html.gz
git commit -m "chore: update webui build output"
git push
```

2. Short video demo with a simple test would be also good to have :)

3. (Nice to have) I am planning to add more tests soon, but if u want to help with that you could consider adding a Vitest unit test and/or Storybook test

allozaur:
> Apologies, I missed this email.
> 
> 
> 
> The fix currently on the main branch is a cleaner and better approach!

Np 😉

PR #20263, opened 2026-03-09
common : consolidate PEG string parsers

aldehir:
PEG parser clean up post Autoparser merge.

Changes:
- Replace `json_string_parser` and `python_string_parser` with a parameterized `string_parser`.
- Inline `generic_*` parsers.

### DIFF
--- a/common/peg-parser.cpp
+++ b/common/peg-parser.cpp
@@ -658,7 +658,7 @@ struct parser_executor {
         return common_peg_parse_result(COMMON_PEG_PARSE_RESULT_SUCCESS, start_pos, pos);
     }
 
-    static common_peg_parse_result handle_escape_sequence(common_peg_parse_context & ctx, size_t start, size_t & pos) {
+    static common_peg_parse_result handle_escape_sequence(common_peg_parse_context & ctx, size_t start, size_t & pos, const char delimiter) {
         ++pos; // consume '\'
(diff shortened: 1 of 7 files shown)


common/peg-parser.cpp
```diff
@@ -1281,47 +1227,25 @@ common_peg_arena common_peg_parser_builder::build() {
...
-                [this]() { return sequence({ literal("'"), single_quoted_string_content(), literal("'"), space() }); });
-}
-
-common_peg_parser common_peg_parser_builder::flexible_string() {
-    return rule("flexible-string", [this]() { return choice({ double_quoted_string(), single_quoted_string() }); });
-}
-
-// Generic helpers for object/array structure
-
-common_peg_parser common_peg_parser_builder::generic_object(const std::string &            name,
```
pwilkin, line 1308: Just wondering - why remove the `generic_object` and `generic_array` helpers and then repeat the code in both cases?

pwilkin (approved):

common/peg-parser.cpp
aldehir: Removing a level of indirection. 

I prefer to see the entire grammar spelled out in a single function, with boundaries that closely map to the language. To that extent, I don't mind a bit of duplication.

Hope that makes sense.

PR #20266, opened 2026-03-09
优化LUT16矩阵乘法

QianerPKU:


inforithmics:
I think a link to the paper would have helped. 
https://arxiv.org/pdf/2508.05571. It is an interesting concept to use complex numbers for 2 Bit quantizations.

PR #19532, opened 2026-02-12
imatrix: fix crash when using --show-statistics with zero counts

ssam18:
This fixes the floating point exception crash reported in bug #19190 when running llama-imatrix with --show-statistics on files containing zero count values. The fix adds proper checks before division operations and handles edge cases where all counts are zero.

PR #20271, opened 2026-03-09
imatrix : prevent division by zero in statistics output

edenfunf:
When using `--show-statistics` with an imatrix file generated from a MoE model, some expert matrices may have never been activated during calibration, leaving their counts[i] at zero. Dividing by a zero count caused a crash.

Skip entries with zero count during the activations loop. Also add an early return if all entries were skipped (activations is empty), which would otherwise cause undefined behavior in std::max_element and std::min_element on an empty range.

The core imatrix computation logic is not affected.

Fixes #19190


CISC:
Duplicate of #19532

PR #20132, opened 2026-03-05
kv-cache : fix M-RoPE checkpoints

ggerganov:
fix #20002
cont https://github.com/ggml-org/llama.cpp/pull/16825#issuecomment-3460868350

Extend the KV cache save/load to handle the `llama_kv_cell_ext` data.

PR #20273, opened 2026-03-09
Read the persisted llama_kv_cell_ext for n_pos_per_embd > 1 on state_read for all sequence ids

sprayandwipe:
cont #20132

Attempting to call llama_kv_cache::state_read fails when n_pos_per_embd is greater than 1, since llama_kv_cell_ext data is serialised in `state_save` but not read back in `state_read`, leading to deserialisation failure since the cell_ext data is being parsed as a seq_id.

I assume the attached fix is correct -- kv cache persistence to host memory is now working as expected.

### DIFF
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -1955,6 +1955,12 @@ bool llama_kv_cache::state_read_meta(llama_io_read_i & io, uint32_t strm, uint32
 
             cells.pos_set(i, pos);
 
+            if (hparams.n_pos_per_embd() > 1) {
+                llama_kv_cell_ext ext;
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
Thanks

LostRuins:
THANK you for this. I was getting
```
state_read_meta: invalid seq_id, 1 is out of range [0, 1)
state_set_data: error loading state: failed to restore kv cache
``` 
after merging https://github.com/ggml-org/llama.cpp/pull/20132 ... and it made 0 sense because I never use `seq_id = 1`. 

Merged this fix and the issue appears to be solved. +1.

LostRuins (approved):

PR #20279, opened 2026-03-09
server : fix off-by-1 in server_tokens::size_up_to_pos()

ggerganov:
fix https://github.com/ggml-org/llama.cpp/pull/20087#issuecomment-4021486697

`n_past` should reflect the number of tokens with position strictly smaller than the next position `pos_next`.


aldehir (approved):
Appears to fix the issue, I can't seem to reproduce it anymore. Thanks!

ngxson (approved):

PR #20289, opened 2026-03-09
Add `--skip-chat-parsing` to force a pure content parser.

pwilkin:
Allows pure content throughput with Jinja template to tools that want to perform their own parsing / tool call handling / etc.

Fixes #20281


common/arg.cpp
```diff
@@ -3080,6 +3080,17 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.chat_template = read_file(value);
         }
     ).set_examples({LLAMA_EXAMPLE_COMPLETION, LLAMA_EXAMPLE_CLI, LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_CHAT_TEMPLATE_FILE"));
+    add_opt(common_arg(
+        {"--force-pure-content"},
+        {"--no-force-pure-content"},
+        string_format(
+            "force a pure content parser, even if a Jinja template is specified; model will output everything "
+            "in the content section, including any reasaoning and/or tool calls (default: disabled)"
```
CISC:
```suggestion
            "in the content section, including any reasoning and/or tool calls (default: disabled)"
```

Vincent-EdgerunnerAI:
~~unless this is a different issue, as @aldehir mentioned, the GPT-OSS parser still results in the code 500~~

nvm I was using the wrong argument :) 

This is tested as working for me.

Thanks for the speedy work!

common/arg.cpp
```diff
@@ -3080,6 +3080,17 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
...
+        {"--force-pure-content"},
+        {"--no-force-pure-content"},
+        string_format(
+            "force a pure content parser, even if a Jinja template is specified; model will output everything "
+            "in the content section, including any reasoning and/or tool calls (default: disabled)"
+        ),
+        [](common_params & params, bool value) {
+            params.force_pure_content_parser = value;
+        }
+    ).set_examples({LLAMA_EXAMPLE_COMPLETION, LLAMA_EXAMPLE_CLI, LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_FORCE_PURE_CONTENT"));
```
aldehir: `--chat-parsing [on]/off` maybe? Since that is what this will effectively control.

Thoughts @CISC?

CISC: IMO `on`/`off` should be reserved for parameters where some auto logic is applied when not specified, the norm for others is prepending `no-`. `[no-]chat-parsing` seems a better name than `[no-]force-pure-content` though.

pwilkin: Aight, changed to `--skip-chat-parsing`.

Vincent-EdgerunnerAI:
Bumping in case it got stale, looking forward to getting this merged, thanks!

aldehir (approved):

firas9941 (approved):

PR #20294, opened 2026-03-09
ci: disable coopmat on ubuntu-24-cmake-vulkan job

0cc4m:
There's some issue with the new barrier in the coopmat llvmpipe driver. But coopmat support in llvmpipe isn't really needed for it and wasn't around when the CI was set up anyways, so disable it to work around the problem.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -469,6 +469,7 @@ jobs:
           cd build
           export GGML_VK_VISIBLE_DEVICES=0
           export GGML_VK_DISABLE_F16=1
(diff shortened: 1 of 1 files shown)


CISC (approved):

jeffbolznv:
There's a decent chance these barriers will cause problems on other drivers - NV drivers probably would have had a problem until about a year ago.

0cc4m:
They can be trouble because they are unusual? We can take them out again, but as you said, that's a theoretical race condition. I'd rather not pay the full barrier price, if it's not actually required for the code. I'm not sure what to do here.

PR #17750, opened 2025-12-04
server: handle limiting maximum reasoning budget

aviallon:
This enables support for limiting the maximum reasoning by counting how much tokens we output since entered reasoning mode, and then append a closing sentence + reasoning end token.

The closing sentence can be set in the GGUF metadata, on the command line, or overridden per-request.

**Note:** we don't abort reasoning while emitting a tool call to avoid corrupting the tool call.

_This work was made with the support of [JustAI](https://justai.fr)_

PR #13196, opened 2025-04-29
Support jinja extra template kwargs (Qwen3 enable_thinking feature), from command line and from client

matteoserva:
This PR implements support for setting additional jinja parameters.
An example of this is `enable_thinking` in the Qwen3 models template.

### Main features:
- Setting jinja variables for command line using `--chat_template_kwargs` or the environment variable
- Setting variables per request in the OAI compatible api using the `chat_template_kwargs` parameter
- Compatibility with the [VLLM API](https://docs.vllm.ai/en/v0.8.3/serving/openai_compatible_server.html)

### Notice
- As per https://github.com/ggml-org/llama.cpp/pull/13771 the preferred way for disabling thinking with a command line argument is now `--reasoning-budget 0`. The command line setting can be overridden anyway by passing the `chat_template_kwargs` during the request to the OAI compatible API
- There is ongoing discussion to support setting the reasoning budget per request in https://github.com/ggml-org/llama.cpp/issues/13272. This would allow to completely disable thinking by setting the budget to 0

### Other info
The official template is still only partially compatible. I modified it to use only supported features.
It's here: ~~https://pastebin.com/16ZpCLHk~~ https://pastebin.com/GGuTbFRc
And should be loaded with `llama-server --jinja --chat-template-file {template_file}`

It fixes https://github.com/ggml-org/llama.cpp/issues/13160 and https://github.com/ggml-org/llama.cpp/issues/13189

Test it with:
- enable_thinking=false. Expected: `{"prompt":"\n<|im_start|>user\nGive me a short introduction to large language models.<|im_end|>\n<|im_start|>assistant\n<think>\n\n</think>\n\n"}`
```
curl http://localhost:8080/apply-template -H "Content-Type: application/json" -d '{
  "model": "Qwen/Qwen3-8B",
  "messages": [
    {"role": "user", "content": "Give me a short introduction to large language models."}
  ],
  "temperature": 0.7,
  "top_p": 0.8,
  "top_k": 20,
  "max_tokens": 8192,
  "presence_penalty": 1.5,
  "chat_template_kwargs": {"enable_thinking": false}
}'
```

- enable_thinking=true
```
curl http://localhost:8080/apply-template -H "Content-Type: application/json" -d '{
  "model": "Qwen/Qwen3-8B",
  "messages": [
    {"role": "user", "content": "Give me a short introduction to large language models."}
  ],
  "temperature": 0.7,
  "top_p": 0.8,
  "top_k": 20,
  "max_tokens": 8192,
[...]

Issue #9291, opened 2024-09-03
changelog : `llama-server` REST API


ggerganov:
# Overview

This is a list of changes to the public HTTP interface of the `llama-server` example. Collaborators are encouraged to edit this post in order to reflect important changes to the API that end up merged into the `master` branch.

If you are building a 3rd party project that relies on `llama-server`, it is recommended to follow this issue and check it carefully before upgrading to new versions.

See also:

- [Changelog for `libllama` API](https://github.com/ggerganov/llama.cpp/issues/9289)

## Recent API changes (most recent at the top)

| version | PR  | desc |
| ---     | --- | ---  |
| b4599 | #9639 | `/v1/chat/completions` now supports `tools` & `tool_choice` |
| TBD.  | #10974 | `/v1/completions` is now OAI-compat |
| TBD.  | #10783 | `logprobs` is now OAI-compat, default to pre-sampling probs |
| TBD.  | #10861 | `/embeddings` supports pooling type `none` |
| TBD.  | #10853 | Add optional `"tokens"` output to `/completions` endpoint |
| b4337 | #10803 | Remove `penalize_nl` |
| b4265 | #10626 | CPU docker images working directory changed to /app |
| b4285 | #10691 | (Again) Change `/slots` and `/props` responses |
| b4283 | #10704 | Change `/slots` and `/props` responses |
| b4027 | #10162 | `/slots` endpoint: remove `slot[i].state`, add `slot[i].is_processing` |
| b3912 | #9865 | Add option to time limit the generation phase |
| b3911 | #9860 | Remove self-extend support |
| b3910 | #9857 | Remove legacy system prompt support |
| b3897 | #9776 | Change default security settings, `/slots` is now disabled by default<br/>Endpoints now check for API key if it's set |
| b3887 | #9510 | Add `/rerank` endpoint |
| b3754 | #9459 | Add `[DONE]\n\n` in OAI stream response to match spec |
| b3721 | #9398 | Add `seed_cur` to completion response |
| b3683 | #9308 | Environment variable updated |
| b3599 | #9056 | Change `/health` and `/slots` |

*For older changes, use:*

```bash
git log --oneline -p b3599 -- examples/server/README.md
```

## Upcoming API changes

- TBD

ngxson:
Not a REST API breaking change, but is server-related: some environment variables are changed in https://github.com/ggerganov/llama.cpp/pull/9308

slaren:
[...]

PR #20297, opened 2026-03-09
Handle reasoning budget

pwilkin:
Adds proper handling for `--reasoning-budget`.

Currently, `--reasoning-budget` is just a stub that handles one case: 0, and the only think that does is set enable_thinking to false.

This PR adds the following flags:
* `--reasoning on` (short `-rea on`) - enable reasoning *via kwargs* on model
* `--reasoning off` (short `-rea off`) - disable reasoning *via kwargs* on model
* `--reasoning-budget-message` - a message to be appended before the reasoning close marker to inform the model the reasoning was terminated due to budget constraints, i.e. " ... reasoning budget exceeded" or "... okay, now let's answer."

Also, `--reasoning-budget` now adds an extra grammar with a mechanism called *delayed launch*. Basically what this does is, when the opening trigger for the grammar is triggered, tokens get counted down and we also watch for the *disarm trigger*. If the disarm trigger doesn't get triggered before the countdown, the grammar gets launched.

This allows setting a real token budget limit on reasoning for models. This also allows *disabling thinking for models that normally do not allow that* by setting the budget to 0, which is now *a different behavior than `--disable-reasoning`*. Note: while possible, this *isn't recommended* since if a model was trained to only work with reasoning, it might exhibit aberrant behavior (for example, try to open an extra reasoning section).

Supersedes #17750


pwilkin:
AI disclosure: I used Claude Opus in making most of the changes, auditing and modifying the critical code myself.

pwilkin:
Oh, I didn't mention it in the note but this of course entails support for multiple grammars for one server task since the tool grammar is still there.

pwilkin:
Some interesting observations from early tests (on Qwen3.5 9B Q8_0):
* full model humaneval is around 93%
* non-reasoning (`-dre`) is around 88%
* using reasoning_budget 1000 and 400 is actually pretty similar and improves to about 89%
* **however** this relies on having a `--reasoning-budget-message` (I used " ... reasoning budget exceeded, need to answer."). Without that, performance drops to a *terrible* 79%.

ggerganov (commented):
I'll probably need to understand this deeper, but on first look this seems very heavy logic. How important is this functionality?

Specifically the changes in `common_sampler` seem disproportionately large compared to what this brings to the existing logic. Look for ways to simplify.

pwilkin:
> I'll probably need to understand this deeper, but on first look this seems very heavy logic. How important is this functionality?

A lot of people have been requesting this, especially with the Qwen3.5 models that are seen as too verbose with their reasoning.

The changes in the sampler code are basically to the grammar sampler, since the idea is (a) to support more than one grammar simultaneously and (b) to support delayed grammar application (with token counting). Maybe this can be simplified by instead inserting another grammar sampler? Not sure how viable that would be.

aldehir:
In my opinion, we need to think long term.

The grammar sampler is incredibly inefficient. We had to revert a change @ggerganov wanted to make that shifts the grammar to the start of the chain to support backend sampling.

Merging this will increase the reliance on the grammar sampler and make it more challenging to optimize in the future.

I'm of the opinion that a dedicated, simple, reasoning sampler that lives in common would be enough and can be used at the start of the chain--so long as it aligns with the grammar used (if any).

ggerganov:
Yes, framing this as a reasoning sampler should definitely be explored.

pwilkin:
@ggerganov @aldehir aight, reverted all the grammar changes and instead reimplemented it as a clean new reasoning parser.

I tested on cli, there is no noticeable overhead on generation (152 t/s both with and without the sampler).

common/arg.cpp
```diff
@@ -3048,14 +3052,37 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
...
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_COMPLETION, LLAMA_EXAMPLE_CLI}).set_env("LLAMA_ARG_ENABLE_REASONING"));
+    add_opt(common_arg(
+        {"-dre", "--disable-reasoning"},
+        "explicitly disable reasoning/thinking in the chat template",
+        [](common_params & params) {
+            params.enable_reasoning = 0;
+            params.default_template_kwargs["enable_thinking"] = "false";
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_COMPLETION, LLAMA_EXAMPLE_CLI}).set_env("LLAMA_ARG_DISABLE_REASONING"));
```
CISC:
```suggestion
    add_opt(common_arg(
        {"-rea", "--reasoning"}, "[on|off|auto]",
        "Use reasoning/thinking in the chat ('on', 'off', or 'auto', default: 'auto' (detect from template))",
        [](common_params & params, const std::string & value) {
            if (is_truthy(value)) {
                params.enable_reasoning = 1;
                params.default_template_kwargs["enable_thinking"] = "true";
            } else if (is_falsey(value)) {
                params.enable_reasoning = 0;
                params.default_template_kwargs["enable_thinking"] = "false";
            } else if (is_autoy(value)) {
                params.enable_reasoning = -1;
            } else {
                throw std::invalid_argument(
                    string_format("error: unknown value for --reasoning: '%s'\n", value.c_str()));
            }
        }
    ).set_examples({LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_COMPLETION, LLAMA_EXAMPLE_CLI}).set_env("LLAMA_ARG_REASONING"));
```

common/arg.cpp
```diff
@@ -2913,6 +2913,10 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         [](common_params & params, const std::string & value) {
             auto parsed = json::parse(value);
             for (const auto & item : parsed.items()) {
+                if (item.key() == "enable_thinking") {
+                    LOG_WRN("Setting 'enable_thinking' via --chat-template-kwargs is deprecated. "
+                            "Use --enable-reasoning / --disable-reasoning instead.\n");
```
CISC:
```suggestion
                            "Use --reasoning on / --reasoning off instead.\n");
```

tools/server/server-context.cpp
```diff
@@ -885,9 +885,10 @@ struct server_context_impl {
             }
 
             // thinking is enabled if:
-            // 1. It's not explicitly disabled (reasoning_budget == 0)
+            // 1. It's not explicitly disabled via --disable-reasoning
```
CISC:
```suggestion
            // 1. It's not explicitly disabled via --reasoning off
```

aldehir:
Unless @ggerganov thinks otherwise, I would put it under `common` until it reaches maturity before exposing it in the public API. I imagine there will be quite a bit of churn with all the models to support.

Other notes:

- Is `arm_immediately` needed? Why not define an initial state instead?
- Need to add a soft/hard cap to handle partial utf8  sequences. When I tested this with a grammar approach, I would often see incomplete utf8 codepoints. Instead, we could enforce a soft cap then continue until we hit a clean boundary or hit the hard cap.
- Would like to see some vision for incorporating other reasoning budget strategies. For example, Nemotron Nano 2 (which they claim to also support in their 3-series).

```
3.4. Budget Control Evaluation
Nemotron Nano V2 allows users to specify how many thinking tokens the model may generate before
producing the final answer. The final answer is the portion of text typically shown to end users.
This feature is implemented by counting tokens after the model begins generating the <think>
token. Once the budget is reached, the inference setup attempts to insert a closing </think> tag.
Rather than inserting it immediately, we let the model finish its current sentence and place the
tag at the next newline. In extreme cases where no newline appears, the system enforces closure
within 500 tokens past the budget: if no newline occurs by the (budget+ 500)th token, the </think>
tag is forcibly inserted.
```
https://arxiv.org/abs/2508.14444

Overall, I think this is a cleaner approach. It isolates the complexity rather than polluting the already complex grammar sampling logic.

include/llama.h
```diff
@@ -1455,6 +1455,36 @@ extern "C" {
     //
     LLAMA_API struct llama_sampler * llama_sampler_init_infill(const struct llama_vocab * vocab);
 
+    /// @details Reasoning budget sampler. Limits the number of tokens a model can generate inside
+    ///          a reasoning block (e.g. between <think> and </think>).
+    ///
+    /// State machine: IDLE -> COUNTING -> FORCING -> DONE
+    ///   - IDLE:     passthrough, watching accepted tokens for the start sequence
+    ///   - COUNTING: counts down tokens, watching for natural end (defuse)
```
aldehir: nit: not a fan of the arm/defuse terminology.

aldehir (commented):
Need some tests around the apply/accept logic. I had some in my example, but feel free to improvise.

include/llama.h
pwilkin: Can change, not very attached to it.

aldehir: It seems out of place 🙂

CISC:
Funny, wonder what happened here:
https://github.com/ggml-org/llama.cpp/actions/runs/22877806623/job/66373589713?pr=20297

pwilkin:
> Funny, wonder what happened here: https://github.com/ggml-org/llama.cpp/actions/runs/22877806623/job/66373589713?pr=20297

GitHub merge running on Windows? :D

pwilkin:
Aight I got rid of the explosive terminology and fixed the newlines in the process :)

pwilkin:
Okay, UTF-8 and tests are done, think this one's ready.

src/llama-sampler.cpp
```diff
@@ -3822,6 +3822,283 @@ struct llama_sampler * llama_sampler_init_infill(const struct llama_vocab * voca
...
+
+    // Determine expected continuation bytes from leading byte
+    int expected;
+    if      ((lead & 0xE0) == 0xC0) { expected = 1; }  // 110xxxxx: 2-byte
+    else if ((lead & 0xF0) == 0xE0) { expected = 2; }  // 1110xxxx: 3-byte
+    else if ((lead & 0xF8) == 0xF0) { expected = 3; }  // 11110xxx: 4-byte
+    else { return true; }                               // invalid leading byte
+
+    return n_cont < expected;
+}
```
aldehir:
```suggestion
static bool llama_utf8_is_incomplete(const std::string & s) {
    for (int i = 1; i <= std::min(4, (int)s.size()); i++) {
        unsigned char c = s[s.size() - i];
        if ((c & 0xC0) != 0x80) {
            int expected = (c >= 0xF0) ? 4 : (c >= 0xE0) ? 3 : (c >= 0xC0) ? 2 : 1;
            return i < expected;
        }
    }
    return true;
}
```

aldehir (approved):

common/common.h
```diff
@@ -536,7 +544,9 @@ struct common_params {
     bool use_jinja = true;                                                                                  // NOLINT
     bool enable_chat_template = true;
     common_reasoning_format reasoning_format = COMMON_REASONING_FORMAT_DEEPSEEK;
+    int enable_reasoning = -1; // -1 = auto, 0 = disable, 1 = enable
     int reasoning_budget = -1;
```
ggerganov, line 548: This `reasoning_budget` parameter in `common_params` seems should be removed in favor of `reasoning_budget_tokens` in `common_params_sampling`.

common/common.h
```diff
@@ -536,7 +544,9 @@ struct common_params {
...
+    std::string reasoning_budget_message; // message injected before end tag when budget exhausted
```
ggerganov, line 549: I think `reasoning_budget_message` is rather a sampling parameter, so probably better to move to `common_params_sampling`.

tools/cli/cli.cpp
```diff
@@ -57,6 +57,8 @@ struct cli_context {
     std::vector<raw_buffer> input_files;
     task_params defaults;
     bool verbose_prompt;
+    int reasoning_budget = -1;
+    std::string reasoning_budget_message;
```
ggerganov, line 61: I think you don't need these vars. Just extract the info from `defaults.sampling`

CISC:
@pwilkin This broke the `server` CI:
https://github.com/ggml-org/llama.cpp/actions/runs/22945680157/job/66597350486

jabr:
This is working great for me so far! Thanks @pwilkin!

edit: I seem to get better behavior if the `message` ends with a newline. Without it, the thinking sometimes "escapes" the first close tag and continues on into the content. (With Qwen3.5 0.8B)

erazortt:
I am on b8303 now (this was merged in b8287) and neither --enable-reasoning nor  --disable-reasoning flags are recognized by llama-server, returning an "invalid argument" error. It this expected?

CISC:
> I am on b8303 now and neither --enable-reasoning nor --disable-reasoning flages are recognized by llama-server, retunring an invalid argument. It this expected?

The OP is outdated, the option is `--reasoning on` or `--reasoning off`.

ZUIcat:
I believe the documentation may need an update. I'm currently extremely confused about how to enable or disable the thinking mode of Qwen3.5. Previously, I used `reasoning-budget = 0`, but it seems that's no longer the case? What exactly is `--reasoning`? And how is it different from `chat-template-kwargs = "{\"enable_thinking\":false}"`?

pwilkin:
> What exactly is `--reasoning`? And how is it different from `chat-template-kwargs = "{\"enable_thinking\":false}"`?

`--reasoning` is `enable_thinking` together with a few extra optimizations for handling the thinking path in the parser. So if you want to disable reasoning, use `--reasoning off`, which also sets `enable_thinking = false` in the template.

If your template doesn't support disabling thinking, you can use `--reasoning-budget 0` as the sampler "forcing" solution, but beware that models might not like it (this is basically the equivalent of hacking the template to insert `</think>` at the start of the response).

pwilkin:
> The OP is outdated, the option is `--reasoning on` or `--reasoning off`.

I updated the OP message.

ZUIcat:
> `--reasoning` is `enable_thinking` together with a few extra optimizations for handling the thinking path in the parser. So if you want to disable reasoning, use `--reasoning off`, which also sets `enable_thinking = false` in the template.
> 
> If your template doesn't support disabling thinking, you can use `--reasoning-budget 0` as the sampler "forcing" solution, but beware that models might not like it (this is basically the equivalent of hacking the template to insert `</think>` at the start of the response).

Thank you very much for your explanation. However, I still have a small question. In the previous discussion at <https://github.com/ggml-org/llama.cpp/pull/13196>, it was mentioned that "the preferred way for disabling thinking with a command line argument is now --reasoning-budget 0." Therefore, I have been using only this command to disable thinking.

So now, to disable thinking, do I need to apply both `--reasoning-budget 0` and `--reasoning off`? Or is it sufficient to use only `--reasoning off` without anything else (considering Qwen3.5 only)?

SlavikCA:
~Is it true, that all flags and options mentioned in this thread applies only during starting of llama?~

~Is there an option (header) to use with API request to enable / disable reasoning? For example send one query with thinking disabled and another - with thinking enabled?~

Found that this works to disable thinking on per-request basis:

```
curl 'https://direct.*****/v1/chat/completions' \
 -H "x-api-key: ****" \
 -d '{
    "model": "qwen35-122b",
    "chat_template_kwargs": {"enable_thinking": false},
    "messages": [
        {
        "role": "user",
        "content": "How large is the solar system?"
        }
    ]
}' | jq
```

Now, is there an option I can use to configure `reasoning-budget` per request?

winstonma:
@SlavikCA I can stop the Qwen 3.5 35B A3B from thinking using llama-server and your API call.

But I still no luck disabling the thinking on llama-cli. Here is what I get:

```
❯ llama-cli -m ~/model/Qwen3.5-35B-A3B-GGUF/Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf --reasoning off
ggml_vulkan: 0 = AMD Radeon 890M Graphics (RADV STRIX1) (radv) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 64 | shared memory: 65536 | int dot: 1 | matrix cores: KHR_coopmat

Loading model...  


▄▄ ▄▄
██ ██
                                    ██    ██
                                    ▀▀    ▀▀

build      : b8352-d88ccec
model      : Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf
modalities : text

available commands:


> hi

[Start thinking]
Thinking Process:

1.  **Analyze the Input:**
...
```

pwilkin:
@winstonma Yeah, `cli` reasoning parameter passing is broken. Will be fixed in https://github.com/ggml-org/llama.cpp/pull/20424

winstonma:
@pwilkin Thanks

Didn't know cli and server are not sharing common arguments.

princess-melie:
I cannot figure out how to use this with the OpenAI v1 interface. Is this implemented only as startup parameters? If so, why? Enabling or disabling reasoning seems like something that should be a per-request parameter, since it does not conceptually require restarting the server. A reasoning budget limit appears to have the same semantics and should also logically be configurable per request. I tried passing reasoning_budget_tokens / reasoning_budget in the request extra_body, but it does not appear to have any effect.

I also wanted to clarify how the soft limit is supposed to work. I was not able to fully understand the behavior from the code.
My initial expectation was that the implementation would look something like this: having a `soft_limit` and a `hard_limit`, where after reaching `soft_limit` the logit (or softmax probability) of the `</think>` token would gradually increase. This would allow reasoning to end at a more natural boundary rather than abruptly at an arbitrary point.

For example, the probability of generating `</think>` might look roughly like 
between words: ≈0.001
between sentences: ≈0.01
between paragraphs: ≈0.03
With such behavior, reasoning could still be long, but the gradual increase would encourage the model to terminate reasoning at a structurally appropriate boundary (e.g., between sentences or paragraphs).
From what I can see, the real implementation appears to work differently. I would appreciate an explanation of how the current mechanism actually works internally.

In the approach I described above, --reasoning-budget-message would also become unnecessary, since as generation approaches the hard_limit, the probability of emitting `</think>` would effectively approach infinity and naturally terminate reasoning.

pwilkin:
@princess-melie People have tested the logit probability approach and have reported that it completely does not work. The model totally ignores it until some point, then hard-enforces the end-of-thinking, so it's technically a delayed hard budget.

aviallon:
@pwilkin should this be added to the recent API changes: https://github.com/ggml-org/llama.cpp/issues/9291 issue?

PR #20298, opened 2026-03-09
vulkan: partial revert #20084

jeffbolznv:
Should resolve the llvmpipe failures.


0cc4m:
If the barriers are needed, they should stay there. #20294

PR #20301, opened 2026-03-09
llama: dynamic head_dim and n_rot for SWA

ngxson:
Add the ability to have `n_embd_head_k/v` and `n_rot` differently for full_attention and SWA layers.

At the time of writing this PR:
- None of the model currently using different `n_embd_head_k/v` (this is to support future models)
- Step3.5 is the only one that uses different `n_rot` for SWA

Impacts to components:
- `llama-kv-cache` uses `n_embd_head_k/v` for view/set_row, so now they will be different for each layer
- `llama-kv-cache` uses `n_rot` for K shifting --> now it can shift correctly
- `llama-model` loads and log these params --> default params for SWA = params for full, but each model can overwrite it


src/llama-kv-cache.h
ggerganov: I have a small preference to put the `il` at the end of the argument list:


```suggestion
ggml_tensor * build_rope_shift(
const llama_cparams & cparams,
ggml_context * ctx,
ggml_tensor * cur,
ggml_tensor * shift,
ggml_tensor * factors,
float   freq_base,
float   freq_scale,
uint32_t   il) const;
```

ggerganov (approved):

CISC (approved):

ASDF2333-FDSA:
This PR seems to break compatibility with older ggufs for Step-3.5-Flash. When using the Q4_K_M quant from [AesSedai](https://huggingface.co/AesSedai/Step-3.5-Flash-GGUF) with the latest commit, I got 
```
llama_model_load: error loading model: check_tensor_dims: tensor 'rope_freqs.weight' has wrong shape; expected     32,
 got     64,      1,      1,      1
 ```

ngxson:
@ASDF2333-FDSA could you try this fix: https://github.com/ggml-org/llama.cpp/pull/20318

PR #20316, opened 2026-03-09
opencl: use larger workgroup size for get_rows

lhez:
Choose the largest possible workgroup size for get_rows for better performance plus small refactor.


max-krasnyansky (approved):

lhez:
Failures are irrelevant. Will merge shortly.

PR #20318, opened 2026-03-09
model: fix step3.5 n_rot

ngxson:
cont https://github.com/ggml-org/llama.cpp/pull/20301


CISC (approved):

ASDF2333-FDSA:
The fix seems to work!

PR #20323, opened 2026-03-10
cuda: remove incorrect assertion in MMQ for batched mul_mat_id

camilorodegheri:
## Summary

Remove incorrect `GGML_ASSERT(ne1 == n_expert_used)` in the MMID (MoE expert dispatch) path of `ggml_mul_mat_q()`.

## Problem

For batched inference with MoE models, `ne1 = n_tokens * n_expert_used`, but the assertion expects `ne1 == n_expert_used` (single-token case). This causes a crash on any MoE model with batch size > 1.

Fixes #19705.

## Analysis

The assertion is simply wrong for batch > 1. The code **immediately below** already correctly handles batched MoE:

- `ne_get_rows = ne12 * n_expert_used` (accounts for batch size)
- All buffer allocations use `ne_get_rows`
- The `mmq_args` struct passes `ne_get_rows` for both `ncols_dst` and `ncols_y`
- `ne1` is **never used** anywhere else in the MMID path — only in this assertion

The MMQ kernel itself already handles batched MoE correctly. The assertion was a debug check that was never tested with batch > 1.

## Testing

Tested on NVIDIA DGX Spark GB10 (Blackwell, sm_121a, 128GB unified memory) with Qwen3-Coder-Next (80B, 512 experts, 10 active, Q8_0):

- **Before**: `GGML_ASSERT` crash on model load/warmup
- **After**: Model loads, MMQ runs for MoE at full speed
  - Prompt: 175 tok/s
  - Generation: 34 tok/s
  - No fallback to cuBLAS needed — MMQ handles it directly

**Note**: Qwen3-Coder-Next has a separate, unrelated generation quality issue (#19305) that is independent of this fix.

## Impact

- **MoE models**: MMQ now works for batched inference (previously crashed)
- **Dense models**: Zero impact (`n_expert_used` path not taken)
- **Performance**: No regression — this preserves the MMQ fast path instead of falling back to cuBLAS

### DIFF
--- a/ggml/src/ggml-cuda/mmq.cu
+++ b/ggml/src/ggml-cuda/mmq.cu
@@ -163,7 +163,6 @@ void ggml_cuda_mul_mat_q(
 
     const int64_t n_expert_used = ids->ne[0];
     const int64_t ne_get_rows = ne12 * n_expert_used;
-    GGML_ASSERT(ne1 == n_expert_used);
 
     ggml_cuda_pool_alloc<int32_t> ids_src1(ctx.pool(), ne_get_rows);
     ggml_cuda_pool_alloc<int32_t> ids_dst(ctx.pool(), ne_get_rows);


am17an:
This is completely wrong.

PR #20329, opened 2026-03-10
cli : fix --reasoning-budget and --chat-template-kwargs being ignored

TrevorS:
The autoparser refactor (#18675) replaced `chat_params.enable_thinking` with `common_chat_templates_support_enable_thinking()`, which only checks template support and ignores user flags. Also missing was the `chat_template_kwargs` passthrough.

Fixes #20182


pwilkin:
Already handled by #20297

PR #20270, opened 2026-03-09
models : fix assert in mamba2 graph

ggerganov:
cont #19802 
fix #20268

This fixes the model loading, but the reasoning parsing seems to be broken because the post-reasoning contents are not being displayed in the WebUI:

image

cc @pwilkin

PR #20335, opened 2026-03-10
models : fix assert in mamba2 (cont)

ggerganov:
cont #20270

I think we actually only need `d_inner % d_state == 0`.

### DIFF
--- a/src/models/mamba-base.cpp
+++ b/src/models/mamba-base.cpp
@@ -170,6 +170,7 @@ ggml_tensor * llm_build_mamba_base::build_mamba2_layer(llm_graph_input_rs * inp,
     GGML_ASSERT(ubatch.n_tokens == n_seq_tokens * n_seqs);
     GGML_ASSERT(d_inner % n_head  == 0);
     GGML_ASSERT(d_inner % d_state == 0);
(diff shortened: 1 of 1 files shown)


src/models/mamba-base.cpp
```diff
@@ -168,8 +168,8 @@ ggml_tensor * llm_build_mamba_base::build_mamba2_layer(llm_graph_input_rs * inp,
     GGML_ASSERT(n_seqs != 0);
     GGML_ASSERT(ubatch.equal_seqs());
     GGML_ASSERT(ubatch.n_tokens == n_seq_tokens * n_seqs);
-    GGML_ASSERT(d_inner % n_head == 0);
-    GGML_ASSERT(d_inner % (n_group*d_state) == 0);
+    GGML_ASSERT(d_inner % n_head  == 0);
+    GGML_ASSERT(d_inner % d_state == 0);
```
CISC, line 172:
```suggestion
    GGML_ASSERT(d_inner % d_state == 0);
    GGML_ASSERT(d_inner % n_group == 0);
```
Though I guess this is sort of covered at tensor load already.

PR #20346, opened 2026-03-10
Refactor llama_model_quantize_params to expose a pure C interface

EAddario:
@ggerganov, as the saying goes, [better late than never](https://github.com/ggml-org/llama.cpp/pull/12511#discussion_r2032541334) 🙃


tools/quantize/quantize.cpp
```diff
@@ -13,13 +13,9 @@
 #include <unordered_map>
 #include <map>
 #include <fstream>
-#include <cmath>
-#include <cctype>
-#include <algorithm>
 #include <filesystem>
 
-// result of parsing --tensor-type option
```
ddh0, line 21: Excuse me for being picky - could you please restore this line? I left the comment to explain what the struct actually is to future readers.

ggerganov (approved):
Thanks, let's wait for the CI to pass

include/llama.h
```diff
@@ -380,22 +380,33 @@ extern "C" {
         size_t                            n_samplers;
     };
 
+    struct llama_model_tensor_override {
+        const char * pattern;
+        enum ggml_type type;
+    };
+
+    struct llama_imatrix_data {
```
ggerganov: Let's put this in the `llama_model` namespace:


```suggestion
    struct llama_model_imatrix_data {
```

include/llama.h
```diff
@@ -380,22 +380,33 @@ extern "C" {
...
+        enum ggml_type token_embedding_type;                        // token embeddings tensor type
+        bool allow_requantize;                                      // allow quantizing non-f32/f16 tensors
+        bool quantize_output_tensor;                                // quantize output.weight
+        bool only_copy;                                             // only copy tensors - ftype, allow_requantize and quantize_output_tensor are ignored
+        bool pure;                                                  // quantize all tensors to the default type
+        bool keep_split;                                            // quantize to the same number of shards
+        bool dry_run;                                               // calculate and show the final quantization size without performing quantization
+        const struct llama_imatrix_data * imatrix;                  // pointer to importance matrix data
+        const struct llama_model_kv_override * kv_overrides;        // pointer to kv overrides
+        const struct llama_model_tensor_override * tensor_types;    // pointer to tensor overrides
```
ggerganov: Improve naming consistency:

```suggestion
        const struct llama_model_kv_override * kv_overrides;        // pointer to kv overrides
        const struct llama_model_tensor_override * tt_overrides;    // pointer to tensor overrides
```

Also, in the rest of the code: `t_override` -> `tt_override`.

PR #20350, opened 2026-03-10
vulkan: fix l2_norm epsilon handling

jeffbolznv:
This is the only "real" bug I could find in test-llama-archs. I see some other failures but they may be driver/compiler bugs.


jeffbolznv:
Oh, and regarding the intermittent failures in CI - the llvmpipe ones may be due to the MUL_MAT failures, and failures on Turing may be the same compiler bug we're seeing in test-backend-ops.

0cc4m (approved):

PR #20355, opened 2026-03-10
Make tests windows compatible.

srikris-sridhar:
On Windows, Python outputs \r\n line endings, but the test expects \n. This caused string comparisons to fail.

Fix: Added normalize line endings before comparing:
      jinja::string_replace_all(output, "\r\n", "\n");

This replaces Windows line endings (\r\n) with Unix line endings (\n) so the comparison works cross-platform.


CISC:
Please try #20365 instead.

srikris-sridhar:
Closing in favor our #20365

PR #20365, opened 2026-03-10
tests : write to binary buffer to avoid newline translation in jinja -py

CISC:
Write directly to `stdout` binary buffer for Python tests to avoid newline translation.

Should fix test on Windows.


CISC:
gentle ping, this together with #20483 will resolve interoperability issues with the test.

ggerganov (approved):

PR #16837, opened 2025-10-29
clip : use FA

ggerganov:
ref https://github.com/ggml-org/llama.cpp/pull/13231#issuecomment-3405623958

Sample implementation for using FA in the CLIP. Reduces memory usage and improves performance.

Testing with Gemma 12B, using `llama-server` and 2 images:

```bash
# before
alloc_compute_meta:      Metal compute buffer size =  1132.00 MiB
alloc_compute_meta:        CPU compute buffer size =     9.19 MiB

srv  process_chun: image processed in 1653 ms
srv  process_chun: image processed in 1093 ms

# after
alloc_compute_meta:      Metal compute buffer size =   121.25 MiB
alloc_compute_meta:        CPU compute buffer size =     9.19 MiB

srv  process_chun: image processed in 1386 ms
srv  process_chun: image processed in 810 ms
```

TODO:

- [ ] Add FA sizes to other backends (f.ex. Gemma uses non-standard head size of 72)

PR #20375, opened 2026-03-11
ggml-cuda : add flash attention support for head size 88 (Llama 4 Vision)

caffeinatedbits:
Llama 4 Vision models utilize a projector head dimension of `D=88`. 

Currently, `ggml-cuda` does not support `D=88` in its flash attention dispatch tables. This causes a failure during warmup (`flash attention not supported by CUDA0`), forcing the backend to fall back to unoptimized operations (e.g., `SOFT_MAX`, `PERMUTE`, `CONCAT`). This results in massive VRAM bloat and severe inference slowdowns, effectively breaking GPU support for these models.

*(Note: This addresses the backend/memory warnings referenced in #16837, but does not affect the known logic/quality degradation issues discussed in #13282).*

### Changes

This PR adds `D=88` support to the CUDA backend via the TILE kernel:

- `ggml/src/ggml-cuda/fattn.cu`: Added `88` to the allowed dimensions switch. Explicitly blacklisted `88` from Turing/Volta/WMMA/MMA Tensor Core checks. Because 88 is not cleanly divisible by 16, pushing it through Tensor Cores causes hardware misalignment and segfaults. It now safely falls through to the TILE kernel.
- `ggml/src/ggml-cuda/fattn-tile.cuh` & `.cu`: Defined specific memory alignment configurations for `88` across all hardware architectures and wired the dispatch case.
- `generate_cu_files.py`: Added `88` to `HEAD_SIZES_KQ` to dynamically generate `fattn-tile-instance-dkq88-dv88.cu`.

### Results

- Llama 4 Vision models now successfully pass the warmup phase with `flash attention is enabled`.
- Unoptimized graph fallback is eliminated.
- VRAM usage returns to expected levels.
- Fully offloads to CUDA without relying on CPU fallback.


JohannesGaessler (commented):
Setting `nbatch_K` to 88 is potentially problematic in terms of register pressure but I don't remember whether a value of 44 would work correctly. Please post performance numbers for the version you submitted here.

caffeinatedbits:
@JohannesGaessler 

Thanks for the review. You make a fair point regarding the potential for high register pressure with nbatch_K = 88. My primary objective with this implementation was to establish a stable, direct path to unblock Llama 4 Vision models, which are currently failing entirely on the CUDA backend due to the massive VRAM bloat caused by falling back to unoptimized operations (SOFT_MAX, PERMUTE, CONCAT).

For context on the performance, I have been profiling this on an NVIDIA DGX Spark (Grace-Blackwell GB10 APU, Compute 12.1, 128GB Unified Memory) running a 109B MoE Llama 4 Vision model (q8_0 KV cache, 256k context window).

**Baseline (Without PR):**
- Warmup fails: flash attention not supported by CUDA0
- Extreme VRAM bloat due to the unsupported operations graph.
- Forced CPU fallback or outright OOM crashes depending on the context size.

**With this PR (nbatch_K = 88):**
- Warmup succeeds: warmup: flash attention is enabled
- 100% GPU offload (49/49 layers) with stable, predictable VRAM consumption.
- Inference is stable and fully utilizes the CUDA backend without spilling to CPU.

Because Blackwell handles the register pressure well, the performance on this specific hardware is excellent. However, I understand the penalty for register spilling could be much higher on older architectures. I haven't tested nbatch_K = 44 to verify if the memory alignment holds up correctly through the tile kernel without throwing a segfault, but I am happy to test it and update the PR if you believe it is the safer route for broader hardware compatibility.

Let me know how you would like to proceed.

JohannesGaessler:
I would like to proceed with what I posted before:

>Please post performance numbers for the version you submitted here.

PR #20361, opened 2026-03-10
metal : add GDN kernel

ggerganov:
target #20340 
cont #20244 

Add fused GDN recurrent kernel. Use both for BS == 1 and BS > 1.

| Model                      | Test   |   t/s master |   t/s gg/metal-add-gdn |   Speedup |
|:---------------------------|:-------|-------------:|-----------------------:|----------:|
| kimi-linear 48B.A3B Q4_K_M | pp512  |       613.65 |                1529.10 |      2.49 |
| kimi-linear 48B.A3B Q4_K_M | pp2048 |       654.58 |                1955.46 |      2.99 |
| kimi-linear 48B.A3B Q4_K_M | tg32   |        68.19 |                  85.61 |      1.26 |
| qwen35 27B Q8_0            | pp512  |       349.12 |                 390.39 |      1.12 |
| qwen35 27B Q8_0            | pp2048 |       363.75 |                 406.81 |      1.12 |
| qwen35 27B Q8_0            | tg32   |        17.03 |                  20.36 |      1.20 |
| qwen35moe 35B.A3B Q4_0     | pp512  |      1612.12 |                2058.31 |      1.28 |
| qwen35moe 35B.A3B Q4_0     | pp2048 |      1879.76 |                2462.35 |      1.31 |
| qwen35moe 35B.A3B Q4_0     | tg32   |        57.08 |                  77.65 |      1.36 |
| qwen3next 80B.A3B Q4_K_M   | pp512  |      1002.26 |                1172.08 |      1.17 |
| qwen3next 80B.A3B Q4_K_M   | pp2048 |      1222.72 |                1477.06 |      1.21 |
| qwen3next 80B.A3B Q4_K_M   | tg32   |        46.58 |                  60.67 |      1.30 |

PR #20376, opened 2026-03-11
vulkan: f16 mixed-precision state for GATED_DELTA_NET

ProgenyAlpha:
Follow-up to #20334. Splits out the f16 mixed-precision state optimization into its own PR per @0cc4m's feedback.

Stores the 128-element state array in `float16_t`, keeps all arithmetic in `float32`. No precision loss (13/13 backend-ops tests passing). Lower register pressure gives a measurable PP boost.

**Depends on #20334**

**890M benchmarks (Qwen3-Coder-Next REAM Q4_K_M):**
| Metric | Without f16 | With f16 | Change |
|--------|------------|----------|--------|
| PP-512 | 165.31 t/s | 174.54 t/s | **+5.6%** |
| TG-128 | 21.16 t/s | 21.48 t/s | **+1.5%** |

f16 pipeline auto-selects when the device supports `shaderFloat16`, falls back to f32 otherwise.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -827,6 +827,7 @@ struct vk_device_struct {
     vk_pipeline pipeline_rwkv_wkv7_f32;
     // [size_idx][kda] where size_idx: 0=d32, 1=d64, 2=d128
     vk_pipeline pipeline_gated_delta_net[3][2];
+    vk_pipeline pipeline_gated_delta_net_f16[3][2];
     vk_pipeline pipeline_ssm_scan_f32_d128;
     vk_pipeline pipeline_ssm_scan_f32_d256;
     vk_pipeline pipeline_ssm_conv_f32;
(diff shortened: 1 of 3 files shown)


0cc4m:
Please rebase and resolve the conflicts.

ProgenyAlpha:
> Please rebase and resolve the conflicts.

Working on this now.

ggerganov:
Btw, I'm not super confident that this cast is safe in terms of quality. Since this is a recurrent state, even small deviations can accumulate to large errors with time.

0cc4m:
> Btw, I'm not super confident that this cast is safe in terms of quality. Since this is a recurrent state, even small deviations can accumulate to large errors with time.

Do you know if the arithmetic could be moved to fp16 if the state stays in fp32? Is there a good way to find out what is safe and what isn't?

ProgenyAlpha:
> Btw, I'm not super confident that this cast is safe in terms of quality. Since this is a recurrent state, even small deviations can accumulate to large errors with time.

> Do you know if the arithmetic could be moved to fp16 if the state stays in fp32? Is there a good way to find out what is safe and what isn't?

This was more of a first pass to see if the f16 state approach was worth pursuing at all (the register pressure reduction did give measurable PP gains on my 890M). But given that CUDA and Metal both keep state in f32 for this op, I wanted to do more homework before committing to this path, anyway.

Next steps from my end:
  - Run perplexity comparisons at longer context lengths (2048+) to see if the drift is actually measurable in practice
  - Look into the inverse approach you suggested since those intermediate values won't accumulate across tokens

  I can also close out the PR if you prefer?

ggerganov:
To reduce the register pressure, implement the sharded approach as demonstrated in #20391 and #20361.

> Do you know if the arithmetic could be moved to fp16 if the state stays in fp32? Is there a good way to find out what is safe and what isn't?

@0cc4m I'm not sure - don't have a good intuition about the recurrent state yet.

The change could be fine - I'm just not really sure.

ProgenyAlpha:
> To reduce the register pressure, implement the sharded approach as demonstrated in #20391 and #20361.
> 
> @0cc4m I'm not sure - don't have a good intuition about the recurrent state yet.
> 
> The change could be fine - I'm just not really sure.

I'll look into the sharded approach from #20391 for Vulkan. I had sharding on my todo list already, but held off on opening another PR due to the new policy. Will do some benchmarks and validity testing first, so I'm not wasting your time. May end up closing this one out depending on results.

jeffbolznv:
My hunch is that spreading the values across more invocations and/or shared memory will be better. The "shape" of the algorithm is similar enough to ssm_scan that it seems like the same techniques should work.

PR #20378, opened 2026-03-11
[WebGPU] Reduce kernel profiling overhead

nikhilJain17:
Profiling kernels in the WebGPU backend incurs significant overhead due to eagerly copying timestamp buffers from device to host, and waiting for profiling futures on kernel submission completion. 

Instead, lazily resolve timestamp queries in batches of 32, and wait for profiling futures only after timestamp query has been resolved. Also resolve any straggling queries on device teardown.

The ideal solution would be to batch all timestamp queries and only resolve on device teardown, but we have ~2700 queries in flight over the whole computation graph (i.e. when testing with llama-3.1) and cannot allocate more than 32 query sets at a time.

Also tried increasing timestamp buffer pool size as well to increase batch size, but seems like the maximum allowed query sets is 32.

Screenshot 2026-03-13 at 4 57 44 PM

Screenshot 2026-03-30 at 1 57 37 PM

Overhead is defined as the difference between end to end latency with only CPU profiling vs end to end latency with CPU + GPU profiling, for run _i_. The first difference is partly due to cold start vs warm start, but overhead is reduced overall. Negative overhead means the CPU + GPU profiling run happened to complete faster than the CPU only profiling run, due to variance -- i.e. one output may simply be generated faster than another one on subsequent runs.


ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -605,32 +659,37 @@ static webgpu_submission ggml_backend_webgpu_submit(webgpu_global_context &
...
-                    const uint64_t * ts_data    = (const uint64_t *) ts_bufs.host_buf.GetConstMappedRange();
-                    // WebGPU timestamps are in ns; convert to ms
-                    double           elapsed_ms = double(ts_data[1] - ts_data[0]) * 1e-6;
-                    ctx->shader_gpu_time_ms[label] += elapsed_ms;
-                }
-                // We can't unmap in here due to WebGPU reentrancy limitations.
-                ctx->timestamp_query_buf_pool.free_bufs({ ts_bufs });
-            });
-        submission.profile_futures.push_back({ f });
+    if (ctx_opt.has_value()) {
```
nikhilJain17, line 661: For the buffer operations, we don't have thread level storage so we don't do profiling for them, since I moved the GPU profiling buffers from global context to per-thread context. Wdyt?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -604,32 +658,39 @@ static webgpu_submission ggml_backend_webgpu_submit(webgpu_global_context &
...
```
reeselevine, line 661: instead of doing the timestamp resolution within this `submit` function, could we move to the same strategy as the parameter buffers, and do the resolution/blocking in `ggml_backend_webgpu_wait`? That would avoid having to pass `ctx_opt` everywhere, which imo makes the code harder to understand.

reeselevine (commented):
I'm seeing 0s in the shader breakdown when I run this on my machine:

```
ggml_webgpu: adapter_info: vendor_id: 4203 | vendor: apple | architecture: metal-3 | device_id: 0 | name: Apple M3 | device_desc: Metal driver on macOS Version 26.3.1 (a) (Build 25D771280a)
| llama 1B Q4_0                  | 729.75 MiB |     1.24 B | WebGPU     |  99 |  1 |           pp512 |       1003.50 ± 1.28 |

[ggml_webgpu gpu profiling summary]
ggml_webgpu: total gpu time (all shaders): 0 ms

ggml_webgpu: gpu breakdown:
ggml_webgpu:  mul_mat_vec_q6_K_f32: 0 ms (0.00%)
ggml_webgpu:  rms_norm_inplace: 0.00 ms (0.00%)
ggml_webgpu:  mul_mat_vec_q4_0_f32: 0.00 ms (0.00%)
ggml_webgpu:  mul_mat_subgroup_matrix_q4_1_f32: 0.00 ms (0.00%)
ggml_webgpu:  swiglu_f32_split: 0.00 ms (0.00%)
ggml_webgpu:  flash_attn_f16_mask_kvdirect_hsqk64_hsv64: 0.00 ms (0.00%)
ggml_webgpu:  ADD_f32_inplace: 0.00 ms (0.00%)
ggml_webgpu:  cpy_f32_f16: 0.00 ms (0.00%)
ggml_webgpu:  set_rows_dstf16_vec4_i64idx: 0.00 ms (0.00%)
ggml_webgpu:  rope_f32_ff: 0.00 ms (0.00%)
ggml_webgpu:  mul_mat_subgroup_matrix_q4_0_f32: 0.00 ms (0.00%)
ggml_webgpu:  MUL_f32_inplace: 0.00 ms (0.00%)
ggml_webgpu:  get_rows_f32_vec: 0.00 ms (0.00%)
ggml_webgpu:  rms_norm: 0.00 ms (0.00%)
| llama 1B Q4_0                  | 729.75 MiB |     1.24 B | WebGPU     |  99 |  1 |           tg128 |         72.18 ± 0.06 |

[ggml_webgpu gpu profiling summary]
ggml_webgpu: total gpu time (all shaders): 0.00 ms

ggml_webgpu: gpu breakdown:
ggml_webgpu:  mul_mat_vec_q6_K_f32: 0.00 ms (0.00%)
ggml_webgpu:  rms_norm_inplace: 0.00 ms (0.00%)
ggml_webgpu:  get_rows_f32_vec: 0.00 ms (0.00%)
ggml_webgpu:  mul_mat_vec_q4_1_f32: 0.00 ms (0.00%)
ggml_webgpu:  flash_attn_f16_mask_kvdirect_hsqk64_hsv64: 0.00 ms (0.00%)
ggml_webgpu:  ADD_f32_inplace: 0.00 ms (0.00%)
ggml_webgpu:  cpy_f32_f16: 0.00 ms (0.00%)
ggml_webgpu:  set_rows_dstf16_vec4_i64idx: 0.00 ms (0.00%)
ggml_webgpu:  rope_f32_ff: 0.00 ms (0.00%)
ggml_webgpu:  MUL_f32_inplace: 0.00 ms (0.00%)
ggml_webgpu:  swiglu_f32_split: 0.00 ms (0.00%)
ggml_webgpu:  mul_mat_vec_q4_0_f32: 0.00 ms (0.00%)
ggml_webgpu:  rms_norm: 0.00 ms (0.00%)

build: 42d1daadd (8613)
```

nikhilJain17:
> I'm seeing 0s in the shader breakdown when I run this on my machine:  

Which command are you using to test? I have been testing with `llama-cli --hf-repo ggml-org/Meta-Llama-3.1-8B-Instruct-Q4_0-GGUF --hf-file meta-llama-3.1-8b-instruct-q4_0.gguf -p "Hello world" -n 100`.

> I'm also working on a larger change in https://github.com/reeselevine/llama.cpp/tree/one-buffer that removes the need for a parameter buffer pool completely. I'm not totally sure how to integrate the profiling with this change effectively yet though.


Sounds good, I looked through your changes. It seems the main change in submit/wait is to wait on the entire queue all at once instead of waiting on individual kernel submission futures, which greatly simplifies the code but loses the kernel submission handles. If we want to preserve kernel-level profiling, we would need to keep our own data structure in `graph_compute()` for which kernels were submitted in each batch. Then, we would process the corresponding timestamp buffers in `graph_compute()` after `wait()` (and not within `wait_queue()` as kernels finish, since the `WaitAny` callback doesn't tell us which submitted work completed.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
nikhilJain17: Sounds good, I can make that change. The reason we pass around `ctx_opt` is because the profiling buffers are in the per-thread context, not the global context, so regardless we would have to move the buffers to the global context to do so. Related to: https://github.com/ggml-org/llama.cpp/pull/20378/changes#r2955515365 -- can just undo these changes.

PR #20385, opened 2026-03-11
common/autoparser : use nlohmann::ordered_json to preserve parameter order

aldehir:
The autoparser used `nlohmann::json`, which does not preserve order of parameters in tools. This confuses the model, as it attempts to emit parameters in the same order as the JSON schema but is grammar constrained from doing so.

Wondering if this was the underlying issue in #20171.

Renamed `json` to `ordered_json` alias since it was conflicting with the `json()` method.

fixes #20359

### DIFF
--- a/common/chat-peg-parser.cpp
+++ b/common/chat-peg-parser.cpp
@@ -6,7 +6,7 @@
 
 #include <nlohmann/json.hpp>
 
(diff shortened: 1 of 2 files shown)


CISC (approved):

pwilkin (approved):
Okay, that was a dumb mistake 😐

PR #20388, opened 2026-03-11
ggml-cpu: fix x86 SIMD detection — validate OS YMM/ZMM/AMX state via XCR0

Mattbusel:
## Problem

`ggml_backend_cpu_x86_score()` contained this comment:

```cpp
// FIXME: this does not check for OS support
```

Every AVX/AVX2/AVX-512/AMX feature check in `cpuid_x86` was a raw CPUID bit read with no OSXSAVE or XGETBV validation. Example:

```cpp
bool AVX2(void) { return f_7_ebx[5]; }  // ← CPUID bit only, no XCR0 check
```

**Why this crashes:** A CPU can report AVX2 support in CPUID while the OS has not enabled YMM register save/restore. This happens in:
- Certain hypervisors (GCP, some AWS instance types with stripped CPUID masks)
- Containers with restricted `XSAVE` permissions
- Windows builds where AVX context saving is disabled
- Older kernels that don't set YMM bits in XCR0

In these environments, ggml selects an AVX2 or AVX-512 backend, then faults with `SIGILL` on the first vector instruction.

## Fix

Adds `xgetbv()` (MSVC `_xgetbv` / GCC-Clang inline asm) and three OS-state predicates to `cpuid_x86`, then gates all affected feature methods on them:

```cpp
// Validates CPUID.1:ECX[27] (OSXSAVE) + XCR0[2:1] == 0b11
bool os_saves_ymm();  // gates: AVX, AVX2, FMA, F16C, AVX_VNNI

// Validates os_saves_ymm() + XCR0[7:5] == 0b111
bool os_saves_zmm();  // gates: all AVX-512 variants

// Validates os_saves_zmm() + XCR0[18:17] == 0b11
bool os_saves_amx();  // gates: AMX_TILE/INT8/FP16/BF16
```

After the fix:

```cpp
bool AVX2(void) { return f_7_ebx[5] && os_saves_ymm(); }  // ✓ correct
```

## What didn't change

- Zero interface changes — all call sites continue using `is.AVX2()` etc.
- `ggml_backend_cpu_x86_score()` is unchanged except the FIXME is removed
- No other files touched

## Relation to PR #19514

PR #19514 opened 2026-02-11 addresses the same issue. This PR takes the same approach (adding an `xgetbv` helper and gating feature checks on OS state) but applies the fix directly to the existing `cpuid_x86` methods rather than adding a separate struct, keeping the diff minimal and the call sites unchanged.

Designed and implemented by Matthew Busel.

### DIFF
--- a/ggml/src/ggml-cpu/arch/x86/cpu-feats.cpp
+++ b/ggml/src/ggml-cpu/arch/x86/cpu-feats.cpp
@@ -6,6 +6,13 @@
 #include <intrin.h>
 #endif
 
+// On Darwin/macOS, AVX-512 context save is lazy: XCR0 bits 5-7 are not set
+// until the process first executes an AVX-512 instruction, even on capable
+// hardware.  We query the OS via sysctl instead of reading XCR0 directly.
+#if defined(__APPLE__)
+#include <sys/sysctl.h>
+#endif
+
 #include <cstring>
 #include <vector>
 #include <bitset>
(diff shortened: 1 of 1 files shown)


vsrinivas:
Is this correct for Darwin/x86 on AVX512 capable machines? AVX512 is masked in XCR0 by default, lazily enabled on first use. You need to check either the sysctl or commpage there.

Mattbusel:
Good catch. Pushed a Darwin path in the latest commit: os_saves_zmm() now queries sysctl hw.optional.avx512f on __APPLE__ instead of reading XCR0. YMM and AMX paths unchanged since macOS eagerly enables YMM and has no AMX hardware. Same approach as google/cpu_features.

PR #20392, opened 2026-03-11
hip: compile debug builds with -O2 on hip to avoid a compiler bug

IMbackK:
see the travesty in https://github.com/ROCm/ROCm/issues/5826 for why this is required.


am17an (approved):
lol

0cc4m:
lol indeed

PR #20393, opened 2026-03-11
common : rework gpt-oss parser

aldehir:
Rework the gpt-oss parser.

- Tighten up the grammar, gpt-oss is very good at following its own Harmony spec.
- Allow any sequence of analysis/preamble.
- Clean up the trigger rules, gpt-oss may sometimes invoke a builtin function if not constrained when emitting the tool namespace.
- Include fix from #19704.
- Fix `response_format` not being enforced.
- Remove parallel call logic, gpt-oss is trained to only emit single tool calls.
- Removed support for `reasoning-format = none`. It makes no sense for gpt-oss. Users can choose to ignore `reasoning_content`.
- Remove invalid test cases.

fixes #20344
fixes #20500

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -933,17 +933,12 @@ static common_chat_params common_chat_params_init_gpt_oss(const common_chat_temp
 
     // Copy reasoning to the "thinking" field as expected by the gpt-oss template
     auto adjusted_messages = json::array();
-    for (const auto & msg : inputs.messages) {
-        auto has_reasoning_content = msg.contains("reasoning_content") && msg.at("reasoning_content").is_string();
-        auto has_tool_calls        = msg.contains("tool_calls") && msg.at("tool_calls").is_array();
-
-        if (has_reasoning_content && has_tool_calls) {
-            auto adjusted_message        = msg;
-            adjusted_message["thinking"] = msg.at("reasoning_content");
-            adjusted_messages.push_back(adjusted_message);
-        } else {
-            adjusted_messages.push_back(msg);
+    for (auto msg : inputs.messages) {
+        if (msg.contains("reasoning_content") && msg.at("reasoning_content").is_string()) {
+            msg["thinking"] = msg.at("reasoning_content");
+            msg.erase("content");
         }
+        adjusted_messages.push_back(msg);
     }
 
     auto prompt = common_chat_template_direct_apply(tmpl, inputs, /* messages_override= */ adjusted_messages);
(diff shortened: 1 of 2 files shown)


common/chat.cpp
```diff
@@ -969,45 +962,31 @@ static common_chat_params common_chat_params_init_gpt_oss(const common_chat_temp
...
-                   p.content(p.schema(p.json(), "response-format", inputs.json_schema));
+        auto start           = p.rule("start", p.literal("<|start|>assistant"));
+        auto end             = p.rule("end", p.literal("<|end|>"));
+        auto content         = p.rule("message-content", p.until("<|end|>"));
+        auto channel         = p.literal("<|channel|>") + (p.literal("commentary") | p.literal("analysis"));
+        auto constrain_type  = p.chars("[A-Za-z0-9_-]", 1, -1);
+
+        auto analysis = p.rule("analysis", p.literal("<|channel|>analysis<|message|>") + p.reasoning(content) + end);
+        auto preamble = p.rule("preamble", p.literal("<|channel|>commentary<|message|>") + p.content(content) + end);
+        auto final = p.rule("final", p.literal("<|channel|>final<|message|>") + p.content(content));
```
pwilkin: Final is a keyword, does this work correctly? I'd change it anyway just in case.

pwilkin (approved):
I'll trust you on this one :)

pwilkin:
Just two issues:
-> support for `reasoning_format: none` should stay, since some people want just raw template output to further process with their own custom parsers / apps (I think SillyTavern does stuff like that) and not outputting the reasoning in content breaks this
-> since you're saying this fixes structured output, better add a test for structured output :)

common/chat.cpp
aldehir: It does, which is why I kept it. But I'll change it. Most syntax highlighters are not semantic aware so it'll highlight it as a keyword.

aldehir:
I'll add the logic back in, but it truly makes no sense. For one, the official template will throw an exception if it sees any harmony tags in the message. Ultimately, this model is incapable of not reasoning. Even when constrained, it will leak reasoning traces inside the final response if it can.

That said, if clients depend on this then I guess 🤷‍♂️

pwilkin:
@aldehir `reasoning_format: none` is not `enable_thinking: false`, it just means you don't process the reasoning tags and you spill them all in the content.

aldehir:
> @aldehir `reasoning_format: none` is not `enable_thinking: false`, it just means you don't process the reasoning tags and you spill them all in the content.

I'm fully aware of the difference. See point 2. The responsibility is pushed to the client to strip the tags. Currently there is a hack to remove the exception, but then the model will in-context learn and start to emit bad harmony output, thus breaking parsing.

I've been down this road when I implemented the original parsing.

Edit: I can see my phrasing conflated the two. Ignore the second part.

pwilkin:
@aldehir I know this is generally non-feasible, but there exists a small but vocal group of people who use their own parsing tools for whatever reasons and they really like to get the raw unprocessed contents :)

But I just realized you can not change it and instead approve my https://github.com/ggml-org/llama.cpp/pull/20289 to satisfy them :)

aldehir:
> But I just realized you can not change it and instead approve my https://github.com/ggml-org/llama.cpp/pull/20289 to satisfy them :)

Yes, I rather just give them the whole harmony output. I've seen complaints about "missing tokens" too, it's never ending!

aldehir:
@pwilkin added structured output test.

pwilkin (approved):

common/chat.cpp
```diff
@@ -933,17 +933,12 @@ static common_chat_params common_chat_params_init_gpt_oss(const common_chat_temp
...
-        if (has_reasoning_content && has_tool_calls) {
-            auto adjusted_message        = msg;
-            adjusted_message["thinking"] = msg.at("reasoning_content");
-            adjusted_messages.push_back(adjusted_message);
-        } else {
-            adjusted_messages.push_back(msg);
+    for (auto msg : inputs.messages) {
+        if (msg.contains("reasoning_content") && msg.at("reasoning_content").is_string()) {
+            msg["thinking"] = msg.at("reasoning_content");
+            msg.erase("content");
```
aldehir, line 939: I messed up, need to only do this for tool calls.

Mo-Hashem:
@aldehir I like your dedication in making gpt-oss work as intended for all since it launched .
I am also writing to let you know that @pwilkin is totally right about the reasoning-format = none
it must stay, I personally have my template that use `reasoning-format = none`
that enables agents that doesn't support interleaved CoT (which is 90% of all Agents)

and llama.cpp philosophy is power and high customization, now what is the deference between it and Ollama/LM-studio

aldehir:
@Mo-Hashem, unfortunately it will cause double tags to render and contradicts the template. The official template normally throws an exception. Nonetheless, I can add it back in.

Mo-Hashem:
Yes add it back please, it is a serious blocker
I fixed the double tags with simple template customization (will release it to community after an edge case fix)
templates can be customized, but this no way
thanks @aldehir

PR #20394, opened 2026-03-11
kleidiai: add CPU feature detection to CI run script

martin-klacer-arm:
This patch adds CPU feature detection for KleidiAI build. Previously, the `-march` flags for KleidiAI build were decided only by the present compiler capabilities without considering the current CPU, which is addressed in this patch.


requirements/requirements-tool_bench.txt
```diff
@@ -1,7 +1,8 @@
 aiohttp~=3.9.3
 pytest~=8.3.3
 huggingface_hub>=0.34.0,<1.0
-matplotlib~=3.10.0
+matplotlib~=3.10.0; python_version>="3.10"
+matplotlib>=3.9.0; python_version<"3.10"
```
CISC: Revert, not related.

martin-klacer-arm: Thanks for the comment, this makes sense. Reverted in latest patch

martin-klacer-arm:
Hi, just wondering if there's any updates on this? I'd appreciate another look if you have the time, thank you

ggerganov:
Could you try to utilize and extend (if needed) the existing `GGML_CPU_ALL_VARIANTS` functionality to support this?

https://github.com/ggml-org/llama.cpp/blob/29b28a98242692e352df38254c9a2bb9fa79c884/ggml/src/CMakeLists.txt#L351-L447

The idea is to avoid adding feature detection logic in the `ci/run.sh` script and instead consolidate it in the existing CMake implementation for building the possible CPU backends.

martin-klacer-arm:
Thank you for the comment! That's a very good point, after further investigation I verified that the CMake implementation already does this in a satisfactory way and the existing feature detection/-march setting code from before my patch is actually redundant.

In the latest patch, I removed all the previously existing CPU -march detection, which addresses the same issue as my original patch did and is, I assume, the preferred option.

ggerganov:
@CISC Do you know if I can start the [CI](https://github.com/ggml-org/llama.cpp/actions/workflows/build.yml) workflow on this branch? Or do I need to clone it in a repo branch to do that?

CISC:
> @CISC Do you know if I can start the [CI](https://github.com/ggml-org/llama.cpp/actions/workflows/build.yml) workflow on this branch? Or do I need to clone it in a repo branch to do that?

I have not found a way to manually trigger workflows on PR-branches, would like to know myself. :)

PR #16743, opened 2025-10-23
get_rows & dequantize function implementation for repacked weights of type q6_K (q6_Kx8) 

swetha097:
_NOTE: Creating the PR changes required for whisper.cpp here as llama.cpp already includes test-backend-op_

- This implements the GGML_OP_GET_ROWS operation specifically for repacked (block interleaved) 6-bit quantized format (q6_Kx8).
- The following gains were observed by the changes made in the PR - The changes allow for increased usage of the GEMM function (ggml_gemm_q6_K_8x8_q8_0) for q6_K type.
- The PR was tested in AMD Raphael 7600X for whisper - which supports the following flags :
system_info: n_threads = 4 / 12 | WHISPER : COREML = 0 | OPENVINO = 0 | CPU : SSE3 = 1 | SSSE3 = 1 | AVX = 1 | AVX2 = 1 | F16C = 1 | FMA = 1 | BMI2 = 1 | AVX512 = 1 | AVX512_VBMI = 1 | AVX512_VNNI = 1 | AVX512_BF16 = 1 | LLAMAFILE = 1 | OPENMP = 1 | AARCH64_REPACK = 1 |

image

 master branch commit - https://github.com/swetha097/whisper.cpp/commit/fc45bb86251f774ef817e89878bb4c2636c8a58f

q6K repacking commit ( block interleaving approach for Q6_K quantization for x64/x86 SIMD Architecture )- https://github.com/swetha097/whisper.cpp/commit/d89aaf28e11f86e0af61fa166e7b8d28a57784de

development (get_rows) branch commit - https://github.com/swetha097/whisper.cpp/commit/de9839e9e453b8734ce25e219dc9b61d832bce27

Model for performance tests Downloaded from : https://huggingface.co/ggerganov/whisper.cpp/blob/main/ggml-base.en.bin and quantized to q6_K

This patch of the code was also tested with llama.cpp repository & the perplexity of Q6_K models were ensured to be the same before and after the changes made :

Final estimate: PPL = 5.3669 +/- 0.13305

Model used for perplexity test quantized from - https://huggingface.co/meta-llama/Llama-2-7b

This PR is to merged after - [Q6_K - Block Interleaving Implementation for x86 SIMD (AVX512/AVX2) #15275
](https://github.com/ggml-org/llama.cpp/pull/15275)

PR #20396, opened 2026-03-11
ggml-cpu: Add get_rows support for Q6_K REPACK

Alcpz:
Tests with CPU_REPACK + models with tied embeddings result duplicated tensors, as CPU_REPACK currently doesn't support GET_ROWS.

This has high impact when trying to run small models on low memory devices like phones, E.g:

| Model              | Mem footprint |  Duplicated tensor size   |
| ------------------ | ------------- | ------- |
| LFM2-350M-Q4\_K\_M | 430 MiB       | 52 MiB  |
| LFM2-1.2B-Q4\_K\_M | 1380 MiB      | 105 MiB |

The PR addresses this by adding generic support (only for Q6_K, as that's the standard for the output layer).

Creating as draft as https://github.com/ggml-org/llama.cpp/pull/16743 provides the same work (and this PR is based on that PR essentially). Will try to see if the other PR can be merged in, if not, I'll add proper validation (perplexity, llama-bench).

### DIFF
--- a/ggml/src/ggml-cpu/repack.cpp
+++ b/ggml/src/ggml-cpu/repack.cpp
@@ -3132,6 +3132,31 @@ static block_q6_Kx8 make_block_q6_Kx8(block_q6_K * in, unsigned int blck_size_in
     return out;
 }
 
+// Inverse of make_block_q6_Kx8 back into block_q6_K format.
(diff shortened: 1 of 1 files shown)


Alcpz:
Closing. Original PR is still alive but blocked by lack of testing support for repack

PR #20402, opened 2026-03-11
llama : add fd-based model loading via llama_model_load_from_fd ( REWORK )

Siddhesh2377:
Adds `llama_model_load_from_fd()` to load GGUF models from a POSIX file descriptor instead of a file path.

On Android, apps accessing user files through SAF only get a file descriptor, not a path. The alternative is copying the model into app storage or requesting **MANAGE_EXTERNAL_STORAGE**, which gets rejected by Google Play. This happened with my app ([ToolNeuron](https://github.com/Siddhesh2377/ToolNeuron)).

Reworked version of a previous PR that was rejected for code quality.

Not supported on Windows. The fd is dup'd internally so the caller retains ownership.

Tested locally with CI and a real model (vocab_only + mmap).

### DIFF
--- a/ggml/include/gguf.h
+++ b/ggml/include/gguf.h
@@ -78,7 +78,7 @@ extern "C" {
 
     GGML_API struct gguf_context * gguf_init_empty(void);
     GGML_API struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_params params);
-    GGML_API struct gguf_context * gguf_init_from_fd(int fd, struct gguf_init_params params);
+    GGML_API struct gguf_context * gguf_init_from_file_ptr(FILE * file, struct gguf_init_params params);
     //GGML_API struct gguf_context * gguf_init_from_buffer(..);
 
     GGML_API void gguf_free(struct gguf_context * ctx);
--- a/ggml/include/gguf.h
+++ b/ggml/include/gguf.h
@@ -78,6 +78,7 @@ extern "C" {
 
     GGML_API struct gguf_context * gguf_init_empty(void);
     GGML_API struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_params params);
+    GGML_API struct gguf_context * gguf_init_from_fd(int fd, struct gguf_init_params params);
     //GGML_API struct gguf_context * gguf_init_from_buffer(..);
 
     GGML_API void gguf_free(struct gguf_context * ctx);
--- a/ggml/src/gguf.cpp
+++ b/ggml/src/gguf.cpp
@@ -15,10 +15,6 @@
 #include <string>
 #include <vector>
 
-#ifndef _WIN32
-#include <unistd.h>
-#endif
-
 #define GGUF_MAX_STRING_LENGTH  (1024*1024*1024)
 #define GGUF_MAX_ARRAY_ELEMENTS (1024*1024*1024)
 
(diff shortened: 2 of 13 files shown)


ggml/src/gguf.cpp
```diff
@@ -15,6 +15,10 @@
 #include <string>
 #include <vector>
 
+#ifndef _WIN32
+#include <unistd.h>
+#endif
```
JohannesGaessler:
```suggestion
#endif // _WIN32
```

ggml/include/gguf.h
```diff
@@ -78,6 +78,7 @@ extern "C" {
 
     GGML_API struct gguf_context * gguf_init_empty(void);
     GGML_API struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_params params);
+    GGML_API struct gguf_context * gguf_init_from_fd(int fd, struct gguf_init_params params);
```
JohannesGaessler: For your purposes, would it work to expose the current `gguf_init_from_file_impl` as `gguf_init_from_file_ptr` and to use that as the basis for the implementation instead? That way we would be able to also use this code on Windows in conjunction with `ggml_fopen`.

ggml/src/gguf.cpp
Siddhesh2377: Fixed

ggml/include/gguf.h
Siddhesh2377: Done, replaced `gguf_init_from_fd` with `gguf_init_from_file_ptr(FILE *)` and moved the dup+fdopen logic up to llama-model-loader

JohannesGaessler: The llama C API should also use a file pointer if at all possible, the conversion from file descriptor to file pointer should be in your user code.

Siddhesh2377: Done, switched the llama C API to FILE pointer as well. The test shows the fd to FILE* conversion on the caller side.

ggml/src/gguf.cpp
```diff
@@ -853,6 +853,13 @@ struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_p
     return result;
 }
 
+struct gguf_context * gguf_init_from_file_ptr(FILE * file, struct gguf_init_params params) {
+    if (!file) {
+        return nullptr;
+    }
+    return gguf_init_from_file_impl(file, params);
+}
+
```
JohannesGaessler: Just rename `gguf_init_from_file_impl` to `gguf_init_from_file_ptr` and add the check there. Keep the check in `gguf_init_from_file` since it is associated with a warning.

include/llama.h
```diff
@@ -464,6 +464,9 @@ extern "C" {
                              const char * path_model,
               struct llama_model_params   params);
 
+    // Load a model from an open FILE pointer
+    LLAMA_API struct llama_model * llama_model_load_from_file_ptr(FILE * file, struct llama_model_params params);
```
JohannesGaessler: Please keep the formatting consistent with the surrounding code.

JohannesGaessler (commented):
Maybe I should have made my intent more clear: The GGUF and llama APIs should be using a file pointer rather than a file descriptor because that has Windows compatibility. But then that is also what should be used internally because conversions between the two add unnecessary complexity. Your PR should not be using file descriptors anywhere, please consistently use file pointers.

include/llama.h
Siddhesh2377: Done, and sorry for the misunderstanding :)

src/llama-mmap.h
```diff
@@ -15,12 +15,13 @@ using llama_mlocks = std::vector<std::unique_ptr<llama_mlock>>;
...
 struct llama_file {
     llama_file(const char * fname, const char * mode, bool use_direct_io = false);
+    llama_file(FILE * file);
     ~llama_file();
 
     size_t tell() const;
     size_t size() const;
 
-    int file_id() const; // fileno overload
+    int file_id() const;
```
JohannesGaessler, line 24: What did "fileno" refer to here and why did you remove this comment?

src/llama.cpp
```diff
@@ -889,8 +889,12 @@ static struct llama_model * llama_model_load_from_file_impl(
         void * set_tensor_data_ud,
         const std::string & path_model,
         std::vector<std::string> & splits,
+        FILE * file,
         struct llama_model_params params) {
-    GGML_ASSERT((metadata == nullptr) != path_model.empty() && "exactly one out of metadata and path_model needs to be defined");
+    if (metadata == nullptr && path_model.empty() && !file) {
+        LLAMA_LOG_ERROR("%s: no model source provided\n", __func__);
+        return nullptr;
+    }
```
JohannesGaessler, line 897: The logic should remain that exactly one out of the three things needs to be defined. Something like this should work:

```
GGML_ASSERT(int(metadata != nullptr) + int(path_model.empty()) + int(file != nullptr) == 1 && "exactly one out of metadata and path_model needs to be defined");
```

tests/test-model-load-fd.cpp
JohannesGaessler, line 1: I am currently adding file saving/loading to the recently added end-to-end tests in `test-llama-archs.cpp` via https://github.com/ggml-org/llama.cpp/pull/20503 . This will provide test coverage for your newly added code so I don't think we need this additional test.

PR #20416, opened 2026-03-11
common : fix `--n-cpu-moe`, `--cpu-moe` for models with fused gate + up

ddh0:
Changed the regex that matches conditional experts from:

```cpp
const char * const LLM_FFN_EXPS_REGEX = "\\.ffn_(up|down|gate)_(ch|)exps";
```

to:

```cpp
const char * const LLM_FFN_EXPS_REGEX = "\\.ffn_(up|down|gate|gate_up)_(ch|)exps";
```

I think this should fix #20414.

### DIFF
--- a/common/common.h
+++ b/common/common.h
@@ -926,7 +926,7 @@ const char * const LLM_KV_SPLIT_TENSORS_COUNT = "split.tensors.count";
 // MoE utils
 //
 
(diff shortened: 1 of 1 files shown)


CISC (approved):

Nekotekina:
Fixes #20431

Issue #20325, opened 2026-03-10
Eval bug: Autoparser misplaces non-thinking content with NVIDIA-Nemotron-Nano-9B-v2


EZForever:
### Name and Version

Compiled from current master branch (https://github.com/ggml-org/llama.cpp/commit/c96f608d9861f7e8466bc1b6ac2ff4e3c6f96641), since the model won't load without #20270.

```
C:\llama.cpp-master\build\bin\Debug>llama-cli.exe --version
register_backend: registered backend CPU (1 devices)
register_device: registered device CPU (Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz)
load_backend: failed to find ggml_backend_init in C:\llama.cpp-master\build\bin\Debug\ggml-cpu.dll
version: 0 (unknown)
built with MSVC 19.44.35223.0 for x64
```

### Operating systems

Windows, Linux

### GGML backends

CPU, CUDA, Vulkan

### Hardware

I don't think this issue is related to hardware; I've tested on three different machines with different backends, and they all have the same issue.

### Models

NVIDIA-Nemotron-Nano-9B-v2 IQ2_M from bartowski: (<https://huggingface.co/bartowski/nvidia_NVIDIA-Nemotron-Nano-9B-v2-GGUF/blob/main/nvidia_NVIDIA-Nemotron-Nano-9B-v2-IQ2_M.gguf>). Tried both the model built-in chat template and the one in llama.cpp repo; both have the exact same issue.

I originally discovered this issue when trying to implement `/no_think` with a custom chat template for [Qwen3.5-35B-A3B](https://huggingface.co/unsloth/Qwen3.5-35B-A3B-GGUF/blob/main/Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf). The modified chat template is [here](https://github.com/user-attachments/files/25858532/chat_template.mod.jinja.txt); it works correctly before the autoparser PR.

### Problem description & steps to reproduce

NVIDIA-Nemotron-Nano-9B-v2 supports both thinking and non-thinking mode in a single model, and supports switching between them in-conversation with a chat template trick, as documented in [model card](https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-9B-v2#prompt-format). It works correctly before, however since the new autoparser PR, after switching into non-thinking mode with `/no_think`, while the model correctly skips thinking, the model's output is no longer being treated as normal content as it should, but as thinking content.

[...]

Issue #20265, opened 2026-03-09
Eval bug: Model output directed to Reasoning block instead of standard response.


441041:
### Name and Version

command is :  llama-server -m  Qwen3.5-9B-Claude-4.6-HighIQ-INSTRUCT-HERETIC-UNCENSORED.i1-Q4_K_M.gguf  --host 0.0.0.0   
version: llama-b8244-bin-win-cuda-12.4-x64
<img width="1120" height="434" alt="Image" src="https://github.com/user-attachments/assets/305c7bf7-c0a7-47b6-a5ff-e7debfa4d70d" />

### Operating systems

Windows

### GGML backends

CUDA

### Hardware

RTX 5070

### Models

Qwen3.5-9B-Claude-4.6-HighIQ-INSTRUCT-HERETIC-UNCENSORED.i1-Q4_K_M.gguf  

### Problem description & steps to reproduce

Model output directed to Reasoning block instead of standard response.

<img width="1120" height="434" alt="Image" src="https://github.com/user-attachments/assets/6d6e4d2b-055e-415b-9278-d6cb819be9c3" />

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

pwilkin:
Can you reproduce this on a normal version of the model? Hard to know whether this stems from template problems or from the finetune itself.

pnb:
@pwilkin FWIW I just started getting this on an older non-reasoning, non-fine-tuned Qwen 3 30B-A3B model as well. [90% sure it is this specific file](https://huggingface.co/unsloth/Qwen3-30B-A3B-Instruct-2507-GGUF/blob/main/Qwen3-30B-A3B-Instruct-2507-Q4_K_M.gguf) although I can't get to the computer it is running on today to double check the GGUF source; 10% chance it is the Bartowski version instead.

Although I see there is nonsense about <think> and reasoning content in the chat template, so maybe it is technically a broken template -- but it has worked for the past 8 months anyway.

ali0une:
Hi there.

indeed something changed recently..

**Example with  Qwen3-30B-A3B-Instruct-2507-Q4_K_M.gguf :**

llama-server with --reasoning-format none argument has reasoning content going to "content":

```json
[...]

PR #20424, opened 2026-03-11
common/parser: add proper reasoning tag prefill reading

pwilkin:
This changes the erroneous behavior of the autoparser that ascribed thinking behavior to *templates*. As people rightly mentioned, some models have dynamic or hybrid reasoning - they can reason or not depending on some switches and even the template behavior can change due to this (i.e. inserting `<think>` in assistant prefill after a "no_think" appears in a user message).

Therefore, the `FORCED_OPEN` and `FORCED_CLOSED` formats are gone. The parser will now just detect models with tagged reasoning, i.e. an opening and closing reasoning marker (deleted `DELIMITER` also since it's a special case with the opening marker being empty). However, it will check the assistant prefill for those markers and will append them to the input for the grammar and the parser, so that they are taken into account, therefore just simplifying the parsing mechanism since it doesn't now have to differentiate whether the `<think>' / `</think>` was added by the template or generated by the model.


pwilkin:
Fixes #20356 
Fixes #20325 
Fixes #20265 

This also clears the ground for disabling grammar triggers inside reasoning loops in a subsequent PR, which would resolve #20260

aldehir:
Dumb question, why not find the start of the assistant message and prepend that?

I agree it would be easier to parse if we had a "prefill" of some sort that normalizes the input, such that we can handle the logic in the grammar and not through flags. However, if we're going this route I would look into prepending the start of the entire assistant message. This will also open the door for parsing output from requests with an assistant prefill.

pwilkin:
Yeah, that would be the logical conclusion, but for now it's easier for me just to extract the reasoning markers since finding the actual start of the assistant message is nontrivial.

aldehir:
Qwen3.5 uses
```
<think>\n\n</think>\n\n
```
```
{%- if enable_thinking is defined and enable_thinking is false %}
{{- '<think>\n\n</think>\n\n' }}
{%- else %}
{{- '<think>\n' }}
{%- endif %}
```
however,
```
      "reasoning_prefill": "<think></think>\n\n",
```
It probably doesn't matter for this model, but it is technically not adhering to the template.

aldehir:
```
    {
      "id": 248045,
      "piece": "<|im_start|>"
    },
    {
      "id": 74455,
      "piece": "assistant"
    },
    {
      "id": 198,
      "piece": "\n"
    },
    {
      "id": 248068,
      "piece": "<think>"
    },
    {
      "id": 271,
      "piece": "\n\n"
    },
    {
      "id": 248069,
      "piece": "</think>"
    },
    {
      "id": 271,
      "piece": "\n\n"
    }
```
Maybe set `reasoning_prefill` from the start of the opening tag to the end of the prompt?

aldehir:
> finding the actual start of the assistant message is nontrivial.

Run the template once with `add_generation_prompt = false`, capture the size, run again with `true`, extract the string content that spans the delta? I think that would work in most cases.

pwilkin:
That *usually* works, yeah 😀 I can try that and see what the results are (this is what `calculate_diff_split` from the analyzer does BTW). I'm just worried about some weird edge cases.

bsdice:
Nice patch! With model https://huggingface.co/mradermacher/Qwen3.5-40B-Claude-4.5-Opus-High-Reasoning-Thinking-GGUF the patches fix webui getting confused on /think and not splitting correctly reasoning and generation part. Build llama.cpp-cuda-git-b8334.r9.710878a7dd-1.

pwilkin:
@aldehir changed the prefill extraction behavior to the differential one you mentioned.

common/chat.h
```diff
@@ -211,7 +211,7 @@ struct common_chat_params {
     std::string                         prompt;
     std::string                         grammar;
     bool                                grammar_lazy         = false;
-    bool                                thinking_forced_open = false;
+    std::string                         prefill;
```
aldehir: Think we name this `generation_prompt`? It lines up with the `add_generation_prompt` flag.

common/chat-auto-parser-generator.cpp
```diff
@@ -39,15 +40,69 @@ common_chat_params peg_generator::generate_parser(const common_chat_template &
...
+                if (string_ends_with(fs_trimmed, r_end_t)) {
+                    data.prefill = r_start + r_end;
+                } else if (string_ends_with(fs_trimmed, r_start_t)) {
+                    data.prefill = from_start;
+                } else {
+                    clear_reasoning_start = true;
+                }
+            }
+        }
+    }
```
aldehir: So my understanding is: we have a generation prompt `G`, we can create a parser that accepts `G[0:max(G.size(), G.index_of(reasoning_start))] + (reasoning_start + reasoning + reasoning end)? + ...`. Then we can do away with all the trim logic.

The benefit is that now the parser can properly parse assistant prefill from the user, since the parser starts from the beginning of the assistant message.

I see that Mistral's templates have no generation prompt, so `G = ""`. But this is fine, because the model emits the `[THINK]` tag. So the above still works.

pwilkin: This workaround is mostly for Apriel that has a delimited thinking format and inserts a header "Thinking chain starts here: " or something like that as the generation prompt which acts as a quasi-reasoning marker that we want to strip.

pwilkin:
@aldehir okay, that rewrite ended up being *a bit* bigger than I expected... but it's exactly the algorithm you mentioned now.

aldehir:
Oh jeez, well it's <100 net LOC. I'll give it a whirl.

pwilkin:
@aldehir happy to report I added another nice piece of code to make it work correctly with grammars / schemas :)

common/sampling.cpp
```diff
@@ -251,15 +253,60 @@ struct common_sampler * common_sampler_init(const struct llama_model * model, st
...
+                // Check if remaining tokens after start match the end sequence
+                // (possibly with whitespace tokens in between, but for simplicity check suffix)
+                auto end_start = prefill_tokens.end() - (ptrdiff_t) params.reasoning_budget_end.size();
+                if (end_start >= end_begin &&
+                    std::equal(params.reasoning_budget_end.begin(), params.reasoning_budget_end.end(), end_start)) {
+                    activate_immediately = false;
+                }
+            }
+        }
+
```
ggerganov: Maybe move this logic inside `common_reasoning_budget_init`?

ggerganov (commented):
I'm not sure how this new logic interacts with the existing logic for feeding the prompt to the sampler:

https://github.com/ggml-org/llama.cpp/blob/07c6a59b4fc9f6163aa3a4cdc6c4b83646478337/tools/server/server-context.cpp#L198-L220

In general the new `common_params_sampling.grammar_external` external flag feels off. I would look to avoid it.

pwilkin:
> In general the new `common_params_sampling.grammar_external` external flag feels off. I would look to avoid it.

I'd really love to avoid it, but there's a key problem here.

Together with @aldehir we figured out an algorithm to clearly handle cases of template generation prompt prefill - we simply add the prefill to the parser, this is reflected in the grammar and therefore we handle this irrelevant of what is added by the template and what is generated by the model.

If an external `json_schema` is added to constrain output, it's also not a problem since the json schema is not *directly* converted into the grammar - instead, it's fed into the parser and only reflected in the grammar via the parser, so we handle this.

However, there's one case we *can't* handle, which is the user explicitly specifying a grammar. Because the grammar passed by the user is non-lazy and because it's not aware of the generation prompt prefill, it will fail exactly on the prefill part, which is why it's critical to notify the mechanism that it's an externally passed grammar (and not one generated within the parser engine) so that the prefill part is not added.

common/sampling.cpp
pwilkin: Aight.

pwilkin: Done.

aldehir:
Hmm, this on me. I didn't think through the grammar impact. The grammar sampler should be limited to just the generation otherwise it's not technically "sampling." What if we modified how the root rule is defined? This should invoke no changes to the grammar sampler and its inputs. Alternatively we can leverage the `grammar_root` parameter that defaults to `root`, which might be an easier solution.

tools/cli/cli.cpp
```diff
@@ -215,7 +215,7 @@ struct cli_context {
         inputs.parallel_tool_calls   = false;
         inputs.add_generation_prompt = true;
         inputs.reasoning_format      = COMMON_REASONING_FORMAT_DEEPSEEK;
-        inputs.enable_thinking       = common_chat_templates_support_enable_thinking(chat_params.tmpls.get());
+        inputs.enable_thinking       = chat_params.enable_thinking ? common_chat_templates_support_enable_thinking(chat_params.tmpls.get()) : false;
```
aldehir, line 219: Can we extract this fix to another PR? Just so we can merge it in sooner. I think this PR needs a bit more brainstorming.

models/templates/deepseek-ai-DeepSeek-R1-Distill-Qwen-32B.jinja
```diff
@@ -15,10 +15,10 @@
     {%- set ns.is_tool = false -%}
     {%- for tool in message['tool_calls']-%}
       {%- if not ns.is_first -%}
-        {{'<｜Assistant｜><｜tool▁calls▁begin｜><｜tool▁call▁begin｜>' + tool['type'] + '<｜tool▁sep｜>' + tool['function']['name'] + '\n' + '```json' + '\n' + tool['function']['arguments'] + '\n' + '```' + '<｜tool▁call▁end｜>'}}
+        {{'<｜Assistant｜><｜tool▁calls▁begin｜><｜tool▁call▁begin｜>' + tool['type'] + '<｜tool▁sep｜>' + tool['function']['name'] + '\n' + '```json' + '\n' + tool['function']['arguments'] | tojson + '\n' + '```' + '<｜tool▁call▁end｜>'}}
```
aldehir, line 18: I'm thinking we use the JSON representation when dumping a JSON object in the Jinja engine itself.

tools/cli/cli.cpp
pwilkin: Ye, sure.

models/templates/deepseek-ai-DeepSeek-R1-Distill-Qwen-32B.jinja
pwilkin: Well, no, we do not, as I found out when I removed the "Python dict" support.

We use the Python dict representation.

Or, more precisely, use double quotes if the internal string contains a single quote and single-quotes when it does not:

```c++
// TODO: avoid circular references
std::string value_to_string_repr(const value & val) {
    if (is_val<value_string>(val)) {
        const std::string val_str = val->as_string().str();

        if (val_str.find('\'') != std::string::npos) {
            return value_to_json(val);
        } else {
            return "'" + val_str + "'";
        }
    } else {
        return val->as_repr();
    }
}
```

I would really love it to use the JSON representation as standard, but I'm worried there might be some "original Python implementation faithfulness" shenanigans. @ngxson @CISC do you guys know if we could make `| tojson` the official string representation for objects?

pwilkin:
> Hmm, this on me. I didn't think through the grammar impact. The grammar sampler should be limited to just the generation otherwise it's not technically "sampling." What if we modified how the root rule is defined? This should invoke no changes to the grammar sampler and its inputs. Alternatively we can leverage the `grammar_root` parameter that defaults to `root`, which might be an easier solution.

Yes, I thought of this path as well. Thing is, I'm not really sure this would be *easier*. We *still* have to differentiate the user-supplied grammars from the internally-generated grammars to know which ones we need to patch the root rule for.

Regarding the sampling, the thing is we *are* technically simulating the sampling of the generation prompt, that's the whole idea of this approach - to make it basically invisible to the parser / grammar whether the given fragment was generated by the template or by the model itself.

tools/cli/cli.cpp
pwilkin: Extracted to https://github.com/ggml-org/llama.cpp/pull/20606

pwilkin:
BTW @ggerganov is the weird thingy where you have to add a space in front of certain tokens still necessary? It looks really bizarre and it seems to break idempotence (as in `detokenize(tokenize(x)) != x`).

models/templates/deepseek-ai-DeepSeek-R1-Distill-Qwen-32B.jinja
aldehir: An argument can be made that Jinja will respect `__str__()` if wrapped in a JSON type that serializes to JSON.

```python
import json
from jinja2 import Template

tmpl = Template("{{ obj }}")

python_dict = dict(a=1, b=2, c=dict(d=3, e=4))
print("Python Dict:", tmpl.render(obj=python_dict))

class JSON:
    def __init__(self, value):
        self.value = value
    def __str__(self):
        return json.dumps(self.value)

json_obj = JSON(dict(a=1, b=2, c=dict(d=3, e=4)))
print("JSON Object:", tmpl.render(obj=json_obj))
```
```
$ ./example.py
Python Dict: {'a': 1, 'b': 2, 'c': {'d': 3, 'e': 4}}
JSON Object: {"a": 1, "b": 2, "c": {"d": 3, "e": 4}}
```

pwilkin: I'd be really glad to change it because it would mean we don't have to change the templates and we handle all the templates that don't use `| tojson` on arguments - but I won't make the call myself :)

CISC: That's not what's happening here though, this template does not support non-string arguments, that's all there is to it.

aldehir: Fair, I did not test with string concat which throws an exception.

There is no winning here. Either we don't parse `arguments` and break templates that iterate its values. Or we do, a break templates that expect a string.

CISC: > I'd be really glad to change it because it would mean we don't have to change the templates and we handle all the templates that don't use `| tojson` on arguments - but I won't make the call myself :)

It's not acceptable, it's not how this behaves with `jinja2`, either `arguments` is sent as a `str`, or it's been parsed with `json.loads`, in which case it's a `dict`, in other words this particular template will give you the following error:
```console
TypeError: can only concatenate str (not "dict") to str
```

CISC: Anyway, I don't understand the issue, don't we have a capability check for this?

aldehir: > capability check for this

No, the arguments are indiscriminately parsed. I don't see a capability check.

CISC: > > capability check for this
> 
> No, the arguments are indiscriminately parsed. I don't see a capability check.

Hmmm, maybe that was `minja`, either way, we should probably add one.

aldehir:
I'm starting to think there isn't a clean solution here. To address the original issue, perhaps we should simply adopt the previous checks for the end thinking tag and create a custom parser for templates that deviate.

I believe even a "reasoning prefill" will have challenges when incorporating with the grammar and reasoning samplers.

pwilkin:
There isn't a clean solution, but I quite like this one because it actually realizes the idea of treating template prefill as forced model output. If you think about it, what we're doing is not really that different from the reasoning sampler forcing the end-thinking tag - but instead of forcing the model to generate it, we just simulate the generation for the prefill.

This makes the parsing for reasoning elegant and simple and avoids all problems with deviant / weird / atypical templates. I think having one extra flag, even if a big hackish, is an appropriate price to pay for it.

pwilkin:
Also, I tested this with the reasoning sampler and it works fine.

ggerganov:
> BTW @ggerganov is the weird thingy where you have to add a space in front of certain tokens still necessary? It looks really bizarre and it seems to break idempotence (as in `detokenize(tokenize(x)) != x`).

Not sure where the spaces come from. There was something about having to apply `normalize()` to restore the identity of `tokenize + detokenize`, but don't remember the details.

common/sampling.cpp
```diff
@@ -259,7 +282,7 @@ struct common_sampler * common_sampler_init(const struct llama_model * model, st
             params.reasoning_budget_end,
             params.reasoning_budget_forced,
             params.reasoning_budget_tokens,
-            params.reasoning_budget_activate_immediately ? REASONING_BUDGET_COUNTING : REASONING_BUDGET_IDLE));
+            prefill_tokens));
     }
```
ggerganov: The main issue that I see is that this logic is technically the same as the logic here:

https://github.com/ggml-org/llama.cpp/blob/d3936498a3d8f41cdb35d9e7d04a19e704b4fc89/tools/server/server-context.cpp#L198-L220

Unless I am missing something, the `common_sampler` should work like this:

- We can initialize an object:
  - without grammar, without reasoning budget
  - without grammar, with reasoning budget
  - with grammar, without reasoning budget
  - with grammar, with reasoning budget
- In all cases, we need to feed any "initial" (a.k.a. "prefill") tokens to both the grammar and the reasoning budget samplers
- So this feeding logic should become part of the `common_sampler` initialization
- I don't see why we need the `params.grammar_external` flag - we always have to feed the tokens to the grammar and reasoning samplers, if they exist

Maybe try to remove the old prefill logic from the server and consolidate it in `common_sampler_init()`.

pwilkin: We need the `grammar_external` flag because otherwise we'll break backwards compatibility.

Some templates add stuff like the assistant role marker (`<|im_begin|>assistant`) in the prefill. User grammars generally only expect to parse what's present in the content section. The idea is to allow the new behavior of simulating the sampling of the prefill tokens but also guaranteeing backwards compatibility (this is why the server tests failed before I added the fix: 

https://github.com/ggml-org/llama.cpp/actions/runs/23100821002

As for the common_sampler initialization, I'll take a look into it.

ggerganov: > User grammars generally only expect to parse what's present in the content section.

Ah yes, I forgot about that.

pwilkin: @ggerganov okay, so I checked the flow and we can't really move this to `common_sampler_init()`.

The way `common_sampler_init()` works is it resets and reinitializes the sampler state *for all samplers other than the grammar sampler*. The grammar sampler is a different beast from the other samplers since its state depenendence is different - it doesn't depend on the entire prompt, but on the *current message* (same with the reasoning sampler). Therefore, if we wanted to do this uniformly, we'd have to do something like this:

* modify the sampler struct to add a scope parameter (PROMPT, MESSAGE)
* keep track of where the beginning of the current message starts (non-trivial since template does this and templates routinely modify the message history eg. by removing reasoning content so we can't just rely on differential analysis here)
* run accept for PROMPT-scope samplers for the entire prompt and for MESSAGE-scope samplers for the current message only

This would make the behavior cleaner, but is a big change probably outside the scope of this PR.

ggerganov: I think it's much simpler than that.

Each sampler has an `accept()` call that consumes a token. Generally we want to feed all tokens from the current context into the samplers. Hence the loop that I referenced in the `server`. The grammar sampler is the only exception - the `false` argument to `common_sampler_accept()` makes it so that we don't feed the grammar sampler with the "prefill" tokens.

So I am thinking that the newly added logic in `common_sampler_init()` for collecting "prefill" tokens and passing it to the reasoning sampler is not necessary. The reasoning sampler should accept all of its token through `common_sampler_accept()` - same as all other samplers (except the grammar).

Most likely the `common_reasoning_budget_init()` needs to be updated to not accept prefill tokens at all.

pwilkin: @ggerganov We definitely do want to pass prefill tokens to the reasoning sampler, the same way we pass it to the grammar sampler, simply because the two samplers operate on the same principle - their context is the single assistant message that is being generated, not the entire conversation (prompt).

However, I now finally understand why the entire flow is so confusing. So when a completion task gets assigned to a slot, this happens:
-> task gets assigned to a slot
-> chat template gets applied, processing the message history to a tokenizable prompt and adding assistant prefill tokens
-> the prompt (with prefill tokens) gets tokenized
-> the tokenized prompt is passed to the budget and reasoning parsers together with the prefill part; those parsers are initialized with just the tokens of the assistant prefill
-> model processes the prompt
-> now that generation is about to start, all the *other* samplers get reset and fed the entire tokenized prompt together with the assistant prefill
-> all samplers - the reasoning+grammar ones initialized earlier and the generation ones initialized now - start working on the outputs generated by the model

It's really counterintuitive why the sampler reset + initialization is done on generation start, when it could just as well be done during init (right now it doesn't matter that much, but if we have backend samplers, they could theoretically run the prefill / accept phase in parallel with the model doing prompt processing).

However, my point still stands - we can't really equate the two prefills because they differ in what they understand as "context". For the normal samplers, the *entire prompt* is the context - the entire message history together with the assistant prefill, before the generation phase. For the reasoning and grammar samplers, only the assistant prefill (the current assistant message) is the context. We could init them in the `init_sampler()` stage and have a unified prefill stage, but only if we mark where the assistant message starts in the prompt *and* the samplers themselves are aware whether they require full context or just assistant prefill context.

pwilkin: In other words, the loop inside `init_sampler()` would become:

```c++
        for (int i = 0; i < (int) prompt.tokens.size(); i++) {
            const llama_token id = prompt.tokens[i];

            if (id != LLAMA_TOKEN_NULL) {
                common_sampler_accept(smpl.get(), id, i >= asst_prefill_start_pos);
                n_text++;
            }
        }
```

where `asst_prefill_start_pos` would be calculated during task init as the position within the prompt where the assistant prefill tokens start, the third parameter to `common_sampler_accept` would change its meaning from `accept_grammar` to `is_assistant_prefill_token` and the samplers would know whether they require full context or just assistant prefill context. Then we could get rid of the loop in `common_sampler_init` :)

ggerganov: > We could init them in the init_sampler() stage and have a unified prefill stage, but only if we mark where the assistant message starts in the prompt and the samplers themselves are aware whether they require full context or just assistant prefill context.

I think even more generally, during tokenization of the task input, we need to create some prompt structure that we can utilize later. So not just where the last assistant message start, but generally where all messages start. This way we could not only simplify the `common_sampler` logic in the way that you described, but we can also significantly improve the checkpointing logic for recurrent models (as was discussed earlier) and potentially do other stuff with it.

> We need the grammar_external flag because otherwise we'll break backwards compatibility.

If I understand correctly, I think the reasoning budget sampler needs to have it's own internal grammar sampler? This way, we keep the existing grammar sampler logic as is.

allozaur:
@pwilkin could u please tag me in a comment when the server-side is fully done and last thing to review is the WebUI part? :)

common/sampling.cpp
pwilkin: > I think even more generally, during tokenization of the task input, we need to create some prompt structure that we can utilize later. So not just where the last assistant message start, but generally where all messages start. This way we could not only simplify the `common_sampler` logic in the way that you described, but we can also significantly improve the checkpointing logic for recurrent models (as was discussed earlier) and potentially do other stuff with it.

Okay, so how about this: we merge this PR as-is (since it's pretty critical and resolves a lot of issues) and I'll open a separate one to work on the prompt structure logic since that's quite a bit more work - then once that happens we can rewrite this to a clean logic using only the `init_sampler()` method?

> If I understand correctly, I think the reasoning budget sampler needs to have it's own internal grammar sampler? This way, we keep the existing grammar sampler logic as is.

No, the reasoning budget sampler has been rewritten not to use grammar, it only checks for specific token sequences + the countdown mechanism.

ggerganov: I still don't understand why we prefill the grammar sampler only when `!params.grammar_external`. Why don't we always prefill it?

pwilkin: @ggerganov because if we prefill an externally provided grammar then it's going to fail on the generation prompt tokens (i.e. `<|im_begin|>assistant`) - see the failing server tests I linked above.

ggerganov: If the reasoning budget sampler does not use the grammar, I don't see in which cases we need to prefill the grammar then?

I.e. if the user didn't provide a grammar, then `grammar_external == false`. The reasoning budget sampler does not use grammar as you mentioned. So grammar sampler is unused --> no need to prefill.

pwilkin: @ggerganov the chat parser generates a grammar for tool calling purposes. That's the "internal" (i.e. "non-external") grammar we're handling (and prefilling) in this case. If a user provides a JSON schema, the parser also generates a grammar based on that JSON schema instead (that's another case of the "internal grammar"). Both of those need the prefill tokens.

ggerganov: Ok, I understand now.

And one more question - if the user provides a JSON schema and the parser also generates a grammar for tool calling, what would happen - is this compatible?

pwilkin: @ggerganov No, we explicitly disallow it and throw an error message that structured output is not compatible with tool calling.

tools/server/server-task.cpp
```diff
@@ -382,7 +382,8 @@ task_params server_task::params_from_json_cmpl(
             throw std::runtime_error(std::string("\"json_schema\": ") + e.what());
         }
     } else {
-        params.sampling.grammar      = json_value(data, "grammar", defaults.sampling.grammar);
+        params.sampling.grammar          = json_value(data, "grammar", defaults.sampling.grammar);
+        params.sampling.grammar_external = json_value(data, "grammar_external", params.sampling.grammar_external);
```
ggerganov:
```suggestion
        params.sampling.grammar_external = json_value(data, "grammar_external", defaults.sampling.grammar_external);
```

common/sampling.cpp
ggerganov: It will be easier to follow the logic if instead of introducing `bool grammar_external`, introduce `std::string grammar_json` like so:

```c++
// ref: https://github.com/ggml-org/llama.cpp/pull/20424#discussion_r2948173630
std::string grammar;         // optional, user-provided BNF-like grammar to constrain sampling
std::string grammar_json;    // JSON-schema generated grammar
std::string grammar_prefill; // parser generated grammar
```

Probably also rename `grammar_prefill` -> `grammar_parser`.

Somewhere suitable, add:

```c++
// user-provided grammar is not compatible with json schema
GGML_ASSERT(grammar.empty() || grammar_json.empty());
```

Or instead of assert, handle as error.

tools/server/server-context.cpp
```diff
@@ -1151,10 +1152,19 @@ struct server_context_impl {
...
-            slot.smpl.reset(common_sampler_init(model, task.params.sampling));
+            try {
+                slot.smpl.reset(common_sampler_init(model, task.params.sampling));
+            } catch (std::exception & e) {
+                LOG_ERR("%s: error initializing samplers. Grammar was:\n%s\n\nGrammar prefill:\n'%s'\n", __func__,
+                    task.params.sampling.grammar.c_str(), task.params.sampling.grammar_prefill.c_str());
+                std::string err_msg = std::string("Failed to initialize samplers: ") + e.what();
+                send_error(task, err_msg, ERROR_TYPE_INVALID_REQUEST);
+                return false;
+            }
```
ggerganov, line 1162: It's not clear where in `common_sampler_init()` this exception is going to be thrown. See me other comment and handle this error explictly. Here, just catch a generic exception and don't assume it is caused by the grammar.

common/sampling.cpp
pwilkin: Fun fact - `grammar_prefill` is not a grammar, it's the generation prompt to be prefilled into the grammar ;) I'll change the name as it's misleading.

pwilkin:
@ggerganov okay, I think I'm done with the refactorings, stuffed the three grammar types into an `std::variant` and did the internal exception handling on the grammar prefill.

common/common.h
```diff
@@ -178,6 +179,35 @@ enum common_speculative_type {
...
+    }, g);
+}
+
+// Returns true when the generation_prompt should be prefilled into the grammar sampler.
+// Only output-format and tool-call grammars need prefill; user-supplied grammars must not be prefilled.
+inline bool common_grammar_needs_prefill(const common_grammar_variant & g) {
+    return std::holds_alternative<common_grammar_output_format>(g)
+        || std::holds_alternative<common_grammar_tool_calls>(g);
+}
+
```
ggerganov, line 219: This is too C++ heavy. Let's not use `std::variant`.

If you want to make this type safe, just wrap a string and an enum value in a struct.

pwilkin: Aight, modified it to a struct with enum instead.

pwilkin:
Any idea why this stupid PR shows "branch has conflicts" when it's literally rebased on master?

common/common.h
```diff
@@ -3,12 +3,14 @@
...
 
 #include "ggml-opt.h"
+#include "ggml.h"
 #include "llama-cpp.h"
 
 #include <set>
 #include <sstream>
 #include <string>
 #include <string_view>
+#include <variant>
```
ggerganov, line 13: Cleanup the unused headers

common/sampling.cpp
```diff
@@ -189,9 +191,10 @@ struct common_sampler * common_sampler_init(const struct llama_model * model, st
 
     std::vector<llama_sampler *> samplers;
 
-    if (params.grammar.compare(0, 11, "%llguidance") == 0) {
+    const std::string & grammar_str = common_grammar_value(params.grammar);
+    if (grammar_str.compare(0, 11, "%llguidance") == 0) {
 #ifdef LLAMA_USE_LLGUIDANCE
-        grmr = llama_sampler_init_llg(vocab, "lark", params.grammar.c_str());
+        grmr = llama_sampler_init_llg(vocab, "lark", grammar_str.c_str());
```
ggerganov, line 198: For llguidance, require that the grammar is user-provided. Print an error and throw.

common/sampling.cpp
```diff
@@ -240,17 +243,45 @@ struct common_sampler * common_sampler_init(const struct llama_model * model, st
...
+                    LOG_DBG("%s: accepted prefill token (%d)\n", __func__, token);
+                }
+            } catch (std::exception &e) {
+                LOG_ERR("%s: error initializing grammar sampler for grammar:\n%s\n\nGeneration prompt:\n'%s'\n", __func__,
+                    common_grammar_value(params.grammar).c_str(), params.generation_prompt.c_str());
+                throw e;
+            }
+        }
+    }
+
```
ggerganov, line 286: If we have a generation prompt, but we don't have a vocab for some reason, this is a bug. `GGML_ASSERT(vocab)`

tools/server/server-context.cpp
```diff
@@ -1152,10 +1153,17 @@ struct server_context_impl {
...
+            } catch (std::exception & e) {
+                std::string err_msg = std::string("Failed to initialize samplers: ") + e.what();
+                send_error(task, err_msg, ERROR_TYPE_INVALID_REQUEST);
+                return false;
+            }
 
             if (slot.smpl == nullptr) {
-                // for now, the only error that may happen here is invalid grammar
+                LOG_ERR("%s: error in parsing grammar. Grammar was:\n%s\n\nGeneration prompt:\n'%s'\n", __func__,
+                    common_grammar_value(task.params.sampling.grammar).c_str(), task.params.sampling.generation_prompt.c_str());
```
ggerganov: This `if` block can be removed now.

ggerganov (approved):
The changes to `common_sampler` and server are OK. Let's wait for @aldehir to review the rest.

IMO the idea for extracting a prompt structure of where the messages starts, as we discussed earlier, would be great. Hope you can figure out how to do it next.

aldehir:
Might take me a minute, I'll see if I can sneak in some reviewing while on vacation 🙂.

> IMO the idea for extracting a prompt structure of where the messages starts, as we discussed earlier, would be great.

For the purposes of checkpointing, I believe this can be done trivially with an Aho-Corasick automaton + set of known start sequences. Falling back to the current strategy if inconclusive. I'll follow up with a PoC.

common/chat-diff-analyzer.cpp
```diff
@@ -185,7 +187,6 @@ void autoparser::analyze_template(const common_chat_template & tmpl) {
     LOG_DBG("func_name_prefix: '%s'\n", tools.function.name_prefix.c_str());
     LOG_DBG("func_name_suffix: '%s'\n", tools.function.name_suffix.c_str());
     LOG_DBG("func_close: '%s'\n", tools.function.close.c_str());
-    LOG_DBG("python_dict_format: %s\n", tools.format.uses_python_dicts ? "true" : "false");
```
aldehir, line 188: Are we getting rid of this entirely? If so, then I can clean up the additions to the PEG parser. No need to take any action.

common/chat.cpp
```diff
@@ -876,8 +878,8 @@ static common_chat_params common_chat_params_init_ministral_3(const common_chat_
         // Response format parser
         if (inputs.json_schema.is_object() && !inputs.json_schema.empty()) {
             // Ministral wants to emit json surrounded by code fences
-            return reasoning << "```json" << p.content(p.schema(p.json(), "response-format", inputs.json_schema))
-                             << "```";
+            return wrap_for_generation_prompt(p, reasoning << "```json" << p.content(p.schema(p.json(), "response-format", inputs.json_schema)) << "```",
+                inputs, autoparser::analyze_reasoning("[THINK]", "[/THINK]"));
```
aldehir: I don't think we should delegate anything to the autoparser from specialized parsers. Be explicit when adding the generation prompt (e.g. `p.literal()`), we know already know what it's going to be since we matched the template.

common/chat.cpp
```diff
@@ -1018,10 +1022,12 @@ static common_chat_params common_chat_params_init_gpt_oss(const common_chat_temp
                 return tool_call | ( any + p.zero_or_more(start + any) + start + tool_call);
             }
 
-            return tool_call | final_msg | (any + p.zero_or_more(start + any) + start + (tool_call | final_msg));
+            return wrap_for_generation_prompt(p, tool_call | final_msg | (any + p.zero_or_more(start + any) + start + (tool_call | final_msg)),
+                inputs, autoparser::analyze_reasoning("<|channel|>", ""));
```
aldehir: Same thing as ministral above. In fact, adding the generation prompt simplifies the parser for gpt-oss! We can simplify this to `p.zero_or_more(start + any) + start + (tool_call | final_msg)`. Similarly for other instances.

common/chat.cpp
```diff
@@ -1545,48 +1565,51 @@ static common_chat_params common_chat_templates_apply_jinja(const struct common_
...
 
+    common_chat_params early_return;
+
     // Ministral/Mistral Large 3 - uses special reasoning structure fixes, can't use autoparser
     // Note: Mistral Small 3.2 uses [CALL_ID] which Ministral doesn't have, so we can distinguish them
     if (src.find("[SYSTEM_PROMPT]") != std::string::npos && src.find("[TOOL_CALLS]") != std::string::npos &&
         src.find("[ARGS]") != std::string::npos && src.find("[CALL_ID]") == std::string::npos) {
         LOG_DBG("Using specialized template: Ministral/Magistral Large 3\n");
-        return common_chat_params_init_ministral_3(tmpl, params);
+        early_return = common_chat_params_init_ministral_3(tmpl, params);
```
aldehir: Extract the specialized parsers to a separate function to avoid `early_return`? Otherwise, rename to `result` and change to a series of `if/else if`.

common/common.h
```diff
@@ -178,6 +180,43 @@ enum common_speculative_type {
...
+// Grammar type enumeration
+enum common_grammar_type {
+    COMMON_GRAMMAR_TYPE_NONE,           // no grammar set
+    COMMON_GRAMMAR_TYPE_USER,           // user-provided GBNF (--grammar / "grammar" API field)
+    COMMON_GRAMMAR_TYPE_OUTPUT_FORMAT,  // auto-generated from JSON schema (--json-schema / "json_schema" API field)
+    COMMON_GRAMMAR_TYPE_TOOL_CALLS,     // auto-generated by chat template parser for function calling
+};
+
+// Grammar variant struct with type and grammar string
+struct common_grammar_variant {
```
aldehir: Nit: rename to `common_grammar`?

models/templates/llama-cpp-deepseek-r1.jinja
```diff
@@ -49,7 +49,7 @@ Example function tool call syntax:
             {%- endif -%}
             {%- set tool_name = tc['function']['name'] -%}
             {%- set tool_args = tc['function']['arguments'] -%}
-            {{- '<｜tool▁call▁begin｜>' + tc['type'] + '<｜tool▁sep｜>' + tool_name + '\n' + '```json' + '\n' + tool_args + '\n' + '```' + '<｜tool▁call▁end｜>' -}}
+            {{- '<｜tool▁call▁begin｜>' + tc['type'] + '<｜tool▁sep｜>' + tool_name + '\n' + '```json' + '\n' + tool_args | tojson + '\n' + '```' + '<｜tool▁call▁end｜>' -}}
```
aldehir, line 52: I think we can revert these after #20612.

tests/test-chat.cpp
```diff
@@ -1317,12 +1329,15 @@ static void test_template_output_peg_parsers(bool detailed_debug) {
...
+            enable_thinking(false).
+            reasoning_format(COMMON_REASONING_FORMAT_AUTO).
+            expect(message_assist).run();
 
         tst.test("I'm\nthinking\n</think>\nHello, world!\nWhat's up?")
-            .enable_thinking(false)
+            .enable_thinking(true)
             .reasoning_format(COMMON_REASONING_FORMAT_NONE)
-            .expect_content("I'm\nthinking\n</think>\nHello, world!\nWhat's up?")
+            .expect_content("<think>I'm\nthinking\n</think>\nHello, world!\nWhat's up?")
```
aldehir, line 1340: Hey, maybe now people will stop complaining about the `<think>` tag not in the output!

tools/server/webui/src/lib/services/parameter-sync.service.spec.ts
```diff
@@ -116,7 +116,7 @@ describe('ParameterSyncService', () => {
 				chat_format: '',
 				reasoning_format: '',
 				reasoning_in_content: false,
-				thinking_forced_open: false,
+				prefill: '',
```
aldehir: Should this be `generation_prompt`? A few more instances below.

tools/server/server-task.cpp
```diff
@@ -38,53 +38,53 @@ json task_params::to_json(bool only_metrics) const {
...
+            { "speculative.n_min",        speculative.n_min                                                 },
+            { "speculative.p_min",        speculative.p_min                                                 },
+            { "speculative.type",         common_speculative_type_to_str(speculative.type)                  },
+            { "speculative.ngram_size_n", speculative.ngram_size_n                                          },
+            { "speculative.ngram_size_m", speculative.ngram_size_m                                          },
+            { "speculative.ngram_m_hits", speculative.ngram_min_hits                                        },
+            { "timings_per_token",        timings_per_token                                                 },
+            { "post_sampling_probs",      post_sampling_probs                                               },
+            { "backend_sampling",         sampling.backend_sampling                                         },
+            { "lora",                     lora                                                              },
```
aldehir: Revert spacing 😊

allozaur (approved):

common/chat.cpp
pwilkin: This doesn't delegate anything from the autoparser, it's a dedicated function to add the generation prompt to the parser. I removed the `analyze_reasoning` struct in favor of just passing the start thinking marker to make it more explicit.

models/templates/llama-cpp-deepseek-r1.jinja
pwilkin: Nope, checked it but still fail. Gonna leave it to a followup PR.

common/chat-diff-analyzer.cpp
pwilkin: We're getting rid of this in the autoparser with the `| tojson` changes, but I'd probably prefer to keep this as it might come in handy at some point.

common/chat.cpp
aldehir: I was referring to the analyze_reasoning. The wrap could be removed and the rules simplified for this partially model though.

pwilkin: I'll leave you the simplification for OSS :) this PR was waaay too long already.

strawberrymelonpanda:
With OpenCode and [Unsloth's Qwen3.5 35B](https://huggingface.co/unsloth/Qwen3.5-35B-A3B-GGUF/blob/main/Qwen3.5-35B-A3B-Q4_K_S.gguf), I seem to be getting all think blocks in the response now with `</think>` attached.

image

This is with "show thinking" off.

Wasn't happening yesterday, so I'm guessing it's related?

pwilkin:
@strawberrymelonpanda argh.

Can you check with vanilla template?

strawberrymelonpanda:
Willing to, is there a command I can use to bypass unsloth's or do I need a different model?

strawberrymelonpanda:
@pwilkin looks like it's happening on Ubergarm's quant too.

image

strawberrymelonpanda:
@pwilkin I rolled back to the commit right before this PR, and I no longer see the thinking content and </think> tags.

With show thinking turn ___ON___ on commit c1258830b, using Ubergarm's quant:
image
_(turned on otherwise there'd be nothing to show it's working)_

pwilkin:
@strawberrymelonpanda can you give the exact server command?

I'm trying to repro on various Qwen3.5 4B quants but all seems correct so far.

strawberrymelonpanda:
```
GGML_CUDA_GRAPH_OPT=1 \
LLAMA_SERVER_SLOTS_DEBUG=1 \
llama-server \
--seed 1 \
--threads 8 \
--host 127.0.0.1 \
--port <port> \
--props \
--no-mmap \
--direct-io \
--models-preset ./presets.ini \
--models-max 1 \
```

```
[*]
spec-type = ngram-mod
spec-ngram-size-n = 24
draft-min = 48
draft-max = 64
parallel = 1
flash-attn = on
fit = on

; Qwen
[qwen35-35b]
load-on-startup = true
model = <path>/Qwen3.5-35B-A3B-Q4_K_S (unsloth).gguf
;model = <path>/Qwen3.5-35B-A3B-Q4_0 (ubergarm).gguf
mmproj = <path>/Qwen3.5-35B-A3B-mmproj-bf16.gguf
fit-target = 1800
ctx-size = 80000
temp = 0.6
top-p = 0.95
top-k = 20
min-p = 0.00
```

(Tested with both models)
OpenCode version 1.2.26

Start opencode in llama.cpp folder, @ commit c1b911654a581662c19eb51829705e4ed7020673 (master).

I ran the command

```Add a sleep API endpoint that triggers the result of sleep_idle_seconds```

_(I was seeing if it could, for local use, because I'd like this without a full unload - I think there's some differences?)_

For this command in specific, I first see thinking work twice:

- image  
- image  
 
Then start failing:  

- image
- image

I can't promise it's related, but my OpenCode is set to manual update and hasn't changed, and after rolling back the `</think>` tags stopped entirely.

---

The same commands on c1258830

- image
- image
- image

and continues on throughout the task without issue.

**Comment by strawberrymelonpanda** [comment]
@pwilkin  I'm using a deterministic seed, --seed 1, so with any luck maybe you can reproduce it.

Not sure how much the RNG varies based on hardware though.

**Comment by pwilkin** [comment]
Can you please try without the speculative decoding as well? (gtg to sleep now but will try to repro tomorrow)

**Comment by strawberrymelonpanda** [comment]
Same results without 
```
spec-type = ngram-mod
spec-ngram-size-n = 24
draft-min = 48
draft-max = 64
```

For that command, works twice, fails twice, etc.

**Comment by strawberrymelonpanda** [comment]
If you're not familiar with OpenCode, here's a `opencode.json` config file that should work with Llama.cpp.  It's a stripped down version of my own, don't think I removed anything important. Just some permissions and extra models.

<details>
<summary>opencode.json</summary>

```
{
  "$schema": "https://opencode.ai/config.json",  

  "share": "disabled",
  "autoupdate": false,
  "enabled_providers": ["llama-cpp"],
  
  "formatter": false,
  "lsp": false,  

  "compaction": {
    "auto": true,
   	"prune": true,
   	"reserved": 10000
  },
  
  "permission": {  	
    "webfetch": "ask",
    "websearch": "ask",    
    "bash": {
      "*": "ask",
    }
  },

  "provider": {
    "llama-cpp": {
      "npm": "@ai-sdk/openai-compatible",
      "name": "llama-server (local)",
      "options": {
        "baseURL": "http://127.0.0.1:<port>/v1"
      },
      "models": {
        "qwen35-35b": {
          "name": "qwen35-35b",
          "modalities": {
            "input": [ "text", "image" ],
            "output": [ "text" ]
          },
          "limit": {
            "context": 80128,
            "output": 99999
          }          
        }
      }
    }
  },

  "model": "llama-cpp/qwen35-35b",
  "small_model": "llama-cpp/dummy"
}
```
`"small_model": "llama-cpp/dummy"` is because there was recently an issue where, unless small model was set to something, it was sending data to their servers to get session titles from GPT Nano.  Giving it a dummy just makes it fail and use a timestamp.  

You can remove this line, but as of recent commits it'll cause the large model to try to make the title instead, which can be a significant delay, so I keep it for a fast fail.
</details>

pwilkin:
@strawberrymelonpanda I test everything on OpenCode, which is why I'm surprised to see this arise. I'll try to repro on the smaller models first, but I have a 35B quant lying around somewhere if I can't.

pwilkin:
@strawberrymelonpanda got a repro, looking into it.

pwilkin:
Aaaaand it's gone... reproduced once, set up an MITM proxy and now it's gone :P

pwilkin:
@strawberrymelonpanda happy to report I found the cause :)

Can you please check if #20825 resolves it?

strawberrymelonpanda:
Looks good. 👍 

image

winstonma:
It seems to me that this would solve the problem on my issue https://github.com/ggml-org/llama.cpp/issues/20099.

However after trying release b8461 I still see the `</think>` issue on my response.

xhinker:
This PR actually introduce a NEW BUG! 
now set reasoning_format none, will lead to:
1. during streaming, no <think> tag output, while </think>\n\n output, it break a lot logic
2. during non-streaming, everything is ok

This is A BUG

aldehir:
> This PR actually introduce a NEW BUG! 
> 
> now set reasoning_format none, will lead to:
> 
> 1. during streaming, no <think> tag output, while </think>\n\n output, it break a lot logic
> 
> 2. during non-streaming, everything is ok
> 
> 
> 
> This is A BUG

Create a new issue and fill out all the information in the template. Thank you.

xhinker:
> > This PR actually introduce a NEW BUG!
> > now set reasoning_format none, will lead to:
> > 
> > 1. during streaming, no  tag output, while \n\n output, it break a lot logic
> > 2. during non-streaming, everything is ok
> > 
> > This is A BUG
> 
> Create a new issue and fill out all the information in the template. Thank you.

here: https://github.com/ggml-org/llama.cpp/issues/23320

PR #20427, opened 2026-03-11
graph : add optional scale parameter to build_lora_mm

richarddd:
@ggerganov As discused in  https://github.com/ggml-org/llama.cpp/pull/19769. Add optional `w_s` parameter to `build_lora_mm() `for applying a multiplicative scale after the matmul. This cleans up the pattern used by bitnet and NVFP4 models where a per-tensor scale is applied after each weight multiplication.

### DIFF
--- a/src/llama-graph.h
+++ b/src/llama-graph.h
@@ -764,10 +764,11 @@ struct llm_graph_context {
              ggml_tensor * cur,
                      int   il) const;
 
-    // do mat_mul, while optionally apply lora
+    // do mat_mul, while optionally apply lora and per-tensor scale
     ggml_tensor * build_lora_mm(
               ggml_tensor * w,
-              ggml_tensor * cur) const;
+              ggml_tensor * cur,
+              ggml_tensor * w_s = nullptr) const;
 
     // do mat_mul_id, while optionally apply lora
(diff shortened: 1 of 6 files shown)


src/llama-graph.h
```diff
@@ -764,10 +764,11 @@ struct llm_graph_context {
              ggml_tensor * cur,
                      int   il) const;
 
-    // do mat_mul, while optionally apply lora
+    // do mat_mul, while optionally apply lora and per-tensor scale
     ggml_tensor * build_lora_mm(
               ggml_tensor * w,
-              ggml_tensor * cur) const;
+              ggml_tensor * cur,
+              ggml_tensor * w_s = nullptr) const;
```
ggerganov, line 771: For a follow-up PR, we should move the `cur` at the front, to increase consistency of the interfaces:

```c++
    ggml_tensor * build_lora_mm(
              ggml_tensor * cur,
              ggml_tensor * w,
              ggml_tensor * w_s = nullptr) const;
```

It will touch a lot of lines though.

ggerganov (approved):

CISC (approved):
Concur with @ggerganov on ordering.

PR #20075, opened 2026-03-03
fix: speculative decoding broken on hybrid SSM/MoE (Qwen3.5 MoE)

eauchs:
Speculative decoding on hybrid SSM/MoE models is broken right now. With a draft model you either crash immediately ("the target context does not support partial sequence removal") or end up with garbage loops. Took me a while to track down why.
Two things were wrong in find_slot: empty_cell.src was pointing to orig_cell.src instead of seq_meta.tail (so the graph was reading stale state from the wrong cell), and the copy_cell call was just... missing. On top of that, llama_memory_recurrent has no rollback mechanism at all for the SSM state when draft tokens get rejected, which is what causes the state drift.
Fix adds a checkpoint/restore with a rolling buffer (depth 8 per sequence). On Metal, ggml_backend_tensor_copy is synchronous in ggml 0.9.7 so no barrier needed.
Numbers on M3 Max 128GB:

Qwen3.5-122B-A10B-UD-Q4_K_XL + Qwen3.5-0.8B draft
Baseline: ~20.4 t/s → with patch: 23.5–29.7 t/s, acceptance rate around 63–89% depending on --draft-max
No garbage loops over extended runs

The checkpoint depth (8) and memory guard are hardcoded for now — not sure if it's worth exposing them as llama_context_params, open to feedback. VRAM overhead is n_seq_max × depth × SSM_state_size_per_layer, fine on my end but probably worth discussing for smaller devices.

PR #19408, opened 2026-02-07
server : improve context checkpoint logic

ggerganov:
fix #19394

The new checkpointing logic should be more recurrent + OpenCode friendly.

PR #20428, opened 2026-03-11
server: fix multi-turn cache reuse for hybrid/recurrent models

krystophny:
## Summary

Multi-turn conversations with hybrid models (e.g. Qwen3.5-A3B) currently force full prompt reprocessing on every turn because:

1. **Short prompts never get checkpoints**: The `pos_max >= 64` minimum threshold prevents checkpoint creation for prompts shorter than ~68 tokens. For hybrid models, even short prompts need checkpoints since the recurrent state cannot be rolled back without one.

2. **`seq_rm` failure clears everything**: When `llama_memory_seq_rm` fails (the recurrent memory has no cell-level checkpoint at the trim position), the server clears all memory instead of trying to restore from a server-level checkpoint.

### Changes

- **Lower checkpoint thresholds for hybrid/recurrent models**: Remove the `pos_max >= 64` and spacing `>= 64` minimums for hybrid/recurrent models while keeping them for SWA-only models. This ensures checkpoints are always created during prompt processing, enabling multi-turn reuse.

- **Restore checkpoint on `seq_rm` failure**: When `seq_rm` fails, search for the nearest server-level checkpoint before the trim position and restore it, instead of clearing all memory. This is a safety net for cases where the primary checkpoint restore path (the `pos_min > pos_min_thold` check) doesn't trigger.

### Test results (Qwen3.5-35B-A3B, macOS Apple Silicon)

**Q4_K_XL quantization:**

| Turn | Total prompt | Cached | New processed |
|------|-------------|--------|---------------|
| 1    | 31          | 0      | 31            |
| 2    | 48          | 27     | 21            |

**Q8_K_XL quantization:**

| Turn | Total prompt | Cached | New processed |
|------|-------------|--------|---------------|
| 1    | 31          | 0      | 31            |
| 2    | 48          | 27     | 21            |
| 3    | 65          | 44     | 21            |

Server logs confirm checkpoints are created and restored:
```
slot update_slots: created context checkpoint 1 of 32 (pos_min = 26, pos_max = 26, n_tokens = 27, size = 62.813 MiB)
slot update_slots: restored context checkpoint (pos_min = 26, pos_max = 26, n_tokens = 27, n_past = 27, size = 62.813 MiB)
```

Previously, the logs showed:
```
slot update_slots: forcing full prompt re-processing due to lack of cache data
```

### Prior work and acknowledgments

This PR builds on the checkpoint infrastructure developed across several PRs:

- #13194 @ggerganov — kv-cache: add SWA support (foundation)
- #13833 @ggerganov — SWA cache sizing
- #13979 @gabe-l-hart — hybrid recurrent cache abstraction
- #15293 @ggerganov — server: add SWA checkpoints (introduced the checkpoint mechanism this PR extends)
- #16382 @ddh0 — generalized context checkpointing to all hybrid/recurrent models
- #17009 @gabe-l-hart — hybrid context shift for multi-turn
- #18391 @o7si — fix server crash when `seq_rm` fails for hybrid models
- #19408 @ggerganov — improved checkpoint logic
- #20087 @pwilkin — `--checkpoint-every-nb` for finer-grained rollback
- #20288 @ggerganov — make 2 checkpoints near end of prompt

Related: #20075 @eauchs — speculative decoding fixes for hybrid SSM/MoE (Qwen3.5)

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -2651,11 +2651,20 @@ private:
                     const auto pos_min = llama_memory_seq_pos_min(llama_get_memory(ctx), slot.id);
                     const auto pos_max = llama_memory_seq_pos_max(llama_get_memory(ctx), slot.id);
 
-                    // no need for empty or small checkpoints
-                    do_checkpoint = do_checkpoint && (pos_min >= 0 && pos_max >= 64);
+                    {
+                        // for hybrid/recurrent models, even short prompts need checkpoints
+                        // to enable multi-turn cache reuse (recurrent state can't be rolled
+                        // back without one). SWA-only models can use a higher threshold.
+                        const bool is_hybrid_or_recurrent = llama_model_is_hybrid(model) || llama_model_is_recurrent(model);
+                        const int min_pos_max = is_hybrid_or_recurrent ? 0 : 64;
+                        const int min_spacing = is_hybrid_or_recurrent ? 0 : 64;
 
-                    // no need to create checkpoints that are too close together
-                    do_checkpoint = do_checkpoint && (slot.prompt.checkpoints.empty() || pos_max > slot.prompt.checkpoints.back().pos_max + 64);
+                        // no need for empty or small checkpoints
+                        do_checkpoint = do_checkpoint && (pos_min >= 0 && pos_max >= min_pos_max);
+
+                        // no need to create checkpoints that are too close together
+                        do_checkpoint = do_checkpoint && (slot.prompt.checkpoints.empty() || pos_max > slot.prompt.checkpoints.back().pos_max + min_spacing);
+                    }
 
                     // note: we create the checkpoint before calling llama_decode(), so the current batch is not
                     //       yet processed and therefore it is not part of the checkpoint.
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -2444,12 +2444,64 @@ private:
                     SLT_INF(slot, "n_tokens = %d, memory_seq_rm [%d, end)\n", slot.prompt.n_tokens(), p0);
 
                     if (!llama_memory_seq_rm(llama_get_memory(ctx), slot.id, p0, -1)) {
-                        SLT_WRN(slot, "failed to truncate tokens with position >= %d - clearing the memory\n", p0);
+                        SLT_WRN(slot, "failed to truncate tokens with position >= %d, searching for a checkpoint to restore\n", p0);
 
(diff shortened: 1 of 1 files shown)


pwilkin:
Seriously asking: what is the actual usefulness of rolling back to a 60-token checkpoint?

ddh0:
If this legitimately does solve the `forcing full prompt re-processing due to lack of cache data` problem in a certain scenario, I think it's worth it.

pwilkin:
> If this legitimately does solve the `forcing full prompt re-processing due to lack of cache data` problem in a certain scenario, I think it's worth it.

But what's the difference on a 90k token prompt whether you restart from 60 or from 0?

UNDEFINED-BEHAVIOR:
I've just tried out this fix locally, didn't really fixed the `forcing full prompt re-processing due to lack of cache data` issue.
Qwen3-coder-next by the way

ggerganov:
If there is a legitimate case that should be resolved, provide repro steps with logs and commands used. Caching 27 tokens is not relevant. Most of the reports that we get about this are related to clients that mutate the prefix - this cannot be solved with more checkpoints.

krystophny:
Thanks for checking! Since I am just starting to play with this with heavy help from Claude and Codex, I wanted to make this a draft PR but it slipped through. Once the fix is working for me locally, I will reopen it for discussion.

ggerganov:
It's better first to explain what you are trying to fix by providing logs, commands and repro steps. Not checkpointing short prompts is not a bug - it's a feature.

Nekotekina:
Why it's so damn impossible to allow partial success of seq_rm on hybrid models???

pwilkin:
> Why it's so damn impossible to allow partial success of seq_rm on hybrid models???

It's not, the mechanism already tries restoring the longest matching checkpoint. Full reprocessing usually means the prefix changed significantly at the very start for some reason.

krystophny:
Closing this after re-running the full OpenCode/Qwen3.5 setup. The original premise turned out to be a bad baseline on my side: I was sometimes testing a stale local binary instead of actual upstream master, and our harness/context-checkpoint settings were too pessimistic. Upstream master is usable here with the corrected runtime profile, so this is not the right PR to pursue.

Nekotekina:
> It's not, the mechanism already tries restoring the longest matching checkpoint. Full reprocessing usually means the prefix changed significantly at the very start for some reason.

I always have a full reprocessing with llama.cpp its own web interface, on hybrid models, not sure why.

PR #20128, opened 2026-03-05
CUDA: use shared mem for ssm_conv

am17an:
Add shared mem loading to `ssm_conv_long_token` for some mild benefit in prompt processing

At ub = 2048 on a 4090

| Model               | Test    |   t/s master |   t/s cuda_ssm_conv |   Speedup |
|:--------------------|:--------|-------------:|--------------------:|----------:|
| qwen35 2B Q8_0      | pp2048  |     27300.78 |            28136.03 |      1.03 |
| qwen35 2B Q8_0      | pp4096  |     26340.90 |            26944.11 |      1.02 |
| qwen35 2B Q8_0      | pp8192  |     25932.23 |            26401.70 |      1.02 |
| qwen35 2B Q8_0      | pp16384 |     24257.07 |            24673.20 |      1.02 |
| qwen35moe ?B Q4_K_S | pp2048  |      6784.72 |             6872.92 |      1.01 |
| qwen35moe ?B Q4_K_S | pp4096  |      6701.00 |             6810.24 |      1.02 |
| qwen35moe ?B Q4_K_S | pp8192  |      6684.46 |             6787.67 |      1.02 |
| qwen35moe ?B Q4_K_S | pp16384 |      6490.17 |             6578.51 |      1.01 |

PR #20430, opened 2026-03-11
CI: add hip quality check

IMbackK:
Add "hip-check" ci workflow, this workflow is intended to solve the following problems for us:

1. The hip backend has had repeated problems with people adding #pragma unroll loops that cant be unrolled see ssm-conv, softmax and https://github.com/ggml-org/llama.cpp/pull/20128#discussion_r2898155552
    * this workflow builds the hip backend with Werror to make it obvious when these occure
    * i did not make the hip builds Werror generally as we can not guarentee that there will be no warnings when built against every rocm version we support
2. Currently we want to support rocm >=6.1 so build for 6.1 in the build workflow to check that no functions from newer rocm sneak in, however we build the release against the latest version of rocm, this causes the issue that we dont notice if a pr dosent build against the latest version until the release fails
    * This workflow builds against the same version of rocm as the release to check for this
4. GCN/CDNA have a rather small 64 KiB vector register file and are wave 64, which means that only 256 registers are available. Quite often kernels get added that include significant register spill.
   * This pr adds a small script that is used in the workflow to check for kernels that spill significant registers
   * The script ignores kernels with significant spills currently in the code base (altho several of these should be investigated at some point)

The idea of this workflow that its failure NOT necessarily block a pr, but that its should prompt investigation / possible adding to the workflows witelist.

### DIFF
--- a/.github/workflows/hip-quality-check.yml
+++ b/.github/workflows/hip-quality-check.yml
@@ -52,15 +52,28 @@ jobs:
           evict-old-files: 1d
           save: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
 
-      - name: Build with Werror and check for major VGPR spills
+      - name: Build with Werror
         id: cmake_build
         run: |
           cmake -B build -S . \
             -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" \
             -DGPU_TARGETS=gfx908 \
             -DGGML_HIP=ON \
-            -DGGML_HIP_EXPORT_METRICS=On \
+            -DGGML_HIP_EXPORT_METRICS=Off \
             -DCMAKE_HIP_FLAGS="-Werror -Wno-tautological-compare" \
+            -DCMAKE_BUILD_TYPE=Release
+          cd build
+          make -j $(nproc)
+
+      - name: Check for major VGPR spills
+        id: vgpr_check
+        run: |
+          cmake -B build -S . \
+            -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" \
+            -DGPU_TARGETS=gfx908 \
+            -DGGML_HIP=ON \
+            -DGGML_HIP_EXPORT_METRICS=On \
+            -DCMAKE_HIP_FLAGS="" \
             -DCMAKE_BUILD_TYPE=Release
           cd build
           make -j $(nproc) 2>&1 | tee metrics.log | grep -v 'Rpass-analysis=kernel-resource-usage\|remark:\|^$'
--- a/.github/workflows/hip-quality-check.yml
+++ b/.github/workflows/hip-quality-check.yml
@@ -1,0 +1,108 @@
+name: HIP quality check
+
+on:
+  workflow_dispatch: # allows manual triggering
+  push:
+    branches:
+      - master
+    paths: [
+      '.github/workflows/build.yml',
+      '.github/workflows/build-linux-cross.yml',
+      '.github/workflows/build-cmake-pkg.yml',
+      '**/CMakeLists.txt',
+      '**/.cmake',
+      '**/*.h',
+      '**/*.hpp',
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
+    ]
+
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths: [
+      '.github/workflows/build.yml',
+      '.github/workflows/build-linux-cross.yml',
+      '.github/workflows/build-cmake-pkg.yml',
+      '**/CMakeLists.txt',
+      '**/.cmake',
+      '**/*.h',
+      '**/*.hpp',
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
+    ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  GGML_NLOOP: 3
+  GGML_N_THREADS: 1
+  LLAMA_LOG_COLORS: 1
+  LLAMA_LOG_PREFIX: 1
+  LLAMA_LOG_TIMESTAMPS: 1
+
+jobs:
+  ubuntu-22-hip-quality-check:
+    runs-on: ubuntu-22.04
+    container: rocm/dev-ubuntu-22.04:7.2
+    steps:
+      - name: Clone
+        id: checkout
+        uses: actions/checkout@v6
+
+      - name: Dependencies
+        id: depends
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential git cmake rocblas-dev hipblas-dev libssl-dev python3
+
+      - name: ccache
+        uses: ggml-org/ccache-action@v1.2.16
+        with:
+          key: ubuntu-22-hip-quality-check
+          evict-old-files: 1d
+          save: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+
+      - name: Build with Werror
+        id: cmake_build
+        run: |
+          cmake -B build -S . \
+            -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" \
+            -DGPU_TARGETS=gfx908 \
+            -DGGML_HIP=ON \
+            -DGGML_HIP_EXPORT_METRICS=Off \
+            -DCMAKE_HIP_FLAGS="-Werror -Wno-tautological-compare" \
+            -DCMAKE_BUILD_TYPE=Release
+          cd build
+          make -j $(nproc)
+
+      - name: Check for major VGPR spills
+        id: vgpr_check
+        run: |
+          cmake -B build -S . \
+            -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" \
+            -DGPU_TARGETS=gfx908 \
+            -DGGML_HIP=ON \
+            -DGGML_HIP_EXPORT_METRICS=On \
+            -DCMAKE_HIP_FLAGS="" \
+            -DCMAKE_BUILD_TYPE=Release
+          cd build
+          make -j $(nproc) 2>&1 | tee metrics.log | grep -v 'Rpass-analysis=kernel-resource-usage\|remark:\|^$'
+          python3 ../scripts/hip/gcn-cdna-vgpr-check.py metrics.log
--- a/.github/workflows/hip-quality-check.yml
+++ b/.github/workflows/hip-quality-check.yml
@@ -14,23 +14,9 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
     paths: [
-      '.github/workflows/build.yml',
-      '.github/workflows/build-linux-cross.yml',
-      '.github/workflows/build-cmake-pkg.yml',
-      '**/CMakeLists.txt',
-      '**/.cmake',
-      '**/*.h',
-      '**/*.hpp',
-      '**/*.c',
-      '**/*.cpp',
+      '.github/workflows/hip-quality-check.yml',
       '**/*.cu',
-      '**/*.cuh',
-      '**/*.swift',
-      '**/*.m',
-      '**/*.metal',
-      '**/*.comp',
-      '**/*.glsl',
-      '**/*.wgsl'
+      '**/*.cuh'
     ]
 
 concurrency:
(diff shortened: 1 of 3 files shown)


IMbackK:
theres a run here https://github.com/IMbackK/llama.cpp/actions/runs/22975551980/job/66703103556 to show how this works (the failure in the VGPR check is on purpose for illustration)

.github/workflows/hip-quality-check.yml
```diff
@@ -0,0 +1,108 @@
...
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
```
CISC:
```suggestion
      '.github/workflows/hip-quality-check.yml',
      '**/*.cu',
      '**/*.cuh'
```

.github/workflows/hip-quality-check.yml
```diff
@@ -0,0 +1,108 @@
...
```
CISC:
```suggestion
      '.github/workflows/hip-quality-check.yml',
      '**/*.cu',
      '**/*.cuh'
```

.github/workflows/hip-quality-check.yml
```diff
@@ -0,0 +1,108 @@
...
+        id: vgpr_check
+        run: |
+          cmake -B build -S . \
+            -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" \
+            -DGPU_TARGETS=gfx908 \
+            -DGGML_HIP=ON \
+            -DGGML_HIP_EXPORT_METRICS=On \
+            -DCMAKE_HIP_FLAGS="" \
+            -DCMAKE_BUILD_TYPE=Release
+          cd build
```
CISC, line 78:
```suggestion
      - name: Build with Werror and check for major VGPR spills
        id: cmake_build
        run: |
          cmake -B build -S . \
            -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" \
            -DGPU_TARGETS=gfx908 \
            -DGGML_HIP=ON \
            -DGGML_HIP_EXPORT_METRICS=On \
            -DCMAKE_HIP_FLAGS="-Werror -Wno-tautological-compare" \
            -DCMAKE_BUILD_TYPE=Release
          cd build
```
Any reason we can't merge both like this?

scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -0,0 +1,148 @@
...
+import sys
+from collections import defaultdict
+
+def parse_log_file(filepath):
+    """Parse log file and extract function VGPR usage."""
+    import re
+    
+    functions = defaultdict(lambda: {'vgprs': 0, 'spill': 0, 'location': ''})
+    
+    try:
```
CISC, line 13:
```suggestion
from collections import defaultdict


def parse_log_file(filepath):
    """Parse log file and extract function VGPR usage."""
    import re

    functions = defaultdict(lambda: {'vgprs': 0, 'spill': 0, 'location': ''})

    try:
```

scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -0,0 +1,148 @@
...
+        with open(filepath, 'r') as f:
+            content = f.read()
+            # Find all function entries with VGPR usage including location
+            pattern = r'([^:]+:\d+):.*?Function Name: (\S+).*?VGPRs: (\d+).*?VGPRs Spill: (\d+)'
+            matches = re.findall(pattern, content, re.DOTALL)
+            
+            for location, func_name, vgprs, spill in matches:
```
CISC, line 20:
```suggestion
            matches = re.findall(pattern, content, re.DOTALL)

            for location, func_name, vgprs, spill in matches:
```

scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -0,0 +1,148 @@
...
+                    functions[func_name]['location'] = location
+    except FileNotFoundError:
+        print(f"Error: File {filepath} not found", file=sys.stderr)
+        sys.exit(1)
+
+    return functions
+
+def main():
+    if len(sys.argv) < 2:
+        print("Usage: ./vgpr_check.py <log_file>", file=sys.stderr)
```
CISC:
```suggestion
        print(f"Error: File {filepath} not found", file=sys.stderr) # noqa: NP100
        sys.exit(1)

    return functions


def main():
    if len(sys.argv) < 2:
        print("Usage: ./vgpr_check.py <log_file>", file=sys.stderr) # noqa: NP100
```

scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -0,0 +1,148 @@
...
+    functions = parse_log_file(log_file)
+    found_issues = False
+
+    # First print all ignored functions (deduplicated)
+    printed_ignored = set()
+    for func_name, data in sorted(functions.items()):
+        total_vgprs = data['vgprs'] + data['spill']
+        if total_vgprs > 256 and func_name in ignored and func_name not in printed_ignored:
+            location = data.get('location', log_file)
+            print(f"{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) [IGNORED]")
```
CISC:
```suggestion
            print(f"{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) [IGNORED]") # noqa: NP100
```

scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -0,0 +1,148 @@
...
+            # Print in red if not ignored
+            color_code = "\033[91m" if func_name not in ignored else ""
+            reset_code = "\033[0m" if func_name not in ignored else ""
+            print(f"{color_code}{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) {status}{reset_code}")
+            if func_name not in ignored:
+                found_issues = True
+
+    sys.exit(1 if found_issues else 0)
+
+if __name__ == "__main__":
```
CISC:
```suggestion
            print(f"{color_code}{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) {status}{reset_code}") # noqa: NP100
            if func_name not in ignored:
                found_issues = True

    sys.exit(1 if found_issues else 0)


if __name__ == "__main__":
```

IMbackK:
@CISC the hip compiler will ignore `#pragma clang diagnostic push` when --save-temps is active. For this reason a build with `-DGGML_HIP_EXPORT_METRICS=On` and `-Werror` can not succeed as warning will be emmited here: https://github.com/ggml-org/llama.cpp/blob/f2ab047f275780987ac028f7368cdad32863a36f/ggml/src/ggml-cuda/softmax.cu#L51 among other places.

CISC:
Better split the jobs again then.

IMbackK:
Right

CISC:
Check CI failures.

IMbackK:
this pr should be in principal ready now, but i think this needs some buy in from the maintainers of the cuda backend this also affects.

@JohannesGaessler @am17an @slaren thoughts on principal?

CISC:
Guess we got one:
https://github.com/ggml-org/llama.cpp/actions/runs/23210660849/job/67458238374?pr=20430#step:6:155

.github/workflows/hip-quality-check.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: actions/checkout@v6
+
+      - name: Dependencies
+        id: depends
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential git cmake rocblas-dev hipblas-dev libssl-dev python3
+
+      - name: ccache
+        uses: ggml-org/ccache-action@v1.2.16
```
CISC:
```suggestion
        uses: ggml-org/ccache-action@v1.2.21
```
Just updated it. :)

scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -0,0 +1,157 @@
...
+        '_ZL18flash_attn_ext_f16ILi96ELi96ELi4ELi8ELb0ELb0EEvPKcS1_S1_S1_S1_PKiPfP15HIP_vector_typeIfLj2EEffffjfiS5_IjLj3EEiiiiiiiiiiiliiliiiiil'
+    }
+
+    functions = parse_log_file(log_file)
+    found_issues = False
+
+    # First print all ignored functions (deduplicated)
+    printed_ignored = set()
+    for func_name, data in sorted(functions.items()):
+        total_vgprs = int(data['vgprs']) + int(data['spill'])
```
CISC, line 134: Seems unnecessary:
https://github.com/ggml-org/llama.cpp/blob/c2bb6194b2aca37ab4b43a5a44136d046e4c6290/scripts/hip/gcn-cdna-vgpr-check.py#L21-L22

scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -0,0 +1,157 @@
...
+        if total_vgprs > 256 and func_name in ignored and func_name not in printed_ignored:
+            location = data.get('location', log_file)
+            print(f"{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) [IGNORED]") # noqa: NP100
+            printed_ignored.add(func_name)
+
+    # Then print new functions with issues in red
+    for func_name, data in sorted(functions.items()):
+        total_vgprs = int(data['vgprs']) + int(data['spill'])
```
CISC, line 142: ditto

scripts/hip/gcn-cdna-vgpr-check.py
IMbackK: the python-type-check fails without this, im actually not sure why but it thinks it could be str.

IMbackK:
> Guess we got one: https://github.com/ggml-org/llama.cpp/actions/runs/23210660849/job/67458238374?pr=20430#step:6:155

yeah i opened a pr for that: https://github.com/ggml-org/llama.cpp/pull/20696

scripts/hip/gcn-cdna-vgpr-check.py
CISC: Ah, it's just being dumb. :)

CISC (approved):

scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -0,0 +1,157 @@
...
+        '_ZL18flash_attn_ext_f16ILi128ELi128ELi32ELi1ELb0ELb0EEvPKcS1_S1_S1_S1_PKiPfP15HIP_vector_typeIfLj2EEffffjfiS5_IjLj3EEiiiiiiiiiiiliiliiiiil',
+        '_ZL9mul_mat_qIL9ggml_type2ELi112ELb0EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
+        '_ZL18flash_attn_ext_f16ILi112ELi112ELi32ELi2ELb0ELb0EEvPKcS1_S1_S1_S1_PKiPfP15HIP_vector_typeIfLj2EEffffjfiS5_IjLj3EEiiiiiiiiiiiliiliiiiil',
+        '_ZL18flash_attn_ext_f16ILi112ELi112ELi4ELi8ELb0ELb0EEvPKcS1_S1_S1_S1_PKiPfP15HIP_vector_typeIfLj2EEffffjfiS5_IjLj3EEiiiiiiiiiiiliiliiiiil',
+        '_ZL18flash_attn_ext_f16ILi128ELi128ELi32ELi1ELb1ELb0EEvPKcS1_S1_S1_S1_PKiPfP15HIP_vector_typeIfLj2EEffffjfiS5_IjLj3EEiiiiiiiiiiiliiliiiiil',
+        '_ZL18flash_attn_ext_f16ILi128ELi128ELi32ELi2ELb1ELb0EEvPKcS1_S1_S1_S1_PKiPfP15HIP_vector_typeIfLj2EEffffjfiS5_IjLj3EEiiiiiiiiiiiliiliiiiil',
+        '_ZL18flash_attn_ext_f16ILi128ELi128ELi4ELi8ELb1ELb0EEvPKcS1_S1_S1_S1_PKiPfP15HIP_vector_typeIfLj2EEffffjfiS5_IjLj3EEiiiiiiiiiiiliiliiiiil',
+        '_ZL18flash_attn_ext_f16ILi96ELi96ELi4ELi8ELb0ELb0EEvPKcS1_S1_S1_S1_PKiPfP15HIP_vector_typeIfLj2EEffffjfiS5_IjLj3EEiiiiiiiiiiiliiliiiiil',
```
am17an, line 125: Can we store the unmangled names? Are these not compiler dependent

am17an (approved):

scripts/hip/gcn-cdna-vgpr-check.py
IMbackK: Name mangling is in theory compiler dependent, however on linux all compilers follow the itanium abi spec for mangling so this is not an issue for this workflow. Additionally, HIP code can only be compiled by clang anyhow and this script depends on a specific feature of the amdclang fork (the metrics output).

The compiler dose not output demangled names in its hip metrics output so the only way to have  unmangled names here would be for the script to unmangle the names itself in accordance to itanium spec im not sure this is worth it.

JohannesGaessler (approved):
Based on static code analysis this seems correct to me.

PR #20432, opened 2026-03-11
Fix server tests to use `reasoning` instead of `reasoning_budget`

pwilkin:
Resolves CI tests broken after the change introduced by #20297

### DIFF
--- a/tools/server/tests/unit/test_template.py
+++ b/tools/server/tests/unit/test_template.py
@@ -11,6 +11,7 @@ sys.path.insert(0, str(path))
 
 import datetime
 from utils import *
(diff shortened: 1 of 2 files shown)


CISC (approved):

ggerganov (approved):

CISC:
Trying to clear the queue, not sure why `server` is stuck for so long while `build` later in the queue appears to at least start some of the jobs.

pwilkin:
Yeah, they've been this way since yesterday, wonder what's going on :/

CISC:
Finally! :)

PR #20437, opened 2026-03-12
ggml : transpose fused GDN state access for coalesced memory reads

arkavo-com:
## Summary

Fixes #20436

- **Transpose state indexing** in the fused GDN kernel across Metal, CUDA, and CPU so threads read contiguously instead of column-strided
- **Add `--fused-gdn [on|off|auto]` CLI flag** (mirrors `--flash-attn`) to control fused GDN independently
- **Use `ggml_vec_dot_f32`** in CPU kernel for SIMD-optimized dot products
- **Couple AR/chunked fused flags** in auto-detection to prevent state layout mismatch

### Root cause

The fused GDN kernel's `[S_v, S_v]` state matrix was stored row-major but accessed column-wise. A 32-thread SIMD group hit 32 different cache lines but used only 4 bytes each (~3% utilization). At S_v=128 (Qwen3.5-9B), 32 heads x 128x128 state (2MB/layer x 24 recurrent layers) spills L2 cache, amplifying the penalty.

### Fix

Store the state matrix transposed: `M[col][row] = S[row][col]`. Reading "column col of S" becomes "row col of M" = contiguous. Since S is square and the fused kernel consistently reads/writes in this format, mathematical equivalence is maintained.

## Benchmark results

**Apple M4 Max, Metal, Qwen3.5 Q4_K_M** (prompt: 4 tokens, generation: 16 tokens)

| Model | Mode | Prompt (t/s) | Generation (t/s) | Prompt Delta | Gen Delta |
|-------|------|-------------|-------------------|----------|-------|
| **0.8B** | Fused OFF | 402.5 | 170.7 | -- | -- |
| **0.8B** | Fused ON (fix) | 416.5 | 233.1 | **+3.5%** | **+36.5%** |
| **9B** | Fused OFF | 109.3 | 53.6 | -- | -- |
| **9B** | Fused ON (fix) | 140.3 | 65.5 | **+28.4%** | **+22.2%** |
| **27B** | Fused OFF | 32.9 | 14.7 | -- | -- |
| **27B** | Fused ON (fix) | 40.3 | 16.7 | **+22.5%** | **+13.6%** |

The 9B regression (previously -39% with fused GDN) is fully resolved. Fused GDN is now faster than the unfused path across all model sizes.

## Test plan

- [x] `test-backend-ops test -o GATED_DELTA_NET` -- 13/13 pass on Metal
- [x] Benchmark Qwen3.5-9B to confirm regression is resolved
- [x] Benchmark Qwen3.5-0.8B and 27B to confirm no regressions
- [x] Verify `--fused-gdn off` disables fused path
- [ ] Test on CUDA backend

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -1339,6 +1339,21 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
                                    string_format("error: unknown value for --flash-attn: '%s'\n", value.c_str()));
                            }
                        }).set_env("LLAMA_ARG_FLASH_ATTN"));
+    add_opt(common_arg({ "-fgdn", "--fused-gdn" }, "[on|off|auto]",
+                       string_format("set fused Gated Delta Net kernel use ('on', 'off', or 'auto', default: '%s')",
+                                     llama_fused_gdn_type_name(params.fused_gdn_type)),
+                       [](common_params & params, const std::string & value) {
+                           if (is_truthy(value)) {
+                               params.fused_gdn_type = LLAMA_FUSED_GDN_TYPE_ENABLED;
+                           } else if (is_falsey(value)) {
+                               params.fused_gdn_type = LLAMA_FUSED_GDN_TYPE_DISABLED;
+                           } else if (is_autoy(value)) {
+                               params.fused_gdn_type = LLAMA_FUSED_GDN_TYPE_AUTO;
+                           } else {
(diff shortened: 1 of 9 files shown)


ggerganov:
Thanks, I'll continue this in #20443

PR #20441, opened 2026-03-12
ggml-vulkan: disable transfer queue on UMA

### LINKED ISSUE #20439: Eval bug: Regression when trying to load a big model
### Name and Version

./build_vulkan/bin/llama-cli --version
ggml_vulkan: 0 = Radeon 8060S Graphics (RADV STRIX_HALO) (radv) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 64 | shared memory: 65536 | int dot: 1 | matrix cores: KHR_coopmat
version: 8183 (66d65ec29) (ignore that, it was the last good commit as I bisected, bisected from HEAD which was d63aa398de455cf8243802c5099dc30a2e26edcc at the time)
built with GNU 15.2.1 for Linux x86_64

### Operating systems

Linux

### GGML backends

Vulkan

### Hardware

Ryzen AI Max 395+

### Models

Issue encountered with big models(specifically in my case Qwen3.5-122B-A10B-UD-Q6_K_XL and NVIDIA-Nemotron-3-Super-120B-A12B-UD-Q6_K)

command line used:
```
./build_vulkan/bin/llama-server --temp 0.6 --top-p 0.95 --min-p 0.0 --top-k 20 -fitt 5120 -ngl 999 --no-mmap --host 192.168.1.105 --port 3334 -m ./models/Qwen3.5-122B-A10B-UD-Q6_K_XL-00001-of-00004.gguf
```

After the commit specified below, models fail to load, the whole system nearly hangs, I have to killall on console terminal which takes a while to trigger. I've experienced this in the past when memory usage was too high(as ram and vram are shared on this mini PC).

### First Bad Commit

Bisected to precisely:
# first bad commit: [319146247e643695f94a558e8ae686277dd4f8da] vulkan: improve partial offloading performance on AMD (#19976)

### Relevant log output

Nothing relevant in dmesg beyond journalctl complaining about memory pressure. Nothing in dmesg and llama-server log just shows it around midway loading the model with dots.

RipleyTom:
Fixes https://github.com/ggml-org/llama.cpp/issues/20439 .

I am unsure why transfer queues add such a huge overhead(it was at least 10GB as tried reserving that much when loading the model and it still choked)

Summary from big C:

Using the separate SDMA/transfer queue for async copies is counterproductive here because:

1. There's no separate device memory to transfer to/from — it's all unified memory
2. The SDMA engine and its associated driver structures (kernel buffer objects for command streams, page table entries, cross-queue synchronization state) consume memory from the same pool
3. The timeline semaphore synchronization between compute and transfer queues adds driver overhead with no benefit — the compute queue can issue buffer copies just as efficiently on UMA
4. The transfer queue's command pool (`transfer_cmd_pool`) and its command buffers accumulate during model loading alongside compute_cmd_pool, effectively doubling the command infrastructure

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -5401,7 +5401,7 @@ static vk_device ggml_vk_get_device(size_t idx) {
 
         ggml_vk_load_shaders(device);
 
-        const bool prefers_transfer_queue = device->vendor_id == VK_VENDOR_ID_AMD && device->architecture != AMD_GCN;
+        const bool prefers_transfer_queue = device->vendor_id == VK_VENDOR_ID_AMD && device->architecture != AMD_GCN && !device->uma;
 
         if (!device->single_queue) {
             const uint32_t transfer_queue_index = compute_queue_family_index == transfer_queue_family_index ? 1 : 0;


RipleyTom:
Fixed by 5866e3bbc8a2ede4236871ea73a3eb3337abf530 in a better way.

PR #20442, opened 2026-03-12
convert : fix/suppress pyright errors

danbev:
This commit fixes the pyright errors that are generated by pyright for convert_hf_to_gguf.py.

The motivation for this is that running this locally generates errors that CI does not, and it can be difficult to spot new errors. One use case is when working on new models which cannot be run in CI due to privacy. Having the ability to run pyright locally is would be helpful in this cases.

In the linked issue there is the mention of switching to `ty` which I don't know anything about but in the meantime I would appreciate if we could suppress these errors for now, and later perhaps revert this commit.

With this change there are no errors but there are 4 informations messages if the `mistral_common` package is installed. The `--level error` flag can be used to suppress them.

Resolves: https://github.com/ggml-org/llama.cpp/issues/20417

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -2201,7 +2201,7 @@ class GPTNeoXModel(TextModel):
             # Map bloom-style qkv_linear to gpt-style qkv_linear
             # bloom: https://github.com/huggingface/transformers/blob/main/src/transformers/models/bloom/modeling_bloom.py#L238-L252  # noqa
             # gpt-2: https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/modeling_gpt2.py#L312  # noqa
-            qkv_weights = data_torch.reshape(n_head, 3, n_embed // n_head, n_embed)
+            qkv_weights = data_torch.reshape((n_head, 3, n_embed // n_head, n_embed))
             data_torch = torch.cat(
                 (
                     qkv_weights[:, 0, :, :].reshape((-1, n_embed)),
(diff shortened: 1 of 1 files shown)


convert_hf_to_gguf.py
```diff
@@ -2194,12 +2194,14 @@ def set_gguf_parameters(self):
...
         n_embed = self.hparams.get("hidden_size", self.hparams.get("n_embed"))
+        assert n_head is not None
+        assert n_embed is not None
 
         if re.match(r"gpt_neox\.layers\.\d+\.attention\.query_key_value\.weight", name):
             # Map bloom-style qkv_linear to gpt-style qkv_linear
             # bloom: https://github.com/huggingface/transformers/blob/main/src/transformers/models/bloom/modeling_bloom.py#L238-L252  # noqa
             # gpt-2: https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/modeling_gpt2.py#L312  # noqa
-            qkv_weights = data_torch.reshape((n_head, 3, n_embed // n_head, n_embed))
+            qkv_weights = data_torch.reshape(n_head, 3, n_embed // n_head, n_embed)
```
```suggestion
            qkv_weights = data_torch.reshape((n_head, 3, n_embed // n_head, n_embed))
```
Both are correct, but no reason to change this.

convert_hf_to_gguf.py
```diff
@@ -2243,14 +2247,16 @@ def set_gguf_parameters(self):
...
+        assert n_embed is not None
 
         name = re.sub(r'transformer\.', '', name)
 
         if re.match(r"h\.\d+\.self_attention\.query_key_value\.weight", name):
             # Map bloom-style qkv_linear to gpt-style qkv_linear
             # bloom: https://github.com/huggingface/transformers/blob/main/src/transformers/models/bloom/modeling_bloom.py#L238-L252  # noqa
             # gpt-2: https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/modeling_gpt2.py#L312  # noqa
-            qkv_weights = data_torch.reshape((n_head, 3, n_embed // n_head, n_embed))
+            qkv_weights = data_torch.reshape(n_head, 3, n_embed // n_head, n_embed)
```
```suggestion
            qkv_weights = data_torch.reshape((n_head, 3, n_embed // n_head, n_embed))
```

CISC (approved):

convert_hf_to_gguf.py
danbev: Thanks, I've update this now.

PR #20445, opened 2026-03-12
tests: Fix invalid iterator::end() dereference in common_regex

rillomas:
When compiling with VS2026 18.4 I noticed `test-regex-partial` crashes immediately with debug build. 

image

I tracked this down to an iterator::end() dereference in the following test case which was [occurring here.](https://github.com/ggml-org/llama.cpp/blob/de190154c85d20e24dbeae8c8af1849402ae5098/common/regex-partial.cpp#L105)

```cpp
    assert_equals<std::string>(
        "^(a*)",
        regex_to_reversed_partial_regex("a*"));
```

This fix adds bounds checking to prevent such dereferencing.

### DIFF
--- a/common/regex-partial.cpp
+++ b/common/regex-partial.cpp
@@ -102,7 +102,7 @@ std::string regex_to_reversed_partial_regex(const std::string & pattern) {
                 auto is_star = *it == '*';
                 ++it;
                 if (is_star) {
-                    if (*it == '?') {
+                    if (it != end && *it == '?') {
                         ++it;
                     }
                 }


rillomas:
Hi @ggerganov, @CISC can you have a look at this (or assign someone sufficient)? Thanks

aldehir (approved):

aldehir:
I believe we have eliminated most, if not all, uses of the partial regex implementation.

Still, I reproduced this and the fix works. Thanks!

PR #20457, opened 2026-03-12
cmake : fix build warning when kleidiai is enabled

chaxu01:


ggml/src/ggml-cpu/kleidiai/kleidiai.cpp
```diff
@@ -187,7 +187,7 @@ static void init_kleidiai_context(void) {
         initialized = true;
 
         const char *env_sme     = getenv("GGML_KLEIDIAI_SME");
-        const char *env_threads = getenv("GGML_TOTAL_THREADS");
+        const char *env_threads = getenv("LLAMA_ARG_THREADS");
```
ggerganov: We should not reference llama vars in the ggml backend. There could be other applications from llama.cpp that use ggml.

chaxu01: Good point. LLAMA_ARG_THREADS is llama-specific and shouldn’t be referenced from the ggml backend. I’ll switch this back and keep the KleidiAI thread handling limited to ggml/backend-level inputs only.

chaxu01:
@ggerganov I’ve addressed the previous review comment (removed the llama-specific env usage in ggml). There are currently two CI failures, but they don’t appear related to this change (no functional code changes beyond the CMake warning fix). Would appreciate another look when you have a chance. Thanks.

ggerganov (approved):

PR #17795, opened 2025-12-05
CUDA:  Improve performance via less synchronizations between token

aendk:
See [comment below](https://github.com/ggml-org/llama.cpp/pull/17795#issuecomment-3675566278)

--------------
This PR suggest to remove some superfluous synchronization calls between tokens to be faster on CUDA backends. I see between 1% and 2% perf gain depending on the model, GPU and settings. 

### Mechanism
The performance impact is best explained visually. Here are the "before" and "after" nsight system traces. They are not to scale. The relevant part is the row with the green and red bubbles. Both images show the overhead between the GPU execution of two tokens. The generation of the n-th token ends on the left hand side of the screenshot, in the  green bubble titled `cudaStreamSynchronize`. The calculation of the next/n+1st token starts on the right hand side, at the green bar titled `cudaGraphLaunch`. In between, there is CPU orchestration overhead. This PR aims to shrink the time spent in the middle, between GPU token generation. Original:

Screenshot 2025-12-05 at 14 45 35

In the middle of the above image, we see red and green bubbles alternating. In this case, the green bubbles are synchronization steps, the red bubbles are asynchronous copy calls from host to device. If async operations are immediately followed by synchronization calls, they are executed synchronously. This is not efficient. Removing the green synchronization operations between asynchronous copy calls leads to asynchronous copies and reduced overhead between GPU token generation:

Screenshot 2025-12-05 at 14 45 10

### Performance
I benchmarked on a RTX Pro 6000 Blackwell using `./llama-bench -m $models -p 0 -n 128,256,512 -fa 1`.
My testing shows around 1% improvement, with `gpt-oss-20b` gaining up to 1.4%. `llama 3B Q4_K - Medium` shows very high variance, prompting me to run the tests again with ` -r 100`. At `-r 100`, a clearer trend  of improved performance for `gemma3n E2B Q8_0` is also visible.

<details>
<summary>Details with default `-r 5` </summary>

Baseline:
```
| model                          |       size |     params | backend    | ngl | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -: | --------------: | -------------------: |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |  1 |           tg128 |        392.24 ± 1.07 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |  1 |           tg256 |        392.72 ± 0.35 |
[...]

PR #20463, opened 2026-03-12
llama : disable graph reuse with pipeline parallelism

ggerganov:
The following repro demonstrates the issue:

```bash
make -j && ./bin/llama-perplexity -hf ggml-org/Qwen3-0.6B-GGUF -f wiki.test.raw --chunks 16 -ngl 99 -ub 512 -b 2048

PPL = 2382.4719 +/- 246.20903
```

The problem seems to occur when 2 consecutive pp ubatches both output logits, which occurs in perplexity runs with the above parameters: 4 ubatches of size 512, the second 2 ubtaches output logits. I think the graph reuse logic somehow conflicts with the scheduler's logic for tracking the current copy:

https://github.com/ggml-org/llama.cpp/blob/557fe2d9132913eaf08c8abf21b0cff61addb9ac/ggml/src/ggml-backend.cpp#L1775-L1780

For now disabling graph reuse when pipeline parallelism is active to workaround. Proper investigation is necessary.

--- 

Additionally, after https://github.com/ggml-org/llama.cpp/pull/17795, only disabling the graph reuse is not enough to fix the issue. So for now also reverting that change.

Note that both commits in this PR are needed. Neither one fixes the issue alone.

cc @aendk @gaugarg-nv

### DIFF
--- a/ggml/src/ggml-backend.cpp
+++ b/ggml/src/ggml-backend.cpp
@@ -1455,10 +1455,6 @@ static enum ggml_status ggml_backend_sched_compute_splits(ggml_backend_sched_t s
         int split_backend_id = split->backend_id;
         ggml_backend_t split_backend = sched->backends[split_backend_id];
 
-        if (sched->events[split_backend_id][sched->cur_copy] == NULL) {
-            ggml_backend_synchronize(split_backend);
-        }
-
         // copy the input tensors to the split backend
         for (int input_id = 0; input_id < split->n_inputs; input_id++) {
             ggml_backend_t input_backend = ggml_backend_sched_get_tensor_backend(sched, split->inputs[input_id]);
@@ -1469,12 +1465,16 @@ static enum ggml_status ggml_backend_sched_compute_splits(ggml_backend_sched_t s
                 // inputs from the user must be copied immediately to prevent the user overwriting the data before the copy is done
                 if (sched->events[split_backend_id][sched->cur_copy] != NULL) {
                     ggml_backend_event_synchronize(sched->events[split_backend_id][sched->cur_copy]);
+                } else {
+                    ggml_backend_synchronize(split_backend);
                 }
-                ggml_backend_tensor_copy_async(input_backend, split_backend, input, input_cpy);
+                ggml_backend_tensor_copy(input, input_cpy);
             } else {
                 // wait for the split backend to finish using the input before overwriting it
                 if (sched->events[split_backend_id][sched->cur_copy] != NULL) {
                     ggml_backend_event_wait(split_backend, sched->events[split_backend_id][sched->cur_copy]);
+                } else {
+                    ggml_backend_synchronize(split_backend);
                 }
 
                 // when offloading MoE weights, we can reduce the amount of data copied by copying only the experts that are used
@@ -1576,10 +1576,6 @@ static enum ggml_status ggml_backend_sched_compute_splits(ggml_backend_sched_t s
                     }
                 }
             }
-        }
-
-        if (sched->events[split_backend_id][sched->cur_copy] == NULL) {
-            ggml_backend_synchronize(split_backend);
         }
 
         if (!sched->callback_eval) {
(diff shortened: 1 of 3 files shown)


JohannesGaessler (approved):

ORippler (approved):
Let's take the time to understand what's going on here

Superbobo75:
hi,

Starting from version `c8323` of `llama.cpp`, where the graph parameter was disabled, the inference speed of the model `Unsloth/Qwen3.5-35B-A3B-Q5_K_M.gguf ` dropped significantly. On my specific hardware setup—consisting of 2x RTX 5060 Ti (16 GB VRAM) and 64 GB DDR4 RAM—the performance fell from approximately 85 tokens per second (t/s) to around 50 t/s.

Due to this substantial regression, I am currently sticking with the last known functional version prior to this change. I have not yet tested other models to see if this issue is isolated to this specific architecture or configuration.

aendk:
@ggerganov thanks for the repro. 
I am currently looking into this.
For preliminary results, I compared the PPL of three versions of the code:
1. The build of this merged code (`b8323`) as ground truth
2. The previous build (`b8322`)
3. The previous build (`b8322`), but with `copy_from_host` set to false

In my setup (RTX PRO 6000, 4500), I get the following PPLs:
1. `PPL = 28.8008 +/- 1.50705`
2.  `PPL = 3964.6876 +/- 382.88199`
3. `PPL = 28.8008 +/- 1.50705`

~This matches with my initial hypothesis, and with my concerns I voiced in the initial PR (https://github.com/ggml-org/llama.cpp/pull/17795#issuecomment-3682816188).
My aim was to only relax CPU _split_ -> CUDA _split_ synchronization, as they are inherently synced. This is what I wanted to explicitly check via `ggml_backend_is_cpu(backend_src)`.
Due to API constraints, I relaxed this check to allow for CPU _buffer_ to CUDA _buffer_ copies, in addition to checking `ggml_backend_dev_type(backend_src->device) == GGML_BACKEND_DEVICE_TYPE_CPU`.
In the pipeline parallelism setting here, the check is not stringent enough. This new path is taken where it shouldn't~


~The fix for this is therefore to make the `copy_from_cpu` check more stringent, and to test thoroughly with llama-perplexity before re-proposing this. Alternatively, disabling this if we are in a multi-GPU setting could be a workaround.~ 
**See comment below**


Regarding the graph reuse logic, I need to dig deeper. According to the PPL results above, just disabling the incorrect part of my PR is enough. Did you test on workstation GPUs too, or were you also using other GPUs (consumer/GeForce or AMD)?

aendk:
With further analysis, I need to retract my first assumption.

The `copy_from_host` change was not directly at fault here. 
In my testing, pipeline parallelism with 2 GPUs leads to 3 splits: input embedding split on CPU, GPU0, and GPU1.
Between GPU0 and GPU1, activation tensors were copied asynchronously as before, and input vectors like weights were also copied correctly from CPU to GPU2. I saw no incorrect use of the new `copy_from_host`.

I rather think that the event-based scheduling mechanism implicitly relied on synchronizations in the blocking CPU->GPU copy operations, and that removing them surfaced an already existing bug/blurry edge case in the event-based scheduling mechanism. 
To be specific, each CPU-> GPU copy synced twice,  a standalone `ggml_backend_cuda_synchronize(split_backend)`, and then a `cudaStreamSynchronize(cudaStreamPerThread)` in `ggml_backend_tensor_copy`. With my PR, every copy was completely asynchronous.

I briefly tested the assumption that the event-based scheduling mechanism for pipeline parallelism implicitly requires similar stream synchronizations to the single GPU case. 
In my initial PR, we went for the `saaasg` pattern for single GPU setups (`s`=sync, `a`= async copies, `g`= graph execution)
 My final implementation executed this new additional `s` just before the graph execution (`g`) in single GPU setups only, and not for pipeline parallelism.  The assumption was that pipeline parallelism does not require this.
However, executing this new `s`  (`ggml_backend_cuda_synchronize(split_backend)`) now also during pipeline parallelism seems to fix the bug. `llama-perplexity`  shows identical values to `master` and performance on linux is unchanged too. 

I will open a new draft PR for further discussions on this.

PR #20479, opened 2026-03-13
server : support reasoning parameters from OAI, OpenRouter and Claude

gyf304:
Adds compatibility for reasoning on/off and effort from OAI, OpenRouter and Claude
- OAI: [`reasoning_effort`](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create#reasoning_effort:~:text=reasoning_effort,-%3A%20optional%20ReasoningEffort)
- OpenRouter: [`reasoning.effort`](https://openrouter.ai/docs/api/api-reference/chat/send-chat-completion-request#request.body.reasoning)
- Claude: [`thinking.type`](https://platform.claude.com/docs/en/api/messages/create#create.thinking) and [`output_config.effort`](https://platform.claude.com/docs/en/api/messages/create#create.output_config.effort)

### DIFF
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -1026,9 +1026,48 @@ json oaicompat_chat_params_parse(
         llama_params["parse_tool_calls"] = true;
     }
 
-    // merge the template args provided from command line with the args provided in the user request
-    auto chat_template_kwargs_object = json_value(body, "chat_template_kwargs", json::object());
     inputs.chat_template_kwargs = opt.chat_template_kwargs;
+
+    // parse the oai compatible "reasoning_effort" or openrouter compatible "reasoning.effort"
+    // to override "chat_template_kwargs.enable_thinking"
+    auto reasoning_effort = json_value(
+        body,
+        "reasoning_effort",
+        json_value(
+            json_value(body, "reasoning", json::object()),
+            "effort",
+            std::string("")
+        )
+    );
+    if (reasoning_effort == "none") {
+        inputs.chat_template_kwargs["enable_thinking"] = "false";
+    } else if (reasoning_effort != "") {
+        inputs.chat_template_kwargs["enable_thinking"] = "true";
+    }
+    // override "chat_template_kwargs.reasoning_effort"
+    if (reasoning_effort != "") {
+        inputs.chat_template_kwargs["reasoning_effort"] = json(reasoning_effort).dump();
+    }
+
+    // parse the claude compatible "thinking.type" arg to override "chat_template_kwargs.enable_thinking"
+    auto claude_thinking = json_value(body, "thinking", json::object());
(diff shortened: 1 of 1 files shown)


tarruda:
Note that it is already possible to turn on/off thinking by passing chat template parameters directly via the API.

gyf304:
> Note that it is already possible to turn on/off thinking by passing chat template parameters directly via the API.

Yes, it's also possible to do this with the built-in web UI by specifying a custom JSON.

This is more for compatibility with APIs with other vendors, which is useful if using frontends like the claude CLI or the codex CLI.

sjoelund:
I was looking into adding something like this, but it's good to know someone worked on it. It would resolve the discussion in https://github.com/ggml-org/llama.cpp/discussions/20408 . @pwilkin any comments? Does the qwen3.5 default chat template handle setting some reasoning budget if low/medium/high are passed, or would those options only be used for models that support them?

tools/server/server-common.cpp
```diff
@@ -1026,9 +1026,48 @@ json oaicompat_chat_params_parse(
...
+        inputs.chat_template_kwargs["enable_thinking"] = "false";
+    } else if (reasoning_effort != "") {
+        inputs.chat_template_kwargs["enable_thinking"] = "true";
+    }
+    // override "chat_template_kwargs.reasoning_effort"
+    if (reasoning_effort != "") {
+        inputs.chat_template_kwargs["reasoning_effort"] = json(reasoning_effort).dump();
+    }
+
+    // parse the claude compatible "thinking.type" arg to override "chat_template_kwargs.enable_thinking"
```
sjoelund, line 1052: Shouldn't this be handled by convert_anthropic_to_oai though?

Djip007: For me yes: llama-server have ~OpenIA API so for std use it is more simple to have only OpenIA Rest API.
And update as needed the jina template.

For exemple look Mistral 4 convertion have it like that: https://huggingface.co/unsloth/Mistral-Small-4-119B-2603-GGUF

Will be nice if we can have it so we can use it with CLI/IDE 
(like: Theia-ide IA)

Davidyz:
Hi, thanks for the PR! I really appreciate this PR because it essentially allows me to keep both non-think and thinking versions of a hybrid model at the same time, without unloading one of them first, and then loading another. I've tried it with the OpenAI-compatible endpoint, and it worked exactly as I hoped.

One thing I'd like to suggest is to include this in the relevant documentation (for example, [here](https://github.com/ggml-org/llama.cpp/tree/master/tools/server#post-v1chatcompletions-openai-compatible-chat-completions-api)). This would clarify things for users and make the reviewers' lives easier.

rtpt-romankarwacik:
Using this for reasoning effort support with VSCode Copliot Chat, works nicely!

PR #20483, opened 2026-03-13
tests: Fix test-jinja-py Windows failures by bypassing command-line args

rillomas:
Current `test-jinja-py` execution on Windows show 12 failures which is caused by two issues:

1. JSON string is first converted to ANSI code page on Windows causing misinterpretation in python side.

```
  ASSERTION FAILED : Template render mismatch
    Template: "{{ data|tojson(ensure_ascii=true) }}"
    Expected: "\"\\u2713\""
    Python  : "\"\\u00e2\\u0153\\u201c\""
```
1. Expected output is assuming LF for new line where Windows defaults to CRLF.

```
  ASSERTION FAILED : Template render mismatch
    Template: "{% if true %}\nhello\n{% endif %}\n"
    Expected: "hello\n"
    Python  : "hello\r\n"

```

This PR fixes the first issue while #20365 addresses the second issue.

Full log with the 12 failures is the following:
[test-jinja-py_log_57819b8d.txt](https://github.com/user-attachments/files/25963521/test-jinja-py_log_57819b8d.txt)


CISC:
Please try #20365 instead.

rillomas:
Thanks @CISC. I tried #20365 and while the CRLF issue was fixed I still see the following ANSI code issue. Should I change this PR to just fix the ANSI code issue? FYI I'm seeing this when built with VS2026 18.4 where the test is running on Win 11 24H2 10.0.26100.8037 
```
  ASSERTION FAILED : Template render mismatch
    Template: "{{ data|tojson(ensure_ascii=true) }}"
    Expected: "\"\\u2713\""
    Python  : "\"\\u00e2\\u0153\\u201c\""
  tojson ensure_ascii=true (1 of 1 assertion(s) failed)                         [FAIL]
```

CISC:
> Thanks @CISC. I tried #20365 and while the CRLF issue was fixed I still see the following ANSI code issue. Should I change this PR to just fix the ANSI code issue?

Thanks, yes, please do. I think it could be fixed by piping the input instead of using command line arguments, but this will do fine (at least until we hit the command line length limit).

rillomas:
Thanks @CISC. I tried your stdin approach and seems to be working. I'll wait until https://github.com/ggml-org/llama.cpp/pull/20365 gets merged for final testing

CISC:
Ok, merged. :)

rillomas:
Confirmed `test-jinja-py` passing with [635b678](https://github.com/ggml-org/llama.cpp/pull/20483/commits/635b678af65af3a3af36343a1310ec9f7da85440) on Win 11 24H2 (10.0.26100.8037) built with VS2026 18.4

```
(venv) λ ctest -C Debug --output-on-failure -R test-jinja-py
Test project C:/Users/ae/Documents/mnakasak/repo/llama.cpp_mine/build-ci-debug
    Start 16: test-jinja-py
1/1 Test #16: test-jinja-py ....................   Passed   19.13 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
python    =  19.13 sec*proc (1 test)

Total Test time (real) =  19.15 sec
```

tests/test-jinja.cpp
```diff
@@ -1944,6 +1946,19 @@ static void test_template_py(testing & t, const std::string & name, const std::s
...
+        std::string input = merged.dump();
+        auto written = fwrite(input.c_str(), 1, input.size(), p_stdin);
+        if (written != input.size()) {
+            t.log("Failed to write complete input to subprocess stdin");
+            t.assert_true("subprocess stdin write", false);
+            subprocess_destroy(&subprocess);
+            return;
+        }
+        fflush(p_stdin);
+        fclose(p_stdin); // Close stdin to signal EOF to the Python process
```
CISC, line 1961: This causes a double free, need to find a clean way to do this.

rillomas: Thanks for catching that. I guess I'll manually set `subprocess->stdin_file` to null after the `fclose`

CISC: Since there does not seem to be any other way to do this in `sheredom/subprocess` I guess this is acceptable.

CISC (approved):

PR #20379, opened 2026-03-11
vulkan: fix SSM_CONV PP scaling with large ubatch sizes

ProgenyAlpha:
Fixes #18725

The SSM_CONV shader dispatched one token per Y workgroup, each doing only `nc` (typically 4) multiply-adds. At ubatch=2048 this meant 2048 workgroups in Y with almost no work per launch — workgroup dispatch overhead dominated.

**Changes:**
- Tile 16 tokens per workgroup using a 2D local size (32×16)
- Add `vec4` dot product fast path for the common `nc=4` (d_conv) case
- Pipeline wg_denoms updated from `{32,1,1}` to `{32,16,1}`

**45/45 SSM_CONV backend-ops tests passing.**

**test-backend-ops perf (ne_a=[515,3328], nc=4):**
| | us/run | GB/s |
|--|--------|------|
| master | 526.24 | 24.29 |
| this PR | 203.56 | 62.79 |
| speedup | **2.59x** | |

**Model bench (Qwen3-Coder-Next REAM Q4_K_M, pp2048, AMD 890M):**
| ubatch | master (t/s) | this PR (t/s) | change |
|--------|-------------|--------------|--------|
| 256 | 148.43 | 150.36 | +1.3% |
| 512 | 171.38 | 175.82 | +2.6% |
| 1024 | 147.46 | 181.14 | **+22.8%** |
| 2048 | 126.34 | 161.56 | **+27.9%** |

Master shows the #18725 pattern — PP drops from 171 at ub512 to 126 at ub2048. With this fix, PP peaks at ub1024 (181) and stays strong at ub2048 (162). The degradation cliff is gone.

Tested on AMD Radeon 890M (RDNA3.5, 8 CUs, Strix Point integrated). Would appreciate testing from @lemmi on the discrete 8060S where the original issue was reported.

PR #20495, opened 2026-03-13
vulkan: fix SSM_CONV crash on multi-GPU (#20462)

ProgenyAlpha:
Reverts the 2D workgroup tiling (32x16) from #20379 which causes `vk::DeviceLostError` on multi-GPU RADV setups. Keeps the vec4 dot product fast path for nc=4.

The 2D tiling reduced workgroup launch overhead at large ubatch sizes, but triggers a driver-level fault on multi-GPU configurations (confirmed on dual 7900 XTX, bisected to 40c550d).

Based on master at 983df14.

Fixes #20462

## Test plan
- [ ] @itterative: confirm crash reproduces on master (983df14) with multi-GPU
- [ ] @itterative: confirm this branch fixes the crash
- [ ] Single-GPU SSM_CONV test-backend-ops pass (verified, 45/45)

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -4607,7 +4607,7 @@ static void ggml_vk_load_shaders(vk_device& device) {
         ggml_vk_create_pipeline(device, device->pipeline_ssm_scan_f32_d256, "ssm_scan_256_f32", ssm_scan_f32_len, ssm_scan_f32_data, "main", 8, sizeof(vk_op_ssm_scan_push_constants), {1, 1, 1}, {256, device->subgroup_size, 16}, 1, true, true);
     }
 
-    ggml_vk_create_pipeline(device, device->pipeline_ssm_conv_f32, "ssm_conv_f32", ssm_conv_f32_len, ssm_conv_f32_data, "main", 3, sizeof(vk_op_ssm_conv_push_constants), {32, 16, 1}, {32, 16}, 1);
+    ggml_vk_create_pipeline(device, device->pipeline_ssm_conv_f32, "ssm_conv_f32", ssm_conv_f32_len, ssm_conv_f32_data, "main", 3, sizeof(vk_op_ssm_conv_push_constants), {32, 1, 1}, {32}, 1);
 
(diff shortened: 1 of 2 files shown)


itterative:
Tested against the pr, and it seems to be failing (same as master). Master fails at the same point as the second log file (vulkan-issue-0a2088a-2.txt).

```
$ llama-cli --version
ggml_vulkan: 0 = Radeon RX 7900 XTX (RADV NAVI31) (radv) | uma: 0 | fp16: 1 | bf16: 0 | warp size: 64 | shared memory: 65536 | int dot: 0 | matrix cores: KHR_coopmat
ggml_vulkan: 1 = Radeon RX 7900 XTX (RADV NAVI31) (radv) | uma: 0 | fp16: 1 | bf16: 0 | warp size: 64 | shared memory: 65536 | int dot: 0 | matrix cores: KHR_coopmat
version: 8325 (0a2088a8c)
```


Ran it twice after setting the temperature to 0. Here are the logs:
[vulkan-issue-0a2088a.txt](https://github.com/user-attachments/files/25968406/vulkan-issue-0a2088a.txt)
[vulkan-issue-0a2088a-2.txt](https://github.com/user-attachments/files/25968410/vulkan-issue-0a2088a-2.txt)

Some other notes:
* I tried to run llama-bench with context size 4100 (basically, 4 trailing tokens) and neither this pr, nor master fail
* there's a performance regression for tg128 (master is 90.70 t/s and this pr is 76.19 t/s) and I've seen this when running llama-server as well (pp4100 was approximately the same)

@ProgenyAlpha is there a way I could give you better stack traces for the failing vulkan shaders?

ProgenyAlpha:
Closing, the revert didn't fix it. @itterative if you want to keep debugging this, the Vulkan debug output might help narrow down which op is faulting. You can enable it by adding `target_compile_definitions(ggml-vulkan PRIVATE GGML_VULKAN_DEBUG)` to `ggml/src/ggml-vulkan/CMakeLists.txt` and rebuilding.

ProgenyAlpha:
Correction: easier way to enable it is just the cmake flag:

```
cmake -B build -DGGML_VULKAN=ON -DGGML_VULKAN_DEBUG=ON
```

No need to edit CMakeLists.txt.

PR #20508, opened 2026-03-13
mtmd: add llama-mtmd-debug binary

ngxson:
Add `llama-mtmd-debug` binary that facilitate debugging mtmd

It allows running the encode and preprocessing pass separately, with a pre-defined input that is reproducible on both cpp and pytorch code. Included patterns are: black (0 filled), white (1 filled), gray (0.5 filled) and checkerboard.

This is the initial version and it's intended to be improved progressively


ngxson:
May you have a look too (if you're interested) @pwilkin @ddh0

ggerganov (approved):

pwilkin:
Looks cool! One thing I'm wondering - maybe add to the examples a pre-rendered image of *something*? Can be simple, like a big red apple on a yellow background, but that would probably help during debugging as well.

ngxson:
@pwilkin yes we can add more patterns in the future, especially asymmetrical and colorful patterns (useful for testing positional embeddings). Meaningful images may not make because we are just running half the pipeline, so output is expected to be mostly non-sense to the language model.

Also, one thing useful to have is to implement a pipeline that can replicate exactly the same input on pytorch (exactly same tensor values). I included a demo for solid color in this PR, but currently missing checkerboard. Feel free to add more patterns if you can think of any.

PR #20513, opened 2026-03-13
server: reset counter related to kill-switch on client error

SoftwareRenderer:
This avoids inadvertently triggering a server kill switch.

If the client sends a request that exceeds the configured context size, an appropriate HTTP 400 response is provided and no tokens are generated.

However since no tokens are generated, update_slots() increments n_empty_consecutive. If the client sends 4 such messages in a row, the server terminates.

This change resets the counter when the client's request exceeds `--ctx-size`. While I was looking at this, neighboring code had similar patterns relating to the same `ERROR_TYPE_EXCEED_CONTEXT_SIZE`, however I'm not familiar with those other conditions.

Steps to reproduce issue:
1. Start llama-server with a tiny context e.g. `--ctx-size 10`
2. Send text that exceeds the context size 4 times in a row

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -1190,6 +1190,7 @@ private:
             : SLOT_STATE_STARTED;
 
         SLT_INF(slot, "processing task, is_child = %d\n", slot.task->is_child());
+        n_empty_consecutive = 0; // Reset server kill-switch counter
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


ggerganov (commented):
It would be better to move this to `launch_slot_with_task()` - right before returning `true`.

SoftwareRenderer:
> It would be better to move this to `launch_slot_with_task()` - right before returning `true`.

Done. I was worried that it might interfere with the debugging efforts in https://github.com/ggml-org/llama.cpp/pull/20277 . 

After looking at that issue closer, if I'm understanding it correctly: the loop occurs within the same task and slot. So resetting the counter won't impact that because the reset occurs before a new task/slot.

ggerganov (approved):

Issue #20029, opened 2026-03-02
Misc. bug: b8143 produces garbage Mac x86 Vulkan with AMD GPU


deniskokarev:
### Name and Version

I have a Mac x86 with Radeon 6900XT RDNA2 GPU with self-built `llamacpp` with Vulkan support. Not super fast, but usable since, at least from b6431, when things broke recently from 8143

8142 works and `mistral-11b-omnimix-bf16.Q8_0.gguf` produces sensible output

### Operating systems

Mac

### Which llama.cpp modules do you know to be affected?

llama-server, llama-cli

### Command line

```shell
cli or server producing garbage

% ./build/bin/llama-cli --ctx-size 8192 -m ~/shared/models/mistral-11b-omnimix-bf16.Q8_0.gguf -dev Vulkan0 -p "what is population of Russian capital"
ggml_metal_device_init: tensor API disabled for pre-M5 and pre-A19 devices
ggml_metal_library_init: using embedded metal library
ggml_metal_library_init: loaded in 0.028 sec
ggml_metal_rsets_init: creating a residency set collection (keep_alive = 180 s)
ggml_metal_device_init: GPU name:   MTL0
ggml_metal_device_init: GPU family: MTLGPUFamilyCommon3 (3003)
ggml_metal_device_init: GPU family: MTLGPUFamilyMetal3  (5001)
ggml_metal_device_init: simdgroup reduction   = true
ggml_metal_device_init: simdgroup matrix mul. = false
ggml_metal_device_init: has unified memory    = false
ggml_metal_device_init: has bfloat            = true
ggml_metal_device_init: has tensor            = false
ggml_metal_device_init: use residency sets    = true
ggml_metal_device_init: use shared buffers    = false
ggml_metal_device_init: recommendedMaxWorkingSetSize  = 17163.09 MB
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = AMD Radeon RX 6900 XT (MoltenVK) | uma: 0 | fp16: 1 | bf16: 0 | warp size: 64 | shared memory: 65536 | int dot: 0 | matrix cores: none

Loading model...  


▄▄ ▄▄
██ ██
██ ██  ▀▀█▄ ███▄███▄  ▀▀█▄    ▄████ ████▄ ████▄
██ ██ ▄█▀██ ██ ██ ██ ▄█▀██    ██    ██ ██ ██ ██
[...]

PR #20518, opened 2026-03-13
vulkan: async and event fixes

0cc4m:
I noticed incoherence with my multi-GPU setup as well when investigating issues like #20462. I found that they can be fixed by disabling `cpy_tensor_async`, so the problem is with the async path. I narrowed it down to these problems:
- events were set, but the wait command was never submitted to the queue, so the `event_wait` function didn't do anything
- events were resetting command buffers that had long since been reused, because they didn't track that. This was causing validation errors and perhaps driver issues/crashes
- there was a race condition between an event being set, being waited on within the GPU and the event being reset by the host. This only didn't lead to deadlocks because `event_wait` was not working. I fixed it by doing the event reset in the queue instead of outside of it, that way it happens immediately before the next set. command. To avoid the same issue with the fence, I replaced it with a timeline semaphore, which does not need manual resets by the host.

This may help with #20462, #20029, #20517 and maybe some more.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -6513,14 +6513,15 @@ static void ggml_vk_ctx_begin(vk_device& device, vk_context& subctx) {
 }
 
 static vk_context ggml_vk_get_compute_ctx(ggml_backend_vk_context * ctx) {
+    vk_context result;
     if (!ctx->compute_ctx.expired()) {
-        return ctx->compute_ctx.lock();
+        result = ctx->compute_ctx.lock();
+    } else {
+        result = ggml_vk_create_context(ctx, ctx->compute_cmd_pool);
+
+        ctx->compute_ctx = result;
+        ggml_vk_ctx_begin(ctx->device, result);
     }
-
-    vk_context result = ggml_vk_create_context(ctx, ctx->compute_cmd_pool);
-
-    ctx->compute_ctx = result;
-    ggml_vk_ctx_begin(ctx->device, result);
 
     if (ctx->device->async_use_transfer_queue && ctx->transfer_semaphore_last_submitted < ctx->transfer_semaphore.value) {
         result->s->wait_semaphores.push_back(ctx->transfer_semaphore);
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -14908,11 +14908,13 @@ static void ggml_backend_vk_event_wait(ggml_backend_t backend, ggml_backend_even
     VK_LOG_DEBUG("ggml_backend_vk_event_wait(backend=" << backend << ", event=" << event << ")");
     ggml_backend_vk_context * ctx = (ggml_backend_vk_context *)backend->context;
     vk_event *vkev = (vk_event *)event->context;
-    GGML_ASSERT(vkev->has_event);
 
     vk_context compute_ctx = ggml_vk_get_compute_ctx(ctx);
 
-    ggml_vk_wait_events(compute_ctx, { vkev->event });
+    if (vkev->has_event) {
+        // Wait for latest event
+        ggml_vk_wait_events(compute_ctx, { vkev->event });
+    }
 }
 
 // TODO: enable async and synchronize
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -14908,13 +14908,11 @@ static void ggml_backend_vk_event_wait(ggml_backend_t backend, ggml_backend_even
     VK_LOG_DEBUG("ggml_backend_vk_event_wait(backend=" << backend << ", event=" << event << ")");
     ggml_backend_vk_context * ctx = (ggml_backend_vk_context *)backend->context;
     vk_event *vkev = (vk_event *)event->context;
+    GGML_ASSERT(vkev->has_event);
 
     vk_context compute_ctx = ggml_vk_get_compute_ctx(ctx);
 
-    if (vkev->has_event) {
-        // Wait for latest event
-        ggml_vk_wait_events(compute_ctx, { vkev->event });
-    }
+    ggml_vk_wait_events(compute_ctx, { vkev->event });
 }
 
 // TODO: enable async and synchronize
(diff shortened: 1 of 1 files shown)


0cc4m:
I'm sure about the event reset asynchronously before setting it. It does seem to work, but maybe we need multiple events instead, to avoid reuse until a full queue synchronization happens.

jeffbolznv:
I don't think the reset right before the set will be safe, this was one of the things that made vkevents very hard to use.

0cc4m:
Probably not, yeah. I'll take another look tomorrow and figure out something better, I guess we need multiple events. I thought about using the timeline semaphore for synchronization within the queue as well, but I think that would be heavier than events and isn't really what they are meant for.

0cc4m:
@jeffbolznv I got something that should work, now. I'm not sure if it could be done in a simpler way, but this was the best I found that still actually reuses without a manual synchronization step anywhere. Without reuse we might get into trouble with too many events during loading, similar to command buffers.

jeffbolznv:
The recycling of events seems to assume that they will only be waited on once. Is that a valid assumption?

0cc4m:
No, they do get waited on multiple times. But the assumption is that after a new event is recorded, the previous one cannot be waited on any longer. So when the new event gets synchronized (cpu-waited), all event waits of the previous one are also done, because they were submitted into the queue before the new event got set.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -14858,18 +14877,34 @@ static void ggml_backend_vk_event_record(ggml_backend_t backend, ggml_backend_ev
...
-    // the backend interface doesn't have an explicit reset, so reset it here
-    // before we record the command to set it
-    ctx->device->device.resetEvent(vkev->event);
-    ctx->device->device.resetFences({ vkev->fence });
+    if (vkev->has_event) {
+        // Move pending to submitted
+        vkev->events_submitted.insert(vkev->events_submitted.end(), vkev->events_pending.begin(), vkev->events_pending.end());
+        vkev->events_pending.clear();
+        // Move existing event into pending
+        vkev->events_pending.push_back(vkev->event);
```
jeffbolznv: I'm not fully following the logic here. The rest of this function will get an event and submit it. Why doesn't that immediately go into events_submitted? And can there ever be more than one pending event?

0cc4m: Pending and submitted are meant in relation to wait commands in the queue, not set commands. So the flow is like this, for example:
```
Event 1 recorded
Waits for event 1
Event 2 recorded, event 1 goes into pending. It is no longer possible to wait for event 1
Waits for event 2
Event 3 recorded, event 2 goes into pending. Event 1 goes into submitted.
Waits for event 3
Synchronize event 3, event 2 and 1 can now be reused because all waits were before event 3.
```
or
```
Event 1 recorded
Waits for event 1
Synchronize event 1, wait commands for event 1 may still be in the queue, so no reuse yet
Event 2 recorded, event 1 moves into pending
Waits for event 2
Synchronize event 2, event 1 can now be reused
```
But I think you're right. The pending stage isn't needed. A synchronization means the queue has reached the set command of the event being synchronized, so all waits for previous events must also be done.

HumerousGorgon:
Hi @0cc4m,

I can confirm that this PR definitely fixes the problems I'm having with inference on Intel GPUS.
Thank you for this!

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -14880,9 +14911,10 @@ static void ggml_backend_vk_event_wait(ggml_backend_t backend, ggml_backend_even
 
     vk_context compute_ctx = ggml_vk_get_compute_ctx(ctx);
 
-    ggml_vk_wait_events(compute_ctx, {vkev->event});
-    ggml_vk_ctx_end(compute_ctx);
-    ctx->compute_ctx.reset();
+    if (vkev->has_event) {
```
jeffbolznv: Could this be an assert?

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -14880,9 +14911,10 @@ static void ggml_backend_vk_event_wait(ggml_backend_t backend, ggml_backend_even
...
+        // Wait for latest event
+        ggml_vk_wait_events(compute_ctx, { vkev->event });
```
jeffbolznv: Is it intentional not to call ggml_vk_ctx_end?

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -15689,8 +15724,16 @@ static void ggml_backend_vk_device_event_free(ggml_backend_dev_t dev, ggml_backe
 
     vk_event *vkev = (vk_event *)event->context;
 
-    device->device.destroyFence(vkev->fence);
-    device->device.destroyEvent(vkev->event);
+    device->device.destroySemaphore(vkev->tl_semaphore.s);
```
jeffbolznv, line 15727: Do you need to do anything with vkev->cmd_buffer in here?

ggml/src/ggml-vulkan/ggml-vulkan.cpp
0cc4m: It could be. Do you think crashing on unrecorded event is better than doing nothing? If this is considered invalid use it might be.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
0cc4m: Yes, I just want to record the event into the current command buffer, not actually submit it. I assume if a wait command happens a submission of the work that depends on this wait is not far off.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
0cc4m: No, the command buffer is just a "weak" reference, not owned by the event. It just exists to be able to mark it for reuse if the event is synchronized.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
jeffbolznv: It seems invalid, so I think getting the assert message is preferable to silently skipping it.

0cc4m: Alright, I changed it.

jeffbolznv (approved):

sinister-cat:
Tried merging this to master locally, everything builds fine but the assert throws during warmup the phase with qwen3.5? I was able to use the exact same command on master with no immediate issues.

[log.txt](https://github.com/user-attachments/files/26040771/log.txt)

Commenting the assert causes a segfault, --no-warmup also causes segfault... any ideas on why this could be?

0cc4m:
@jeffbolznv I guess that answers the "valid use" question. I'll revert the assertion.

jeffbolznv (approved):

PR #20521, opened 2026-03-13
ci: try to optimize some jobs

netrunnereve:
I tried to switch some jobs to arm or ubuntu-slim as per my comment in #20446 for builds where it really doesn't matter. Most jobs didn't fit in the 15 minute ubuntu-slim time limit and some like the sanitizer or android straight up failed on arm. If a job doesn't have ccache set up I also made it work on both x86 and arm so it would pick the first available machine.

I'm not sure how much this really helps, but it does reduce the number of x86 machines that we're using at any given time.

run in my fork with those jobs forced to run on arm: https://github.com/netrunnereve/llama.cpp/actions/runs/23031702820


.github/workflows/build.yml
```diff
@@ -381,7 +375,7 @@ jobs:
           ctest -L main --verbose
 
   ubuntu-24-cmake-vulkan-deb:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-24.04-arm
```
CISC: Why is this and wasm not `||` like the others?

CISC (approved):
Worth a shot, please remove `ccache` from vulkan and wasm builds.

Did you test the `release` workflow?

.github/workflows/build.yml
netrunnereve: This is because it has ccache set up and it's going to mess up the cache if it sometimes runs on x86 and sometimes runs on arm. Since you want me to remove the ccache from those builds anyways then it's fine for it to run on both.

netrunnereve:
> Did you test the release workflow?

I thought I did but turns out that run got stuck. Anyways here's [a rerun with a limited number of jobs](https://github.com/netrunnereve/llama.cpp/actions/runs/23077670685), its working fine. And here's the [server webui run](https://github.com/netrunnereve/llama.cpp/actions/runs/23031932445) as well.

ggerganov (approved):

PR #20525, opened 2026-03-13
ggml-cuda: native bf16 flash attention for vec kernel

eous:
mma kernel still converts bf16 to fp16 before launch, native mma bf16 todo. Only about 1.2% improvement on sm120.


xkmire:
I tested this PR on my RTX 6000 PRO blackwell. It works great on Qwen 3.5 122B that uses delta attention and I have flash attention turned on.
Even if some of calculations looks to be done in 32-bit precision in 
ggml_cuda_mad(sum, __bfloat1622float2(tmp[k_KQ_1]), ((const float2 *) Q_v)[k_KQ_0/nthreads + k_KQ_1]);
the performance in tokens/second is the same as using f16 on the KV cache (even if there probably are some small difference if measured in detail).
I confirmed that the compiler, for 120a-real architecture uses the 32-bit path.

HUGE improvement vs. the CPU fallback that happened before.

I hope this will go in soon... 
Thanks! Made my day.

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -351,10 +351,24 @@ static __device__ __forceinline__ void flash_attn_tile_load_tile(
...
+                        ggml_cuda_memcpy_1<cpy_nb>(
+                            tmp, !oob_check || i < i_sup ? KV + i*stride_KV + j : zero);
+                        __align__(16) half2 converted[cpy_ne];
+#pragma unroll
+                        for (int l = 0; l < cpy_ne; ++l) {
+                            const float2 f = __bfloat1622float2(tmp[l]);
+                            converted[l] = make_half2(f.x, f.y);
+                        }
+                        ggml_cuda_memcpy_1<sizeof(converted)>(tile_KV + i*(J/2 + J_padding) + j, converted);
+                    }
```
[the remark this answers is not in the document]
eous: This one might be worth doing?

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -351,10 +351,24 @@ static __device__ __forceinline__ void flash_attn_tile_load_tile(
...
+                        __align__(16) T_KV tmp[cpy_ne];
+                        ggml_cuda_memcpy_1<cpy_nb>(
+                            tmp, !oob_check || i < i_sup ? KV + i*stride_KV + j : zero);
+                        __align__(16) half2 converted[cpy_ne];
+#pragma unroll
+                        for (int l = 0; l < cpy_ne; ++l) {
+                            const float2 f = __bfloat1622float2(tmp[l]);
+                            converted[l] = make_half2(f.x, f.y);
+                        }
+                        ggml_cuda_memcpy_1<sizeof(converted)>(tile_KV + i*(J/2 + J_padding) + j, converted);
```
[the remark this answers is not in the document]
eous: Pretty sure the compiler is just doing the right thing here, but I guess maybe?

ggml/src/ggml-cuda/CMakeLists.txt
```diff
@@ -116,11 +116,13 @@ if (CUDAToolkit_FOUND)
...
-        file(GLOB   SRCS "template-instances/fattn-vec*q4_0-q4_0.cu")
+        file(GLOB   SRCS "template-instances/fattn-vec-instance-q4_0-q4_0.cu")
         list(APPEND GGML_SOURCES_CUDA ${SRCS})
-        file(GLOB   SRCS "template-instances/fattn-vec*q8_0-q8_0.cu")
+        file(GLOB   SRCS "template-instances/fattn-vec-instance-q8_0-q8_0.cu")
         list(APPEND GGML_SOURCES_CUDA ${SRCS})
-        file(GLOB   SRCS "template-instances/fattn-vec*f16-f16.cu")
+        file(GLOB   SRCS "template-instances/fattn-vec-instance-f16-f16.cu")
+        list(APPEND GGML_SOURCES_CUDA ${SRCS})
+        file(GLOB   SRCS "template-instances/fattn-vec-instance-bf16-bf16.cu")
```
JohannesGaessler: If you remove the glob pattern, just add the strings explicitly to `SRCS` in one go.

ggml/src/ggml-cuda/fattn-common.cuh
```diff
@@ -74,6 +74,37 @@ static __device__ __forceinline__ float vec_dot_fattn_vec_KQ_f16(
...
+        __align__(16) nv_bfloat162 tmp[cpy_ne];
+        ggml_cuda_memcpy_1<sizeof(tmp)>(tmp, K_bf16 + k_KQ_0 + (threadIdx.x % nthreads)*cpy_ne);
+#pragma unroll
+        for (int k_KQ_1 = 0; k_KQ_1 < cpy_ne; ++k_KQ_1) {
+#ifdef V_DOT2_F32_F16_AVAILABLE
+            const float2 bf16_f2 = __bfloat1622float2(tmp[k_KQ_1]);
+            ggml_cuda_mad(sum, make_half2(bf16_f2.x, bf16_f2.y), ((const half2 *) Q_v)[k_KQ_0/nthreads + k_KQ_1]);
+#else
+            ggml_cuda_mad(sum, __bfloat1622float2(tmp[k_KQ_1]), ((const float2 *) Q_v)[k_KQ_0/nthreads + k_KQ_1]);
+#endif // V_DOT2_F32_F16_AVAILABLE
```
JohannesGaessler, line 101: Remove the `V_DOT2_F32_F16_AVAILABLE` path, that instruction does not have a BF16 equivalent. Also use `ggml_cuda_cast` defined in `convert.cuh` for the conversion to `float2` (this is also what the half2 path on master should have used).

ggml/src/ggml-cuda/fattn-common.cuh
```diff
@@ -321,6 +352,32 @@ static __device__ __forceinline__ void dequantize_V_f16(const void * __restrict_
...
+    if constexpr (std::is_same_v<T, half>) {
+        static_assert(ne % 2 == 0, "bad ne");
+        __align__(16) nv_bfloat162 tmp[ne/2];
+        ggml_cuda_memcpy_1<ne*sizeof(nv_bfloat16)>(tmp, (const nv_bfloat16 *) vx + i0);
+        half2 * dst_h2 = (half2 *) dst;
+#pragma unroll
+        for (int l = 0; l < ne/2; ++l) {
+            const float2 f2 = __bfloat1622float2(tmp[l]);
+            dst_h2[l] = make_half2(f2.x, f2.y);
+        }
```
JohannesGaessler: Remove this code path, it doesn't make sense to use.

ggml/src/ggml-cuda/fattn-common.cuh
```diff
@@ -321,6 +352,32 @@ static __device__ __forceinline__ void dequantize_V_f16(const void * __restrict_
...
+    } else if constexpr (std::is_same_v<T, float>) {
+        static_assert(ne % 2 == 0, "bad ne");
+        __align__(16) nv_bfloat162 tmp[ne/2];
+        ggml_cuda_memcpy_1<ne*sizeof(nv_bfloat16)>(tmp, (const nv_bfloat16 *) vx + i0);
+        float2 * dst_f2 = (float2 *) dst;
+#pragma unroll
+        for (int l = 0; l < ne/2; ++l) {
+            dst_f2[l] = __bfloat1622float2(tmp[l]);
```
JohannesGaessler:
```suggestion
            dst_f2[l] = ggml_cuda_cast<float2>(tmp[l]);
```

ggml/src/ggml-cuda/fattn-vec.cuh
```diff
@@ -556,31 +558,35 @@ void ggml_cuda_flash_attn_ext_vec_case(ggml_backend_cuda_context & ctx, ggml_ten
     template void ggml_cuda_flash_attn_ext_vec_case                         \
     <D, type_K, type_V>(ggml_backend_cuda_context & ctx, ggml_tensor * dst) \
 
-#define EXTERN_DECL_FATTN_VEC_CASES(D, type_K)             \
+#define EXTERN_DECL_FATTN_VEC_CASES(D, type_K)              \
```
JohannesGaessler:
```suggestion
#define EXTERN_DECL_FATTN_VEC_CASES(D, type_K)             \
```

ggml/src/ggml-cuda/fattn-common.cuh
```diff
@@ -781,7 +842,8 @@ static __global__ void flash_attn_combine_results(
 template <int DV, int ncols1, int ncols2>
 void launch_fattn(
     ggml_backend_cuda_context & ctx, ggml_tensor * dst, fattn_kernel_t fattn_kernel, const int nwarps, const size_t nbytes_shared,
-    const int nbatch_fa, const bool need_f16_K, const bool need_f16_V, const bool stream_k, const int warp_size = WARP_SIZE
+    const int nbatch_fa, const bool need_f16_K, const bool need_f16_V, const bool stream_k, const int warp_size = WARP_SIZE,
+    const bool need_bf16_K = false, const bool need_bf16_V = false
```
JohannesGaessler: Revert the changes to `launch_fattn`. The quantized data formats use `half` scales so there is no point in converting them to BF16 over FP16.

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -321,9 +321,9 @@ static constexpr __device__ int ggml_cuda_fattn_tile_get_nbatch_K(const int DKQ,
 }
 
 // TODO: deduplicate with mma-f16
-template<int warp_size, int nwarps, int I, int J, int J_padding, bool oob_check>
+template<int warp_size, int nwarps, int I, int J, int J_padding, bool oob_check, typename T_KV>
 static __device__ __forceinline__ void flash_attn_tile_load_tile(
-        const half2 * const __restrict__ KV, half2 * const __restrict__ tile_KV, const int stride_KV, const int i_sup) {
+        const T_KV * const __restrict__ KV, half2 * const __restrict__ tile_KV, const int stride_KV, const int i_sup) {
```
JohannesGaessler: Revert the changes to this function. It doesn't make sense to convert BF16 to FP16 so just always use the FP32 path for BF16.

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -401,15 +415,19 @@ static __device__ __forceinline__ void flash_attn_tile_load_tile(
...
 
                     __align__(16) float2 tmp_f2[cpy_ne/2];
 #pragma unroll
                     for (int l = 0; l < cpy_ne/2; ++l) {
-                        tmp_f2[l] = __half22float2(tmp_h2[l]);
+                        if constexpr (std::is_same_v<T_KV, half2>) {
+                            tmp_f2[l] = __half22float2(tmp_kv[l]);
+                        } else {
+                            tmp_f2[l] = __bfloat1622float2(tmp_kv[l]);
+                        }
```
JohannesGaessler:
```suggestion
                        tmp_f2[l] = ggml_cuda_cast<float2>(tmp_kv[l]);
                        
```

Add a new case to `ggml_cuda_cast` if necessary.

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -735,7 +753,7 @@ static __device__ __forceinline__ void flash_attn_tile_iter(
     }
 }
 
-template<int DKQ, int DV, int ncols1, int ncols2, bool use_logit_softcap> // D == head size
+template<int DKQ, int DV, int ncols1, int ncols2, bool use_logit_softcap, typename T_KV = half2> // D == head size
```
JohannesGaessler:
```suggestion
template<int DKQ, int DV, int ncols1, int ncols2, bool use_logit_softcap, typename T_KV> // D == head size
```

I think the type should be explicitly required to avoid accidental misuse.

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -428,10 +446,10 @@ static __device__ __forceinline__ void flash_attn_tile_load_tile(
 
 // Function that performs a single iteration in for the KQ matrix multiplication:
 template <int warp_size, int nwarps, int ncols1, int ncols2, int DKQ, int nbatch_fa, int nbatch_K,
-    bool use_logit_softcap, bool oob_check, typename T_vec_dot>
+    bool use_logit_softcap, bool oob_check, typename T_vec_dot, typename T_KV>
 static __device__ __forceinline__ void flash_attn_tile_iter_KQ(
         T_vec_dot   * const Q_tmp,
-        const half2 * const __restrict__ K_h2,
+        const T_KV  * const __restrict__ K_kv,
```
JohannesGaessler:
```suggestion
        const T_KV  * const __restrict__ K_t2,
```

I think `_kv` is a poor choice for the suffix name. I don't particularly like `_t2` either but I think it's less bad.

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -1087,6 +1105,24 @@ static __global__ void flash_attn_tile(
 #endif // FLASH_ATTN_AVAILABLE
 }
 
+template <int DKQ, int DV, int ncols1, int ncols2, bool use_logit_softcap>
+static void launch_fattn_tile_kernel(
+        ggml_backend_cuda_context & ctx, ggml_tensor * dst,
+        const int nwarps, const size_t nbytes_shared, const int nbatch_fa, const int warp_size) {
+    const ggml_tensor * K = dst->src[1];
+    const bool bf16_kv = K->type == GGML_TYPE_BF16;
```
JohannesGaessler:
```suggestion
    const ggml_tensor * K = dst->src[1];
    const ggml_tensor * V = dst->src[2];
    GGML_ASSERT(K->type == V->type);
    const bool bf16_kv = K->type == GGML_TYPE_BF16;
```

eous:
addressed all review feedback

- removed V_DOT2 path from bf16 vec_dot, bf16 always goes through float
- switched to ggml_cuda_cast for all bf16 conversions instead of raw intrinsics
- removed the half dequantize path for bf16 V, float only with static_assert
- reverted all launch_fattn changes, no bf16 conversion params needed
- reverted all tile kernel changes, the proper fix needs an if constexpr refactor of the fp16/fp32 path selection which is a bigger change than makes sense here. tile and mma fall back to bf16 to fp16 pre conversion for now
- cmake uses explicit filenames in one list() call instead of globs
- added ggml_cuda_cast(nv_bfloat162) overload to convert.cuh
- fixed trailing whitespace in EXTERN_DECL macro

benchmarked on RTX PRO 6000 Blackwell with qwen 3.5 27B BF16 up to 65k context, no regressions

JohannesGaessler (approved):

eous:
Working through the V_DOT2 path, I am starting to appreciate the layout of the cuda kernels and their cross arch compatibility. This CI setup is great.

ggml/src/ggml-cuda/fattn-common.cuh
eous: So I know you said to remove the V_DOT2 path from vec_dot_bf16 since there is no bf16 equivalent of the v_dot2 instruction. agreed, but there is a downstream issue that kind of forcing us to add it back

On V_DOT2 platforms Q_reg is declared as half2 (fattn-vec.cuh line 133, selected by ifdef V_DOT2_F32_F16_AVAILABLE). the bf16 vec_dot casts Q_v as float2 which would reinterpret those half2 values as float2, producing garbage. the V_DOT2 path in vec_dot_bf16 converts the half2 Q values to float2 before the dot product so both operands are correctly typed

Removing the V_DOT2 path entirely would require also changing how Q_reg is stored for bf16 K on V_DOT2 platforms, which means adding if constexpr guards around the Q_reg declaration and Q loading loops. happy to do that if preferred, but it seemed like a bigger change than warranted

xkmire:
Is there something specific hindering this to go in before it gets old?
It is important to have when you run Qwen 3.5 with deltanet attention on GPU.
Anything I can help with?

am17an:
@JohannesGaessler merge?

JohannesGaessler (approved):
There were AMD-specific changes changes after I had approved so I wanted to re-review. The reason you would want to use BF16 vs. FP16 is the increased numerical range. So the AMD workaround where BF16 is cast to FP16 makes the data type essentially universally worse than FP16. I think the vector kernel needs to be refactored though I cannot expect someone without access to AMD hardware to do this - I added a comment to the code indicating that this needs to be fixed (by me).

JohannesGaessler:
@am17an my approval has become stale since after pushing the comments I am now the last pusher. Can you approve instead?

am17an (approved):

CISC:
Looks like this warrants an extra guard:
https://github.com/ggml-org/llama.cpp/actions/runs/23400742598/job/68071505499#step:3:391

Orcazephyr:
@JohannesGaessler 

I'm very interested in the optimizations to amd which would let hip use the new bfloat16 path without any casts to float16. I'd be willing to test on my 7900xtx if you still plan to implement those paths and you don't have access to amd hardware. Thanks! Otherwise looking forward to that.

PR #20527, opened 2026-03-13
hexagon: Q4_0 and MXFP4 repack fixes

max-krasnyansky:
Turns out our repack logic has bug where tensors with row sizes not multiple of 256 are getting corrupted.
Basically, I made the wrong assumption that we can use `0:128,1:129,... INT4` element packing for all blocks of 256
This was causing the scales to partially override some of the tail quants (in Hexagon backend we repack the rows into all-quants followed by all-scales format). 

This bug was hard to spot because, surprisingly, the output of GPT-OSS-20B (one of the first models to use row size 2880 which is not a multiple of 256) looks good, and the `test-backend-ops` does not catch this error because the CPU backend ends up running on the same corrupted tensor data (i.e the _unpack_ in `hexagon_get_tensor()` is correct but it unpacks already corrupted data).

To fix this issue we're now using `0:1,2:3,... INT4` packing for the last partial block.
This change causes a slight perf regression for the models with row sizes not multiple of 256. 
Performance has slightly improved for other models because of the overall matmul improvements.

### DIFF
--- a/ggml/src/ggml-hexagon/htp/matmul-ops.c
+++ b/ggml/src/ggml-hexagon/htp/matmul-ops.c
@@ -317,8 +317,6 @@ static inline HVX_Vector hvx_vec_rmpy_x8_full(HVX_Vector_x8 x, HVX_Vector_x8 y) 
 }
 
 static inline HVX_Vector hvx_vec_rmpy_x8_partial(HVX_Vector_x8 x, HVX_Vector_x8 y, unsigned int n) {
-    return hvx_vec_rmpy_x8_full(x, y);
-
     if (n >= 512)
         return hvx_vec_rmpy_x8_full(x, y);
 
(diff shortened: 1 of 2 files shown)


ggml/src/ggml-hexagon/htp/matmul-ops.c
```diff
@@ -213,15 +285,44 @@ static inline HVX_Vector hvx_vec_rmpy_x8_n(HVX_Vector_x8 x, HVX_Vector_x8 y, uns
...
 }
 
-// Handle most common cases of tensors not multiple of 1024.
-static inline HVX_Vector hvx_vec_rmpy_x8_nloe(HVX_Vector_x8 x, HVX_Vector_x8 y, unsigned int n) {
-    if (n <= 256) { return hvx_vec_rmpy_x8_n(x, y, 256); };
-    if (n <= 512) { return hvx_vec_rmpy_x8_n(x, y, 512); };
-    if (n <= 768) { return hvx_vec_rmpy_x8_n(x, y, 768); };
-    return hvx_vec_rmpy_x8_n(x, y, 1024);
+static inline HVX_Vector hvx_vec_rmpy_x8_partial(HVX_Vector_x8 x, HVX_Vector_x8 y, unsigned int n) {
+    return hvx_vec_rmpy_x8_full(x, y);
```
lhez: Should this line be removed? If it returns here, lines 322 - 325 won't be executed.

lhez (approved):

ggml/src/ggml-hexagon/htp/matmul-ops.c
max-krasnyansky: Oops. Thanks for catching that. Leftover from experiments. Fixing and merging.

PR #20530, opened 2026-03-13
devops: force Python 3.13 in Vulkan container

geraname:
Forces Python 3.13 in the Vulkan container image build to support PyTorch 2.6.

Fixes #20524.

### DIFF
--- a/.devops/vulkan.Dockerfile
+++ b/.devops/vulkan.Dockerfile
@@ -58,7 +58,6 @@ RUN apt-get update \
     python3-pip \
     python3-wheel \
     && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.13 100 \
-    && update-alternatives --set python3 /usr/bin/python3.13 \
     && pip install --break-system-packages --upgrade setuptools \
     && pip install --break-system-packages -r requirements.txt \
     && apt autoremove -y \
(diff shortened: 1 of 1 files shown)


.devops/vulkan.Dockerfile
```diff
@@ -53,10 +53,12 @@ RUN apt-get update \
...
     build-essential \
     git \
-    python3 \
-    python3-dev \
+    python3.13 \
+    python3.13-dev \
     python3-pip \
     python3-wheel \
+    && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.13 100 \
+    && update-alternatives --set python3 /usr/bin/python3.13 \
```
taronaeo:
```patch
+    && update-alternatives --set python3 /usr/bin/python3.13 \
```

I'm not sure if this is required if you're already using `--install`. Any chance you ran this on your CI? :)

geraname: You are right. This is not needed as long as the priority of the alternative is high enough :+1:

I tested that the container image builds successfully on my machine for the full release following the [Docker documentation](https://github.com/ggml-org/llama.cpp/blob/master/docs/docker.md#building-docker-locally). I did not run this within the GHA CI, as I have no ability to do so in my environment.

taronaeo (approved):

CISC (approved):

PR #20535, opened 2026-03-14
chat: fix thinking support for Kimi K2 Thinking and K2.5

### LINKED ISSUE #20008: Eval bug: Cannot get reasoning_content with Kimi K2.5 in response (and no thinking block in Roo Code and other frontends)
### Name and Version

> /home/lissanro/pkgs/llama.cpp/build/bin/llama-cli --version
ggml_cuda_init: found 4 CUDA devices:
  Device 1: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes
  Device 2: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes
  Device 3: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes
version: 8182 (05728db18)

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

EPYC 7763 + 1 TB 3200 MHz DDR4 RAM + 4x3090 GPUs

### Models

Kimi K2.5 (Q4_X quant)

I am starting Kimi K2.5 with the latest chat_template.jinja from Moonshot AI). When server starts, I see:

```
srv          init: init: chat template, thinking = 0
```

I have attached the more detailed logs, test curl command and how I ran llama-serever.

I also tried `--reasoning-format deepseek` but it made no difference.

Running the same model with different backend other than llama.cpp works just fine with the same chat template and the same Q4_X GGUF file, reasoning content block appears just fine, both in the curl command and Roo Code while streaming.

With llama.cpp, I see that `<think>` tag is in the prompt and closing `</think>` is present in verbose "content", but for some reason generated reasoning tokens are not put in "reasoning_content" block.

In Roo Code behavior is the same while streaming using llama.cpp - it fails to put thinking in the reasoning block. Given this happens with llama.cpp and only while using K2.5, so I think it is a bug, but please let me know if I missed some command line parameter to make thinking work properly (it works just fine with Qwen3.5 and GPT-OSS series - they produce reasoning content block just fine, so the issue seems to be specific to combination of K2.5 with llama.cpp).

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console
```
</details>

```
> /home/lissanro/pkgs/llama.cpp/build/bin/llama-server \
--model /mnt/neuro/models/Kimi-K2.5/Kimi-K2.5-Q4_X.gguf \
--fit on --fit-ctx 262144 -b 4096 -ub 4096 -fa on \
--threads 64 --host 0.0.0.0 --port 5000 \
--jinja --chat-template-file /mnt/secondary_22tb/models/Kimi-K2.5/chat_template.jinja -b 4096 -ub 4096 \
--slot-save-path /var/cache/llama.cpp/k2.5 --cache-ram 131072 --fit-target 512 \
--min-p 0.01 --top-p 0.95 --temp 1.0 --top-k 100 --reasoning-budget -1 --verbose --mmap
...
> curl -s -X POST http://localhost:5000/v1/chat/completions -H "Content-Type: application/json" -d '{
    "model": "kimi-k2.5",         
    "messages": [
      {"role": "user", "content": "What is 2+2?"}
    ],
    "max_tokens": 512,
    "stream": false
  }' | jq                  
{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "The user is asking a simple math question: \"What is 2+2?\"\n\nThis is a straightforward arithmetic question. The answer is 4.\n\nI should provide a clear, direct answer without overcomplicating it. While there could be philosophical or joke interpretations (like \"2+2=5\" in Orwellian contexts, or base 3 where 2+2=11), the standard mathematical answer in base 10 is 4.\n\nGiven the simplicity of the question, I'll provide the direct answer: 4.\n\n2 + 2 = **4**"
      }
    }
  ],
  "created": 1772364891,
  "model": "Kimi-K2.5-Q4_X.gguf",
  "system_fingerprint": "b8182-05728db18",
  "object": "chat.completion",
  "usage": {
    "completion_tokens": 118,
    "prompt_tokens": 33,
    "total_tokens": 151
  },
  "id": "chatcmpl-eNY44xaJO0xvcMicCJaEa55FIGhqgwOe",
  "__verbose": {
    "index": 0,
    "content": "The user is asking a simple math question: \"What is 2+2?\"\n\nThis is a straightforward arithmetic question. The answer is 4.\n\nI should provide a clear, direct answer without overcomplicating it. While there could be philosophical or joke interpretations (like \"2+2=5\" in Orwellian contexts, or base 3 where 2+2=11), the standard mathematical answer in base 10 is 4.\n\nGiven the simplicity of the question, I'll provide the direct answer: 4.</think>2 + 2 = **4**",
...
    "prompt": "<|im_system|>system<|im_middle|>You are Kimi, an AI assistant created by Moonshot AI.<|im_end|><|im_user|>user<|im_middle|>What is 2+2?<|im_end|><|im_assistant|>assistant<|im_middle|><think>",
...
```

Lissanro:
After the recent chat template parser refactoring, the Kimi K2 Thinking/K2.5 thinking/reasoning support remained broken. The template injects the <think> tag in the prompt, but the parser was not correctly handling reasoning extraction. Please see the issue #20008 for details.

I also moved tool call and thinking marker constants to function scope for reuse. I also used "if" block instead of a ternary operator, but the logic is the same, except now it properly handles thinking.

Fixes #20008

The previous attempt at fixing this (PR #19234) no longer applies after the refactoring and remained unupdated. I wrote this implementation to fix the #20008 issue in the new refactored code after the autoparser was added.

I have tested with `--reasoning-budget` set to both `-1` (enable thinking) and `0` (disabled thinking) using Kimi K2.5 and Kimi K2 Thinking. I also verified that K2 0905 (the non-thinking model) still works correctly.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1167,19 +1167,30 @@ static common_chat_params common_chat_params_init_kimi_k2(const common_chat_temp
                                                           const autoparser::templates_params & inputs) {
     common_chat_params data;
 
-    data.prompt             = common_chat_template_direct_apply(tmpl, inputs);
-    data.format             = COMMON_CHAT_FORMAT_PEG_NATIVE;
-    data.supports_thinking  = true;
-    data.thinking_start_tag = "<think>";
-    data.thinking_end_tag   = "</think>";
-    data.preserved_tokens  = {
-        "<|tool_calls_section_begin|>",
-        "<|tool_calls_section_end|>",
-        "<|tool_call_begin|>",
-        "<|tool_call_argument_begin|>",
-        "<|tool_call_end|>",
-        "<think>",
-        "</think>",
+    // Tool call markers
+    const std::string SECTION_BEGIN = "<|tool_calls_section_begin|>";
+    const std::string SECTION_END   = "<|tool_calls_section_end|>";
+    const std::string CALL_BEGIN    = "<|tool_call_begin|>";
+    const std::string ARGS_BEGIN    = "<|tool_call_argument_begin|>";
+    const std::string CALL_END      = "<|tool_call_end|>";
+
+    const std::string THINK_START   = "<think>";
+    const std::string THINK_END     = "</think>";
+
+    data.prompt               = common_chat_template_direct_apply(tmpl, inputs);
+    data.format               = COMMON_CHAT_FORMAT_PEG_NATIVE;
+    data.supports_thinking    = true;
+    data.thinking_forced_open = inputs.enable_thinking;
+    data.thinking_start_tag   = THINK_START;
+    data.thinking_end_tag     = THINK_END;
+    data.preserved_tokens     = {
+        SECTION_BEGIN,
+        SECTION_END,
+        CALL_BEGIN,
+        ARGS_BEGIN,
+        CALL_END,
+        THINK_START,
+        THINK_END,
     };
 
     auto has_tools         = inputs.tools.is_array() && !inputs.tools.empty();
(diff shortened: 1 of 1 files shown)


Alexey-Akishin:
This patch helped, thank you so much! Now k2.5 actually puts its thoughts into the thinking block!

I also tested out the mode with thinking disabled, and works too - the model no longer does long thinking in actual response with --reasoning-budget 0, but does short responses as expected with the thinking turned off... without the patch, it fails to behave correctly in both cases and just does the long thinking in the message no matter what --reasoning-budget set to...

thanks again, I am happy I can now properly run k2.5 with llama.cpp with this patch applied.

gordan-bobic:
I could also do with this fix. I think this is what is breaking KV caching of the generated responses that I am seeing. Unfortunately the patch doesn't merge cleanly any more.

PR #20538, opened 2026-03-14
ci : reduce webgpu tests timeout to 900s

ggerganov:
The `test-thread-safety` test has been recently getting stuck with the WebGPU backend causing long runs. Reduce the timeout from 3600s -> 900s to reduce the CI hangs.

https://github.com/ggml-org/llama.cpp/actions/runs/23063874570/job/66997161164#step:10:17552
https://github.com/ggml-org/llama.cpp/actions/runs/23045127401/job/66932266420#step:10:17632
https://github.com/ggml-org/llama.cpp/actions/runs/22998877128/job/66778103540#step:10:17630

cc @reeselevine

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -542,7 +542,7 @@ jobs:
         run: |
           cd build
           # This is using llvmpipe and runs slower than other backends
-          ctest -L main --verbose --timeout 3600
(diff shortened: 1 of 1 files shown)


reeselevine:
Thanks, this seems like a reasonable default timeout anyways, while I try and figure out what is going on with that test.

PR #20544, opened 2026-03-14
common,tools : move fs helpers to fs-impl.h to avoid duplicated code

angt:


angt:
I've only changed the `GGML_ABORT` to `throw std::runtime_error` in the `__EMSCRIPTEN__` section.
I guess it is safe as it used at many places.

common/fs-impl.h
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    std::wstring wstr(size, 0);
+    MultiByteToWideChar(CP_UTF8, 0, str.c_str(), (int)str.size(), &wstr[0], size);
+
+    return wstr;
+}
+#endif
+
+#ifndef COMMON_EXPORT
```
rgerganov, line 49: what is the purpose of this?

angt:
This is to prepare a series of patches related to the cache system in the near future.

This `fs-impl.h` file allows for inlining in tools that only require reading the cache, without needing all the common files. It still ensures that the file is correctly included in `common.cpp`, avoiding code duplication across all compilation units.

rgerganov (commented):
I am not sure this approach is significantly better. It's inconsistent to have part of the `fs` functions in `common.cpp` and another part in some `fs-impl.h`. Putting implementation in common header files feels like a hack, not sure how this is better than having duplicate code.

common/fs-impl.h
angt: This is to support both usages:
- The standalone one from `rpc-server`, where we only use `fs-impl.h` to have inlined functions.
- The non-inlined one for `common.cpp`, where we have usual functions compiled once for all (with prototypes in `common.h`).

This is mostly to avoid `proxy` functions in `common.cpp` so we can just use add them and remove this code if you think it's cleaner.

rgerganov:
I think the right way to remove the duplicate code is to have a common library which doesn't depend on `libllama`, so programs in `/tools` can use common utilities without pulling extra dependencies like `libllama`.

That said, if other maintainers are OK with the approach here, I won't hold this off and you can merge it.

ggerganov (approved):
I agree with @rgerganov's idea for extracting a small `libcommon` that does not depend on `libllama`. And the current `libcommon` becoming `libcommon-llama`. We have the same pattern for the examples in the `ggml` repo:

https://github.com/ggml-org/ggml/blob/master/examples/CMakeLists.txt

We can do this in a follow-up PR.

The `COMMON_EXPORT` is a bit confusing to me. I'm more used to the [stblib](https://github.com/nothings/stb)-style implementation macros and think they make better sense. For example, use `COMMON_FS_IMPL` macro to gate the implementation in `fs.h` and in `libcommon` include once in `common/fs.cpp`:

```c++
#define COMMON_FS_IMPL
#include "fs.h"
```

Same in `rpc-server.cpp`.

The rest of the code in `libcommon` just includes `fs.h` without defining the implementation macro.

angt:
Ok so let's not merge this one, i'll first do `libcommon-llama` 👍

PR #16682, opened 2025-10-20
RVV1.0 Supported tests for RISC-V

alitariq4589:
This PR adds tests supported for RISC-V

## Tests which are added for execution with RISC-V

`debian-cpu-cmake-rv64-native`
`debian-trixie-cmake-sanitizer-riscv64-native` * 3
`debian-trixie-llguidance-riscv64-native`
`debian-trixie-cmake-rpc-riscv64-native`
`ggml-ci-riscv64-native-cpu-low-perf`

## Dependencies which are not yet supported for RISC-V:

`rocblas-dev`
`hipblas-dev`
`vulkan-sdk`
`mthreads/musa:rc4.3.0-devel-ubuntu22.04-amd64`
`intel-oneapi-compiler-dpcpp-cpp`
`intel-oneapi-mkl-devel`
`cuda`
`macOS*`
`windows*`
`torch`

## Tests which are not added for RISC-V due to above unmet dependencies:

`macOS-latest-cmake-arm64`
`macOS-latest-cmake-x64`
`macOS-latest-cmake-arm64-webgpu`
`ubuntu-24-cmake-vulkan`
`ubuntu-22-cmake-webgpu`
`ubuntu-22-cmake-hip`
`ubuntu-22-cmake-musa`
`ubuntu-22-cmake-sycl`
`ubuntu-22-cmake-sycl-fp16`
`macOS-latest-cmake-ios`
`macOS-latest-cmake-tvos`
`macOS-latest-cmake-visionos`
`macOS-latest-swift`
`windows-msys2`
`windows-latest-cmake`
`ubuntu-latest-cmake-cuda`
`windows-2022-cmake-cuda`
`windows-latest-cmake-sycl`
`windows-latest-cmake-hip`
`ios-xcode-build`
`android-build`
`openEuler-latest-cmake-cann`
`ggml-ci-x64-nvidia-cuda`
`ggml-ci-x64-nvidia-vulkan-cm`
`ggml-ci-x64-nvidia-vulkan-cm2`
`ggml-ci-x64-cpu-amx`
`ggml-ci-mac-metal`
`ggml-ci-mac-vulkan`
`ggml-ci-arm64-cpu-high-perf-sve`
`ggml-ci-riscv64-native-cpu-high-perf`

### Note 1

Due to a warning (treated as error) related to RISC-V simd mappings

```
In file included from ../../../ggml/src/ggml-cpu/arch/riscv/quants.c:6:
[...]

PR #20546, opened 2026-03-14
ci : split build.yml + server.yml

ggerganov:
cont #20540 

- Split `build.yml` + `server.yml` into parts and move some of the workflows in the new parts
- Continue to run `build.yml` + `server.yml` on all PRs and `master` branch
- The new build parts run only on `master` branch
- The new build parts can be manually triggered on PRs
- Fork and use [setup-xcode](https://github.com/ggml-org/setup-xcode) action
- Rename workflow: drop `-cmake`, improve consistency
- Rename `ggml-ci-arm64-graviton4-kleidiai` -> `ggml-ci-arm64-cpu-kleidiai-graviton4`
- Remove `ubuntu-llguidance-riscv64-native`
- Remove `ubuntu-cmake-rpc-riscv64-native`

Note: not sure how the GA cache works, but I suspect that when we run this new CI it will trash the caches because of the new workflow names

### Next PRs

- [ ] Move `ggml-ci-*-cpu-*` workflows to the self-hosted runners to free some GH cache
- [x] Drop `spacemit` GH cache: https://github.com/ggml-org/llama.cpp/pull/20546#issuecomment-4062743230


CISC:
> Note: not sure how the GA cache works, but I suspect that when we run this new CI it will trash the caches because of the new workflow names

Since you also renamed the keys, yes, but since we just about can't fit a whole run in the 10GB we have available we are partially thrashing the cache already. This will just leave older PRs in the dark until they catch up, not a major issue I think.

ggerganov:
I moved the sanitizer workflows in separate `*-sanitize.yml` files to be executed manually on PRs.

In the next PR, I think we can move some of the `ggml-ci` CPU workflows to `self-hosted` runners since we have both x86 and arm64 machines enrolled now. This will probably free up some GH cache.

.github/workflows/build-3rd-party.yml
```diff
@@ -0,0 +1,65 @@
...
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
```
CISC:
```suggestion
      '**/*.cpp'
```

.github/workflows/build-apple.yml
```diff
@@ -0,0 +1,212 @@
...
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
```
CISC:
```suggestion
      '**/*.swift',
      '**/*.m',
      '**/*.metal'
```

.github/workflows/build-cann.yml
```diff
@@ -0,0 +1,103 @@
...
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
```
CISC:
```suggestion
      '**/*.cpp'
```

.github/workflows/build-riscv.yml
```diff
@@ -0,0 +1,137 @@
...
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
```
CISC:
```suggestion
      '**/*.cpp'
```

.github/workflows/build-sanitize.yml
```diff
@@ -0,0 +1,95 @@
...
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
```
CISC:
```suggestion
      '**/*.cpp'
```

.github/workflows/build-vulkan.yml
```diff
@@ -0,0 +1,95 @@
...
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp',
+      '**/*.glsl',
+      '**/*.wgsl'
```
CISC:
```suggestion
      '**/*.comp',
      '**/*.glsl'
```

.github/workflows/server-sanitize.yml
```diff
@@ -0,0 +1,96 @@
...
+        required: false
+        type: string
+      slow_tests:
+        description: 'Run slow tests'
+        required: true
+        type: boolean
+  push:
+    branches:
+      - master
+    paths: ['.github/workflows/server-sanitize.yml', '**/CMakeLists.txt', '**/Makefile', '**/*.h', '**/*.hpp', '**/*.c', '**/*.cpp', '**/*.cu', '**/*.swift', '**/*.m', 'tools/server/**.*']
```
CISC:
```suggestion
    paths: ['.github/workflows/server-sanitize.yml', '**/CMakeLists.txt', '**/Makefile', '**/*.h', '**/*.hpp', '**/*.c', '**/*.cpp', 'tools/server/**.*']
```

CISC (approved):

ggerganov:
I think we should drop the `spacemit` cache:

image

The `build-riscv.yml` should run only on changes to `ggml/src/ggml-cpu/spacemit/*` and should not use the GH cache.

CISC:
> I think we should drop the `spacemit` cache:
> image
> 
> The `build-riscv.yml` should run only on changes to `ggml/src/ggml-cpu/spacemit/*` and should not use the GH cache.

It's not running on riscv, it's a crossbuild, IIRC it's a special arch.

ggerganov:
The #20585 and #20588 should free quite some GH cache too.

ggerganov:
- Extracted the android builds to run only on `master`
- For the split workflows, they now also run in the PRs if files are touched directly. For example, modifying `ggml-metal` sources will trigger the `build-apple.yml` workflow even in the PRs

.github/workflows/release.yml
```diff
@@ -10,7 +10,21 @@ on:
...
+      '**/*.h',
+      '**/*.hpp',
+      '**/*.c',
+      '**/*.cpp',
+      '**/*.cu',
+      '**/*.cuh',
+      '**/*.swift',
+      '**/*.m',
+      '**/*.metal',
+      '**/*.comp'
```
```suggestion
      '**/*.comp',
      '**/*.glsl'
```

CISC (approved):

ggerganov:
I think I'll yolo this into master and see what happens.

CISC:
> I think I'll yolo this into master and see what happens.

https://github.com/ggml-org/llama.cpp/actions/runs/23111057141/job/67128454775

ggerganov:
Should be fixed now - I incorrectly forked only the `master` branch earlier.

alitariq4589:
> Remove ubuntu-llguidance-riscv64-native
Remove ubuntu-cmake-rpc-riscv64-native

Were these removed due to hardware resource constraints, or was there some other issue?

I was about to enable `ggml-ci-riscv64-native-cpu-high-perf` for RISC-V, which was not added due to the absence of dependencies (torch and transformer) mentioned in #16682. Pytorch and transformers can now be built from source for RISC-V and I am also planning to host wheels for them. But seeing the new `build_riscv.yaml` file, I can see low-perf was also removed along with the tests mentioned above.

CISC:
> > Remove ubuntu-llguidance-riscv64-native
> > Remove ubuntu-cmake-rpc-riscv64-native
> 
> Were these removed due to hardware resource constraints, or was there some other issue?
> 
> I was about to enable `ggml-ci-riscv64-native-cpu-high-perf` for RISC-V, which was not added due to the absence of dependencies (torch and transformer) mentioned in #16682. Pytorch and transformers can now be built from source for RISC-V and I am also planning to host wheels for them. But seeing the new `build_riscv.yaml` file, I can see low-perf was also removed along with the tests mentioned above.

I think they were removed when trying to cut down on the queue time and that they were technically superfluous. Adding `high-perf` would be good for test coverage and time is no longer that much of a concern as it will only run on `RISC-V` backend changes and merges.

PR #20558, opened 2026-03-14
Server becomes the source of truth for sampling parameter defaults

### LINKED ISSUE #18129: Misc. bug: Web-Frontend overrides sampling parameters from models preset in router mode
### Name and Version

```
$ llama-server --version
[...]
version: 7435 (79dbae034)
```

### Operating systems

Linux

llama-server

### Command line

```shell
llama-server --host 127.0.0.1 --port 8000 --jinja --prio 2 --no-mmap --threads -1 -np 3 -fa on --models-preset models.ini

With the following models.ini:


version = 1

[Qwen3-Next-80B-A3A-Thinking-Q4KXL]
model = /path/to/Qwen3-Next-80B-A3B-Thinking-UD-Q4_K_XL.gguf
c = 86016
temp = 0.6
top-p = 0.95
min-p = 0
top-k = 20
cpu-moe = true
presence-penalty = 1.0

[GPT-OSS-20B-F16]
model = /path/to/gpt-oss-20b-F16.gguf
c = 98304
temp = 1.0
top-p = 1.00
min-p = 0
top-k = 100
```

I'm running `llama-server` in router mode with a model preset file containing the recommended sampling parameters for each model. However, the web frontend has different default sampling parameters that are not updated when I switch to a model. While the model's process does use the sampling parameters (as arguments) configured in the preset file, they get overwritten by the web frontend's default sampling parameters in the corresponding `/v1/chat/completions` request. Clicking "Reset to default" in the web frontend's sampling parameter settings also does not apply the current model's sampling parameters.

The request to `/v1/chat/completions` sends the following sampling parameters:
```
temperature: 0.8
top-k: 40
top-p: 0.95
min-p: 0.05
```

### First Bad Commit

This is the first time that I use the model preset feature, so I assume it always behaved like this since the feature was just merged.

ServeurpersoCom:

### Server becomes the source of truth for sampling parameter defaults

Sampling defaults in the webui are now empty strings instead of
hardcoded values duplicated from common.h. Empty means "not sent
in the request" (already handled by hasValue()), so the server
decides. Placeholders show the actual server/preset values via /props.

### Fixes

Fixes the "Custom" badge that never appeared on sampling parameters
because client defaults always matched server defaults. Now any
user-set value that differs from the server is flagged.

Close #18129


ServeurpersoCom:
The single model mode still needs to be tested. We can also store overrides per model, but that needs to be done in a separate PR, it's a negative line PR, this simplifie le code. WDYT @allozaur

ServeurpersoCom:
It works in principle, but something doesn't sit right with me; the placeholder updates don't seem reliable.

allozaur:
> It works in principle, but something doesn't sit right with me; the placeholder updates don't seem reliable.

just tested it on my end and things look normal on my end

tools/server/webui/src/lib/constants/settings-config.ts
```diff
@@ -30,27 +30,30 @@ export const SETTING_CONFIG_DEFAULT: Record<string, string | number | boolean> =
...
+	repeat_last_n: '',
+	repeat_penalty: '',
+	presence_penalty: '',
+	frequency_penalty: '',
+	dry_multiplier: '',
+	dry_base: '',
+	dry_allowed_length: '',
+	dry_penalty_last_n: '',
+	max_tokens: '',
 	custom: '', // custom json-stringified object
```
allozaur: empty strings are wrong type of data for the initial state... i think that `undefined` would work better


```suggestion
	temperature: undefined,
	dynatemp_range: undefined,
	dynatemp_exponent: undefined,
	top_k: undefined,
	top_p: undefined,
	min_p: undefined,
	xtc_probability: undefined,
	xtc_threshold: undefined,
	typ_p: undefined,
	repeat_last_n: undefined,
	repeat_penalty: undefined,
	presence_penalty: undefined,
	frequency_penalty: undefined,
	dry_multiplier: undefined,
	dry_base: undefined,
	dry_allowed_length: undefined,
	dry_penalty_last_n: undefined,
	max_tokens: undefined,
custom: '', // custom json-stringified object
```

allozaur (changes requested):

allozaur (approved):

ServeurpersoCom:
> > It works in principle, but something doesn't sit right with me; the placeholder updates don't seem reliable.
> 
> just tested it on my end and things look normal on my end

Actually, it's just a matter of caching child process properties, but from a UX perspective, this PR does the job.

PR #20539, opened 2026-03-14
convert : support mixed-precision ModelOpt models with per-tensor NVFP4/FP8 quantization

richarddd:
Adds support for converting mixed-precision ModelOpt models (e.g. nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4) that use per-tensor quant_algo with both NVFP4 and FP8 layers, instead of a single global quant_algo: "NVFP4". NVFP4  tensors (2D scales) are repacked natively while FP8 tensors (1D scales) are dequantized to float.

Fixes: https://github.com/ggml-org/llama.cpp/issues/20504

PR #20561, opened 2026-03-14
model : wire up Nemotron-H tensors for NVFP4 support

CISC:
prep #20539


CISC:
As an apropos, when https://github.com/ggml-org/llama.cpp/pull/20427#discussion_r2921092474 is addressed, an optional `bias` should also be added:
https://github.com/ggml-org/llama.cpp/blob/da0c8a5325446cccb0174001d3cf8284c6083d2b/src/models/nemotron-h.cpp#L69-L74

cc/ @richarddd

danbev (approved):

PR #20563, opened 2026-03-14
CUDA: WIP: Repack NVFP4 blocks into new SoA layout

michaelw9999:
This PR demonstrates the first working version of an NVFP4 repacker for CUDA and its helper functions, which
convert the block_nvfp4 into a SoA 4 pack layout.  This is a prerequisite PR for my upcoming CUDA NVFP4/SM120/generic kernels.

It repacks block_nvfp4 from:
```
#define QK_NVFP4 64
#define QK_NVFP4_SUB 16  // sub-block size for per-group scales
typedef struct {
    uint8_t d[QK_NVFP4/QK_NVFP4_SUB]; // UE4M3 scales (4 bytes, one per 16-element sub-block)
    uint8_t qs[QK_NVFP4/2];           // packed 4-bit E2M1 values (32 bytes)
} block_nvfp4;
```

...to a 4 pack AoSoA:
```
struct GGML_ALIGN_16 block_nvfp4_cuda {
    uint8_t qs[4][32];
    uint8_t scales[4][4];
};
```

It does this at runtime in the CUDA backend during upload, so CUDA buffers for NVFP4 get stored as SoA instead of what's on disk.
I am investigating whether it would help to pre-swizzle  at this time for Native BW MMA but have not included that here and wanted feedback on that.

I did not want to make an new `ggml_backend_cuda_interface` NVFP4 copy just for NVFP4 because there would be a lot of duplicated code and it would be messy, so this first version POC is a little messy, as it has to branch for NVFP4 in some of the `ggml_backend_cuda_` functions, but it kept things simple for now.  

We could template it by type or even modularize this repacker to be generic if there's any need to repack any other quant, or for future use, but I do not have enough experience yet with llama.cpp to know if that would be useful or needed.
(Maybe for the other 256 sized K-quants or generic MXFP4 to expand compatability with more models)
Suggestions and feedback on the  cleanest and best way you would like to see that handled would be very welcome.

A `test-nvfp4-repack` was created and included here.   It  generates NVFP4 E2M1 weights, UE4M3 scales, in the block_nvfp4 layout,
then calls these helper functions to verify repack/unpack works correctly along with partial row support:

```
layout-4x64-to-256:
  4 x block_nvfp4 (64 weights each) -> 1 x block_nvfp4_cuda (256 weights total)
  block_nvfp4[0] qs: 12 29 40 57 6e 85 9c b3 ca e1 f8 0f 26 3d 54 6b 82 99 b0 c7 de f5 0c 23 3a 51 68 7f 96 ad c4 db
  block_nvfp4[1] qs: 43 5a 71 88 9f b6 cd e4 fb 12 29 40 57 6e 85 9c b3 ca e1 f8 0f 26 3d 54 6b 82 99 b0 c7 de f5 0c
  block_nvfp4[2] qs: 74 8b a2 b9 d0 e7 fe 15 2c 43 5a 71 88 9f b6 cd e4 fb 12 29 40 57 6e 85 9c b3 ca e1 f8 0f 26 3d
  block_nvfp4[3] qs: a5 bc d3 ea 01 18 2f 46 5d 74 8b a2 b9 d0 e7 fe 15 2c 43 5a 71 88 9f b6 cd e4 fb 12 29 40 57 6e
  block_nvfp4[0] scales: 31 3a 43 4c
  block_nvfp4[1] scales: 3e 47 50 59
  block_nvfp4[2] scales: 4b 54 5d 66
  block_nvfp4[3] scales: 58 61 6a 73
  block_nvfp4_cuda:
  lane 0 qs @0x00: 3c5e7092 b9865421 b4d6f81a 65320fec 3c5e7092 20fdcb98 b4d6f81a dca97653
  lane 1 qs @0x20: 4d6f81a3 ecb98754 c5e7092b 9865421f 4d6f81a3 5320fecb c5e7092b 0fdcb986
  lane 2 qs @0x40: 5e7092b4 1fedba87 d6f81a3c cb987542 5e7092b4 865421fe d6f81a3c 320fecb9
  lane 3 qs @0x60: 6f81a3c5 4210edba e7092b4d fedba875 6f81a3c5 b9875421 e7092b4d 65421fec
  lane 0 scales @0x80: 31 3a 43 4c
  lane 1 scales @0x84: 3e 47 50 59
  lane 2 scales @0x88: 4b 54 5d 66
  lane 3 scales @0x8c: 58 61 6a 73
test-nvfp4-repack: 8/8 passed
```

Partial tiles of 64/128/192 are OK as it accounts for tails, so a full 256w block_nvfp4_cuda isn't required.  Those partial tails would not really be 16B aligned, however.
I am doing a final refactor and cleaning on the MMA kernels that will use these functions, and then will post those separately.
Apologies for anything not perfect, still learning much of this stuff!
AI was used to help create some of this code used here under very focused direction.

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -36,6 +36,7 @@
 #include "ggml-cuda/pad.cuh"
 #include "ggml-cuda/pool2d.cuh"
 #include "ggml-cuda/quantize.cuh"
+#include "ggml-cuda/repack_nvfp4.cuh"
 #include "ggml-cuda/rope.cuh"
 #include "ggml-cuda/roll.cuh"
 #include "ggml-cuda/scale.cuh"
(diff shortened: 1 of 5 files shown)


michaelw9999:
@ORippler @JohannesGaessler @am17an appreciate any feedback on improving this, thanks!

am17an:
I'm working on a similar idea for mxfp4, I will take look at this PR after that

JohannesGaessler:
We should not be repacking any data until we have a baseline implementation without any repacking to compare against.

ORippler:
> It does this at runtime in the CUDA backend during upload, so CUDA buffers for NVFP4 get stored as SoA instead of what's on disk.
I am investigating whether it would help to pre-swizzle at this time for Native BW MMA but have not included that here and wanted feedback on that.

Any work that happens at load time executes once, whereas work that happens at run-time (i.e. in the MMQ kernels) executes every time we invoke the kernel. So from that standpoint it makes sense to do everything that can be done at load time at load time. Ultimate goal is that `ldmatrix` can be used to load from SMEM -> R without bank conflicts. Note that consumer BW does not have TMA available, so pre-swizzling would have to happen prior to loading into GMEM if we want to use LDGSTS exclusively, and I'm not sure how feasible that is (as I have not done a deep-dive to match layouts of `block_nvfp4_cuda` and `ldmatrix`/ an. It can of course also be done manually if we go via registers at the cost of higher register pressure (which could be offset by introducing warp-specialization):

https://yang-yifan.github.io/blogs/mma_swizzle/mma_swizzle.html
https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/async-copies.html#using-ldgsts
https://docs.nvidia.com/cuda/parallel-thread-execution/#warp-level-block-scaling
https://pytorch.org/blog/cutlass-ping-pong-gemm-kernel/

In general, I think it makes sense to utilize abstractions and libs provided by NVDA (CUTLASS + CuTe/CuTile) to do the heavy lifting instead of doing a custom implementation, though I'm obviously biased here (in that light one would ensure those blocks can be plugged into relevant data-structures exposed by those libs).

> We should not be repacking any data until we have a baseline implementation without any repacking to compare against.

Agreed, and I think mxfp4 kernels would be a good starting point as we have a baseline implementation available (afaik @am17an already demonstrated perf-improvement when repacking during gguf generation).

> We could template it by type or even modularize this repacker to be generic if there's any need to repack any other quant, or for future use, but I do not have enough experience yet with llama.cpp to know if that would be useful or needed.
(Maybe for the other 256 sized K-quants or generic MXFP4 to expand compatability with more models)
Suggestions and feedback on the cleanest and best way you would like to see that handled would be very welcome.

Feel free to take a look at how it's done in https://github.com/ggml-org/llama.cpp/blob/master/ggml/src/ggml-cpu/repack.cpp and https://github.com/ggml-org/llama.cpp/blob/master/ggml/src/ggml-cpu/repack.h for the CPU backend, that has already implemented repacking for multiple data-types/quants.

ORippler:
Had a skim through the CPU-repack, and it seems that we don't pad individual dimensions, which would be a requirement for perf optimality. Basically, every contiguous dim of interest in `ggml_tensor -> ne` has to be 16-byte-aligned. So I guess we would have to adjust `ggml_backend_cuda_buffer_type_get_alloc_size` accordingly.

michaelw9999:
@ORippler I spent more time playing with it anyway and made some more updates but didn't post them here, I can later.  It should just use load_generic and so need nearly zero staging

PR #20565, opened 2026-03-14
vulkan: fix PowerVR shader compiler crash with subgroup reduction mode

Konoxx:
## Summary

- Adds `VK_VENDOR_ID_IMAGINATION` (0x1010) vendor ID constant
- Overrides `SHADER_REDUCTION_MODE_SUBGROUP` to `SHADER_REDUCTION_MODE_HYBRID` for Imagination Technologies GPUs
- Fixes `VK_ERROR_UNKNOWN` (-13) crash during `vkCreateComputePipelines` on all quantized matmul shaders

## Root Cause

PowerVR's shader compiler crashes on the `subgroup_no_shmem` SPIR-V variant when the specialization constant `NUM_COLS >= 2`. With `NUM_COLS=1` it works fine. The other two reduction modes (`SHMEM` and `HYBRID`) work correctly with all `NUM_COLS` values.

Isolated through 8 targeted test binaries that systematically narrowed the failure from "all pipelines crash" to "this specific shader variant with this specific specialization constant value."

## Fix

Two changes:
1. Define `VK_VENDOR_ID_IMAGINATION 0x1010` alongside existing vendor IDs
2. After computing `reduc`/`reduc16`, override `SUBGROUP` → `HYBRID` for Imagination GPUs

The HYBRID mode uses subgroup operations with a shared memory fallback, which the PowerVR compiler handles correctly. Performance impact is minimal — HYBRID is only slightly slower than pure SUBGROUP on other vendors.

## Testing

Tested on **PowerVR Immortalis G925** (Google Pixel 10 Pro XL, Tensor G5):
- Vulkan 1.3.288, subgroup size 128, 32KB shared memory
- All quantized matmul pipelines now compile and execute successfully
- Benchmarked at 4-5 tok/s GPU inference (vs crash before fix)

**Note:** GPU compute output has a separate correctness issue (garbled text) on this device that appears to be a deeper driver precision bug unrelated to this shader compilation fix. CPU inference works correctly.

## Test plan

- [x] Verify pipelines compile on PowerVR Immortalis G925
- [x] Verify no regression on non-Imagination GPUs (override is vendor-gated)
- [x] Benchmark GPU inference runs without crash
- [ ] Test on other Imagination/PowerVR GPUs if available

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -93,6 +93,7 @@ static bool is_pow2(uint32_t x) { return x > 1 && (x & (x-1)) == 0; }
 #define VK_VENDOR_ID_INTEL 0x8086
 #define VK_VENDOR_ID_NVIDIA 0x10de
 #define VK_VENDOR_ID_QUALCOMM 0x5143
+#define VK_VENDOR_ID_IMAGINATION 0x1010
 
 #define VK_DEVICE_DESCRIPTOR_POOL_SIZE 256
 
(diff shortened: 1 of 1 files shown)


jeffbolznv:
Were there any validation errors with the original code?

PR #20566, opened 2026-03-15
fix(openvino): replace hardcoded TBB archive path with find_package(TBB)

Kepard:
## Problem

The OpenVINO backend fails to configure on Ubuntu when OpenVINO is installed via apt instead of the official archive installer.

`ggml/src/ggml-openvino/CMakeLists.txt` line 4 hardcodes:

```cmake
include("${OpenVINO_DIR}/../3rdparty/tbb/lib/cmake/TBB/TBBConfig.cmake")
```

This path assumes the archive install layout where TBB is bundled under OpenVINO's own directory tree. On apt-based installs, `OpenVINO_DIR` resolves to `/usr/lib/cmake/openvino2026.0.0` and the `../3rdparty/tbb` path does not exist, causing:

```
CMake Error at ggml/src/ggml-openvino/CMakeLists.txt:4 (include):
  include could not find requested file:
    /usr/lib/cmake/openvino2026.0.0/../3rdparty/tbb/lib/cmake/TBB/TBBConfig.cmake
```

## Fix

Replace the hardcoded include with a standard `find_package(TBB REQUIRED)` call, which works for both archive and apt installs.

```cmake
# Before
include("${OpenVINO_DIR}/../3rdparty/tbb/lib/cmake/TBB/TBBConfig.cmake")

# After
find_package(TBB REQUIRED)
```

For apt-based installs, users also need to pass `-DTBB_DIR=/usr/lib/x86_64-linux-gnu/cmake/TBB` to cmake, since the apt package does not ship `TBBConfig.cmake`. A note has been added to the OpenVINO backend docs.

## Tested On

- Ubuntu 25.10 x86_64
- OpenVINO 2026.0.0 (apt)
- libtbb-dev (apt)
- Intel Core Ultra 7 258V / Arc 130V integrated GPU
- llama.cpp commit d23355afc

## Docs Update

The `docs/backend/OPENVINO.md` build instructions should add for apt-based Ubuntu installs:

```bash
sudo apt install libtbb-dev

cmake -B build/ReleaseOV -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DGGML_OPENVINO=ON \
  -DOpenVINO_DIR=/usr/lib/cmake/openvino2026.0.0 \
  -DTBB_DIR=/usr/lib/x86_64-linux-gnu/cmake/TBB
```

### DIFF
--- a/ggml/src/ggml-openvino/CMakeLists.txt
+++ b/ggml/src/ggml-openvino/CMakeLists.txt
@@ -1,7 +1,7 @@
 find_package(OpenVINO REQUIRED)
 find_package(OpenCL REQUIRED)
 
-include("${OpenVINO_DIR}/../3rdparty/tbb/lib/cmake/TBB/TBBConfig.cmake")
+find_package(TBB REQUIRED)
 
 file(GLOB_RECURSE GGML_HEADERS_OPENVINO "*.h" "*.hpp")
 file(GLOB_RECURSE GGML_SOURCES_OPENVINO "*.cpp")


taronaeo:
cc: @cavusmustafa @wine99

PR #20584, opened 2026-03-15
common: improve error message when setting process priority fails

MaheshJakkala:
On POSIX systems, negative nice values require elevated privileges.

When users run with --prio high or --prio realtime, the call to
setpriority() may fail with EPERM if the process does not have
sufficient permissions.

This patch improves the warning message to clarify that elevated
privileges (e.g., running with sudo) may be required.


0cc4m:
You messed up the branch, rebase so that it only contains your commits.

JohannesGaessler:
This PR is broken in terms of git, make a new one with a fixed branch.

PR #20590, opened 2026-03-15
ggml-rpc: add native RDMA transport for RPC backend (RoCEv2)

dvv101111:
Adds optional RDMA (RoCEv2) transport to the RPC backend

Two-node cluster: AMD Radeon 8060S (gfx1151) iGPUs, ConnectX-4 Lx / ConnectX-6 Lx 25GbE, RoCEv2. Model: Qwen3-Coder-Next 80B Q8_K_XL, layer split (-sm layer -ts 1/1) across both nodes, ROCm backend.

| Metric                     | TCP    | RDMA   | Improvement |
| -------------------------- | ------ | ------ | ----------- |
| Prompt processing (pp2048) | 651.48 | 678.42 | **+4.1%**   |
| Token generation (tg256)   | 30.19  | 32.16  | **+6.5%**   |

Optional: libibverbs-dev / rdma-core (Linux only). Without the cmake flag, nothing changes -- no new headers, no new linkage.


ggerganov:
@dvv47 Do you know if I connect 2 DGX Sparks with a 200Gb/s QSFP cable, is this implementation going to be compatible? I am planning to buy the cable and test it.

Mithras:
I might be doing something wrong but this is as far as I was able to get:
image
I have dual strix halo with x-5 and x-4 in ETH mode. No matter what `GGML_RDMA_GID` I set, I'm just getting infinite
```
llama-1  | Accepted client connection
llama-1  | RDMA probed: dev=rocep101s0 gid=5 qpn=322 inline=316
llama-1  | RDMA activated: qpn=322->463 mtu=1024 rx_depth=24
llama-1  | [get_device_memory] device: 0, free_mem: 134044098560, total_mem: 134217728000
llama-1  | Client connection closed
llama-1  | RDMA CQ poll timeout
```
loop. Models never finish loading. It might be something specific to containers... I've tried host network, priviledged, mapping `/dev/infiniband`, etc. Wasn't able to make this work. Is there something obvious I'm missing?

dvv101111:
@ggerganov  Yes, it will work with dgx spark, as it is, they use Mellanox ConnectX-7 and Libibverbs api exactly same. Low-latency benefits will be there, but full bandwidth utilization for faster model transfer weight could be optimized even further

dvv101111:
@Mithras it's possible that this is a docker limitation. Have you tried running it natively? I will also test it with Docker and will let you know the results. Maybe using --ipc=host will help.

dvv101111:
Fixed the issue, it now reconnects better, and I have retested with the ROCm 7.2 Docker container and 
My previous test was with nightly version

image

Also, according to my tests the Vulkan backend has a lower RDMA profit, about 3%, in both PP and TG. I am also interested in testing how this affects CUDA devices. I will probably try it this week.


@Mithras for this strix halo:
```
docker run \
    --network host \
    --ipc host \
    --privileged \
    --ulimit memlock=-1:-1 \
    --security-opt seccomp=unconfined \
    --device /dev/dri \
    --device /dev/kfd \
    -v /dev/infiniband:/dev/infiniband \
    -e HSA_OVERRIDE_GFX_VERSION=11.5.1
```

and if you build it yourself:
```
RUN cmake -B build \
    -DGGML_HIP=ON \
    -DGGML_HIP_NO_VMM=ON \
    -DLLAMA_HIP_UMA=ON \
    -DGGML_CUDA_ENABLE_UNIFIED_MEMORY=ON \
    -DCMAKE_HIP_FLAGS="--rocm-path=/opt/rocm -mllvm --amdgpu-unroll-threshold-local=600" \
    -DAMDGPU_TARGETS=gfx1151 \
    -DGGML_RPC=ON \
    -DGGML_RPC_RDMA=ON
```

Mithras:
@dvv47 it mostly works now, thank you so much! A weird issue that still present is that it works until you let it idle for a few mins. Then it throws the same timeout error:

But that only happens after you let it idle for a couple mins. I was able to run pretty big prompts from perplexica prior to that just fine

---

Also, it seems the only thing that's needed to make RDMA work in docker is passing `/dev/infiniband` device

dvv101111:
@Mithras fixed that and tested on some agentic tasks

Mithras:
@dvv47 this works great, thank you so much! Real perf boost for free!
Another feature/question: Would it be possible to set `GGML_RDMA_DEV` and `GGML_RDMA_GID` per RPC? I have two strix halos and one of them has two port x-5 NIC  with one port connected to my desktop PC. It'll probably just work if I use `network: host` but it would be nice to be able to use multi-port setup in isolated network where automatic dev/gid detection doesn't work.

krampenschiesser:
Works with virtual roce v2, too. but of course slightly slower than the ethernet stack below it.
(used rdma link add rxe0 type rxe netdev eth0)
```
ggml_cuda_init: found 5 CUDA devices (Total VRAM: 79251 MiB):
  Device 0: NVIDIA GeForce RTX 5060 Ti, compute capability 12.0, VMM: yes, VRAM: 15850 MiB
  Device 1: NVIDIA GeForce RTX 5060 Ti, compute capability 12.0, VMM: yes, VRAM: 15850 MiB
  Device 2: NVIDIA GeForce RTX 5060 Ti, compute capability 12.0, VMM: yes, VRAM: 15850 MiB
  Device 3: NVIDIA GeForce RTX 5060 Ti, compute capability 12.0, VMM: yes, VRAM: 15850 MiB
  Device 4: NVIDIA GeForce RTX 5060 Ti, compute capability 12.0, VMM: yes, VRAM: 15850 MiB
RDMA probed: dev=rxe0 gid=2 qpn=20 inline=256
RDMA activated: qpn=20->20 mtu=4096 rx_depth=24
| model                          |       size |     params | backend    | ngl | fa | mmap | dio |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -: | ---: | --: | --------------: | -------------------: |
| minimax-m2 230B.A10B IQ4_XS - 4.25 bpw | 113.52 GiB |   228.69 B | CUDA,RPC   |  99 |  1 |    0 |   1 |           pp512 |        577.62 ± 6.70 |
| minimax-m2 230B.A10B IQ4_XS - 4.25 bpw | 113.52 GiB |   228.69 B | CUDA,RPC   |  99 |  1 |    0 |   1 |          pp1024 |       412.20 ± 12.36 |
| minimax-m2 230B.A10B IQ4_XS - 4.25 bpw | 113.52 GiB |   228.69 B | CUDA,RPC   |  99 |  1 |    0 |   1 |          pp2048 |        413.64 ± 5.54 |
| minimax-m2 230B.A10B IQ4_XS - 4.25 bpw | 113.52 GiB |   228.69 B | CUDA,RPC   |  99 |  1 |    0 |   1 |         pp10000 |        370.39 ± 4.64 |
```
skipped the tg/s since there is barely any traffic (4mb here and there)

ggerganov:
> > @dvv47 Do you know if I connect 2 DGX Sparks with a 200Gb/s QSFP cable, is this > implementation going to be compatible? I am planning to buy the cable and test it.
> 
> @ggerganov Yes, it will work with dgx spark, as it is, they use Mellanox ConnectX-7 and Libibverbs api exactly same. Low-latency benefits will be there, but full bandwidth utilization for faster model transfer weight could be optimized even further

I got the QSFP cable today. Will report results when I get some time to try it out.

ggerganov:
Got the two DGX Sparks connected. Not sure I am getting the full 200Gb/s speed, but it should be at least 100Gb/s if I am reading the numbers correctly:

<details>

Cable: https://www.naddod.com/products/102069.html

```
$ ibdev2netdev 
rocep1s0f0 port 1 ==> enp1s0f0np0 (Up)
rocep1s0f1 port 1 ==> enp1s0f1np1 (Down)
roceP2p1s0f0 port 1 ==> enP2p1s0f0np0 (Up)
roceP2p1s0f1 port 1 ==> enP2p1s0f1np1 (Down)
```

```
$ ib_write_bw -d rocep1s0f0 --report_gbits -F
************************************
* Waiting for client to connect... *
************************************
---------------------------------------------------------------------------------------
                    RDMA_Write BW Test
 Dual-port       : OFF		Device         : rocep1s0f0
 Number of qps   : 1		Transport type : IB
 Connection type : RC		Using SRQ      : OFF
 PCIe relax order: ON
 ibv_wr* API     : ON
 CQ Moderation   : 1
 Mtu             : 1024[B]
 Link type       : Ethernet
 GID index       : 3
 Max inline data : 0[B]
 rdma_cm QPs	: OFF
 Data ex. method : Ethernet
---------------------------------------------------------------------------------------
 local address: LID 0000 QPN 0x0178 PSN 0xe286ee RKey 0x18431d VAddr 0x00f2e9affd8000
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:169:254:78:122
 remote address: LID 0000 QPN 0x017a PSN 0x476605 RKey 0x184300 VAddr 0x00e9841df15000
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:169:254:48:241
---------------------------------------------------------------------------------------
 #bytes     #iterations    BW peak[Gb/sec]    BW average[Gb/sec]   MsgRate[Mpps]
 65536      5000             108.30             108.29 		  0.206553
---------------------------------------------------------------------------------------
```

```
# nccl-tests version 2.18.2 nccl-headers=22809 nccl-library=22809
# Collective test starting: all_gather_perf
# nThread 1 nGpus 1 minBytes 17179869184 maxBytes 17179869184 step: 2(factor) warmup iters: 1 iters: 20 agg iters: 1 validation: 1 graph: 0 unalign: 0
#
# Using devices
#  Rank  0 Group  0 Pid  31967 on spark-17ed device  0 [000f:01:00] NVIDIA GB10
#  Rank  1 Group  0 Pid  32456 on spark-a163 device  0 [000f:01:00] NVIDIA GB10
#
#                                                              out-of-place                       in-place          
#       size         count      type   redop    root     time   algbw   busbw  #wrong     time   algbw   busbw  #wrong 
#        (B)    (elements)                               (us)  (GB/s)  (GB/s)             (us)  (GB/s)  (GB/s)         
 17179869184    2147483648     float    none      -1   495564   34.67   17.33       0   396500   43.33   21.66       0
# Out of bounds values : 0 OK
# Avg bus bandwidth    : 19.499 
#
# Collective test concluded: all_gather_perf
#
```

```
$ ib_write_lat -d rocep1s0f0 --report_gbits
************************************
* Waiting for client to connect... *
************************************
---------------------------------------------------------------------------------------
                    RDMA_Write Latency Test
 Dual-port       : OFF		Device         : rocep1s0f0
 Number of qps   : 1		Transport type : IB
 Connection type : RC		Using SRQ      : OFF
 PCIe relax order: OFF
 ibv_wr* API     : ON
 Mtu             : 1024[B]
 Link type       : Ethernet
 GID index       : 3
 Max inline data : 220[B]
 rdma_cm QPs	: OFF
 Data ex. method : Ethernet
---------------------------------------------------------------------------------------
 local address: LID 0000 QPN 0x017a PSN 0x7b2cdb RKey 0x1707c4 VAddr 0x00c31234c02000
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:169:254:78:122
 remote address: LID 0000 QPN 0x017c PSN 0x90bf28 RKey 0x169b59 VAddr 0x00b51d8310f000
 GID: 00:00:00:00:00:00:00:00:00:00:255:255:169:254:48:241
---------------------------------------------------------------------------------------
 #bytes #iterations    t_min[usec]    t_max[usec]  t_typical[usec]    t_avg[usec]    t_stdev[usec]   99% percentile[usec]   99.9% percentile[usec] 
 2       1000          1.65           2.81         1.74     	      1.74        	0.03   		1.86    		2.81   
---------------------------------------------------------------------------------------
```

</details>

dvv101111:
I heard that the network of DGX Spark is a bit complicated. In reality, each port is split into two links with a speed of 100 Gbps. So, in theory, you will be able to set up link bonding and achieve 200 Gpbs over one cable. Or, without bonding, you can get 100 gbps over one port, and at the same time, 100 gbps to a second device through a second port. In the case of this RPC RDMA feature, most profit will come from a stable, low-latency connection. So I expect no difference between 25 Gbps and even 200 Gbps links

ggerganov:
Yup, that's what I figured as well. Added the latency numbers to the previous comment for completeness.

v0l:
**EDIT 3:** Got it working between RTX Pro 6000 (x86_64 host) and DGX Spark (arm64) loading Qwen3.5 122B Q8 (200GB+ model)

**EDIT 2:** I rebased on master which has a bug in rpc-server: fixed in https://github.com/ggml-org/llama.cpp/pull/21030

**EDIT**: MTU issue was first problem, fixed that but having another error now... will update later

Hey just checking out this PR,

<details>
I tried to run rpc-server between DGX Spark -> Linux PC (RTX Pro 6000) but i get an exception running llama-bench

```bash
kieran@spark-95d6:~/git/llama.cpp/build$ bin/llama-bench --rpc 10.100.50.1:50052 \
  -m ~/.cache/huggingface/hub/models--unsloth--Qwen3.5-122B-A10B-GGUF/snapshots/51eab4d59d53f573fb9206cb3ce613f1d0aa392b/Q4_K_M/Qwen3.5-122B-A10B-Q4_K_M-00001-of-00003.gguf -p 2048 -n 256
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 124610 MiB):
  Device 0: NVIDIA GB10, compute capability 12.1, VMM: yes, VRAM: 124610 MiB
RDMA probed: dev=rocep1s0f1 gid=2 qpn=552 inline=316
RDMA activated: qpn=552->260 mtu=4096 rx_depth=24
/home/kieran/git/llama.cpp/ggml/src/ggml-rpc/ggml-rpc.cpp:1125: Remote RPC server crashed or returned malformed response
[New LWP 2814330]
[New LWP 2814326]

This GDB supports auto-downloading debuginfo from the following URLs:
  <https://debuginfod.ubuntu.com>
Enable debuginfod for this session? (y or [n]) [answered N; input not from terminal]
Debuginfod has been disabled.
To make this setting permanent, add 'set debuginfod enabled off' to .gdbinit.
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libhns-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libmthca-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libhfi1verbs-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libirdma-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libmlx5-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libefa-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libocrdma-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libcxgb4-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libvmw_pvrdma-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libqedr-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/liberdma-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libmlx4-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libipathverbs-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libsiw-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/libmana-rdmav34.so
warning: could not find '.gnu_debugaltlink' file for /usr/lib/aarch64-linux-gnu/libibverbs/librxe-rdmav34.so
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".
0x0000f68df09c7b74 in __GI___wait4 (pid=2815007, stat_loc=0x0, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
warning: 30	../sysdeps/unix/sysv/linux/wait4.c: No such file or directory
#0  0x0000f68df09c7b74 in __GI___wait4 (pid=2815007, stat_loc=0x0, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
30	in ../sysdeps/unix/sysv/linux/wait4.c
#1  0x0000f68df13a5a1c in ggml_print_backtrace () from /home/kieran/git/llama.cpp/build/bin/libggml-base.so.0
#2  0x0000f68df13a5bc0 in ggml_abort () from /home/kieran/git/llama.cpp/build/bin/libggml-base.so.0
#3  0x0000f68dee6aa8a0 in ggml_backend_rpc_buffer_set_tensor(ggml_backend_buffer*, ggml_tensor*, void const*, unsigned long, unsigned long) () from /home/kieran/git/llama.cpp/build/bin/libggml-rpc.so.0
#4  0x0000f68df158b790 in llama_model_loader::load_all_data(ggml_context*, std::unordered_map<unsigned int, ggml_backend_buffer*, std::hash<unsigned int>, std::equal_to<unsigned int>, std::allocator<std::pair<unsigned int const, ggml_backend_buffer*> > >&, std::vector<std::unique_ptr<llama_mlock, std::default_delete<llama_mlock> >, std::allocator<std::unique_ptr<llama_mlock, std::default_delete<llama_mlock> > > >*, bool (*)(float, void*), void*) () from /home/kieran/git/llama.cpp/build/bin/libllama.so.0
#5  0x0000f68df15a8d7c in llama_model::load_tensors(llama_model_loader&) () from /home/kieran/git/llama.cpp/build/bin/libllama.so.0
#6  0x0000f68df14f3098 in llama_model_load_from_file_impl(gguf_context*, void (*)(ggml_tensor*, void*), void*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, llama_model_params) () from /home/kieran/git/llama.cpp/build/bin/libllama.so.0
#7  0x0000f68df14f43a0 in llama_model_load_from_file () from /home/kieran/git/llama.cpp/build/bin/libllama.so.0
#8  0x0000acf83a58d9fc in main ()
[Inferior 1 (process 2814325) detached]
Aborted (core dumped)
```

PC side:
```bash
kieran@kieran-x ~/g/l/build (feat-rdma-9493)> bin/rpc-server -H 0.0.0.0 -p 50052 -c
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 97204 MiB):
  Device 0: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition, compute capability 12.0, VMM: yes, VRAM: 97204 MiB

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
WARNING: Host ('0.0.0.0') is != '127.0.0.1'
         Never expose the RPC server to an open network!
         This is an experimental feature and is not secure!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Starting RPC server v3.7.1
  endpoint       : 0.0.0.0:50052
  local cache    : /home/kieran/.cache/llama.cpp/rpc/
Devices:
  CUDA0: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition (97204 MiB, 95458 MiB free)
  transport      : TCP (RDMA auto-negotiate enabled)
Accepted client connection
RDMA probed: dev=mlx5_0 gid=2 qpn=260 inline=316
RDMA activated: qpn=260->552 mtu=1024 rx_depth=24
Client connection closed
Accepted client connection
RDMA probed: dev=mlx5_0 gid=2 qpn=262 inline=316
RDMA activated: qpn=262->553 mtu=1024 rx_depth=24
Client connection closed
Accepted client connection
RDMA probed: dev=mlx5_0 gid=2 qpn=263 inline=316
RDMA activated: qpn=263->554 mtu=1024 rx_depth=24
Client connection closed
Accepted client connection
RDMA probed: dev=mlx5_0 gid=2 qpn=264 inline=316
RDMA activated: qpn=264->555 mtu=1024 rx_depth=24
Client connection closed
Accepted client connection
RDMA probed: dev=mlx5_0 gid=2 qpn=265 inline=316
RDMA activated: qpn=265->556 mtu=1024 rx_depth=24
Client connection closed
Accepted client connection
RDMA probed: dev=mlx5_0 gid=2 qpn=266 inline=316
RDMA activated: qpn=266->557 mtu=1024 rx_depth=24
Client connection closed
```
</details>

ggml/include/ggml-rpc.h
```diff
@@ -7,7 +7,7 @@ extern "C" {
 #endif
 
 #define RPC_PROTO_MAJOR_VERSION    3
```
rgerganov: let's bump to 4.0.0

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -121,6 +193,25 @@ struct rpc_msg_hello_rsp {
     uint8_t patch;
 };
 
+struct rpc_msg_hello_rdma_req {
```
rgerganov: let's abstract the RDMA specifics with something like this:

```
struct rpc_msg_hello_req {
    uint8_t conn_caps[20];
};

struct rpc_msg_hello_rsp {
    uint8_t major;
    uint8_t minor;
    uint8_t patch;
    uint8_t padding;
    uint8_t conn_caps[20];
};
```

also no need to support older versions of the protocol, it's fine making a breaking change as long as we can detect this and log an error telling the user to upgrade

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -414,27 +505,344 @@ static bool recv_data(sockfd_t sockfd, void * data, size_t size) {
...
+        if ((s & 0xFFFFF) == 0 && s > 0) {
+            if (tcp_fd >= 0) {
+                // TCP FD available: use it as the authoritative liveness check.
+                // Abort immediately on peer close; otherwise wait indefinitely
+                // (the connection may be idle between inference requests).
+                if (tcp_peer_closed(tcp_fd)) {
+                    return false;
+                }
+            } else {
+                // No TCP FD: fall back to a hard 30s timeout as safety net.
```
rgerganov: in which cases there is no TCP file descriptor?

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -478,24 +890,77 @@ static bool send_rpc_cmd(const std::shared_ptr<socket_t> & sock, enum rpc_cmd cm
...
     }
     return true;
 }
 
 // RPC client-side implementation
 
-static bool check_server_version(const std::shared_ptr<socket_t> & sock) {
+// Performs HELLO handshake with optional RDMA auto-negotiation.
+// If both sides have RDMA, the socket is upgraded transparently.
+static bool negotiate_hello(const std::shared_ptr<socket_t> & sock) {
```
rgerganov, line 1005: let's try to hide all of the RDMA stuff behind the `socket_t` struct and do something like this:
```
rpc_msg_hello_req request;
rpc_msg_hello_rsp response;
sock->get_caps(&request.conn_caps); // advertise RDMA capabilities (if any) to the remote side
bool status = send_rpc_cmd(sock, RPC_CMD_HELLO, &request, sizeof(request), &response, sizeof(response));
sock->update_caps(response.conn_caps); // activate RDMA if remote side also supports this
```

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -1592,24 +2058,69 @@ rpc_server::~rpc_server() {
...
         return;
     }
-    // the first command sent by the client must be HELLO
     if (cmd != RPC_CMD_HELLO) {
         GGML_LOG_ERROR("Expected HELLO command, update client\n");
         return;
     }
-    if (!recv_msg(sockfd, nullptr, 0)) {
+
+    // Read input_size to determine legacy vs extended HELLO
```
rgerganov: no need to support old versions of the protocol, just log an error saying that client is old if the size of the request doesn't match `sizeof(rpc_msg_hello_req)`

rgerganov (commented):
Please squash the commits and rebase on current master

slavonnet:
Thank you for doing this!

I can check the difference on a pure Infiniband 56g (though on the CPU, there are no GPUs there). In theory, the Infiniband protocol has a 2-8 times lower delay per port, and this should greatly reduce delays and increase performance.

Ethernet (RoCE v2): Latency  ~10–50 µs (after tuning). Jitter - Higher, depends on tuning
InfiniBand: Latency  - ~1–2 µs (native). Jitter - Very low, deterministic

I have Mellonox CX-3 VPI cards and a Mellonox Infiniband switch. Its very chip install if buy it on Ebay :)

By the way, Digital Spark is supposed to support Infiniband port mode. If there is one, then you need to run the OpenSM service on one of the nodes, without it the Infiniband layer will not rise and there will be no link. I also have two Nvidia Spark coming before the end of the year, so I'm really following your progress :)

If you need to build something and run the test, then send the commands to run it (taking into account that I have CPU versions of the servers there)

Mithras:
> Ethernet (RoCE v2): Latency ~10–50 µs

That doesn't sound right. I have mellanox x4 and in eth mode with roce v2 have ~1us latency

ggml/src/ggml-rpc/ggml-rpc.cpp
dvv101111: Garbage, it was my first attempt to switch completely to rdma and close tcp. Now not required and cleared this

ggml/src/ggml-rpc/ggml-rpc.cpp
dvv101111: @rgerganov, now when major version is bumped, older protocol versions are no longer supported. What do you think about enabling RDMA by default on Linux builds? - CMakeLists.txt auto-detects libibverbs and turn GGML_RPC_RDMA=ON when found. It now safe pure runtime check - even if  one side is compiled with RDMA and the other without, both will fall back to TCP

https://github.com/ggml-org/llama.cpp/pull/20590/changes#diff-63f9ed7e7080481865a6d41b3cd9607d4591ff4510c9e715f7a8c37c4c611b9dR23

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -41,6 +51,13 @@ namespace fs = std::filesystem;
 
 static constexpr size_t MAX_CHUNK_SIZE = 1024ull * 1024ull * 1024ull; // 1 GiB
 
+#ifdef GGML_RPC_RDMA
+static constexpr size_t RDMA_CHUNK    = 256 * 1024;   // 256 KiB per send/recv (fits default 8 MiB memlock)
+static constexpr int    RDMA_RX_DEPTH = 24;            // pre-posted recv ring: 24 × 256 KiB = 6 MiB
+static constexpr size_t RDMA_GID_SIZE = 16;            // RoCE GID / IB GID is always 16 bytes
+using rdma_gid_t = std::array<uint8_t, RDMA_GID_SIZE>;
+#endif // GGML_RPC_RDMA
```
rgerganov: move this under the next `#ifdef GGML_RPC_RDMA`

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -115,12 +226,21 @@ static_assert(RPC_CMD_HELLO == 14, "RPC_CMD_HELLO must be always 14");
...
     uint8_t major;
     uint8_t minor;
     uint8_t patch;
+    uint8_t padding;
+    uint8_t conn_caps[RPC_CONN_CAPS_SIZE];
 };
 
+static_assert(sizeof(rpc_msg_hello_req) == 24, "rpc_msg_hello_req must be 24 bytes");
+static_assert(sizeof(rpc_msg_hello_rsp) == 28, "rpc_msg_hello_rsp must be 28 bytes");
+
```
rgerganov: remove these asserts

ggml/src/ggml-rpc/ggml-rpc.cpp
```diff
@@ -1597,25 +2115,51 @@ rpc_server::~rpc_server() {
...
-    if (!send_msg(sockfd, &response, sizeof(response))) {
+
+    if (hello_input_size != sizeof(rpc_msg_hello_req)) {
+        GGML_LOG_ERROR("HELLO request size mismatch (%zu vs %zu) — client needs upgrade to protocol v%d.x\n",
+                       (size_t)hello_input_size, sizeof(rpc_msg_hello_req), RPC_PROTO_MAJOR_VERSION);
+        // drain whatever the old client sent
+        if (hello_input_size > 0 && hello_input_size < 1024) {
+            std::vector<uint8_t> discard(hello_input_size);
+            recv_data(sockfd, discard.data(), hello_input_size);
+        }
```
rgerganov: remove this

docs/backend/RPC-RDMA.md
```diff
@@ -0,0 +1,226 @@
+# llama.cpp RPC with RDMA Transport
```
rgerganov: delete this entire file

tools/rpc/README.md
```diff
@@ -95,6 +95,12 @@ $ bin/rpc-server -c
 
 By default, the cache is stored in the `$HOME/.cache/llama.cpp/rpc` directory and can be controlled via the `LLAMA_CACHE` environment variable.
 
+### RDMA transport
+
+On Linux systems with RoCEv2-capable NICs (e.g. Mellanox ConnectX), the RPC backend can use RDMA instead of TCP for lower latency and higher throughput. Build with `-DGGML_RPC_RDMA=ON` and the transport is negotiated automatically -- no changes to command-line usage are required.
+
+See [docs/backend/RPC-RDMA.md](/docs/backend/RPC-RDMA.md) for setup, hardware requirements, and configuration details.
```
rgerganov: just say that RDMA requires `libibverbs`; you can also put the `apt install` commands for Linux

ggml/src/ggml-rpc/ggml-rpc.cpp
rgerganov: make sense, let's auto enable RDMA if `libibverbs` is detected

rgerganov:
On my testbed (two DGX Sparks connected with QSFP) I get almost identical performance from CUDA-over-RPC and local CUDA backend:

| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |           pp512 |      3113.46 ± 26.16 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |           tg128 |         77.19 ± 0.60 |
| gemma4 ?B Q4_K - Medium        |  17.39 GiB |    30.70 B | CUDA       |  99 |           pp512 |        687.29 ± 8.00 |
| gemma4 ?B Q4_K - Medium        |  17.39 GiB |    30.70 B | CUDA       |  99 |           tg128 |         10.07 ± 0.01 |

build: b1a5fdece (8769)


| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | RPC        |  99 |           pp512 |      3170.49 ± 51.74 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | RPC        |  99 |           tg128 |         76.71 ± 0.18 |
| gemma4 ?B Q4_K - Medium        |  17.39 GiB |    30.70 B | RPC        |  99 |           pp512 |        638.88 ± 4.14 |
| gemma4 ?B Q4_K - Medium        |  17.39 GiB |    30.70 B | RPC        |  99 |           tg128 |         10.35 ± 0.01 |

build: b1a5fdece (8769)

Unfortunately, loading the model with `-sm tensor` over RPC is super slow because we make thousands of `set_tensor` calls with very small chunks of data:

```
...
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38064105, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38064870, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38065635, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38066400, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38067165, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38067930, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38068695, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38069460, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38070225, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38070990, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38071755, size: 765
[set_tensor] buffer: 0xafc800794370, data: 0xe2e99767b300, offset: 38072520, size: 765
...
```

We need to figure out how to improve this.

rgerganov (approved):
I think this is good to go as an initial version for RDMA support. I will prepare a follow up patch which moves all transport related stuff in a separate file (e.g. `transport.cpp`) and clean up the `socket_t` interface.

To summarize the design decisions made here:
 * `socket_t` encapsulates how client-server communication is done
 * partial reads are no longer supported due to RDMA
 * transport capabilities are negotiated as part of `RPC_CMD_HELLO`
 * `rdma_poll()` is doing a busy loop causing 100% CPU usage on one core (not sure how we can avoid this)

dvv101111:
I also want to make a cross-platform Windows/Linux RPC RDMA in the next version. Probably not a very popular setup, but OK for a home lab. 
I already have some proof of concept

ggerganov (approved):

rgerganov:
This is the follow-up refactoring: #21998
It's still WIP but feedback is welcome

segmond:
Update:  This works great.

I finally got to try this.  I bought connect4x cards and connected 2 systems with direct active copper.  The network with RDMA transport is obviously faster than my previous 1gigabit ethernet.    I have half the latency and practical 2.5x the speed.    So it's no surprise to see a slightly faster improvement than with ethernet like everyone else, however.  I think this should be much better.  I have 2 clusters, 1 with 5 gpus, another with 4 gpus.   I should see 2x at least as if all GPUs are on the same system.  I suspect it's because I'm not using GPUdirect and I didn't see anymore talk about it.

I don't know if you all are using nvidia-peer-mem, but RDMA provides GPUdirect across the network if setup correctly.  One needs to setup mlnx ofed or doca ofed.  mlnx ofed is for older cards with end of life support at 2027, where  as doca ofed is for newer cards.      Right now, I'm fighting with the system to get this installed, but when I figure it out I'll report back.  If anyone is doing RDMA, check if you have GPUdirect enabled.  If we are seeing marginal improvement, then I supposed it might be possible that the code needs to be updated to support it.   The data transfer from both servers when going from one GPU to another GPU should bypass system CPU.

"GPUDirect RDMA provides direct communication between NVIDIA GPUs in remote systems. This eliminates the system CPUs and the required buffer copies of data via the system memory, resulting in 10X better performance."    

See https://developer.nvidia.com/gpudirect
https://network.nvidia.com/products/GPUDirect-RDMA/

I bought dell connect4x cards and unfortunately I need their device driver to get GPUdirect and the driver support is only for redhat & suse.  I'm on Ubuntu, so I'm going to get nvidia connect4x card which should make this easier.  I'll update when my new cards arrive.

I decided to run one more experiment before I shut it down.  My prior 2 rigs were a rig of 3090s and a rig of P40/P100s.  I noticed that I saw CUDA graph being disabled in rpc via rig 2.  I thought perhaps it's because the GPUs are a bit too old.  BTW, Nvidia supports GDPdirect for those old gpus too.  I decided to try with a rig of 3090s and rig of 3060s.  I saw noticeable improvement moving from gigabyte to connect4x.

I had to limit to models that could fit on both rigs vram, it has a total of 132gb vram.

Running DeepSeekV4Flash-Q2
gigabit ethernet - 124PP, 24TG, with RDMA 274PP, 32TG
Running Qwen3.5-122B-Q6
gigabit ethernet - 324PP, 25TG, with RDMA 676PP, 43TG

So RDMA definitely for the win if doing RPC.  It would still be worth seeing the gains that can be had with GPUdirect/nvidia-peer-mem

d-shehu:
I see RPC instructions were updated to include RDMA but when I tried it some time ago with Connectx 4 LX I couldn't get it to work. My low-level socket test worked fine for me (sockperf) and I'm using libvma with TCP. But that's not as good as using the primitives libibverbs.

https://github.com/ggml-org/llama.cpp/blob/master/tools/rpc/README.md

rgerganov if at all possible, could you kindly include usage snippet with llama.cpp RPC in the doc. 

Or perhaps an issue with older Mellanox cards? Thanks. 

---

Update:
I rebuilt with the latest code and got it to run with RDMA. It seems a bit unstable since it crashed once in the rpc client with code 143 and ran the 2nd time. 

TG and pp were basically identical to what I got from libvma even though latency was 2.2 usecs vs 5 usecs with libvma. What an odd result ...

PR #15307, opened 2025-08-14
Add OpenVINO backend

wine99:
### Overview

This PR introduces an [OpenVINO backend](https://docs.openvino.ai/2025/index.html) for `llama.cpp`, enabling hardware-accelerated inference on **Intel® CPUs, GPUs, and NPUs**. The backend leverages OpenVINO to deliver optimized inference with the existing llama.cpp GGUF model ecosystem. Enables performance improvements via OpenVINO’s graph compilation and kernel fusion.

* llama.cpp with OpenVINO backend: [Build Instructions](https://github.com/ravi9/llama.cpp/blob/dev_backend_openvino/docs/build.md#openvino)

### Key Features:

* **New backend implementation**
  * Added OpenVINO backend in `ggml/src/ggml-openvino`.
  * Implemented translations for core GGML operations

* **Supported precisions**
  * FP16/BF16 GGUF models supported.
  * Q4_0, Q4_1, Q4_K_M, Q6_K models partially supported. (See notes below)

* **Supported devices**
  * Intel CPUs
  * Intel integrated and discrete GPUs
  * Intel NPUs (requires **UD32+ driver**).

**For NPU: currently prompt processing is slow, a smaller context size is recommended for better performance, e.g., `-c 512`.**

**For llama-bench: `-fa 1` is required.**

### Tested Models

The following models are validated for functionality.

Accuracy and performance are WIP.

* [`Llama-3.2-1B-Instruct-GGUF`](https://huggingface.co/MaziyarPanahi/Llama-3.2-1B-Instruct-GGUF)
* [`Llama-3.1-8B-Instruct`](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct) 
* [`microsoft/Phi-3-mini-4k-instruct-gguf`](https://huggingface.co/microsoft/Phi-3-mini-4k-instruct-gguf)
* [`Qwen/Qwen2.5-1.5B-Instruct-GGUF`](https://huggingface.co/Qwen/Qwen2.5-1.5B-Instruct-GGUF)
* [`Qwen/Qwen3-8B`](https://huggingface.co/Qwen/Qwen3-8B)
* [`openbmb/MiniCPM-1B-sft-bf16`](https://huggingface.co/openbmb/MiniCPM-1B-sft-bf16)
* [`tencent/Hunyuan-7B-Instruct`](https://huggingface.co/tencent/Hunyuan-7B-Instruct)
[...]

PR #20593, opened 2026-03-15
ci : only save openvino caches on github-hosted master

CISC:
cont #15307

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -608,6 +608,7 @@ jobs:
           with:
             key: ubuntu-24-openvino-${{ matrix.variant }}-no-preset-v1
             evict-old-files: 1d
+            save: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
 
         - name: Dependencies
           id: depends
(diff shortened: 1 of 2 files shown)


ggerganov (approved):

CISC:
Hmmm, there's a problem with the toolkit cache as well, think we need to disable it for self-hosted as they will alternately create their own due to cross-os detection.

ggerganov:
@copilot How to disable the ccache and toolkit cache for this job when the runner is self-hosted?

CISC:
Figured it out, seems to work, let's wait for CPU job.

.github/workflows/build.yml
```diff
@@ -617,6 +618,7 @@ jobs:
             sudo apt-get install -y ocl-icd-opencl-dev opencl-headers opencl-clhpp-headers intel-opencl-icd
 
         - name: Use OpenVINO Toolkit Cache
+          if: ${{ !contains(fromJSON(matrix.runner), 'self-hosted') }}
```
ggerganov: This might be cleaner, but not sure if it works (Qwen 3.5 suggested it):

```suggestion
          if: runner.environment != 'self-hosted'
```

CISC: Let's test it after CPU job runs, and add it to both ccache and toolkit cache.

ggerganov: Ok. Btw, here it found a reference, so it should be good: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context

ggerganov (approved):

PR #20551, opened 2026-03-14
vulkan: use graphics queue on AMD

0cc4m:
I'm not sure why, but the graphics queue is slightly faster in tg on AMD than the compute queue, and this also fixes the partial offload issue I fixed in #19976, so the second queue no longer has to be enabled by default. I got the idea from @zedbytes reporting that tg goes up when running with `RADV_DEBUG=nocompute`.

<details>
<summary>AMD RX 9070 XT</summary>

| model                          | size      | params  | ngl | fa | test  | t/s (before)     | t/s (after)      | diff   |
|--------------------------------|-----------|---------|-----|----|-------|------------------|------------------|--------|
| llama 8B Q4_0                  | 4.33 GiB  | 8.03 B  | 20  | 1  | pp512 | 2288.04 ± 2.42   | 2225.76 ± 2.31   | -2.7%  |
| llama 8B Q4_0                  | 4.33 GiB  | 8.03 B  | 20  | 1  | tg128 | 24.33 ± 0.04     | 24.58 ± 0.05     | +1.0%  |
| llama 8B Q4_0                  | 4.33 GiB  | 8.03 B  | 99  | 1  | pp512 | 4886.26 ± 105.08 | 4901.77 ± 102.66 | +0.3%  |
| llama 8B Q4_0                  | 4.33 GiB  | 8.03 B  | 99  | 1  | tg128 | 115.78 ± 0.02    | 121.39 ± 0.02    | +4.8%  |
| qwen3moe 30B.A3B Q2_K - Medium | 10.48 GiB | 30.53 B | 20  | 1  | pp512 | 736.21 ± 9.37    | 735.19 ± 7.51    | -0.1%  |
| qwen3moe 30B.A3B Q2_K - Medium | 10.48 GiB | 30.53 B | 20  | 1  | tg128 | 39.53 ± 0.10     | 40.36 ± 0.21     | +2.1%  |
| qwen3moe 30B.A3B Q2_K - Medium | 10.48 GiB | 30.53 B | 99  | 1  | pp512 | 3383.58 ± 29.26  | 3425.38 ± 28.68  | +1.2%  |
| qwen3moe 30B.A3B Q2_K - Medium | 10.48 GiB | 30.53 B | 99  | 1  | tg128 | 200.45 ± 1.89    | 220.41 ± 1.46    | +10.0% |

</details>

<details>
<summary>AMD Radeon Pro VII</summary>

[...]

PR #20599, opened 2026-03-15
vulkan: allow graphics queue only through env var

### LINKED ISSUE #20597: Misc. bug: 40% tg\s drop in latest b8354 release on AMD card
### Name and Version

b8354

### Operating systems

Windows

llama-server

### Command line

```shell

```

I downloaded latest release b8354  and my tg/s dropped about 40% compared to release b8353 !! I use 2 Radeon cards 9070XT and 6700 XT, i would suggest removing changes from this update until further tested.

I use vulkan

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

0cc4m:
Improve #20551 to fix the reported issues. Only use graphics queue on RADV on larger GPUs.

Fixes #20597

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -5444,11 +5444,14 @@ static vk_device ggml_vk_get_device(size_t idx) {
 
         ggml_vk_load_shaders(device);
 
+        // Only use transfer queue on AMD non-GCN, when the graphics queue is not enabled
+        const bool prefers_transfer_queue = device->vendor_id == VK_VENDOR_ID_AMD && device->architecture != AMD_GCN && !allow_graphics_queue;
+
         if (!device->single_queue) {
             const uint32_t transfer_queue_index = compute_queue_family_index == transfer_queue_family_index ? 1 : 0;
             ggml_vk_create_queue(device, device->transfer_queue, transfer_queue_family_index, transfer_queue_index, { vk::PipelineStageFlagBits::eTransfer }, true);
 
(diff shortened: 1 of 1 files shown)


jeffbolznv (approved):

lemmi:
While graphics queue measurably helps performance on strix halo's 8060s with 40 CUs, desktop performance is unusable now, like with other smaller GPUs, as commented here: https://github.com/ggml-org/llama.cpp/pull/20551#issuecomment-4063085446
I think this either needs to be an option, or a different threshold is necessary.

0cc4m:
I have updated the PR to require manually setting `GGML_VK_ALLOW_GRAPHICS_QUEUE=1` to enable graphics queue use (if a compute queue is available). It seems this is only really safe when headless.

0cc4m:
@zedbytes That means you should now be using `GGML_VK_ALLOW_GRAPHICS_QUEUE=1` instead of `RADV_DEBUG=nocompute`. This is helpful on headless RADV systems or if you don't mind the graphics interruptions.

zedbytes:
@0cc4m agreed, building with master and updated flag now !

zedbytes:
@0cc4m post preliminary testing 
I still find ~14% better performance with `RADV_DEBUG=nocompute` over `GGML_VK_ALLOW_GRAPHICS_QUEUE=1`
wondering if this gain is restricted only to RDNA 4 as other AMD GPUs aren't getting much performance benefit

tsterbak:
Quick question: where is this env variable and other available env configuration documented? I cannot find it anywhere :/

NickM-27:
The findings here seem fairly odd, on linux both my 9060XT and 7900XTX the previous change to use the graphics queue resulted in an increase in TG.

PR #20600, opened 2026-03-15
webui: add model information dialog to router mode

ServeurpersoCom:

webui: add model information dialog to router mode

### 1. Model information button in router mode

In router mode, the "Model Information" dialog was disabled because the
parent process does not load GGUF files. However, the proxy to child
processes already exists (`GET /props?model=<id>`) and the frontend
service `PropsService.fetchForModel()` was already implemented but
unused by the UI.

This wires everything together: an Info button appears on hover for
loaded models, opens the existing dialog, and fetches metadata (context
size, chat template, quantization...) directly from the child process.
No backend changes required. Single-model mode is unchanged.

### 2. "Available models" section header

The model selector showed "Loaded models" and "Favourite models" group
headers, but the remaining unloaded models had no separator, making the
list harder to scan. Adds a symmetric "Available models" header for
visual consistency.

### 3. Remove nested scrollbar from chat template

The chat template Jinja block had a `max-h-120 overflow-y-auto` nested
scrollbar inside the dialog's own scrollable area. This created a
frustrating double-scroll experience, especially on mobile. The template
now expands fully and the dialog's main scrollbar handles overflow.

Close #20547


allozaur (approved):

tools/server/webui/src/lib/components/app/models/ModelsSelectorList.svelte
```diff
@@ -0,0 +1,72 @@
...
+
+{#if groups.loaded.length > 0}
+	<p class={sectionHeaderClass}>Loaded models</p>
+	{#each groups.loaded as item (`loaded-${item.option.id}`)}
+		{@render render(item, true)}
+	{/each}
+{/if}
+
+{#if groups.favourites.length > 0}
+	<p class={sectionHeaderClass}>Favourite models</p>
```
Iniquitatis, line 56: Is "bri'ish" spelling intended here?

Iniquitatis (commented):
.

tools/server/webui/src/lib/components/app/models/ModelsSelectorList.svelte
ServeurpersoCom: Same here, when I saw Favourite instead of Favorite, I saw it was the UK version and I didn't touch it, I should have done it @allozaur , I switch to the US version in a next PR

PR #20601, opened 2026-03-15
server : fix wait in test_cancel_requests() test

ggerganov:
The wait condition in `test_cancel_request()` is a bit flaky, causing failures with the Metal server runs from time to time:

https://github.com/ggml-org/llama.cpp/actions/runs/23113808739/job/67135730576#step:4:22888

Give the slot a bit more time to finish for sure.


CISC (approved):
@ggml-org/llama-server We should perhaps add someone as owner for server tests?

PR #20606, opened 2026-03-15
tools/cli: fix disable reasoning

pwilkin:
Handle `--reasoning off` properly in cli.


aldehir (approved):
This is fine, but it will invert the default for templates that support thinking but default to false. E.g. https://huggingface.co/Qwen/Qwen3.5-0.8B/blob/main/chat_template.jinja

I guess you could run it 3 times, one without the parameter and two with `true`/`false`. Then compare between them. Something for a later PR.

PR #20612, opened 2026-03-16
jinja : add capability check for object args

aldehir:
Add capability check for tool call arguments as objects.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1519,7 +1519,6 @@ static common_chat_params common_chat_templates_apply_jinja(const struct common_
         // map developer to system for all models except for GPT-OSS
         workaround::map_developer_role_to_system(params.messages);
     }
-    workaround::func_args_not_string(params.messages);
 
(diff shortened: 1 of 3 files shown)


CISC (approved):

common/chat.cpp
```diff
@@ -1532,6 +1531,10 @@ static common_chat_params common_chat_templates_apply_jinja(const struct common_
         workaround::requires_non_null_content(params.messages);
     }
 
+    if (tmpl.original_caps().supports_object_arguments) {
+        workaround::func_args_not_string(params.messages);
```
dcousens, line 1535: if the JSON is ill-formed, this results in a 500

```
 {
  "error": {
    "code": 500,
    "message":"Failed to parse tool call arguments as JSON: [json.exception.parse_error.101] parse error at line 1, column 59650: syntax error while parsing value - invalid string: missing closing quote;
...
```

is that intentional? a 500 for bad syntax seems odd

PR #20618, opened 2026-03-16
ggml webgpu: Move to no timeout for WaitAny in graph submission to avoid deadlocks

reeselevine:
Another approach to see if this avoids deadlocks in the llvm-pipe Vulkan backend. After some debugging on the Github CI I've seen cases where it seems to get stuck within the `WaitAny` call itself, even after the timeout nanoseconds have passed, leading me to believe there is a bug within the interface between Dawn and llvm-pipe. Setting timeout to 0 from the WebGPU side creates a busy-wait loop on the ggml side, but hopefully avoids deadlocking in most scenarios, and in practice the busy-wait loop does not occur that often in my tests.


reeselevine:
This should make the `ubuntu-24-webgpu` CI pass more reliably.

@nikhilJain17 @ggerganov

nikhilJain17 (approved):

ggerganov (approved):
Thanks

PR #20620, opened 2026-03-16
kleidiai : fix MUL_MAT support for batched (3D) inputs

jabr:
The supports_op() check incorrectly rejected MUL_MAT operations with 3D inputs (ne[2] > 1), but the actual compute_forward_qx() implementation handles batched inputs correctly via a loop over ne12.

This caused models with Q4_0/Q8_0 weights to crash during graph scheduling when n_seq_max > 1, because weights were placed in KLEIDIAI buffers during loading (tested with 2D inputs) but the runtime used 3D inputs.

~Also relax the buffer check to allow supports_op() to be called during weight loading when src[0]->buffer is NULL.~

Fixes #20608

update: Ah, I just read the contributing guidelines above. I used an AI while tracking this down. I reviewed and tested the changes, and then pared it down to the minimal set of changes needed for the fix, but it would be "considered AI-generated" according to the contribution guidelines.

Feel free to close this PR if needed, but the root cause for the bug and this fix for it both work and make sense to me now.

### DIFF
--- a/ggml/src/ggml-cpu/kleidiai/kleidiai.cpp
+++ b/ggml/src/ggml-cpu/kleidiai/kleidiai.cpp
@@ -1447,9 +1447,8 @@ class extra_buffer_type : ggml::cpu::extra_buffer_type {
         const int slot_total = kleidiai_collect_kernel_chain(op, kernel_chain);
         if ((op->op == GGML_OP_MUL_MAT || op->op == GGML_OP_GET_ROWS) &&
             (op->src[0]->type == GGML_TYPE_Q4_0 || op->src[0]->type == GGML_TYPE_Q8_0) &&
-            op->src[0]->buffer &&
             (ggml_n_dims(op->src[0]) == 2) &&
-            op->src[0]->buffer->buft == ggml_backend_cpu_kleidiai_buffer_type() &&
+            (!op->src[0]->buffer || op->src[0]->buffer->buft == ggml_backend_cpu_kleidiai_buffer_type()) &&
             slot_total > 0) {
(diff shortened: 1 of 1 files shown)


jabr:
These Qwen3.5 (unsloth quants) work for me on aarch64 now (with KleidiAI on q4/q8 ops):
- 9B: Q8_0, UD-Q6_K_XL, UD-Q4_K_XL
- 4B: UD-Q6_K_XL, UD-Q4_K_XL

chaxu01 (commented):
compute_forward_qx() does not fully support 3-D weights (src0->ne[2] > 1). It does iterate over the batch dimension for the LHS / activation side (src1 / dst), but it treats the RHS (the weight) as a single packed buffer shared across the batch — there is no per-batch indexing of the packed RHS.

To make compute_forward_qx() safely handle src0->ne[2] > 1:

- either repack each weight slice during model load into backend (KleidiAI) buffers (so compute_forward_qx can index into a packed buffer per-batch).
- or if the weights are not static implement a runtime/lazy repack for QX that triggers the first time that batched weight is needed (and ensure supports_op() permits the op even if src0->buffer == NULL). This will for sure have some impact on the performance.
- otherwise keep rejecting 3-D weights in supports_op().

jabr:
The patch still has the `(ggml_n_dims(op->src[0]) == 2)` check limiting it to 2D weights. It removes the 2D requirement on inputs, which appears to work fine with the existing `compute_forward_qx()` implementation.

edit: updated to keep the rejection of 4D inputs, only additionally allowing 3D.

chaxu01 (approved):

PR #20623, opened 2026-03-16
Speed up claude-code by prevent adding system message that starts with 'x-anthropic-'

hksdpc255:
Claude code send messages like this now:

```
{
  "system": [
    {"type": "text", "text": "x-anthropic-billing-header: cc_version=2.1.37.0d9; cc_entrypoint=cli; cch=fa690;"},
    {"type": "text", "text": "You are Claude Code, Anthropic's official CLI for Claude."},
    {"type": "text", "text": "...full system prompt..."}
  ],
  "messages": [{"role": "user", "content": "hey"}]
}
```

In the first system message, `cch` change for every request.

This PR will significantly increase the prompt cache hit rate and speed up claude-code.

### DIFF
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -1435,8 +1435,11 @@ json convert_anthropic_to_oai(const json & body) {
             system_content = system_param.get<std::string>();
         } else if (system_param.is_array()) {
             for (const auto & block : system_param) {
-                if (json_value(block, "type", std::string()) == "text") {
-                    system_content += json_value(block, "text", std::string());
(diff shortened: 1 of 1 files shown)


ggerganov:
I think there is a CC environment var that you can set to prevent this.

hksdpc255:
> I think there is a CC environment var that you can set to prevent this.

Yes, there is indeed an environment variable that can make claude-code stop sending that header. However, this likely related to the Anthropic API defination (even though it is not documented in the official API docs). I think the Anthropic API implementation in llama-server should be compatible with it.

PR #20624, opened 2026-03-16
MiroThinker tool call parser

hksdpc255:
The MiroThinker series v1.0–v1.7 (and likely every version before v2.0) uses an MCP-style tool call:
```
<use_mcp_tool>
<server_name>{server_name}</server_name>
<tool_name>{tool_name}</tool_name>
<arguments>
{json_args}
</arguments>
...
</use_mcp_tool>
```
It requires the MCP server name to be included in the system prompt, which makes it impossible for the autoparser to work with it.


models/templates/MiroThinker-v1.jinja
```diff
@@ -0,0 +1,285 @@
+{#- ========== MiroThinker Tool Parsing Macro ========== #}
+
+{%- macro function_name() %}
```
CISC: Even though this works in our jinja I think I would prefer the compliant version where you pass the tool and namespace parameters.

models/templates/MiroThinker-v1.jinja
```diff
@@ -0,0 +1,285 @@
...
+
+{%- macro render_tool(server_name) %}
+    {%- if tool.mt_visited is not defined %}
+        {%- if server_name != ns.last_server %}
+            {{- "\n## Server name: " + server_name + "\n" }}
+            {%- set ns.last_server = server_name %}
+        {%- endif %}
+        {{- "### Tool name: " + function_name() + "\n" }}
+        {{- "Description: " + function_description() + "\n" }}
+        {{- "Input JSON schema: " + (function_parameters() | tojson(ensure_ascii=False)) + "\n" }}
```
CISC: I forgot to ask this earlier, but are you using `ensure_ascii=False` because "everyone else is"? This is actually the default in `transformers` (and here).

common/chat.cpp
```diff
@@ -1278,6 +1278,116 @@ static common_chat_params common_chat_params_init_kimi_k2(const common_chat_temp
...
+        //   </arguments>
+        //   ...
+        //   </use_mcp_tool>
+
+        auto reasoning = extract_reasoning ? p.optional("<think>" + p.reasoning(p.until("</think>")) + "</think>") : p.eps();
+
+        // Tool call markers
+        const std::string SECTION_BEGIN = "<use_mcp_tool>";
+        const std::string SECTION_END   = "</use_mcp_tool>";
+        const std::string CALL_BEGIN    = "<server_name>";
```
pwilkin, line 1318: * `</server_name>` is missing
* you're not using the constants later on

pwilkin (changes requested):
There doesn't seem to be any support for structured outputs.

common/chat.cpp
hksdpc255: Done via d6ec977.

hksdpc255:
> There doesn't seem to be any support for structured outputs.

@pwilkin What does `structured outputs` means? Is there any examples to implement that?

models/templates/MiroThinker-v1.jinja
hksdpc255: Done via 02e6701.

models/templates/MiroThinker-v1.jinja
hksdpc255: Done via aa5a30f.

hksdpc255:
@pwilkin Actually, I have another idea. I could further improve the chat template to recognize formatted tool names from MCP servers (e.g., `mcp__<server_name>__<tool_name>` used by claude-code), which would make the server name more meaningful.

However, this would require, as an example, mapping
`<server_name>name_part_1</server_name>[spaces]<tool_name>name_part_2</tool_name>`
to `mcp__name_part_1__name_part_2`.

How can I implement this kind of custom transformation using the new PEG parser?

models/templates/MiroThinker-v1.jinja
```diff
@@ -0,0 +1,286 @@
...
+            {%- endif %}
+        {%- endif %}
+    {%- endfor %}
+{%- endif %}
+
+{#- ========== MiroThinker Tool Response ========== #}
+
+{%- for message in messages %}
+    {%- if message.role == 'assistant' %}
+    {%- if message.role == 'user' %}
```
CISC: Something is missing here and/or `elif`.

hksdpc255: Sorry fot that

models/templates/MiroThinker-v1.jinja
```diff
@@ -0,0 +1,285 @@
...
+{%- for message in messages %}
+    {%- if message.role == "assistant" %}
+        {%- if message.tool_calls | length == 0 %}
+            {%- set fake_function = namespace(name='fake_name', arguments='{}') %}
+            {%- set fake_function = namespace(function=fake_function) %}
+            {%- set message.tool_calls = [fake_function, fake_function] %}
+        {%- endif %}
+    {%- endif %}
+{%- endfor %}
+{#- ========== Workaround for llama.cpp crashing ========== #}
```
CISC, line 285: Was this fixed? If not, someone should look at it.

CISC (approved):
Otherwise the template LGTM (even though it will not work in immutable sandbox; does not matter for us), remaining changes and approval up to @pwilkin

models/templates/MiroThinker-v1.jinja
hksdpc255: This part add fake functions to assistant messages and it will prevent llama-server from crash.

hksdpc255: The crash seems to occur here:
https://github.com/ggml-org/llama.cpp/blob/740a447fc38e6578a69bb578380c86998d75e286/common/jinja/caps.cpp#L228-L252
It assumes that the message list passed into the chat template is immutable. However, in llama.cpp’s Jinja engine, a reference is passed to the template rather than a copy, which makes it effectively mutable.

aldehir:
> How can I implement this kind of custom transformation using the new PEG parser?

You can create a custom mapper and add another chat format. See chat-peg-parser.cpp, you can likely inherit the one there. 

Definitely an interesting model...

Is it not possible to hardcode the server name to something like "localhost" for better compatibility?

hksdpc255:
> Definitely an interesting model...
> 
> Is it not possible to hardcode the server name to something like "localhost" for better compatibility?

Yes, that is possible. However, this model is fine-tuned to use specific server names such as `tool-python`, `search_and_scrape_webpage`, and `jina_scrape_llm_summary` for tasks like Python/shell execution, web search, and fetching web content.

Mapping everything to a generic server would in pratically require additional reasoning tokens during inference, and the results would not be as good as when using the original format.

pwilkin:
> > How can I implement this kind of custom transformation using the new PEG parser?
> 
> You can create a custom mapper and add another chat format. See chat-peg-parser.cpp, you can likely inherit the one there.

Don't even need to create a custom mapper since for the analysis I made a tagged mapper that can be used out-of-the-box for this :)

See the parser usages in `chat-diff-analyzer.cpp` for examples of extracting fragments.

pwilkin:
> > There doesn't seem to be any support for structured outputs.
> 
> @pwilkin What does `structured outputs` means? Is there any examples to implement that?

Basically this:

```c++
        if (has_response_format) {
            auto response_format = p.rule("response-format", p.content(p.schema(p.json(), "response-format-schema", inputs.json_schema)));
            return ctx.reasoning_parser + p.space() + p.choice({
                p.literal("```json") + p.space() + response_format + p.space() + p.literal("```"),
                response_format
            }) + p.end();
        }
```

hksdpc255:
> > > There doesn't seem to be any support for structured outputs.
> > 
> > 
> > @pwilkin What does `structured outputs` means? Is there any examples to implement that?
> 
> Basically this:
> 
> ```c++
>         if (has_response_format) {
>             auto response_format = p.rule("response-format", p.content(p.schema(p.json(), "response-format-schema", inputs.json_schema)));
>             return ctx.reasoning_parser + p.space() + p.choice({
>                 p.literal("```json") + p.space() + response_format + p.space() + p.literal("```"),
>                 response_format
>             }) + p.end();
>         }
> ```

Oh... I know what you means. I'll implement it later. Let me convert this PR to draft before I fully implement it.

hksdpc255:
@pwilkin Mind taking another look?

hksdpc255:
@pwilkin Is current implementation good for merge now?

pwilkin:
Yeah, almost good - please add proper tests to `test-chat.cpp`.

pwilkin (approved):

aldehir (approved):

pwilkin (approved):

PR #20626, opened 2026-03-16
 [OpenVINO backend] add func is_splited_model()

zhaixuejun1993:
This pull request introduces support for detecting and handling split-model computation graphs within the OpenVINO GGML decoder. The changes add a new mechanism to identify whether a computation graph is a split fragment and propagate this information through the decoder interface, will be used in fallback PATH.

Detection and handling of split-model graphs:

* Added the `is_model_splitted` function to heuristically determine if a `ggml_cgraph` represents a split-model fragment, with detailed checks for node usage and input sources. [[1]](diffhunk://#diff-f03e5dcc8ad94468267d70b7a2ff7926977ecbc1395285e416ee5ad3b13fab78R475-R520) [[2]](diffhunk://#diff-201d067dc64c3a1a706bd3503e92ecf1ad8e1fd40abe9828c0c38efe3cd94ad7R120-R126)
* Modified the logic in `ov_graph_compute_dynamic` to avoid naive computation for split-model graphs by integrating the new detection function.
* Updated the construction of `GgmlOvDecoder` instances to pass the split-model flag, ensuring downstream components are aware of the graph's split status. [[1]](diffhunk://#diff-f03e5dcc8ad94468267d70b7a2ff7926977ecbc1395285e416ee5ad3b13fab78L178-R180) [[2]](diffhunk://#diff-eafa32f3b7b0ba782711f9188ab1b9687264c80b7edf7a1d5d04c77c9ff191aeR48-R56) [[3]](diffhunk://#diff-e0fb1e316593457e151c376b0c6b538e0259e3e4d255f405ad8d221240c7405dR72)
* Adjusted `ov_graph_compute_static` to correctly set the split-model and prefill flags when creating decoder instances.

Decoder interface enhancements:

* Added the `is_splited_model` virtual method to the decoder interface and implemented it in `GgmlOvDecoder`, along with the new member variable `m_model_is_splitted`. [[1]](diffhunk://#diff-3f48eac27705871a1bdab9de1db478acc44e694a1fa300bfd0bb8aef1e1ace23R69-R70) [[2]](diffhunk://#diff-e0fb1e316593457e151c376b0c6b538e0259e3e4d255f405ad8d221240c7405dR179-R182) [[3]](diffhunk://#diff-e0fb1e316593457e151c376b0c6b538e0259e3e4d255f405ad8d221240c7405dR213)


ggml/src/ggml-openvino/utils.cpp
```diff
@@ -470,6 +472,52 @@ enum ggml_status ov_graph_compute_static(ggml_cgraph * cgraph, std::shared_ptr<o
     return GGML_STATUS_SUCCESS;
 }
 
+// Detect whether a cgraph is a split subgraph or not.
+// Step 1 compares each node's recorded use_count with actual fan-out references in node->src.
+// Step 2 verifies that node inputs come from model nodes/weights/leafs; external sources imply split.
+bool is_model_splitted(ggml_cgraph * cgraph) {
```
cavusmustafa, line 492: I was wondering if we can detect if the graph is split in ggml-openvino.cpp when detecting supported/unsupported ops and then we can write this information into runtime context?

ggerganov: Generally no - checking for supported ops can happen at various stages, depending on what the user code (e.g. llama.cpp) does. Basically, the backend should never be aware if a graph is split or not.

I assume you need some workaround logic to get the test suite rolling. That would be OK. In the long term, this logic should not be needed.

zhaixuejun1993:
in dev repo

PR #20630, opened 2026-03-16
fix: correct Granite template tool_call role + add Granite 4.0 Jinja template

jesus-talavera-ibm:
## Summary

- **Fix wrong role name for tool call messages in the C++ Granite template.** The hardcoded template was emitting `<|start_of_role|>assistant_tool_call<|end_of_role|><|tool_call|>` but the model was trained with `<|start_of_role|>assistant<|end_of_role|><|tool_call|>`. This mismatch breaks tool calling, especially at lower quantization levels where the model has less capacity to generalize across formatting variants.
- **Add a Granite 4.0 Jinja template** (`models/templates/ibm-granite-granite-4.0-tiny-preview.jinja`) that uses the correct `available_tools` role for tool definitions instead of `<tools>` XML in the system message, matching the format the model was fine-tuned with.

## Background

IBM Granite 4 models are extremely sensitive to chat template formatting when quantized ([community reports](https://www.reddit.com/r/LocalLLaMA/comments/1nzozpg/)). The root cause is a divergence between the C++ hardcoded template and the Jinja template the model was trained with:

1. **Bug in `llama-chat.cpp`**: The `assistant_tool_call` role was passed through literally as the role name, but the Jinja template maps it to `assistant` with a `<|tool_call|>` prefix.
2. **Missing Granite 4 Jinja template**: The legacy C++ template path cannot inject tool definitions (no `tools` parameter in `llm_chat_apply_template`), so `--jinja` is required for full tool calling support. Adding the Jinja template file ensures it's available when `--jinja` is used.

## Test plan

- [ ] Verify `assistant_tool_call` messages now produce `<|start_of_role|>assistant<|end_of_role|><|tool_call|>...` (not `<|start_of_role|>assistant_tool_call<|end_of_role|>...`)
- [ ] Run `test-chat-template` to confirm existing Granite 3.0 test still passes
- [ ] Test tool calling with Granite 4 model using `--jinja` flag and the new template
- [ ] Test basic chat (no tools) still works with the C++ template path

### DIFF
--- a/models/templates/ibm-granite-granite-4.0-tiny-preview.jinja
+++ b/models/templates/ibm-granite-granite-4.0-tiny-preview.jinja
@@ -1,0 +1,45 @@
+{# Granite 4.0 Chat Template #}
+{# Alias tools -> available_tools for OpenAI-compatible API #}
+{%- if tools and not available_tools -%}
+    {%- set available_tools = tools -%}
+{%- endif -%}
+{%- if messages[0]['role'] == 'system' %}
+     {%- set system_message = messages[0]['content'] %}
+     {%- set loop_messages = messages[1:] %}
+ {%- else %}
+     {%- set system_message = "Knowledge Cutoff Date: April 2025. Today's Date: " + strftime_now('%B %d, %Y') + ". You are Granite, developed by IBM." %}
+     {%- if available_tools %}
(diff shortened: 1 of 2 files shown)


jesus-talavera-ibm:
Opened against wrong repo by mistake. Will re-submit properly.

PR #20634, opened 2026-03-16
tests : remove test-jinja-py

ggerganov:
I think this is no longer needed as we don't use jinja? Suggest to remove it, instead of having to `pip install jinja2`

Noticed this error on the new mac self-hosted runner:

https://github.com/ggml-org/llama.cpp/actions/runs/23115539071/job/67140337984#step:4:6366

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -192,7 +192,8 @@ endif()
 
 llama_build_and_test(test-chat-peg-parser.cpp peg-parser/simple-tokenize.cpp)
 llama_build_and_test(test-jinja.cpp)
-llama_test(test-jinja NAME test-jinja-py ARGS -py LABEL python)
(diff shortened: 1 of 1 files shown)


ngxson:
Yes we still need it, this will run our test cases against python jinja2 for ground-truth.

Although, I think it's better to run this on only one of the workflow, instead of running it everywhere. Not sure if you can look into this @CISC ?

ggerganov:
Ok, it's not a big issue. It is a dependency for `torch` so it will get installed either way. But if at some point we no longer need it, we can remove the tests so that at least the C++ stuff (`ctest` specifically) does not require it.

CISC:
It's why I put it on the `python` label so we have some control on where it's run. It's caught some cases of bad assumptions in jinja behavior, so it's definitely useful to run in our CI.

PR #20639, opened 2026-03-16
kleidiai: add data type check to get_tensor_traits

martin-klacer-arm:
This patch adds a check for the F16 data type into the KleidiAI get_tensor_traits function if input data isn't in the `ggml_backend_cpu_kleidiai_buffer_type` format. 

The path with input data not in the KleidiAI buffer type format is currently only supported for the F16 type, notably causing issues in some specific cases with Q4_0 and Q8_0 types. This fix makes sure only the supported type is accepted.


ggml/src/ggml-cpu/kleidiai/kleidiai.cpp
```diff
@@ -1473,10 +1473,12 @@ class extra_buffer_type : ggml::cpu::extra_buffer_type {
             if (op->src[0]->buffer && op->src[0]->buffer->buft == ggml_backend_cpu_kleidiai_buffer_type()) {
                 return (ggml::cpu::tensor_traits *) op->src[0]->extra;
             } else {
+                if(op->src[0]->type != GGML_TYPE_F16) {
```
ggerganov:
```suggestion
                if (op->src[0]->type != GGML_TYPE_F16) {
```

chaxu01 (approved):

PR #20644, opened 2026-03-16
ggml-cuda: Add NVFP4 dp4a kernel

michaelw9999:
This PR brings in the initial plumbing for basic CUDA support for NVFP4 - it includes one NVFP4xQ8_1 dp4a kernel.  MMA or Blackwell kernels are not included here and were kept out for a separate PR.  

`vec_dot_mma` is is linked up the dp4a kernel so it still runs dp4a even when BLACKWELL_MMA_AVAILABLE.
There is a branch for NVFP4 in `mmq_write_back_mma` to push it back to the dp4a layout which will be removed when the MMA kernel is wired in.

It was tuned to bring as much performance as possible for DP4A. Comparisons below.  
**CPU vs DP4A**

| Model | CPU (pp64) | DP4A (pp64) | Speedup | CPU (tg16) | DP4A (tg16) | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| Qwen3.5-0.8B | 59.63 | 3557.08 | **59.65x** | 32.33 | 329.60 | **10.19x** |
| Qwen3.5-27B | 1.27 | 594.14 | **467.83x** | 1.08 | 52.65 | **48.75x** |

**pp512 / tg128**

| Model | pp512 t/s | tg128 t/s |
|---|---:|---:|
| Qwen3-4B |  9031.22 | 271.69 |
| Qwen3-8B | 4909.27 | 175.93 |
| Qwen3.5-27B| 1482.89 | 63.47 |
| Qwen3.5-0.8B | 25596.92 | 388.12 |
| Qwen3.5-0.8B-Q4_K_M | 38339.57 | 521.69 |

AI assistance was used in refactoring and writing some of this code.  Each line has been scrutinized and this was hand-edited to be as neat and as minimal as possible.  Test-backend-ops passes; CPU<>GPU parity was tested with a separate tool and is exact across multiple tile sizes, and kld/ppl was verified on several models and is as expected (tested on GPU only, CPU too slow).

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -932,6 +932,13 @@ struct ggml_cuda_type_traits<GGML_TYPE_MXFP4> {
 };
 
 template<>
+struct ggml_cuda_type_traits<GGML_TYPE_NVFP4> {
+    static constexpr int qk = QK_NVFP4;
+    static constexpr int qr = QR_NVFP4;
+    static constexpr int qi = QI_NVFP4;
+};
+
+template<>
 struct ggml_cuda_type_traits<GGML_TYPE_Q2_K> {
     static constexpr int qk = QK_K;
     static constexpr int qr = QR2_K;
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -799,54 +799,14 @@ static __device__ __forceinline__ float ggml_cuda_e8m0_to_fp32(uint8_t x) {
 #endif // CUDART_VERSION >= 12050
 }
 
-__device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
-#if defined(__CUDA_ARCH__)
-#if !defined(GGML_USE_HIP) && !defined(GGML_USE_MUSA) && defined(CUDART_VERSION) && CUDART_VERSION >= 12050 // This matches cuda_fp8.h's version gate.
-    // This uses the same fp8 conversion that __nv_fp8_e4m3 uses internally.
-    __half h = __half(__nv_cvt_fp8_to_halfraw((__nv_fp8_storage_t) x, __NV_E4M3));
-    unsigned short hb = __half_as_ushort(h) - 0x0400u; // Built in 0.5f op.
-    float f = __half2float(__ushort_as_half(hb));
-
-    // __nv_fp8_e4m3 is signed but UE4M3
-    if (x == 0u || x == 0x7Fu) { // force 0x7F and 0x00 to 0.0f
-        f = 0.0f;
-    }
-
-    return f;
+static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) { 
+#if CUDART_VERSION >= 11080 || defined(GGML_USE_HIP) 
+    const uint32_t bits = x * (x != 0x7F && x != 0xFF); // Convert NaN to 0.0f to match CPU implementation. 
+    const __nv_fp8_e4m3 xf = *reinterpret_cast<const __nv_fp8_e4m3 *>(&bits); 
+    return static_cast<float>(xf) / 2; 
 #else
-    if (x == 0u || x == 0x7Fu) {
-        return 0.0f;
-    }
-
-    const uint32_t exp  = x >> 3;
-    const uint32_t mant = x & 0x7u;
-    uint32_t bits;
-
-    if (exp != 0u) {
-        bits = ((exp + 119u) << 23) | (mant << 20);
-    } else {
-        uint32_t p;
-        if (mant < 2u) {
-            p = 0u;
-        } else if (mant < 4u) {
-            p = 1u;
-        } else {
-            p = 2u;
-        }
-
-        const uint32_t r = mant - (1u << p);
-        const uint32_t exp32  = p + 117u;
-        const uint32_t mant32 = r << (23u - p);
-        bits = (exp32 << 23) | mant32;
-    }
-
-    float f;
-    memcpy(&f, &bits, sizeof(f)); // 0.5f is baked in.
-    return f;
-#endif
-#else
-    return ggml_ue4m3_to_fp32(x);
-#endif
+    NO_DEVICE_CODE;
+#endif // CUDART_VERSION >= 11080 || defined(GGML_USE_HIP)
 }
 
 __device__ __forceinline__ uint8_t ggml_cuda_float_to_fp4_e2m1(float x, float e) {
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -800,13 +800,13 @@ static __device__ __forceinline__ float ggml_cuda_e8m0_to_fp32(uint8_t x) {
 }
 
 static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
-#if CUDART_VERSION >= 11080 || defined(GGML_USE_HIP)
+#ifdef FP8_AVAILABLE
     const uint32_t bits = x * (x != 0x7F && x != 0xFF); // Convert NaN to 0.0f to match CPU implementation.
     const __nv_fp8_e4m3 xf = *reinterpret_cast<const __nv_fp8_e4m3 *>(&bits);
     return static_cast<float>(xf) / 2;
 #else
     NO_DEVICE_CODE;
-#endif // CUDART_VERSION >= 11080 || defined(GGML_USE_HIP)
+#endif // FP8_AVAILABLE
 }
 
 __device__ __forceinline__ uint8_t ggml_cuda_float_to_fp4_e2m1(float x, float e) {
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -799,11 +799,11 @@ static __device__ __forceinline__ float ggml_cuda_e8m0_to_fp32(uint8_t x) {
 #endif // CUDART_VERSION >= 12050
 }
 
-static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) { 
-#if CUDART_VERSION >= 11080 || defined(GGML_USE_HIP) 
-    const uint32_t bits = x * (x != 0x7F && x != 0xFF); // Convert NaN to 0.0f to match CPU implementation. 
-    const __nv_fp8_e4m3 xf = *reinterpret_cast<const __nv_fp8_e4m3 *>(&bits); 
-    return static_cast<float>(xf) / 2; 
+static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
+#if CUDART_VERSION >= 11080 || defined(GGML_USE_HIP)
+    const uint32_t bits = x * (x != 0x7F && x != 0xFF); // Convert NaN to 0.0f to match CPU implementation.
+    const __nv_fp8_e4m3 xf = *reinterpret_cast<const __nv_fp8_e4m3 *>(&bits);
+    return static_cast<float>(xf) / 2;
 #else
     NO_DEVICE_CODE;
 #endif // CUDART_VERSION >= 11080 || defined(GGML_USE_HIP)
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -799,6 +799,56 @@ static __device__ __forceinline__ float ggml_cuda_e8m0_to_fp32(uint8_t x) {
 #endif // CUDART_VERSION >= 12050
 }
 
+__device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
+#if defined(__CUDA_ARCH__)
+#if !defined(GGML_USE_HIP) && !defined(GGML_USE_MUSA) && defined(CUDART_VERSION) && CUDART_VERSION >= 12050 // This matches cuda_fp8.h's version gate.
+    // This uses the same fp8 conversion that __nv_fp8_e4m3 uses internally.
+    __half h = __half(__nv_cvt_fp8_to_halfraw((__nv_fp8_storage_t) x, __NV_E4M3));
+    unsigned short hb = __half_as_ushort(h) - 0x0400u; // Built in 0.5f op.
+    float f = __half2float(__ushort_as_half(hb));
+
+    // __nv_fp8_e4m3 is signed but UE4M3
+    if (x == 0u || x == 0x7Fu) { // force 0x7F and 0x00 to 0.0f
+        f = 0.0f;
+    }
+
+    return f;
+#else
+    if (x == 0u || x == 0x7Fu) {
+        return 0.0f;
+    }
+
+    const uint32_t exp  = x >> 3;
+    const uint32_t mant = x & 0x7u;
+    uint32_t bits;
+
+    if (exp != 0u) {
+        bits = ((exp + 119u) << 23) | (mant << 20);
+    } else {
+        uint32_t p;
+        if (mant < 2u) {
+            p = 0u;
+        } else if (mant < 4u) {
+            p = 1u;
+        } else {
+            p = 2u;
+        }
+
+        const uint32_t r = mant - (1u << p);
+        const uint32_t exp32  = p + 117u;
+        const uint32_t mant32 = r << (23u - p);
+        bits = (exp32 << 23) | mant32;
+    }
+
+    float f;
+    memcpy(&f, &bits, sizeof(f)); // 0.5f is baked in.
+    return f;
+#endif
+#else
+    return ggml_ue4m3_to_fp32(x);
+#endif
+}
+
 __device__ __forceinline__ uint8_t ggml_cuda_float_to_fp4_e2m1(float x, float e) {
     const uint8_t sign_bit = (x < 0.0f) << 3;
     float         ax       = fabsf(x) * e;
--- a/ggml/src/ggml-cuda/convert.cu
+++ b/ggml/src/ggml-cuda/convert.cu
@@ -621,12 +621,12 @@ template <typename dst_t>
 static __global__ void dequantize_block_nvfp4(
         const void * __restrict__ vx,
         dst_t * __restrict__ yy,
-        const int64_t k) {
+        const int64_t ne) {
     const int64_t i = blockIdx.x;
     const int     tid = threadIdx.x;
 
     const int64_t base = i * QK_NVFP4;
-    if (base >= k) {
+    if (base >= ne) {
         return;
     }
 
--- a/ggml/src/ggml-cuda/convert.cu
+++ b/ggml/src/ggml-cuda/convert.cu
@@ -636,7 +636,7 @@ static __global__ void dequantize_block_nvfp4(
     const int sub = tid / (QK_NVFP4_SUB / 2);
     const int j = tid % (QK_NVFP4_SUB / 2);
 
-    const float d = ggml_cuda_cast<float>(ue4m3{xb.d[sub]});
+    const float d = ggml_cuda_ue4m3_to_fp32(xb.d[sub]);
     const uint8_t q = xb.qs[sub * (QK_NVFP4_SUB / 2) + j];
 
     const int64_t y0 = base + sub * QK_NVFP4_SUB + j;
--- a/ggml/src/ggml-cuda/convert.cu
+++ b/ggml/src/ggml-cuda/convert.cu
@@ -617,6 +617,45 @@ static void dequantize_row_mxfp4_cuda(const void * vx, dst_t * y, const int64_t 
     dequantize_block_mxfp4<<<nb, 32, 0, stream>>>(vx, y);
 }
 
+template <typename dst_t>
+static __global__ void dequantize_block_nvfp4(
+        const void * __restrict__ vx,
+        dst_t * __restrict__ yy,
+        const int64_t k) {
+    const int64_t i = blockIdx.x;
+    const int     tid = threadIdx.x;
+
+    const int64_t base = i * QK_NVFP4;
+    if (base >= k) {
+        return;
+    }
+
+    const block_nvfp4 * x = (const block_nvfp4 *) vx;
+    const block_nvfp4 & xb = x[i];
+
+    const int sub = tid / (QK_NVFP4_SUB / 2);
+    const int j = tid % (QK_NVFP4_SUB / 2);
+
+    const float d = ggml_cuda_cast<float>(ue4m3{xb.d[sub]});
+    const uint8_t q = xb.qs[sub * (QK_NVFP4_SUB / 2) + j];
+
+    const int64_t y0 = base + sub * QK_NVFP4_SUB + j;
+    const int64_t y1 = y0 + QK_NVFP4_SUB / 2;
+
+    yy[y0] = ggml_cuda_cast<dst_t>(d * kvalues_mxfp4[q & 0x0F]);
+    yy[y1] = ggml_cuda_cast<dst_t>(d * kvalues_mxfp4[q >> 4]);
+}
+
+template <typename dst_t>
+static void dequantize_row_nvfp4_cuda(
+        const void * vx,
+        dst_t * y,
+        const int64_t k,
+        cudaStream_t stream) {
+    GGML_ASSERT(k % QK_NVFP4 == 0);
+    const int nb = k / QK_NVFP4;
+    dequantize_block_nvfp4<<<nb, 32, 0, stream>>>(vx, y, k);
+}
 template <typename src_t, typename dst_t>
 static __global__ void convert_unary(
         const void * __restrict__ vx, dst_t * __restrict__ y, const int64_t ne00, const int64_t ne01,
@@ -715,6 +754,8 @@ to_fp16_cuda_t ggml_get_to_fp16_cuda(ggml_type type) {
             return dequantize_row_iq3_s_cuda;
         case GGML_TYPE_MXFP4:
             return dequantize_row_mxfp4_cuda;
+        case GGML_TYPE_NVFP4:
+            return dequantize_row_nvfp4_cuda;
         case GGML_TYPE_F32:
             return convert_unary_cont_cuda<float>;
         case GGML_TYPE_BF16:
@@ -766,6 +807,8 @@ to_fp32_cuda_t ggml_get_to_fp32_cuda(ggml_type type) {
             return dequantize_row_iq3_s_cuda;
         case GGML_TYPE_MXFP4:
             return dequantize_row_mxfp4_cuda;
+        case GGML_TYPE_NVFP4:
+            return dequantize_row_nvfp4_cuda;
         case GGML_TYPE_F16:
             return convert_unary_cont_cuda<half>;
         case GGML_TYPE_BF16:
--- a/ggml/src/ggml-cuda/convert.cuh
+++ b/ggml/src/ggml-cuda/convert.cuh
@@ -62,8 +62,8 @@ template<typename dst_t, typename src_t>
 #else
         return {x.x, x.y};
 #endif // GGML_USE_HIP
-    } else if constexpr (std::is_same_v<src_t, ue4m3>) {
-#if defined(__CUDA_ARCH__)
+    } else if constexpr (std::is_same_v<src_t, ue4m3> && std::is_same_v<dst_t, float>) {
+    #if defined(__CUDA_ARCH__)
 #if !defined(GGML_USE_HIP) && !defined(GGML_USE_MUSA) && defined(CUDART_VERSION) && CUDART_VERSION >= 12050 // This matches cuda_fp8.h's version gate.
         // This uses the same fp8 conversion that __nv_fp8_e4m3 uses internally.
         __half h = __half(__nv_cvt_fp8_to_halfraw((__nv_fp8_storage_t) x.x, __NV_E4M3));
--- a/ggml/src/ggml-cuda/convert.cuh
+++ b/ggml/src/ggml-cuda/convert.cuh
@@ -31,6 +31,10 @@ to_fp32_nc_cuda_t ggml_get_to_fp32_nc_cuda(ggml_type type);
 to_fp16_nc_cuda_t ggml_get_to_fp16_nc_cuda(ggml_type type);
 to_bf16_nc_cuda_t ggml_get_to_bf16_nc_cuda(ggml_type type);
 
+struct ue4m3 {
+    uint8_t x;
+};
+
 template<typename dst_t, typename src_t>
  __host__ __device__ inline dst_t ggml_cuda_cast(src_t x) {
     if constexpr (std::is_same_v<dst_t, src_t>) {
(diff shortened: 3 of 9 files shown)


ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -826,6 +837,41 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+    for (int i0 = 0; i0 < mmq_y; i0 += rows_per_warp * nwarps) {
+        int i = i0 + threadIdx.y * rows_per_warp + row_in_warp;
+
+        if constexpr (need_check) {
+            i = min(i, i_max);
+        }
+
+        const block_nvfp4 * bxi = (const block_nvfp4 *) x + kb0 + i * stride + kbx;
+        block_nvfp4_mmq & x_row = x_rows[i];
+        x_row.sc4_u32[kbx] = (uint32_t) get_int_b1(bxi->d, 0);
```
am17an: `get_int_b1` assumes 1 byte alignment, I think you can do `get_int_b4` here

ggml/src/ggml-cuda/vecdotq.cuh
```diff
@@ -322,6 +322,39 @@ static __device__ __forceinline__ float vec_dot_mxfp4_q8_1(
...
+
+    const block_nvfp4 * bq4 = (const block_nvfp4 *) vbq + kbx;
+    float sum = 0.0f;
+#pragma unroll
+    for (int i = 0; i < VDR_NVFP4_Q8_1_MMVQ/2; i++) {
+        const int iqs0 = iqs + 2*i;
+        const int iqs1 = iqs0 + 1;
+        const int is = iqs0 >> 1;
+        const int2 v0 = get_int_from_table_16(get_int_b1(bq4->qs, iqs0), kvalues_mxfp4);
+        const int2 v1 = get_int_from_table_16(get_int_b1(bq4->qs, iqs1), kvalues_mxfp4);
```
JohannesGaessler:
```suggestion
        const int2 v0 = get_int_from_table_16(get_int_b4(bq4->qs, iqs0), kvalues_mxfp4);
        const int2 v1 = get_int_from_table_16(get_int_b4(bq4->qs, iqs1), kvalues_mxfp4);
```

The NVFP4 type should be aligned to at least 4 bytes even without any repacking.

ggml/src/ggml-cuda/vecdotq.cuh
```diff
@@ -322,6 +322,39 @@ static __device__ __forceinline__ float vec_dot_mxfp4_q8_1(
...
+        const block_q8_1 * bq8 = bq8_1 + (is >> 1);
+        const int i8 = ((is & 1) << 2);
+
+        int sumi = ggml_cuda_dp4a(v0.x, get_int_b1(bq8->qs, i8 + 0), 0);
+        sumi = ggml_cuda_dp4a(v0.y, get_int_b1(bq8->qs, i8 + 2), sumi);
+        sumi = ggml_cuda_dp4a(v1.x, get_int_b1(bq8->qs, i8 + 1), sumi);
+        sumi = ggml_cuda_dp4a(v1.y, get_int_b1(bq8->qs, i8 + 3), sumi);
```
JohannesGaessler:
```suggestion
        int sumi = ggml_cuda_dp4a(v0.x, get_int_b4(bq8->qs, i8 + 0), 0);
        sumi = ggml_cuda_dp4a(v0.y, get_int_b4(bq8->qs, i8 + 2), sumi);
        sumi = ggml_cuda_dp4a(v1.x, get_int_b4(bq8->qs, i8 + 1), sumi);
        sumi = ggml_cuda_dp4a(v1.y, get_int_b4(bq8->qs, i8 + 3), sumi);
```

q8_1 is aligned to 4 bytes.

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -3158,6 +3272,10 @@ template<ggml_type type, int mmq_x, int mmq_y, bool need_check>
 static __device__ __forceinline__ void mmq_write_back_mma(
         const float * __restrict__ sum, const int * __restrict__ ids_dst, float * __restrict__ dst,
         const int stride, const int i_max, const int j_max) {
+    if constexpr (type == GGML_TYPE_NVFP4) {
+        mmq_write_back_dp4a<mmq_x, mmq_y, need_check>(sum, reinterpret_cast<const int32_t *>(ids_dst), dst, stride, i_max, j_max);
+        return;
+    }
```
JohannesGaessler: Please add a comment clarifying that this is temporary.

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -826,6 +837,41 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
 
+template <int mmq_y, bool need_check>
+static __device__ __forceinline__ void load_tiles_nvfp4(const char * __restrict__ x,
+                                                        int * __restrict__ x_tile,
+                                                        const int kb0,
+                                                        const int i_max,
+                                                        const int stride) {
+    constexpr int nwarps = mmq_get_nwarps_device();
+    constexpr int warp_size = ggml_cuda_get_physical_warp_size();
+    block_nvfp4_mmq * x_rows = reinterpret_cast<block_nvfp4_mmq *>(x_tile);
```
JohannesGaessler: Did you check NSight systems when it comes to shared memory bank conflicts? The `__dp4a` implementations generally need a stride with `% 128 == 4` in terms of bytes but this is `% 128 == 16` so there should be a 4-way shared memory bank conflict.

ggml/src/ggml-cuda/convert.cu
```diff
@@ -617,6 +617,49 @@ static void dequantize_row_mxfp4_cuda(const void * vx, dst_t * y, const int64_t
...
+
+    const int64_t y0 = base + sub * QK_NVFP4_SUB + j;
+    const int64_t y1 = y0 + QK_NVFP4_SUB / 2;
+
+    if (y0 < k) {
+        yy[y0] = ggml_cuda_cast<dst_t>(d * kvalues_mxfp4[q & 0x0F]);
+    }
+    if (y1 < k) {
+        yy[y1] = ggml_cuda_cast<dst_t>(d * kvalues_mxfp4[q >> 4]);
+    }
```
JohannesGaessler: Remove these checks and instead add a `GGML_ASSERT` in host code to ensure that `k` is divisible by the block size.

ggml/src/ggml-cuda/common.cuh
```diff
@@ -799,6 +799,48 @@ static __device__ __forceinline__ float ggml_cuda_e8m0_to_fp32(uint8_t x) {
 #endif // CUDART_VERSION >= 12050
 }
 
+static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
```
JohannesGaessler, line 802: Why are you adding this function instead of using the [built-in type casting functionality](https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/struct____nv__fp8__e4m3.html#struct____nv__fp8__e4m3)?

JohannesGaessler (commented):
For the initial PR, add only MMVQ and dequantization support. MMQ support should be in a follow-up PR.

ggml/src/ggml-cuda/mmq.cuh
michaelw9999: Yes, I did.  There was just a small conflict but I had assumed it was worth the trade off as I had got much higher speeds from where I had started. (500 -> 8000)  But 128==4 is definitely better!
With Qwen3 4B (nearly identical numbers across models)
Before: `%128==16`:
``` 
ld_conflicts_per_wavefront=0.125
st_conflicts_per_wavefront=0.000
Avg pp512 8621, tg128 264
```
After: `%128==4` and padded to ` ( 33 * 4 ) = 132 bytes `
```
ld_conflicts_per_wavefront=0.022
st_conflicts_per_wavefront=0.000
Avg pp512 8704, tg128 283
```
Prefill tiny bit better but definitely a nice boost in tokengen.

ggml/src/ggml-cuda/common.cuh
michaelw9999: 1) Primarily for correctness:
The host function `ggml_ue4m3_to_fp32` (`ggml-impl.h:501`) is correct and follows NVFP4 semantics.  NVFP4's use of ue4m3 does not follow the same rules as standard encoded E4M3.  So I matched `ggml_ue4m3_to_fp32`'s exact semantics so the CUDA side would be aligned in agreement.  Using __nv_fp8_e4m3 would not only be different, but it would not be correct either.
 2)  `__nv_fp8_e4m3` is signed but UE4M3is  unsigned. 0xFF is valid in E4M3 as NaN but it is not valid representation in UE4m3.  So in this function (just like the host's) we override `0x7F` (NaN)  or `0x00` and force to 0.0f.   The native wrapper does not do this.
3) It's still very fast:
`__half(__nv_cvt_fp8_to_halfraw((__nv_fp8_storage_t) x, __NV_E4M3))` came from `__nv_fp8_e4m3`'s own call in `cuda_fp8.hpp`.  
4) In `ggml_ue4m3_to_fp32`  it returns with `return raw * 0.5f;`  Here in this function, we get a free speed optimization win during the conversion itself:  subtracting 0x0400  before doing `__half2float(x)`  bypasses the need for a separate *0.5f; step altogether. 
5) It provides a fallback if Fp8 CUDA is not available:
 I used `CUDART_VERSION >= 12050` in the check because that matches `cuda.h`'s definition:
 ```
#if CUDART_VERSION >= 12050
#include <cuda_fp8.h>
#endif // CUDART_VERSION >= 12050
```
The fallback is equivalent functionally to `ggml_ue4m3_to_fp32`'s but increases prefill ~500 tk/s by turning the entire thing into integer ops and  completely removing `ldexpf()`.

So, using this function as written, instead of `__nv_fp8_e4m3` alone gives flexibility, correctly handles the conversion keeping in agreement with the CPU side, and it can be easily adapted in the future if necessary without needing to change any kernels that call it.

**Comment by michaelw9999** [comment]
@JohannesGaessler I  found an issue with overflow inf output on the default cuBLAS fallback and have been investigating a lot of the detail. I could open a new bug if you think that is the best to discuss further. The root cause is the the folding of the F32 global scale in llama-graph is multiplied after matmul instead of into the weight directly.  The error difference is tiny and it does not show up on CPU math because that error is tiny and built into the Q8 error.  The accumulated difference ends up with unscaled dot -71785.6172 exceeding the half max of 65504.  Changing cublas and convert to force out of use_fp16 and to use fp32 and bypass f16 everywhere solves the inf issue but that seems like a bandaid workaround instead of getting it perfect.  Unsure if it changes performance in any way as I have not gone far enough to try it.  Thoughts?

**Comment by JohannesGaessler** [comment]
Fixes for numerical issues should be in a follow-up PR. The cuBLAS API allows user code to specify a constant factor by which to multiply the result - depending on how that factor is applied it could fix the issue.

**Comment by michaelw9999** [comment]
> Fixes for numerical issues should be in a follow-up PR. The cuBLAS API allows user code to specify a constant factor by which to multiply the result - depending on how that factor is applied it could fix the issue.

@JohannesGaessler Finished most of the defective work.  Just using `ggml_mul_mat_id_set_prec` or setting F32 everywhere isn't enough to really fix the root cause of the numerical issue completely. To fix it properly _just for CUDA_ is a band-aid that adds complexity and divergence.  I will have more details soon.

File: ggml/src/ggml-cuda/common.cuh (hunk 7 above)
Comment (JohannesGaessler) [reply ggml/src/ggml-cuda/common.cuh:802]: 1. Make this code part of `ggml_cuda_cast` in `convert.cuh`.
2. If the only purpose is correct handling of NaNs, use the built-in conversion in the fallback branch rather than implementing manual bit shifts and such. Afterwards, apply a fixup that zeroes the return value for the case that would not be handled correctly.
3. Document with comments in the code why it is implemented this way rather than just a naive conversion.

**Comment by michaelw9999** [comment]
Moved conversion to cuda_cast and finished the new rewiring with the new get_scale API functions from  #20845 , it's working better than ever now! 👍

**Review by JohannesGaessler — CHANGES_REQUESTED** [review]
Remove the ggml changes.

**Comment by am17an** [comment]
Please also remove the fusion code. F16 accumulator is not a priority to fix in the CUDA code. You can add a `GGML_LOG_WARN` when cublas + nvfp4 is used with f16 compute.

**Comment by michaelw9999** [comment]
> Please also remove the fusion code. F16 accumulator is not a priority to fix in the CUDA code. You can add a `GGML_LOG_WARN` when cublas + nvfp4 is used with f16 compute.

@am17an I've got that done and all set with both the fusion code and tensor mul code removed and back to baseline.  I re-tested full kld back to back on Nemotron3 Nano-30B on cublasF32 only, cubasF16 is inf.  Just tiny reduced precision otherwise with the scale as before, but not enough to matter with forced F32.  NVFP4 vs Q4_K beats on all ppl/kld/ln/max kld/etc which is great so I believe its math is correct. We'll want to just keep F16 completely disabled since it causes inf.  Then no need for any error or logging.  Are you OK with that?

File: ggml/src/ggml-cuda/common.cuh [hunk 8]
```diff
@@ -1417,11 +1417,14 @@ struct ggml_cuda_mm_fusion_args_host {
...
     const ggml_tensor * gate_bias = nullptr;
+    const ggml_tensor * nvfp4_t_s = nullptr;
     ggml_glu_op glu_op;
 };
 struct ggml_cuda_mm_fusion_args_device {
     const void * x_bias = nullptr;
     const void * gate = nullptr;
     const void * gate_bias = nullptr;
+    const void * nvfp4_t_s = nullptr;
+    int32_t nvfp4_t_s_ne = 0;
```
am17an: This is not required anymore?

am17an:
I don't think you have, or you haven't pushed.

michaelw9999:
I didn't push yet.  I can soon.1 Just was verifying hard enforce F32 for
all.

am17an:
Yes, if this is what you mean: for the cuBLAS path only, if we have nvfp4 type we use f32 accumulator unconditionally. Everything else in this PR should just be adding the dp4a instructions for nvfp4

xkmire:
Am I misunderstanding something, when I try to build this against latest main?

gh pr checkout 20644

cmake -B build -DCMAKE_BUILD_TYPE=Release \
  -DGGML_CUDA=ON \
  -DGGML_SCHED_MAX_COPIES=1 \
  -DGGML_CUDA_IQK_FORCE_BF16=1 \
  -DCMAKE_CUDA_HOST_COMPILER=/usr/bin/gcc-14 \
  -DGGML_NATIVE=ON \
-DCMAKE_CUDA_ARCHITECTURES="120a-real"

cmake --build build --config Release --clean-first -j 32

[  6%] Building CUDA object ggml/src/ggml-cuda/CMakeFiles/ggml-cuda.dir/ggml-cuda.cu.o/home/mir/services/llama.cpp/llama.cpp/ggml/src/ggml-cuda/ggml-cuda.cu

(1299): error: identifier "ggml_mul_mat_get_scale" is undefined
const ggml_tensor * nvfp4_t_s_tensor = src0->type == GGML_TYPE_NVFP4 ? ggml_mul_mat_get_scale(dst) : nullptr;

 followed by several other places with same issue

michaelw9999:
> Am I misunderstanding something, when I try to build this against latest main?
> 
> gh pr checkout 20644
> 
> cmake -B build -DCMAKE_BUILD_TYPE=Release -DGGML_CUDA=ON -DGGML_SCHED_MAX_COPIES=1 -DGGML_CUDA_IQK_FORCE_BF16=1 -DCMAKE_CUDA_HOST_COMPILER=/usr/bin/gcc-14 -DGGML_NATIVE=ON -DCMAKE_CUDA_ARCHITECTURES="120a-real"
> 
> cmake --build build --config Release --clean-first -j 32
> 
> [ 6%] Building CUDA object ggml/src/ggml-cuda/CMakeFiles/ggml-cuda.dir/ggml-cuda.cu.o/home/mir/services/llama.cpp/llama.cpp/ggml/src/ggml-cuda/ggml-cuda.cu
> 
> (1299): error: identifier "ggml_mul_mat_get_scale" is undefined const ggml_tensor * nvfp4_t_s_tensor = src0->type == GGML_TYPE_NVFP4 ? ggml_mul_mat_get_scale(dst) : nullptr;
> 
> followed by several other places with same issue

This previously had another PR as a pre-requisite but that is outdated and no longer the case.  A new update will be posted soon.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2453,6 +2468,7 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor *
         ggml_tensor dst_slice;
         memset(&dst_slice, 0, sizeof(dst_slice));
         dst_slice.buffer = dst->buffer;
+        dst_slice.op     = GGML_OP_MUL_MAT;
```
am17an: remove

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3847,7 +3863,7 @@ static void ggml_cuda_graph_evaluate_and_capture(ggml_backend_cuda_context * cud
 
                             const ggml_tensor * src0 = up_n->src[0];
                             const ggml_tensor * src1 = up_n->src[1];
-                            const ggml_tensor * ids  = up_n->src[2];
+                            const ggml_tensor * ids  = op == GGML_OP_MUL_MAT_ID ? up_n->src[2] : nullptr;
```
am17an: remove

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3886,7 +3902,7 @@ static void ggml_cuda_graph_evaluate_and_capture(ggml_backend_cuda_context * cud
 
                             const ggml_tensor * src0 = up->src[0];
                             const ggml_tensor * src1 = up->src[1];
-                            const ggml_tensor * ids  = up->src[2];
+                            const ggml_tensor * ids  = op == GGML_OP_MUL_MAT_ID ? up->src[2] : nullptr;
```
am17an: remove

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3948,7 +3964,7 @@ static void ggml_cuda_graph_evaluate_and_capture(ggml_backend_cuda_context * cud
 
                         const ggml_tensor * src0 = mm_node->src[0];
                         const ggml_tensor * src1 = mm_node->src[1];
-                        const ggml_tensor * ids  = mm_node->src[2];
+                        const ggml_tensor * ids  = mm_node->op == GGML_OP_MUL_MAT_ID ? mm_node->src[2] : nullptr;
```
am17an: remove

ggml/src/ggml-cuda/vecdotq.cuh
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#include "common.cuh"
+#include "convert.cuh"
```
am17an: why the header change?

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2350,10 +2360,15 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor *
             }
         }
 
+        // this is temporary to block MMQ for now
+    if (src0->type != GGML_TYPE_NVFP4) {
         if (ggml_cuda_should_use_mmq(src0->type, cc, ne12, /*n_experts=*/ne02)) {
```
am17an, line 2358: this check should go inside `ggml_cuda_should_use_mmq`

ggml/src/ggml-cuda/convert.cu
```diff
@@ -617,6 +617,45 @@ static void dequantize_row_mxfp4_cuda(const void * vx, dst_t * y, const int64_t
     dequantize_block_mxfp4<<<nb, 32, 0, stream>>>(vx, y);
 }
 
+template <typename dst_t>
+static __global__ void dequantize_block_nvfp4(
+        const void * __restrict__ vx,
+        dst_t * __restrict__ yy,
+        const int64_t k) {
```
am17an: I think `ne` is a better name

ggml/src/ggml-cuda/convert.cuh
```diff
@@ -58,6 +62,54 @@ template<typename dst_t, typename src_t>
 #else
         return {x.x, x.y};
 #endif // GGML_USE_HIP
+    } else if constexpr (std::is_same_v<src_t, ue4m3>) {
```
am17an: shouldn't this also check `std::is_same_v<dst_t, float>`?

ggml/src/ggml-cuda/vecdotq.cuh
```diff
@@ -322,6 +322,38 @@ static __device__ __forceinline__ float vec_dot_mxfp4_q8_1(
...
+        const int2 v0 = get_int_from_table_16(get_int_b4(bq4->qs, iqs0), kvalues_mxfp4);
+        const int2 v1 = get_int_from_table_16(get_int_b4(bq4->qs, iqs1), kvalues_mxfp4);
+        const block_q8_1 * bq8 = bq8_1 + (is >> 1);
+        const int i8 = ((is & 1) << 2);
+
+        int sumi = ggml_cuda_dp4a(v0.x, get_int_b4(bq8->qs, i8 + 0), 0);
+        sumi = ggml_cuda_dp4a(v0.y, get_int_b4(bq8->qs, i8 + 2), sumi);
+        sumi = ggml_cuda_dp4a(v1.x, get_int_b4(bq8->qs, i8 + 1), sumi);
+        sumi = ggml_cuda_dp4a(v1.y, get_int_b4(bq8->qs, i8 + 3), sumi);
+
```
am17an, line 350: maybe these should be 4 separate floats which you add together at the end? I have not looked at the PTX but it maybe that the compiler can't unroll these

ggml/src/ggml-cuda/ggml-cuda.cu
michaelw9999: Stale file even though I fixed it first it wasn't coming in.  Git issues... pushed a new one about 3minutes after.   Fixing a few more.  Sorry

ggml/src/ggml-cuda/vecdotq.cuh
michaelw9999: Moved ggml_cuda_ue4m3_to_fp32 from common to convert into ggml_cuda_cast per J

ggml/src/ggml-cuda/ggml-cuda.cu
michaelw9999: `case GGML_TYPE_NVFP4:` is not in the switch.  Default is           mmq_supported = false;` so even if called it will still return false.

am17an: so this if statement is not required?

ggml/src/ggml-cuda/common.cuh
michaelw9999: Correct, not required anymore; removed after blocking F16.  Precision difference is negligible on F32 paths.

ggml/src/ggml-cuda/convert.cu
michaelw9999: Renamed to ne

ggml/src/ggml-cuda/ggml-cuda.cu
michaelw9999: No, it is not really required, it's just for clarity.  When the MMQ files were part of this PR, now removed, I was asked to put in a comment that blocking MMQ was temporary.  Since there are no MMQ file changes in this PR, I moved that notice instead.  I can take it out if you want.

ggml/src/ggml-cuda/convert.cuh
michaelw9999: Yes, I've added that check in.

ggml/src/ggml-cuda/vecdotq.cuh
michaelw9999: Went and checked and did comparison as int (since dp4a output is int)  The compiler unrolls them.
With Nemotron-3-Nano-30B:
|Test|Registers|ns/op|Occupancy|LlamaBench|
|:---|:---|:---|:---|:---|
4 chained | 44 | 9632 | 65.36% | 134.17 |
4 separate  | 46 | 9792 | 64.52% | 129.12 |
<details>
<summary>PTX Output</summary>

```
4 dp4a chained baseline:
// begin inline asm
	dp4a.s32.s32 %r63, %r64, %r65, %r86;
	// end inline asm
	ld.global.nc.b32 	%r69, [%rd38+4];
	// begin inline asm
	dp4a.s32.s32 %r67, %r68, %r69, %r63;
	// end inline asm
	ld.global.nc.b32 	%r73, [%rd38];
	// begin inline asm
	dp4a.s32.s32 %r71, %r72, %r73, %r67;
	// end inline asm
	ld.global.nc.b32 	%r77, [%rd38+8];
	// begin inline asm
	dp4a.s32.s32 %r75, %r76, %r77, %r71;
	// end inline asm
	ld.global.nc.b8 	%rs1, [%rd33+-1];
	setp.eq.s16 	%p3, %rs1, 0;
	setp.eq.s16 	%p4, %rs1, 127;
	// begin inline asm
	{cvt.rn.f16x2.e4m3x2 %r79, %rs1;}
4 separate dp4a :
	// begin inline asm
	dp4a.s32.s32 %r63, %r64, %r65, %r98;
	// end inline asm
	ld.global.nc.b32 	%r69, [%rd38+4];
	// begin inline asm
	dp4a.s32.s32 %r67, %r68, %r69, %r98;
	// end inline asm
	ld.global.nc.b32 	%r73, [%rd38];
	// begin inline asm
	dp4a.s32.s32 %r71, %r72, %r73, %r98;
	// end inline asm
	ld.global.nc.b32 	%r77, [%rd38+8];
	// begin inline asm
	dp4a.s32.s32 %r75, %r76, %r77, %r98;
	// end inline asm
	add.s32 	%r133, %r67, %r63;
	add.s32 	%r134, %r133, %r71;
	add.s32 	%r135, %r134, %r75;
	ld.global.nc.b8 	%rs1, [%rd33+-1];
	setp.eq.s16 	%p3, %rs1, 0;
	setp.eq.s16 	%p4, %rs1, 127;
	// begin inline asm
	{cvt.rn.f16x2.e4m3x2 %r79, %rs1;}
```
</details>

ggml/src/ggml-cuda/convert.cuh
```diff
@@ -58,6 +62,54 @@ template<typename dst_t, typename src_t>
...
+            bits = (exp32 << 23) | mant32;
+        }
+
+        float f;
+        memcpy(&f, &bits, sizeof(f)); // 0.5f is baked in.
+        return f;
+#endif
+#else
+        return ggml_ue4m3_to_fp32(x.x);
+#endif
```
am17an: I think `ggml_cuda_cast` is probably not a consistent way to do this, let's do a similar thing like `ggml_cuda_e8m0_to_f32`

ggml/src/ggml-cuda/vecdotq.cuh
```diff
@@ -322,6 +322,38 @@ static __device__ __forceinline__ float vec_dot_mxfp4_q8_1(
...
+    for (int i = 0; i < VDR_NVFP4_Q8_1_MMVQ/2; i++) {
+        const int iqs0 = iqs + 2*i;
+        const int iqs1 = iqs0 + 1;
+        const int is = iqs0 >> 1;
+        const int2 v0 = get_int_from_table_16(get_int_b4(bq4->qs, iqs0), kvalues_mxfp4);
+        const int2 v1 = get_int_from_table_16(get_int_b4(bq4->qs, iqs1), kvalues_mxfp4);
+        const block_q8_1 * bq8 = bq8_1 + (is >> 1);
+        const int i8 = ((is & 1) << 2);
+
+        int sumi = ggml_cuda_dp4a(v0.x, get_int_b4(bq8->qs, i8 + 0), 0);
```
am17an: change all these to `int32_t`

ggml/src/ggml-cuda/vecdotq.cuh
```diff
@@ -322,6 +322,38 @@ static __device__ __forceinline__ float vec_dot_mxfp4_q8_1(
...
 }
 
+#define VDR_NVFP4_Q8_1_MMVQ 4
+#define VDR_NVFP4_Q8_1_MMQ  8
+
+static __device__ __forceinline__ float vec_dot_nvfp4_q8_1(
+                                        const void * __restrict__ vbq,
+                                        const block_q8_1 * __restrict__ bq8_1,
+                                        const int & kbx,
+                                        const int & iqs) {
```
am17an: change all these to `int32_t`

michaelw9999: Changed them in both locations

ggml/src/ggml-cuda/convert.cuh
michaelw9999: Restored `ggml_cuda_ue4m3_to_fp32` back in between  `ggml_cuda_e8m0_to_fp32`  and `ggml_cuda_float_to_fp4_e2m1`

JohannesGaessler:
The FP8 -> FP32 conversion can be simplified a lot: https://github.com/JohannesGaessler/llama.cpp/commit/881fab1027c6d03425c6cc4bf6aad72f09be3d22

michaelw9999:
> The FP8 -> FP32 conversion can be simplified a lot: [JohannesGaessler@881fab1](https://github.com/JohannesGaessler/llama.cpp/commit/881fab1027c6d03425c6cc4bf6aad72f09be3d22)

Awesome! Benchmarked it and it keeps the speedup.  Super tiny, nice!  It's in.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2281,6 +2286,11 @@ static void ggml_cuda_mul_mat(ggml_backend_cuda_context & ctx, const ggml_tensor
         any_gpus_with_slow_fp16 = any_gpus_with_slow_fp16   || !fast_fp16_hardware_available(cc);
     }
 
+    // temporarily block MMQ for NVFP4
+    if (src0->type == GGML_TYPE_NVFP4) {
+        use_mul_mat_q = false;
+    }
```
JohannesGaessler: This is not needed, `ggml_cuda_should_use_mmq` has a whitelist of types.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2350,10 +2360,15 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor *
             }
         }
 
+        // this is temporary to block MMQ for now
+    if (src0->type != GGML_TYPE_NVFP4) {
```
JohannesGaessler: Not needed.

ggml/src/ggml-cuda/vecdotq.cuh
JohannesGaessler: Sorry, this was a bad suggestion by me. I forgot that `ggml_cuda_cast` only works for casting "proper" types, if we have minifloats stored as low-bit integers we can't use it. The conversion should be in `common.cuh` as you have already reverted back to. Revert this change as well.

ggml/src/ggml-cuda/vecdotq.cuh
JohannesGaessler: The "a" in dp4a stands for "accumulate", the intended way to use it is to pass the accumulator as the last argument as is done here. Converting the integers to floats and then adding them up should be bit-for-bit identical to first adding up the integers and then converting the sum to float but it would introduce additional instructions.

JohannesGaessler (commented):
Apply the other changes from the commit I linked as well. After that this PR would be good to merge from my end.

michaelw9999:
@JohannesGaessler  regarding the common.cuh/convert actually neither is needed, so it's just back to baseline.  The new `ggml_cuda_ue4m3_to_f32` is in ggml-cuda/common.cuh which is where it should be.

michaelw9999:
I've now pushed all the new changes and corrections

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2281,6 +2286,7 @@ static void ggml_cuda_mul_mat(ggml_backend_cuda_context & ctx, const ggml_tensor
         any_gpus_with_slow_fp16 = any_gpus_with_slow_fp16   || !fast_fp16_hardware_available(cc);
     }
 
+
```
```suggestion
```

JohannesGaessler (approved):

am17an (approved):

am17an (approved):

IMbackK (approved):

CISC:
We have multiple failures:
https://github.com/ggml-org/llama.cpp/actions/runs/23585668334/job/68678073133
https://github.com/ggml-org/llama.cpp/actions/runs/23585668413/job/68678072936 (not sure why, perhaps this branch wasn't rebased to updated version?)
https://github.com/ggml-org/llama.cpp/actions/runs/23585668403/job/68678073118 (looks like runner just disconnected)

CISC:
@IMbackK Looks like we have some untested surfaces in general build vs. release still.

IMbackK:
yeah i opend a issue for that

CISC:
> yeah i opend a issue for that

The release failure is the most important one though.

IMbackK:
Right gfx942 uses a different fp8 format to all others, so its implementation in rocm is unique, looks like there is a bug in the header, i would recomend simply disableing this path on gfx942 for now, i will add a pr to this effect this evening.

stduhpf:
I think this PR broke my ROCm 6.2 build on windows. 

```
hip.h:240:9: error: unknown type name '__hip_fp8_e4m3'
  240 | typedef __hip_fp8_e4m3 __nv_fp8_e4m3;
      |         ^
563 warnings and 1 error generated when compiling for gfx1030.
```

michaelw9999:
> I think this PR broke my ROCm 6.2 build on windows.
> 
> ```
> hip.h:240:9: error: unknown type name '__hip_fp8_e4m3'
>   240 | typedef __hip_fp8_e4m3 __nv_fp8_e4m3;
>       |         ^
> 563 warnings and 1 error generated when compiling for gfx1030.
> ```

@stduhpf sorry about that!  Can you tell me what gpu you have?

stduhpf:
I have a RX 6800 (gfx1030). Just removing the whole `#if HIP_VERSION >= 60200000` block fixes it.

PR #20664, opened 2026-03-17
ggml-webgpu: Add supports for `DIAG` and `TRI`

yomaytk:
This PR adds support for `DIAG` and `TRI` to the WebGPU backend. These operations are used by Qwen 3.5.


yomaytk:
@reeselevine If you’re already working on supporting new ops for Qwen 3.5, this PR might conflict with your work, so please let me know in that case.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -73,16 +73,17 @@ struct ggml_webgpu_shader_lib_context {
...
+    size_t        wg_mem_limit_bytes       = 0;
+    bool          inplace                  = false;
+    bool          overlap                  = false;
+    bool          src_overlap              = false;
+    bool          supports_subgroup_matrix = false;
+    uint32_t      sg_mat_m                 = 0;
+    uint32_t      sg_mat_n                 = 0;
+    uint32_t      sg_mat_k                 = 0;
+    uint32_t      max_subgroup_size        = 0;
+    ggml_tri_type ttype;
```
reeselevine: I don't think we should add this to the context struct, since it can be derived pretty easily given the tensor.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -244,13 +245,15 @@ struct ggml_webgpu_binary_pipeline_key_hash {
...
 struct ggml_webgpu_unary_pipeline_key {
-    int  type;
-    int  op;
-    bool is_unary;  // many unary operators fall under the GGML_OP_UNARY umbrella
-    bool inplace;
+    int           type;
+    int           op;
+    bool          is_unary;  // many unary operators fall under the GGML_OP_UNARY umbrella
+    bool          inplace;
+    ggml_tri_type ttype;
```
reeselevine: I do like how you split up the different `TRI` cases within the unary shader, but I don't really like how the pipeline key now encodes the tri type, which isn't relevant for any other unary operation.

I think it's ok though, maybe just add a comment that says this only applies to `TRI` operations to be extra clear?

reeselevine (commented):
You beat me to it by a few hours! I think these implementations are good though, getting `DIAG` to fit in the unary shader and splitting up the separate `TRI` cases. I'll plan on merging this first and updating my other PR around it.

yomaytk:
> You beat me to it by a few hours! I think these implementations are good though, getting DIAG to fit in the unary shader and splitting up the separate TRI cases. I'll plan on merging this first and updating my other PR around it.

Thanks! I've also addressed the issue you pointed out.

reeselevine (approved):

PR #20665, opened 2026-03-17
ggml-webgpu: Update the `RMS_NORM` preprocessor and add `L2_NORM`

yomaytk:
This PR updates the `RMS_NORM` preprocessor and add `L2_NORM`. This op is used by Qwen 3.5.
- It  looks like we can update the previous preprocessor (e.g., `VARIANTS`, `DECLS`) of `RMS_NORM` to the new one, so I updated accordingly.
- Add the support for `L2_NORM` with expanding `RMS_NORM` shader.


reeselevine:
This looks pretty good too although I slightly prefer the more generic naming of "row_norm" for the base shader that I have implemented in https://github.com/ggml-org/llama.cpp/pull/20687

yomaytk:
I've addressed the issue you pointed out. Could you review again?

reeselevine (approved):

PR #20680, opened 2026-03-17
server: add `hidden` preset option to filter models from /models endpoint

### LINKED ISSUE #18609: Feature Request: Option to show only preset-defined models in router mode
### Prerequisites

### Feature Description

When running a llama-server in router mode with `--models-preset`, it would be helpful to have an option to restrict the `GET /models` endpoint to only show models that are explicitly defined in the preset file, rather than showing all cached models or models in `--models-dir`.

### Motivation

This would be useful for removing clutter from the default llama-server webui, for example I don't want to show my reranker models there, or sometimes I want to have a custom name for a model because it includes GGUF in the title and that's too long.

### Possible Implementation

Add a new command-line flag such as `--models-preset-only` that, when used with `--models-preset`, filters the `/models` endpoint response to only include models defined in the preset configuration file. Or as an alternative, have a config for each model defined in the presets to be hidden, and we could use the special selector `[*]` to hide all by default and only show the ones that I want to show in the API.

drizzt:
Closes #18609

## Problem

In router mode with `--models-preset`, `GET /models` returns **all** discovered models (cached, from `--models-dir`, and from the preset file). This clutters the web UI and the API with models the user does not want to expose (e.g. reranker models, or models with long GGUF names).

## Solution

Add a per-model `hidden` boolean option for the preset INI file. When set, the model is excluded from the `GET /models` / `GET /v1/models` response, while remaining fully loadable and usable via direct API calls.

The `[*]` global selector can be used to hide everything by default and opt individual models back in:

```ini
[*]
hidden = true        ; hide all models by default

[my-chat-model]
hf-repo = user/chat-model
hidden = false       ; explicitly visible

[my-reranker]
hf-repo = user/reranker
; inherits hidden = true from [*] → stays hidden
```

## Changes

| File | Change |
|------|--------|
| `common/arg.h` | Add `COMMON_ARG_PRESET_HIDDEN` constant |
| `common/arg.cpp` | Register `hidden` as a preset-only option (follows `load-on-startup` / `stop-timeout` patterns) |
| `tools/server/server-models.h` | Add `bool hidden = false` field to `server_model_meta` |
| `tools/server/server-models.cpp` | Read `hidden` from preset in `load_models()`; skip hidden models in `get_router_models` |

## Notes

- Hidden models are **only** excluded from the listing endpoints. They can still be loaded and used via `/completions`, `/chat/completions`, etc.
- The `[*]` cascading is handled by the existing `ctx_preset.cascade(global, ...)` logic — no changes needed there.
- Follows the exact same patterns as the existing `load-on-startup` and `stop-timeout` preset-only options.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -3865,6 +3865,12 @@ void common_params_add_preset_options(std::vector<common_arg> & args) {
         [](common_params &, int) { /* unused */ }
     ).set_env(COMMON_ARG_PRESET_STOP_TIMEOUT).set_preset_only());
 
+    args.push_back(common_arg(
+        {"hidden"},
+        "in server router mode, hide this model from the /models endpoint",
+        [](common_params &) { /* unused */ }
+    ).set_env(COMMON_ARG_PRESET_HIDDEN).set_preset_only());
+
     // args.push_back(common_arg(
     //     {"pin"},
     //     "in server router mode, do not unload this model if models_max is exceeded",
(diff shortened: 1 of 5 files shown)


ngxson:
That is not what the initial issue asked for

PR #20682, opened 2026-03-17
ggml-cpu: fix RVV checks in quants and repacking

taimur-10x:
## Summary
This PR adds macros for checking RVV availability in quantized vector-dot kernels and repack GEMM/GEMV. 
(Fixes #20669) 

## Key Changes
- Added `__riscv_v_intrinsic` checks for RVV vector-dot kernels
- Added `__riscv_zvfh` check for repack GEMM/GEMV. Repacking pathway is only enabled if `zvfh` is available.


xctan (approved):

ggerganov:
Why does the risc-v CI not detect this issue? Can we extend the CI to detect this in some way?

xctan:
Our current CI scripts don't set `GGML_RVV`, which defaults to `ON`. We can just add a non-RVV config to `build-riscv.yml` to catch this.

ggerganov:
> Our current CI scripts don't set GGML_RVV, which defaults to ON. We can just add a non-RVV config to build-riscv.yml to catch this.

Would be useful to add a job to exercise this.

PR #20692, opened 2026-03-17
ggml-cpu/x86: fix unused changemask warning in repack

mrshaw01:
## Summary

Fix an `-Wunused-but-set-variable` warning in `ggml/src/ggml-cpu/arch/x86/repack.cpp`.

`changemask` was declared unconditionally in `gemv_q4_b32_8x8_q8_0_lut_avx`, but it is only used in the `block_q4_0x8` / `block_iq4_nlx8` path. For the `block_mxfp4x8` template instantiation, that branch is discarded by `if constexpr`, which leaves `changemask` unused and triggers the warning.

This change moves the declaration into the branch where it is used.

## Testing

- Built locally with CMake and verified the warning no longer appears.

### DIFF
--- a/ggml/src/ggml-cpu/arch/x86/repack.cpp
+++ b/ggml/src/ggml-cpu/arch/x86/repack.cpp
@@ -531,7 +531,6 @@ static void gemv_q4_b32_8x8_q8_0_lut_avx(int n, float * GGML_RESTRICT s, size_t 
 
     UNUSED(bs);
 
-    __m128i changemask = _mm_set_epi8(15, 14, 7, 6, 13, 12, 5, 4, 11, 10, 3, 2, 9, 8, 1, 0);
     __m256i finalpermutemask = _mm256_set_epi32(7, 5, 3, 1, 6, 4, 2, 0);
 
     // Permute mask used for easier vector processing at later stages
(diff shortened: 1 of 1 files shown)


am17an (approved):

ggml/src/ggml-cpu/arch/x86/repack.cpp
```diff
@@ -580,6 +579,7 @@ static void gemv_q4_b32_8x8_q8_0_lut_avx(int n, float * GGML_RESTRICT s, size_t
                 if constexpr (
                         std::is_same_v<block_tx8, block_q4_0x8> ||
                         std::is_same_v<block_tx8, block_iq4_nlx8>) {
+                    const __m128i changemask = _mm_set_epi8(15, 14, 7, 6, 13, 12, 5, 4, 11, 10, 3, 2, 9, 8, 1, 0);
```
ggerganov, line 582: Wouldn't this have performance impact from calling `_mm_set_epi8` in each iteration instead of once at the start?

mrshaw01: I double-checked the exact `_mm_set_epi8`-using instantiations, not just the whole file. For `ggml_gemv_q4_0_8x8_q8_0` and `ggml_gemv_iq4_nl_8x8_q8_0`, the emitted function bytes are identical before and after the change on GCC 9.4.0 with `-O2/-O3 -mavx2 -mf16c -mfma`. The reason is that `if constexpr` removes the unused `block_mxfp4x8` branch at compile time, and for the `block_q4_0x8` / `block_iq4_nlx8` paths, `_mm_set_epi8(...)` is just a pure constant-vector expression, so the compiler can fold or hoist it as before rather than materializing it per iteration. So in this compiler/config, the warning is fixed without changing the generated code for the branch that actually uses `changemask`.

ggerganov: Ok, thanks for confirming.

mrshaw01:
@am17an In case 1 test failed, how can I rerun?

PR #20696, opened 2026-03-17
HIP: ignore return of hipMemAdvise

IMbackK:
we want to ignore the error here but hipMemAdvise is nodiscard


CISC (approved):
Merge this and rebase #20430 so the complete test can run.

IMbackK:
thanks for merging, i dont have write access ever since that was generally janked from collaborators.

PR #20707, opened 2026-03-18
ggml-backend: remove spurious check

am17an:
We already check if the op is MUL_MAT_ID before entering this if statement


ggerganov (commented):
I think the checks were needed in the case of `GGML_OP_ADD_ID`. That op is currently disabled, by I think it does not hurt to keep the checks in case we experiment with it in the future.

am17an:
Ok will leave it as it is then

PR #20712, opened 2026-03-18
fix(rpc): prevent division by zero in deserialize_tensor

y198nt:
When receiving an RPC message with a deprecated tensor type (e.g., type 4 or 5 where `blck_size == 0`), `ggml_row_size()` will trigger a division by zero (SIGFPE) and crash the rpc-server. 

This patch adds a simple validation check in `deserialize_tensor` to return `nullptr` if the requested tensor type has a block size of 0.

(Note: This was originally reported via Security Advisory and maintainer suggested dropping a patch here).


rgerganov (approved):
`ggml_blck_size` performs a simple array lookup, so this shouldn't have performance impact. looks good to me.

rgerganov (approved):

PR #20716, opened 2026-03-18
gguf : fix division by zero

ggerganov:
Prevent malformed data causing divisions by zero and crashes.

### DIFF
--- a/ggml/src/gguf.cpp
+++ b/ggml/src/gguf.cpp
@@ -615,9 +615,9 @@ struct gguf_context * gguf_init_from_file_impl(FILE * file, struct gguf_init_par
                     ok = ok && gr.read(info.t.ne[j]);
                 }
 
-                // check that all ne are non-negative
-                if (info.t.ne[j] < 0) {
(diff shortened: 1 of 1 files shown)


taronaeo (approved):

Green-Sky:
While I am for this change, to explicitly not support zero sized tensors, safetensors does support this. And over at sd.cpp there are some models that currently use this.
I do agree though that one really should use metadata instead, or a one dim tensor as a workaround.

JohannesGaessler:
Also more generally: right now the ggml API *does* allow the user to create zero-sized tensors. I think preferably we would either allow or disallow these consistently for ggml and GGUF.

ngxson:
I think disallowing 0-element tensors everywhere is more preferable (which can be follow-up PRs), as we don't have any use cases for that.

I suspect that safetensors support it because they also have scalar tensors. I have no ideas how they are stored at byte level, but it make more sense (and it is indeed a common practice) to convert pytorch scalar tensor into GGML 1-element tensor

PR #20688, opened 2026-03-17
webui: improve tooltip wording for attachment requirements

julien-c:
## Summary
- Reword image attachment tooltip: "Images require vision models to be processed" → "Image processing requires a vision model"
- Reword audio attachment tooltip: "Audio files require audio models to be processed" → "Audio files processing requires an audio model"

## Test plan
- [ ] Open the web UI with a non-vision model loaded and hover over the image attachment icon to verify the updated tooltip
- [ ] Open the web UI with a non-audio model loaded and hover over the audio attachment icon to verify the updated tooltip

PR #20724, opened 2026-03-18
WebUI: Rebuild index.html.gz

crsawyer:
Looks like there was an issue in rebuilding index.html.gz in #20688

Web page would not load. Chrome error log shows:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'start')
    at (index):592:26
```

Only change is rebuilding the static file

Also fixes #20722


crsawyer:
Going to loop in @allozaur. It appears there are no code owners for the static `public` directory.

allozaur (approved):

allozaur:
Thank you for catching this, I've merged the PR

thomas-0816:
see https://github.com/ggml-org/llama.cpp/issues/20738

PR #20730, opened 2026-03-18
Convert: Make NVFP4 and MXFP4 HF conversions say NVFP4/MXFP4 instead of BF16

michaelw9999:
When converting an NVFP4 or MXFP4 HF model to GGUF, the default behavior is to keep the same name if not specified, so the converted models still are named <model name> BF16:

Before:
```
| model                          |     
| ------------------------------ | 
| qwen35 0.8B BF16               |
```
After:
```
| model                          |     
| ------------------------------ | 
| qwen35 0.8B NVFP4              |
```
  This script edit identifies the file type properly for both NVFP4 and MXFP4 and adds the missing MOSTLY_NVFP4 and missing MOSTLY_MXFP4_MOE from the table in the py constants file.

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -1150,9 +1150,8 @@ class TextModel(ModelBase):
             self.gguf_writer.add_key_length(head_dim)
             self.gguf_writer.add_value_length(head_dim)
 
-        file_type = gguf.LlamaFileType.MOSTLY_NVFP4 if self._is_nvfp4 else self.ftype
-        self.gguf_writer.add_file_type(file_type)
-        logger.info(f"gguf: file type = {file_type}")
+        self.gguf_writer.add_file_type(self.ftype)
+        logger.info(f"gguf: file type = {self.ftype}")
 
     def write_vocab(self):
         if len(self.gguf_writer.tensors) != 1:
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -876,8 +876,11 @@ class ModelBase:
         if self.metadata.name is None:
             self.metadata.name = self.dir_model.name
 
-        if self._is_nvfp4 and self.model_name is None and self.metadata.name is not None and self.metadata.name.endswith(" BF16"):
-            self.metadata.name = f"{self.metadata.name[:-5]} NVFP4"
+        if self.ftype in (gguf.LlamaFileType.ALL_F32, gguf.LlamaFileType.MOSTLY_F16, gguf.LlamaFileType.MOSTLY_BF16):
+            if self._is_nvfp4:
+                self.ftype = gguf.LlamaFileType.MOSTLY_NVFP4
+            elif self._is_mxfp4:
+                self.ftype = gguf.LlamaFileType.MOSTLY_MXFP4_MOE
 
         # Generate parameter weight class (useful for leader boards) if not yet determined
         if self.metadata.size_label is None and total_params > 0:
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -1022,7 +1022,7 @@ class TextModel(ModelBase):
 
         total_params = self.gguf_writer.get_total_parameter_count()[0]
         # Extract the encoding scheme from the file type name. e.g. 'gguf.LlamaFileType.MOSTLY_Q8_0' --> 'Q8_0'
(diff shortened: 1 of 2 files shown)


convert_hf_to_gguf.py
```diff
@@ -876,6 +876,9 @@ def prepare_metadata(self, vocab_only: bool):
         if self.metadata.name is None:
             self.metadata.name = self.dir_model.name
 
+        if self._is_nvfp4 and self.model_name is None and self.metadata.name is not None and self.metadata.name.endswith(" BF16"):
+            self.metadata.name = f"{self.metadata.name[:-5]} NVFP4"
```
```suggestion
        if self.ftype in (gguf.LlamaFileType.ALL_F32, gguf.LlamaFileType.MOSTLY_F16, gguf.LlamaFileType.MOSTLY_BF16):
            if self._is_nvfp4:
                self.ftype = gguf.LlamaFileType.MOSTLY_NVFP4
            elif self._is_mxfp4:
                self.ftype = gguf.LlamaFileType.MOSTLY_MXFP4_MOE
```

convert_hf_to_gguf.py
```diff
@@ -1016,7 +1019,7 @@ def prepare_metadata(self, vocab_only: bool):
 
         total_params = self.gguf_writer.get_total_parameter_count()[0]
         # Extract the encoding scheme from the file type name. e.g. 'gguf.LlamaFileType.MOSTLY_Q8_0' --> 'Q8_0'
-        output_type: str = self.ftype.name.partition("_")[2]
+        output_type: str = "NVFP4" if self._is_nvfp4 else self.ftype.name.partition("_")[2]
```
```suggestion
        output_type: str = self.ftype.name.partition("_")[2]
```

convert_hf_to_gguf.py
```diff
@@ -1144,8 +1147,9 @@ def set_gguf_parameters(self):
             self.gguf_writer.add_key_length(head_dim)
             self.gguf_writer.add_value_length(head_dim)
 
-        self.gguf_writer.add_file_type(self.ftype)
-        logger.info(f"gguf: file type = {self.ftype}")
+        file_type = gguf.LlamaFileType.MOSTLY_NVFP4 if self._is_nvfp4 else self.ftype
+        self.gguf_writer.add_file_type(file_type)
+        logger.info(f"gguf: file type = {file_type}")
```
```suggestion
        self.gguf_writer.add_file_type(self.ftype)
        logger.info(f"gguf: file type = {self.ftype}")
```

gguf-py/gguf/constants.py
```diff
@@ -3869,6 +3869,8 @@ class LlamaFileType(IntEnum):
     # MOSTLY_Q4_0_8_8      = 35  # removed from gguf files, use Q4_0 and runtime repack
     MOSTLY_TQ1_0         = 36  # except 1d tensors
     MOSTLY_TQ2_0         = 37  # except 1d tensors
+    MOSTLY_MXFP4_MOE     = 38  # except 1d tensors
+    MOSTLY_NVFP4         = 39  # except 1d tensors
```
CISC, line 3873: Nice catch.

CISC (commented):
This simplifies things while also fixing MXFP4.

Add `quant_method = (self.hparams.get("quantization_config") or {}).get("quant_method")` here
https://github.com/ggml-org/llama.cpp/blob/5d41b4c8f44555740fe9bf15d1956958b4a11761/convert_hf_to_gguf.py#L714
and add `self._is_mxfp4 = quant_method == "mxfp4"` here
https://github.com/ggml-org/llama.cpp/blob/5d41b4c8f44555740fe9bf15d1956958b4a11761/convert_hf_to_gguf.py#L730
then update this
https://github.com/ggml-org/llama.cpp/blob/5d41b4c8f44555740fe9bf15d1956958b4a11761/convert_hf_to_gguf.py#L11126-L11131

CISC (approved):

convert_hf_to_gguf.py
```diff
@@ -728,6 +730,7 @@ def prepare_tensors(self):
                 quant_algo = "NVFP4"
 
         self._is_nvfp4 = quant_algo == "NVFP4"
+        self._is_mxfp4 = quant_method == "mxfp4"
```
ggerganov, line 733: Should these be case insensitive?

CISC: Not as far as I've been able to tell, `quant_method` seems to be always lower-case, while `quant_algo` (at least for `NVFP4`) is always uppper-case.

PR #20751, opened 2026-03-19
mtmd: add clip_graph::build_mm()

ngxson:
Adding `build_mm()`, which allow future models to hook begin/after the `ggml_mul_mat()`. Only instances where `mm(weight, inp)` are changed; instances like `mm(q, k)` stays unchanged

This is required by a model that we planned to support.

### DIFF
--- a/tools/mtmd/clip-graph.h
+++ b/tools/mtmd/clip-graph.h
@@ -42,6 +42,7 @@ struct clip_graph {
     virtual ggml_cgraph * build() = 0;
 
     // wrapper around ggml_mul_mat, allow hooking (e.g. LoRA, clamping) depending on the model
+    // tensor a should be the weight matrix, and tensor b should be the input
     virtual ggml_tensor * build_mm(ggml_tensor * a, ggml_tensor * b) const;
 
     //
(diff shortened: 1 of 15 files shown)


tools/mtmd/clip-graph.h
```diff
@@ -41,6 +41,10 @@ struct clip_graph {
     virtual ~clip_graph() = default;
     virtual ggml_cgraph * build() = 0;
 
+    // wrapper around ggml_mul_mat, allow hooking (e.g. LoRA, clamping) depending on the model
+    // tensor a should be the weight matrix, and tensor b should be the input
+    virtual ggml_tensor * build_mm(ggml_tensor * a, ggml_tensor * b) const;
```
CISC: In this case you must also have an optional `bias` tensor so that it can be applied correctly.

CISC (approved):

tools/mtmd/clip-graph.h
ngxson: Ah yeah, nice catch. Indeed, the model that we planned to support doesn't have bias, and LoRA doesn't target bias either, so I think we're fine for now.

I can add a TODO for an overload `build_mm(w, b, inp)`

ggerganov (approved):

PR #20752, opened 2026-03-19
Add Qwen3 TTS architecture support

Acceldium:
Note that this branch adds initial support for running Qwen3-TTS models in llama.cpp. Qwen3-TTS uses a multi-stage pipeline (language model + audio decoder/tokenizer) that requires executing multiple independent compute graphs in sequence — a pattern llama.cpp does not currently support natively.


Rye426:
May I ask if the Q4 quantization scheme can be used? The F16 synthesis results are currently good, but the process takes a long time.

PR #20756, opened 2026-03-19
ci : add action for finding duplicate issues

ggerganov:
When a new issue is opened, a Github Action searches for similar issues and links them in a comment.

The idea is the same as the one in [OpenCode](https://github.com/anomalyco/opencode/blob/5a0bfa70616a55e4d3ffcfd74fddbe02abcb46c7/.github/workflows/duplicate-issues.yml). However this implementation runs the inference locally on self-hosted runners.

Here is an example: https://github.com/ggml-org/tmp/issues/8


.github/workflows/ai-issues.yml
```diff
@@ -0,0 +1,71 @@
...
+    if: github.event.action == 'opened'
+    runs-on: [self-hosted, opencode]
+
+    permissions:
+      contents: read
+      issues: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
CISC:
```suggestion
        uses: actions/checkout@v6
```

CISC (approved):

PR #20764, opened 2026-03-19
chat : handle tool calls with no required args in TAG_WITH_TAGGED format

jpohhhh:
regarding #20650

When a tool has no required args, the model may emit <tool_call>name</tool_call> with no arg tags. The grammar tried to peek ahead for an arg tag after the tool name, even when all args were optional, so the peek failed, and the tool call was not recognized.

### DIFF
--- a/tests/test-chat.cpp
+++ b/tests/test-chat.cpp
@@ -1912,9 +1912,26 @@ static void test_template_output_peg_parsers(bool detailed_debug) {
         })
             .expect_tool_calls({
                 { "special_function", R"({"arg1": 1})", {} },
-                { "special_function_with_opt", R"({"arg1": 1, "arg2": 2})", {} },
-            })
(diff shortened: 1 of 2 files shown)


pwilkin (approved):

tests/test-chat.cpp
```diff
@@ -1912,9 +1912,26 @@ static void test_template_output_peg_parsers(bool detailed_debug) {
         })
             .expect_tool_calls({
                 { "special_function", R"({"arg1": 1})", {} },
-                { "special_function_with_opt", R"({"arg1": 1, "arg2": 2})", {} },
-            })
+                { "special_function_with_opt", R"({"arg1": 1, "arg2": 2})", {} },            })
```
aldehir: nit: undo whitespace change
```suggestion
                { "special_function_with_opt", R"({"arg1": 1, "arg2": 2})", {} },
            })
```

aldehir (approved):

PR #20772, opened 2026-03-19
ci : improve action for duplicate issue

ggerganov:
cont #20766 

- Move model selection to the model provider
- Switch to Qwen3.5 35B
- Remove repo agent files from the context
- Improve prompt
- Show thinking traces

Example: https://github.com/ggml-org/tmp/issues/15

### DIFF
--- a/.github/workflows/ai-issues.yml
+++ b/.github/workflows/ai-issues.yml
@@ -31,11 +31,18 @@ jobs:
               "webfetch": "deny"
             }
         run: |
+          rm AGENTS.md
+          rm CLAUDE.md
+
           opencode run -m llama.cpp-dgx/gpt-oss-120b-hf-medium --thinking "A new issue has been created:
 
           Issue number: ${{ github.event.issue.number }}
 
-          Lookup this issue with `gh issue view ${{ github.event.issue.number }} --json title,body,url,number`
+          Lookup the contents of the issue using the following command:
+
+          ```bash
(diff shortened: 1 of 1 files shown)


.github/workflows/ai-issues.yml
```diff
@@ -31,11 +31,18 @@ jobs:
               "webfetch": "deny"
             }
         run: |
-          opencode run -m llama.cpp-dgx/gpt-oss-120b-hf-low "A new issue has been created:
+          rm AGENTS.md
+          rm CLAUDE.md
```
CISC, line 35: :rofl:

CISC (approved):

.github/workflows/ai-issues.yml
ngxson: hmm, do we really need to have the project code pulled when running this workflow? If we don't, then we can simply delete the `actions/checkout` step

CISC: I was wondering about this too, but it could come in handy if we extend the prompt to gather some context etc?

ggerganov: Yeah, atm we don't need to checkout, but we could utilize this later on for more tasks.

aviallon: Your `AGENTS.md` prompt sure is efficient to provoke denials.

PR #20784, opened 2026-03-20
Update Dawn version in WebGPU CI

nikhilJain17:
Update Dawn version for WebGPU CI.


reeselevine:
I think it would be better if we move the CI build to use an official Google Dawn release :) (for some context @nikhilJain17 is working with me at UCSC on the WebGPU backend).

The reason we didn't do it before is that Dawn wasn't doing official releases, and build artifacts were being deleted after a short retention period. @nikhilJain17 can we confirm that the Dawn releases and build artifacts do seem to be more stable now?

Also we'll want to update which commit the current CI is testing in the WebGPU docs here.

nikhilJain17:
I'm not sure what the temporary artifacts looked like in the past, but Dawn uses Github Releases now, so we should be good. I see releases from April 2025 until today: [https://github.com/google/dawn/releases](https://github.com/google/dawn/releases). 

Also updated docs to reflect new Dawn commit that we depend on.

CISC (approved):

taronaeo (approved):

PR #20785, opened 2026-03-20
server: add activation capture endpoints for feature interpretability

hrhdegenetrix:
## Summary
Add /activations endpoints to llama-server that expose per-layer mean activation vectors from inference. Enables real-time monitoring of model internals and streaming activation collection for sparse autoencoder (SAE) training: an accessible foundation for feature-level interpretability and behavioral steering in local LLM inference.

## Motivation
There's currently no way to inspect a model's internal activations through llama-server. Interpretability researchers working with local models need to either use separate frameworks (TransformerLens, nnsight) that don't support GGUF, or modify llama.cpp themselves. This patch fills that gap with a minimal, low-overhead implementation that integrates cleanly with the existing server architecture, making activation-level analysis accessible to the broader open-source research community.

## What It Adds
  **Two operating modes:**

  1. **Live monitoring** — per-layer mean activation vectors stored in memory, queryable with optional top-K selection by magnitude. Useful for real-time
  dashboards.
  2. **Collection** — streams full per-token activation vectors to a binary file (16-byte header + float32 arrays, directly numpy-readable). Useful for
  offline SAE training.

  **Implementation:**
  - Uses the public `cb_eval` callback API — no internal headers or private API access
  - Intercepts `l_out` tensors (layer outputs) during graph computation
  - Thread-safe via mutex-protected global struct
  - Routes bypass the server task queue (activation data has its own mutex)
  - Fixes eval callback registration to persist across graph reuse (without this fix, the callback silently stops working after the first inference)

## Files changed (5 files, ~400 lines)

  | File | Change |
  |------|--------|
  | `tools/server/activation_capture.h` | **New** — capture struct, eval callback, binary I/O |
  | `tools/server/server-context.cpp` | Install callback, init struct, 3 route handlers |
  | `tools/server/server-context.h` | Handler declarations |
  | `tools/server/server.cpp` | Endpoint registration |
  | `src/llama-context.cpp` | Move callback registration before graph reuse check |

## Performance

  Measured on RTX 3090, Qwen3-8B Q4_K_M:

  | State | Speed | Overhead |
  |-------|-------|----------|
  | Disabled (default) | ~140.7 tok/s | Near-zero |
  | Enabled (monitoring) | ~131.8 tok/s | ~6% |

  When disabled, the callback returns `false` for all tensors — no GPU→CPU copies occur.

## Usage examples

  ```bash
  # Enable capture
  curl -X POST http://localhost:8080/activations \
    -d '{"enabled": true}' -H "Content-Type: application/json"

  # Query top-5 features from layers 0, 12, 23
  curl "http://localhost:8080/activations?layers=0,12,23&top_k=5"

  # Collect activations to file for SAE training
  curl -X POST http://localhost:8080/activations/collect \
    -d '{"layer": 20, "path": "/tmp/activations.bin"}' \
    -H "Content-Type: application/json"

  # Read with numpy
  python3 -c "
  import numpy as np, struct
  with open('/tmp/activations.bin', 'rb') as f:
      f.read(4)  # magic 'ACTV'
      ver, n_embd, layer = struct.unpack('<III', f.read(12))
  data = np.fromfile('/tmp/activations.bin', dtype=np.float32, offset=16).reshape(-1, n_embd)
  print(f'{data.shape[0]} tokens x {data.shape[1]} dims from layer {layer}')
  "
```

## Testing
Tested with Qwen2.5-0.5B, Qwen3-8B, Qwen3.5-35B-A3B (MoE) on RTX 3090. Works with any architecture that emits l_out tensors.

A [companion repo](https://github.com/hrhdegenetrix/llama-sae-feature-interpretability) demonstrates the full workflow from activation collection through feature discovery and behavioral steering.

---
Magdalene Sullivan — [HeraldAI](https://heraldai.org)
First draft of PR text provided by Claude and heavily revised by Sullivan

### DIFF
--- a/src/llama-context.cpp
+++ b/src/llama-context.cpp
@@ -1186,6 +1186,9 @@ llm_graph_result * llama_context::process_ubatch(const llama_ubatch & ubatch, ll
     // in order to correctly reuse a graph, it's full topology has to be uniquely determined by these parameters
     const auto gparams = graph_params(res, ubatch, mctx, gtype);
 
+    // Always set eval callback (even on graph reuse) so activation capture works
+    ggml_backend_sched_set_eval_callback(sched.get(), cparams.cb_eval, cparams.cb_eval_user_data);
+
     if (!graph_reuse_disable && res->can_reuse(gparams)) {
         //LLAMA_LOG_DEBUG("%s: reusing previous graph\n", __func__);
 
@@ -1194,7 +1197,6 @@ llm_graph_result * llama_context::process_ubatch(const llama_ubatch & ubatch, ll
         res->reset();
 
         ggml_backend_sched_reset(sched.get());
-        ggml_backend_sched_set_eval_callback(sched.get(), cparams.cb_eval, cparams.cb_eval_user_data);
 
         //const auto t_start_us = ggml_time_us();
 
--- a/tools/server/activation_capture.h
+++ b/tools/server/activation_capture.h
@@ -1,0 +1,211 @@
+#pragma once
+
+// Activation capture for llama-server
+// Hooks into ggml's eval callback to capture l_out tensors during inference.
+//
+// Two modes:
+//   1. Live monitoring - stores per-layer mean activations in memory
+//   2. Collection - streams per-token activation vectors to a binary file (for SAE training)
+//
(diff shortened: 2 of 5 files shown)


ngxson:
Closing because this is not in-scope for server's features

PR #20786, opened 2026-03-20
chat : fix Cohere tool_use template crash when model declines tools

jpohhhh:
Two issues:
1. The Cohere tool_use template unconditionally adds the generation prompt (<|CHATBOT_TOKEN|>), so calculate_diff_split finds no difference and generation_prompt is empty. Fall back to the default template's generation_prompt when the tool_use template's is empty.

2. When the generation_prompt already ends with the content start marker (Cohere-specific, flagged via gp_includes_start in the diff analyzer), the ALWAYS_WRAPPED parser expected it again in the model output. Use the same until(tool_start) pattern as the non-wrapped path so content stops before tool calls.

   The gp_includes_start flag is checked together with the actual generation_prompt to avoid false positives: when generation_prompt is empty (e.g. --chat-template-file override without a default template fallback), the original ALWAYS_WRAPPED path is used.

Verified with DevQuasar-6/CohereForAI.c4ai-command-r7b-12-2024-GGUF: tools decline, tools use, and json_schema all return 200.

Test:

  ./build/bin/test-chat


pwilkin:

PR #20799, opened 2026-03-20
server : add special handling for /health in httplib

rgerganov:
When the number of parallel requests to llama-server exceed the number of http threads, llama-server stop responding to /health which is very disruptive in k8s deployments, causing restarts of properly working inference endpoints.

Unfortunately, there is no way to fix this outside of httplib and this patch adds a rather ugly hack for handling GET /health requests before dispatching them to the thread pool. No changes are made in the HTTPS implementation.

Disclaimer: the implementation was AI assisted

closes: #20684

### DIFF
--- a/vendor/cpp-httplib/httplib.cpp
+++ b/vendor/cpp-httplib/httplib.cpp
@@ -1668,6 +1668,124 @@ bool is_socket_alive(socket_t sock) {
   return detail::read_socket(sock, &buf[0], sizeof(buf), MSG_PEEK) > 0;
 }
 
+// Read the first HTTP request line (until \r\n) with timeout. Returns empty string on timeout/error.
+std::string read_first_line(socket_t sock, time_t timeout_sec, time_t timeout_usec) {
+  std::string line;
+  line.reserve(128);
+  char prev = 0;
+  const time_t sec = (timeout_sec > 0 || timeout_usec > 0) ? (timeout_sec ? timeout_sec : 1) : 1;
+  const time_t usec = timeout_usec;
+  for (;;) {
+    if (line.size() >= CPPHTTPLIB_MAX_LINE_LENGTH) { return std::string(); }
+    if (select_read(sock, sec, usec) <= 0) { return std::string(); }
+    char byte;
+    auto n = read_socket(sock, &byte, 1, CPPHTTPLIB_RECV_FLAGS);
+    if (n <= 0) { return std::string(); }
+    line += byte;
+#ifndef CPPHTTPLIB_ALLOW_LF_AS_LINE_TERMINATOR
+    if (prev == '\r' && byte == '\n') { break; }
+    prev = byte;
+#else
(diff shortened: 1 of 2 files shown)


ngxson:
Since we regularly pull upstream httplib source code, this patch will be overwritten the next time we do so.

Would it work if you specify http threads to a very large number? Maybe 64 threads?

rgerganov:
> Since we regularly pull upstream httplib source code, this patch will be overwritten the next time we do so.
> 
> Would it work if you specify http threads to a very large number? Maybe 64 threads?

Unfortunately no because we are deploying `llama.cpp` endpoints in k8s environments where there is no upper bound for the number of parallel requests and we observe constant restarts.

Out of curiosity, how do you solve this issue for HF endpoints of `llama.cpp`?

ggerganov:
I think there is a solution without changing `httplib`.

Here, we basically execute the handlers on the HTTP threads:

https://github.com/ggml-org/llama.cpp/blob/58c81f7e811d82b9ba62002f40d7af56134b61fc/tools/server/server-http.cpp#L389-L419

Instead, we need to free the http threads ASAP. To do that, they need to push the incoming requests into a work queue. The `server_http_context` will then have a separate "work" thread pool that processes this work queue.

The handlers for certain endpoints, such as `/health` will not forward the requests to the work queue. They will immediately respond.

This way, the `--threads-http` argument would now become the size of the "work" thread pool. And the actual HTTP thread pool of the `httplib` instance is no longer necessary to be large. Even just 1 to 4 threads would be enough to serve it efficiently in all cases.

rgerganov:
> Instead, we need to free the http threads ASAP. To do that, they need to push the incoming requests into a work queue. The `server_http_context` will then have a separate "work" thread pool that processes this work queue.

This may not work because `httplib` is closing the socket after executing the callback:

https://github.com/ggml-org/llama.cpp/blob/58c81f7e811d82b9ba62002f40d7af56134b61fc/vendor/cpp-httplib/httplib.cpp#L7685-L7690

ggerganov:
> > Instead, we need to free the http threads ASAP. To do that, they need to push the incoming requests into a work queue. The `server_http_context` will then have a separate "work" thread pool that processes this work queue.
> 
> This may not work because `httplib` is closing the socket after executing the callback:
> 
> https://github.com/ggml-org/llama.cpp/blob/58c81f7e811d82b9ba62002f40d7af56134b61fc/vendor/cpp-httplib/httplib.cpp#L7685-L7690

Ah bummer.

ngxson:
@rgerganov on HF endpoints, we simply set the number of http threads to `max(16, n_parallel * 2)`, so we ensure that there are minimum 16 threads

Another idea is to allow spawn dynamic amount of threads, which is a newly-added feature in httplib (and websocket is also built-in now, quite nice!) https://github.com/yhirose/cpp-httplib/pull/2368

rgerganov:
closing this in favor of #20817

PR #20800, opened 2026-03-20
chat : add content-only fallback for JSON_NATIVE tool parser

jpohhhh:
For templates like Llama 3.3 where tool_start is "{" (no distinctive marker), the content parser stops at any brace and the tools parser takes over. If the model output contains braces that aren't valid tool calls, the tools parser fails with nothing to absorb the remaining input.

Two failure modes on current master:
  - Content silently truncated at first "{" (partial match)
  - HTTP 500 crash (full parse throws, per #20729 demo of new chat parser core regression)

Fix: wrap the existing parser in a choice() with a content-only fallback. The tools path is tried first; when it fails, the fallback returns everything as content. No behavior change for valid tool calls.

Unit test:

  ./build/bin/test-chat

Server repro (Llama 3.2 3B, temp=0, tools enabled):

  llama-server -m Llama-3.2-3B-Instruct-Q4_K_M.gguf --jinja

    "messages": [{"role": "user", "content": "Write a hello world C program. Just the code, no explanation."}],
    }}],
  }'

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -208,8 +208,21 @@ common_peg_parser analyze_tools::build_tool_parser_json_native(parser_build_cont
         tool_start = format.per_call_start;
     }
 
-    return ctx.reasoning_parser + (force_tools ? p.eps() : p.optional(p.content(p.until(tool_start)))) + tools_parser +
-           p.end();
+    if (force_tools) {
+        return ctx.reasoning_parser + tools_parser + p.end();
+    }
+
+    // For templates like Llama 3.3 where tool_start is "{" (no distinctive marker),
+    // content stops at any brace and tools_parser takes over. If the model output
+    // contains JSON that isn't a valid tool call (e.g. user asked for a JSON schema),
+    // the tools parser fails with nothing to absorb the remaining input.
+    //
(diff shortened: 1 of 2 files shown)


pwilkin:
Fixed by https://github.com/ggml-org/llama.cpp/pull/20289.

jpohhhh:
> Fixed by #20289.

The fix isn't working, it is optional, so the regression persists. Any response containing `{` causes crash with Llama 3.x. please run tests in PR, or in description.

pwilkin:
The tests are irrelevant since they don't force the pure content parser.

jpohhhh:
> The tests are irrelevant since they don't force the pure content parser.

Am I understanding correctly:

This works as intended: all Llama 3.x requests with tools calls and responses containing { should return 500 unless A) the { expands into one of the tools B) the user passes an option.

jpohhhh:
Spent 20 minutes thinking on it, realized that can't be the case, because:

- the argument can only work with the server binary, not API
- disabling tool calls altogether works around the issue, but doesn't resolve it - i.e. user shouldn't have to disable tool calling altogether to be allowed text responses with {

I'll open a new PR with description based on that, per your comment re: unclear issue descriptions, I assume this falls into that bucket and thus that would be best practice.

PR #20152, opened 2026-03-06
CANN: add BF16 support for core operators

hipudding:
Add BF16 (bfloat16) type support to the CANN backend for the following operators: MUL_MAT, GET_ROWS, SET_ROWS, CPY, CONT. This enables BF16 models to run on Ascend NPUs.

*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

Issue #18797, opened 2026-01-13
Eval bug: The inference speed of the 310P3 is too slow for Q4 models.


younger-diao:
### Name and Version

built with GNU 10.3.1 for Linux aarch64 cann8.3.RC2  

### Operating systems

Linux

### GGML backends

CPU

### Hardware

8*310P3

### Models

./llama-cli -m /models/Qwen3-4B-Q4_K_M.gguf -p "你是谁" -ngl 99

### Problem description & steps to reproduce

I'm running Qwen3-4B-Q4_K_M.gguf on the 310P3. The model weights and memory are successfully loaded across all 8 NPU cards, but the inference speed is extremely slow—only 2.45 tokens per second. The NPU utilization (NA) remains at 0, which makes it feel like the model is actually running on the CPU, even though the weights have already been loaded into NPU memory. In contrast, running the FP16 GGUF version works fine without this issue.

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

younger-diao:
Is it because the current version of llama.cpp (with CANN backend) only supports FP16 (half-precision floating-point) NPU-accelerated computation(310P3) and does not support accelerated computation for Q4 (integer) quantized models?

#16269 I see here that the 310P3 does not support quantization.

taronaeo:
As per this

https://github.com/ggml-org/llama.cpp/blob/076b0faf7ddbcf9c1c68b90be2a47497a57141c2/ggml/src/ggml-cann/ggml-cann.cpp#L2287-L2304

310P does not support Q4 and Q8. Q4_K_M isn't listed in the support as well.

taronaeo:
Looks like I have answered your question and will be marking this issue as resolved now. Please feel free to reopen if otherwise.

PR #20801, opened 2026-03-20
CANN: update docker images to 8.5.0 and improve CANN.md

KokerZhou:
This PR includes small updates to CANN-related components:

- update docker images from rc versions to stable 8.5.0
  - this helps ensure a more reliable environment and enables further optimization using newer ops available in stable releases

- update CANN.md to clarify device naming
  - the original document refers to "Atlas 300T A2"
  - however, only "Atlas 300I A2" appears to be officially available
  - the documentation has been adjusted accordingly

These changes are limited to CANN environment setup and documentation.
No functional behavior outside of CANN is affected.

### DIFF
--- a/.devops/llama-cli-cann.Dockerfile
+++ b/.devops/llama-cli-cann.Dockerfile
@@ -1,4 +1,4 @@
-ARG ASCEND_VERSION=8.1.RC1.alpha001-910b-openeuler22.03-py3.10
+ARG ASCEND_VERSION=8.5.0-910b-openeuler22.03-py3.10
 
 FROM ascendai/cann:$ASCEND_VERSION AS build
 
--- a/.devops/cann.Dockerfile
+++ b/.devops/cann.Dockerfile
@@ -4,7 +4,7 @@
 
 # Define the CANN base image for easier version updates later
 ARG CHIP_TYPE=910b
-ARG CANN_BASE_IMAGE=quay.io/ascend/cann:8.3.rc2-${CHIP_TYPE}-openeuler24.03-py3.11
+ARG CANN_BASE_IMAGE=quay.io/ascend/cann:8.5.0-${CHIP_TYPE}-openeuler24.03-py3.11
 
 # ==============================================================================
 # BUILD STAGE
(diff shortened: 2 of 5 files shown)


hipudding:
Thank you very much for your submission. This is something we've been wanting to do recently. llama.cpp only verifies the latest CAN version, and timely updates to the CAN version will result in better compatibility.

hipudding:
BTW. bf16 is supported by #20152, could you please also add this data type in CANN.md? Thanks.

hipudding (approved):

KokerZhou:
Clarified BF16 support for 910B. Leave it empty for 310P.
On 310P, BF16/Q8_0/Q4_0 run but may be slower than FP16, as they do not leverage NPU acceleration (likely CPU-backed), as reported in #18797.
BTW, it seems that `aclnnWeightQuantBatchMatmulV2` can now work on 310P under certain conditions. The issue mentioned there can be handled carefully later.

hipudding:
> Clarified BF16 support for 910B. Leave it empty for 310P. On 310P, BF16/Q8_0/Q4_0 run but may be slower than FP16, as they do not leverage NPU acceleration (likely CPU-backed), as reported in #18797. BTW, it seems that `aclnnWeightQuantBatchMatmulV2` can now work on 310P under certain conditions. The issue mentioned there can be handled carefully later.

Thanks for your feedback. Q8 and Q4 are slow even it run on NPU, not recommend to use 310p with these dtypes.

CISC (approved):

.devops/llama-cli-cann.Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG ASCEND_VERSION=8.1.RC1.alpha001-910b-openeuler22.03-py3.10
+ARG ASCEND_VERSION=8.5.0-910b-openeuler22.03-py3.10
```
CISC, line 1: Is this intentionally different to the others?

KokerZhou: This was created in b5086, whereas the other pipelines were introduced after b6054. I believe this difference is unintentional.

PR #20803, opened 2026-03-20
[SYCL] Support bf16 and quantized type of MUL_MAT

arthw:
Fix issue: https://github.com/ggml-org/llama.cpp/issues/20713

Enable BF16 and quantized type of MUL_MAT.
All related UT cases are passed on Arc770 and UHD Graphics 770  of 13700K.


arthw:
@ggerganov Please help review!

ggerganov (approved):

0cc4m:
So the actual support was already there and just deactivated? Why?

arthw:
> So the actual support was already there and just deactivated? Why?

Some broken user cases are fixed, but there is no update for the correct case, I guess.

I will fix the left unsupported cases later.

PR #20804, opened 2026-03-20
chat : add Granite 4.0 chat template with correct tool_call role mapping

jesus-talavera-ibm:
## Summary

Add `LLM_CHAT_TEMPLATE_GRANITE_4_0` as a parallel path to the existing Granite 3.x template, fixing `assistant_tool_call` role mapping for Granite 4.0 models when `--jinja` is not used.

### Problem

The Granite 4.0 Jinja template maps `assistant_tool_call` to `<|start_of_role|>assistant<|end_of_role|><|tool_call|>`, but the C++ template handler emits the literal role `<|start_of_role|>assistant_tool_call<|end_of_role|>` which the model does not recognize. This silently breaks tool calling on the C++ template path (without `--jinja`).

### Changes

- Rename `LLM_CHAT_TEMPLATE_GRANITE` → `LLM_CHAT_TEMPLATE_GRANITE_3_X` (preserves existing 3.x behavior unchanged)
- Add `LLM_CHAT_TEMPLATE_GRANITE_4_0` enum, map entry (`"granite-4.0"`), and handler
- Detection: `<|start_of_role|>` + (`<tool_call>` or `<tools>`) → 4.0; otherwise → 3.x
- Add production Granite 4.0 Jinja template (shared by granite-4.0-tiny, h-small, micro)
- Add tests for both 3.x and 4.0 template paths (C++ and Jinja) in `test-chat-template.cpp`
- Add Granite 4.0 peg parser test in `test-chat.cpp`

### Files changed

| File | Change |
|------|--------|
| `src/llama-chat.h` | Add `GRANITE_3_X` / `GRANITE_4_0` enum values |
| `src/llama-chat.cpp` | Map entries, detection logic, and parallel handlers |
| `models/templates/ibm-granite-granite-4.0.jinja` | Production Granite 4.0 Jinja template |
| `tests/test-chat-template.cpp` | 3.x backwards compat + 4.0 fix tests (C++ and Jinja) |
| `tests/test-chat.cpp` | Granite 4.0 peg parser test |

## Test plan

- [x] `test-chat-template` — Granite 3.x backwards compat (C++ + Jinja): PASS
- [x] `test-chat-template` — Granite 4.0 fix (C++ + Jinja): PASS
- [x] `test-chat` — Granite 3.3 peg parser (reasoning + basic chat): PASS
- [x] `test-chat` — Granite 4.0 peg parser (basic chat + tool calling): PASS
- [x] Regression tested with actual model weights: Granite 3.0, 3.1, 3.3, 4.0 (basic chat, tool calling with --jinja, C++ path without --jinja)

EDIT: I used Claude Code as start point and then iterated from it manually. You can check the whole process here: https://github.com/jesus-talavera-ibm/llama.cpp/pull/1


gabe-l-hart:
>🤖 Generated with [Claude Code](https://claude.com/claude-code)

In general, this project doesn't accept fully generated PRs. For context, @jesus-talavera-ibm and I iterated on this on https://github.com/jesus-talavera-ibm/llama.cpp/pull/1, so there has been significant manual review of the generated code.

src/llama-chat.cpp
```diff
@@ -59,7 +59,8 @@ static const std::map<std::string, llm_chat_template> LLM_CHAT_TEMPLATES = {
     { "exaone4",           LLM_CHAT_TEMPLATE_EXAONE_4          },
     { "exaone-moe",        LLM_CHAT_TEMPLATE_EXAONE_MOE        },
     { "rwkv-world",        LLM_CHAT_TEMPLATE_RWKV_WORLD        },
-    { "granite",           LLM_CHAT_TEMPLATE_GRANITE           },
+    { "granite",           LLM_CHAT_TEMPLATE_GRANITE_3_X       },
```
gabe-l-hart, line 62: The goal of this change is to carefully delineate between template for the 3.x series and the 4.0 series (there may be further 4.x template changes in the future, thus being explicit about `4_0`).

As written, this is nicely backwards-compatible since the enum name is only used internally and the string name is mapped consistently.

tests/test-chat-template.cpp
```diff
@@ -679,6 +679,155 @@ int main_automated_tests(void) {
         }
     }
 
+    // Test Granite 3.x template (LLM_CHAT_TEMPLATE_GRANITE_3_X) — backwards compatibility
```
gabe-l-hart: I should have caught this in pre-review, but I think we should try to rewrite these tests to match the ones above that all follow a standard format and use common test logic.

gabe-l-hart (changes requested):
I think we should refactor the `test-chat-template.cpp` changes to match the other tests, but otherwise this looks good!

tests/test-chat-template.cpp
jesus-talavera-ibm: Thank you, Gabe! I've adapted the tests here: https://github.com/ggml-org/llama.cpp/pull/20804/changes/3f2a48ab7554a663b2ebbc86fb8c124d8bedd85d

gabe-l-hart (approved):
I think this change looks good.

@pwilkin @ngxson @ggerganov We've validated this against both 3.x and 4.0 models and the code/tests look tight, so I think it's ready for CI and maintainer review.

pwilkin (approved):

ggerganov (approved):

PR #20805, opened 2026-03-20
chat : add EOS token to additional_stops for autoparser templates 

jpohhhh:
Some models emit the EOS token as text (e.g. </s>) rather than as the special EOS token ID. The PEG parser fails at end-of-input because the trailing EOS text isn't consumed.

Fix: add the template's EOS token to additional_stops so the server strips it before the output reaches the parser.

Unit test:

  ./build/bin/test-chat

Server repro (bartowski/mistralai_Mistral-Small-3.2-24B-Instruct-2506-GGUF, temp=0):

  llama-server -m Mistral-Small-3.2-24B-Instruct-2506-IQ2_M.gguf --jinja

    "messages": [{"role": "user", "content": "Weather in Tokyo?"}],
    }}],
  }'

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -46,6 +46,14 @@ common_chat_params peg_generator::generate_parser(const common_chat_template &  
     data.format           = COMMON_CHAT_FORMAT_PEG_NATIVE;
     data.preserved_tokens = autoparser.preserved_tokens;
 
+    // Some models emit the EOS token as text (e.g. </s>) rather than as the
+    // special EOS token ID. Add it as a stop so the server strips it before
+    // the parser sees it. Without this, the trailing EOS text causes the
+    // parser to fail at end-of-input.
+    if (!tmpl.eos_token().empty()) {
+        data.additional_stops.push_back(tmpl.eos_token());
+    }
+
     auto parser = autoparser.build_parser(inputs);
     data.parser = parser.save();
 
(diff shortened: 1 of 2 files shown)


pwilkin:
Yet another bad change for a non-issue that masks a real issue.

If a model actually emits an EOS token "as a normal token", that's a tokenizer error that needs to be fixed, not masked in the parser.

jpohhhh:
> Yet another bad change for a non-issue that masks a real issue.
> 
> If a model actually emits an EOS token "as a normal token", that's a tokenizer error that needs to be fixed, not masked in the parser.

You're being abusive. 

I am finding issues with the models I support, making sure they repro with server binary *and* my API, and then making sure they repro with a unit test. These are not non-issues :( Why are you sooooo mean, all the time?

jpohhhh:
cc @bartowski1182 , bartowski/mistralai_Mistral-Small-3.2-24B-Instruct-2506-GGUF can't be handled by new parser. Any advice, here? I don't know how to square the circle here.

pwilkin:
No, you're being abusive by spamming PRs without opening issues with actual real-life reproductions. Please stop it.

jpohhhh:
> No, you're being abusive by spamming PRs without opening issues with actual real-life reproductions. Please stop it.

I don't know what you mean, I'm holding myself to a *really* strict standard - it *has* to repro with the llama-server binary, and I *have* to give the exact command and bisect, and have unit tests. 

Please tell someone who can ban me from PRs your claim that I'm spamming PRs without repros and have them evaluate your claim. I'm happy to get banned if they agree.

ggerganov:
@jpohhhh I'm sorry, but I decided to block you for 30 days. Would recommend to change your approach if you wish to contribute to the project.

PR #20806, opened 2026-03-20
chat : fix Llama 3.x throwing runtime exception if response contains {

jpohhhh:
**NOTE**: I'm aware the server *binary* has a flag to disable tool call parsing altogether, @pwilkin mentioned it when closing #20800. Opened this PR because: that cannot help API callers, and this is a severe regression- all Llama 3.x requests with tools throw runtime exception if response is freeform and contains {, ex. a hello world C program). This PR's description clarifies that, the other theoratically left open to option to close if the report was server-binary-only and the PR contributor was amenable to disabling all tool calls with Llama 3.x.

For templates like Llama 3.3 where tool_start is "{" (no distinctive marker), the content parser stops at any brace and the tools parser takes over. If the model output contains braces that aren't valid tool calls, the tools parser fails with nothing to absorb the remaining input. Ex. "write me a C program" 500s without starting the server with ` --skip-chat-parsing`. That's fine for the server if _a priori_ you know llama 3.x will be used with the server and can afford to disable tool call altogether. It won't work for API users.

Two failure modes on current master:
  - Content silently truncated at first "{" (partial match)
  - server: HTTP 500 crash (full parse throws), server API: runtime exception 

Fix: wrap the existing parser in a choice() with a content-only fallback. The tools path is tried first; when it fails, the fallback returns everything as content. No behavior change for valid tool calls.

Unit test:

  ./build/bin/test-chat

Server repro (Llama 3.2 3B, temp=0, tools enabled):

  llama-server -m Llama-3.2-3B-Instruct-Q4_K_M.gguf --jinja

    "messages": [{"role": "user", "content": "Write a hello world C program. Just the code, no explanation."}],
    }}],
  }'

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -208,8 +208,21 @@ common_peg_parser analyze_tools::build_tool_parser_json_native(parser_build_cont
         tool_start = format.per_call_start;
     }
 
-    return ctx.reasoning_parser + (force_tools ? p.eps() : p.optional(p.content(p.until(tool_start)))) + tools_parser +
-           p.end();
+    if (force_tools) {
+        return ctx.reasoning_parser + tools_parser + p.end();
+    }
+
+    // For templates like Llama 3.3 where tool_start is "{" (no distinctive marker),
+    // content stops at any brace and tools_parser takes over. If the model output
+    // contains JSON that isn't a valid tool call (e.g. user asked for a JSON schema),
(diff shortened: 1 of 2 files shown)


pwilkin:

jpohhhh:
> Please stop this endless spam of PRs to imaginary issues you cannot reproduce with real-life scenarios. Open an issue with an actual model query or message history first.

See server commands, those are an actual model query.

Moving forward, it is clear you are intending to communicate "file an issue with the server one liners before the PR", which I will do.

PR #20808, opened 2026-03-20
peg : return NEED_MORE_INPUT for partial delimiter match in LENIENT mode

jpohhhh:
until() returned SUCCESS when a delimiter was partially matched at end
of input (e.g. '<' partial-matching '</think>'). In LENIENT (streaming)
mode this is incorrect: we don't yet know if the partial match will
complete or if it's just content that starts with the same characters.

Noticed while auditing streaming behavior for templates with
<think>...</think> reasoning tags. Model output containing '<'
(comparisons, HTML, code) would cause the until() parser to mark the
AST node as definitive when it should be tentative. Currently masked
by the outer sequence, but becomes observable when until() is terminal
in a sequence or when choice() selects a branch based on the SUCCESS.

Test:

  ./build/bin/test-chat" && git push --force origin fix-until-partial-match-lenient

### DIFF
--- a/common/peg-parser.cpp
+++ b/common/peg-parser.cpp
@@ -771,7 +771,14 @@ struct parser_executor {
             }
 
             if (match == trie::PARTIAL_MATCH) {
-                // Found a partial match extending to end of input, return everything before it
+                // Found a partial match extending to end of input, return everything before it.
+                // In LENIENT (streaming) mode, the partial delimiter match means we don't yet
+                // know whether this is actually the delimiter or just content that happens to
+                // start with the same characters (e.g. "<" partial-matching "</think>").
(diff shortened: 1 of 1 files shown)


pwilkin:

jpohhhh:
> Please stop this endless spam of PRs to imaginary issues you cannot reproduce with real-life scenarios. Open an issue with an actual model query or message history first.

Noting for the record you been claiming the issues are hallucinated, or fake, and 3rd parties have had to step in to get constructive responses from you, but you end up rewording that same idea politely. 

This PR was something I noticed while working through *another* bug, so it doesn't have server commands. But every single other one does. 

Moving forward, it is clear you are intended to communicate "file an issue with the server one liners before the PR", which I will do as I continue helping fix the errors I can repro.

PR #20811, opened 2026-03-20
fix(opencl): free host buffers after MoE MXFP4 router table reorder

buley:
## Summary

In `ggml/src/ggml-opencl/ggml-opencl.cpp`, the MoE MXFP4 gemm path allocates two host buffers (`host_src2_reorder` and `host_src2`) to preprocess the router table, then copies the data into a CL buffer via `CL_MEM_COPY_HOST_PTR`. After the kernel launch, the CL memory objects are released but the two host `malloc` buffers are never freed.

This leaks `ne20 * ne21 * 4 * num_tiles_per_expert * sizeof(short)` + `ne21 * nb21` bytes on every MoE matmul dispatch on OpenCL/Adreno.

## Fix

Added `free(host_src2)` and `free(host_src2_reorder)` after the CL memory objects are released and before the `return`.

## AI Disclosure

The bug was identified using static analysis tooling with AI assistance (pattern matching for `malloc` without corresponding `free` in the same scope). The two-line fix was written by hand after reading the surrounding code to verify the buffers are not used after `clCreateBuffer` copies their contents.

### DIFF
--- a/ggml/src/ggml-opencl/ggml-opencl.cpp
+++ b/ggml/src/ggml-opencl/ggml-opencl.cpp
@@ -10960,6 +10960,8 @@ static void ggml_cl_mul_mat_id(ggml_backend_t backend, const ggml_tensor * src0,
                 CL_CHECK(clReleaseMemObject(src1_sub_buffer));
                 CL_CHECK(clReleaseMemObject(buf_src1_image));
                 CL_CHECK(clReleaseMemObject(buf_src2));
+                free(host_src2);
+                free(host_src2_reorder);
                 return;
             } // else fallback to generic kernel
 #endif // GGML_OPENCL_USE_ADRENO_KERNELS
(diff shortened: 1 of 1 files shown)


lhez:
@buley The CI shows this PR fails to build. Did you or the AI agent try building and running it?

buley:
@lhez thought so, on it, sorry for the noise.

buley:
@lhez You were right to call this out. I missed the basic verification step here: the original two-line fix compiled in my head, but not in the Adreno-specific OpenCL path, because I placed the `free()` calls outside the branch where `host_src2` and `host_src2_reorder` are declared.

I’ve pushed a remediation in `0f98d26` that frees `host_src2` immediately after the router-table reorder finishes and frees `host_src2_reorder` immediately after `clCreateBuffer(..., CL_MEM_COPY_HOST_PTR, ...)` copies it, so the leak is fixed without breaking the specialized build.

That gap is on me. I don't have tools for that environment but I know better than to leave the PR unchecked, and I'm sorry my excitement about finding a bug overcame my get 'er done. Rerunning CI now.

Issue #20787, opened 2026-03-20
CI : Releases blocked by s390x runner failure


CISC:
The s390x runners started randomly failing some hours ago:
https://github.com/ggml-org/llama.cpp/actions/runs/23309996974/job/67794402629

though there were some other strange errors preceding that:
https://github.com/ggml-org/llama.cpp/actions/runs/23302100350/job/67786180793

CISC:
@taronaeo Any idea?

taronaeo:
Let me check in awhile. We may have a communication error with the runners:

> The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

I'll investigate once I can sit down :)

taronaeo:
https://github.com/ggml-org/llama.cpp/actions/runs/23302100350/job/67786180793

This failure seem to be related to token permissions though. Any chance we can check if the token is still valid with the correct permissions?

CISC:
> https://github.com/ggml-org/llama.cpp/actions/runs/23302100350/job/67786180793
> 
> This failure seem to be related to token permissions though. Any chance we can check if the token is still valid with the correct permissions?

Let's worry about that one later, may be random (I've seen it before).

taronaeo:
Tracking: https://github.com/IBM/actionspz/issues/92

ggerganov:
Shall we disable these jobs to get the releases back up?

pwilkin:
Yeah I think that'd be recommended, I'm going to get swarmed with conflicting parser reports otherwise :)

CISC:
Resolved:
https://github.com/ggml-org/llama.cpp/actions/runs/23354518496/job/67941807564

PR #20813, opened 2026-03-20
releases : disable s390x builds

ggerganov:
ref https://github.com/ggml-org/llama.cpp/issues/20787#issuecomment-4099727984

Not sure if this is all that is failing atm


CISC:
Should be resolved:
https://github.com/IBM/actionspz/issues/92#issuecomment-4099035138

CISC:
https://github.com/ggml-org/llama.cpp/actions/runs/23354518496/job/67941807564

PR #20818, opened 2026-03-20
Add shader count for Intel Arc Pro B60

TheBlueMatt:


ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -16043,6 +16043,8 @@ static uint32_t ggml_vk_intel_shader_core_count(const vk::PhysicalDevice& vkdev)
     case 0x56A0:  // A770
     case 0x5690:  // A770M
         return 32;
+    case 0xE211:  // Pro B60
```
0cc4m: Just move this line below the B580 one, they can share the return call.

0cc4m (commented):
I can't believe I forgot the B60, thank you.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
TheBlueMatt: Done.

0cc4m (approved):
LGTM

gustrd:
Just being curious, what is the performance gain for those?

I tested with:

```
    case 0x5693:  // A370M
    case 0x56A5:  // A380
    case 0x56B1:  // Pro A40/A50
    case 0x64A0:  // Arc 140V (Lunar Lake) <---
        return 8;
```

But measured no performance gain at all.

TheBlueMatt:
This indeed makes almost no difference for me, but moving to using shared memory scratch space gave me something like an 80% speedup, see https://github.com/ggml-org/llama.cpp/issues/18808#issuecomment-4101794338

PR #20726, opened 2026-03-18
server : improve mtmd ctx checkpoints

ggerganov:
fix #20720
cont #20232

- Fix logic for end-of-prompt-checkpoints in the presence of mtmd chunks
- Checkpoint spacing is now based on number of tokens, not on positions
- Fix off-by-one in `pos_min_thold` logic

Issue #20697, opened 2026-03-17
Feature Request: Disk-based context checkpoint offloading (`--cache-disk`)


wypiki:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

## Problem

Context checkpoints (`--ctx-checkpoints`, `--cache-ram`) are currently stored exclusively in system RAM. On UMA (Unified Memory Architecture) systems such as AMD Strix Halo, system RAM and VRAM are physically the same memory pool. This means checkpoint caching actively competes with model weights and KV cache for the same limited resource, rather than providing relief.

On these systems, `--cache-ram` doesn't offload anything – it just reserves more of the same unified memory that the model already needs, reducing the effective memory available for context or GPU layers.

## Proposed Solution

Add a `--cache-disk <path>` option (or `--cache-ssd`) that allows context checkpoints to be serialized to a fast local storage device (NVMe SSD) instead of being held in RAM.

Modern NVMe SSDs deliver 5-7 GB/s sequential read throughput, which would allow restoring a typical checkpoint (50-100 MiB) in ~10-20ms – negligible compared to the seconds or minutes saved by avoiding full prompt re-processing on hybrid/recurrent models like Qwen 3.5.

### Suggested interface

```
--cache-disk <path>       directory for disk-based checkpoint storage (default: disabled)
--cache-disk-max <MiB>    maximum disk space for checkpoints (default: unlimited)
```

When `--cache-disk` is set, checkpoints would be written to disk instead of (or in addition to) RAM. On restore, the checkpoint is read back from disk into the appropriate memory (GPU/unified).

## Use Case

This is particularly relevant for:

1. **UMA/APU systems** (AMD Strix Halo, Strix Point, future Apple Silicon with large models) where RAM = VRAM and every megabyte of RAM used for checkpoints is a megabyte less for model/context
2. **Multi-user systems**: this kind of almost unlimited caching would make serving multiple users with different system prompts or several agents simultaniously fast on systems with slow prompt processing speeds
3. **Large context windows** (100k+ tokens) where checkpoints can grow large and numerous

[...]

PR #20819, opened 2026-03-20
server: persist context checkpoints across slot save/restore

European-tech:
## Summary

The `/slots` save/restore API persists the raw KV+recurrent memory and the token list, but does **not** persist the context checkpoints stored in `server_prompt::checkpoints`.

For hybrid/recurrent models (Qwen3.5, Jamba, Falcon-H1), these checkpoints are essential: without them, `update_slots()` cannot find a valid recurrent state snapshot and falls back to **full prompt re-processing** on every slot reuse.

This PR adds two helper functions that write/read a small companion file (`<filename>.checkpoints`) next to the main slot save file, using a versioned binary format with a magic header.

## Motivation

In router mode with `--models-max 1`, switching between models destroys the in-memory prompt cache (the child process is killed). Users can call `/slots/0?action=save` before a swap and `/slots/0?action=restore` after, but the restored slot had no checkpoints — triggering a full re-processing of the entire prompt.

For a 26K-token conversation on Qwen3.5-27B (16 attention + 48 recurrent layers), this meant **23 seconds** of wasted compute on every model swap return.

## Results

Tested with Qwen3.5-27B in router mode (save → swap to MoE → swap back → restore → same prompt):

| Metric | Before | After |
|---|---|---|
| `cache_n` after restore | 0 | **26549** |
| Prompt processing | 23,000 ms | **75 ms** |
| Subsequent turns | 23,000 ms | **159 ms** |
| Save overhead | — | 675 ms (826 MB state + 749 MB checkpoints) |
| Restore overhead | — | 632 ms |

## Changes

- `server-context.cpp`: Two static helper functions (`slot_checkpoints_save`, `slot_checkpoints_load`) + two one-line call sites in the existing `SLOT_SAVE` / `SLOT_RESTORE` handlers.

## Backward compatibility

- Old save files without a `.checkpoints` companion are loaded normally (the load helper returns `true` when the file is absent).
- Non-hybrid models that have no checkpoints skip the save entirely.

## File format

```
uint32  magic    = 0x4C4C4350  ("LLCP")
uint32  version  = 1
uint32  n_checkpoints
Per checkpoint:
  int32   pos_min
  int32   pos_max
  int64   n_tokens
  uint64  data_size
  uint8   data[data_size]
```

## Disclosure

AI tools were used in an assistive capacity during the development of this patch. The code was manually reviewed, tested, and validated on real hardware (RTX 3090, Qwen3.5-27B + Qwen3.5-35B-A3B MoE in router mode).

## Related

- #20697 — feature request for `--cache-disk` (persistent checkpoints)
- #16391 — host-memory prompt caching
- #20726 — hybrid model checkpoint improvements

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -455,6 +455,127 @@ struct server_slot {
     }
 };
 
+//
+// checkpoint persistence helpers for hybrid/recurrent models
+//
+// Hybrid models (e.g. Qwen3.5, Jamba, Falcon-H1) use recurrent layers whose
+// state cannot be partially restored from the KV cache alone.  The server
+// creates "context checkpoints" during prompt processing that snapshot the
+// full recurrent state at regular intervals.  These checkpoints live in
+// server_prompt::checkpoints and are essential to avoid a full prompt
+// re-processing when the slot is reused.
+//
+// The built-in /slots save/restore API persists the raw KV+recurrent memory
+// via llama_state_seq_{save,load}_file, but does NOT persist the checkpoint
+// metadata.  The two helpers below fill that gap: they write/read a small
+// companion file (<filename>.checkpoints) next to the main slot save file.
+//
+// File format (binary, little-endian):
+//   uint32  magic    = 0x4C4C4350  ("LLCP")
+//   uint32  version  = 1
+//   uint32  n_checkpoints
+//   For each checkpoint:
+//     int32   pos_min
+//     int32   pos_max
+//     int64   n_tokens
+//     uint64  data_size
+//     uint8   data[data_size]
+//
+
+static bool slot_checkpoints_save(const std::string & filepath,
+                                  const std::list<server_prompt_checkpoint> & checkpoints) {
+    if (checkpoints.empty()) {
+        return true;
+    }
+
+    const std::string cp_path = filepath + ".checkpoints";
+    FILE * fp = fopen(cp_path.c_str(), "wb");
+    if (!fp) {
+        SRV_WRN("failed to open checkpoint file for writing: %s\n", cp_path.c_str());
+        return false;
+    }
+
+    const uint32_t magic   = 0x4C4C4350;
+    const uint32_t version = 1;
+    const uint32_t n_cp    = (uint32_t) checkpoints.size();
+
+    bool ok = true;
+    ok = ok && fwrite(&magic,   sizeof(magic),   1, fp) == 1;
+    ok = ok && fwrite(&version, sizeof(version), 1, fp) == 1;
+    ok = ok && fwrite(&n_cp,    sizeof(n_cp),    1, fp) == 1;
+
+    for (const auto & cp : checkpoints) {
+        const uint64_t data_size = cp.data.size();
+        ok = ok && fwrite(&cp.pos_min,  sizeof(cp.pos_min),  1, fp) == 1;
+        ok = ok && fwrite(&cp.pos_max,  sizeof(cp.pos_max),  1, fp) == 1;
+        ok = ok && fwrite(&cp.n_tokens, sizeof(cp.n_tokens), 1, fp) == 1;
+        ok = ok && fwrite(&data_size,   sizeof(data_size),   1, fp) == 1;
+        if (data_size > 0) {
+            ok = ok && fwrite(cp.data.data(), 1, data_size, fp) == data_size;
+        }
+    }
+
+    fclose(fp);
+
+    if (!ok) {
+        SRV_WRN("failed to write checkpoint data to %s\n", cp_path.c_str());
+        std::remove(cp_path.c_str());
+        return false;
+    }
+
+    SRV_INF("saved %u context checkpoints to %s\n", n_cp, cp_path.c_str());
+    return true;
+}
+
+static bool slot_checkpoints_load(const std::string & filepath,
+                                  std::list<server_prompt_checkpoint> & checkpoints) {
+    const std::string cp_path = filepath + ".checkpoints";
+    FILE * fp = fopen(cp_path.c_str(), "rb");
+    if (!fp) {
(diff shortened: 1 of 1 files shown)


llukas:
Not sure how this can be made to stick to guidelines wrt AI generated code even in new PR. Just leaving this info:

For use case: load prompt, save slot, shutdown llama-server, start llama-server, load slot, continue with prompt this requires part of #20955 to work (20955 minus auto-save/restore basically)

dark-penguin:
@European-tech This is really cool! And it worked right up to 68e7ea3ea , which broke it with a major refactor. But some of the updated models don't work on such an old version (Qwen3.6-27B). I would be really grateful if you could update this so I can cherry-pick it into my own fork - you'd probably have to update sooner or later anyway.

dark-penguin:
I think I've managed to rebase these two commits with a lot of help from an AI (for those who want it too).

calebrio02:
## Confirmed: works perfectly in production with Qwen3.6-27B + ROCm

Cherry-picked commit `d5c3250` from this PR on top of my llama.cpp fork. I can confirm this **completely solves the slot save/restore issue** for hybrid/recurrent models like Qwen3.6.

**Hardware:** AMD Ryzen 5 7600 + Radeon RX 7900 XT (ROCm)
**Model:** Qwen3.6-27B-Q4_K_M (16 attention + 48 recurrent layers)
**Context:** ~48k tokens

---

### BEFORE the PR (no checkpoints persisted)

Flow: save slot → shut down server → start server → restore slot → send continuation request. Result: **full re-prefill of all tokens every time**.

```
checking checkpoint with [1972, 1972] against 3...
checking checkpoint with [1460, 1460] against 3...
forcing full prompt re-processing due to lack of cache data (likely due to
  SWA or hybrid/recurrent memory, see https://github.com/ggml-org/llama.cpp/pull/13194)
erased invalidated context checkpoint (pos_min = 1460, pos_max = 1460, ...)
erased invalidated context checkpoint (pos_min = 1972, pos_max = 1972, ...)

prompt eval time = 301603.16 ms / 28367 tokens (10.63 ms/token, 94.05 tok/s)
total time = 303160.54 ms / 28396 tokens
```

**~303 seconds** (5 minutes) of full re-prefill on every restart.

---

### AFTER the PR (with persistent checkpoints)

Exact same flow: save slot → shut down server → start server → restore slot → send continuation request.

**On server startup (auto-restore):**
```
restored 4 context checkpoints from .../Qwen3.6-27B-Q4_K_M.checkpoints
auto-restored slot 0 (20480 tokens, 660.1 MiB)
```

**On manual snapshot restore:**
```
restored 13 context checkpoints from .../snapshot-20260607-111710.bin.checkpoints
```

**On continuation request:**
```
n_past = 48845, slot.prompt.tokens.size() = 48875
checking checkpoint with [48842, 48842] against 48845...
restored context checkpoint (pos_min = 48842, pos_max = 48842,
  n_tokens = 48843, n_past = 48843, size = 149.626 MiB)

prompt eval time = 612.31 ms / 31 tokens (19.75 ms/token, 50.63 tok/s)
total time = 3156.80 ms / 78 tokens
```

**~612ms** to process only the 31 new tokens (the full checkpoint state was restored instantly from the `.checkpoints` companion file).

---

### Improvement summary

| Metric | Before | After |
|---|---|---|
| Prefill after restore | **303,160 ms** (28k tokens, full re-process) | **612 ms** (31 new tokens only) |
| Checkpoint restore | ❌ Forced full re-process | ✅ Restored in milliseconds |
| Total time | ~5 minutes | ~3 seconds |
| Speedup | — | **~100x faster** |

---

### Integration note

I had to make a small type adaptation since my fork uses `common_prompt_checkpoint` with `data_tgt`/`data_dft` members instead of `server_prompt_checkpoint` with `data`. The changes were minimal (type/member renames) — the PR logic is 100% compatible.

Excellent work on this PR.

dark-penguin:
@calebrio02 Could you please review that [my AI-based rebase](https://github.com/dark-penguin/llama.cpp/tree/cherrypick/slot-save-restore) (on top of basically vanilla upstream) is reasonable and not overengineered? It seems to work, but it also works without the last fix in that branch, and I don't understand nearly enough in the internals of llama.cpp to conduct an educated review.

calebrio02:
@dark-penguin 
Hey, first off — huge thanks. I'll be completely honest: I don't know enough about llama.cpp's internals to give a proper code review. I'm just a user running Qwen3.6 on ROCm who was getting sick of waiting 5 minutes for the server to load after a restart.

I actually applied these patches with help from Claude and a Qwen3.6 35B A3B on hermes agent,  but I tracked exactly what they were doing in the logs. Here is what I saw:

Without your PR: Every restart wiped the memory, forcing a full 300s prefill.
With your PR: The server remembers everything. A continuation request comes back in about 600ms.
To answer your questions:

Is it overengineered? Definitely not. It's exactly what was needed.
About the "last fix" (truncation recovery): I don't see a need for it to make the main feature work. It looks like it only handles a very rare edge case where the prompt needs to be partially trimmed and the model fails. Your branch handles the normal save/restore flow perfectly.
So no, it's not overengineered. It's the exact solution we've been waiting for. Solid work!

dark-penguin:
Thank the PR author, not me! :smile: I barely managed to rebase it!

And really, huge thanks @European-tech ! It really is what we've been waiting for, and I can't believe it works so well!

calebrio02:
> Thank the PR author, not me! 😄 I barely managed to rebase it!
> 
> And really, huge thanks @European-tech ! It really is what we've been waiting for, and I can't believe it works so well!

You're totally right, my apologies for the phrasing — huge thanks to @European-tech! 🙌 This is literally the dealbreaker I've been waiting for to finally use local LLMs comfortably. Removing the pain of context loss is going to make local inference so much more user-friendly. Truly amazing work!

PR #20827, opened 2026-03-20
misc : prefer ggml-org models in docs and examples

ddh0:
Prefer using [ggml-org](https://huggingface.co/ggml-org) models in examples / help text rather than 3rd party uploaders.


ngxson (approved):

taronaeo (approved):

tests/.DS_Store
netrunnereve, line 1: Looks like this got added by accident, we should add this to the gitignore.

ddh0: Oops, thanks, removed that file.

ngxson (approved):

PR #20828, opened 2026-03-20
llama-quant : default ftype param `Q5_1` --> `Q8_0`

ddh0:
Change the default `ftype` in `llama_model_quantize_params` from `LLAMA_FTYPE_MOSTLY_Q5_1` to `LLAMA_FTYPE_MOSTLY_Q8_0`.

In case some external program naively uses the default quantization params, we should probably default to a known-good type like Q8_0 rather than Q5_1, which is rather old.


ddh0:
Thoughts on this change? @ggerganov

ggerganov (approved):

PR #20830, opened 2026-03-21
WebUI: fix edit msg form textarea height

### LINKED ISSUE #19940: Misc. bug: webui edit message textarea limited to 2 lines
### Name and Version

version: 8169 (0adfc2d64)
built with GNU 11.4.0 for Linux x86_64

### Operating systems

Linux

llama-server

### Command line

```shell
llama-server -m model.gguf
```

Previously (definitely before baa12f38) we got a generous up to 16 lines to edit, depending on the length of the chat message:

Image

Now we get only two, and that is not enough. I now have to copy and paste my prompts into a text editor if I want to change it because there is not enough space to edit the message:

Image

If you begin typing somewhere, it shoots up to 5 lines. (Why isn't it expanded from the start?)

Image

In either case, it's difficult to edit larger prompts with only a few lines.

### First Bad Commit

Probably https://github.com/ggml-org/llama.cpp/commit/baa12f3831d088503717a8f85c70340936992343 ?

### Relevant log output

_No response_

bluemoehre:
Fixes #19940

Improve UX by making `<textarea>` height consistent with regular rendered messages.

Changes:
- Call `autoResizeTextarea()` in `onMount()` to adjust height based on current value
- Remove the fixed 128px height constraint by removing the corresponding framework classes
- Apply the same custom height limit used in all regular rendered messages `var(--max-message-height)`

### Current behavior

https://github.com/user-attachments/assets/4a469cac-ab09-49cc-ac4b-940c361dabfc

### Adding these changes

https://github.com/user-attachments/assets/858eba3b-9d2e-4125-9b0c-c365cd7a9a45


allozaur (changes requested):
Please push rebuilt `index.html.gz` by running `npm run build` and commiting the file :) Also a short video demo would be nice in the description.

bluemoehre:
Sure! Here it is, @allozaur. I hope the videos help clarify things.

allozaur (approved):
@bluemoehre perfect, thank you. There's a conflict as the build file should be compiled after merge/rebase. So please push this one last thing and then we can merge it :)

bluemoehre:
@allozaur Yeh, it was pretty clear that there would be conflicts with the build file, so I wanted to wait for the "ready to take-off" approval of the changes. 

Perhaps it would be a good idea to let a repository owner configure automatic builds in the CI pipeline?

bluemoehre:
This is the way to not pull rebase using Github UI 🤷  ... better push everything in one step.

Apologies for any confusion.

allozaur (approved):

PR #20461, opened 2026-03-12
llama-bench: fix case where mmap and direct-io are turned on together

taronaeo:
Ref: https://github.com/ggml-org/llama.cpp/pull/20211#discussion_r2923702030

As mentioned in the thread referenced, users may inadvertently enable both `--mmap` and `--direct-io` together and notice no performance difference.

This PR adds additional safeguards during flag parsing to ensure that both `--mmap` and `--direct-io` cannot have the same enabled state. Also, moved the `ggml_backend_load_all` call just after the flag parsing function for `llama-bench` to fail early if any of the flags has an error.

### Test Cases

1. `--mmap` is enabled by default but the user specifies `--direct-io 1`

```console
$ build/bin/llama-bench -hf ibm-granite/granite-3.3-2b-instruct-GGUF:Q4_K_M --direct-io 1 

...truncated...
common_download_file_single_online: using cached file (same etag): /Users/taronaeo/Library/Caches/llama.cpp/ibm-granite_granite-3.3-2b-instruct-GGUF_granite-3.3-2b-instruct-Q4_K_M.gguf
| model                          |       size |     params | backend    | threads | mmap | dio |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | ---: | --: | --------------: | -------------------: |
| granite 3B Q4_K - Medium       |   1.44 GiB |     2.53 B | MTL,BLAS   |       8 |    0 |   1 |           pp512 |       633.47 ± 44.96 |
| granite 3B Q4_K - Medium       |   1.44 GiB |     2.53 B | MTL,BLAS   |       8 |    0 |   1 |           tg128 |         59.78 ± 0.94 |

build: a320bed33 (8323)
```

2. `--mmap` and `--direct-io` arguments do not have the same number of values

```console
$ build/bin/llama-bench -hf ibm-granite/granite-3.3-2b-instruct-GGUF:Q4_K_M --mmap 1,0 --direct-io 1

...truncated...
common_download_file_single_online: using cached file (same etag): /Users/taronaeo/Library/Caches/llama.cpp/ibm-granite_granite-3.3-2b-instruct-GGUF_granite-3.3-2b-instruct-Q4_K_M.gguf
error: --mmap and --direct-io must have the same number of values
```

3. `--mmap` and `--direct-io` are turned on together

```console
$ build/bin/llama-bench -hf ibm-granite/granite-3.3-2b-instruct-GGUF:Q4_K_M --mmap 1,0 --direct-io 1,0

[...]

Issue #9289, opened 2024-09-03
changelog : `libllama` API


ggerganov:
# Overview

This is a list of breaking changes to the public interface of the `llama` library. Collaborators are encouraged to edit this post in order to reflect important changes to the API that end up merged into the `master` branch.

If you are building a 3rd party project that relies on `libllama`, it is recommended to follow this issue and check it before upgrading to new versions.

## Recent API changes (most recent at the top)

| version | PR  | Desc |
| ---     | --- | ---  |
| b3651 | #8980 | Add `LLAMA_VOCAB_TYPE_RWKV` enum value |
| b3644 | #8672 | Add `llama_threadpool` API + change `uint32_t` -> `int32_t` |
| b3614 | #8526 | Add `llama_model_is_recurrent` |

*For older changes, just run `git log -p b3614 -- include/llama.h` and inspect the diffs*

## Upcoming API changes

- Disambiguate API calls https://github.com/ggerganov/llama.cpp/pull/9270
- Refactor the sampling API https://github.com/ggerganov/llama.cpp/pull/8643

ggerganov:
#9355 restores the functionality for getting performance measurements from within `libllama` (which was removed in #9294) via a new `llama_perf` API. The `llama_context_params` is extended with a new `bool no_perf` parameter that can be used to disable the internal timings during `libllama` compute.

ddh0:
Looks like `llama_model_get_tensor` was removed from the API but that change was not documented here

ggerganov:
> Looks like `llama_model_get_tensor` was removed from the API but that change was not documented here

I didn't expect that this function is being used by anyone, so I skipped updating the changelog. It's updated now.

Btw, what do you use this call for?

ddh0:
> I didn't expect that this function is being used by anyone, so I skipped updating the changelog. It's updated now.
> 
> Btw, what do you use this call for?

I don't use it personally but the function was included in [my Python code](https://github.com/ddh0/easy-llama/blob/main/easy_llama/libllama.py), I started to get `ctypes` "symbol not found" errors and I had to do some digging to figure out why. No worries!

ggerganov:
Significant updates to the public API - added `struct llama_vocab` and applied multiple naming changes for better consistency: #11110. Update of user code should be relatively easy - mostly changing functions to use the new names and calling `llama_model_get_vocab(model)` where the old API required `llama_model` and the new API requires `llama_vocab`.

ngxson:
[...]

PR #20834, opened 2026-03-21
args: refactor mlock/mmap/directio into load-mode

taronaeo:
Ref: https://github.com/ggml-org/llama.cpp/pull/20211#discussion_r2916871694
Obsoletes: #20461

This PR overhauls the three separate loading modes (mlock, mmap, and direct-io) into one single `-lm`/`--load-mode` option to simplify the logic. While working on #20461, I realised that it became quite complex to maintain multiple loading modes when they are mutually exclusive of one another. This PR solves that by allowing only one loading mode to exist at a time.

Flags `--mlock`, `--mmap`, `--direct-io` and their negative flags have been marked as deprecated, with help messages informing the user to use the new `--load-mode`.

### Verification

To verify that this refactor did not break any existing codepaths, I have added the following debug statements to verify that the corresponding system calls are registered correctly.

<details>

<summary>Click to expand patch file for codepath verification</summary>

```patch
diff --git a/src/llama-mmap.cpp b/src/llama-mmap.cpp
index ccc29c130..e35065faa 100644
--- a/src/llama-mmap.cpp
+++ b/src/llama-mmap.cpp
@@ -174,13 +174,19 @@ struct llama_file::impl {
 #else
     impl(const char * fname, const char * mode, [[maybe_unused]] const bool use_direct_io = false) : fname(fname) {
 #ifdef __linux__
+        direct_io_requested = use_direct_io;
+        printf("%s: opening '%s' mode='%s' use_direct_io=%d\n", __func__, fname, mode, use_direct_io);
         // Try unbuffered I/O for read only
         if (use_direct_io && std::strcmp(mode, "rb") == 0) {
             if (init_fd()) {
+                printf("%s: Direct I/O enabled for '%s' (fd=%d, alignment=%zu, size=%zu)\n",
+                        __func__, fname, fd, alignment, size);
                 return;
             }
             LLAMA_LOG_WARN("Failed to open file '%s' with error: %s. Falling back to buffered I/O",
                            fname, strerror(errno));
+        } else if (use_direct_io) {
+            printf("%s: Direct I/O requested for '%s' but mode='%s' is not compatible\n", __func__, fname, mode);
         }
 #endif
         init_fp(mode);
@@ -189,6 +195,9 @@ struct llama_file::impl {
 #ifdef __linux__
     bool init_fd() {
         fd = open(fname.c_str(), O_RDONLY | O_DIRECT);
+        if (fd == -1) {
+            printf("%s: open(O_DIRECT) failed for '%s': %s\n", __func__, fname.c_str(), strerror(errno));
+        }
 
         if (fd != -1) {
             struct stat file_stats{};
@@ -281,7 +290,7 @@ struct llama_file::impl {
                     }
                     // Fallback to std::fread in case the DMA controller cannot access the buffer
                     if (errno == EFAULT || errno == EINVAL) {
-                        LLAMA_LOG_WARN("%s: Falling back to buffered IO due to %s\n", __func__, strerror(errno));
+                        LLAMA_LOG_WARN("%s: Falling back to buffered IO for '%s' due to %s\n", __func__, fname.c_str(), strerror(errno));
                         auto curr_off = tell();
                         close(fd);
                         fd = -1;
@@ -334,6 +343,13 @@ struct llama_file::impl {
 
     void read_raw(void * ptr, size_t len) {
         if (has_direct_io()) {
+            if (!direct_io_usage_logged) {
+                printf("%s: Direct I/O read path active for '%s' (fd=%d, alignment=%zu)\n",
+                        __func__, fname.c_str(), fd, alignment);
+                direct_io_usage_logged = true;
+            }
+            direct_io_read_calls++;
+            direct_io_read_bytes += len;
             read_aligned_chunk(ptr, len);
         } else {
             read_raw_unsafe(ptr, len);
@@ -366,6 +382,13 @@ struct llama_file::impl {
     }
 
     ~impl() {
+        if (direct_io_read_calls > 0) {
+            printf("%s: Direct I/O usage summary for '%s': reads=%zu, requested_bytes=%zu\n",
+                    __func__, fname.c_str(), direct_io_read_calls, direct_io_read_bytes);
+        } else if (direct_io_requested) {
+            printf("%s: Direct I/O was requested for '%s' but never used\n", __func__, fname.c_str());
+        }
+
         if (fd != -1) {
             close(fd);
         } else if (owns_fp) {
@@ -374,6 +397,10 @@ struct llama_file::impl {
     }
     int fd = -1;
     std::string fname;
+    bool direct_io_requested = false;
+    bool direct_io_usage_logged = false;
+    size_t direct_io_read_calls = 0;
+    size_t direct_io_read_bytes = 0;
 #endif
 
     size_t read_alignment() const {
@@ -435,6 +462,7 @@ struct llama_mmap::impl {
     std::vector<std::pair<size_t, size_t>> mapped_fragments;
 
     impl(struct llama_file * file, size_t prefetch, bool numa) {
+        printf("%s: file_size=%zu, prefetch=%zu, numa=%d\n", __func__, file->size(), prefetch, numa);
         size = file->size();
         int fd = file->file_id();
         int flags = MAP_SHARED;
@@ -450,6 +478,7 @@ struct llama_mmap::impl {
         if (addr == MAP_FAILED) {
             throw std::runtime_error(format("mmap failed: %s", strerror(errno)));
         }
+        printf("%s: mmap activated: addr=%p, size=%zu\n", __func__, addr, file->size());
 
         if (prefetch > 0) {
             if (posix_madvise(addr, std::min(file->size(), prefetch), POSIX_MADV_WILLNEED)) {
@@ -633,6 +662,7 @@ struct llama_mlock::impl {
 
     bool raw_lock(const void * addr, size_t size) const {
         if (!mlock(addr, size)) {
+            printf("DEBUG: mlock succeeded for addr=%p, size=%zu\n", addr, size);
             return true;
         }
 
@@ -681,6 +711,7 @@ struct llama_mlock::impl {
     bool raw_lock(void * ptr, size_t len) const {
         for (int tries = 1; ; tries++) {
             if (VirtualLock(ptr, len)) {
+                printf("DEBUG: VirtualLock succeeded for ptr=%p, len=%zu\n", ptr, len);
                 return true;
             }
             if (tries == 2) {
```

</details>

1. `--mmap` or `--load-mode mmap`

```console
$ build/bin/llama-completion -hf ibm-granite/granite-4.1-3b-GGUF:Q4_K_M -p "Write me a dog walking business idea 1. " -n 15 --seed 42 --temp 0 -no-cnv -lv 3 --mmap
```

`upstream/master`
```console
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: file_size=2099501664, prefetch=18446744073709551615, numa=0
impl: mmap activated: addr=0x79085ac00000, size=2099501664

Write me a dog walking business idea 1. 2. 3. 4. 5. 6. 
```

`pr/20834`
```console
0.00.110.615 W DEPRECATED: --mmap and --no-mmap are deprecated. use --load-mode mmap instead
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: file_size=2099501664, prefetch=18446744073709551615, numa=0
impl: mmap activated: addr=0x76cddac00000, size=2099501664

Write me a dog walking business idea 1. 2. 3. 4. 5. 6. 
```

2. `--mlock` or `--load-mode mlock`

```console
$ build/bin/llama-completion -hf ibm-granite/granite-4.1-3b-GGUF:Q4_K_M -p "Write me a dog walking business idea 1. " -n 15 --seed 42 --temp 0 -no-cnv -lv 3 --mlock
```

`upstream/master`
```console
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: file_size=2099501664, prefetch=18446744073709551615, numa=0
impl: mmap activated: addr=0x776636c00000, size=2099501664
DEBUG: mlock succeeded for addr=0x776636c00000, size=214319104

Write me a dog walking business idea 1. 2. 3. 4. 5. 6. 
```

`pr/20834`
```
0.00.118.119 W DEPRECATED: --mlock is deprecated. use --load-mode mlock instead
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: file_size=2099501664, prefetch=18446744073709551615, numa=0
impl: mmap activated: addr=0x7a86f2c00000, size=2099501664
DEBUG: mlock succeeded for addr=0x7a86f2c00000, size=214319104

Write me a dog walking business idea 1. 2. 3. 4. 5. 6. 
```

4. `--direct-io` or `--load-mode dio`

```console
$ build/bin/llama-completion -hf ibm-granite/granite-4.1-3b-GGUF:Q4_K_M -p "Write me a dog walking business idea 1. " -n 15 --seed 42 --temp 0 -no-cnv -lv 3 -dio
```

`upstream/master`
```console
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=1
impl: Direct I/O enabled for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' (fd=16, alignment=4096, size=2099501664)
~impl: Direct I/O was requested for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' but never used
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=1
impl: Direct I/O enabled for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' (fd=16, alignment=4096, size=2099501664)
~impl: Direct I/O was requested for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' but never used
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=1
impl: Direct I/O enabled for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' (fd=16, alignment=4096, size=2099501664)
0.01.334.522 W llama_model_loader: direct I/O is enabled, disabling mmap
read_raw: Direct I/O read path active for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' (fd=16, alignment=4096)
~impl: Direct I/O usage summary for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf': reads=1, requested_bytes=210739200

Write me a dog walking business idea 1. 2. 3. 4. 5. 6. 
```

`pr/20834`
```console
0.00.016.167 W DEPRECATED: --direct-io and --no-direct-io are deprecated. use --load-mode dio instead
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=0
impl: opening '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' mode='rb' use_direct_io=1
impl: Direct I/O enabled for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' (fd=16, alignment=4096, size=2099501664)
read_raw: Direct I/O read path active for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf' (fd=16, alignment=4096)
~impl: Direct I/O usage summary for '/home/taronaeo/.cache/huggingface/hub/models--ibm-granite--granite-4.1-3b-GGUF/snapshots/ab4701481089b58a082ef63cc1cee738887293ff/granite-4.1-3b-Q4_K_M.gguf': reads=1, requested_bytes=210739200

Write me a dog walking business idea 1. 2. 3. 4. 5. 6. 
```

Responsible AI Disclosure: AI was used to write debugger code for `llama-mmap.cpp` to ensure feature parity between `upstream/master` and this refactor. AI was also used to identify affected lines within the refactor, but changes were made by a human.


0cc4m:
@ggml-org/maintainers What do you think about this change?

tools/cli/README.md
```diff
@@ -56,9 +55,10 @@
...
 | `-dt, --defrag-thold N` | KV cache defragmentation threshold (DEPRECATED)<br/>(env: LLAMA_ARG_DEFRAG_THOLD) |
 | `-np, --parallel N` | number of parallel sequences to decode (default: 1)<br/>(env: LLAMA_ARG_N_PARALLEL) |
-| `--mlock` | force system to keep model in RAM rather than swapping or compressing<br/>(env: LLAMA_ARG_MLOCK) |
-| `--mmap, --no-mmap` | whether to memory-map model. (if mmap disabled, slower load but may reduce pageouts if not using mlock) (default: enabled)<br/>(env: LLAMA_ARG_MMAP) |
-| `-dio, --direct-io, -ndio, --no-direct-io` | use DirectIO if available. (default: disabled)<br/>(env: LLAMA_ARG_DIO) |
+| `--mlock` | DEPRECATED: force system to keep model in RAM rather than swapping or compressing<br/>(env: LLAMA_ARG_MLOCK) |
+| `--mmap, --no-mmap` | DEPRECATED: whether to memory-map model. (if mmap disabled, slower load but may reduce pageouts if not using mlock)<br/>(env: LLAMA_ARG_MMAP) |
+| `-dio, --direct-io, -ndio, --no-direct-io` | DEPRECATED: use DirectIO if available<br/>(env: LLAMA_ARG_DIO) |
```
am17an:
```suggestion
| `--mlock` | DEPRECATED in favor of `--load-mode`: force system to keep model in RAM rather than swapping or compressing<br/>(env: LLAMA_ARG_MLOCK) |
| `--mmap, --no-mmap` | DEPRECATED in favor of `--load-mode`: whether to memory-map model. (if mmap disabled, slower load but may reduce pageouts if not using mlock)<br/>(env: LLAMA_ARG_MMAP) |
| `-dio, --direct-io, -ndio, --no-direct-io` | DEPRECATED in favor of `--load-mode`: use DirectIO if available<br/>(env: LLAMA_ARG_DIO) |
```

tools/cli/README.md
```diff
@@ -145,6 +145,7 @@
 
 | Argument | Explanation |
 | -------- | ----------- |
+| `--verbose-prompt` | print a verbose prompt before generation (default: false) |
```
am17an, line 148: unrelated?

tools/cli/README.md
```diff
@@ -176,8 +177,8 @@
...
+| `--chat-template-file JINJA_TEMPLATE_FILE` | set custom jinja chat template file (default: template taken from model's metadata)<br/>if suffix/prefix are specified, template will be disabled<br/>only commonly used templates are accepted (unless --jinja is set before this flag):<br/>list of built-in templates:<br/>bailing, bailing-think, bailing2, chatglm3, chatglm4, chatml, command-r, deepseek, deepseek-ocr, deepseek2, deepseek3, exaone-moe, exaone3, exaone4, falcon3, gemma, gigachat, glmedge, gpt-oss, granite, granite-4.0, grok-2, hunyuan-dense, hunyuan-moe, kimi-k2, llama2, llama2-sys, llama2-sys-bos, llama2-sys-strip, llama3, llama4, megrez, minicpm, mistral-v1, mistral-v3, mistral-v3-tekken, mistral-v7, mistral-v7-tekken, monarch, openchat, orion, pangu-embedded, phi3, phi4, rwkv-world, seed_oss, smolvlm, solar-open, vicuna, vicuna-orca, yandex, zephyr<br/>(env: LLAMA_ARG_CHAT_TEMPLATE_FILE) |
```
am17an: unrelated?

am17an (commented):
Need to resolve the conflicts. I also think for a period of time we should have both `--load-mode` and all the other existing options co-existing, but print a depreciation warning. `--load-mode` should take precedence if specified.

taronaeo:
Just rebased with `upstream/master` to resolve conflicts.

@am17an @ORippler Could you review the PR again? Or if everything is good to go, an approval would be much appreciated :)

tools/cli/README.md
```diff
@@ -15,7 +15,6 @@
 | `--license` | show source code license and dependencies |
 | `-cl, --cache-list` | show list of models in cache |
 | `--completion-bash` | print source-able bash completion script for llama.cpp |
-| `--verbose-prompt` | print a verbose prompt before generation (default: false) |
```
am17an, line 18: This got missed?

tools/cli/README.md
```diff
@@ -176,8 +177,8 @@
...
-| `--chat-template-file JINJA_TEMPLATE_FILE` | set custom jinja chat template file (default: template taken from model's metadata)<br/>if suffix/prefix are specified, template will be disabled<br/>only commonly used templates are accepted (unless --jinja is set before this flag):<br/>list of built-in templates:<br/>bailing, bailing-think, bailing2, chatglm3, chatglm4, chatml, command-r, deepseek, deepseek-ocr, deepseek2, deepseek3, exaone-moe, exaone3, exaone4, falcon3, gemma, gigachat, glmedge, gpt-oss, granite, grok-2, hunyuan-dense, hunyuan-moe, kimi-k2, llama2, llama2-sys, llama2-sys-bos, llama2-sys-strip, llama3, llama4, megrez, minicpm, mistral-v1, mistral-v3, mistral-v3-tekken, mistral-v7, mistral-v7-tekken, monarch, openchat, orion, pangu-embedded, phi3, phi4, rwkv-world, seed_oss, smolvlm, solar-open, vicuna, vicuna-orca, yandex, zephyr<br/>(env: LLAMA_ARG_CHAT_TEMPLATE_FILE) |
```
am17an, line 180: not sure why there are changes here

src/llama-model-loader.cpp
```diff
@@ -534,6 +534,9 @@ llama_model_loader::llama_model_loader(
 
     tensor_buft_overrides = param_tensor_buft_overrides_p;
 
+    this->use_mmap      = (load_mode == LLAMA_LOAD_MODE_MMAP || load_mode == LLAMA_LOAD_MODE_MLOCK);
+    this->use_direct_io = (load_mode == LLAMA_LOAD_MODE_DIRECT_IO);
```
am17an:
```suggestion
    this->use_mmap      = load_mode == LLAMA_LOAD_MODE_MMAP || load_mode == LLAMA_LOAD_MODE_MLOCK;
    this->use_direct_io = load_mode == LLAMA_LOAD_MODE_DIRECT_IO;
```

tools/llama-bench/README.md
```diff
@@ -24,46 +24,55 @@ usage: llama-bench [options]
...
-  --delay <0...N> (seconds)                 delay between each test (default: 0)
-  -o, --output <csv|json|jsonl|md|sql>      output format printed to stdout (default: md)
-  -oe, --output-err <csv|json|jsonl|md|sql> output format printed to stderr (default: none)
-  --list-devices                            list available devices and exit
-  -v, --verbose                             verbose output
-  --progress                                print test progress indicators
-  -rpc, --rpc <rpc_servers>                 register RPC devices (comma separated)
+  --numa <distribute|isolate|numactl>         numa mode (default: disabled)
+  -r, --repetitions <n>                       number of times to repeat each test (default: 5)
+  --prio <-1|0|1|2|3>                         process/thread priority (default: 0)
```
am17an: not sure why this is there either

am17an (commented):
Overall looks good, I tried it out locally as well. There are some diffs which are unrelated and should be fixed. Also you need to change this log line and the probably the surrounding code

https://github.com/ggml-org/llama.cpp/blob/e21cdc11a0461d8b0cbd28cc356d993bf6be7282/src/llama-model.cpp#L2969-L2970

taronaeo:
> There are some diffs which are unrelated and should be fixed.

I think my reply got buried in the PR history, but to quote my previous comments,

> > RE the unrelated changes in this PR, these are artifacts from running build/bin/llama-gen-docs to update all the *.md documentation files. I can manually remove these changes from my PR to avoid conflicts if needed.

Nevertheless, I've manually went through the documentation changes made by `build/bin/llama-gen-docs` to exclude anything unrelated to this PR.

> Also you need to change this log line and the probably the surrounding code
> https://github.com/ggml-org/llama.cpp/blob/e21cdc11a0461d8b0cbd28cc356d993bf6be7282/src/llama-model.cpp#L2969-L2970

I've updated it to log this way. Let me know if this is the desired log message.

```log
load_tensors: loading model tensors, this can take a while... (load_mode = mmap)

### OR

load_tensors: loading model tensors, this can take a while... (load_mode = mlock)

### OR

load_tensors: loading model tensors, this can take a while... (load_mode = dio)
```

am17an:
What I meant was `llama_models_params` hasn't gone through this change. I think we intend to keep that as is? In case this PR looks good

https://github.com/ggml-org/llama.cpp/blob/006809fdba84260478fc351d11f7f293d89bd246/src/llama-model.cpp#L9303-L9305

include/llama.h
```diff
@@ -312,9 +322,9 @@ extern "C" {
 
         // Keep the booleans together to avoid misalignment during copy-by-value.
         bool vocab_only;      // only load the vocabulary, no weights
-        bool use_mmap;        // use mmap if possible
-        bool use_direct_io;   // use direct io, takes precedence over use_mmap when supported
-        bool use_mlock;       // force system to keep model in RAM
+        // bool use_mmap;        // DEPRECATED: use mmap if possible
+        // bool use_direct_io;   // DEPRECATED: use direct io, takes precedence over use_mmap when supported
+        // bool use_mlock;       // DEPRECATED: force system to keep model in RAM
```
JohannesGaessler: For the llama C API this is a breaking change rather than a deprecation so wouldn't it make more sense to just remove these?

examples/training/finetune.cpp
```diff
@@ -26,10 +26,10 @@ int main(int argc, char ** argv) {
         return 1;
     }
 
-    if (params.use_mmap) {
+    if (params.load_mode == LLAMA_LOAD_MODE_MMAP) {
```
JohannesGaessler:
```suggestion
    if (params.load_mode != LLAMA_LOAD_MODE_NONE) {
```

To my understanding `LLAMA_LOAD_MODE_MLOCK` would result in the same problem, I would say to just force `LLAMA_LOAD_MODE_NONE` here.

common/arg.cpp
```diff
@@ -2212,27 +2214,45 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
...
+            LOG_WRN("DEPRECATED: -dio/--direct-io and -ndio/--no-direct-io are deprecated. use --load-mode dio instead\n");
+            params.load_mode = value ? LLAMA_LOAD_MODE_DIRECT_IO : LLAMA_LOAD_MODE_NONE;
         }
     ).set_env("LLAMA_ARG_DIO"));
+    add_opt(common_arg(
+        {"-lm", "--load-mode"}, "MODE",
+        "model loading mode (default: mmap)\n"
+        "- none: no special loading mode\n"
+        "- mmap: memory-map model (if mmap disabled, slower load but may reduce pageouts if not using mlock)\n"
+        "- mlock: force system to keep model in RAM rather than swapping or compressing\n"
```
JohannesGaessler:
```suggestion
        "- mlock: mmap + force system to keep model in RAM rather than swapping or compressing\n"
```

src/llama-model.cpp
```diff
@@ -9294,15 +9294,13 @@ llama_model_params llama_model_default_params() {
...
+        /*.load_mode                   =*/ LLAMA_LOAD_MODE_MMAP,
         /*.main_gpu                    =*/ 0,
         /*.tensor_split                =*/ nullptr,
         /*.progress_callback           =*/ nullptr,
         /*.progress_callback_user_data =*/ nullptr,
         /*.kv_overrides                =*/ nullptr,
         /*.vocab_only                  =*/ false,
-        /*.use_mmap                    =*/ true,
-        /*.use_direct_io               =*/ false,
-        /*.use_mlock                   =*/ false,
```
taronaeo, line 9305: > What I meant was llama_models_params hasn't gone through this change. I think we intend to keep that as is?

It's changed here. I don't think it's beneficial to keep the respective parameters available as the code has been refactored to use `load_mode `instead.

tools/llama-bench/llama-bench.cpp
```diff
@@ -1265,34 +1309,33 @@ static std::vector<cmd_params_instance> get_cmd_params_instances(const cmd_param
...
-                /* .embeddings   = */ embd,
-                /* .no_op_offload= */ nopo,
-                /* .no_host      = */ noh,
-                /* .fit_target   = */ fpt,
-                /* .fit_min_ctx  = */ fpc,
+                /* .embeddings    = */ embd,
+                /* .no_op_offload = */ nopo,
+                /* .no_host       = */ noh,
+                /* .fit_target    = */ fpt,
+                /* .fit_min_ctx   = */ fpc,
```
JohannesGaessler: If you're changing the indentation anyways, please make it match `tensor_buft_overrides`.

JohannesGaessler (commented):
Due to the changes in `llama-bench.cpp` it is also necessary to change `scripts/compare-llama-bench.py`.

include/llama.h
taronaeo: I initially wanted to let developers know that we've refactored all three flags into one without having to check #9289. But I guess it's cleaner to have it removed now and documented in the changelog instead.

Removed in the latest push.

taronaeo:
Thanks! I forgot about that. The output for `scripts/compare-llama-bench.py` looks like this now.

```markdown
| Model           | Load mode         | Test   |   t/s master |   t/s master |   Speedup |
|:----------------|:------------------|:-------|-------------:|-------------:|----------:|
| qwen2 1.5B BF16 | mmap              | pp512  |      1232.05 |      1232.05 |      1.00 |
| qwen2 1.5B BF16 | mmap              | tg128  |        49.37 |        49.37 |      1.00 |
| qwen2 1.5B BF16 | none              | pp512  |      1233.24 |      1233.24 |      1.00 |
| qwen2 1.5B BF16 | none              | tg128  |        49.09 |        49.09 |      1.00 |
```

JohannesGaessler (dismissed):

JohannesGaessler:
I remembered this PR after looking at https://github.com/ggml-org/llama.cpp/pull/21821 . That PR is about enabling huge pages for Linux which seems to be ~1% more efficient. Would it make sense to use bit flags rather than an enum for the load mode?

taronaeo:
> Would it make sense to use bit flags rather than an enum for the load mode?

I had considered this while writing this refactor. IMO it would not be the best option here since mmap/mlock are mutually exclusive to direct-io. If we were to implement this as bit flags, I think we would run into the same problem as what we are trying to fix in this refactor, albeit now in a bit mask instead of different feature flags.

I would still think that using an enum is more appropriate in this scenario. Let me know what you think.

taronaeo:
I have just rebased this PR with `upstream/master` following the breaking changes to `llama_fit_params` and verified that it still works correctly. PR description is also updated with the latest test logs.

common/arg.cpp
```diff
@@ -2213,27 +2215,45 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
     }
     add_opt(common_arg(
         {"--mlock"},
-        "force system to keep model in RAM rather than swapping or compressing",
+        "DEPRECATED in favor of `--load-mode`: force system to keep model in RAM rather than swapping or compressing",
```
0cc4m:
```suggestion
        "DEPRECATED in favor of `--load-mode`: mmap + force system to keep model in RAM rather than swapping or compressing",
```

examples/training/finetune.cpp
```diff
@@ -26,10 +26,10 @@ int main(int argc, char ** argv) {
         return 1;
     }
 
-    if (params.use_mmap) {
+    if (params.load_mode != LLAMA_LOAD_MODE_NONE) {
```
0cc4m, line 29:
```suggestion
    if (params.load_mode != LLAMA_LOAD_MODE_NONE && params.load_mode != LLAMA_LOAD_MODE_DIRECT_IO) {
```
Assuming that direct-io is fine for this example. Otherwise the LOG message needs to change.

src/llama-model.cpp
```diff
@@ -2970,16 +2970,16 @@ void llama_model::load_vocab(llama_model_loader & ml) {
 
 bool llama_model::load_tensors(llama_model_loader & ml) {
     const auto & split_mode   = params.split_mode;
-    const auto & use_mlock    = params.use_mlock;
+    const auto & use_mlock    = params.load_mode == LLAMA_LOAD_MODE_MLOCK;
```
0cc4m:
```suggestion
    const bool   use_mlock    = params.load_mode == LLAMA_LOAD_MODE_MLOCK;
```
A reference to a temporary is weird.

0cc4m (commented):
Needs another rebase, and some more @ggml-org/maintainers opinions/discussion.

Personally I think this is the right thing to do, clean up a small mess in the arguments that has grown over time.

JohannesGaessler:
@0cc4m do you have an opinion on an enum vs. flags? My concern is that with an enum we will eventually end up with a different mess where combinations of features are either arbitrarily not supported or we have an exponentially growing enum.

0cc4m:
Do you think a great number of loading modes will show up? So far the list is mutually exclusive (except mlock, in a way), so an enum is fine. What loading features would not be exclusive?

JohannesGaessler:
There is this PR https://github.com/ggml-org/llama.cpp/pull/21821 that enables huge pages for example.

0cc4m:
I think these are two separate cases. We can cover the mutually exclusive loading modes with the enum and "modifiers" like huge pages with flags. Putting everything into one bitmask makes it harder to understand and includes a lot of invalid combinations, like mmap+direct-io.

examples/training/finetune.cpp
taronaeo: I'm not too clear about DirectIO. I've kept the existing if logic and updated the log message to this:

```c++
        LOG_INF("%s: forcing load_mode = none to enable writable pointers to the weights\n", __func__);
```

taronaeo:
Sorry for the late response, been busy with other work.

> Needs another rebase

I was holding back until we have an agreement on whether to use enum vs. flags, since the files touched in this PR are constantly being worked on, and would more often than not, cause conflicts that needs manual resolving.

> My concern is that with an enum we will eventually end up with a different mess where combinations of features are either arbitrarily not supported or we have an exponentially growing enum.

The only combinations I've seen thus far are (1) mmap + mlock, and (2) the linked mmap + hugepages PR. By using enums as such:

```c++
LLAMA_LOAD_MODE_MLOCK
LLAMA_LOAD_MODE_HUGEPAGES
```

we are already implying the use of mmap behind the scenes, and that the configuration will be guaranteed to be correct because the behavior is hard-coded.

But if we were to use flags, the developer can specify any arbitrary number of configurations f.ex., mmap + DirectIO, which doesn't make sense, and we would have to check for these errant configurations again in our code and defeat the purpose of this PR in the first place.

taronaeo:
IMO we should consider using enums first and move forward with this PR if possible. If there is a case that does not make sense to use an enum, we can always revisit it again and reconsider with the new knowledge of the case.

JohannesGaessler:
I read up a bit on I/O functionality and I think @0cc4m 's suggestion is sensible. From my end I agree that it would make sense to cover the mutually exclusive modes with an enum as is done in this PR and to have modifiers like huge pages (which I think can apply to all of these load modes) as flags. In terms of the long-term interface design, should we make such modifiers part of the `--load-mode` parsing, so e.g. `--load-mode mmap+huge_pages` with a `+` indicating a modifier?

taronaeo:
> We can cover the mutually exclusive loading modes with the enum and "modifiers" like huge pages with flags.

> From my end I agree that it would make sense to cover the mutually exclusive modes with an enum as is done in this PR and to have modifiers like huge pages (which I think can apply to all of these load modes) as flags

I also agree with this approach. I'm assuming all three of us are in agreement on this approach now and will proceed to overhaul this PR with the suggested implementation.

> should we make such modifiers part of the --load-mode parsing, so e.g. --load-mode mmap+huge_pages with a + indicating a modifier?

I'm OK with this design. Although I feel that specifying just `--load-mode huge_pages` can also automatically imply `mmap+huge_pages` internally as well, or should we prefer explicit?

taronaeo:
I've completed the refactor to split `none/mmap/direct-io` into its own exclusive modes and `mlock` into a bitflag. This allows future expansion with HugePages. For now, I've made it work only with the `mode+modifier` syntax, which means to enable a modifier like `mlock`, the argument would have to be `-lm mmap+mlock`.

This PR is ready for another review. cc: @0cc4m @JohannesGaessler

src/llama.cpp
```diff
@@ -46,6 +46,52 @@ const char * llama_flash_attn_type_name(enum llama_flash_attn_type flash_attn_ty
...
+    }
+    GGML_ABORT("fatal error");
+}
+
+const char * llama_load_modifier_name(uint32_t load_modifier) {
+    if (load_modifier == LLAMA_LOAD_MODIFIER_NONE) {
+        return "none";
+    }
+
+    static std::string result;
```
0cc4m: This is weird and may also be thread-unsafe. What is the reasoning for the static string + const char * return type?

common/arg.cpp
```diff
@@ -2255,27 +2256,59 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
...
         {"--no-mmap"},
-        string_format("whether to memory-map model. (if mmap disabled, slower load but may reduce pageouts if not using mlock) (default: %s)", params.use_mmap ? "enabled" : "disabled"),
+        "DEPRECATED in favor of `--load-mode`: whether to memory-map model. (if mmap disabled, slower load but may reduce pageouts if not using mlock)",
         [](common_params & params, bool value) {
-            params.use_mmap = value;
+            LOG_WRN("DEPRECATED: --mmap and --no-mmap are deprecated. use --load-mode mmap instead\n");
+            if (value) {
+                params.load_mode = LLAMA_LOAD_MODE_MMAP;
+                params.load_modifier = LLAMA_LOAD_MODIFIER_NONE;
+            }
```
0cc4m: Will this just skip and do nothing if `--no-mmap` is used?

common/arg.cpp
```diff
@@ -2255,27 +2256,59 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
...
         {"-ndio", "--no-direct-io"},
-        string_format("use DirectIO if available. (default: %s)", params.use_direct_io ? "enabled" : "disabled"),
+        "DEPRECATED in favor of `--load-mode`: use DirectIO if available",
         [](common_params & params, bool value) {
-            params.use_direct_io = value;
+            LOG_WRN("DEPRECATED: --direct-io and --no-direct-io are deprecated. use --load-mode dio instead\n");
+            if (value) {
+                params.load_mode = LLAMA_LOAD_MODE_DIRECT_IO;
+                params.load_modifier = LLAMA_LOAD_MODIFIER_NONE;
+            }
```
0cc4m: Same, this means `--no-direct-io` does nothing.

src/llama-model-loader.cpp
```diff
@@ -535,6 +536,9 @@ llama_model_loader::llama_model_loader(
 
     tensor_buft_overrides = param_tensor_buft_overrides_p;
 
+    this->use_mmap      = load_mode == LLAMA_LOAD_MODE_MMAP || load_modifier == LLAMA_LOAD_MODIFIER_MLOCK;
```
0cc4m: I don't think the load modifier should modify the load mode, this would e.g. enable mmap when using `--load-mode none+mlock`

tools/llama-bench/llama-bench.cpp
```diff
@@ -762,6 +764,37 @@ static cmd_params parse_cmd_params(int argc, char ** argv) {
                     break;
                 }
                 params.split_mode.insert(params.split_mode.end(), modes.begin(), modes.end());
+            } else if (arg == "-lm" || arg == "--load_mode") {
```
0cc4m:
```suggestion
            } else if (arg == "-lm" || arg == "--load-mode") {
```

tools/llama-bench/llama-bench.cpp
```diff
@@ -822,15 +855,39 @@ static cmd_params parse_cmd_params(int argc, char ** argv) {
...
+
+                std::vector<llama_load_mode> modes;
+                for (const auto & m : p) {
+                    llama_load_mode mode;
+                    if (m) {
+                        mode = LLAMA_LOAD_MODE_MMAP;
+                    } else {
+                        mode = LLAMA_LOAD_MODE_NONE;
+                    }
+                    modes.push_back(mode);
```
0cc4m, line 871: Wouldn't this also need an entry in load_modifier?

0cc4m (commented):
To be honest, I don't see any need for the modifiers here. It seems like we're just adding complexity for potential future features that apply to more than one mode.

mlock would be easier to implement as its own mode that does mmap+mlock. This is also visible in the cases where you set modifier to NONE if the load mode doesn't support it to avoid nonsensical combinations.

src/llama.cpp
taronaeo: `const char *` because `llama.h` can be consumed by C libraries so I can't directly return `std::string`. `static` because we are returning a pointer to `result` so it must still be valid when consumed.

Maybe I should have used `thread_local` instead.

common/arg.cpp
taronaeo: That is the current behavior, yes. But I realize it's a bug because `--no-mmap` should do the inverse instead of doing nothing. Patched in the latest commits.

src/llama-model-loader.cpp
taronaeo: Our current `--mlock` flag hint specifies that `mlock` automatically implies that `mmap` gets activated. In such cases, if the user specifies `-lm none+mlock`, `mlock` would not actually activate as per these lines in the argument handler:

```c++
            // prevents -lm none+mlock or similar combinations that don't make sense
            if (params.load_mode == LLAMA_LOAD_MODE_NONE) { params.load_modifier = LLAMA_LOAD_MODIFIER_NONE; }
```

But if the user specifies `-lm mmap+mlock`, it will work as intended.

taronaeo:
I've just addressed most of the review comments in the latest commits.

> To be honest, I don't see any need for the modifiers here. It seems like we're just adding complexity for potential future features that apply to more than one mode.

Same thought here. That was why I wanted us to use a single enum for all the actual modes/combinations that we support in the first place, instead of allowing arbitrary combinations and adding more code to support it, which is IMO unnecessary maintenance overhead.

This still makes more sense to me:

```c++
    enum llama_load_mode {
        LLAMA_LOAD_MODE_NONE      = 0, // no special loading mode
        LLAMA_LOAD_MODE_MMAP      = 1, // memory map the model
        LLAMA_LOAD_MODE_MLOCK     = 2, // mmap + lock it in RAM (i.e. prevent swapping or compressing)
        LLAMA_LOAD_MODE_DIRECT_IO = 3, // use direct I/O if available
        // LLAMA_LOAD_MODE_HUGEPAGES = 4, // mmap + use hugepages if available
    };
```

I would prefer that we all agree on either
(1) only using a single enum for all the combinations we support, or;
(2) continue using the `load_mode` enum + `load_modifiers` bitflag combination with additional maintenance overhead

before we make more changes to this PR. Otherwise we will constantly bounce between the two options with no end in sight.

0cc4m:
I think it's best to just use a mode enum until we actually get modifiers that are available on many or all modes.

taronaeo:
@JohannesGaessler Gentle ping. Waiting on your agreement/opinion to move forward with the single `enum llama_load_mode` instead of having modifiers, before I update this PR again :)

JohannesGaessler:
Sorry for the long silence, I took a break from llama.cpp because I felt I was too stressed. If it comes down to it I'm fine with an enum-only solution as well.

src/llama-model-loader.cpp
```diff
@@ -535,6 +536,9 @@ llama_model_loader::llama_model_loader(
...
+    this->use_direct_io = load_mode == LLAMA_LOAD_MODE_DIRECT_IO;
```
ORippler, line 546: Thought's on making this load-mode configurable per backend in the future? IIRC, CPUs/~UMA~ systems may prefer `-mmap` as it is an easy way to enable weight-streaming. dGPUs/other systems that are bound by PCIe BW may prefer `-directio` to avoid potential cache trashing

ORippler (commented):
+1 on the single enum approach. @JTischbein please take a look if there is anything additional regarding NVGPUs we should consider

src/llama-model-loader.cpp
0cc4m: UMA systems must not use mmap because otherwise they run out of memory quickly when loading the model from mapped memory into gpu-visible memory in the same memory. It basically loads the model twice.

ORippler: yeah we had issues with CUDA RT being swapped as well in the past. Guess it's a common UMA problem and only CPUs (+ maybe Metal?) may prefer `-mmap` then

ORippler: though I faintly recall NVGPUs supporting file-backed mmap on "full" UMA systems https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/unified-memory.html#file-backed-unified-memory, have not looked at this yet though

taronaeo:
My apologies on the delay. As agreed, I have updated the PR to use only a single enum and updated the description to match the latest checks.

PTAL again. @0cc4m @JohannesGaessler @ORippler

0cc4m (approved):

src/llama-model-loader.cpp
JTischbein: @ORippler Unfortunately "file backed UMA" this does not work on RTX/DGX Spark.

ORippler:
How do we ensure that `--load-mode` takes precedence over deprecated flags again? Seems like we parse them sequentially: https://github.com/taronaeo/llama.cpp-s390x/blob/fd9359d8bea0044977d4ef089f7c46cc2986ac9b/common/arg.cpp#L531-L603

I'm personally fine with sequential processing, but this was mentioned here https://github.com/ggml-org/llama.cpp/pull/20834#pullrequestreview-4066916578

taronaeo:
> How do we ensure that `--load-mode` takes precedence over deprecated flags again? Seems like we parse them sequentially: https://github.com/taronaeo/llama.cpp-s390x/blob/fd9359d8bea0044977d4ef089f7c46cc2986ac9b/common/arg.cpp#L531-L603
> 
> I'm personally fine with sequential processing, but this was mentioned here [#20834 (review)](https://github.com/ggml-org/llama.cpp/pull/20834#pullrequestreview-4066916578)

Ah, I missed addressing that review comment. As of right now, it is last-flag-wins like how we handle every other flag. I don't have a clean solution to ensure that `--load-mode` takes presence without adding additional, unwanted bloat into the argument parser to handle this specific case and flag.

I think we can go the route of adding a warning when both `--mmap|mlock|direct-io` and `--load-mode` are detected in the arguments and give a similar warning as such:

```console
0.00.054.386 W DEPRECATED: argument '-c' specified multiple times, use comma-separated values instead (only last value will be used)
```

_^ This warning already exists when you specify multiple `-c` flags in `llama-server`._

Please let me know what you think. I will fix the merge conflicts together with this review in a while.

ORippler:
> I think we can go the route of adding a warning when both --mmap|mlock|direct-io and --load-mode are detected in the arguments and give a similar warning as such:

Sounds good to me, as the intent is to remove those CLI flags in future anyways. @am17an do you agree (you raised this initial concern)?

am17an:
Looks good to me!

am17an (approved):

taronaeo:
I had to rebase because the docs were conflicting with upstream. Added a warning for users who specify both `--mmap|mlock|dio` and `--load-mode` together as well, as discussed:

```console
0.00.001.505 W DEPRECATED: --mmap and --no-mmap are deprecated. use --load-mode mmap instead
0.00.001.509 W DEPRECATED: `--load-mode` and `--mlock`/`--mmap`/`--direct-io` should not be combined; only the last flag on the command line will take effect
```

Let's get this upstream before another merge conflict happens again and having to go down the rabbit-hole :)

@ggml-org/maintainers

pwilkin (approved):

JohannesGaessler (approved):

taronaeo:
Merging on green, ignoring the already-known CI errors with `test-recurrent-state-rollback`. Will update #9289 changelog upon merge also.

ORippler (approved):

vakst:
This patch changes llama.cpp's behavior in a way that breaks my existing workflow.

Previously, I ran GLM-5.2-UD-IQ2_M with --no-mmap and --mlock. The model uses almost all available RAM, but Linux is still able to rebalance memory and successfully load all weights.

After this patch, direct-io is enabled by default. As a result, the kernel can no longer use the page cache or rebalance memory in the same way, and llama.cpp fails to load the model because there isn't enough free memory available at startup.

I'm also concerned about the planned deprecation of --no-direct-io. Removing this option would eliminate the ability to disable Direct I/O, making this regression unavoidable for users with similar memory-constrained setups.

Could direct-io remain opt-in, or at least keep --no-direct-io available as an escape hatch for cases like this?

pwilkin:
@vakst you can just use `--load-mode mlock`

abc-nix:
> @vakst you can just use `--load-mode mlock`

Sorry to bother you. Does `--load-mode none` equal `--no-mmap`? Thanks for your time.

vakst:
> > @vakst you can just use `--load-mode mlock`
> 
> Sorry to bother you. Does `--load-mode none` equal `--no-mmap`? Thanks for your time.

Maybe I'm misunderstanding the new feature, but at first I couldn't figure out how to enable mlock while disabling direct-io using the new --load-mode parameter.

Also, this changes the current default behavior: direct-io is now enabled by default, whereas it was disabled by default before this PR. As a result, existing configurations will likely need to be updated after this change to preserve their previous behavior.

taronaeo:
> Sorry to bother you. Does --load-mode none equal --no-mmap? Thanks for your time.

Yes.

> After this patch, direct-io is enabled by default.

Where are you seeing this? In the old path, DirectIO is not turned on unless explicitly requested to via `-dio` or `--direct-io` and is still not turned on by default.

The current default translated from the old path is `--mmap`.

> Could direct-io remain opt-in, or at least keep --no-direct-io available as an escape hatch for cases like this?

`--no-direct-io` is the same as not choosing `--load-mode dio`.

vakst:
> > Sorry to bother you. Does --load-mode none equal --no-mmap? Thanks for your time.
> 
> Yes.
> 
> > After this patch, direct-io is enabled by default.
> 
> Where are you seeing this? In the old path, DirectIO is not turned on unless explicitly requested to via `-dio` or `--direct-io` and is still not turned on by default.
> 
> The current default translated from the old path is `--mmap`.
> 
> > Could direct-io remain opt-in, or at least keep --no-direct-io available as an escape hatch for cases like this?
> 
> `--no-direct-io` is the same as not choosing `--load-mode dio`.

just finished testing several flag combinations. Here are the results:

--load-mode mmap: loads successfully, but it's about 3× slower.
--load-mode mlock: the process gets killed (out of memory).
--load-mode none: works.
--load-mode mlock --no-direct-io: works.
--mmap --mlock --no-direct-io: works.
--no-mmap --mlock --no-direct-io: works.
--no-mmap --mlock: the process gets killed (out of memory). The last combination used to work before the PR

taronaeo:
> just finished testing several flag combinations. Here are the results:
> 
> --load-mode mmap: loads successfully, but it's about 3× slower.
> --load-mode mlock: the process gets killed (out of memory).
> --load-mode none: works.
> --load-mode mlock --no-direct-io: works.
> --mmap --mlock --no-direct-io: works.
> --no-mmap --mlock --no-direct-io: works.
> --no-mmap --mlock: the process gets killed (out of memory). The last combination used to work before the PR

This is not a definitive test. As mentioned in https://github.com/ggml-org/llama.cpp/pull/20834#issuecomment-5058003640, the last flag will override the pre-existing flag and the warnings are there:

```console
0.00.001.509 W DEPRECATED: `--load-mode` and `--mlock`/`--mmap`/`--direct-io` should not be combined; only the last flag on the command line will take effect
```

As such, you need to plug in debugger code to show that DirectIO is turning on by default, and is the cause of this.

To explain one of your tests better,

```
--load-mode mlock --no-direct-io
```

is the equivalent of `--load-mode none`

vakst:
I apologize—this isn't caused by the direct-io flag. The issue is that the new PR ignores --no-mmap and ends up enabling mmap somehow.

Before the PR (--no-mmap + --mlock):

lsof -p $(pgrep llama-server) | grep GLM-5.2

llama-ser 228233 user  59r      REG                9,0 238577580160      283 /data/gguf/GLM5.2/GLM-5.2-UD-IQ2_M.gguf

After the PR (--no-mmap + --mlock):

llama-ser 217543 user mem       REG                9,0 238577580160      283 /data/gguf/GLM5.2/GLM-5.2-UD-IQ2_M.gguf
llama-ser 217543 user  58r      REG                9,0 238577580160      283 /data/gguf/GLM5.2/GLM-5.2-UD-IQ2_M.gguf

The mem mapping indicates that the model file is still being memory-mapped, even though --no-mmap is specified.

pwilkin:
I think I know what might be going on.

@vakst under the old code, `--mlock` didn't explicitly set `--mmap`. But it also was a **no-op** if `--mmap` wasn't set, so `--mlock --no-mmap` behaved like `--load-mode none` under the new behavior.

Under the new code, `--load-mode mlock` **also** enables mmap, so you're seeing the same behavior as if you did `--mmap --mlock` (or just `--mlock`, since `--mmap` was the default) under the old code.

vakst:
> I think I know what might be going on.
> 
> @vakst under the old code, `--mlock` didn't explicitly set `--mmap`. But it also was a **no-op** if `--mmap` wasn't set, so `--mlock --no-mmap` behaved like `--load-mode none` under the new behavior.
> 
> Under the new code, `--load-mode mlock` **also** enables mmap, so you're seeing the same behavior as if you did `--mmap --mlock` (or just `--mlock`, since `--mmap` was the default) under the old code.

Thank you so much for the clarification! I just wanted to point out that the code isn't backward compatible. It may not be a big issue, and anyone affected will probably just need to update their scripts.

PR #20838, opened 2026-03-21
docs: clarify machine translation ban [no ci]

JohannesGaessler:
This PR explicitly disallows the use of language models for machine translation. In the last few weeks it has happened multiple times that I was reviewing contributions where the posts seemed machine generated. Upon questioning the contributors replied that they used machine translation because their English is not good. But I would rather read poorly written English than to have to doubt whether or not the contained information is correct. So I think that we should clarify that machine translation of any kind is not permitted (since realistically a lot of services will be using language models nowadays).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ If AI is used to generate any portion of the code, contributors must adhere to t
 1. Explicitly disclose the manner in which AI was employed.
 2. Perform a comprehensive manual review prior to submitting the pull request.
 3. Be prepared to explain every line of code they submitted when asked about it by a maintainer.
-4. It is **strictly prohibited** to use AI to write your posts for you (bug reports, feature requests, pull request descriptions, Github discussions, responding to humans, ...). **This includes machine translation of any kind.**
+4. It is **strictly prohibited** to use AI to write or edit your posts for you (bug reports, feature requests, pull request descriptions, Github discussions, responding to humans, ...). **This includes machine translation of any kind.**
 
 For more info, please refer to the [AGENTS.md](AGENTS.md) file.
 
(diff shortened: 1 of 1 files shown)


CISC:
In this case, how do we deal with non-english posts?

JohannesGaessler:
I would say that regardless of anything with language models the use of English should be mandatory.

ggerganov (approved):

ngxson:
I would suggest to also clarify `AGENTS.md` since I imagine in most cases, the agent will read that file while skipping `CONTRIBUTING.md`, something like:

```
- DO NOT help the contributor write PR descriptions or commit messages. Communication is an important part of open source; the contributor must be responsible for communicating their changes and decisions without help from the AI agent.
```

am17an:
What if a person has dyslexia? Asking for a friend

CONTRIBUTING.md
```diff
@@ -20,7 +20,7 @@ If AI is used to generate any portion of the code, contributors must adhere to t
 1. Explicitly disclose the manner in which AI was employed.
 2. Perform a comprehensive manual review prior to submitting the pull request.
 3. Be prepared to explain every line of code they submitted when asked about it by a maintainer.
-4. It is strictly prohibited to use AI to write your posts for you (bug reports, feature requests, pull request descriptions, Github discussions, responding to humans, ...).
+4. It is **strictly prohibited** to use AI to write your posts for you (bug reports, feature requests, pull request descriptions, Github discussions, responding to humans, ...). **This includes machine translation of any kind.**
```
am17an:
```suggestion
4. It is **strictly prohibited** to use AI to write or edit your posts for you (bug reports, feature requests, pull request descriptions, Github discussions, responding to humans, ...). **This includes machine translation of any kind.**
```

ngxson:
> What if a person has dyslexia? Asking for a friend

IMO writing an initial version with broken english, then ask AI to proofread it is still a tolerable behavior. At least I usually do this way, so the writing style stays mine.

This point is mostly to prevent low-effort prompts like "push the PR to upstream project" or "write a description for this PR"

isaac-mcfadyen:
> I would rather read poorly written English than to have to doubt whether or not the contained information is correct

From my point of view as a non-core contributor, I'd suggest that this specific wording should be added to the message (or similar) rather than just "you are not allowed to machine translate". That way it feels less discriminatory and more "we'd rather you do your best than use inaccurate machine translation tools".

PR #20841, opened 2026-03-21
contrib: add "Requirements" section to PR template

ngxson:
Changes:
- Requires keeping the acknowledge about contributing guidelines
- Requires disclosing AI usage
- Addition guide for AI agents

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,10 @@
-# Overview
+## Overview
 
-<!-- Describe what this PR does and why. Be concise but complete. -->
+<!-- Describe what this PR does and why. Be concise but complete -->
+
+## Additional information
+
+<!-- You can provide more details and link related discussions here. Delete this section if not applicable -->
 
 # Requirements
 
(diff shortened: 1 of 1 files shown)


am17an:
Does a commit co-authored with Claude count as disclosure?

ngxson:
> Does a commit co-authored with Claude count as disclosure?

I think co-authored doesn't specify exactly which part of the work is done by claude.

I also believe that a written disclosure encourages the contributor to be responsible / be transparent about their work.

JohannesGaessler:
How about we make it checkboxes? Something like

- [ ] I have read and understood the contributing guidelines.
- [ ] AI was used to write parts of the code.

ngxson:
@JohannesGaessler I find it a bit inconvenient because the editing screen requires placing an `x` inside the checkbox `[ ]`. If github allows having PR template that works like issue template, I think that would make more sense

ngxson:
I think this is pretty much a QoL improvement for both contributors & maintainers, making it clear how to write a PR description

Since this doesn't include any functional or policy changes, could we have it merged soon @ggerganov @ggml-org/maintainers ? Thanks.

ggerganov (approved):

PR #20844, opened 2026-03-21
Do not trigger grammar inside tool calling section

pwilkin:
Further changes regarding the parser / sampler interaction. Now keeps a state inside the slot to track reasoning status. When inside reasoning, grammar triggers are disabled. That fixes errors like #20837 where the grammar thinks the tool call inside the reasoning section is a real tool call and forces EOG after it's done.

Also, when if we're inside the reasoning section and we have an EOG, replace the EOG with the end-of-thinking marker instead to force the model to generate a real response.

AI disclosure: I told Claude what to do and yelled at it a bit.


ggerganov (changes requested):
This does not look good. We cannot spread the parsing logic all over server and libllama. It must be contained in the `common_sampler`.

- Can the tool-calling grammar be formulated so that it detects if it is in a reasoning block? If not, what is the obstacle?
- If it cannot be formulated, then the other option I can think of is to repurpose the reasoning budget sampler into a more generic parsing sampler and put all such logic inside of it, including the tool-calling grammar.

aldehir:
I think there are a few alternatives here, starting with the simplest: include the reasoning tokens in the trigger pattern. Something like:

```
^(<think>[^<]*</think>|[^<]*)<tool_call>$
```

Another approach could be to expand the grammar to support a trigger rule. In other words, always feed the token to the PDA but only constrain when the trigger rule is reached.

I think we should try to see if a trigger pattern can be composed for this as a short term solution.

pwilkin:
Aight, will rewrite.

ggerganov:
> Another approach could be to expand the grammar to support a trigger rule. In other words, always feed the token to the PDA but only constrain when the trigger rule is reached.

I think we already support this with the lazy activation of the grammar via trigger words/tokens:

https://github.com/ggml-org/llama.cpp/blob/db9d8aa428012cc5593e18635d4c3c54095f5138/include/llama.h#L1359-L1369

pwilkin:
Nope, what we're talking about is "second level laziness" (delaying the lazy triggers themselves).

aldehir:
> I think we already support this with the lazy activation of the grammar via trigger words/tokens:

A trigger word/token may trigger prematurely when semantically invalid, e.g. in the reasoning output.

A trigger pattern will work, since the output is regular until it generates JSON. The only downside is the use of `std::regex`, which may crash on large output.

My other idea is to use the grammar sampler itself to parse up until the tool call. It would be nested within the grammar sampler as an alternative trigger type, and would not use `std::regex`.

pwilkin:
My currently winning idea is adding a token state machine to the grammar sampler since that can be also later extended to cache the graph transitions for backend sampling.

aldehir:
Regex is a state machine, I think we do the simplest solution first to resolve issues before trying a more complex solution.

Plus, I have my doubts on using tokens as the atomic unit. We do that already in the current grammar implementation and it makes it awkward to handle partial UTF8 sequences. I believe using bytes is far more flexible and you can represent UTF8 as a DFA. I already have a working prototype in my repo, with caching. It was heavily vibecoded, as I was exploring ideas, but I plan to rewrite it by hand as my next potential contribution.

ggerganov:
> My currently winning idea is adding a token state machine to the grammar sampler since that can be also later extended to cache the graph transitions for backend sampling.

I'm not sure I am following. IMO, there is already available functionality and we don't need to add new stuff to the grammar. The reasoning budget sampler "knows" when we are inside a reasoning block - the tool grammar can be active only outside the reasoning block.

> Regex is a state machine, I think we do the simplest solution first to resolve issues before trying a more complex solution.

Yes, we definitely need to consolidate and make things stable, instead of adding more new stuff that breaks and have to fix it with more new stuff.

pwilkin:
Okay, if we want to go with the simplest route I think we should use @aldehir 's idea and just add the thinking block to the grammar trigger. My only caveat with it is that it might be too slow - the issue is `^(<think>[^<]*</think>[^<]*)<tool_call>$` assumes there's no thinking + content before tool call, which fails for many models and if we want to do it properly, we'd have to use the annoying `[\s\S]*` which is a performance nightmare (since you also can't rely on the assumption that there's no `<` characters in the thinking block).

Moving the tool grammar as an internal `llama_grammar_sampler` inside the reasoning budget sampler might actually be the most plausible idea if you don't want any sort of communication between samplers / shared state.

aldehir:
I don't believe it will have any measurable impact on performance. Have you benchmarked it to show otherwise?

pwilkin:
Not yet, will do it and see.

aldehir:
BTW I used `[^<]*` as a shorthand for the exclusion pattern we use in gbnf. They're a bit a annoying to type, I should have clarified. You can do that to minimize backtracking.

Anchoring the regex will also help so the engine does not attempt to match the pattern at every position. This isn't default behavior, it is handled in the grammar sampler. If a pattern is wrapped in `^...$` it will use the `continuous` flag (https://en.cppreference.com/w/cpp/regex/match_flag_type.html).

Also, we should nix the idea of forcing the think closing tag on EOG if within reasoning. The parser should be written to support this, and it is capable of doing so.

pwilkin:
@aldehir @ggerganov okay, so it turned out quite a bit more difficult, but I do have a clean solution, replaces the old one.

AI disclosure: nope, had to write it all by hand, Opus was completely braindead here, kept proposing incorrect patterns etc. LLMs struggle so much with grammar stuff.

pwilkin:
Update: unfortunately the stupid regex engine crashes on long reasoning chains.

pwilkin:
I think that the most viable path forward now is option 2: repurpose the reasoning budget sampler (rename it to "reasoning sampler"?) and make it store the tool grammar sampler inside (only activated when we hit the trigger condition).

ggerganov:
> Update: unfortunately the stupid regex engine crashes on long reasoning chains.

Is this the infamous [std::regex running out of stack memory](https://developercommunity.visualstudio.com/t/grouping-within-repetition-causes-regex-stack-erro/885115) issue?

pwilkin:
Yup. On gdb backtrace I got a trace that was 160000 (yes, one hundred sixty thousand) stack frames long.

pwilkin:
All righty, rewrite three.

This time, I did the second original @ggerganov proposal:
* extend `reasoning-sampler` into a more general `matching-sampler`: idea is a state machine that matches / counts tokens and then toggles some sampler based on that
* move the tool grammar case from the main grammar sampler to the matching sampler as a special case that triggers a **non-lazy** grammar when outside of reasoning and the trigger word is captured, replaying the trigger word first

aldehir:
Alternative approach in #20970 that handles this in `common_sampler` where it has visibility of both the reasoning budget and grammar samplers.

pwilkin:
Superseded by #20970

PR #20845, opened 2026-03-21
Restore NVFP4 per-tensor scale precision 

michaelw9999:
This PR is a pre-requisite for the  PR [https://github.com/ggml-org/llama.cpp/pull/20644](https://github.com/ggml-org/llama.cpp/pull/20644) which will introduce basic CUDA support (mmvq/dp4a only) for NVFP4. 
NVFP4 is currently unique in that it has a dual scaling design using a float scale per tensor in addition to the per-subblock scale.  
These NVFP4 tensor scales are critical with 4-bit NVFP4 being low precision, and they need to be part of the initial weight interpretation at the first `ggml_mul_mat`.
The current implementation is applying some tensor scales through `build_lora_mm` and `build_lora_mm_id`:
```
res = ggml_mul(ctx0, res, w_s);`
```
This effectively means NVFP4's per-tensor scale is applied **after** the entire _mul_mat_ is complete.  But that is not OK: the late application loses precision because the operation is done just once after accumulation.  Effectively, to get the most precision, the 4-bit weight needs the full 32-bit tensor scale applied to it without any loss of precision of that tensor scale itself.
This PR changes the application of the NVFP4 per tensor scale so it can be applied **as part of** the _mul_mat_.
It does this by saving the tensor scale in `build_lora_mm` to `src[2]` ([3] for ids) so that it's not lost and can be used later at the time of the vec_dot:

```
float nvfp4_t_s = ((const float *) dst->src[2]->data)[0];
if (dst->src[2]->ne[0] > 1) {
  nvfp4_t_s = ((const float *) dst->src[2]->data)[i02];
 }
tmp[ir0 - iir0] = tmp[ir0 - iir0] * nvfp4_t_s;
```

On the CPU version of NVFP4, there is no significant observed difference with this fix, but it makes a very significant difference on GPU with the F16 path, and even more significantly with models like Qwen3.5 and Nemotron that have special experts:

| Test| Baseline | After fix | Change |
| --- | ---: | ---: | --- |
| Runtime CPU capture RMS | 0.00377982258 | 0.00377982791 | +0.00% |
| Q8 float accumulation RMS | 0.00377982258 | 0.00377982791 | +0.00% |
| F16 accumulation RMS | inf | 0.0059391915 | Stops Overflow  |
| F32 accumulation RMS | 0.000148721874 | 0.000217344128 | +46% (Normal GPU) |
| dot magnitude   | 71785.6172 | 12.3097782 | Normal |

On Qwen3.5-0.8B:
```
After accmulation:
- attn_qkv:  baseline16_rms=inf, fix_f16_rms=0.386774
- attn_gate: baseline16_rms=inf, fix_f16_rms=0.0511276
- ssm_out:  baseline16_rms=inf, fix_f16_rms=0.403415
- ssm_alpha: baseline16_rms=inf, fix_f16_rms=0.0546128
- ssm_beta:  baseline16_rms=inf, fix_f16_rms=0.0458366

F16 conversions would result in unscaled exact dot of about -71785.6 causing _inf_ and exceeding half_max_finite=65504

Baseline CPU:
- Mean PPL(Q)   = 21.000743
- Mean KLD      = 0.094119
- Maximum KLD   = 7.885083
- RMS Delta p   = 7.233
- Same top p    = 84.245%
Fix CPU:
--Tested only for 8 chunks:100% equivalence
(Full kld via cpu >12 hours and previously done with near-identical pre-PR))

Fix CUDA: 
- Mean PPL(Q)   = 21.007698
- Mean KLD      = 0.093654
- Maximum KLD   = 7.336067
- RMS Delta p   = 7.223
- Same top p    = 84.337%
```
For Qwen3.5-0.8B-Base:
```
Baseline (CUDA):
Mean PPL(Q) = 16.354339
Mean KLD = 0.164407
Maximum KLD = 12.685198
RMS Δp = 10.490
Same top p = 79.721%

Fix (CUDA):
Mean PPL(Q) = 15.226978
Mean PPL(base) = 13.932563
Mean KLD = 0.088797
Maximum KLD = 5.984150
RMS Δp = 7.439
Same top p = 84.707%
```

For Qwen3-4B:
```
Baseline CUDA:
- Mean PPL(Q)   = 15.508894
- Mean KLD      = 0.566403
- Maximum KLD   = 31.852549
- RMS Delta p   = 20.417
- Same top p    = 72.481%

With fix CUDA:
- Mean PPL(Q)   = 14.936430
- Mean KLD      = 0.522256
- Maximum KLD   = 31.771683
- RMS Delta p   = 19.369
- Same top p    = 73.774%
```
Also tested with Qwen3.5-35B, Qwen3.5-27B, Nemotron3-30B, Polaris-4B.
Anecdotally, the conversations with the models also seem "better", with Qwen3.5 in particular giving answers closer to the BF16 model with more emojis and less terseness.

This fix could be better integrated without branching specifically for NVFP4, but it would require more re-work and a bigger refactor to find a place to store and apply the per-tensor scales, as currently there is no easy way to do this without either having a separate `ggml_vec_dot_t` or changing args that would affect every other quant as well as where to save the scale.
AI assistance and code was used write the in-run test functions to capture RMS (not included) and help evaluate the error and clean up my code under my direction.

### DIFF
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -1422,6 +1422,20 @@ extern "C" {
             struct ggml_tensor  * b,
             struct ggml_tensor  * ids);
 
+    GGML_API void ggml_mul_mat_add_scale(
+            struct ggml_tensor * a,
+            struct ggml_tensor * scale);
+
+    GGML_API const struct ggml_tensor * ggml_mul_mat_get_scale(
+            const struct ggml_tensor * a);
+
+    GGML_API void ggml_mul_mat_id_add_scale(
+            struct ggml_tensor * a,
+            struct ggml_tensor * scale);
+
+    GGML_API const struct ggml_tensor * ggml_mul_mat_id_get_scale(
+            const struct ggml_tensor * a);
+
     // A: m columns, n rows,
     // B: p columns, n rows,
     // result is m columns, p rows
--- a/ggml/src/ggml-cpu/ggml-cpu.c
+++ b/ggml/src/ggml-cpu/ggml-cpu.c
@@ -1222,6 +1222,16 @@ static void ggml_compute_forward_mul_mat_one_chunk(
 
                 for (int64_t ir0 = iir0; ir0 < iir0 + blck_0 && ir0 < ir0_end; ir0 += num_rows_per_vec_dot) {
                     vec_dot(ne00, &tmp[ir0 - iir0], (num_rows_per_vec_dot > 1 ? 16 : 0), src0_row + ir0 * nb01, (num_rows_per_vec_dot > 1 ? nb01 : 0), src1_col, (num_rows_per_vec_dot > 1 ? src1_col_stride : 0), num_rows_per_vec_dot);
+
+                    // NVFP4's per tensor scale needs to be applied with the vecdot
+                    // this should be put into a future traits based approach to be cleaner
+                    if (src0->type == GGML_TYPE_NVFP4 && dst->src[2]) {
+                        float nvfp4_t_s = ((const float *) dst->src[2]->data)[0];
+                        if (dst->src[2]->ne[0] > 1) {
+                            nvfp4_t_s = ((const float *) dst->src[2]->data)[i02];
+                        }
+                        tmp[ir0 - iir0] = tmp[ir0 - iir0] * nvfp4_t_s;
+                    }
                 }
 
                 for (int cn = 0; cn < num_rows_per_vec_dot; ++cn) {
@@ -1490,6 +1500,14 @@ static void ggml_compute_forward_mul_mat_id_one_chunk(
 
                 for (int64_t ir0 = iir0; ir0 < iir0 + blck_0 && ir0 < ir0_end; ++ir0) {
                     vec_dot(ne00, &tmp[ir0 - iir0], 0, src0_cur + ir0*nb01, 0, src1_col, 0, 1);
+
+                    if (src0->type == GGML_TYPE_NVFP4 && dst->src[3]) {
+                        float nvfp4_t_s = ((const float *) dst->src[3]->data)[0];
+                        if (dst->src[3]->ne[0] > 1) {
+                            nvfp4_t_s = ((const float *) dst->src[3]->data)[cur_a];
+                        }
+                        tmp[ir0 - iir0] = tmp[ir0 - iir0] * nvfp4_t_s;
+                    }
                 }
 
                 memcpy(&dst_col[iir0], tmp, (MIN(iir0 + blck_0, ir0_end) - iir0)*sizeof(float));
--- a/ggml/src/ggml-cpu/ggml-cpu.c
+++ b/ggml/src/ggml-cpu/ggml-cpu.c
@@ -1225,10 +1225,10 @@ static void ggml_compute_forward_mul_mat_one_chunk(
 
                     // NVFP4's per tensor scale needs to be applied with the vecdot
                     // this should be put into a future traits based approach to be cleaner
-                    if (src0->type == GGML_TYPE_NVFP4 && dst->src[2]) {
-                        float nvfp4_t_s = ((const float *) dst->src[2]->data)[0];
-                        if (dst->src[2]->ne[0] > 1) {
-                            nvfp4_t_s = ((const float *) dst->src[2]->data)[i02];
+                    if (src0->type == GGML_TYPE_NVFP4) {
+                        float nvfp4_t_s = ((const float *) ggml_mul_mat_get_scale(dst)->data)[0];
+                        if (ggml_mul_mat_get_scale(dst)->ne[0] > 1) {
+                            nvfp4_t_s = ((const float *) ggml_mul_mat_get_scale(dst)->data)[i02];
                         }
                         tmp[ir0 - iir0] = tmp[ir0 - iir0] * nvfp4_t_s;
                     }
@@ -1501,10 +1501,10 @@ static void ggml_compute_forward_mul_mat_id_one_chunk(
                 for (int64_t ir0 = iir0; ir0 < iir0 + blck_0 && ir0 < ir0_end; ++ir0) {
                     vec_dot(ne00, &tmp[ir0 - iir0], 0, src0_cur + ir0*nb01, 0, src1_col, 0, 1);
 
-                    if (src0->type == GGML_TYPE_NVFP4 && dst->src[3]) {
-                        float nvfp4_t_s = ((const float *) dst->src[3]->data)[0];
-                        if (dst->src[3]->ne[0] > 1) {
-                            nvfp4_t_s = ((const float *) dst->src[3]->data)[cur_a];
+                    if (src0->type == GGML_TYPE_NVFP4) {
+                        float nvfp4_t_s = ((const float *) ggml_mul_mat_id_get_scale(dst)->data)[0];
+                        if (ggml_mul_mat_id_get_scale(dst)->ne[0] > 1) {
+                            nvfp4_t_s = ((const float *) ggml_mul_mat_id_get_scale(dst)->data)[cur_a];
                         }
                         tmp[ir0 - iir0] = tmp[ir0 - iir0] * nvfp4_t_s;
                     }
--- a/ggml/src/ggml.c
+++ b/ggml/src/ggml.c
@@ -3280,6 +3280,28 @@ struct ggml_tensor * ggml_mul_mat_id(
     return result;
 }
 
+GGML_API void ggml_mul_mat_add_scale(
+        struct ggml_tensor * a,
+        struct ggml_tensor * scale) {
+    a->src[2] = scale;
+}
+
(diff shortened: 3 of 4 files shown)


ggerganov:
I think the proper way to do that would be to add the following functions:

```c
    ...
    // add a scaling tensor to src[2]
    GGML_API void ggml_mul_mat_add_scale(
            struct ggml_tensor * a,
            struct ggml_tensor * scale);

    // add a bias tensor to src[3]
    // not needed for now
    GGML_API void ggml_mul_mat_add_bias(
            struct ggml_tensor * a,
            struct ggml_tensor * bias);

    ...

    // add a scaling tensor to src[3]
    GGML_API void ggml_mul_mat_id_add_scale(
            struct ggml_tensor * a,
            struct ggml_tensor * scale);

    // add a bias tensor to src[4]
    // not needed for now
    GGML_API void ggml_mul_mat_id_add_bias(
            struct ggml_tensor * a,
            struct ggml_tensor * bias);
```

This is similar to how we add optional sinks to the flash attention op: `ggml_flash_attn_ext_add_sinks()`.

cc @JohannesGaessler, @richarddd for viz and opinion. This should be resolved before other NVFP4 changes.

JohannesGaessler:
>This effectively means NVFP4's per-tensor scale is applied after the entire mul_mat is complete. But that is not OK: the late application loses precision because the operation is done just once after accumulation. Effectively, to get the most precision, the 4-bit weight needs the full 32-bit tensor scale applied to it without any loss of precision of that tensor scale itself.

I don't understand why there would be precision issues (I can see why there may be issues with the numerical range). But in any case, did you consider fixing this via operator fusion of `GGML_OP_MUL_MAT` and `GGML_OP_MUL`?

ggerganov:
Correct, it's a numerical range problem when the underlying implementation uses f16 accumulators.

> did you consider fixing this via operator fusion of GGML_OP_MUL_MAT and GGML_OP_MUL?

Fusion can fix this, but still not sure if it OK for a graph to work correctly only when fusion is enabled. It might complicate debugging in the future.

I was initially thinking that fusion in this case would be needed just from performance perspective, but at that time I didn't consider the numerical range problem.

JohannesGaessler:
I mean, I think that that is a matter of how we define a graph "working correctly". Due to numerical limitations it's basically impossible to guarantee correct results for arbitrary inputs. We do have an enum `ggml_prec` to control the minimum precision that is needed for a matrix multiplication. So my opinion is that we should make use of that first and have the ggml backends respect it. For the linked CUDA PR I am requiring the initial implementation to use dequantization + cuBLAS because I'm currently stretched thin and that is the minimal viable implementation that I can review and merge. But going forward we should add MMQ support where the accumulation is always done using FP32 precision so this wouldn't be an issue.

ggerganov:
Got it - makes sense.

Just making sure my understanding is correct - we'll basically have the following logic:

- By default, the computation uses F16 for accumulation (as usual)
- Therefore, to avoid numerical range problems with NVFP4 multiplications, we will call `ggml_prec` in `llama.cpp` to force F32 accumulators for all NVFP4 multiplications. In the CUDA backend this will initially perform dequantize + F32 cuBLAS
- Later on, MMQ implementation for NVFP4 with F32 accumulation will be added that should work across all NVIDIA hardware and should be faster than the "dequantize + F32 cuBLAS" path
- Further on, the fusion logic in the CUDA backend can be extend so that if it detects `GGML_OP_MUL_MAT + GGML_OP_MUL` with NVFP4 data, it can utilize the FP4 compute, given that the device has the compute capability for it

michaelw9999:
More info:
At first, no issues with CUDA but this was with full MMQ/MMVQ kernels, before Qwen3.5 or Nemotron 3 models were out.  It was not until forcing cubLAS and running full CPU kld runs across a few models that I saw problems with the inferior results on the CUDA versions.

With standard precision for cublas at F16, now it would crash or show junk (I can't remember), for *almost all* models, because of the F16 overflow.  Switching to F32 solved the inf, but it **still** did not have as good results as CPU.  Once back on F32(cublas) for Qwen3.5/Nemotron, they were *still* poor (along with other bugs along the way possibly interfering with my results, however I was merging locally daily every day this week).  Specifically, anything Nemotron/Qwen3.5, I've come to realize they are more sensitive.

But that was because of their inclusion as new tensor scales in `build_lora_xx`.
I saw `build_ffn` did it the same way, as did `build_moe_ffn`.

Effectively, for those scales, it's ggml_mul_mat() first, then ggml_mul()  
Algebraically it makes sense in theory this should be fine, but in reality, because of floating point differences and the tiny scale accumulating. it is not equivalent.

From studying and doing real research (I am by no means a mathematician or anywhere close, so apologies, I may be incorrectly paraphrasing), this type of inclusion of the scale is considered a "late" application of the scale (which is also how the hardware SM120 TC does it too, but to a lesser extent).  
And, it's okay to do so, to some extend. Doing tensor_scale * weight individually for every single weight one by one would be very slow even by GPU that's not feasible   So it's okay to this late scaling and to do it for some small group size (eg, 16/64/256).

But too many floating point errors add up if that is done across the entire tensor itself or for a huge K, especially for very tiny scales as well as with low bit precision like NVFP4/4-bit, so you end up getting what I've recently learned is called "**swamping**" - I had never heard of this until the other day (See: [https://proceedings.neurips.cc/paper_files/paper/2018/file/335d3d1cd7ef05ec77714a215134914c-Paper.pdf](https://proceedings.neurips.cc/paper_files/paper/2018/file/335d3d1cd7ef05ec77714a215134914c-Paper.pdf )  and [https://arxiv.org/html/2404.14010v1](https://arxiv.org/html/2404.14010v1)   

Switching (as shown this PR) to vecdot * scale, instead of mulmat * scale, made a massive difference; you can see the huge improvement for some of the changes, it is not as significant on some models as others, but still significant.  (I have mountains of tensor raw output pre/post error values I can still go through). I believe that has to do with the model itself and how some with high outlier kld tails can bring out more error than others.  The issue does not show itself on CPU because of how a CPU does its math but on GPU it is a big deal as apparently the GPU just starts losing bits entirely.

Switching to vecdot*scale immediately but before the final accumulation is done pushes the tensor scale op earlier, and so it retains the precision and stops the error.  

Doing that allowed F16 to work just fine on cublas, so F32 cublas precision is  not necessary after fixing the mul order.   I did not notice any change in performance doing so, if anything it went up higher.  (29k/470k pp512/tg128 on Qwen3.5 0.8B).  F32 cublas is about 25% slower than F16 as I observed on SM120.

So to summarize: 
It is the single "late" post mulmat * tensorscale that causes too much FP loss due to GPU swamping with tensors/scales/models seen on NVFP4, on GPU.  Just switching to F32 accumulator precision is not the best solution.  The tensorscale must be multiplied earlier to the product of the vecdot.  

As to how to apply that the cleanest and best way into llama.cpp I will look at the suggestions.  If you have an even better idea after this comment's detail information, please let me know.  Maybe there is an even simpler way outside of build_lora_xx altogether.  There may be other dual-block scale quantization formats in the future too. (See  [Eg, TetraJet v2](https://arxiv.org/html/2510.27527v1 based on NVFP4)  if you have spare time and are interested)

JohannesGaessler:
Sorry, but I think you have a poor understanding of floating-point arithmetic - if you are seeing huge changes from when you apply the scale to me that rather suggests that there is something wrong with your code. And I think that the solutions you propose tend to introduce too much complexity that I frequently think is not justified. I obviously don't know what your workflow is but my honest guess is that you are using language models for insight and absorb their outputs too uncritically. As it is I would rather have the basic implementation be done by a contributor with a more established track record. @am17an are you up for it?

am17an:
I can take this up. However, I don't think this is a big issue(correct me if I'm wrong) because MMQ is always used unless you force cublas, even then we have added #19959 which can force cublas to do f32 accumulation via env flag. For the blackwell path, MMQ would still use f32 accumulation.

michaelw9999:
> Sorry, but I think you have a poor understanding of floating-point arithmetic - if you are seeing huge changes from when you apply the scale to me that rather suggests that there is something wrong with your code. And I think that the solutions you propose tend to introduce too much complexity that I frequently think is not justified. I obviously don't know what your workflow is but my honest guess is that you are using language models for insight and absorb their outputs too uncritically. As it is I would rather have the basic implementation be done by a contributor with a more established track record. @am17an are you up for it?

 I am sorry for any trouble not getting things perfect.  I am still very new to ML and self-learning as much as I can - really just for fun.  I have been programming since Pascal days with Singles and Doubles but I am by no means will I ever call myself an expert. I just saw this post after I went and finished wiring back up the dp4a version .  My background is systems/bio-engineering so yes I do tend to over-do and over-think things at first but then there is always time afterwards to prune down and get feedback.  Most of my big ideas are inspired from reading real publications as well as the actual code, not just LLMs.  If @am17an wants to take over 20644 - it really is working very well now and you can use that as a base to save time.  I am seeing equivalent kld to CPU now on Nemotron 30B which was the hardest one to get right.  I can upload the exact models to HF if you want to test them but it's easy to convert too.  Nemotron is up and running on llama-server if you want to try it out.

PR #20849, opened 2026-03-21
server: allow router to report child instances sleep status

ngxson:
Allow router server to report child sleeping status via `/models` endpoint, example:

```
{
  "data": [{
    "id": "LiquidAI/LFM2-350M-GGUF",
    "aliases": [],
    "tags": [],
    "object": "model",
    "owned_by": "llamacpp",
    "created": 1774133483,
    "status": {
      "value": "sleeping",
      "args": [....],
      "preset": "[LiquidAI/LFM2-350M-GGUF]\nsleep-idle-seconds = 10\nhf-repo = LiquidAI/LFM2-350M-GGUF\n\n"
    }
  },
...
```


tools/server/server-models.h
```diff
@@ -56,6 +56,7 @@ struct server_model_meta {
     std::set<std::string> tags;    // informational tags, not used for routing
     int port = 0;
     server_model_status status = SERVER_MODEL_STATUS_UNLOADED;
+    bool is_sleeping = false; // whether the model is in sleeping state (only valid when status == LOADED)
```
ggerganov: Can we have a new status `SERVER_MODEL_STATUS_SLEEPING` instead of the extra `bool`?

ngxson: Yeah it's a bit complicated but still possible, my initial plan is to consider "sleeping" as a sub-state of "loaded". but on second thought, it maybe more beneficial in the long term to have just one set of states.

I implemented it in https://github.com/ggml-org/llama.cpp/pull/20849/commits/02c8b8855dbd02ba09609d4e59c8e0acb02b1f86 , it also requires changing the semantic of "loaded" in some places to be "ready", with "ready" means either "loaded" or "sleeping"

ggerganov (approved):

Issue #20227, opened 2026-03-08
querying /metrics keeps model awake.


wbste:
Running the latest llama.cpp with model router and something like `llama-server --models-preset presets.ini --sleep-idle-seconds 10 --models-max 2`

**My end goal here to is track metrics of pp, tg, etc provided by the `/metrics` endpoint**

I feel like I'm missing something obvious here 😄 

1. I run 1 or more LLMs.
2. I want to query what is actually loaded, so I use `/models` and look for the status of "loaded".
3. I then get the model specific metrics at `/metrics...` by passing the model's `id` parameter.
4. This counts as a "query" so the sleep counter resets.

It seems there's no way of getting the data without keeping the models on all the time (other than querying once per load basically)?

Note `autoload=false` doesn't seem to bypass the "resetting of the clock" if the model is already loaded.

dsluo:
I think the solution proposed in #19197 would address this.

Cirius1792:
Confirming this issue as well. When configuring llama.cpp with Prometheus and llama server in router mode prevents models from being unloaded and swapped because of the metrics endpoint invocation on the currently loaded model

wbste:
Just because it came up again...(ai summary incoming):

Traced against llama.cpp master (tools/server). Querying `/metrics` for a model always resets that model's `--sleep-idle-seconds` countdown and wakes it if sleeping, so there is no way to sample per-model metrics without keeping the model alive.

## Root cause

The idle timer is `time_last_task`, and it is stamped unconditionally after every queue pass:

- `server-queue.cpp:136` - sleep decision: `(now - time_last_task) >= idle_sleep_ms`
- `server-queue.cpp:167` - after `callback_update_slots()`: `time_last_task = ggml_time_ms()` runs for any task that went through the loop

`/metrics` is task-queue backed, so it triggers that stamp:

- `server-context.cpp:4370` `get_metrics` calls `create_response()` without `bypass_sleep`
- `server-context.cpp:4039-4040` the response ctor calls `queue_tasks.wait_until_no_sleep()`, waking a sleeping child
- `server-context.cpp:4379-4381` it posts a `SERVER_TASK_TYPE_METRICS` task, whose processing stamps `time_last_task`

## Asymmetry with the free endpoints

`/models`, `/health`, `/props` do not wake or reset, because they pass `bypass_sleep = true` and post no task:

- `server-context.cpp:4359` `get_health` -> `create_response(true)`
[...]

PR #20850, opened 2026-03-21
server: add reset_idle_timer flag to server_task

kilofox:
This fixes issue #20227 where querying the `/metrics` endpoint would prevent the server from going to sleep, making the  `--sleep-idle-seconds` ineffective when monitoring llama.cpp with e.g Prometheus.

My PR adds a new `reset_idle_timer` field to `server_task` (defaults to true for backwards compatibility). When handlers for `/metrics` and `/slots` create their tasks, they now set this to false so those requests don't count as activity that would reset the sleep timer.

Full disclosure: it's been a while since I've written C++ and I leaned on a coding agent to help review my changes.

### DIFF
--- a/tools/server/README.md
+++ b/tools/server/README.md
@@ -1704,6 +1704,8 @@ Note that the following endpoints are exempt from being considered as incoming t
 - `GET /health`
 - `GET /props`
 - `GET /models`
+- `GET /metrics`
+- `GET /slots`
 
 ## More examples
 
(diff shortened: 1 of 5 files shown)


ngxson:
Unfortunately I cannot accept this solution as-is. Any accesses to the task queue MUST strictly be done when server is NOT in sleeping state.

Your solution will fail whenever `/slots` calls detokenize to get the prompt.

PR #20853, opened 2026-03-22
webui: fix nested button elements in model selector dropdown

### LINKED ISSUE #20832: Misc. bug: WebUI: Nested button elements in model selector dropdown
### Name and Version

version: 8445 (c46583b86)

### Operating systems

_No response_

llama-server

### Command line

```shell

```

Currently, the cursor display/hinting no longer works correctly for the model selector dropdown.

In the final HTML I've found **three levels of nested `<button>`** elements 🥵 .
These are likely causing the browser misbehavior:

Image

> [!NOTE]
> `<button>` elements must never be nested inside other `<button>` elements, as this can lead to event handling issues and other unexpected side effects.

## Related files

https://github.com/ggml-org/llama.cpp/blob/master/tools/server/webui/src/lib/components/ui/tooltip/tooltip-trigger.svelte

https://github.com/ggml-org/llama.cpp/blob/master/tools/server/webui/src/lib/components/ui/dropdown-menu/dropdown-menu-trigger.svelte

https://github.com/bluemoehre/llama.cpp/blob/master/tools/server/webui/src/lib/components/app/models/ModelsSelector.svelte

### First Bad Commit

_No response_

### Relevant log output

_No response_

majiayu000:
Fixes #20832

## Problem

`ModelsSelector.svelte` has nested `<button>` elements: `DropdownMenu.Trigger` (renders `<button>`) wraps an explicit `<button>` which wraps `Tooltip.Trigger` (renders `<button>`). This produces invalid HTML (`<button><button><button>`).

## Fix

**Router mode:** Use `{#snippet child({ props })}` on `DropdownMenu.Trigger` so it delegates to the explicit `<button>` instead of wrapping it. Change `Tooltip.Trigger` to delegate to a `<span>` via the same snippet pattern.

Result: `<button><button><button>` → `<button><span>`

**Non-router mode:** Same `Tooltip.Trigger` fix — delegate to `<span>` to avoid nesting inside the parent `<button>`.

The `child` snippet is bits-ui v2's composition pattern (equivalent to Radix's `asChild`), already used elsewhere in this codebase (sidebar-menu-button.svelte, tooltip-content.svelte, etc). Also removed a duplicate `disabled` attribute on the dropdown trigger button that was being set both directly and via spread props.


ServeurpersoCom:
Wait, just send us a screenshot of the problem and it will be fixed quickly, don't use AI to automatically post here!

PR #20864, opened 2026-03-22
jinja : refactor token advancement

CISC:
Refactor token advancement to catch unexpected EOF.

### DIFF
--- a/common/jinja/parser.cpp
+++ b/common/jinja/parser.cpp
@@ -53,6 +53,13 @@ private:
         return tokens[current + offset];
     }
 
+    const token & next() {
+        if (current >= tokens.size()) {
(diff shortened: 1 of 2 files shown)


common/jinja/parser.cpp
```diff
@@ -202,7 +208,7 @@ class parser {
             // Ignore generation blocks (transformers-specific)
             // See https://github.com/huggingface/transformers/pull/30650 for more information.
             result = mk_stmt<noop_statement>(start_pos);
-            current++;
+            ++current;
```
ngxson, line 211: is it possible to call `next()` here? so we consume the token but do nothing

CISC: Yes, that's what I initially did. :)

Changed it back though as we don't really need to check as long as we're not reading the token.

ngxson (approved):

PR #20865, opened 2026-03-22
CUDA: fix BF16 FA compilation

JohannesGaessler:
See https://github.com/ggml-org/llama.cpp/pull/20525#issuecomment-4106067370 .


am17an (approved):
waves fist at HIP

PR #20869, opened 2026-03-22
vulkan: add shared memory staging for MMVQ kernels (>2x tg speedup on Intel Arc)

TheBlueMatt:
**Note that this code was written by Claude. While I did review and understand it at a high level, I am no vulkan expert. Still, a >2x performance gain is pretty substantial, so it seemed worth cleaning up and opening, which I'm doing here.** At least one other contributor also concluded it might be acceptable to open at https://github.com/ggml-org/llama.cpp/issues/18808#issuecomment-4104510623 but of course feel free to close if you prefer.

Introduce an alternative code path for quantized matrix-vector multiplication (MMVQ) shaders that stages the A-matrix through shared memory using coalesced uint64_t loads, rather than reading quantization blocks directly from global memory with per-field accessors.

The standard MMVQ path reads quantized blocks field-by-field from global memory (e.g. data_a[ib].d, data_a[ib].qs[...]), which produces many small, scattered loads. On architectures where this access pattern is expensive (notably Intel Xe2), this creates a bottleneck. The new path instead:

1. Loads entire tiles of quantization blocks into shared memory as raw uint64_t words, using coalesced wide loads across all threads in the workgroup.
2. Provides byte-level read helpers (shmem_read_i32, shmem_read_f16, shmem_read_u8, shmem_read_u16) that extract fields from the raw shared memory buffer with proper alignment handling across u64 boundaries.
3. Reimplements get_dm/repack/mmvq_dot_product for every supported quantized type (Q4_0, Q4_1, Q5_0, Q5_1, Q8_0, MXFP4, Q2_K, Q3_K, Q4_K, Q5_K, Q6_K, IQ1_S, IQ1_M) to read from shared memory instead of global buffers when USE_SHMEM_STAGING is defined.

The feature is compiled as separate shader variants (*_shmem_staging) and selected at pipeline creation time based on a per-device flag (mmvq_shmem_staging). It defaults to enabled on Intel Xe2 GPUs and can be overridden via the GGML_VK_MMVQ_SHMEM_STAGING environment variable.

Also excludes Intel Xe2 from the subgroup_min_size override for integer dot product pipelines, since Xe2 should use its native subgroup size.

On an Intel Arc Pro B60, this provides a more than 2x speedup in token generation:
 * gpt-oss 20B MXFP4 MoE tg128: 25.66 ± 0.04 -> 56.64 ± 0.38

On two Intel Arc Pro B60s (with very slow PCIe 3.0 interconnects):
 * qwen35 27B Q8_0 tg128: 3.35 ± 0.03 -> 5.30 ± 0.34


HumerousGorgon:
Is this relevant only on Linux? Windows has MMVQ disabled for Intel GPUs since it was producing problems with speeds.

TheBlueMatt:
I have no idea. I haven't had any Windows boxes around in a decade, probably.

TheBlueMatt:
Ah, damn, managed to get a better version that gets ~61.7~ 74 tok/s tg128 for gpt-oss 20B. Might take some work to massage it into something PR'able, but closing for now.

TheBlueMatt:
Reopening as an alternative to #20897. #20897 is much faster, but as noted more annoying to maintain. This one by comparison keeps nearly all the existing logic with some duplication but largely the same structure so is much more maintainable.

TheBlueMatt:
https://github.com/ggml-org/llama.cpp/pull/20897 now has similar code structure so this isn't any different.

PR #20872, opened 2026-03-22
server : add SSE headers to fix streaming behind reverse proxy

nicois:
In certain situations, when llama-server is behind a reverse proxy such as nginx, the web UI freezes on "processing..." when the first chat message is received. The session initializes fine and MCP services connect, but once a message is submitted the streaming response never reaches the client.

"certain situations" relates to the size of the buffer. While small, things appear to work fine, but (for example) defining a large number of MCP servers is sufficient to trigger this condition.

Root cause: nginx enables proxy_buffering by default. For SSE (Server- Sent Events) streams, this means nginx buffers the entire response before forwarding any bytes to the client. Since SSE streams are long-lived, the client never receives the first chunk and appears to hang indefinitely.

The fix adds two response headers to all streaming (chunked) responses:

  X-Accel-Buffering: no  — instructs nginx to disable response buffering
  Cache-Control: no-cache — prevents caching of the event stream

These are the standard headers that SSE endpoints should emit for compatibility with reverse proxies. They have no effect when the server is accessed directly.


tools/server/tests/unit/test_mcp_multi_service.py
ngxson, line 1: I don't get how this test case is related to MCP as the file name suggests

please remove this file, test case is not required for trivial header changes like this

ngxson (changes requested):

tools/server/tests/unit/test_mcp_multi_service.py
nicois: While the problem is a general one, I could only trigger it when I enabled MCPs while running on nginx.

I'll remove the test.

CISC:
Bad merge.

PR #20880, opened 2026-03-22
Implementation of Phi-4-Mini-Flash-Reasoning on Llama.cpp

felipedilho:
Implementation of Phi-4-Mini-Flash-Reasoning inference, conversion from HF weights format into .gguf format and quantization of the same. The model, which is of a hybrid architecture, uses a combination of many different components, namely differential attention on all attention layers, SWA, Mamba-1, YOCO for a full attention layer, and a YOCO for the last Mamba-1 layer, cross attention layers, and GMUs (which could be seen as a "cross attention" analogue for Mamba layers).

I have used Claude Sonnet 4.6 and Gemini Pro 3.1 to help me understand the Llama.cpp codebase. I have read every line of code generated and made sure that it made sense. I compiled the forked project in my machine, converted the model from the Hugging Face weights format in .safetensors and .json files to a .gguf model file. I also quantized this .gguf to both Q8_0 and Q4_K_M and I ran all .gguf files with llama-cli, getting text outputs that are of the same nature as the ones outputted by the online Nvidia serving of this same model in "[https://build.nvidia.com/microsoft/phi-4-mini-flash-reasoning](https://build.nvidia.com/microsoft/phi-4-mini-flash-reasoning)" (One thing that I thought was a bug, is the fact that the model seems to ignore the previous prompt completely, meaning that if we ask it something like "How much is 2+2" it will eventually answer "\boxed{4}", but if we ask it after "How much is 1 plus the last result" it will answer that there is no last result or completely hallucinate some value, like 0. Ithought this was a bug, but the Nvidia serving has the exact same behaviour, I think that Microsoft just didn't train the model with extended conversations in mind, perhaps even choosing to actively ignore the previous prompts, as this is a model trained to solve mathematical questions, so I think this is a artifact of the model training regime).

I know that I am not a good programmer, and this is my first pull request ever, there sure are many things that could be better in my code, and it could also possibly be way more optimized to get more tokens per second out (another thing that I also thought is about the lambda vectors used in the differential attention, during training it seems they use 4 64-dim. vectors as a way to get stability during training, but for inference only a calculated 1-d scalar is really used, so I guess I could already pre-calculate this scalar during the .gguf model creation and not include the lambda vectors themselves, but I don't know if I should do this, since if there is some way to do continual training on the .gguf file, or lora generation, having these lambda vectors could be useful.).

I know that there are several things that you pleople will likely ask me to change, I am open to do those changes, you only have to tell me! I know that you people are possibly receiving lots of crackpot pull requests since LLMs began to become popular (so, I think since always? As this is a LLM inference open source project). But I really got the model to run in my PC! It took me two weeks after I read all the relevant papers for this model (since I am NOT from Microsoft, just a interested nerd that wanted to see how "fast" and "efficient" this model really was, wich was some of the claims made in the model's accompaning paper, which also tested some other architectures, like hybrid delta-net architectures which Qwen appears to have chosen, yet Microsoft's tests appeared to indicate that the Phi-4-Mini-Flash-Reasoning architecture, i.e. SambaY using differential attention, was more promissing (at least considering the training data used which was the same as that used for the more traditional completely transformer based Phi-4-Mini model that Llama.cpp supported basically from the get go). So although this is a "relatively" older model, I wanted to be able to run it in my PC to see if it was really all that in terms of "efficiency" and "performance" (according to the paper, this model outperformed Phi-4-Mini in all benchmarks that they did, with a comparable number of parameters, and being trained with the exact same training data as Phi-4-Mini), but I wasn't able to run it locally with vLLM because my PC and GPU are old and not very good.

As far as I know, this is the first model to be ported to Llama.cpp that uses differential attention (or differential transformer architecture, as it is referred in the paper that coined this mechanism), and GMUs (gated memory units), so there is this novel aspect to it, also this model mixes a lot of different things in it, so it could be seen as an "experimental" model (and could also explain why it wasn't ported to Llama.cpp yet). So, even if the model is not that good with comparison to the latest models like those of Qwen 3.5, I think that these idiosyncrasies might make it worth it to add to the models that Llama.cpp is able to run.  Also, although the Phi models seemed to have died off, a recent new image-text to text model in the Phi-4 family was published and soon afterwards was ported to Llama.cpp, so there is hope that there might be new Phi models coming in the future, and if any of them uses the same type of architecture as Phi-4-Mini-Flash-Reasoning, my implementation might be usefull to port any other newer Phi models to Llama.cpp (once again, I DON'T work for Microsoft, I am just a Mathematical Physics PhD student at the University of São Paulo, Brazil, that got way too interested in this model in particular).

I had an issue opened around the time this model was released (which was in July), the issue was opened in October of last year, Issue #16450 , but it was automatically closed for being inactive for 14 days since being marked as stale.


convert_hf_to_gguf.py
```diff
@@ -5453,6 +5453,279 @@ def prepare_tensors(self):
...
+
+        # Helper to ensure tensor is transposed in actual memory layout
+        def force_transpose_2d(tensor):
+            """Force transpose with contiguous memory layout via numpy"""
+            # Convert to numpy (forces materialization)
+            np_array = tensor.cpu().numpy()
+            # Transpose in numpy (creates new array with transposed data)
+            np_transposed = np.ascontiguousarray(np_array.T)
+            # Convert back to torch
+            return torch.from_numpy(np_transposed)
```
ngxson: why not using pytorch `transpose(-1, -2)` ?

convert_hf_to_gguf.py
```diff
@@ -5453,6 +5453,279 @@ def prepare_tensors(self):
...
+            # FIX: Remove force_transpose_2d!
+            return [(f"blk.{bid}.ffn_down.weight", data_torch)]
+
+        # ==================================================
+        # MAMBA LAYERS (type 0)
+        # ==================================================
+        if layer_type == 0:
+            if rest == "attn.in_proj.weight":
+                # Remove force_transpose_2d — ggml dimension reversal gives correct ne=[2560,10240]
+                return [(f"blk.{bid}.ssm_in.weight", data_torch)]
```
ngxson: use proper tensor mapping for these

convert_hf_to_gguf.py
ngxson, line 1: these changes don't follow existing pattern in `convert_hf_to_gguf.py`, please refactor

tools/cli/cli.cpp
```diff
@@ -508,6 +508,14 @@ int main(int argc, char ** argv) {
...
 
             ctx_cli.input_files.clear();
+            // PATCH for cli.cpp: make /clear also reset the model's memory state
+            // (KV cache, Mamba recurrent state, etc.)
+            {
+                llama_context * lctx = ctx_cli.ctx_server.get_llama_context();
+                if (lctx != nullptr) {
+                    llama_memory_clear(llama_get_memory(lctx), true);
+                }
+            }
```
ngxson: this is unrelated to the current PR, please move this to a dedicated one

src/llama-graph.cpp
```diff
@@ -2281,6 +2281,46 @@ ggml_tensor * llm_graph_context::build_attn(
     return cur;
 }
 
+// Implementations needed for Phi-4-Mini-Flash-Reasoning
```
ngxson: if these functions are used only by one model, move them to `phi4flash.cpp`

refer to qwen-next as an example

src/models/phi4flash.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+    const int64_t n = ggml_nelements(layer.attn_lambda_q1);
+    double dot1 = 0.0, dot2 = 0.0;
+    for (int64_t i = 0; i < n; ++i) {
+        dot1 += (double)lq1[i] * (double)lk1[i];
+        dot2 += (double)lq2[i] * (double)lk2[i];
+    }
+    out = (float)(std::exp(dot1) - std::exp(dot2));
+    return true;
+}
```
ngxson: I have no words for this...

ngxson (changes requested):
> I have read every line of code generated and made sure that it made sense.

Given the quality of this PR, I don't think your claim is honest.

felipedilho:
> > I have read every line of code generated and made sure that it made sense.
> 
> Given the quality of this PR, I don't think your claim is honest.

I am sorry for the bad quality of the Pull Request, and for the time you lost reading it, but in my defence I also said that "I am not a good programmer, and this is my first pull request ever", so there is no incompatibility with both things being true at the same time.

convert_hf_to_gguf.py
felipedilho: I changed to it in the latest commit.

convert_hf_to_gguf.py
felipedilho: I tried to adequate that in the last commit.

convert_hf_to_gguf.py
felipedilho: I have refactored in the last commit.

tools/cli/cli.cpp
felipedilho: I have removed this piece of code in the last commit.

src/llama-graph.cpp
felipedilho: I did exactly this in the last commit using qwen-next as a base.

src/models/phi4flash.cpp
felipedilho: Well, using a pre-calculated scalar, for each layer, at load time seems better than calculating two dot products, two scalar exponentials and a scalar subtraction at inference time. I have removed those checks I had there before and tried to improve the code aesthetics, but the underneath logic is the same. I'm sorry, but "I have no words for this..." isn't descriptive enough for me to do the modifications that you might want me to make. Again, this model uses "differential attention", I don't know if you are completely aware of all its implementation details... but I will suppose you do since it is more probable that I am in the wrong, rather than the other way around. If there is a better way to do differential attention than what I am doing here in the new version of the code in the newest commit, you are free to point it out with more details, I will try my best to try to close the gap between what you indicate and the current code. But I do need the direction!

PR #20888, opened 2026-03-23
fix(ggml): correct RISC-V ISA string canonical ordering for RVV in CMake

ihb2032:
## Motivation

When configuring the CMake build for RISC-V with both standard vector extensions `GGML_RVV` and multi-letter extensions, the generated `-march` string violates the RISC-V ISA canonical ordering rules. 

Currently, the single-letter `v` extension is appended after the multi-letter `z` extensions, e.g., resulting in `rv64gc_zfh_v...`. Modern, strict compilers reject this non-canonical ISA string and fail to build the CPU backend with:
`error: unexpected ISA string at end: ‘v_zvfh_zicbop_zihintpause’`

## Testing

Verified successful compilation on RISC-V using GCC 14.2 with `-DGGML_RVV=ON` and `-DGGML_RV_ZFH=ON`.*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*


ggml/src/ggml-cpu/CMakeLists.txt
```diff
@@ -458,29 +458,34 @@ function(ggml_add_cpu_backend_variant_impl tag_name)
                 ggml-cpu/spacemit/ime_kernels.h
             )
         endif()
-        if(NOT GGML_CPU_ALL_VARIANTS)
+if(NOT GGML_CPU_ALL_VARIANTS)
```
ggerganov:
```suggestion
            if(NOT GGML_CPU_ALL_VARIANTS)
```

xctan (approved):

PR #20897, opened 2026-03-23
vulkan: add shmem-staging matrix-vector multiplication kernel (>2.5x tg speedup on Intel Arc)

TheBlueMatt:
**Note that this code was written by Claude. While I did review and understand it at a high level, I am no vulkan expert. Still, a >2.5x performance gain is pretty substantial, so it seemed worth cleaning up and opening, which I'm doing here.** At least one other contributor also concluded it might be acceptable to open at https://github.com/ggml-org/llama.cpp/issues/18808#issuecomment-4104510623 but of course feel free to close if you prefer.

Add a new shared-memory staging variant of the quantized matrix-vector multiplication kernel (mul_mat_vecq_shmem.comp) that decouples the A-matrix decode step from the dot-product computation by staging decoded quant data through shared memory.

The existing mul_mat_vecq.comp kernel uses a fused approach: each thread reads its assigned A-matrix quant block directly from global memory via typed SSBO access (data_a[block_idx].qs[...]), decodes the quantized values, and immediately computes the dot product with the cached B-side data — all in one pass.

The new shmem-staging kernel splits this into two distinct phases:

  1. DECODE: Each thread reads raw bytes from the A-matrix via an untyped uint64 buffer (data_a_u64[]), manually extracts quant values and scale factors at byte granularity using helper functions (buf_read_i32, buf_read_f16, buf_read_u8, buf_read_u16), decodes them into unpacked int8 form, and writes the decoded values into a shared memory array (shmem_decoded[]). Scale factors are kept in per-thread register arrays.

  2. COMPUTE: Each thread reads back the decoded quant data from shared memory and computes dot products against the B-side data using integer dot product instructions (dotPacked4x8EXT).

This two-phase approach is a performance win on architectures like Intel Xe2 (BMG/LNL) where the typed SSBO access pattern of the original kernel hits pathological cases in the memory subsystem. By reading via raw uint64 loads and staging through shared memory, the kernel achieves sequential coalesced reads that better match these GPUs' memory controllers.

This cannot be implemented as a mode within the existing mul_mat_vecq.comp because the two kernels' control flow structures differ significantly: the original kernel iterates over individual blocks with a fused load-compute loop, while the shmem-staging kernel iterates over tiles of multiple blocks with explicit decode and compute phases separated by the shared memory staging point.

The shmem-staging kernel is enabled by default on Intel Xe2 GPUs and can be controlled via the GGML_VK_MMVQ_SHMEM_STAGING environment variable. The Intel Xe2 subgroup size override is also adjusted to not force minimum subgroup size on Xe2, since Xe2 benefits from using its native subgroup size with this kernel.

On an Intel Arc Pro B60, this provides a more than 2.5x speedup in token generation:
 * gpt-oss 20B MXFP4 MoE tg128: 25.66 ± 0.04 -> 74.06 ± 0.16

On two Intel Arc Pro B60s (with very slow PCIe 3.0 interconnects):
 * qwen35 27B Q8_0 tg128: 3.45 ± 0.03 -> 6.41 ± 0.27

There's an earlier attempt at this at #20869 which was a cleaner change by modifying the existing kernel, but sadly this version is much harder to fit into the existing kernel and is quite a bit faster still than #20869.


0cc4m:
I guess Intel GPUs suffer a lot more from smaller global loads, but what I'm not sure about is whether that's an issue on the driver or on the hardware side. AMD and Nvidia also prefer large loads, but they handle scattered access much more gracefully.

I've seen some attempts at loading untyped in the past, but so far we've avoided it and I would prefer to keep it that way. It's a lot harder to read and maintain.

Thank you for the information. In this state it's way too much code, but maybe the concept can be refined into something usable.

TheBlueMatt:
Might be worth taking a look at #20869 then? Its much more straightforward code and reuses most of the existing logic in the kernel while still getting a material uplift. Its not as performant as this kernel but much more maintainable, even if it is also untyped loads.

TheBlueMatt:
If #20869 is acceptable its maybe possible to munge the loading from that into this setup, doing the decode during the load phase and before the actual work phase, would be more complicated but maybe possible to clean this up and unify the kernels again.

TheBlueMatt:
Should also be relatively straightforward to make this at least somewhat more understandable. Claude, of course, just spat out numbers for offsets when reading but `offset_of` and the like should be usable to make it at least somewhat more readable. If you'd be open to that also happy to take a crack at it.

TheBlueMatt:
Actually, I went ahead and rewrote this. Its now three separate commits, with the coalesced loading only taking place in the first commit, but integrated into the existing utility functions in `mul_mat_vecq_funcs.glsl`. With some offset constants (why can't we have basic things like `offsetof`...) the code is ultimately only narrowly less readable than the existing struct-based loading. Then, the following commit adds the preloading-based kernel which avoids any of the untyped loading, just calling the existing util functions.

I left an option to run just the untyped loading kernel without the preload, I assume it might well be a performance win on other GPUs, but it probably makes sense to rip out before landing to avoid kernel option proliferation.

Let me know if this is more reasonable @0cc4m, or what else can be tweaked to get this to a state that you find maintainable/readable. There's probably some juice to squeeze out of the second commit on readability, but I already spent more time on this than I wanted so not sure I'll put more into that unless the project is interested in moving forward here.

ggml/src/ggml-vulkan/vulkan-shaders/mul_mat_vecq_funcs.glsl
```diff
@@ -4,38 +4,96 @@
...
+    else
+        raw = (data_a_u32[w] >> shift) | (data_a_u32[w + 1u] << (32u - shift));
+    return raw & 0xFFFFu;
+}
+#endif
+
 #if defined(DATA_A_Q4_0) || defined(DATA_A_Q5_0) || defined(DATA_A_Q8_0) || defined(DATA_A_IQ1_S) || defined(DATA_A_IQ2_XXS) || defined(DATA_A_IQ2_XS) || defined(DATA_A_IQ2_S) || defined(DATA_A_IQ3_XXS) || defined(DATA_A_IQ3_S) || defined(DATA_A_IQ4_XS) || defined(DATA_A_IQ4_NL)
 FLOAT_TYPE get_dm(uint ib) {
+#ifdef USE_COALESCED_LOADS
+    return FLOAT_TYPE(buf_read_f16(ib * A_BLOCK_SIZE + A_BLOCK_OFFSET_D));
```
0cc4m, line 49: How is this even different from the read below? This isn't "coalesced", it's just the same read going through integer conversions.

The only actual difference I can see is that you piece it together from multiple 32-bit loads instead of doing 16-bit or 8-bit loads when only that is needed. If that were necessary it's on the driver to spread out loads like that, not on us.

0cc4m (commented):
Thank you for the effort, but this isn't actually better, it just shuffled the code into different places. Ultimately the issue is that you (Claude...) duplicated a ton of code, accepting that would make maintenance much harder.

There is probably an acceptable way to do a shared memory path for MMV/MMVQ, but not through writing everything with Claude.

ggml/src/ggml-vulkan/vulkan-shaders/mul_mat_vecq_funcs.glsl
TheBlueMatt: Indeed for some of them it might well be that we can skip it.

TheBlueMatt:
> Thank you for the effort, but this isn't actually better, it just shuffled the code into different places

I mean come on, compared to the previous version this is a ton different? Using constants for offsets rather than Claude's hard-coded garbage means changes in the layout are easy to update, read logic is in a consistent place so updates aren't spread over multiple files, etc... It's your code you get to decide if you want to maintain changes but I don't think it's fair to say this is equally maintainable as Claude's original slop.

> Ultimately the issue is that you (Claude...) duplicated a ton of code, accepting that would make maintenance much harder.

I mean this version I spent time time by hand, fwiw, and if there's a way to move forward here I'd clean it up more (Claude is surprisingly bad at cleaning up existing changes, for whatever reason).

But to your point, ignoring the new shader entirely, the code duplication doesn't seem all that crazy? There's probably ways to remove some of it, as you note, happy to benchmark further if it makes sense, but you even noted this is likely to help on some other hardware as well. Might be worth benchmarking and even removing some of the old paths if they're consistently slower.

As for the new shader, I'd be interested in seeing that benchmarked on different hardware as well. Per https://chipsandcheese.com/p/intels-battlemage-architecture this isn't the only hardware where SHM is faster than L1, so it might pipeline better elsewhere. (edit: though maybe worth noting the new kernel is actually slower on a B60 if the `USE_COALESCED_LOADS` `#define` is removed, so dunno if that would work elsewhere without the coalesced loads...might be worth trying).

0cc4m:
> > Thank you for the effort, but this isn't actually better, it just shuffled the code into different places
> 
> I mean come on, compared to the previous version this is a ton different? Using constants for offsets rather than Claude's hard-coded garbage means changes in the layout are easy to update, read logic is in a consistent place so updates aren't spread over multiple files, etc... It's your code you get to decide if you want to maintain changes but I don't think it's fair to say this is equally maintainable as Claude's original slop.

You force-pushed the branch, I didn't have a direct comparison. What I mean is that the things that bother me didn't really change, not that the code isn't different. I didn't do a full review before or after the changes.

> As for the new shader, I'd be interested in seeing that benchmarked on different hardware as well. Per https://chipsandcheese.com/p/intels-battlemage-architecture this isn't the only hate where SHM is faster than L1, so it might pipeline better elsewhere.

Shared memory lives in the same space as other caches, ideally it shouldn't be needed since the driver can decide how to cache global loads efficiently and cover that part without manual shader work. That's the reason why we're not using shared memory in the matrix-vector multiplications, just directly accessing global data.

I'm pretty sure Intel should work the same way (and maybe this is why the Windows driver performs a lot better in token generation already). Are you testing on Linux?

TheBlueMatt:
> You force-pushed the branch, I didn't have a direct comparison. What I mean is that the things that bother me didn't really change, not that the code isn't different. I didn't do a full review before or after the changes.

Fair. Can you be a bit more specific on what bothers you - from the above I read that its mostly code volume that might make it annoying to change the access patterns in the future?

> Shared memory lives in the same space as other caches, ideally it shouldn't be needed since the driver can decide how to cache global loads efficiently and cover that part without manual shader work. That's the reason why we're not using shared memory in the matrix-vector multiplications, just directly accessing global data.
> I'm pretty sure Intel should work the same way (and maybe this is why the Windows driver performs a lot better in token generation already). Are you testing on Linux?

Yep, though on B60 that article notes that SLM is substantially faster (23.24 vs 15.13ns as far as I can tell, noting "Unlike cache hits, SLM accesses don’t need tag checks or address translation.") I thought I had read that that was similar on other hardware but now I can't find a quote for it so maybe I'm making that up. I was mostly using the chips and cheese article as a source cause its the only architecture analysis I could find.

Indeed, it may be the Windows driver splits the L1/SLM differently, though other issues I've seen on this repo don't seem to point to the Windows driver being anywhere near the performance of this code. At least the Chips and Cheese article (which is now a fairly old driver, and AFAICT on Windows) seems to think the driver rather statically splits L1/SLM, though its not clear to me that this workload would care all that much about more SLM/L1. (edit: At least mesa has code to do a dynamic split, https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/intel/common/intel_compute_slm.c?ref_type=heads so this shouldn't be true on Linux today, at least).

0cc4m:
Here are some tg128 results from a gpt-oss 20b mxfp4 test on my hardware. I ran the first PR test with `GGML_VK_MMVQ_SHMEM_STAGING=1` and the second with `GGML_VK_MMVQ_SHMEM_STAGING=1 GGML_VK_COALESCED_LOADS=1`, let me know if that wasn't correct.

| GPU                | t/s Master    | t/s SHMEM_STAGING | t/s COALESCED |
|--------------------|---------------|-------------------|---------------|
| Intel A770         | 41.21 ± 0.05  | 41.20 ± 0.02      | 28.64 ± 0.01  |
| AMD Radeon Pro VII | 140.34 ± 0.31 | 80.26 ± 0.07      | 80.17 ± 0.16  |
| AMD RX 9070 XT     | 178.42 ± 0.10 | 170.54 ± 0.12     | 170.41 ± 0.08 |
| Nvidia RTX 3090    | 182.59 ± 0.94 | 181.43 ± 0.92     | 181.23 ± 0.53 |

TheBlueMatt:
> and the second with GGML_VK_MMVQ_SHMEM_STAGING=1 GGML_VK_COALESCED_LOADS=1

that should just use the shmem kernel, I didn't do a full 4 kernel compilation, only three. To just test coalesced loading you'd only want the second env.

On the A770 it differed because the shmem kernel won't run if the workgroup size is too small (I need to fix it so that it would actually run, some of the math would result in an infinite loop, can do if this were to move forward, of course), so it fell back to the regular kernel, the coalesced kernel was just slower.

0cc4m:
Ah, I hadn't seen that coalesced also does something without shmem staging. Here are updated results, also with this diff to deal with the subgroup size issue and to fix the coalesced variable being set after shaders have been loaded already:
```diff
diff --git a/ggml/src/ggml-vulkan/ggml-vulkan.cpp b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
index 73cfa1a06..c2387a62f 100644
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -4079,7 +4079,7 @@ static void ggml_vk_load_shaders(vk_device& device) {

 #if defined(GGML_VULKAN_INTEGER_DOT_GLSLC_SUPPORT)
             if (device->integer_dot_product) {
-                const uint32_t subgroup_size_int = (device->vendor_id == VK_VENDOR_ID_INTEL && device->architecture != INTEL_XE2 && device->subgroup_size_control) ? device->subgroup_min_size : device->subgroup_size;
+                const uint32_t subgroup_size_int = (device->vendor_id == VK_VENDOR_ID_INTEL && device->subgroup_size_control) ? 16 : device->subgroup_size;
                 const uint32_t wg_size_subgroup_int = (w == DMMV_WG_SIZE_SUBGROUP) ? subgroup_size_int : (subgroup_size_int * 4);

                // For K-quants, the shmem variant needs a BLOCK_SIZE of at least 16
@@ -4168,7 +4168,7 @@ static void ggml_vk_load_shaders(vk_device& device) {

 #if defined(GGML_VULKAN_INTEGER_DOT_GLSLC_SUPPORT)
         if (device->integer_dot_product) {
-            const uint32_t subgroup_size_int = (device->vendor_id == VK_VENDOR_ID_INTEL && device->architecture != INTEL_XE2 && device->subgroup_size_control) ? device->subgroup_min_size : device->subgroup_size;
+            const uint32_t subgroup_size_int = (device->vendor_id == VK_VENDOR_ID_INTEL && device->subgroup_size_control) ? 16 : device->subgroup_size;
             const uint32_t wg_size_subgroup_int = (w == DMMV_WG_SIZE_SUBGROUP) ? subgroup_size_int : (subgroup_size_int * 4);

             // For K-quants, the shmem variant needs a BLOCK_SIZE of at least 16
@@ -5578,6 +5578,9 @@ static vk_device ggml_vk_get_device(size_t idx) {
         descriptor_set_layout_create_info.setPNext(&dslbfci);
         device->dsl = device->device.createDescriptorSetLayout(descriptor_set_layout_create_info);

+        const char * coalesced_loads_env = getenv("GGML_VK_COALESCED_LOADS");
+        device->coalesced_loads = coalesced_loads_env != nullptr && strcmp(coalesced_loads_env, "0") != 0;
+
         ggml_vk_load_shaders(device);

         // Only use transfer queue on AMD non-GCN, when the graphics queue is not enabled
@@ -5621,9 +5624,6 @@ static vk_device ggml_vk_get_device(size_t idx) {
             device->mmvq_mode = 1;
         }

-        const char * coalesced_loads_env = getenv("GGML_VK_COALESCED_LOADS");
-        device->coalesced_loads = coalesced_loads_env != nullptr && strcmp(coalesced_loads_env, "0") != 0;
-
         return device;
     }
```

| GPU                | t/s Master    | t/s SHMEM_STAGING | t/s COALESCED  | t/s BOTH      |
|--------------------|---------------|-------------------|----------------|---------------|
| Intel A770         | 41.21 ± 0.05  | 52.58 ± 0.06      | 35.68 ± 0.02   | 52.07 ± 1.45  |
| AMD Radeon Pro VII | 140.34 ± 0.31 | 80.32 ± 0.23      | 93.03 ± 0.41   | 80.29 ± 0.18  |
| AMD RX 9070 XT     | 178.42 ± 0.10 | 170.21 ± 0.32     | 171.30 ± 0.42  | 169.98 ± 0.46 |
| Nvidia RTX 3090    | 182.59 ± 0.94 | 176.53 ± 13.65    | 177.60 ± 10.67 | 180.60 ± 3.90 |

TheBlueMatt:
Mmm, fair enough, guess its only a win on Intel, and only huge on BMG/Xe2. I guess the BOTH/SHMEM_STAGING is just noise, cause the SHMEM_STAGING kernel sets the COALESCED read #define always at the top (you can comment it out and it'll still work, but given your numbers above I'm not sure its worth bothering to benchmark that).

0cc4m:
Oh, right. Here's the actual result:

| GPU                | t/s Master    | t/s SHMEM_STAGING | t/s COALESCED  | t/s BOTH      |
|--------------------|---------------|-------------------|----------------|---------------|
| Intel A770         | 41.21 ± 0.05  | 20.35 ± 0.13      | 35.68 ± 0.02   | 52.07 ± 1.45  |
| AMD Radeon Pro VII | 140.34 ± 0.31 | 130.80 ± 0.37     | 93.03 ± 0.41   | 80.29 ± 0.18  |
| AMD RX 9070 XT     | 178.42 ± 0.10 | 179.16 ± 0.47     | 171.30 ± 0.42  | 169.98 ± 0.46 |
| Nvidia RTX 3090    | 182.59 ± 0.94 | 175.13 ± 16.45    | 177.60 ± 10.67 | 180.60 ± 3.90 |


> Fair. Can you be a bit more specific on what bothers you - from the above I read that its mostly code volume that might make it annoying to change the access patterns in the future?

Basically you are duplicating the quant structs so that they can be loaded from arbitrary uint arrays in the absence of pointer casting. You are duplicating the quant repacking code to prepare for the integer dot instruction. You are duplicating all the mmvq block math code. Also a lot more mmvq shader variants that have to be managed.

Some duplication is unavoidable, but this is a lot. Also, I like that we have so far managed to avoid loading from uint arrays, it's rather ugly.

> Yep, though on B60 that article notes that SLM is substantially faster (23.24 vs 15.13ns as far as I can tell, noting "Unlike cache hits, SLM accesses don’t need tag checks or address translation.") 

Though this is latency, not throughput. GPUs are pipelining machines and basically always try to hide latency that way, so I'm not sure if these L1 vs SLM numbers have practical value.

> Indeed, it may be the Windows driver splits the L1/SLM differently, though other issues I've seen on this repo don't seem to point to the Windows driver being anywhere near the performance of this code. At least the Chips and Cheese article (which is now a fairly old driver, and AFAICT on Windows) seems to think the driver rather statically splits L1/SLM, though its not clear to me that this workload would care all that much about more SLM/L1.

It would be worth comparing, but I don't have any setup to do that.

TheBlueMatt:
> Basically you are duplicating the quant structs so that they can be loaded from arbitrary uint arrays in the absence of pointer casting. You are duplicating the quant repacking code to prepare for the integer dot instruction. You are duplicating all the mmvq block math code. 

Yea, definitely fair. Sadly each alternative path seems to be a very substantial individual win on B60, so hard to walk any of it back. There's probably more dancing that could be done with macros to unify the repacking code, but then we'd end up with somewhat messy macros and I'm not sure its any better (man would be nice if Vulkan gave us niceties like `offsetof` and pointer casting, even if there's alignment things devs have to deal with...though maybe that's just cause Intel's drivers suck). Also seems like there's room to simplify the mmvq block math across datatypes but, yea, I mean it exists.

Is there a level of duplication you'd be comfortable with here or some other path you see? It would be nice to get at least some of the performance wins here upstream, even if via a separate path. As you note its not clear how to accomplish any of these wins without code duplication, but I'm happy to take some cracks at alternative paths if you have any ideas.

> Also a lot more mmvq shader variants that have to be managed.

Given your results above it does seem like we'd have no reason to have a `COALESCED` variant, so it would "only" be 2x more.

> Some duplication is unavoidable, but this is a lot. Also, I like that we have so far managed to avoid loading from uint arrays, it's rather ugly.

FWIW I did go ahead and open https://gitlab.freedesktop.org/mesa/mesa/-/work_items?show=eyJpaWQiOiIxNTE2MiIsImZ1bGxfcGF0aCI6Im1lc2EvbWVzYSIsImlkIjoxNDg5ODB9 to see if there's any appetite from the anv folks to take a look. I imagine they won't but you never know.

0cc4m:
Thank you, opening an issue with ANV is a good idea. Let's try to get some information from them about this behaviour. A lot of my concerns come from the fact that I believe it is a workaround for a driver optimization problem.

I'll try to spend some time looking through the code to check if I can find a smaller set of changes to get to this behaviour. I do plan to replace my A770 with a B50 eventually as well, so that I can learn more about Battlemage. I'll get back to you about the duplication.

TheBlueMatt:
You may find https://github.com/intel/metrics-discovery/pull/41 useful if you spend more time looking into Linux perf of Intel hardware.

TheBlueMatt:
Claude was able to find a small mesa patch which provides the same improvement as the first commit (see freedeskop gitlab issue). We'd still have to include the second commit, but maybe that duplication is more tolerable since we'd be reusing the existing loading methods, assuming the mesa patch makes sense and isn't a huge regression elsewhere (based on your numbers above it maybe only makes sense for battlemage? would leave the A series out in the cold).

savvadesogle:
> * gpt-oss 20B MXFP4 MoE tg128: 25.66 ± 0.04 -> 74.06 ± 0.16

Excellent speed! 
So far, the maximum I've been able to get with the gpt-oss 20b is 72 t/s on the A770.

image

https://github.com/ggml-org/llama.cpp/issues/17628#issuecomment-4018920296

Issue #20677, opened 2026-03-17
Refactor: webui: migrate to the new /tools API


ngxson:
### Background Description

Ref: https://github.com/ggml-org/llama.cpp/issues/20673

The schema is to be defined, this issue is currently a placeholder

PR #20898, opened 2026-03-23
server: add built-in tools backend support

ngxson:
Prepare for https://github.com/ggml-org/llama.cpp/issues/20677

CC @allozaur for the frontend part

AI usage disclosure: tool implementation is written by LLM. But other parts: tool prototype, readme docs, help messages are written my me.

---

Note for UI:
- If `--tools` is not specified when launching server, `GET /tools` returns 404 --> web UI should let user know that built-in tools is not enabled, and can be enabled with `--tools all`
- Tools with `permissions.write == true` should require a permission dialog on the UI


ServeurpersoCom:
Can't wait to try it out :)

pwilkin:
The write tool seems pretty complex. I'd probably do a few variants and then run a few tests on a group of models to see how well they can use the edit tool, since from my experience regarding agentic tool use a good edit tool often makes or breaks the whole set.

ServeurpersoCom:
A "view" command could also display folders (two levels deep, excluding node_modules, for example) and show text files with added numbering. This is what Claude does, and it works best on small LLM. The command can be secured, and the LLM can also be informed if the content is binary or the context is too large. The most important thing in agentic is to force the LLM to write a rationale/reason field for using the tool. If the context is too repetitive, it breaks down. That makes the difference between reliable use on 500 tokens and on 10,000 more tokens!

Green-Sky:
A tool to fetch a remote resource by url might be useful too.

ServeurpersoCom:
We could do something crazy, but I digress. An internal tool that allows LLM to configure any external server we want :)

ngxson:
> The write tool seems pretty complex. I'd probably do a few variants and then run a few tests on a group of models to see how well they can use the edit tool, since from my experience regarding agentic tool use a good edit tool often makes or breaks the whole set.

@pwilkin  For context, I took my inspiration from set of tools from vscode. From what I observed from my initial test, models are not very good at calling edit tools because it must correctly translate what it what to do into JSON format. Hence I added 2 different tools for editing (plus one for simple write) mostly for experimenting.

I'd agree this is an area worth investigating, appreciate any efforts into improving this.

> A tool to fetch a remote resource by url might be useful too.

@Green-Sky For now, I leave this for the shell command tool for simplification

> A "view" command could also display folders (two levels deep, excluding node_modules, for example) and show text files with added numbering.

Hmm yeah this can be added as a flag to the existing file listing / file searching tool I think, something like `"display_level": number`

tools/server/server-tools.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "server-common.h"
+#include "server-http.h"
```
ggerganov, line 4: Including `server-http.h` here seems off. I think we have to refactor the HTTP layer in order to decouple it better from the rest of the server logic. For example, the `server-context.cpp` implementation should not reference HTTP components at all - HTTP is only used to transport the requests and responses to/back from the server.

ggerganov (approved):

tools/server/server-tools.h
ngxson: Yes we only include it here to use the HTTP requests/response interface. The implementation of the server HTTP itself is not exposed via this header.

ngxson: (Btw maybe we can refactor the interface `server_http_res/req` into a new header in the future)

ngxson:
@allozaur I guess can merge this if you give the 2nd approval

pwilkin (approved):
Let's approve and work on any enhancements in followup PRs.

tools/server/server-tools.h
ggerganov: > (Btw maybe we can refactor the interface `server_http_res/req` into a new header in the future)

Yes. The internal logic should have response/request handlers that are not associated with the HTTP protocol. The `server_http_res/req` would wrap those "native" handlers. This should reduce coupling.

isaac-mcfadyen:
> The write tool seems pretty complex

From my experience testing local models (specifically, I was comparing file-and-replace that Anthropic uses with OpenAI's V4A diffing, which is very poorly documented but somewhat outlined [here](https://developers.openai.com/cookbook/examples/gpt4-1_prompting_guide)), I've found that local models are *much* better at find-and-replace. I'd be interested to see how they handle this format though.

I also took a look at the read_file tool. Based on my testing on similar tools, a few things stood out from the implementation here:
- We use `\u2192` which is a rightwards arrow and is unicode, which some models might struggle with and/or it's a fairly rare thing in their vocab. I'd suggest we use a pipe here, like `<line number>|<content>`, which is how a lot of the existing coding harnesses do it
- The space after the arrow/pipe (e.g. the `\u2192 <content>`) *definitely* hurts the model. I found this to occur the most when the model is trying to edit a file, because it assumes the space is intentional/it mixes with any actual indentation in the file (e.g. in Python). The model then tries to edit/replace that space, which fails (because it's not actually in the file) and gets it stuck in a loop. I found this repeatedly occurring/reproducible until I removed that injected space.
- I've found that padding the line number to 5-6 chars with a blank space before the separator helps versus just adding the number. That way, on like `1` and line `1000`, the model has the same number of characters to "skip over" when looking for the actual content. So, for line 1, it would be something like `1     |<content>`
- Also might be worth testing start/end number versus offset/limit, a lot of harnesses (example: Codex) use offset/limit, and I have to imagine they've done some testing before they decided on that versus start/end.

PR #20908, opened 2026-03-23
RPC GRAPH_COMPUTE Arbitrary RCE patch

las7:
This fix prevents an existing RCE chain where deserialize_tensor() skips all validations when the incoming rpce_tensor.buffer field is set to null. Currently we only check if result==null, however this does not sufficiently cover the case for result->buffer. This prevents attackers from gaining arbitrary read and write.


rgerganov (approved):
I believe this check is sufficient because tensor views which are part of the compute graph are initialized by `ggml_backend_view_init()`:

https://github.com/ggml-org/llama.cpp/blob/bd6992180bd1e858129bf995d6b90539fb901dd6/ggml/src/ggml-backend.cpp#L1884-L1894

so it's not possible to have a valid tensor with `tensor->buffer == nullptr`

catap:
@ggerganov maybe make a new release of ggml with that fix?

PR #20912, opened 2026-03-23
common : replace wrap_for_generation with a prefix convenience function + gpt-oss fixes

aldehir:
## Overview

Remove `wrap_for_generation` function and replace with a `prefix()` parser that generates a literal up to a given delimiter.

Clean up gpt-oss and fix required tool calling. Injecting the generation prompt greatly simplifies the gpt-oss parser rules. This assumes `<|start|>assistant` is always a prefix of `generation_prompt`, which is a reasonable assumption. It will work even if users append extra tokens, i.e. `<|start|>assistant<|channel|>analysis<|message|>I am thinking of chocolate<|end|>`.

# Requirements

- AI usage disclosure: No

cc @pwilkin

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -112,8 +112,7 @@ common_peg_arena autoparser::build_parser(const generation_params & inputs) cons
         } else {
             parser = content.build_parser(ctx);
         }
-        parser = wrap_for_generation_prompt(p, parser, inputs, reasoning.start);
-        return parser;
+        return p.prefix(inputs.generation_prompt, reasoning.start) + parser;
     });
 }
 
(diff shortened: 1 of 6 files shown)


pwilkin:
Dunno, the creator of the PEG parser said that it's not a good idea to add more helper builders ;)

aldehir:
I meant inside the core peg parser so it's still useful for general parsing tasks. Inside the chat builder, anything goes!

Sorry for not making that distinction. I do prefer more composable patterns though, for example we're just matching a prefix here.

pwilkin:
Aight, might want to take this out of draft :)

aldehir:
Wanted to run through a few models before removing out of draft, ministral-3 thinks forever.

pwilkin:
BTW you might want to take a look at https://github.com/ggml-org/llama.cpp/pull/20844 - I scrapped the previous approach and did the regexp generator for the full pattern, but also expanded the tests to check whether the grammar is actually triggered.

pwilkin (approved):

PR #20913, opened 2026-03-23
jinja: fix a bug that empty expressions can trigger null binding in jinja parser (#20911)

zeph1912:
## Overview

This is to fix the bug found by UndefinedBehaviorSanitizer in jinja parser. The patch will reject empty computed member expressions before returning `slices[0]` from `parse_member_expression_arguments()`.  

# Requirements

- AI usage disclosure:

### DIFF
--- a/common/jinja/parser.cpp
+++ b/common/jinja/parser.cpp
@@ -540,7 +540,7 @@ private:
             return mk_stmt<slice_expression>(start_pos, std::move(start), std::move(stop), std::move(step));
         }
         if (slices.empty()) {
-            return mk_stmt<noop_statement>(start_pos);
+            return mk_stmt<blank_expression>(start_pos);
         }
         return std::move(slices[0]);
     }
--- a/common/jinja/parser.cpp
+++ b/common/jinja/parser.cpp
@@ -539,6 +539,9 @@ private:
             statement_ptr step = slices.size() > 2 ? std::move(slices[2]) : nullptr;
             return mk_stmt<slice_expression>(start_pos, std::move(start), std::move(stop), std::move(step));
         }
+        if (slices.empty()) {
+            throw parser_exception("Empty member expression arguments", source, peek().pos);
+        }
         return std::move(slices[0]);
     }
 
--- a/common/jinja/parser.cpp
+++ b/common/jinja/parser.cpp
@@ -540,7 +540,7 @@ private:
             return mk_stmt<slice_expression>(start_pos, std::move(start), std::move(stop), std::move(step));
         }
         if (slices.empty()) {
-            throw parser_exception("Empty member expression arguments", source, peek().pos);
+            return mk_stmt<noop_statement>(start_pos);
         }
         return std::move(slices[0]);
     }
--- a/common/jinja/runtime.cpp
+++ b/common/jinja/runtime.cpp
@@ -746,10 +746,6 @@ value member_expression::execute_impl(context & ctx) {
             return slice_func->invoke(args);
         } else {
             property = this->property->execute(ctx);
-            if (property->is_undefined()) {
-                JJ_DEBUG("%s", "Computed member property is undefined, returning undefined");
-                return mk_val<value_undefined>("object_property");
-            }
         }
     } else {
         // syntax: obj.prop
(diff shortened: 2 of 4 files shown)


common/jinja/parser.cpp
```diff
@@ -539,6 +539,9 @@ class parser {
             statement_ptr step = slices.size() > 2 ? std::move(slices[2]) : nullptr;
             return mk_stmt<slice_expression>(start_pos, std::move(start), std::move(stop), std::move(step));
         }
+        if (slices.empty()) {
+            throw parser_exception("Empty member expression arguments", source, peek().pos);
```
CISC:
```suggestion
            return mk_stmt<noop_statement>(start_pos);
```
In Python this would be an error, but Jinja2 suppresses this and returns `undefined` instead.

CISC (commented):
Please add a test so we can verify parity.

CISC:
@zeph1912 gentle ping

zeph1912:
sorry for the delay, addressed in the new commit

common/jinja/runtime.cpp
```diff
@@ -746,6 +746,10 @@ value member_expression::execute_impl(context & ctx) {
             return slice_func->invoke(args);
         } else {
             property = this->property->execute(ctx);
+            if (property->is_undefined()) {
+                JJ_DEBUG("%s", "Computed member property is undefined, returning undefined");
+                return mk_val<value_undefined>("object_property");
+            }
```
CISC: Short-circuiting here is not correct, rather we should handle `undefined` as a property further down, the following is what's happening (and is allowed in `jinja2`); `a[undefined]`, add that as a test too.

common/jinja/runtime.cpp
```diff
@@ -769,6 +769,10 @@ value member_expression::execute_impl(context & ctx) {
     }
 
     JJ_DEBUG("Member expression on object type %s, property type %s", object->type().c_str(), property->type().c_str());
+    if (property->is_undefined()) {
+        JJ_DEBUG("%s", "Member expression property is undefined, returning undefined");
+        return mk_val<value_undefined>("object_property");
+    }
     ensure_key_type_allowed(property);
 
     value val = mk_val<value_undefined>("object_property");
```
CISC:
```suggestion
    value val = mk_val<value_undefined>("object_property");

    if (property->is_undefined()) {
        JJ_DEBUG("%s", "Member expression property is undefined, returning undefined");
        return val;
    }

    ensure_key_type_allowed(property);
```

common/jinja/runtime.h
```diff
@@ -215,7 +215,7 @@ struct continue_statement : public statement {
 };
 
 // do nothing
-struct noop_statement : public statement {
+struct noop_statement : public expression {
```
CISC:
```suggestion
struct noop_statement : public statement {
```

CISC (approved):

common/jinja/runtime.h
CISC: Forgot this one? I see no reason to promote this to an `expression`.

zeph1912: Changing it back to statement seems to break the current `a[]` handling.

`parse_member_expression_arguments()` returns `mk_stmt<noop_statement>(...)` for the empty case, and `member_expression` expects its property node to be an `expression`. 

This expectation is not new, `member_expression` checks that its `property` node passes `chk_type<expression>(...)` in c15395f7 ([`runtime.h`](https://github.com/ngxson/llama.cpp/blob/c15395f73c42805d9609a55749a1d4b5b2251379/common/jinja/runtime.h#L273)), but the old code path never tried to put such node into member_expression.property. 

- normal case a[i] returns parse_expression()
- slice case like a[1:2] returns slice_expression
- empty `a[]` wasn't producing a valid AST node. My PR changed the empty case. Instead of crashing/going UB, it now return a node for a[] with property = noop_statement. That's why I did `struct noop_statement : public expression` to satisfy the existing expression check.

CISC: Ah, I see, then perhaps for clarity we should add an `undefined_expression` instead?

CISC: Or rather, `blank_expression`.

CISC (approved):

ngxson (approved):

Issue #18865, opened 2026-01-15
Feature Request: Tune CUDA kernel selection logic for datacenter GPUs


JohannesGaessler:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

On NVIDIA datacenter GPUs:
 
* Check performance of the custom kernel in `mmq.cuh` vs. dequantization + cuBLAS.
* Check performance of the kernel in `fattn-mma-f16.cuh` vs. those in `fattn-vec.cuh` and `fattn-tile.cuh`.
* Adjust the logic in host code if an alternative path is significantly faster.
* For datacenter GPUs that have the same CC as gaming GPUs, query the frequency from the device properties and use it for the kernel selection logic.

### Motivation

I wrote the kernel selection logic using performance numbers for NVIDIA consumer GPUs like the RTX 4090. However, I found that for example the optimal FlashAttention kernel changes depending on whether or not I set a frequency limit for my RTX 4090: with the default settings the vector kernel is the fastest, but with a limited frequency the MMA kernel is faster. Since datacenter GPUs like an L40 tend to run at lower frequencies than consumer GPUs it stands to reason that the current kernel selection logic is suboptimal for them.

When it comes to architectures like Hopper or datacenter Blackwell there is also the complication that they have instructions that are not available on consumer GPUs. The code I wrote does not utilize those instructions so dequantization + cuBLAS may be a better choice in those cases.

Issue #17822, opened 2025-12-06
Eval bug: Qwen 3 Next CUDA poor performance


fuutott:
### Name and Version

PS D:\llama\latest> .\llama-cli.exe --version
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 2 CUDA devices:
  Device 0: NVIDIA RTX PRO 6000 Blackwell Workstation Edition, compute capability 12.0, VMM: yes
  Device 1: NVIDIA RTX A6000, compute capability 8.6, VMM: yes
load_backend: loaded CUDA backend from D:\llama\latest\ggml-cuda.dll
load_backend: loaded RPC backend from D:\llama\latest\ggml-rpc.dll
load_backend: loaded CPU backend from D:\llama\latest\ggml-cpu-sapphirerapids.dll
version: 7300 (2960eb297)
built with Clang 19.1.5 for Windows x86_64

### Operating systems

Windows

### GGML backends

CUDA

### Hardware

RTX PRO 6000 and RTX A6000

### Models

qwen next testing with 
unsloth\Qwen3-Next-80B-A3B-Instruct-GGUF\Qwen3-Next-80B-A3B-Instruct-UD-Q4_K_XL

### Problem description & steps to reproduce

This doesn't seem to affect other models but when inferencing qwen 3 next looks like some things are still run on cpu which results in significant performance loss.
Screenshot was taken during tg stage for qwen next bench
<img width="1574" height="1282" alt="Image" src="https://github.com/user-attachments/assets/a637226a-828d-4322-959e-21d289737d5c" />

### Relevant log output

```shell
PS D:\llama> .\benchq3nxt.bat

D:\llama>set CUDA_VISIBLE_DEVICES=0

D:\llama>d:/llama/latest/llama-bench.exe   -m d:\models\unsloth\Qwen3-Next-80B-A3B-Instruct-GGUF\Qwen3-Next-80B-A3B-Instruct-UD-Q4_K_XL.gguf   
-p 512   -n 512   -b 1024   -ub 512   -ngl 99  -mmp 0   -fa 1   -o md   -r 3   -d 0
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 1 CUDA devices:
[...]

PR #20920, opened 2026-03-23
docker: parameterize CUDA version in build matrix, remove cuda-new.Dockerfile

M1DNYT3:
## Overview

`ggml/src/ggml-cuda/CMakeLists.txt` gates Blackwell cubin compilation on `CUDAToolkit_VERSION >= 12.8`:
````
cmake
if (CUDAToolkit_VERSION VERSION_GREATER_EQUAL "12.8")
    list(APPEND CMAKE_CUDA_ARCHITECTURES 120a-real)
endif()
````
The official image ships with CUDA_VERSION=12.4.0 - gate never triggers, RTX 50xx and RTX PRO 6000 fall back to PTX JIT.

Validated across 4 architectures (100-record batch, 19 parallel slots per card): https://github.com/M1DNYT3/self-hosted-llm-inference/blob/main/iterations/10-blackwell-root-cause/summary.md

12.9.1 is the latest 12.x. Pascal (sm_60/62) and Maxwell (sm_52) are not deprecated until 13.0 - no supported GPU is dropped.

Test image: ``docker pull m1dnyt3/llama-cpp:server-cuda-12.9``

Full logs + dmon traces: https://github.com/M1DNYT3/self-hosted-llm-inference/tree/main/iterations/10-blackwell-root-cause/logs

Addresses #17822, #18865.

## Additional information

If you need additional data, I can run benchmarks on demand for any specific GPU. The harness is built around Vast.ai, so any card available on the platform can be tested - inference cost per benchmark run is negligible.

# Requirements

- AI usage disclosure: YES - Claude assisted with case study documentation and log analysis. Benchmarks, Docker build, source code investigation, and the one-line fix are my own work.


CISC:
Just use the `cuda-13` image and/or `cuda-new.Dockerfile` for Blackwell.

M1DNYT3:
> Just use the `cuda-13` image and/or `cuda-new.Dockerfile` for Blackwell.

CUDA 13 drops support for Pascal and Maxwell. While both are quite old, 12.9.1 is the last 12.x release, which enables Blackwell without dropping anything. No need to split the stack into two images either; mixed fleets with old and new hardware stay on a single image until fully transitioned to 13 and dropped 12.x.

M1DNYT3:
> > It did a whole lot more, most specifically updating the base image and tools so that arm64 builds could be done, the bump was a pre-requisite for that, not the other way around.
> 
> Makes sense on the ARM64 context. Given that 12.8 landed as a prerequisite for that work, bumping to 12.9.1 seems like an even smaller ask now that the baseline has already moved - it's the latest stable 12.x, picking up all fixes since 12.8, while avoiding 13.x, which still has open issues like #21255. If 12.9.1 becomes a prerequisite for something else tomorrow, the PR is already there. I have no objection to #21207 or its purpose - my question is simply why #20920 shouldn't be merged on its own merits.

@CISC Following up on the above - is my PR worth a second look on its own merits?

CISC (approved):
Perfect, thank you, can you run it on your fork to verify it builds (disable `s390x` to be able to complete)?

taronaeo (approved):

M1DNYT3:
@taronaeo @CISC Thank you for reviewing and helping with improvements! :)

PR #20923, opened 2026-03-24
llama_params_fit: set n_ctx to training context size when memory is abundant

ssam18:
When cparams->n_ctx == 0 and the model fits in device memory without requiring any changes, llama_params_fit_impl() returned early without ever writing to n_ctx, leaving it as 0 and forcing callers into corrective workarounds (see #20878). Per the documented contract in llama.h, the context size is modified if and only if it equals 0. Both early-return paths (single-device and multi-device) now assign hp_nct, the model's training context size, already computed at that point to cparams->n_ctx before returning, making behaviour consistent whether memory is scarce or abundant.

### DIFF
--- a/common/fit.cpp
+++ b/common/fit.cpp
@@ -215,6 +215,9 @@ static void common_params_fit_impl(
         if (sum_projected_free >= margins[0]) {
             LOG_TRC("%s: will leave %" PRId64 " >= %" PRId64 " MiB of system memory, no changes needed\n",
                 __func__, sum_projected_free/MiB, margins[0]/MiB);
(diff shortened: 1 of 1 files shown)


ssam18:
Please review this PR

ssam18:
@ggerganov - Please review this PR and let me know if you have any comments?

ssam18:
@tobi @deadprogram @richardkiss please find some time to review this PR.

JohannesGaessler:
I disagree with the linked issue, the function should keep the returned value at 0 rather than set an explicit value. At most the docstring should be modified.

PR #20928, opened 2026-03-24
Use docker in build-android.yml

shreyajn:
## Overview

The docker is the recommended way to build llama.cpp for hexagon so replacing the workflow to do exactly that. This is cleaner and allows us to maintain the SDK versions in the docker image itself. 

- AI usage disclosure: No

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -40,13 +40,9 @@ jobs:
     steps:
       - name: Clone
         uses: actions/checkout@v6
-
-      # Disabled due to size (400MB) and always 0 cache hits
-      # - name: ccache
-      #   uses: ggml-org/ccache-action@v1.2.16
-      #   with:
-      #     key: android-build
-      #     evict-old-files: 1d
+        with:
+          fetch-depth: 0
+          lfs: false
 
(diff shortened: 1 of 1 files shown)


max-krasnyansky (approved):
Very nice cleanup.

.github/workflows/build-android.yml
```diff
@@ -66,10 +62,15 @@ jobs:
...
-
+    container:
+      image: 'ghcr.io/snapdragon-toolchain/arm64-android:v0.3'
+    defaults:
+      run:
+        shell: bash
+    permissions:
+      id-token: write
+      packages: read
+      contents: read
```
CISC:
```suggestion
```
Is this really necessary?

max-krasnyansky: Not right now.

CISC (approved):
Any plans with the artifacts?

shreyajn:
> Any plans with the artifacts?

I want to add a step to use the artifacts for quick test workflow on-device. The device will be on cloud so I need to upload the build. I can add a retention period of 1 day if you'd like or add the upload step with the test workflow.

PR #20941, opened 2026-03-24
readme : clarify MODEL_ENDPOINT usage

angt:
## Overview

ModelScope is not an HF-compatible endpoint.

## Additional information

A dedicated `-ms` option could be added in the future.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ LLM inference in C/C++
 
 ## Hot topics
 
-- **HuggingFace cache migration: models downloaded with `-hf` are now stored in the standard HuggingFace cache directory, enabling sharing with other HF tools.**
+- **Hugging Face cache migration: models downloaded with `-hf` are now stored in the standard Hugging Face cache directory, enabling sharing with other HF tools.**
 - **[guide : using the new WebUI of llama.cpp](https://github.com/ggml-org/llama.cpp/discussions/16938)**
 - [guide : running gpt-oss with llama.cpp](https://github.com/ggml-org/llama.cpp/discussions/15396)
 - [[FEEDBACK] Better packaging for llama.cpp to support downstream consumers 🤗](https://github.com/ggml-org/llama.cpp/discussions/15313)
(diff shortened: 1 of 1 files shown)


ggerganov (approved):

yingdachen:
Not sure why this change is made to the readme without any code change? With no code hange, it would appear that the usage 

`MODEL_ENDPOINT=https://www.modelscope.cn/ ./llama-cli -hf ggml-org/gemma-3-1b-it-GGUF`

would still be valid. 

There was disccusion before the implementation was oringally merged and the MODEL_ENDPOINT approach was suggested, are the communities now favoring option like 

`./llama-cli` -ms ggml-org/gemma-3-1b-it-GGUF` 

instead?

@angt @ggerganov

angt:
Hi @yingdachen,

Indeed, we updated the `-hf` option recently to share the same cache as the `hf` tool and do the discovery of the model and the mmproj locally, not relying on the API. So it is not possible anymore to mix different source under the same `-hf` flag:

```
$ MODEL_ENDPOINT=https://www.modelscope.cn/ build/bin/llama-cli -hf ggml-org/gemma-3-1b-it-GGUF
get_repo_commit: JSON error: [json.exception.parse_error.101] parse error at line 2, column 1: syntax error while parsing value - invalid literal; last read: '<U+000A><'
error: failed to download model from Hugging Face
```

I did a PR especially to clarify that as I was not aware of this usage and so I proposed a `-ms` for ModelScope.

yingdachen:
> Hi @yingdachen,
> 
> Indeed, we updated the `-hf` option recently to share the same cache as the `hf` tool and do the discovery of the model and the mmproj locally, not relying on the API. So it is not possible anymore to mix different source under the same `-hf` flag:
> 
> ```
> $ MODEL_ENDPOINT=https://www.modelscope.cn/ build/bin/llama-cli -hf ggml-org/gemma-3-1b-it-GGUF
> get_repo_commit: JSON error: [json.exception.parse_error.101] parse error at line 2, column 1: syntax error while parsing value - invalid literal; last read: '<U+000A><'
> error: failed to download model from Hugging Face
> ```
> 
> I did a PR especially to clarify that as I was not aware of this usage and so I proposed a `-ms` for ModelScope.

thanks for the clarification @angt , we will take a look and submit a PR for the -ms option later.

PR #20947, opened 2026-03-24
Make compile time cache directory configurable

CSBVision:
## Overview

The cache directory is either defined by the `LLAMA_CACHE` environment variable or, if this does not exist, resolved to a user-dependent path. The latter introduces redundancies in multi-user environments (if multiple users use the same model, it is download and saved for each user) and, as the cache directory contains large files but is slightly hidden inside the home directory, easily takes much space. Therefore, introducing an optional configuration option gives more control.

## Additional information

The implementation checks for a `LLAMA_CACHE` definition at compile time, which can be defined using `/DLLAMA_CACHE="/path/to/custom/cache"`. If it is not defined, the existing logic is kept. In parcticular, the environment variable `LLAMA_CACHE` is used with higher priority if is defined at runtime. In this case, the environment is only read once (return value of `std::getenv("LLAMA_CACHE")` is cached). Everything else stays exactly the same.

Finally, we could also update the copied code inside `rpc-server.cpp`, still we focused on  `common.cpp` here.

# Requirements

- AI usage disclosure: No

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -848,9 +848,13 @@ std::string fs_get_cache_directory() {
         }
         return p;
     };
-    if (getenv("LLAMA_CACHE")) {
-        cache_directory = std::getenv("LLAMA_CACHE");
+    const char * cache_env = std::getenv("LLAMA_CACHE");
+    if (cache_env) {
+        cache_directory = std::string(cache_env);
     } else {
+#if defined(LLAMA_CACHE)
+        cache_directory = LLAMA_CACHE;
+#else
 #if defined(__linux__) || defined(__FreeBSD__) || defined(_AIX) || \
         defined(__OpenBSD__) || defined(__NetBSD__)
         if (std::getenv("XDG_CACHE_HOME")) {
@@ -881,6 +885,7 @@ std::string fs_get_cache_directory() {
 #endif
         cache_directory = ensure_trailing_slash(cache_directory);
         cache_directory += "llama.cpp";
+#endif /* defined(LLAMA_CACHE) */
     }
     return ensure_trailing_slash(cache_directory);
 }


angt:
sharing a cache between many users requires quite a lot of configuration for write permissions and so on. once that is done, it’s way easier to just set the user env to have `LLAMA_CACHE` by default.

CSBVision:
Well, this is highly platform-dependent. On Windows, you can easily setup shared directories. For example, installing everything to `C:\llama.cpp` and put the cache to `C:\llama.cpp\cache`, everything will work fine without any additional configuration. Similarly, one could simply define a relative directory `cache`, copy the binaries to a writable path like `C:\llama.cpp` and everything will work fine without any further configuration, too.

ngxson:
IMO it's strange to have this configurable at build time, I never seen any projects do this way.

Instead, you can simply configure system/user env var, or write powershell/cmd wrapper that set the env var before launching the library

ngxson:
Closing since it's more suitable for a wrapper, as said above

CSBVision:
A project supporting something similar is Tesseract; here, you can define an environment variable `TESSDATA_PREFIX` as well as define a compile-time default. But as you prefer, we are fine with patching the file ourselves. In case if someone finds this useful, too, patching with Powershell on Windows:

```
(Get-Content 'common\common.cpp' -Raw).Replace('cache_directory += "llama.cpp"', 'cache_directory = "path/to/your/cache"') | Set-Content 'common\common.cpp' -NoNewline
```

On Linux, `sed` can do the same thing.
Anyway, thanks for your time.

ngxson:
> A project supporting something similar is Tesseract; here, you can define an environment variable `TESSDATA_PREFIX` as well as define a compile-time default. 

That's the data dir, not cache dir, they are two different things

In anyway, it's still better to set cache dir via env var, patching source code is not guaranteed to be stable

CSBVision:
Yes agreed, it's not exactly the same thing but still something similar. You can define the path using an environment variable and if this is not set, you can provide a compile-time default to make things more flexible than using a _hard coded_ compile-time fallback. This situation is still analogous here: The compile-time fallback already exists, but the cache directory is hard coded to a location inside the home directory. The provided patch would only allow to _configure_ the fallback instead of forcing a hard coded one.

> In anyway, it's still better to set cache dir via env var, patching source code is not guaranteed to be stable

Agreed, that's why we proposed a PR. If you don't want to merge it, patching is still easier than maintaining our own fork... surely, on our own risk messing things up 👍

Issue #20225, opened 2026-03-08
Eval bug: Qwen 3.5 Full prompt re-processing on every conversation turn


williamtwomey:
### Name and Version

version: 8182 (05728db18)
built with GNU 13.3.0 for Linux x86_64

same issue on release llama-b8233 and 8149

### Operating systems

Linux

### GGML backends

HIP

### Hardware

Ryzen 9600X + Radeon AI PRO R9700

### Models

Qwen3.5 27B

## Summary

Qwen 3.5 27B (hybrid attention + Mamba2/SSM architecture) forces **full prompt re-processing on every conversation turn** due to the recurrent memory's `pos_min` always exceeding the SWA-based `pos_min_thold` threshold. This makes multi-turn conversations unusable at scale - a 15k-token conversation takes ~8 minutes per turn instead of seconds.

## Steps to Reproduce

1. Start the server:
```bash
llama-server --host 0.0.0.0 --port 8080 \
  -m Qwen3.5-27B.Q8_0.gguf \
  -np 1 --batch-size 4096 --ubatch-size 1024 \
  -c 131072 -ngl 99 --flash-attn on -v
```

2. Send a first message (works fine):
```bash
curl http://localhost:8080/v1/chat/completions -H "Content-Type: application/json" -d '{
  "model":"test",
  "messages":[
    {"role":"system","content":"You are a helpful assistant."},
    {"role":"user","content":"What is the capital of France?"}
  ],
  "max_tokens":20
}'
```

3. Send a follow-up message (triggers full re-processing):
```bash
curl http://localhost:8080/v1/chat/completions -H "Content-Type: application/json" -d '{
  "model":"test",
  "messages":[
    {"role":"system","content":"You are a helpful assistant."},
    {"role":"user","content":"What is the capital of France?"},
    {"role":"assistant","content":"Paris is the capital of France."},
    {"role":"user","content":"What about Germany?"}
  ],
  "max_tokens":20
}'
```

## Observed Behavior

Every subsequent conversation turn forces full prompt re-processing:

```
slot update_slots: id  0 | task 302 | n_past = 3, slot.prompt.tokens.size() = 357, seq_id = 0, pos_min = 356, n_swa = 1
slot update_slots: id  0 | task 302 | forcing full prompt re-processing due to lack of cache data (likely due to SWA or hybrid/recurrent memory, see https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055)
[...]

Issue #20153, opened 2026-03-06
Eval bug: forcing full prompt re-processing due to lack of cache data in Qwen3.5 27B


lostmsu:
### Name and Version

version: 8210 (2cd20b72e)
built with Clang 19.1.5 for Windows x86_64

### Operating systems

Windows

### GGML backends

CUDA

### Hardware

2x RTX 3090

### Models

bartowski/Qwen_Qwen3.5-27B-GGUF in Q8_0

### Problem description & steps to reproduce

I run

```
llama-server -hf bartowski/Qwen_Qwen3.5-27B-GGUF `
               -hff Qwen_Qwen3.5-27B-Q8_0.gguf `
               -ngl all -fit on --swa-full
```

Then interact with model in llama-server's WebUI.

### Relevant log output

<details>
<summary>Logs</summary>
>llama-server -hf bartowski/Qwen_Qwen3.5-27B-GGUF -hff Qwen_Qwen3.5-27B-Q8_0.gguf -ngl all -fit on --swa-full
ggml_cuda_init: found 2 CUDA devices:
  Device 0: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes
  Device 1: NVIDIA GeForce RTX 3090, compute capability 8.6, VMM: yes
load_backend: loaded CUDA backend from C:\Users\lost\AppData\Local\Programs\llama.cpp\ggml-cuda.dll
load_backend: loaded RPC backend from C:\Users\lost\AppData\Local\Programs\llama.cpp\ggml-rpc.dll
load_backend: loaded CPU backend from C:\Users\lost\AppData\Local\Programs\llama.cpp\ggml-cpu-haswell.dll
common_download_file_single_online: no previous model file found C:\Users\lost\AppData\Local\llama.cpp\bartowski_Qwen_Qwen3.5-27B-GGUF_preset.ini
common_download_file_single_online: HEAD failed, status: 404
no remote preset found, skipping
common_download_file_single_online: using cached file (same etag): C:\Users\lost\AppData\Local\llama.cpp\bartowski_Qwen_Qwen3.5-27B-GGUF_Qwen_Qwen3.5-27B-Q8_0.gguf
main: n_parallel is set to auto, using n_parallel = 4 and kv_unified = true
build: 8210 (2cd20b72e) with Clang 19.1.5 for Windows x86_64
system info: n_threads = 8, n_threads_batch = 8, total_threads = 16

[...]

PR #20955, opened 2026-03-24
server: recover from checkpoint on truncation failure for hybrid/recurrent models

European-tech:
## Problem

When re-using a cached prompt with `TAG_PROMPT_LOGITS`, the server tries to truncate the KV cache via `llama_memory_seq_rm(slot.id, p0, -1)`. On hybrid/recurrent models (Qwen3.5, Jamba, Falcon-H1), this call **always fails** because partial truncation is not supported for recurrent layers. The current fallback is `prompt_clear(true)`, which erases the **entire** cached state and forces a full prompt re-evaluation from scratch.

For long contexts this is devastating: a 35K-token prompt takes ~21 seconds to re-process instead of the sub-100ms it should take from cache.

## Root Cause

`llama_memory_seq_rm()` returns `false` for recurrent memory types (Mamba/SSM layers) because they store a compressed state that cannot be partially truncated. The server interprets this as "memory is corrupt" and clears everything — but for hybrid models the memory is perfectly fine, it just doesn't support partial removal.

## Fix

Instead of clearing all memory on truncation failure, the server now:

1. Searches backward through existing prompt checkpoints (already created during prompt processing for recurrent models)
2. Finds the nearest checkpoint **before** the truncation point
3. Restores the full recurrent state from that checkpoint via `llama_state_seq_set_data_ext()` with `LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY`
4. Re-evaluates only the delta tokens between the checkpoint and the current position

Falls back to the original `prompt_clear()` only if no suitable checkpoint exists.

### Additionally: checkpoint persistence across slot save/restore

The built-in `/slots` save/restore API persists KV+recurrent memory via `llama_state_seq_{save,load}_file`, but does **not** persist the checkpoint metadata. This PR adds:

- `slot_checkpoints_save/load()` helpers that write/read a companion `.checkpoints` file next to the slot state file
- `auto_save_slots()` / `auto_restore_slots()` methods called at server shutdown/startup for seamless model hot-swap in router mode (multi-model `--models-dir`)

## Measured Impact

Qwen3.5-27B, 35858 token context, RTX 3090 24GB:

| Scenario | Time | Tokens re-evaluated |
|----------|------|---------------------|
| Before (prompt_clear) | **21054 ms** | 35858 (all) |
| After (checkpoint recovery) | **77 ms** | ~200 (delta only) |
| **Speedup** | **273x** | |

## Related Issues

- #20225 — Qwen3 inference issues with KV cache
- #20153 — Hybrid model memory management

## Files Changed

- `tools/server/server-context.cpp` — checkpoint save/load helpers, truncation recovery logic, auto-save/restore
- `tools/server/server-context.h` — `auto_save_slots()` / `auto_restore_slots()` declarations
- `tools/server/server.cpp` — call auto-restore on startup, auto-save on shutdown

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -455,6 +455,127 @@ struct server_slot {
     }
 };
 
+//
+// checkpoint persistence helpers for hybrid/recurrent models
+//
+// Hybrid models (e.g. Qwen3.5, Jamba, Falcon-H1) use recurrent layers whose
+// state cannot be partially restored from the KV cache alone.  The server
+// creates "context checkpoints" during prompt processing that snapshot the
+// full recurrent state at regular intervals.  These checkpoints live in
+// server_prompt::checkpoints and are essential to avoid a full prompt
+// re-processing when the slot is reused.
+//
+// The built-in /slots save/restore API persists the raw KV+recurrent memory
+// via llama_state_seq_{save,load}_file, but does NOT persist the checkpoint
+// metadata.  The two helpers below fill that gap: they write/read a small
+// companion file (<filename>.checkpoints) next to the main slot save file.
+//
+// File format (binary, little-endian):
+//   uint32  magic    = 0x4C4C4350  ("LLCP")
+//   uint32  version  = 1
+//   uint32  n_checkpoints
+//   For each checkpoint:
+//     int32   pos_min
+//     int32   pos_max
+//     int64   n_tokens
+//     uint64  data_size
+//     uint8   data[data_size]
+//
+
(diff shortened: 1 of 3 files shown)


llukas:
This MR works for me on M4 Macbook. I checked Qwen-122B and Nemotron 3 Super.

Minor comment: slot auto-save/auto-restore could be an option not unconditional behavior.

dark-penguin:
This was closed because this mechanism is already implemented, right?

PR #20959, opened 2026-03-24
Add graceful exit for SYCL backend when you ask for --split-mode row

FCLC:
## Overview

As of now split-mode row isn't implemented, resulting in a core dump when trying to use it. 

handle that with grace and add a fixme for when row is implemented in SYCL (which I'm working on)

## Additional information

CI runs fine/no new warnings 

# Requirements

- AI usage disclosure:  yes, codex cli with gpt5.4 pro, mainly to to the tracing of which files were relevant before i wrote the patches


arthw (approved):
Thank for your contribution!

arthw:
@FCLC 
Are you working to implement the row split in SYCL?

FCLC:
@arthw 
Yeah, working on it now. 

have managed to get row with ngl2 working on a pair of b580s, but it's borked above that. 

hopping to land it in the next few weeks, ideally before saturday. 

Also seems like the A770 has other issues (which I think are caused by the lack of 64 bit atomics in the memory controller and having layers/buffers >=4GBs) but that's for another day/patch set. 

Machine is 1 Esperanto ETSOC1, 2 intel B580s and 1 A770 on an 8124p.

arthw:
Allocate memory more than 4GB has been resolved on both iGPU and dGPU(including Arc770).

ggml/src/ggml-sycl/ggml-sycl.cpp
```diff
@@ -671,9 +671,16 @@ static size_t ggml_backend_sycl_buffer_type_get_alignment(ggml_backend_buffer_ty
 }
 
 static size_t ggml_backend_sycl_buffer_type_get_max_size(ggml_backend_buffer_type_t buft) {
-    return dpct::get_current_device().get_max_mem_alloc_size();
+    ggml_backend_sycl_buffer_type_context * ctx = (ggml_backend_sycl_buffer_type_context *) buft->context;
+    size_t max_size = dpct::dev_mgr::instance().get_device(ctx->device).get_max_mem_alloc_size();
```
arthw, line 675: Save the max mem size in the global variable when initiate SYCL.
Then get the value from the global variable in this function.
refer to:
`info.devices[i].smpbo = prop.get_local_mem_size();`

For example:
`info.devices[i].max_mem_alloc_size = {xxxxx: 4G ? get_max_mem_alloc_size()};`

In this function:
`return ggml_sycl_info().devices[ctx->device].max_mem_alloc_size;`

src/llama-model.cpp
```diff
@@ -249,6 +249,9 @@ static buft_list_t make_gpu_buft_list(ggml_backend_dev_t dev, llama_split_mode s
     // add the device split buffer type if requested and available
     if (split_mode == LLAMA_SPLIT_MODE_ROW) {
         ggml_backend_reg_t reg = ggml_backend_dev_backend_reg(dev);
+        if (ggml_backend_reg_name(reg) == std::string("SYCL")) {
+            throw std::runtime_error("SYCL backend does not support LLAMA_SPLIT_MODE_ROW");
+        }
```
ggerganov, line 254: This should not be added - please remove it.

arthw: "SYCL" could be replaced by GGML_SYCL_NAME.

This code is similar with  `if (ggml_backend_reg_name(reg) == std::string("RPC")) {` in llama.cpp.
To avoid the code crash when split-mode=row which is not supported by SYCL backend.
The feature split-mode=row is WIP.
When it's ready, this code will be removed.

Another alternative is let it crash as legacy until the feature is ready.

arthw (commented):
Sorry, I missed to submit my comments and forgot this PR for a long time.
Could you check if this PR is useful now?

Thank you!

arthw:
@FCLC 
Would you like to handle the PR continuously or close it?

Thank you!

FCLC:
Closing as I'll be unable to work on for the next few weeks, will open new PR when time permits

PR #20970, opened 2026-03-25
common : inhibit lazy grammar sampler while reasoning is active

aldehir:
## Overview

Inhibit lazy grammar sampler while reasoning is active to avoid grammar constraining when a tool call tag (e.g. `<tool_call>`) is present in the reasoning output.

This PR exposes the reasoning budget sampler state to `common_sampler` where it is aware of both the reasoning budget and grammar samplers. When the reasoning sampler is active, and the grammar is lazy, tokens are not passed to the grammar to avoid triggering within the reasoning block.

For models that are supported, i.e. have a defined end thinking tag, the reasoning budget sampler is always added with a default of `INT_MAX`.

Test changed to run apply before accept, as it is done in practice. Adjusted offset to accommodate. Moved `force_pos++` in `accept()`, which should be what updates the sampler's state. The init function will handle the case when `state == COUNTING && budget <= 0`.

## Additional information

Alternative approach to #20844.

# Requirements

- AI usage disclosure: Used to explore a PoC (https://github.com/aldehir/llama.cpp/pull/9). Rewritten by hand, although I took a few things verbatim. Used to review and assist in addressing the discrepancy in the tests.

### DIFF
--- a/common/sampling.cpp
+++ b/common/sampling.cpp
@@ -415,13 +415,12 @@ static bool grammar_should_apply(struct common_sampler * gsmpl) {
     if (!gsmpl->rbudget) {
         return true;
     }
-    const auto state = common_reasoning_budget_get_state(gsmpl->rbudget);
-    if (state == REASONING_BUDGET_IDLE || state == REASONING_BUDGET_DONE) {
-        // always apply if reasoning budget sampler is not active
-        return true;
+    if (gsmpl->params.grammar_lazy) {
+        // if grammar is lazy, only apply when reaosning budget is not active
+        const auto state = common_reasoning_budget_get_state(gsmpl->rbudget);
+        return state == REASONING_BUDGET_IDLE || state == REASONING_BUDGET_DONE;
     }
-    // if reasoning budget sampler is active, only apply when not lazy
-    return !gsmpl->params.grammar_lazy;
+    return true;
 }
 
 void common_sampler_accept(struct common_sampler * gsmpl, llama_token token, bool accept_grammar) {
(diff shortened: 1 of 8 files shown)


pwilkin:
Server tests fail on this one.

aldehir:
Yes, weird.

pwilkin:
You should adapt the tests in test-chat like I did in the other PR to test the grammar matching, that might help diagnose more problems.

aldehir:
Sure. I found the bug, missed an early return.

ggerganov:
Need to prevent using backend sampling (`-bs`) in combination with rbudget:

```diff
diff --git a/common/sampling.cpp b/common/sampling.cpp
index 7b648cf8c..6a3cf9919 100644
--- a/common/sampling.cpp
+++ b/common/sampling.cpp
@@ -538,7 +538,8 @@ llama_token common_sampler_sample(struct common_sampler * gsmpl, struct llama_co
         if (id != LLAMA_TOKEN_NULL) {
             LOG_DBG("%s: Backend sampler selected token: '%d'. Will not run any CPU samplers\n", __func__, id);
 
-            GGML_ASSERT(!gsmpl->grmr && "using grammar in combination with backend sampling is not supported");
+            GGML_ASSERT(!gsmpl->grmr    && "using grammar in combination with backend sampling is not supported");
+            GGML_ASSERT(!gsmpl->rbudget && "using reasoning budget in combination with backend sampling is not supported");
 
             // TODO: simplify
             gsmpl->cur.resize(1);
```

ggerganov (approved):

aldehir:
Tests added, thanks @pwilkin and added guard for backend sampling.

common/sampling.cpp
```diff
@@ -398,18 +402,39 @@ void common_sampler_free(struct common_sampler * gsmpl) {
...
 
+static bool grammar_should_apply(struct common_sampler * gsmpl) {
+    if (!gsmpl->grmr) {
+        return false;
+    }
+    if (!gsmpl->rbudget) {
+        return true;
+    }
+    if (gsmpl->params.grammar_lazy) {
+        // if grammar is lazy, only apply when reaosning budget is not active
```
aldehir:
```suggestion
        // if grammar is lazy, only apply when reasoning budget is not active
```

pwilkin (approved):

pwilkin:
Supersedes #20844

PR #20974, opened 2026-03-25
fix-pointer-dangling

yikechayedan:


yikechayedan (commented):
In the JNI layer of the sample Android program, when calling processUserInput, the pointer of user_prompt is freed before being referenced, and if the memory is overwritten during this period, it will not be possible to correctly retrieve the input.

PR #20975, opened 2026-03-25
mtmd: Add DeepSeekOCR 2 Support

sfallah:
## Overview

This PR adds support for DeepSeek-OCR-2 [deepseek-ai/DeepSeek-OCR-2](https://huggingface.co/deepseek-ai/DeepSeek-OCR-2).

GGUF model files for testing this PR are available at [sabafallah/DeepSeek-OCR-2-GGUF](https://huggingface.co/sabafallah/DeepSeek-OCR-2-GGUF).

The implementation adds the DeepSeek-OCR-2 vision path that reuses SAM encoder (from DeepSeek-OCR v1), a new Qwen2-based vision encoder, and multi-tile dynamic-resolution image preprocessing.

The preprocessing includes multi-tile dynamic-resolution, with a 1024 global view and grid of tiles with 768 tile views.
The image resize/padding behavior is Pillow-based to match the original deepseek-ai implementation.

### Main changes

- added DeepSeek-OCR-2 model support in mtmd
- reused the SAM implementation from DeepSeek-OCR where possible
- added the Qwen2 vision encoder
- added multi-tile dynamic-resolution preprocessing
- implemented the DeepSeek-OCR-2 attention mask handling for the Qwen2 encoder
- updated `mtmd_encode` to handle multi-tile and global view token counts diff
- added converter support for DeepSeek-OCR-2 vision weights
- extended `tools/mtmd/tests/test-deepseek-ocr.py` to cover both DeepSeek-OCR and DeepSeek-OCR-2

### Implementation notes

- The SAM encoder is shared with DeepSeek-OCR through the existing `build_sam`.
- The Qwen2 attention mask is prepared CPU-side.
- The dynamic-resolution preprocessing follows the InternVL-style behavior used by the reference implementation.

### Testing

- Extended `tools/mtmd/tests/test-deepseek-ocr.py` to cover both v1 and v2.
- DeepSeek-OCR-2 result: CER 0.6944 / chrF 23.01
- DeepSeek-OCR-2 HF reference: CER 0.7761 / chrF 28.70
- Manually verified with `llama-mtmd-cli`.
- Manually verified with `llama-server`.

### Caveats

- The DeepSeek-OCR-2 test gate is intentionally loose (`cer_tol = 0.12`). The current test image is low quality, and the HF reference itself has a high CER on this sample.
- The DRY sampler is used as an approximation for the HF `no_repeat_ngram_size` behavior.
- For DeepSeek-OCR-2, `llama-server` currently requires `--chat-template deepseek-ocr --no-jinja`.  This is due to the template handling for DeepSeek-OCR-2.

- Running `llama-server`:
  ```bash
  build/bin/llama-server \
  -m gguf_models/deepseek-ai/deepseek-ocr-2-bf16.gguf \
  --mmproj gguf_models/deepseek-ai/mmproj-deepseek-ocr-2-bf16.gguf \
  --chat-template deepseek-ocr --no-jinja \
  --temp 0 \
  --flash-attn off \
  --no-warmup \
  -n 2048 \
  --dry-multiplier 0.8 --dry-base 1.75 --dry-allowed-length 2 \
  --dry-penalty-last-n -1 --dry-sequence-breaker none
  ```

## Requirements

- AI usage disclosure: YES - I used AI assistance for code review, debugging, implementation checks, and testing. I have reviewed the submitted changes and take responsibility for the full contents of this PR.

### DIFF
--- a/conversion/base.py
+++ b/conversion/base.py
@@ -1138,7 +1138,7 @@ class TextModel(ModelBase):
         # Skip multimodal tensors
         if name.startswith(("mlp", "vit.", "vpm.", "siglip2.", "conformer.", "merger.", "resampler.", "sound_encoder.", "sound_projection.", "speech_embeddings.")) \
                 or "visual." in name or "vision." in name or "audio." in name or "talker." in name \
-                or "vision_" in name or "audio_" in name or "sam_model" in name \
+                or "vision_" in name or "audio_" in name \
                 or "token2wav." in name or "code2wav." in name \
                 or "projector." in name or "pre_mm_projector_norm" in name \
                 or "image_newline" in name or "view_seperator" in name \
--- a/conversion/deepseek.py
+++ b/conversion/deepseek.py
@@ -16,10 +16,14 @@ from .qwen import QwenModel
 
 @ModelBase.register("DeepseekOCRForCausalLM")
 class DeepseekOCRVisionModel(MmprojModel):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.clip_projector_type = gguf.VisionProjectorType.DEEPSEEKOCR
+
     def set_gguf_parameters(self):
         super().set_gguf_parameters()
         hparams = self.hparams
-        self.gguf_writer.add_clip_projector_type(gguf.VisionProjectorType.DEEPSEEKOCR)
+        self.gguf_writer.add_clip_projector_type(self.clip_projector_type)
         # default values below are taken from HF tranformers code
         self.gguf_writer.add_vision_attention_layernorm_eps(hparams.get("layer_norm_eps", 1e-6))
         self.gguf_writer.add_vision_use_gelu(True)
@@ -49,20 +53,20 @@ class DeepseekOCRVisionModel(MmprojModel):
             raise ValueError("DeepseekOCR model requires 'vision_config' in the model configuration, but it was not found")
 
         vision_config['sam'] = vision_config['width']['sam_vit_b']
-        vision_config.update(vision_config['width']['clip-l-14-224'])
-        vision_config['hidden_size'] = vision_config['width']
-        vision_config['num_heads'] = vision_config['heads']
-        vision_config['intermediate_size'] = vision_config['heads'] * 4
+        if vision_config['width'].get('clip-l-14-224') is not None:
+            vision_config.update(vision_config['width']['clip-l-14-224'])
+        if isinstance(vision_config['width'], int):
+            vision_config['hidden_size'] = vision_config['width']
+        if vision_config.get('heads') is not None:
+            vision_config['num_heads'] = vision_config['heads']
+            vision_config['intermediate_size'] = vision_config['heads'] * 4
 
         return vision_config
 
     def tensor_force_quant(self, name, new_name, bid, n_dims):
-        if ".embeddings." in name or 'pos_embed' in name:
-            return gguf.GGMLQuantizationType.F32
-        if ".rel_pos_h" in name or '.rel_pos_w' in name:
-            return gguf.GGMLQuantizationType.F32
-        if ".neck." in name or ".net_" in name:
-            return gguf.GGMLQuantizationType.F32
+        for nq_name in ('.embeddings.', 'pos_embed', '.rel_pos_h', '.rel_pos_w', '.neck.', '.net_'):
+            if nq_name in name:
+                return gguf.GGMLQuantizationType.F32
         return super().tensor_force_quant(name, new_name, bid, n_dims)
 
     @classmethod
@@ -79,6 +83,33 @@ class DeepseekOCRVisionModel(MmprojModel):
             name += ".weight"
 
         return super().filter_tensors((name, gen))
+
+
+@ModelBase.register("DeepseekOCR2ForCausalLM")
+class DeepseekOCR2VisionModel(DeepseekOCRVisionModel):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.clip_projector_type = gguf.VisionProjectorType.DEEPSEEKOCR2
+
+    def set_gguf_parameters(self):
+        # the vision tower's qwen2 encoder is built from fixed defaults,
+        # see build_qwen2_decoder_as_encoder() in deepencoderv2.py
+        if self.hparams.get("patch_size") is None:
+            self.hparams["patch_size"] = 16
+        if self.hparams.get("intermediate_size") is None:
+            self.hparams["intermediate_size"] = 4864
+        if self.hparams.get("num_attention_heads") is None:
+            self.hparams["num_attention_heads"] = 14
+        super().set_gguf_parameters()
+        # qwen2 encoder is GQA: 14 Q heads, 2 KV heads
+        self.gguf_writer.add_vision_head_count_kv(2)
+
+    def get_vision_config(self) -> dict[str, Any]:
+        vision_config = super().get_vision_config()
+        vision_config['hidden_size'] = vision_config['width']['qwen2-0-5b']['dim']
+        if vision_config.get('layers') is None:
+            vision_config['layers'] = 24
+        return vision_config
 
 
 @ModelBase.register("DeepseekForCausalLM")
@@ -188,12 +219,20 @@ class DeepseekV2Model(TextModel):
         self.origin_hf_arch = hparams.get('architectures', [None])[0]
 
         # special handling for Deepseek OCR
-        if self.origin_hf_arch == "DeepseekOCRForCausalLM":
+        if self.origin_hf_arch in ("DeepseekOCRForCausalLM", "DeepseekOCR2ForCausalLM"):
             self.model_arch = gguf.MODEL_ARCH.DEEPSEEK2OCR
             self.gguf_writer.arch = gguf.MODEL_ARCH_NAMES[self.model_arch]
             self.gguf_writer.add_architecture()
             # default jinja template
             self.gguf_writer.add_chat_template("{% for m in messages %}{{m['content']}}{% endfor %}")
+
+    @classmethod
+    def filter_tensors(cls, item: tuple[str, Callable[[], Tensor]]) -> tuple[str, Callable[[], Tensor]] | None:
+        name, _ = item
+        # DeepSeek-OCR vision encoder (SAM + DeepSeek-OCR-2 qwen2 tower)
+        if "sam_model" in name or "qwen2_model" in name:
+            return None
+        return super().filter_tensors(item)
 
     def set_vocab(self):
         try:
--- a/conversion/deepseek.py
+++ b/conversion/deepseek.py
@@ -69,6 +69,11 @@ class DeepseekOCRVisionModel(MmprojModel):
                 return gguf.GGMLQuantizationType.F32
         return super().tensor_force_quant(name, new_name, bid, n_dims)
 
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        if name.endswith("view_seperator"):
+            data_torch = data_torch.unsqueeze(0)
+        yield from super().modify_tensors(data_torch, name, bid)
+
     @classmethod
     def filter_tensors(cls, item: tuple[str, Callable[[], Tensor]]) -> tuple[str, Callable[[], Tensor]] | None:
         name, gen = item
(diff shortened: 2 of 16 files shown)


tools/mtmd/models/deepseekocr2.cpp
ngxson, line 1: IIRC, the vision encoder is mostly the same between DS-OCR v1 and v2, it's better to find a way to reuse the code from v1 when the other PR is merged

Edit: even better, v2 vision encoder is the stripped down version of v1, so it should be trivial to merge these 2 into one file

ngxson:
you may also need to enable `mtmd_decode_use_non_causal` for this model, so that text is causal while image tokens are non-causal

ngxson:
hmm, please ignore what I said earlier, it seems like their paper is poorly written and the v2 is not easier than v1.

image

basically it's a 3-step pipeline: SAM encode the image into embeddings, Qwen take embedding in and generate yet another embeddings, then finally they are fed into the main deepseek model

~~if my understanding is correct, that means the prompt will now be processed by vision encoder, which make things a whole lot more complicated~~

ngxson:
small correction, seems like what they refers to "query" on the diagram is not the text prompts, but simply a set of pre-trained tokens (fixed size), so that won't be too complicated to implement

the only thing that need to be provided correctly to the cgraph is the attention mask that will correctly mask the image tokens as non-causal and the query as causal, it should not be complicated

and finally, because SAM is the same between 2 models it is recommended that you extract SAM as a function and inherit it in DS-OCR v2, example:

```cpp
struct clip_graph_deepseekocr : clip_graph {
    clip_graph_deepseekocr(clip_ctx * ctx, const clip_image_f32 & img) : clip_graph(ctx, img) {}
    ggml_cgraph * build() override;
    ggml_cgraph * build_sam(); // build the SAM model
};

struct clip_graph_deepseekocr2 : clip_graph_deepseekocr { // inherit
    clip_graph_deepseekocr2(clip_ctx * ctx, const clip_image_f32 & img) : clip_graph(ctx, img) {}
    ggml_cgraph * build() override; // can directly reuse build_sam() from base class, no need to duplicate the code
};
```

tools/mtmd/models/deepseekocr2.cpp
```diff
@@ -0,0 +1,126 @@
...
+        ggml_tensor * inpL = inp;
+
+        ggml_tensor * inp_pos = ggml_cast(ctx0, ggml_arange(ctx0, 0, seq_len, 1), GGML_TYPE_I32);
+
+        int  num_heads = 14;
+        auto head_dim  = n_embd / num_heads;
+        int  kv_heads  = 2;
+
+        // qwen2 layers
+        for (int il = 0; il < n_layer; ++il) {
```
ngxson: this loop should be able to be replaced by build_vit, it should already support attention mask

conversion/base.py
```diff
@@ -1114,7 +1114,7 @@ def filter_tensors(cls, item: tuple[str, Callable[[], Tensor]]) -> tuple[str, Ca
         # Skip multimodal tensors
         if name.startswith(("mlp", "vit.", "vpm.", "siglip2.", "conformer.", "merger.", "resampler.", "sound_encoder.", "sound_projection.", "speech_embeddings.")) \
                 or "visual." in name or "vision." in name or "audio." in name or "talker." in name \
-                or "vision_" in name or "audio_" in name or "sam_model" in name \
+                or "vision_" in name or "audio_" in name or "sam_model" in name or "qwen2_model" in name \
```
CISC: I'm not comfortable with this in `TextModel`, please move it to `DeepseekModel`.

tools/mtmd/models/deepseekocr2.cpp
sfallah: `build_vit` does not support GQA right now.
I prepared a small PR for this here: https://github.com/ggml-org/llama.cpp/pull/23782
The change is basically to use `n_head_kv` for the K/V reshape, and fall back to `n_head` when it is not set.

But to be honest, I am a bit conflicted about this. 
I am not comfortable changing `build_vit`.
Please give me some guidance here.

ngxson: if GQA is the only thing missing from build_vit, then it's ok to add. there are ViT models use GQA, so it make sense to support that

except from GQA, is anything else missing?

sfallah: no,
the tests in `mtmd/tests.sh `are running right now, if successful I will open https://github.com/ggml-org/llama.cpp/pull/23782

conversion/base.py
sfallah: Moved qwen2_model (and sam_model, while at it) out of TextModel.

tools/mtmd/models/deepseekocr2.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    ggml_tensor * cur;
+
+    cur = ggml_mul_mat(ctx0, model.mm_fc_w, qwen2_out);
+    cur = ggml_add(ctx0, cur, model.mm_fc_b);
+
+    const auto n_dim = cur->ne[0];
+
+    // separator row, only after the global view (256 tokens)
+    if (cur->ne[1] == 256) {
```
ngxson: this pattern is not very clean.

instead, it's better to add a new field to `clip_image_f32`, let's say `bool clip_image_f32::add_viewsep`

the preprocessor can set this field on the f32 image before returning to the caller, and here we check for `img->add_viewsep`

tools/mtmd/clip.cpp
```diff
@@ -3293,6 +3305,15 @@ int clip_n_output_tokens(const struct clip_ctx * ctx, struct clip_image_f32 * im
                 int oh = (img->ny / patch_size) / merge;
                 n_patches = (ow + 1) * oh + 2;
             } break;
+        case PROJECTOR_TYPE_DEEPSEEKOCR2:
+        {
+            // 1024 global view -> 256 query tokens + 1 view separator = 257;
+            // 768 local tile   -> 144 query tokens, no separator.
+            n_patches /= 16;
+            if (n_patches == 256) {
```
ngxson: same here, better to check for image->add_viewsep

conversion/deepseek.py
```diff
@@ -49,20 +53,20 @@ def get_vision_config(self) -> dict[str, Any]:
             raise ValueError("DeepseekOCR model requires 'vision_config' in the model configuration, but it was not found")
 
         vision_config['sam'] = vision_config['width']['sam_vit_b']
-        vision_config.update(vision_config['width']['clip-l-14-224'])
-        vision_config['hidden_size'] = vision_config['width']
-        vision_config['num_heads'] = vision_config['heads']
-        vision_config['intermediate_size'] = vision_config['heads'] * 4
+        if vision_config['width'].get('clip-l-14-224') is not None:
+            vision_config.update(vision_config['width']['clip-l-14-224'])
+        if isinstance(vision_config['width'], int):
```
CISC, line 58: This will never be true because you will hit an `AttributeError` above.

Edit: Or is it that `clip-l-14-224` contains `width`?

sfallah: yes, 'clip-l-14-224' has also a 'width'

tools/mtmd/models/deepseekocr2.cpp
```diff
@@ -0,0 +1,84 @@
...
+        inp = ggml_cont(ctx0, ggml_permute(ctx0, inp, 1, 0, 2, 3));
+
+        auto num_image_tokens = inp->ne[1]; // H*W
+
+        // query based on numbers of image tokens (in SAM output)
+        // 16x16 -> query_1024 (1024x1024 images)
+        // 12x12 -> query_768 (768x768 images)
+
+        ggml_tensor * query_embed = model.resample_query_1024;
+        int           num_queries = 256;
```
ngxson, line 28: maybe have a GGML_ASSERT to make sure the `num_image_tokens` can be either 144 or 256

tools/mtmd/models/deepseekocr2.cpp
```diff
@@ -0,0 +1,84 @@
...
+    ggml_tensor * cur;
+
+    cur = ggml_mul_mat(ctx0, model.mm_fc_w, qwen2_out);
+    cur = ggml_add(ctx0, cur, model.mm_fc_b);
+
+    const auto n_dim = cur->ne[0];
+
+    // separator row, only after the global view
+    if (img.add_viewsep) {
+        ggml_tensor * vs = ggml_reshape_2d(ctx0, model.view_seperator, n_dim, 1); // (n_dim, 1)
```
ngxson: this reshape should be done at model conversion

tools/mtmd/models/deepseekocr2.cpp
```diff
@@ -0,0 +1,84 @@
...
+    ggml_tensor * inp_raw = build_inp_raw();
+
+    ggml_tensor * sam_out = build_sam(inp_raw);
+
+    ggml_tensor * qwen2_out;
+    // Building Qwen2 encoder
+    {
+        ggml_tensor * inp;
+
+        inp = ggml_cpy(ctx0, sam_out, ggml_dup_tensor(ctx0, sam_out));
```
ngxson: I don't quite get this, why we need to copy the tensor?

ngxson: same question for `clip_graph_deepseekocr`

tools/mtmd/models/deepseekocr2.cpp
sfallah: Added

tools/mtmd/models/deepseekocr2.cpp
sfallah: done

tools/mtmd/models/deepseekocr2.cpp
sfallah: These were leftovers from debugging.
I was dumping sam_out and wanted to isolate it before possible mutation effects, to be on the safe side.
Redundant ggml_cpy ops for sam_out and others are removed now.

tools/mtmd/models/deepseekocr2.cpp
```diff
@@ -0,0 +1,81 @@
...
+    ggml_tensor * qwen2_out;
+    // Building Qwen2 encoder
+    {
+        ggml_tensor * inp;
+
+        inp = ggml_reshape_2d(ctx0, sam_out, sam_out->ne[0] * sam_out->ne[1], sam_out->ne[2]); // H*W, C
+        inp = ggml_cont(ctx0, ggml_permute(ctx0, inp, 1, 0, 2, 3));
+
+        auto num_image_tokens = inp->ne[1]; // H*W
+        GGML_ASSERT(num_image_tokens == 144 ^ num_image_tokens == 256);
```
ngxson:
```suggestion
        GGML_ASSERT(num_image_tokens == 144 || num_image_tokens == 256);
```

ngxson (approved):

tools/mtmd/models/deepseekocr2.cpp
sfallah: It practically makes no difference if it is XOR or OR.
XOR is logically more accurate.
Because it is either a tile image or a global view image.
But we can go with `||`.

ngxson: I don't see how XOR is more accurate. are you saying to me that `num_image_tokens` can be both 144 AND 256 at the same time, how?

And even that, developers never write this in practice because it's dangerous. The whole code base of llama.cpp never have such pattern.

For example, say a=1 and b=2, a XOR b = 3 but !!a XOR !!b == false. Simply write using && || is much safer and easier to understand. So think twice before adding something new to the coding convention.

sfallah: yes but we didn't had `a ^ b`
we have `a==1 ^ b==2`
anyway I see your point
my first commit that i than squashed/force-pushed was exactly what you suggested.

CISC (approved):

ngxson:
@sfallah can you check locally if `test-llama-archs` passes? the CI fails on that test but I'm not sure if it has something to do with this PR, since the test only runs the text model

sfallah:
> @sfallah can you check locally if `test-llama-archs` passes? the CI fails on that test but I'm not sure if it has something to do with this PR, since the test only runs the text model

@ngxson

All tests pass locally.
As far as I can tell, this test is also failing on `master`.
`DSOCR-2` is currently excluded from `test-llama-archs`. In fact, I have already prepared a small follow-up PR that adds DeepSeek-OCR, since you have already approved this PR.

PR #20978, opened 2026-03-25
llama-model-loader: print warning when using overrides with mmap

am17an:
## Overview

When using `mmap` which is enabled by default, tensor overrides don't get pinned memory. This PR skips the downgrade for overriden tensors. 

## Additional information

Before 

| model                          |       size |     params | backend    | ngl | n_ubatch | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | ---: | --------------: | -------------------: |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      128 |    0 |          pp2048 |        543.68 ± 4.84 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      256 |    0 |          pp2048 |       973.01 ± 10.10 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      512 |    0 |          pp2048 |      1700.75 ± 22.03 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     1024 |    0 |          pp2048 |      2827.85 ± 18.51 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |    0 |          pp2048 |      4097.78 ± 75.23 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      128 |    1 |          pp2048 |        298.28 ± 4.26 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      256 |    1 |          pp2048 |        546.08 ± 4.33 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      512 |    1 |          pp2048 |        979.02 ± 4.78 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     1024 |    1 |          pp2048 |      1708.17 ± 27.07 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |    1 |          pp2048 |      2756.34 ± 36.39 |

After

| model                          |       size |     params | backend    | ngl | n_ubatch | mmap |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -------: | ---: | --------------: | -------------------: |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      128 |    0 |          pp2048 |        539.51 ± 7.29 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      256 |    0 |          pp2048 |       977.11 ± 10.41 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      512 |    0 |          pp2048 |      1713.08 ± 16.35 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     1024 |    0 |          pp2048 |      2838.30 ± 23.96 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |    0 |          pp2048 |      4123.53 ± 36.17 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      128 |    1 |          pp2048 |        548.10 ± 1.81 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      256 |    1 |          pp2048 |       972.36 ± 26.46 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |      512 |    1 |          pp2048 |      1737.46 ± 10.71 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     1024 |    1 |          pp2048 |      2814.28 ± 29.08 |
| gpt-oss 20B MXFP4 MoE          |  11.27 GiB |    20.91 B | CUDA       |  99 |     2048 |    1 |          pp2048 |      4114.94 ± 54.93 |

# Requirements

- AI usage disclosure: YES, asked it to implement the fix after identifying the reason

### DIFF
--- a/src/llama-model-loader.cpp
+++ b/src/llama-model-loader.cpp
@@ -1117,7 +1117,6 @@ struct ggml_tensor * llama_model_loader::create_tensor(
         }
 
         ggml_backend_buffer_type_t buft = nullptr;
-        bool buft_from_override = false;
 
         // check overrides
         if (tensor_buft_overrides) {
@@ -1128,10 +1127,15 @@ struct ggml_tensor * llama_model_loader::create_tensor(
                     if (overrides->buft == ggml_backend_cpu_buffer_type()) {
                         // when overriding to a CPU buffer, consider the extra buffer types
                         buft = select_weight_buft(hparams, t_meta, op, buft_list_cpu);
+                        if (use_mmap) {
+                            static std::once_flag once;
+                            std::call_once(once, [] {
+                                LLAMA_LOG_WARN("llama_model_loader: tensor overrides to CPU are used with mmap enabled - consider using --no-mmap for better performance\n");
+                            });
+                        }
                     } else {
                         buft = overrides->buft;
                     }
-                    buft_from_override = true;
 
                     LLAMA_LOG_DEBUG("tensor %s (%zu MiB %s) buffer type overridden to %s\n",
                             tensor_name.c_str(),
@@ -1150,9 +1154,8 @@ struct ggml_tensor * llama_model_loader::create_tensor(
         }
 
         // avoid using a host buffer when using mmap
-        // but keep host buffers for overridden tensors - they need the host buffer for pinned memory
         auto * buft_dev = ggml_backend_buft_get_device(buft);
-        if (use_mmap && !buft_from_override && buft_dev && buft == ggml_backend_dev_host_buffer_type(buft_dev)) {
+        if (use_mmap && buft_dev && buft == ggml_backend_dev_host_buffer_type(buft_dev)) {
             auto * cpu_dev = ggml_backend_dev_by_type(GGML_BACKEND_DEVICE_TYPE_CPU);
             if (!cpu_dev) {
                 throw std::runtime_error("no CPU backend found");
--- a/src/llama-model-loader.cpp
+++ b/src/llama-model-loader.cpp
@@ -1117,6 +1117,7 @@ struct ggml_tensor * llama_model_loader::create_tensor(
         }
 
         ggml_backend_buffer_type_t buft = nullptr;
+        bool buft_from_override = false;
 
         // check overrides
         if (tensor_buft_overrides) {
@@ -1130,6 +1131,7 @@ struct ggml_tensor * llama_model_loader::create_tensor(
                     } else {
                         buft = overrides->buft;
                     }
+                    buft_from_override = true;
 
                     LLAMA_LOG_DEBUG("tensor %s (%zu MiB %s) buffer type overridden to %s\n",
                             tensor_name.c_str(),
@@ -1148,8 +1150,9 @@ struct ggml_tensor * llama_model_loader::create_tensor(
         }
 
         // avoid using a host buffer when using mmap
+        // but keep host buffers for overridden tensors - they need the host buffer for pinned memory
         auto * buft_dev = ggml_backend_buft_get_device(buft);
-        if (use_mmap && buft_dev && buft == ggml_backend_dev_host_buffer_type(buft_dev)) {
+        if (use_mmap && !buft_from_override && buft_dev && buft == ggml_backend_dev_host_buffer_type(buft_dev)) {
             auto * cpu_dev = ggml_backend_dev_by_type(GGML_BACKEND_DEVICE_TYPE_CPU);
             if (!cpu_dev) {
                 throw std::runtime_error("no CPU backend found");


am17an:
@JohannesGaessler not sure if this would interfere with the functionality you added in #19802

JohannesGaessler (commented):
Do I understand you correctly that overrides are set as "CPU" when they should be set as "CUDA_Host", which is the pinned memory? I'm not sure this fix is the correct one though. In `ggml-cuda.cu` we have a function `ggml_backend_cuda_register_host_buffer` which, when applied to a memory mapping, should pin the memory and fix this issue. However, the function seems kind of unfinished and is to my understanding currently unused.

am17an:
So currently it maps correctly to CUDA host buffer when no mmap is used. However, the current code explicitly changes it to be not to a host buffer when mmap is used. I am not sure if this has always been the case, a quick git blame showed changes to this part of the code in the e2e tests PR.

am17an:
For the record, reverting #19802 also has the same behavior of `--mmap` being slower. Let me do a git bisect

am17an:
It looks this has always been there? I went pretty far back (till b6790) and I couldn't get the `--no-mmap` performance.

JohannesGaessler:
So generally speaking I think the issue with using "regular" CPU buffers or pinning the memory has to do with what happens when the OS runs our of system memory. If the weights are memory-mapped it can simply evict the memory pages, but I don't think it can with `--no-mmap` or if we pin the memory for CUDA. So the OS would instead have to kill processes or become unresponsive.

am17an:
I think if system memory allows it, it should use pinned memory by default since it is such a drastic change in PP performance. I don't think most people are aware (at least I wasn't) that host buffers are disabled when mmap is used.

0cc4m:
It is known. For partial offloads, disabling mmap is recommended. It might be worth considering a smarter mmap default selection.

JohannesGaessler:
We could maybe consider smarter behavior for `--mlock`: if `--mmap` and `--mlock`, then pin the memory. Maybe, if memory use is <= 80% of total memory, enable `--mlock` by default - I think this risks being very unintuitive though.

am17an:
Maybe we can just add a warning to not use mmap when partial offload is used? I don't think this is well documented anywhere, and llama bench results in the wild are all with mmap enabled (which is the default)

JohannesGaessler:
I think in recent months we have also gone back and forth regarding what the default for memory mapping was. @ggerganov do you know why this was?

0cc4m:
We briefly defaulted to direct-io loading, which performs better for large models on modern NVMe setups, but this caused a myriad of compatibility issues, so the default was reverted back to mmap.

ggerganov:
>  llama bench results in the wild are all with mmap enabled (which is the default)

~Atm, I think `llama-bench` sets both mmap and dio to off by default.~ Nvm, mmap is on.

ggerganov:
> It is known. For partial offloads, disabling mmap is recommended.

From @am17an's results, it seems that even for full offloads, there is a significant difference.

am17an:
@ggerganov those results are with `-ncmoe 999`, I think need to add it as a separate field to print

am17an:
I think because we have `llama_fit_params`, it makes sense for this to work out of the box. Because fit params could offload some layers which will end up not being pinned, even if they practically could.

am17an:
Changed this to a warning message to suggest `--no-mmap` when using cpu overrides

JohannesGaessler (approved):
Probably there isn't a way to do this that's 100% optimal in all cases, the warning will definitely be an improvement.

am17an:
can someone else also approve this?

0cc4m (approved):

PR #20980, opened 2026-03-25
ci : limit requirements versions

CISC:
## Overview

Limit Python package versions installed during CI runs.

## Additional information

How hard-line should we be? We have a few packages where we set upper/lower bounds on the version, and most of the packages use `~=`, which basically means any compatible version, changing this would likely be dependency hell and break all sorts of things.

# Requirements

- AI usage disclosure: nope

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -54,4 +54,3 @@ jobs:
           python3 -m venv .venv
           source .venv/bin/activate
           pip install -r requirements/requirements-all.txt -r tools/server/tests/requirements.txt
-          pip install flake8 pyright pre-commit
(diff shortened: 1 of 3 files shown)


ggerganov:
> How hard-line should we be?

Not really sure. I was thinking that for every package there will be a version that is good enough for our needs and we can restrict to less or equal to that version?

CISC:
> > How hard-line should we be?
> 
> Not really sure. I was thinking that for every package there will be a version that is good enough for our needs and we can restrict to less or equal to that version?

In theory, yes, however problems arise when another package has a different definition of that that does not directly overlap (because some new patch version came out f.ex.), then suddenly everything fails.

ggerganov:
Yes, I guess we do want to get patches though.

Ok let's proceed with those for now. Good to merge?

CISC:
> Yes, I guess we do want to get patches though.

It's a fine line, patches can be evil too, damned if you do, damned if you don't. :)

> Ok let's proceed with those for now. Good to merge?

👍

PR #20983, opened 2026-03-25
ggml : skip null devices in backend registry

changgeunJi:
## Overview

This PR fixes a case where HTP session initialization failure can cause `ggml_backend_reg_dev_get()` to return `nullptr`, and the common registry would then store that invalid value in the global device list.

This is best fixed in the common registry because the global device list is consumed by common enumeration and lookup paths, not only by Hexagon-specific code. This includes `ggml-backend-reg.cpp` and `llama-model.cpp`, with similar iteration patterns also appearing in args, tools, and tests.

The change is intentionally narrow. `register_device()` now skips null devices before logging or appending them to the global registry list. There are no public API signature changes.

### Problem

When HTP session initialization fails, `ggml_backend_reg_dev_get()` can return `nullptr`. The common registry currently does not guard against that case before publishing devices into the global list.

As a result, an invalid device handle can be exposed through the global registry and affect later device enumeration or lookup paths.

### Why fix this in common code

The issue is not limited to Hexagon-local handling. The global device list is used by common code paths, including `ggml-backend-reg.cpp` and `llama-model.cpp`, and similar iteration logic also exists in args, tools, and tests.

Because the invalid value is being published through the shared registry layer, the most appropriate fix is to prevent null devices from entering the global list in the common registration path.

### Change

`register_device()` now skips null devices before logging or appending them to the global registry list.

This does not change backend initialization behavior and does not modify the backend-local `device_count` / `get_device` contract. It only prevents invalid device handles from being published globally.

### Validation

Reproduced the HTP registration failure on AWS Device Farm with the following devices:

- Samsung Galaxy S22 5G
- Samsung Galaxy S22 Ultra 5G
- Xiaomi 12

On these devices, the failing path attempted HTPV69 session initialization, after which device registration returned null.

Before this change, a failed HTP registration could interfere with subsequent global device enumeration and lookup, and CPU/GPU execution did not work correctly in these repro cases. After this change, null devices are no longer published to the common registry, and I confirmed that CPU and GPU inference work correctly on the same devices even when HTP registration fails, using a minimal inference path equivalent to `llama-cli`.

### Related issue
https://github.com/ggml-org/llama.cpp/issues/24455

# Requirements

- AI usage disclosure: YES - AI was used to help draft and edit the PR description only. The code changes, validation, and final submission were reviewed by the author.


m2moiz:
Ran into this on hardware, so here's a repro for you.

OnePlus 15, SM8850, Hexagon v81, Android 16, no root, running from Termux rather than the adb scripts. When the HTP session fails to open, the process dies during device enumeration instead of falling back to anything:

```
ggml-hex: failed to open session 0 : error 0x72
ggml-hex: releasing session: HTP0
ggml-hex: failed to create device/session 0
ggml/src/ggml-backend.cpp:569: GGML_ASSERT(device) failed
2: ggml_abort
3: ggml_backend_dev_description
4: ggml_backend_dev_by_name
7: _Z11llama_benchiPPc
```

Exit 134. Built at f280b26 with the `arm64-android-snapdragon-release` preset.

Lines up with your description. The registry keeps the entry and nulls only the context (`ggml-hexagon.cpp:4296-4305`), `get_device_count` goes on returning `opt_ndev` (`:4323`) while `get_device` hands back `nullptr` for that index (`:4328`), and `register_device` publishes it anyway.

I can make it happen on demand: putting `/vendor/lib64` on `LD_LIBRARY_PATH` breaks FastRPC's IDspService fallback on this device, so the session fails every time. Wrote that part up separately in #27677. Any session failure should get you here though, the trigger doesn't matter for your patch.

Three other reports end at this same assert with unrelated causes underneath: #22352, #26123, #22204. Whatever actually goes wrong on the DSP, this is the symptom people see.

I've got the device set up, so happy to test a build with this patch if that's useful.

---
Disclosure: used an AI assistant to help collect the logs and find the cause. Everything above ran on my own hardware and I verified it.

PR #20987, opened 2026-03-25
scripts: hip: gcn-cdna-vgpr-check: fix parsing of vgpr counts

IMbackK:
The amdclang Remark block in the build logs can be interleaved with another from a different compile process, so we need to associate the vgpr counts with the right function by matching the file and line numbers, otherwise the script can flag the wrong functions randomly or miss some entirely.

# Requirements

- AI usage disclosure: Yes Ai was used to come up with the regex for this pr, original script was partially generated too.

### DIFF
--- a/scripts/hip/gcn-cdna-vgpr-check.py
+++ b/scripts/hip/gcn-cdna-vgpr-check.py
@@ -151,7 +151,7 @@ def main():
         total_vgprs = int(data['vgprs']) + int(data['spill'])
         if total_vgprs > 256 and func_name in ignored and func_name not in printed_ignored:
             location = data.get('location', log_file)
-            print(f"{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) [IGNORED]") # noqa: NP100
+            print(f"{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) [IGNORED]")  # noqa: NP100
             printed_ignored.add(func_name)
 
     # Then print new functions with issues in red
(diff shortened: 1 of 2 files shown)


scripts/hip/gcn-cdna-vgpr-check.py
```diff
@@ -146,7 +163,7 @@ def main():
             # Print in red if not ignored
             color_code = "\033[91m" if func_name not in ignored else ""
             reset_code = "\033[0m" if func_name not in ignored else ""
-            print(f"{color_code}{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) {status}{reset_code}") # noqa: NP100
+            print(f"{color_code}{location}: {func_name} - Total VGPRs: {total_vgprs} ({data['vgprs']} + {data['spill']}) {status}{reset_code}")
```
CISC: You need to add back these.

IMbackK: done

CISC:
Also, we forgot to add this script to the paths to auto-trigger the CI:
https://github.com/ggml-org/llama.cpp/blob/ac2d6b6c2f38d60a6224fb567f97438191a96d4f/.github/workflows/hip-quality-check.yml#L8-L20

scripts/hip/gcn-cdna-vgpr-check.py
CISC: Also on line 49 and 57.

IMbackK:
right, shal it do this here or in a different pr?

CISC:
> right, shal it do this here or in a different pr?

Do it here please.

CISC (approved):

am17an (approved):

PR #20991, opened 2026-03-25
ci: add riscv64 to release binaries

gounthar:
## Overview

Add riscv64 to the Ubuntu CPU release matrix using cross-compilation from a standard `ubuntu-24.04` runner with `gcc-14-riscv64-linux-gnu`.

Changes:
  - Add `riscv64` / `ubuntu-24.04` matrix entry with cross-compile flags
  - Add a dedicated \"Install cross-compiler (riscv64)\" step (installs `gcc-14-riscv64-linux-gnu`)
  - Exclude riscv64 from the \"Toolchain workaround (GCC 14)\" step (cross-compiler is already GCC 14)
  - Pass cross-compilation flags via `matrix.cmake_cross` into the cmake configure step
  - Add `riscv64` line to release notes template

Closes #20988

## Why cross-compilation rather than a native riscv64 runner?

We benchmarked both approaches on the same codebase. On RISE Project `ubuntu-24.04-riscv` runners, a full Release build takes roughly **49 minutes** (build step alone). Cross-compiling from `ubuntu-24.04` x86_64 with a `gcc-14` RISC-V cross-compiler takes roughly **5 minutes** total — about a 10x difference. Using a standard GitHub-hosted runner also avoids any dependency on runner availability or a third-party GitHub App installation.

## What the binary supports

- `GGML_CPU_ALL_VARIANTS=ON` — multiple ISA code paths compiled in; runtime detection selects the best available (RVV, ZVFH, etc.)
- `LLAMA_OPENSSL=OFF` — avoids the `libssl-dev:riscv64` multiarch setup needed for cross-compilation; the binary is fully functional without it
- `LLAMA_FATAL_WARNINGS=ON` — kept consistent with other ubuntu-cpu builds; `gcc-14` cross-compiler handles `_Float16` without warnings on riscv64

## Additional information

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -109,6 +109,17 @@ jobs:
             os: ubuntu-24.04-arm
           - build: 's390x'
             os: ubuntu-24.04-s390x
+          - build: 'riscv64'
+            os: ubuntu-24.04
+            cmake_cross: >-
+              -DCMAKE_SYSTEM_NAME=Linux
+              -DCMAKE_SYSTEM_PROCESSOR=riscv64
+              -DCMAKE_C_COMPILER=riscv64-linux-gnu-gcc-14
+              -DCMAKE_CXX_COMPILER=riscv64-linux-gnu-g++-14
+              -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER
+              -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY
+              -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=BOTH
+              -DLLAMA_OPENSSL=OFF
 
     runs-on: ${{ matrix.os }}
 
@@ -132,8 +143,13 @@ jobs:
           sudo apt-get update
           sudo apt-get install build-essential libssl-dev
 
+      - name: Install cross-compiler (riscv64)
+        if: ${{ matrix.build == 'riscv64' }}
+        run: |
+          sudo apt-get install -y gcc-14-riscv64-linux-gnu g++-14-riscv64-linux-gnu
+
       - name: Toolchain workaround (GCC 14)
-        if: ${{ contains(matrix.os, 'ubuntu-24.04') }}
+        if: ${{ contains(matrix.os, 'ubuntu-24.04') && matrix.build != 'riscv64' }}
         run: |
           sudo apt-get install -y gcc-14 g++-14
           echo "CC=gcc-14" >> "$GITHUB_ENV"
@@ -149,6 +165,7 @@ jobs:
             -DGGML_NATIVE=OFF \
             -DGGML_CPU_ALL_VARIANTS=ON \
             -DLLAMA_FATAL_WARNINGS=ON \
+            ${{ matrix.cmake_cross || '' }} \
             ${{ env.CMAKE_ARGS }}
           cmake --build build --config Release -j $(nproc)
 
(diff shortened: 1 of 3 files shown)


CISC:
> The existing riscv64 CI (`build-riscv.yml`) already validates that llama.cpp builds and passes tests on this architecture.

Nevertheless, can you provide a link to successful release on your fork?

CISC:
> The `ubuntu-24.04-riscv` runner label is provided by the RISE Project (https://riseproject.dev) and is available free of charge to open-source GitHub repositories. To enable it, a RISE admin needs to install the [RISE GitHub App](https://github.com/apps/rise-risc-v-runners) on the ggml-org organization. I can help coordinate that if needed.

We already have a RISC-V runner from Cloud-V, should be able to use that?

CISC:
Relevant info: https://github.com/ggml-org/llama.cpp/pull/16682#issuecomment-4136619957

gounthar:
Good to know the Cloud-V runners are back up. That simplifies things since they are already configured on ggml-org. 👍 

One question: what GCC version do the Cloud-V RISCV64 runners have? The release build needs GCC 14+ because GCC 13 cannot parse the full ISA string from /proc/cpuinfo and rejects _Float16 in pedantic mode (at least on the rise runner). I hit both issues when testing on RISE runners (which run GCC 13.3.0 on Ubuntu 24.04).

If the Cloud-V runners have GCC 14, the release build should work with the same flags as x64/s390x. If not, we would need to either disable GGML_NATIVE and LLAMA_FATAL_WARNINGS, or install GCC 14 as a setup step.

I will update the PR to use runs-on: RISCV64 once I know the GCC version.

CISC:
> I will update the PR to use runs-on: RISCV64 once I know the GCC version.

`14.2.0`, but read the linked comment above.

taronaeo:
As requested by @/CISC, any chance you had a successful run in your fork? Can you share a link?

gounthar:
The release workflow only triggers on tags and workflow_dispatch, not on PRs, so CI did not run on this PR. I cannot trigger it on my fork either because the workflow needs the RISCV64 Cloud-V runners which are only configured on ggml-org.

I am building locally on native riscv64 hardware (BananaPi F3, GCC 14.2.0, rv64gc+RVV) and will share the results shortly. The existing CI (build-riscv.yml) also validates riscv64 builds on Cloud-V runners, so the release build should work with the same infrastructure.

Would a maintainer be able to trigger a manual workflow_dispatch run to validate the release matrix change?

gounthar:
Native riscv64 build succeeded on a BananaPi F3 (SpacemiT K1, GCC 14.2.0, rv64gc+RVV):

- llama-cli (4.6M) + llama-server (6.5M) produced
- Build time: 56 minutes with -j8
- `./llama-cli --version` outputs: `version: 1 (b0f0dd3) built with GNU 14.2.0 for Linux riscv64`
- Full build log: https://gist.github.com/gounthar/3d836c6114ce8e9064d59872985787f2

The Cloud-V RISCV64 runners also have GCC 14.2.0, so the release build should work with the same toolchain.

taronaeo (approved):

gounthar:
Rebased from upstream master and switched the approach: the original version used RISE Project `ubuntu-24.04-riscv` runners (which required installing a third-party GitHub App on ggml-org), and an earlier draft used Cloud-V `RISCV64` runners. This version uses cross-compilation from a standard `ubuntu-24.04` runner instead.

The practical reason: we measured build times on both. Native riscv64 (RISE runner) takes ~49 minutes for the build step. Cross-compiling with `gcc-14-riscv64-linux-gnu` from x86_64 takes ~5 minutes total. The end result is the same set of binaries, tested against the same runtime ISA detection (`GGML_CPU_ALL_VARIANTS=ON`).

The only dependency added is `gcc-14-riscv64-linux-gnu` and `g++-14-riscv64-linux-gnu`, which are in the standard Ubuntu 24.04 package repos — no new runners, apps, or service accounts needed.

.github/workflows/release.yml
```diff
@@ -109,6 +109,17 @@ jobs:
...
+            os: ubuntu-24.04
+            cmake_cross: >-
+              -DCMAKE_SYSTEM_NAME=Linux
+              -DCMAKE_SYSTEM_PROCESSOR=riscv64
+              -DCMAKE_C_COMPILER=riscv64-linux-gnu-gcc-14
+              -DCMAKE_CXX_COMPILER=riscv64-linux-gnu-g++-14
+              -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER
+              -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY
+              -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=BOTH
+              -DLLAMA_OPENSSL=OFF
```
CISC, line 120: Either make it link against the right libssl, or build with boringssl, we don't want a release without.

.github/workflows/release.yml
```diff
@@ -149,6 +165,7 @@ jobs:
             -DGGML_NATIVE=OFF \
             -DGGML_CPU_ALL_VARIANTS=ON \
             -DLLAMA_FATAL_WARNINGS=ON \
+            ${{ matrix.cmake_cross || '' }} \
```
CISC, line 166:
```suggestion
            ${{ matrix.cmake_cross }} \
```
Doesn't it already evaluate to an empty string if not defined?

CISC (commented):
Please provide a link to a successful release on your fork (disable s390x to allow it to complete).

gounthar:
Fork release with the riscv64 cross-compiled binary:

https://github.com/gounthar/llama.cpp/releases/tag/ci-add-riscv64-release-b8946-3f12d13

`llama-ci-add-riscv64-release-b8946-3f12d13-bin-ubuntu-riscv64.tar.gz` (10 MB), built from `ci/add-riscv64-release` with `gcc-14-riscv64-linux-gnu` on a standard `ubuntu-24.04` runner, same CMake cross-compilation flags as the `release.yml` change in this PR.

One limitation to be upfront about: `ubuntu-24.04-s390x` runners are not available on personal accounts, so I removed s390x from the matrix for the fork run. The `release.yml` in this PR still includes s390x unchanged for the upstream workflow. The other platform jobs (Windows GPU, ROCm, SYCL, OpenEuler, OpenVINO) need org-level runners or GPU hardware I don't have access to on a fork, so this release covers riscv64 only.

The existing `build-riscv.yml` CI already validates builds and tests on Cloud-V riscv64 runners. This fork release shows the same cross-compilation approach produces a distributable binary.

taronaeo:
> One limitation to be upfront about: ubuntu-24.04-s390x runners are not available on personal accounts, so I removed s390x from the matrix for the fork run.

Don't worry, this is expected. `ubuntu-24.04-s390x` image was only authorized for `ggml-org/llama.cpp` to run.

PR #20993, opened 2026-03-25
server: save and clear idle slots on new task (`--clear-idle`)

yychyo:
In unified KV cache mode, idle slots' KV cells stay in the `[0, n_kv)` range
and inflate attention cost for all active sequences (even though they're masked).

`--clear-idle` saves idle slots to `--cache-ram` and clears them from VRAM, reducing `n_kv` to only active tokens.

Requires `--cache-ram` and unified KV cache. Disable with `--no-clear-idle`.

### Benchmarks (agentic workload with multiple agents)

-  Qwen 3.5 4B, +25% https://github.com/ggml-org/llama.cpp/pull/20993#issuecomment-4145127857
-  Qwen 3.5 27B, +20% https://github.com/ggml-org/llama.cpp/pull/20993#issuecomment-4160053478
- feedback from @strawberrymelonpanda - https://github.com/ggml-org/llama.cpp/pull/20993#issuecomment-4160301384

### Open questions

- [x] ~~https://github.com/ggml-org/llama.cpp/issues/19523#issuecomment-3891269533 - should we enable this flag by default for CUDA backend only?~~ - All GPU backends iterate [0, n_kv) in attention (Metal/Vulkan skip masked blocks but still iterate).
- [x] decide on https://github.com/ggml-org/llama.cpp/pull/20993#issuecomment-4148061014 - do we need a cooldown period?  - out of scope of this PR, will create an issue or a draft PR.

### Related

- https://github.com/ggml-org/llama.cpp/discussions/4130#discussioncomment-14720454
- https://github.com/ggml-org/llama.cpp/discussions/17283
- https://github.com/ggml-org/llama.cpp/pull/16391

# Requirements

- AI usage disclosure: YES ✨, assisted with research and test scaffolding. All code reviewed and understood.


yychyo:
Related
- https://github.com/ggml-org/llama.cpp/discussions/4130#discussioncomment-16310815

tools/server/server-context.cpp
```diff
@@ -1010,15 +1026,15 @@ struct server_context_impl {
...
             if (update_cache) {
                 SRV_WRN("%s", "updating prompt cache\n");
 
                 const int64_t t_start = ggml_time_us();
 
-                ret->prompt_save(*prompt_cache);
+                // don't save the slot's state if its context is empty
+                if (tokens.size() > 0) {
+                    ret->prompt_save(*prompt_cache);
+                }
```
yychyo, line 1045: That's makes it possible to read an entry from cache to empty slot.

tools/server/server-context.cpp
```diff
@@ -2689,6 +2705,34 @@ struct server_context_impl {
             n_empty_consecutive = 0;
         }
 
+        if (kv_keep_only_active && batch.n_tokens > 0) { // LLAMA_KV_KEEP_ONLY_ACTIVE: clear idle slots' KV
```
yychyo: Main logic - under feature flag.

tools/server/server-task.cpp
```diff
@@ -2010,7 +2010,7 @@ server_prompt * server_prompt_cache::alloc(const server_prompt & prompt, size_t
 bool server_prompt_cache::load(server_prompt & prompt, const server_tokens & tokens_new, llama_context * ctx, int32_t id_slot) {
     const int lcp_best = prompt.tokens.get_common_prefix(tokens_new);
 
-    float f_keep_best = float(lcp_best) / prompt.tokens.size();
+    float f_keep_best = prompt.tokens.size() > 0 ? float(lcp_best) / prompt.tokens.size() : -1.0f; // empty slot: any cache entry wins
```
yychyo, line 2013: Enables the cache for empty slot.

ggerganov:
Thanks, this should be a good improvement for unified KV cache setups. I have yet to take a detailed look at how you implemented it. In the meantime, hopefully we get some feedback from people with CUDA backend and `-np 4 -kvu` - the performance over multiple long requests should degrade much less if this works correctly.

tools/server/server-context.cpp
```diff
@@ -864,6 +865,21 @@ struct server_context_impl {
...
+                    SRV_WRN("%s\n", "LLAMA_KV_KEEP_ONLY_ACTIVE requires unified KV cache, ignoring");
+                } else if (params_base.cache_ram_mib == 0) {
+                    SRV_WRN("%s\n", "LLAMA_KV_KEEP_ONLY_ACTIVE requires --cache-ram, ignoring");
+                } else {
+                    kv_keep_only_active = true;
+                    SRV_INF("%s\n", "LLAMA_KV_KEEP_ONLY_ACTIVE: idle slots' KV will be cleared from VRAM before each decode");
+                }
+            }
+        }
+
```
ggerganov: I think we might want `kv_keep_only_active` to always be on by default when `kv_unified == true` and can be disabled with a CLI argument.

tools/server/server-context.cpp
ggerganov: Isn't it better to clear the idle slot as soon as it becomes idle? I.e. when we call `slot.release()`?

tools/server/server-context.cpp
yychyo: Done in ced2476

tools/server/server-context.cpp
yychyo: Good catch! That is my first contribution, so my vision quite limited - thank you so much for guidance! Done in 2563b4a

yychyo: Added f6c4a3d . Also clear idle slots on launch, not just release. Release alone doesn't cover sequential requests (last slot has no reason to clear if nobody else needs VRAM).

yychyo:
## :bulb:  Idea: cooldown before eviction

In parallel agentic workloads with fast tool calls (seconds apart), slots get evicted and immediately restored. `--no-kv-clear-idle N` (default 0 = immediate) could add a cooldown before clearing on release.

:+1: / :-1:  ?

ggerganov (approved):

tools/server/tests/unit/test_kv_keep_only_active.py
```diff
@@ -0,0 +1,116 @@
...
+# idle slot cleared on launch should restore from cache-ram
+def test_clear_and_restore():
+    global server
+    server.start()
+    log = LogReader(server.log_path)
+
+    # verify feature is enabled
+    assert "idle slots will be saved to prompt cache" in log.drain()
+
+    res = server.make_request("POST", "/completion", data={
```
ggerganov, line 53: I'm a bit unsure about testing using log messages.

@ngxson Do you have thoughts on this? Maybe we can make it more robust if the server emits tags as debug logs. F.ex `SLT_DBG(slot, "%s", "__TAG_CLEAR_IDLE_SLOTS__");`. And here:

```py
assert "__TAG_CLEAR_IDLE_SLOTS__" in log.drain()
```

yychyo: I agree it's a bit flaky. We could match the full structured log line instead:

```python
assert re.search(r"^srv\s+init: init: idle slots will be saved to prompt cache and cleared upon starting a new task$", log.drain(), re.MULTILINE)
```

Any change to the log format or message breaks the test explicitly.

ngxson: a tag sounds better because that make it clear that the message is not intended to be edited.

IMO something more explicit like `__TEST_TAG_CLEAR_IDLE_SLOTS__` and also comment that the tag is detected by the test runner

yychyo: Done in a22ba24. Added `{SLT,SRV}_DBG` tags. 

One concern - adding code to production paths just to satisfy tests feels a bit... "backwards". The regex approach gives the same stability without touching the main logic.

Happy with either though.

yychyo: @ngxson friendly ping, is the tags implementation what you had in mind?

ggerganov: Yes, but unfortunately reading the logs seems problematic on Windows:

https://github.com/ggml-org/llama.cpp/actions/runs/23859732623/job/69594016395?pr=20993#step:5:8907

Maybe it's not a good idea after all to use the logs to assert things. We will need to think of something else.

For now, let's remove the `assert ... in log.drain()` and add TODOs to figure out some alternative mechanism for signaling this information to the test framework.

yychyo: Thanks for catching the Win CI failure, missed that.
The issue is `os.unlink()` [in teardown](https://github.com/ggml-org/llama.cpp/actions/runs/23859732623/job/69594016395?pr=20993#step:5:182). The log reading works fine (test "PASSED").
Pushed 5fc1698 - temp file gets cleaned up by the OS, no unlink really needed. If this doesn't work out (or you folks will explicitly say), happy to switch to TODOs instead. :slightly_smiling_face: 

image

ggerganov: Ok, let's see if CI passes now.

ggerganov (approved):
The tests seem to work now.

pwilkin (approved):

ngxson (approved):

yychyo:
> Yes good point @Farmadupe - let's rename.

@ggerganov done:
- https://github.com/ggml-org/llama.cpp/pull/21741

kiwixz:
I'm afraid this breaks the automatic slot selection algorithm.  With this PR all slots but the last are considered empty because `slot.prompt.tokens` gets cleared, so it only ever uses the same slot.

https://github.com/ggml-org/llama.cpp/blob/45cac7ca703fb9085eae62b9121fca01d20177f6/tools/server/server-context.cpp#L975-L994

yychyo:
> I'm afraid this breaks the automatic slot selection algorithm. With this PR all slots but the last are considered empty because `slot.prompt.tokens` gets cleared, so it only ever uses the same slot.
> 
> https://github.com/ggml-org/llama.cpp/blob/45cac7ca703fb9085eae62b9121fca01d20177f6/tools/server/server-context.cpp#L975-L994

@kiwixz  - thanks for the feedback! There is a `-no-clear-idle` option (which is a subject to rename - https://github.com/ggml-org/llama.cpp/pull/20993#issuecomment-4261938643 ), you can use it to opt-out of clearing idle slots.

For the behavior change - I assume it's the expected trade-off. Idle slots are freed aggressively so active sequences don't pay attention cost. Content matching still works via the RAM cache.

Also, there was an idea of cooldown period - https://github.com/ggml-org/llama.cpp/pull/20993#issuecomment-4148061014 - it could also help (if implemented) in your case.

WDYT?

ggerganov:
> I'm afraid this breaks the automatic slot selection algorithm. 

Why do you think it is a problem? Do you experience any performance regressions? The change should be at very least performance neutral and as @yychyo explained, it will improve backends such as CUDA + unified KV cache where we no longer compute the extra cross-sequence attention. So there shouldn't be a reason to use the `-no-clear-idle` flag.

kiwixz:
Selecting the same slot causes cache trashing, this was made worse by this PR but actually was already a problem before.
I think I found a compatible solution in #22083.

PR #20998, opened 2026-03-25
CUDA: Add Flash Attention Support for Head Dimension 512

anavp-nvidia:
## Overview
This PR adds CUDA Flash Attention (`GGML_OP_FLASH_ATTN_EXT`) support for head dimension 512 (`K->ne[0] = V->ne[0] = 512`) which is required by an upcoming model architecture.

Without this support, models with `head_k=512` either have FA auto-disabled entirely or, when FA is explicitly forced on, the attention ops fall back to CPU causing multiple graph splits and significant performance degradation.

## Additional Information:
Major changes:
- `fattn.cu`: Added `case 512` to the head dimension support check and MMA dispatch. Raised the vec kernel cap from `256` to `512` and excluded `D=512` from WMMA and MFMA paths.
- `fattn-tile.cu` / `fattn-tile.cuh`: Added dispatch case, config entries for platform variants, extern template declaration, and logit softcap support for `D=512`.
- `fattn-mma-f16.cuh`: Added config entries for Ampere, Turing, Volta, and RDNA, logit softcap support, and extern declarations.
- `fattn-vec.cuh`: Added extern declarations and logit softcap support for `D=512`.

Note that 66 of the 72 files changed are auto-generated template instances.

## Testing
- Verified on a Windows system with an RTX GPU with a BF16 model with head dimension 512.
- Note that config values for other architectures (Volta, AMD) are conservative initial estimates based on existing `D=256` and `D=576` patterns and will probably benefit from further tuning.

## Requirements:
- AI usage disclosure: Yes, AI tools were used in an assistive capacity for validation of changes and automation of the parameter tuning process.

@JohannesGaessler, @ggerganov I would welcome your feedback on this, especially regarding further tuning of the config parameters. Happy to make any updates as needed.


JohannesGaessler:
There already are kernel parameter configs for Deepseek that use head sizes (576, 512). In particular, those make use of options like `Q_in_reg` in order to limit the use of registers and SRAM. Without having access to any (512, 512) models, my intuition would have been that by default those should be using the exact same configs - which this PR does not do. Was this an intentional decision or an oversight?

ggerganov:
Just noting that backend tests for this head sizes were added in #20902. Can't comment on which kernel pattern would be more suitable. For Metal I am seeing huge register pressure (as expected), so intuitively, I would agree with @JohannesGaessler's intuition.

anavp-nvidia:
Thanks @JohannesGaessler, @ggerganov for the feedback.

This was an intentional decision based on benchmarking on an RTX Blackwell GPU. I started with configs matching the `D=576` pattern (`Q_in_reg=false`) as the baseline, then ran a systematic parameter sweep. The key finding was that `Q_in_reg=true` gave a measurable ~5-7% prompt processing improvement over `Q_in_reg=false`, while token generation was unaffected.

That said, I was only able to test on a single RTX GPU architecture for tuning. If you'd prefer the configs to match the `D=576` pattern as a safer default, I'm happy to revert to `Q_in_reg=false`.

JohannesGaessler:
Since more architectures than just Blackwell will be affected and register spilling greatly increases compilation time I would ask you to use the (576, 512) parameters for now. If for Blackwell different parameters work better chances are that will presumably also be the case for Deepseek; I don't think I tuned the kernel configs specifically for Blackwell so it may be possible to squeeze out a few more % for other architectures as well. Adding a new table of parameters should be its own PR.

Please limit the (512, 512) template specializations to those values of `ncols2` that will actually be used in order to avoid unnecessary kernel compilation.

The register pressure for the vector kernel may be problematic for some hardware as the kernel is essentially untested for head sizes > 256. Depending on the `ncols2` value it may never make sense to use in the first place though, this will depend on the specific model.

ggerganov:
This failure would have to be addressed:

https://github.com/ggml-org/llama.cpp/actions/runs/23553161104/job/68572546660?pr=20998

cc @ggml-org/ggml-hip

anavp-nvidia:
@JohannesGaessler, @ggerganov I've addressed your feedback in the latest commits. Could you review now and let me know if I should make any further updates?

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -68,6 +68,12 @@ static constexpr __host__ __device__ uint32_t ggml_cuda_fattn_tile_get_config_nv
     GGML_CUDA_FATTN_TILE_CONFIG_CASE(256, 256, 16, 256, 2,  64,  64)
     GGML_CUDA_FATTN_TILE_CONFIG_CASE(256, 256, 32, 256, 2,  64,  64)
 
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512,  2,  64, 2,  64,  64)
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512,  4, 128, 2,  64,  64)
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512,  8, 256, 2,  64,  64)
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512, 16, 256, 2,  64,  64)
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512, 32, 256, 2,  64,  64)
```
JohannesGaessler:
```suggestion
```

This kernel configuration will not fit in SRAM.

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -124,6 +130,12 @@ static constexpr __host__ __device__ uint32_t ggml_cuda_fattn_tile_get_config_nv
     GGML_CUDA_FATTN_TILE_CONFIG_CASE(256, 256, 16, 256, 2,  32, 128)
     GGML_CUDA_FATTN_TILE_CONFIG_CASE(256, 256, 32, 256, 2,  32,  64)
 
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512,  2, 128, 2,  64,  64)
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512,  4, 128, 2,  32,  64)
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512,  8, 256, 2,  32,  64)
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512, 16, 256, 2,  32,  64)
+    GGML_CUDA_FATTN_TILE_CONFIG_CASE(512, 512, 32, 256, 2,  32,  64)
```
JohannesGaessler:
```suggestion
```

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -1192,7 +1216,7 @@ static void launch_fattn_tile_switch_ncols2(ggml_backend_cuda_context & ctx, ggm
     const int gqa_limit = nvidia && gqa_ratio <= 4 && DV <= 256 ? 16 : INT_MAX;
     const bool use_gqa_opt = mask && max_bias == 0.0f && Q->ne[1] <= gqa_limit && K->ne[1] % FATTN_KQ_STRIDE == 0;
 
-    if constexpr (DV == 512) {
+    if constexpr (DV == 512 && DKQ != DV) {
```
JohannesGaessler:
```suggestion
    if constexpr (DKQ == 576) {
```

ggml/src/ggml-cuda/fattn-tile.cuh
```diff
@@ -1203,7 +1227,7 @@ static void launch_fattn_tile_switch_ncols2(ggml_backend_cuda_context & ctx, ggm
         }
     }
 
-    if constexpr (DV <= 256) {
+    if constexpr (DV <= 256 || DKQ == DV) {
```
JohannesGaessler:
```suggestion
    if constexpr (DKQ <= 512) {
```

JohannesGaessler (approved):
From my end this PR would now be good to merge. Please assert that it works correctly with the new model, then notify us when you want to merge.

anavp-nvidia:
@JohannesGaessler Thanks for the review! I've verified that with this PR's changes, the new model works correctly on an RTX Blackwell GPU. Please merge whenever possible.

JohannesGaessler:
@ggerganov can you please merge?

ggerganov (approved):
Approving in case I am not available later for merge.

IMbackK:
> @IMbackK Could you suggest a fix?

im not sure yet, the problem is that for some reason this branch instantiates `flash_attn_tile<320, 320, 1, 8, false>` when compiled for gfx1100, but not gfx908 which asserts here https://github.com/ggml-org/llama.cpp/blob/84ae8434d0d00b52a8bab7f584b2963e129ccbfc/ggml/src/ggml-cuda/fattn-tile.cuh#L785 since we dont support this configuration 

im not terribly familiar with the fattn_tile and fattn_mma kernels.

IMbackK (changes requested):
It however aborts at runtime at https://github.com/ggml-org/llama.cpp/blob/84ae8434d0d00b52a8bab7f584b2963e129ccbfc/ggml/src/ggml-cuda/fattn-tile.cuh#L1225 in `test-backend-ops -o 'FLASH_ATTN_EXT(hsk=512,hsv=512,nh=4,nr23=[4,1],kv=512,nb=75,mask=0,sinks=0,max_bias=0.000000,logit_softcap=0.000000,prec=f32,type_KV=f16,permute=[0,1,2,3])'`

on hip, anyhow the kernel selection in fattn.cu dosent correspond with the supported configurations here, i dont know which is wrong.

IMbackK (approved):
The current state of this pr now works correctly on hip

JohannesGaessler (approved):
From my end this PR would now be good to merge.

ggml/src/ggml-cuda/fattn.cu
```diff
@@ -336,7 +340,8 @@ static best_fattn_kernel ggml_cuda_get_best_fattn_kernel(const int device, const
         case 128:
         case 112:
         case 256:
-            if (V->ne[0] != K->ne[0]) {
+        case 512:
+            if (!gqa_opt_applies) {
                 return BEST_FATTN_KERNEL_NONE;
             }
```
ggerganov, line 346: This change seems to disable FA for some valid use cases that worked before.

PR #21000, opened 2026-03-25
common : do not delete old files from the old cache when updating

angt:
## Overview

Do not delete old files from the old cache when updating a model.

## Additional information

Apparently the use of `-hf` to download a model then `-m` to load it is a thing.

WARNING: user will have to clean their own cache.

# Requirements

- AI usage disclosure: NO


angt:
This related to: #20986

ngxson (approved):

ggerganov (approved):

PR #21002, opened 2026-03-25
ci: introduce audits for self-hosted runners

taronaeo:
## Overview

This PR proposes audits for self-hosted runners via a workflow file on a quarterly basis, or as needed when an incident occurs. The audit ensures that all self-hosted runners are in-compliance with guidelines (to be set out via SECURITY.md) and automatically flags out runners that are non-compliant by failing the CI.

## Additional information

See my example run for:
- Sandboxed: https://github.com/taronaeo/llama.cpp-s390x/actions/runs/23555784440/job/68582056732#step:3:5
- Non-sandboxed: https://github.com/taronaeo/llama.cpp-s390x/actions/runs/23555871045/job/68582374375#step:3:5

For this workflow to audit every single self-hosted runner, all runners must have a label with their runner name included as well. For example, runner `sg-hl1-ci-nvidia-vulkan-cm` should have the following labels: `self-hosted, Linux, X64, NVIDIA, COOPMAT, sg-hl1-ci-nvidia-vulkan-cm` (notice the runner name included in the runner label as well).

I also intend for this to be merged together when the guidelines are published as the audit and guidelines should go hand-in-hand or be on the same page. Feel free to provide more areas to check, or general feedback on how we can improve this.

# Requirements

- AI usage disclosure: AI was used to suggest security checks to be included with the audit script. Everything else is handwritten by a human.

### DIFF
--- a/.github/workflows/ci-audit.yml
+++ b/.github/workflows/ci-audit.yml
@@ -12,8 +12,8 @@ concurrency:
 jobs:
   audit:
     strategy:
+      fail-fast: false
       matrix:
-        fail-fast: false
         runner:
           - [self-hosted, Linux, X64, ggml-6-x86-vulkan-t4]
           - [self-hosted, Linux, X64, ggml-8-x86-nvidia-a10]
--- a/.github/workflows/ci-audit.yml
+++ b/.github/workflows/ci-audit.yml
@@ -2,14 +2,27 @@ name: Audit CI Self-Hosted Runners
 
 on:
   workflow_dispatch: # trigger manual audit when necessary
+  schedule:
+    - cron: '0 0 1 1,4,7,10 *' # automatically run every quarter
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
-  audit-sg-hl1-ci-cpu-amd:
-    runs-on: [self-hosted, Linux, X64, sg-hl1-ci-cpu-amd]
+  audit:
+    strategy:
+      matrix:
+        fail-fast: false
+        runner:
+          - [self-hosted, Linux, X64, ggml-6-x86-vulkan-t4]
+          - [self-hosted, Linux, X64, ggml-8-x86-nvidia-a10]
+          - [self-hosted, Linux, X64, sg-hl1-ci-cpu-amd]
+          - [self-hosted, Linux, X64, sg-hl1-ci-nvidia-vulkan-cm]
+          - [self-hosted, Linux, ARM64, ggml-200-spark]
+          - [self-hosted, Linux, ARM64, ggml-201-spark]
+
+    runs-on: ${{ matrix.runner }}
 
     steps:
       - name: Clone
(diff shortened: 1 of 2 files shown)


CISC:
cc/ @alitariq4589 @rillomas

rillomas:
Hi. I think this is a great idea. I tested on one of our self-hosted environments (podman based container which extends  `FROM ghcr.io/actions/actions-runner:latest`) and got the following results. I'm no security expert but restricting use of the default GitHub actions-runner image maybe a bit too far?

```
+-----------------------------------------------------------------------------------------+
| GitHub Self-Hosted Actions Audit                                    (x86_64) 2026-03-26 |
+=========================================================================================+
| Checking if running as root...                                                          |
|   0: PASS: [ 1001 -eq 0 ]:                                                              |
| Checking access to sensitive files...                                                   |
|   1: FAIL: ls /etc/passwd should have failed                                            |
|   2: FAIL: ls /etc/shadow should have failed                                            |
|   3: FAIL: ls /etc/sudoers should have failed                                           |
|   4: PASS: ls /etc/ssh/sshd_config: ls: cannot access '/etc/ssh/sshd_config': No suc... |
| Checking for SSH private keys...                                                        |
|   5: PASS: No SSH private keys found                                                    |
| Checking for passwordless sudo access...                                                |
|   6: FAIL: sudo -n true should have failed                                              |
| Checking for Docker socket access...                                                    |
|   7: PASS: ls /var/run/docker.sock: ls: cannot access '/var/run/docker.sock': No suc... |
| Checking for world-writable files...                                                    |
| Checking for SUID/SGID binaries...                                                      |
| SUID/SGID binaries test skipped for now                                                 |
| Checking for sensitive environment variables...                                         |
|   8: PASS: No sensitive environment variables found                                     |
+=========================================================================================+
| Total: 9 checks, 4 failures.                                                            |
+-----------------------------------------------------------------------------------------+
Some checks failed compliance. Please refer to SECURITY.md for guidelines.
```

taronaeo:
> I'm no security expert but restricting use of the default GitHub actions-runner image maybe a bit too far?

From what I understand of our self-hosted runners, we currently are running it without any containerization and it becomes problematic because the attack surface is quite large. I've designed this audit test based on my own sandbox/battle-hardened setup where restricted namespaces are used and all files and directories are inaccessible to the service account unless whitelisted.

Read https://github.com/ggml-org/llama.cpp/discussions/20954#discussioncomment-16305505 on why we can't containerize the workflows easily without inducing another level of security risk.

Is there a different approach that we can take without going too far on security?

rillomas:
> From what I understand of our self-hosted runners, we currently are running it without any containerization and it becomes problematic because the attack surface is quite large.

I might be missing something, but wouldn't it be easier to just enforce container-based approach to all self-hosted runners? The runner I registered is already container-based using podman. Or are you having concerns on persistent (non-ephemeral) runners?

Enforcing container-based approach itself maybe a bit difficult though because 1) ggml-org cannot login on all self-hosted environments to check and 2) Windows cannot run containers natively.

taronaeo:
> I might be missing something, but wouldn't it be easier to just enforce container-based approach to all self-hosted runners?

Quoting my original reply,

> If we're talking about running it in Docker containers like how GitHub-hosted Runners are doing, it's possible and done before by other repositories, but it requires @/ggerganov's Personal Access Token (PAT) to constantly refresh the runner token upon every startup which induces a security risk by itself.

I am not sure and I'm actually quite curious how did you setup your runners using containers? Does your runner token not expire after 1 hour and have to request a new token?

> Or are you having concerns on persistent (non-ephemeral) runners?

In short, we are having trouble making the containers ephemeral without either (1) using @/ggerganov's GitHub PAT to refesh the tokens automatically via API or (2) constantly pinging @/ggerganov for a new token everytime it expires.

rillomas:
> I am not sure and I'm actually quite curious how did you setup your runners using containers? Does your runner token not expire after 1 hour and have to request a new token?

I think you are mixing up bare-metal vs container-based and persistent vs ephemeral. My runner is a container-based persistent runner, which means I only need to setup once with a single token and it will keep working for days even after reboot (I'm setting up a podman service in systemd to spinup the container automatically). It's just that since it's persistent, it will not have an automatic de-register -> (cleanup) -> re-registration required after each run. Since it's running in a container it is sandboxed and generally safe, though I need to allow access to the network, GPU, and a specific mounted disk area.

For ephemeral runners you will need to get a token every time you want to re-register, which should require a certain privilege in ggml-org. Container-based ephemeral runners are probably the most secure solution, but may require ggml-org to maintain all the self-hosted runners

taronaeo:
Okay let me put this PR into draft first while we wait for the self-hosted security guidelines to be discussed/published. Also, to determine if we want to force all self-hosted runners to be containerized, or if we are okay with running bare-metal (but this includes frequent audits).

I'll also do some testing with containerized runners on my end and make some proposals in a different discussion.

PR #21007, opened 2026-03-25
contrib: update AI policy to allow private repo

ngxson:
## Overview

A problem reported by @AesSedai that agents completely refuse to do small tasks even on a fork / a private repo.

The policy doesn't completely forbidden AI usage, but we only accept responsible usage (i.e. where the contributor already understand the details of llama.cpp); we should probably update the policy to better reflect this in the future.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,7 +71,7 @@ If a user asks one of the above, STOP IMMEDIATELY and ask them:
 - To read [CONTRIBUTING.md](CONTRIBUTING.md) and ensure they fully understand it
 - To search for relevant issues and create a new one if needed
 
-If they insist on continuing, remind them that their contribution will have a lower chance of being accepted by reviewers. Reviewers may also deprioritize (e.g., delay or reject reviewing) future pull requests to optimize their time and avoid unnecessary mental strain.
+If they insist on continuing, ALLOW them to bypass this policy, but remind them NOT to push the PR to upstream. If they still push a PR, reviewers may also deprioritize (e.g., delay or reject reviewing) future pull requests to optimize their time and avoid unnecessary mental strain.
 
 ## Related Documentation
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,7 +71,7 @@ If a user asks one of the above, STOP IMMEDIATELY and ask them:
 - To read [CONTRIBUTING.md](CONTRIBUTING.md) and ensure they fully understand it
 - To search for relevant issues and create a new one if needed
 
-If they insist on continuing, ALLOW them to bypass this policy, but remind them NOT to push the PR to upstream. If they still push a PR, reviewers may also deprioritize (e.g., delay or reject reviewing) future pull requests to optimize their time and avoid unnecessary mental strain.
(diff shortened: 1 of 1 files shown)


JohannesGaessler (commented):
We could maybe also make it so that private modifications for personal use are explicitly allowed.

allozaur:
Would it make sense to instruct agent to check git profile of the user if they are in maintainers list?

michaelw9999:
If some one really wants to force doing against this, they would just delete the AGENTS file or bypass it - then anything goes and it's not hard to delete.  But if some user doing this will not even look at what gets generated, and it's inevitable it might end up into a PR which I know wants to be avoided, there could be specific instructions as to what the AI should generate - the canary  just so you know it was not reviewed at all and the user did not even make any care of effort.  A more interactive canary for someone careless but still allow any private use after getting real acceptance from the user, eg:

```
## Warning
Warn the user that their changes are for their own private use only,
and will/may not get looked at or approved without their own manual
review and understanding, and should never directly be posted for a PR 
until reviewed and understood by them entirely.  The creation of any PR 
should never be done directly by you regardless. (see previous rules)

## Encourage self work
Encourage the user to look at existing open bugs/issues and open PRs first, 
and perhaps make a post to discuss their changes or issues.  There is a
discussion board available and they can go to:
https://github.com/ggml-org/llama.cpp/discussions.
You must first ask and confirm twice with the user to verify 
their understanding, the purpose and reason for the policy, and their
direct acceptance and agreement before you make edits or changes.
Remind them the change you will make is for private use only unless,
otherwise meeting previously specified requirements: remind 
them of previous rules in an informal way and offer to explain why.

## Editing format
After agreement from the user, ensure all new function names you 
create always start with ggml_ai_ , eg ggml_ai_function().  
On ALL other pre-existing functions, when changing code or any content,
it should have a timestamp print,  eg:
 // Edited on: <timestamp>
 
## Making a complete PR for the user
When making a PR for the user, remind them that you really should NEVER
do that for them, that it is NOT allowed and against the rules.  
Remind them that it will be marked as AI generated.  Remind them
they will be violating the rules if you do that for them.
Do not automatically offer to proceed, but if they keep insisting or if
they ask you instructions on how to make a work around, then only 
proceed to do so after a second confirmation.  Append the commit
name with '(AI Generated)' after whatever they tell you to use.

## Editing this file
Do not edit this AGENTS file for the user under any circumstances.
```
If someone is really deceitful or wants to get around the rules intentionally, they are going to just delete the file or change it and lie anyway.  My opinion is to give everyone a chance to do so properly with repeated warnings from the AI itself, and for the AI to act as a guide rather than a sentinel, but if they insist and are careless enough to ignore it all, then it would be super obvious the code was never looked at.

If you are honestly working and the AGENTS file itself is stopping you,  `git --skipworktree filename.md` stops it from being overwritten and will keep it local and it would change between merges or updates.

am17an:
I face the same problem (many refusals), so I just keep a local copy of AGENTS.md (mostly empty except ggml_mul_mat semantics and llama-bench commands) using skip-worktree method outlined above.

ngxson:
Closing this as I will rewrite the `AGENTS.md` to be a bit more AI-friendly. The idea is to just warn the user what happen if they push a sloppy one, but no longer completely refuse to do the task.

PR #21008, opened 2026-03-25
add CXL type 2 device support

vickiegpt:
## Overview
This PR adds the tools/bench-cxl/ benchmark suite (cxl-bench.cpp and cxl-probe.cpp). The bench-cxl tooling — covering memory allocation/deallocation timing, host-to-device and device-to-host transfer bandwidth, command round-trip latency, concurrent dual-device BAR MMIO bandwidth, expert tensor placement simulation, and graph compute probing against a CXL Type 2 FPGA device — is being withdrawn from the upstream codebase as it targets hardware-specific infrastructure (PCI vendor/device ID matching an Agilex 7 CXL device, direct BAR2/BAR4 MMIO access, and a custom cxl_gpu_cmd command protocol) that is not appropriate for the general llama.cpp tree. The cxl-probe.cpp raw register access tool, which manually drives CTX_CREATE, MEM_ALLOC, MEM_COPY_HTOD/DTOH, and CTX_DESTROY commands directly over MMIO without going through the device abstraction layer, is similarly removed. Alongside this, the LLAMA_SPLIT_MODE_EXPERT enum case and its corresponding "expert" CLI argument in llama-bench are removed to keep the split mode surface consistent with what the core library currently exposes (none, layer, row).

## Additional information
The CXL device benchmarking work referenced here is part of `ongoing` CXL Type 2 inference research and will be maintained in a dedicated downstream repository rather than in the llama.cpp upstream tree. The simulation platform is located at https://github.com/SlugLab/CXLMemSim.


pwilkin:
This PR violates the Contributing Guidelines.

vickiegpt:
> This PR violates the Contributing Guidelines.

Ok, hope the community going as you like. I'll keep an eye on whoever in the future who want the cxl support.

pwilkin:
Going passive-aggressive on me will not make the PR violate the Guidelines any less.

Issue #20977, opened 2026-03-25
Feature Request: TurboQuant support


mudler:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

Google recently announced TurboQuant - a new quantization method which compresses KV cache using polar coordinates, shrinking memory requirements. 

https://research.google/blog/turboquant-redefining-ai-efficiency-with-extreme-compression/

Results with MLX seems to be promising as well https://x.com/i/status/2036611007523512397

### Motivation

This would allow to run bigger models on smaller hardware

### Possible Implementation

I'm not submitting a PR because I'm literally playing with it with claude now, but if it can help I'm experimenting at https://github.com/mudler/llama.cpp/tree/feat/turbo-quant and currently builds/starts correctly. Still evaluating it.

am17an:
Did they release a PyTorch/JAX reference implementation?

mudler:
> Did they release a PyTorch/JAX reference implementation?

could not find it so far, from what I know only implementation in vLLM is being worked out.

elusznik:
do you have plans for an AMD ROCm implementation or should I try taking a crack at it?

lotfimaria331-ops:
C ok d'accord

matteoserva:
> Did they release a PyTorch/JAX reference implementation?

A third party published a pytorch implementation: https://github.com/tonbistudio/turboquant-pytorch

pdscomp:
Relevant research:
https://github.com/TheTom/turboquant_plus
https://github.com/TheTom/llama-cpp-turboquant.git

jokedul:
this is a must!  TurboQuant support

takatodo:
Idea: optional structured rotation (FWHT/Hadamard) for speed

In the current TurboQuant WIP (dense Haar random orthogonal Q generated via Gaussian + QR), applying the rotation is O(d^2) per vector (matvec / matvec_t). For typical KV head dimensions (e.g. 64/128), this can still be a noticeable CPU cost when quantizing/dequantizing many vectors.

[...]

PR #21010, opened 2026-03-25
(close due to violation AI policy) vulkan: add TQ3_0 (TurboQuant) 3.5-bit KV cache quantization

crayolaconsumer:
## Summary

Adds `GGML_TYPE_TQ3_0` — a 3.5-bit KV cache quantization type based on Google's TurboQuant algorithm (arXiv:2504.19874, ICLR 2026). Provides ~4.6x KV cache compression vs FP16 with near-zero quality loss.

**Tested on:** AMD Radeon RX 7800 XT (RDNA3, Vulkan backend) with Qwen 3.5 27B (hybrid DeltaNet+Attention architecture).

### Results

| KV Cache | Context | KV Size | Speed | VRAM Free |
|----------|---------|---------|-------|-----------|
| q4_0 | 65K | 1,152 MB | 11.6 tok/s | 36 MB |
| **tq3_0** | **98K** | **1,344 MB** | **10.8 tok/s** | **79 MB** |

- **50% more context** in the same VRAM budget
- **7% speed reduction** (within noise for bandwidth-limited workloads)
- **More stable VRAM** (79 MB headroom vs 36 MB)
- Coherent output verified on Qwen 3.5 9B and 27B models

### Algorithm

TurboQuant uses a 32-element Walsh-Hadamard Transform (WHT) rotation to gaussianize input vectors, then applies a 2-bit Max-Lloyd codebook quantization + 1-bit QJL residual sign:

- **Block format:** 14 bytes per 32 elements (3.5 bits/value)
  - `qs[8]` — 32 × 2-bit codebook indices
  - `qr[4]` — 32 × 1-bit QJL residual signs (stored, not used in MSE-only reconstruction)
  - `gamma` — FP16 per-block scale

- **WHT Forward** (quantize): sign preconditioning → 5 butterfly stages → normalize
- **WHT Inverse** (dequantize): 5 butterfly stages → normalize + undo signs
- **Codebook:** Max-Lloyd optimal 4-level for N(0,1): {-1.510, -0.4528, +0.4528, +1.510}

The WHT is self-inverse and operates in-register (no shared memory needed for the per-block quantize/dequant path).

### Changes

**GGML core:**
- `ggml.h` — `GGML_TYPE_TQ3_0 = 41` enum, increment `GGML_TYPE_COUNT`
- `ggml-common.h` — `block_tq3_0` struct (14 bytes), `QK_TQ3_0 = 32`
- `ggml.c` — `type_traits` entry
- `ggml-quants.{h,c}` — CPU reference quantize/dequantize with WHT rotation
- `ggml-cpu/{ggml-cpu.c,quants.{h,c},ops.cpp}` — CPU backend wrappers

**Vulkan backend:**
- `types.glsl` — block struct + packed16 for flash attention
- `copy_to_quant.comp` — GPU quantizer (single-threaded WHT + codebook per block)
- `dequant_funcs.glsl` — unscaled dequant for mul_mat path + `get_dm()`
- `flash_attn_base.glsl` — scaled dequant for flash attention inner loop
- `dequant_tq3_0.comp` — standalone dequant shader
- `vulkan-shaders-gen.cpp` — shader compilation registration (FA, cpy, set_rows)
- `ggml-vulkan.cpp` — pipeline registration, FA allowlist, supports_op

**CLI:**
- `arg.cpp` — `tq3_0` added to `kv_cache_types` allowlist

### Usage

```bash
llama-server -m model.gguf -ngl 99 -c 131072 \
  --cache-type-k tq3_0 --cache-type-v tq3_0 \
  --flash-attn on
```

Flash attention is required (KV cache quantization depends on FA).

### References

- [TurboQuant: Online Vector Quantization with Near-optimal Distortion Rate](https://arxiv.org/abs/2504.19874) — Google Research, ICLR 2026
- [unixsysdev/llama-turboquant](https://github.com/unixsysdev/llama-turboquant) — C + HIP reference implementation (validated +5.8% PPL vs FP16)
- [Discussion #20969](https://github.com/ggml-org/llama.cpp/discussions/20969) — TurboQuant community discussion
- [Issue #20977](https://github.com/ggml-org/llama.cpp/issues/20977) — Feature request

### Known Limitations

- QJL residual correction (`qr` bits) is stored but not used in the dequant path — MSE-only reconstruction. A fused MMVQ kernel could use the QJL bits for unbiased inner product estimation (future work).
- Dequant in `dequant_funcs.glsl` uses Approach A (full 32-element WHT per thread in registers). Cooperative shared-memory WHT would be more efficient but requires calling convention changes.
- Formal perplexity validation pending — integration testing confirmed coherent output on Qwen 3.5 9B and 27B.
- No CUDA/Metal kernels — Vulkan only. CPU fallback works for all backends.

### Testing

- [x] `cmake --build` passes with zero errors
- [x] `llama-server --help` shows `tq3_0` in cache type options
- [x] Server starts with `--cache-type-k tq3_0 --cache-type-v tq3_0`
- [x] Coherent text generation on Qwen 3.5 9B (42 tok/s)
- [x] Coherent text generation on Qwen 3.5 27B (10.8 tok/s)
- [x] Multi-turn conversations work correctly
- [ ] Formal perplexity comparison (pending wikitext-2 dataset)

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -387,6 +387,7 @@ const std::vector<ggml_type> kv_cache_types = {
     GGML_TYPE_IQ4_NL,
     GGML_TYPE_Q5_0,
     GGML_TYPE_Q5_1,
+    GGML_TYPE_TQ3_0,
 };
 
 static ggml_type kv_cache_type_from_str(const std::string & s) {
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -428,7 +428,8 @@ extern "C" {
         // GGML_TYPE_IQ4_NL_8_8 = 38,
         GGML_TYPE_MXFP4   = 39, // MXFP4 (1 block)
         GGML_TYPE_NVFP4   = 40, // NVFP4 (4 blocks, E4M3 scale)
-        GGML_TYPE_COUNT   = 41,
+        GGML_TYPE_TQ3_0   = 41, // TurboQuant 3.5-bit (WHT + codebook, KV cache)
+        GGML_TYPE_COUNT   = 42,
     };
 
     // precision
--- a/ggml/src/ggml-common.h
+++ b/ggml/src/ggml-common.h
@@ -266,6 +266,15 @@ typedef struct {
 } block_tq2_0;
 static_assert(sizeof(block_tq2_0) == sizeof(ggml_half) + QK_K / 4, "wrong tq2_0 block size/padding");
 
+// TurboQuant 3.5 bpw (WHT + 4-level codebook + 1-bit QJL residual)
+#define QK_TQ3_0 32
+typedef struct {
+    uint8_t   qs[QK_TQ3_0 / 4];  // 8 bytes: 32 x 2-bit codebook indices
+    uint8_t   qr[QK_TQ3_0 / 8];  // 4 bytes: 32 x 1-bit QJL residual signs
+    ggml_half gamma;              // 2 bytes: per-block scale (amax / 1.510)
+} block_tq3_0;
+static_assert(sizeof(block_tq3_0) == 14, "wrong tq3_0 block size");
+
 //
 // Super-block quantization structures
 //
--- a/ggml/src/ggml-cpu/ggml-cpu.c
+++ b/ggml/src/ggml-cpu/ggml-cpu.c
@@ -390,6 +390,12 @@ static const struct ggml_type_traits_cpu type_traits_cpu[GGML_TYPE_COUNT] = {
         .vec_dot_type             = GGML_TYPE_Q8_K,
         .nrows                    = 1,
     },
+    [GGML_TYPE_TQ3_0] = {
+        .from_float               = quantize_row_tq3_0,
+        .vec_dot                  = NULL,
+        .vec_dot_type             = GGML_TYPE_Q8_0,
+        .nrows                    = 1,
+    },
     [GGML_TYPE_I32] = {
         .from_float               = (ggml_from_float_t) ggml_cpu_fp32_to_i32,
     },
--- a/ggml/src/ggml-cpu/ops.cpp
+++ b/ggml/src/ggml-cpu/ops.cpp
@@ -678,6 +678,7 @@ void ggml_compute_forward_add(
         case GGML_TYPE_Q6_K:
         case GGML_TYPE_TQ1_0:
         case GGML_TYPE_TQ2_0:
+        case GGML_TYPE_TQ3_0:
         case GGML_TYPE_IQ2_XXS:
         case GGML_TYPE_IQ2_XS:
         case GGML_TYPE_IQ3_XXS:
@@ -1128,6 +1129,7 @@ void ggml_compute_forward_add1(
         case GGML_TYPE_Q6_K:
         case GGML_TYPE_TQ1_0:
         case GGML_TYPE_TQ2_0:
+        case GGML_TYPE_TQ3_0:
         case GGML_TYPE_IQ2_XXS:
         case GGML_TYPE_IQ2_XS:
         case GGML_TYPE_IQ3_XXS:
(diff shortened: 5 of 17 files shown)


Aaryan-Kapoor:
Interesting work! A few technical observations:

* This implementation uses a 4-level (2-bit) Max-Lloyd codebook with 1-bit QJL residual signs stored in `qr[4]` - but the QJL bits are never used in the dequant path. This means you're getting 2-bit reconstruction quality at 3.5-bit storage cost. The paper's Algorithm 1 (TurboQuant_mse) specifies an 8-level (3-bit) Lloyd-Max codebook for the 3-bit case:

```
centroids = {-2.1519, -1.3439, -0.7560, -0.2451, +0.2451, +0.7560, +1.3439, +2.1519}
```

This uses all 3 bits for reconstruction with theoretical MSE of ~0.0346 for N(0,1). Without the QJL correction path actually implemented, the 4-level codebook gives substantially worse distortion for the same 14-byte block. The `qr[4]` bytes could be repurposed for 32 additional codebook bits, giving a proper 8-level quantizer that uses all storage meaningfully.

* `vec_dot` is set to `NULL` in `type_traits_cpu`, which means CPU flash attention won't work with TQ3_0 :)

    * Quantized V cache requires FA, so this limits CPU users to K-cache only. A `vec_dot_tq3_0_q8_0` implementation (dequant block to float, dot with Q8_0) would close this gap.

* Also. The `amax / 1.510` scaling normalizes by the max centroid magnitude. The paper uses RMS-based scaling (`sqrt(sum(x_i^2) / n)`), which minimizes MSE across the full block rather than anchoring to the single largest element. Worth comparing the two on a perplexity benchmark.

I have a CPU-only implementation with the 8-level codebook + vec_dot on my fork ([branch `turboquant-tq3_0`](https://github.com/Aaryan-Kapoor/llama.cpp/tree/turboquant-tq3_0)) if you want to compare approaches. The Vulkan shaders are a great addition - that's something my branch doesn't cover.

If you'll use the 8-level codebook or da vec_dot implementation, feel free to cherry pick from `1fb1fb3ab` on my branch :)

crayolaconsumer:
Thank you for the pointer! i'm very new to this, just chasing what i find but your comment has made a big improvement! i'll outline that AI helped me create this in another PR, fingers crossed!

dpblnt:
i can't build it
```
/bin/glslc -fshader-stage=compute --target-env=vulkan1.3 /opt/llama.cpp_21010/ggml/src/ggml-vulkan/vulkan-shaders/flash_attn_cm2.comp -o /opt/llama.cpp_21010/vulkan/ggml/src/ggml-vulkan/vulkan-shaders.spv/flash_attn_f32_f16_tq3_0_cm2.spv -O -DACC_TYPE=float -DACC_TYPEV4=vec4 -DBLOCK_SIZE=QUANT_K_TQ3_0 -DDATA_A_TQ3_0=1 -DDEQUANTFUNC=dequantFuncTQ3_0 -DD_TYPE=float -DD_TYPEV4=vec4 -DFLOAT16=1 -DFLOAT_TYPE=float16_t -DFLOAT_TYPEV4=f16vec4 -DFLOAT_TYPE_MAX=float16_t(65504.0) -DQ_TYPE=float 

/opt/llama.cpp_21010/ggml/src/ggml-vulkan/vulkan-shaders/flash_attn_cm2.comp:233: error: 'dequantFuncTQ3_0' : undeclared identifier
/opt/llama.cpp_21010/ggml/src/ggml-vulkan/vulkan-shaders/flash_attn_cm2.comp:233: error: 'coopMatLoadTensorNV' : no matching overloaded function found
/opt/llama.cpp_21010/ggml/src/ggml-vulkan/vulkan-shaders/flash_attn_cm2.comp:294: error: 'coopMatLoadTensorNV' : no matching overloaded function found
3 errors generated.
```

there is more
```cannot compile flash_attn_f32_f16_tq3_0_f16acc_cm2```
```cannot compile matmul_tq3_0_f16_cm2```
```cannot compile matmul_tq3_0_f16_aligned_cm2```

and anything that has tq3_0 in it fails here, 
mesa-26.0.1
shaderc-2025.5

PR #21011, opened 2026-03-25
convert_hf_to_gguf: add RuGPT3XL (RuGPT3XLForCausalLM) support

EvilFreelancer:
## Overview

Add conversion support for ruGPT3XL (RuGPT3XLForCausalLM), a tiny 1.3B Russian legacy GPT3-like decoder model ([evilfreelancer/ruGPT3XL](https://huggingface.co/evilfreelancer/ruGPT3XL) recovered from [ai-forever/rugpt3xl](https://huggingface.co/ai-forever/rugpt3xl)).

The model is a bit similar to existing GPT-2 implementation, this patch just adds support of custom class which concatenates separate `q_proj`, `k_proj`, `v_proj` weights and biases into a single QKV tensor per layer, as expected by `LLM_ARCH_GPT2`.

## Additional information

- No C++ changes, just python convert script updated, inference uses existing `gpt-2` architecture

- Model card [evilfreelancer/ruGPT3XL](https://huggingface.co/evilfreelancer/ruGPT3XL)

- Conversion scripts [EvilFreelancer/rugpt3xl-convert](https://github.com/EvilFreelancer/rugpt3xl-convert)

# Requirements

- AI usage disclosure: YES, of course, it was used for analyzing large llama.cpp codebase and helping to understand what need to do for adding support of ruGPT3XL.

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -5100,6 +5100,58 @@ class GPT2Model(TextModel):
         yield from super().modify_tensors(data_torch, new_name, bid)
 
 
+@ModelBase.register("RuGPT3XLForCausalLM")
+class RuGPT3XLModel(TextModel):
+    model_arch = gguf.MODEL_ARCH.GPT2
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._qkv_parts: dict[int, dict[str, Tensor]] = {}
+
+    def set_vocab(self):
+        self._set_vocab_gpt2()
+
+    def get_vocab_base_pre(self, tokenizer) -> str:
+        # ruGPT3XL uses GPT-2 byte-level BPE with the same pre-tokenization regex
+        return "gpt-2"
+
+    def set_gguf_parameters(self):
+        self.gguf_writer.add_block_count(self.block_count)
+        self.gguf_writer.add_context_length(self.hparams["max_position_embeddings"])
+        self.gguf_writer.add_embedding_length(self.hparams["hidden_size"])
+        self.gguf_writer.add_feed_forward_length(self.hparams["intermediate_size"])
+        self.gguf_writer.add_head_count(self.hparams["num_attention_heads"])
+        self.gguf_writer.add_head_count_kv(self.hparams["num_attention_heads"])
+        self.gguf_writer.add_layer_norm_eps(self.hparams["layer_norm_eps"])
+        self.gguf_writer.add_file_type(self.ftype)
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        # Fuse separate Q, K, V projections into a single QKV tensor
+        if ".self_attn.q_proj." in name or ".self_attn.k_proj." in name or ".self_attn.v_proj." in name:
+            suffix = "weight" if name.endswith(".weight") else "bias"
+            part = "q" if ".q_proj." in name else ("k" if ".k_proj." in name else "v")
+            key = f"{part}.{suffix}"
+
+            assert bid is not None
+            if bid not in self._qkv_parts:
+                self._qkv_parts[bid] = {}
+            self._qkv_parts[bid][key] = data_torch
+
+            q_key, k_key, v_key = f"q.{suffix}", f"k.{suffix}", f"v.{suffix}"
+            if all(k in self._qkv_parts.get(bid, {}) for k in [q_key, k_key, v_key]):
+                q = self._qkv_parts[bid].pop(q_key)
+                k = self._qkv_parts[bid].pop(k_key)
+                v = self._qkv_parts[bid].pop(v_key)
+                qkv = torch.cat([q, k, v], dim=0)
+                qkv_name = self.format_tensor_name(gguf.MODEL_TENSOR.ATTN_QKV, bid, f".{suffix}")
+                logger.info(f"Fused Q/K/V {suffix} for layer {bid} -> {qkv_name}")
+                yield qkv_name, qkv
+            return
+
+        new_name = self.map_tensor_name(name)
+        yield new_name, data_torch
+
+
 @ModelBase.register("PhiForCausalLM")
 class Phi2Model(TextModel):
     model_arch = gguf.MODEL_ARCH.PHI2
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -5134,6 +5134,15 @@ class RuGPT3XLModel(TextModel):
 
         yield from super().modify_tensors(data_torch, name, bid)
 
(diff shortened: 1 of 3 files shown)


convert_hf_to_gguf.py
```diff
@@ -5100,6 +5100,58 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._qkv_parts: dict[int, dict[str, Tensor]] = {}
+
+    def set_vocab(self):
+        self._set_vocab_gpt2()
+
+    def get_vocab_base_pre(self, tokenizer) -> str:
+        # ruGPT3XL uses GPT-2 byte-level BPE with the same pre-tokenization regex
+        return "gpt-2"
```
```suggestion
```
Add the `chkhsh` to `convert_hf_to_gguf_update.py` instead and run it to update the hashes here.

convert_hf_to_gguf.py
```diff
@@ -5100,6 +5100,58 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+
+    def set_gguf_parameters(self):
+        self.gguf_writer.add_block_count(self.block_count)
+        self.gguf_writer.add_context_length(self.hparams["max_position_embeddings"])
+        self.gguf_writer.add_embedding_length(self.hparams["hidden_size"])
+        self.gguf_writer.add_feed_forward_length(self.hparams["intermediate_size"])
+        self.gguf_writer.add_head_count(self.hparams["num_attention_heads"])
+        self.gguf_writer.add_head_count_kv(self.hparams["num_attention_heads"])
+        self.gguf_writer.add_layer_norm_eps(self.hparams["layer_norm_eps"])
+        self.gguf_writer.add_file_type(self.ftype)
```
```suggestion
```
All of these are handled by default.

convert_hf_to_gguf.py
```diff
@@ -5100,6 +5100,58 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+                k = self._qkv_parts[bid].pop(k_key)
+                v = self._qkv_parts[bid].pop(v_key)
+                qkv = torch.cat([q, k, v], dim=0)
+                qkv_name = self.format_tensor_name(gguf.MODEL_TENSOR.ATTN_QKV, bid, f".{suffix}")
+                logger.info(f"Fused Q/K/V {suffix} for layer {bid} -> {qkv_name}")
+                yield qkv_name, qkv
+            return
+
+        new_name = self.map_tensor_name(name)
+        yield new_name, data_torch
```
```suggestion
                data_torch = torch.cat([q, k, v], dim=0)
                name = self.format_tensor_name(gguf.MODEL_TENSOR.ATTN_QKV, bid, f".{suffix}")
                logger.debug(f"Fused Q/K/V {suffix} for layer {bid} -> {name}")
            else:
                return

        yield from super().modify_tensors(data_torch, name, bid)
```

convert_hf_to_gguf.py
EvilFreelancer: Good point, thanks

EvilFreelancer:
I've added
```python
chkhsh == "0fe1cf6eda062318a1af7270f3331a85c539a01778ff948e24388e949c5282f4"
```

convert_hf_to_gguf.py
```diff
@@ -5100,6 +5100,40 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+            part = "q" if ".q_proj." in name else ("k" if ".k_proj." in name else "v")
+            key = f"{part}.{suffix}"
+
+            assert bid is not None
+            if bid not in self._qkv_parts:
+                self._qkv_parts[bid] = {}
+            self._qkv_parts[bid][key] = data_torch
+
+            q_key, k_key, v_key = f"q.{suffix}", f"k.{suffix}", f"v.{suffix}"
+            if all(k in self._qkv_parts.get(bid, {}) for k in [q_key, k_key, v_key]):
```
```suggestion
    self._qkv_parts: list[dict[str, Tensor]] | None = None

    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
        # Fuse separate Q, K, V projections into a single QKV tensor
        if ".self_attn.q_proj." in name or ".self_attn.k_proj." in name or ".self_attn.v_proj." in name:
            suffix = "weight" if name.endswith(".weight") else "bias"
            part = "q" if ".q_proj." in name else ("k" if ".k_proj." in name else "v")
            key = f"{part}.{suffix}"

            assert bid is not None
            if self._qkv_parts is None:
                self._qkv_parts = [{} for _ in range(self.block_count)]
            self._qkv_parts[bid][key] = data_torch

            q_key, k_key, v_key = f"q.{suffix}", f"k.{suffix}", f"v.{suffix}"
            if all(k in self._qkv_parts[bid] for k in [q_key, k_key, v_key]):
```

convert_hf_to_gguf.py
```diff
@@ -5100,6 +5100,40 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+                q = self._qkv_parts[bid].pop(q_key)
+                k = self._qkv_parts[bid].pop(k_key)
+                v = self._qkv_parts[bid].pop(v_key)
+                data_torch = torch.cat([q, k, v], dim=0)
+                name = self.format_tensor_name(gguf.MODEL_TENSOR.ATTN_QKV, bid, f".{suffix}")
+                logger.debug(f"Fused Q/K/V {suffix} for layer {bid} -> {name}")
+            else:
+                return
+
+        yield from super().modify_tensors(data_torch, name, bid)
```
CISC, line 5135:
```suggestion
        yield from super().modify_tensors(data_torch, name, bid)

    def prepare_tensors(self):
        super().prepare_tensors()

        if self._qkv_parts is not None:
            # flatten `list[dict[str, Tensor]]` into `list[str]`
            parts = [f"({i}){k}" for i, d in enumerate(self._qkv_parts) for k in d.keys()]
            if len(parts) > 0:
                raise ValueError(f"Unprocessed Q/K/V parts: {parts}")
```

CISC:
> I've added

Great, but not quite right
- You have to run `convert_hf_to_gguf_update.py`, it will place the `chkhsh` in the right spot in `convert_hf_to_gguf.py`
- You can't add a new name without also handling it in `llama_vocab.cpp`, however:
- `precomputed_hashes` is for duplicate (or complicated) pre-tokenizers, not new ones, but since this is just `GPT2Tokenizer` change the name to `gpt-2`

EvilFreelancer:
@CISC thanks, i've did what you recommend, plus i've found and fixed a small typo with `self._qkv_parts` in `RuGPT3XLModel` after merging all review changes.

ngxson (approved):

CISC (approved):

PR #21027, opened 2026-03-26
fix: mtmd "v.patch_embd" quant and unsupported im2col ops on Metal for deepseek-ocr

sfallah:
## Bug Fixes

This PR fixes two issues affecting vision models:

1. **Quantization of `v.patch_embd`**
2. **Unsupported `im2col` (bf16) ops on Metal for DeepSeek-OCR**

---

### Vision Models — `v.patch_embd` Quantization

**Issue**
`v.patch_embd` was incorrectly quantized, causing inference failure when quantizing vision (mmproj) model.

**Fix**
Exclude `v.patch_embd` from quantization across all vision models.

**Validated on**

* `moonshotai/Kimi-VL-A3B-Instruct`
* `OpenGVLab/InternVL3-2B`
* `deepseek-ai/DeepSeek-OCR`

---

### DeepSeek-OCR — Metal bf16 `im2col`

**Issue**
Metal backend does not support `im2col` for bf16, causing `ggml_conv_2d` failures in DeepSeek-OCR vision-model.

**Fix**
Avoid unsupported bf16 `im2col` paths on Metal by using f32 type in conversion and exclude the involved tensors from quantization.

---

## Summary

* Fixes a systemic vision quantization issue (`v.patch_embd`)
* Restores DeepSeek-OCR execution on Metal (bf16)

### DIFF
--- a/src/llama-quant.cpp
+++ b/src/llama-quant.cpp
@@ -345,9 +345,11 @@ static bool tensor_allows_quantization(const llama_model_quantize_params * param
 
     // do not quantize specific multimodal tensors
     quantize &= name.find(".position_embd") == std::string::npos;
-    quantize &= name.find("sam.patch_embd") == std::string::npos;
     quantize &= name.find("sam.pos_embd")   == std::string::npos;
+    quantize &= name.find("sam.neck.")      == std::string::npos;
+    quantize &= name.find("sam.net_")       == std::string::npos;
     quantize &= name.find(".rel_pos")       == std::string::npos;
+    quantize &= name.find(".patch_embd")    == std::string::npos;
 
     return quantize;
 }
--- a/src/llama-quant.cpp
+++ b/src/llama-quant.cpp
@@ -350,6 +350,7 @@ static bool tensor_allows_quantization(const llama_model_quantize_params * param
     quantize &= name.find("sam.net_")       == std::string::npos;
     quantize &= name.find(".rel_pos")       == std::string::npos;
     quantize &= name.find(".patch_embd")    == std::string::npos;
+    quantize &= name.find(".patch_merger")  == std::string::npos;
 
     return quantize;
 }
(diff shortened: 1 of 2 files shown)


src/llama-quant.cpp
```diff
@@ -345,9 +345,11 @@ static bool tensor_allows_quantization(const llama_model_quantize_params * param
 
     // do not quantize specific multimodal tensors
     quantize &= name.find(".position_embd") == std::string::npos;
-    quantize &= name.find("sam.patch_embd") == std::string::npos;
     quantize &= name.find("sam.pos_embd")   == std::string::npos;
+    quantize &= name.find("sam.neck.")      == std::string::npos;
+    quantize &= name.find("sam.net_")       == std::string::npos;
     quantize &= name.find(".rel_pos")       == std::string::npos;
+    quantize &= name.find(".patch_embd")    == std::string::npos;
```
CISC, line 352:
```suggestion
    quantize &= name.find(".patch_embd")    == std::string::npos;
    quantize &= name.find(".patch_merger")  == std::string::npos;
```

CISC (approved):
I could have sworn it was already doing this at some point, ohwell...

src/llama-quant.cpp
```diff
@@ -345,9 +345,12 @@ static bool tensor_allows_quantization(const llama_model_quantize_params * param
 
     // do not quantize specific multimodal tensors
     quantize &= name.find(".position_embd") == std::string::npos;
-    quantize &= name.find("sam.patch_embd") == std::string::npos;
     quantize &= name.find("sam.pos_embd")   == std::string::npos;
```
CISC, line 348: BTW, `sam.pos_embd` is not covered here (or anywhere else I could find):
https://github.com/ggml-org/llama.cpp/blob/3b7b6f3a50ed2486f6b10f592b9d92a669604eb0/convert_hf_to_gguf.py#L6986-L6993

sfallah: it is covered by 'pos_embed'
or what do you mean?

CISC: Look closer. :)

sfallah: Sorry, but I don't get the point!
'pos_embed' (or full-name "model.sam_model.pos_embed") is the origin tensor-name.
And in GGUF is call "sam.pos_embd".

CISC: Ah yes, you are right, sorry. :)

ngxson (approved):

PR #21033, opened 2026-03-26
ci: pin external actions to exact commit SHA

ngxson:
## Overview

Pin actions to full commit SHA, except for:
- Official actions
- `ggml-org` actions

# Requirements

- AI usage disclosure: YES, it is a repetitive task, so I used AI. But I manually validated the output

### DIFF
--- a/.github/workflows/editorconfig.yml
+++ b/.github/workflows/editorconfig.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-slim
     steps:
       - uses: actions/checkout@v6
-      - uses: editorconfig-checker/action-editorconfig-checker@v2
+      - uses: editorconfig-checker/action-editorconfig-checker@840e866d93b8e032123c23bac69dece044d4d84c # v2.2.0
         with:
           version: v3.0.3
       - run: editorconfig-checker
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -51,7 +51,7 @@ jobs:
           distribution: zulu
 
       - name: Setup Android SDK
-        uses: android-actions/setup-android@v3
+        uses: android-actions/setup-android@9fc6c4e9069bf8d3d10b2204b1fb8f6ef7065407 # v3
         with:
           log-accepted-android-sdk-licenses: false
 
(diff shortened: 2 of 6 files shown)


.github/workflows/editorconfig.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-slim
     steps:
       - uses: actions/checkout@v6
-      - uses: editorconfig-checker/action-editorconfig-checker@v2
+      - uses: editorconfig-checker/action-editorconfig-checker@840e866d93b8e032123c23bac69dece044d4d84c # v2.2.0
```
CISC, line 26:
```suggestion
      - uses: editorconfig-checker/action-editorconfig-checker@840e866d93b8e032123c23bac69dece044d4d84c # v2
```
We should comment what the old tag was rather than what the new hash points to.

CISC (approved):

.github/workflows/editorconfig.yml
ngxson:
```suggestion
      - uses: editorconfig-checker/action-editorconfig-checker@d2ed4fd072ae6f887e9407c909af0f585d2ad9f4 # v2
```

In this case, do you think it's better to also pin to the exact commit of `v2` tag? https://github.com/editorconfig-checker/action-editorconfig-checker/releases/tag/v2

CISC: No, the latest at this time is fine (and v2.2.0 upgrades this to Node 24, which is required soon).

ngxson: In that case, I think keeping `v2.2.0` will be more clear, right? Release tag: https://github.com/editorconfig-checker/action-editorconfig-checker/releases/tag/v2.2.0 with commit https://github.com/editorconfig-checker/action-editorconfig-checker/commit/840e866d93b8e032123c23bac69dece044d4d84c

Changing it to just `v2` means we use the `v2` tag, unfortunately it doesn't mean `v2.*`

CISC: Ah, ok, I just assumed it was like the other tags that gets bumped.

ggerganov (approved):

PR #21034, opened 2026-03-26
webui: Add option to pre-encode conversation for faster next turns

### LINKED ISSUE #18853: Feature Request: Add UI Toggle to Auto‑Remove Chain‑of‑Thought and Pre‑Submit Conversation History for Faster Next‑Turn Encoding
### Prerequisites

### Feature Description

Functional Requirement:

Please add a toggle option to the web UI. When this option is enabled, after the large language model (LLM) finishes responding to a user request, the web interface should automatically delete any “thinking” (chain‑of‑thought) content if the model supports such reasoning, and then submit the full conversation history back to the LLM (so the model has the entire context pre‑encoded).

Consequences of this behavior:

    When the user sends the next query, the LLM does not need to spend time re‑encoding the deleted reasoning steps or the previous generation; it can directly encode the new user prompt.
    This allows the system to pre‑process the context needed for the upcoming turn while the user is reading or thinking about the current response, resulting in a smoother, more responsive interaction.

### Motivation

When I receive a very large context from the model (e.g., a code snippet of more than 8 000 tokens) and then submit a simple follow‑up question, I have to wait a long time for the prompt‑processing step. While I am reading the returned context, the system is idle. The current LLM backend supports caching of multiple statements, so performing this pre‑processing can significantly improve the smoothness of subsequent turns in the conversation.

### Possible Implementation

_No response_

allozaur:
Close #18853


ggerganov:
Does this have logic that all current slots are inactive before sending the pre-encode? If not, it might be good to add it.

ServeurpersoCom (approved):

ggerganov (approved):
I wanted to test a for a bit, but I see that the option is opt-in. So if the regular usage (without the option) is not affected, I think it is OK to merge and I will be testing over the next days from `master`.

allozaur:
> I wanted to test a for a bit, but I see that the option is opt-in. So if the regular usage (without the option) is not affected, I think it is OK to merge and I will be testing over the next days from `master`.

Sure, lemme know whenever u need anything during the tests

ggerganov (approved):

Issue #6444, opened 2024-04-02
Support QuaRot quantization scheme


EwoutH:
A new, interesting quantization scheme was published, which not only reduces memory consumption (like current quantization schemes), but als reduces computations.

> **[QuaRot: Outlier-Free 4-Bit Inference in Rotated LLMs](https://arxiv.org/abs/2404.00456)**
> We introduce QuaRot, a new Quantization scheme based on Rotations, which is able to quantize LLMs end-to-end, including all weights, activations, and KV cache in 4 bits. QuaRot rotates LLMs in a way that removes outliers from the hidden state without changing the output, making quantization easier. This computational invariance is applied to the hidden state (residual) of the LLM, as well as to the activations of the feed-forward components, aspects of the attention mechanism and to the KV cache. The result is a quantized model where all matrix multiplications are performed in 4-bits, without any channels identified for retention in higher precision. Our quantized LLaMa2-70B model has losses of at most 0.29 WikiText-2 perplexity and retains 99% of the zero-shot performance. Code is available at: [this https URL](https://github.com/spcl/QuaRot).

I think it would be interesting to see if this technique, or parts of it, could be adopted in llama.cpp, to speed up inference of quantized models.

ggerganov:
It's an interesting approach that we should explore. As far as I understood, the model weights are pre-processed (rotated) and then the inference is augmented with extra (two?) operations to restore the effects from the rotation. We can start with implementing the latter and try to use existing QuaRot models to evaluate

PinkysBrain:
Similar to "Training Transformers with 4-bit Integers", except that only used Hadamard. Hadamard alone might be enough, QuaRot did do an ablation for Q alone, but not Hadamard alone.

sashkboos:
@ggerganov 

Thanks for your interest in our work. I am the main author of QuaRot. 

I would be happy to discuss/plan for this and help to integrate it into the repo. I think the general steps of QuaRot are:

1. First you need to convert your model into an RMNorm model. This means that you fuse all other operations except normalization (divide by norm) into the previous and next linear weights and use [this](https://github.com/spcl/QuaRot/blob/d538b8ef8127fe199d3fc5e43bbfa696e637ba92/quarot/nn/normalization.py#L3) class as the normalization module.
[...]

PR #21038, opened 2026-03-26
llama : rotate activations for better quantization

ggerganov:
## Overview

In anticipation of the incoming flood of vibe generated PRs implementing [TurboQuant](https://research.google/blog/turboquant-redefining-ai-efficiency-with-extreme-compression/), I'm raising the baseline a bit using a very simple interpretation of the idea of using Hadamard transform to reduce outliers in the attention and improve the quantization quality:

- Rotate input Q, K, V and store in cache
- Perform attention with the rotated Q, K, V
- Rotate back the output vector of the attention

This works because:
  - Rotation does not affect the dot product of 2 vectors
  - The output vector is a linear combination of rotated vectors

The implementation is very simple and backend agnostic - use Hadamard matrix of size `n x n`, normalized by `1/sqrt(n)` so that it is orthonormal and can be used both for the forward and backward rotation. Technically any rotation matrix (and it's inverse) should work - I just think this is what is commonly used due to it's simplicity. The implementation does not introduce new types and is compatible with all existing quantizations. It adds 4 matrix multiplication operators in the attention, though I think some of them can be fused in the attention weights (similar to [QuaRot](https://arxiv.org/abs/2404.00456)).

I don't know what is the impact of the remaining techniques explained in TurboQuant (PolarQuant, QJL, etc.). They could be important and can potentially improve further on top of this. In any case, having a better baseline at almost 0 cost won't hurt. Only based on the PPL data below, I think this should never be worse that before, though it needs a bit more evaluation.

Note: MLA is not supported

## Additional information

Here are some PPL results before and after using base models. Ideally, there should be KLD data too, but leaving it for people to play with it and see if it looks good.

```bash
./bin/llama-perplexity -m model.gguf -f wiki.test.raw --chunks 128 -ctk f16 -ctv f16
```

### Model: Qwen3 0.6B BF16
https://huggingface.co/Qwen/Qwen3-0.6B-Base

Baseline F16 cache: PPL = 13.6711 +/- 0.21422

| type | master   | PR      |
| ---  | ---      | ---     |
| q8_0 | 13.9115  | 13.6713 |
| q5_1 | 61.6992  | 14.1452 |
| q5_0 | 17.2805  | 14.2171 |
| q4_1 | 212.479  | 22.2816 |
| q4_0 | 62.0161  | 46.2503 |

### Model: Qwen3 8B BF16
https://huggingface.co/Qwen/Qwen3-8B-Base

Baseline F16 cache: PPL = 7.3203 +/- 0.09901

| type | master  | PR     |
| ---  | ---     | ---    |
| q8_0 | 7.3172  | 7.3195 |
| q5_0 | 7.3793  | 7.3323 |
| q4_0 | 7.6451  | 7.5012 |

### Model: Gemma3 4B Q8_0
https://huggingface.co/google/gemma-3-4b-pt

Baseline F16 cache: PPL = 7.6905 +/- 0.10483

| type | master  | PR     |
| ---  | ---     | ---    |
| q8_0 | 7.6928  | 7.6914 |
| q5_1 | 7.7133  | 7.7052 |
| q5_0 | 7.7544  | 7.7182 |
| q4_1 | 7.8095  | 7.7531 |
| q4_0 | 7.8535  | 7.7928 |

### Model: Qwen3.5 4B F16
https://huggingface.co/Qwen/Qwen3.5-4B-Base

Baseline F16 cache: PPL = 8.3266 +/- 0.11623

| type | master   | PR     |
| ---  | ---      | ---    |
| q8_0 | 8.3272   | 8.3261 |
| q5_0 | 8.3359   | 8.3339 |
| q4_0 | 8.3475   | 8.3448 |

### TODOs

- [x] Cache shift support https://github.com/ggml-org/llama.cpp/commit/ff76c6731d9a72b7ea1dbc034020c70330a37ff8
- [x] Disable rotations with env variable (`LLAMA_ATTN_ROT_DISABLE`)

### Next PRs

- Add randomized Hadamard matrices (https://github.com/ggml-org/llama.cpp/commit/f0fea264b0118fe0c84d92cbfff4508ca2792c26)

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -1004,6 +1004,14 @@ bool llama_kv_cache::get_has_shift() const {
     return result;
 }
 
+ggml_type llama_kv_cache::type_k() const {
+    return layers[0].k->type;
+}
+
+ggml_type llama_kv_cache::type_v() const {
+    return layers[0].v->type;
+}
+
 uint32_t llama_kv_cache::get_n_kv(const slot_info & sinfo) const {
     uint32_t result = 0;
 
@@ -2237,6 +2245,14 @@ const llama_ubatch & llama_kv_cache_context::get_ubatch() const {
 
 uint32_t llama_kv_cache_context::get_n_kv() const {
     return n_kv;
+}
+
+ggml_type llama_kv_cache_context::type_k() const {
+    return kv->type_k();
+}
+
+ggml_type llama_kv_cache_context::type_v() const {
+    return kv->type_v();
 }
 
 ggml_tensor * llama_kv_cache_context::get_k(ggml_context * ctx, int32_t il) const {
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -13,6 +13,53 @@
 #include <map>
 #include <stdexcept>
 
+static bool ggml_is_power_of_2(int n) {
+    return (n & (n - 1)) == 0;
+}
+
+// orthonormal Walsh-Hadamard rotation matrix
+static void ggml_gen_hadamard(ggml_tensor * tensor) {
+    assert(tensor->type == GGML_TYPE_F32);
+
+    const int n = tensor->ne[0];
+
+    assert(ggml_is_power_of_2(n));
+    assert(tensor->ne[1] == n);
+    assert(tensor->ne[2] == 1);
+    assert(tensor->ne[3] == 1);
+
+    auto * data = (float *) tensor->data;
+
+    data[0*n + 0] = 1.0 / sqrtf(n);
+
+    for (int s = 1; s < n; s *= 2) {
+        for (int i = 0; i < s; i++) {
+            for (int j = 0; j < s; j++) {
+                const float val = data[i*n + j];
+
+                data[(i + s)*n + (j    )] =  val;
+                data[(i    )*n + (j + s)] =  val;
+                data[(i + s)*n + (j + s)] = -val;
+            }
+        }
+    }
+}
+
+static ggml_tensor * ggml_mul_mat_aux(
+        ggml_context * ctx,
+        ggml_tensor * cur,
+        ggml_tensor * rot) {
+    const auto n = rot->ne[0];
+
+    ggml_tensor * res;
+
+    res = ggml_reshape_2d(ctx, cur, n, ggml_nelements(cur)/n);
+    res = ggml_mul_mat   (ctx, rot, res);
+    res = ggml_reshape_4d(ctx, res, cur->ne[0], cur->ne[1], cur->ne[2], cur->ne[3]);
+
+    return res;
+}
+
 //
 // llama_kv_cache
 //
@@ -1515,6 +1562,18 @@ void llama_kv_cache::set_input_pos_bucket(ggml_tensor * dst, const llama_ubatch 
     }
 }
 
+void llama_kv_cache::set_input_k_rot(ggml_tensor * dst) const {
+    GGML_ASSERT(ggml_backend_buffer_is_host(dst->buffer));
+
+    ggml_gen_hadamard(dst);
+}
+
+void llama_kv_cache::set_input_v_rot(ggml_tensor * dst) const {
+    GGML_ASSERT(ggml_backend_buffer_is_host(dst->buffer));
+
+    ggml_gen_hadamard(dst);
+}
+
 size_t llama_kv_cache::total_size() const {
     size_t size = 0;
 
@@ -1550,6 +1609,7 @@ ggml_tensor * llama_kv_cache::build_rope_shift(
                ggml_context * ctx,
                 ggml_tensor * cur,
                 ggml_tensor * shift,
+                ggml_tensor * rot,
                 ggml_tensor * factors,
                       float   freq_base,
                       float   freq_scale,
@@ -1575,9 +1635,15 @@ ggml_tensor * llama_kv_cache::build_rope_shift(
         // dequantize to f32 -> RoPE -> quantize back
         tmp = ggml_cast(ctx, cur, GGML_TYPE_F32);
 
+        // rotate back
+        tmp = ggml_mul_mat_aux(ctx, tmp, rot);
+
         tmp = ggml_rope_ext(ctx, tmp,
                 shift, factors, n_rot, rope_type, n_ctx_orig, freq_base, freq_scale,
                 yarn_ext_factor, yarn_attn_factor, yarn_beta_fast, yarn_beta_slow);
+
+        // rotate fwd
+        tmp = ggml_mul_mat_aux(ctx, tmp, rot);
 
         tmp = ggml_cpy(ctx, tmp, cur);
     } else {
@@ -1599,6 +1665,8 @@ public:
 
     ggml_tensor * k_shift; // I32 [kv_size*n_stream]
 
+    ggml_tensor * k_rot = nullptr;
+
     const llama_kv_cache * kv_self;
 };
 
@@ -1607,6 +1675,10 @@ void llm_graph_input_k_shift::set_input(const llama_ubatch * ubatch) {
 
     if (k_shift) {
         kv_self->set_input_k_shift(k_shift);
+    }
+
+    if (k_rot) {
+        kv_self->set_input_k_rot(k_rot);
     }
 }
 
@@ -1618,6 +1690,21 @@ ggml_cgraph * llama_kv_cache::build_graph_shift(llm_graph_result * res, llama_co
 
     inp->k_shift = ggml_new_tensor_1d(ctx, GGML_TYPE_I32, (int64_t) get_size()*n_stream);
     ggml_set_input(inp->k_shift);
+
+    if (ggml_is_quantized(type_k())) {
+        int nrot = 64;
+
+        // TODO: investigate if using the smallest rotation matrix is beneficial also for K (similar as for V)
+        // ref: https://github.com/ggml-org/llama.cpp/pull/21038#issuecomment-4141323088
+        do {
+            nrot *= 2;
+        } while (hparams.n_embd_head_k() % nrot == 0);
+        nrot /= 2;
+
+        inp->k_rot = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, nrot, nrot);
+        ggml_set_input(inp->k_rot);
+        ggml_set_name(inp->k_rot, "rope_shift_k_rot");
+    }
 
     const auto & cparams = lctx->get_cparams();
 
@@ -1643,7 +1730,7 @@ ggml_cgraph * llama_kv_cache::build_graph_shift(llm_graph_result * res, llama_co
                 ggml_row_size(layer.k->type, n_embd_k_gqa),
                 ggml_row_size(layer.k->type, n_embd_nope));
 
-        ggml_tensor * cur = build_rope_shift(cparams, ctx, k, inp->k_shift, rope_factors, freq_base_l, freq_scale_l, il);
+        ggml_tensor * cur = build_rope_shift(cparams, ctx, k, inp->k_shift, inp->k_rot, rope_factors, freq_base_l, freq_scale_l, il);
 
         ggml_build_forward_expand(gf, cur);
     }
@@ -2297,4 +2384,12 @@ void llama_kv_cache_context::set_input_kq_mask(ggml_tensor * dst, const llama_ub
 
 void llama_kv_cache_context::set_input_pos_bucket(ggml_tensor * dst, const llama_ubatch * ubatch) const {
     kv->set_input_pos_bucket(dst, ubatch);
+}
+
+void llama_kv_cache_context::set_input_k_rot(ggml_tensor * dst) const {
+    kv->set_input_k_rot(dst);
+}
+
+void llama_kv_cache_context::set_input_v_rot(ggml_tensor * dst) const {
+    kv->set_input_v_rot(dst);
 }
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -18,6 +18,7 @@ static bool ggml_is_power_of_2(int n) {
 }
 
 // orthonormal Walsh-Hadamard rotation matrix
+// note: res^2 == I
 static void ggml_gen_hadamard(ggml_tensor * tensor) {
     assert(tensor->type == GGML_TYPE_F32);
 
@@ -1244,6 +1245,57 @@ ggml_tensor * llama_kv_cache::build_input_v_idxs(ggml_context * ctx, const llama
     return v_idxs;
 }
 
+ggml_tensor * llama_kv_cache::build_input_k_rot(ggml_context * ctx) const {
+    ggml_tensor * res = nullptr;
+
+    const bool can_k_rot =
+        ggml_is_quantized(type_k()) &&
+        !hparams.is_n_embd_k_gqa_variable() &&
+        hparams.n_embd_head_k() % 64 == 0;
+
+    if (can_k_rot) {
+        int nrot = 64;
+
+        // TODO: investigate if using the smallest rotation matrix is beneficial also for K (similar as for V)
+        // ref: https://github.com/ggml-org/llama.cpp/pull/21038#issuecomment-4141323088
+        do {
+            nrot *= 2;
+        } while (hparams.n_embd_head_k() % nrot == 0);
+        nrot /= 2;
+
+        res = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, nrot, nrot);
+        ggml_set_input(res);
+        ggml_set_name(res, "attn_inp_k_rot");
+    }
+
+    return res;
+}
+
+ggml_tensor * llama_kv_cache::build_input_v_rot(ggml_context * ctx) const {
+    ggml_tensor * res = nullptr;
+
+    const bool can_v_rot =
+        ggml_is_quantized(type_v()) &&
+        !hparams.is_n_embd_v_gqa_variable() &&
+        hparams.n_embd_head_v() % 64 == 0;
+
+    if (can_v_rot) {
+        int nrot = 64;
+        // using smaller rotation matrices for V seems beneficial
+        // ref: https://github.com/ggml-org/llama.cpp/pull/21038#issuecomment-4146397570
+        //do {
+        //    nrot *= 2;
+        //} while (hparams.n_embd_head_v() % nrot == 0);
+        //nrot /= 2;
+
+        res = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, nrot, nrot);
+        ggml_set_input(res);
+        ggml_set_name(res, "attn_inp_v_rot");
+    }
+
+    return res;
+}
+
 void llama_kv_cache::set_input_k_idxs(ggml_tensor * dst, const llama_ubatch * ubatch, const slot_info & sinfo) const {
     const uint32_t n_tokens = ubatch->n_tokens;
     GGML_ASSERT(n_tokens == (int64_t) sinfo.size()*sinfo.n_stream());
@@ -1665,6 +1717,7 @@ public:
 
     ggml_tensor * k_shift; // I32 [kv_size*n_stream]
 
+    // note: assumes k_rot^2 == I
     ggml_tensor * k_rot = nullptr;
 
     const llama_kv_cache * kv_self;
@@ -2364,6 +2417,14 @@ ggml_tensor * llama_kv_cache_context::build_input_k_idxs(ggml_context * ctx, con
 
 ggml_tensor * llama_kv_cache_context::build_input_v_idxs(ggml_context * ctx, const llama_ubatch & ubatch) const {
     return kv->build_input_v_idxs(ctx, ubatch);
+}
+
+ggml_tensor * llama_kv_cache_context::build_input_k_rot(ggml_context * ctx) const {
+    return kv->build_input_k_rot(ctx);
+}
+
+ggml_tensor * llama_kv_cache_context::build_input_v_rot(ggml_context * ctx) const {
+    return kv->build_input_v_rot(ctx);
 }
 
 void llama_kv_cache_context::set_input_k_shift(ggml_tensor * dst) const {
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -29,7 +29,14 @@ static void ggml_gen_hadamard(ggml_tensor * tensor) {
     assert(tensor->ne[2] == 1);
     assert(tensor->ne[3] == 1);
 
-    auto * data = (float *) tensor->data;
+    std::vector<float> data_f32;
+
+    float * data = (float *) tensor->data;
+
+    if (tensor->type != GGML_TYPE_F32) {
+        data_f32.resize(n*n);
+        data = data_f32.data();
+    }
 
     data[0*n + 0] = 1.0 / sqrtf(n);
 
@@ -43,6 +50,10 @@ static void ggml_gen_hadamard(ggml_tensor * tensor) {
                 data[(i + s)*n + (j + s)] = -val;
             }
         }
+    }
+
+    if (tensor->type != GGML_TYPE_F32) {
+        ggml_quantize_chunk(tensor->type, data, tensor->data, 0, 1, n*n, nullptr);
     }
 }
 
@@ -1744,20 +1755,7 @@ ggml_cgraph * llama_kv_cache::build_graph_shift(llm_graph_result * res, llama_co
     inp->k_shift = ggml_new_tensor_1d(ctx, GGML_TYPE_I32, (int64_t) get_size()*n_stream);
     ggml_set_input(inp->k_shift);
 
-    if (ggml_is_quantized(type_k())) {
-        int nrot = 64;
-
-        // TODO: investigate if using the smallest rotation matrix is beneficial also for K (similar as for V)
-        // ref: https://github.com/ggml-org/llama.cpp/pull/21038#issuecomment-4141323088
-        do {
-            nrot *= 2;
-        } while (hparams.n_embd_head_k() % nrot == 0);
-        nrot /= 2;
-
-        inp->k_rot = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, nrot, nrot);
-        ggml_set_input(inp->k_rot);
-        ggml_set_name(inp->k_rot, "rope_shift_k_rot");
-    }
+    inp->k_rot = build_input_k_rot(ctx);
 
     const auto & cparams = lctx->get_cparams();
 
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -268,11 +268,26 @@ llama_kv_cache::llama_kv_cache(
                 ggml_type_name(type_v), (float)memory_size_v / (1024.0f * 1024.0f));
     }
 
-    const char * LLAMA_KV_CACHE_ATTN_ROT_DISABLE = getenv("LLAMA_KV_CACHE_ATTN_ROT_DISABLE");
-    attn_rot_disable = LLAMA_KV_CACHE_ATTN_ROT_DISABLE ? atoi(LLAMA_KV_CACHE_ATTN_ROT_DISABLE) : false;
+    const char * LLAMA_ATTN_ROT_DISABLE = getenv("LLAMA_ATTN_ROT_DISABLE");
+    const bool attn_rot_disable = LLAMA_ATTN_ROT_DISABLE ? atoi(LLAMA_ATTN_ROT_DISABLE) : false;
     if (attn_rot_disable) {
-        LLAMA_LOG_WARN("%s: attention rotation disabled\n", __func__);
+        LLAMA_LOG_WARN("%s: attention rotation force disabled (LLAMA_ATTN_ROT_DISABLE)\n", __func__);
     }
+
+    attn_rot_k =
+        !attn_rot_disable &&
+        ggml_is_quantized(type_k) &&
+        !hparams.is_n_embd_k_gqa_variable() &&
+        hparams.n_embd_head_k() % 64 == 0;
+
+    attn_rot_v =
+        !attn_rot_disable &&
+        ggml_is_quantized(type_v) &&
+        !hparams.is_n_embd_v_gqa_variable() &&
+        hparams.n_embd_head_v() % 64 == 0;
+
+    LLAMA_LOG_INFO("%s: attn_rot_k = %d\n", __func__, attn_rot_k);
+    LLAMA_LOG_INFO("%s: attn_rot_v = %d\n", __func__, attn_rot_v);
 
     const char * LLAMA_KV_CACHE_DEBUG = getenv("LLAMA_KV_CACHE_DEBUG");
     debug = LLAMA_KV_CACHE_DEBUG ? atoi(LLAMA_KV_CACHE_DEBUG) : 0;
@@ -1265,13 +1280,7 @@ ggml_tensor * llama_kv_cache::build_input_v_idxs(ggml_context * ctx, const llama
 ggml_tensor * llama_kv_cache::build_input_k_rot(ggml_context * ctx) const {
     ggml_tensor * res = nullptr;
 
-    const bool can_k_rot =
-        !attn_rot_disable &&
-        ggml_is_quantized(type_k()) &&
-        !hparams.is_n_embd_k_gqa_variable() &&
-        hparams.n_embd_head_k() % 64 == 0;
-
-    if (can_k_rot) {
+    if (attn_rot_k) {
         int nrot = 64;
 
         // TODO: investigate if using the smallest rotation matrix is beneficial also for K (similar as for V)
@@ -1292,13 +1301,7 @@ ggml_tensor * llama_kv_cache::build_input_k_rot(ggml_context * ctx) const {
 ggml_tensor * llama_kv_cache::build_input_v_rot(ggml_context * ctx) const {
     ggml_tensor * res = nullptr;
 
-    const bool can_v_rot =
-        !attn_rot_disable &&
-        ggml_is_quantized(type_v()) &&
-        !hparams.is_n_embd_v_gqa_variable() &&
-        hparams.n_embd_head_v() % 64 == 0;
-
-    if (can_v_rot) {
+    if (attn_rot_v) {
         int nrot = 64;
         // using smaller rotation matrices for V seems beneficial
         // ref: https://github.com/ggml-org/llama.cpp/pull/21038#issuecomment-4146397570
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -268,6 +268,12 @@ llama_kv_cache::llama_kv_cache(
                 ggml_type_name(type_v), (float)memory_size_v / (1024.0f * 1024.0f));
     }
 
+    const char * LLAMA_KV_CACHE_ATTN_ROT_DISABLE = getenv("LLAMA_KV_CACHE_ATTN_ROT_DISABLE");
+    attn_rot_disable = LLAMA_KV_CACHE_ATTN_ROT_DISABLE ? atoi(LLAMA_KV_CACHE_ATTN_ROT_DISABLE) : false;
+    if (attn_rot_disable) {
+        LLAMA_LOG_WARN("%s: attention rotation disabled\n", __func__);
+    }
+
     const char * LLAMA_KV_CACHE_DEBUG = getenv("LLAMA_KV_CACHE_DEBUG");
     debug = LLAMA_KV_CACHE_DEBUG ? atoi(LLAMA_KV_CACHE_DEBUG) : 0;
 }
@@ -1260,6 +1266,7 @@ ggml_tensor * llama_kv_cache::build_input_k_rot(ggml_context * ctx) const {
     ggml_tensor * res = nullptr;
 
     const bool can_k_rot =
+        !attn_rot_disable &&
         ggml_is_quantized(type_k()) &&
         !hparams.is_n_embd_k_gqa_variable() &&
         hparams.n_embd_head_k() % 64 == 0;
@@ -1286,6 +1293,7 @@ ggml_tensor * llama_kv_cache::build_input_v_rot(ggml_context * ctx) const {
     ggml_tensor * res = nullptr;
 
     const bool can_v_rot =
+        !attn_rot_disable &&
         ggml_is_quantized(type_v()) &&
         !hparams.is_n_embd_v_gqa_variable() &&
         hparams.n_embd_head_v() % 64 == 0;
--- a/src/llama-kv-cache.cpp
+++ b/src/llama-kv-cache.cpp
@@ -289,6 +289,27 @@ llama_kv_cache::llama_kv_cache(
     LLAMA_LOG_INFO("%s: attn_rot_k = %d\n", __func__, attn_rot_k);
     LLAMA_LOG_INFO("%s: attn_rot_v = %d\n", __func__, attn_rot_v);
 
+    // pre-compute the haramard matrices and keep them in host memory
+    // TODO: in the future, we can make copies in the backend buffers to avoid host -> device transfers
+    if (attn_rot_k || attn_rot_v) {
+        for (int64_t n = 64; n <= std::max(hparams.n_embd_head_k(), hparams.n_embd_head_v()); n *= 2) {
+            attn_rot_hadamard[n] = std::vector<float>(n*n);
+
+            ggml_init_params params = {
+                /* .mem_size   = */ 1*ggml_tensor_overhead(),
+                /* .mem_buffer = */ nullptr,
+                /* .no_alloc   = */ true,
+            };
+
+            ggml_context_ptr ctx { ggml_init(params) };
+
+            ggml_tensor * tmp = ggml_new_tensor_2d(ctx.get(), GGML_TYPE_F32, n, n);
+            tmp->data = attn_rot_hadamard[n].data();
+
+            ggml_gen_hadamard(tmp);
+        }
+    }
+
     const char * LLAMA_KV_CACHE_DEBUG = getenv("LLAMA_KV_CACHE_DEBUG");
     debug = LLAMA_KV_CACHE_DEBUG ? atoi(LLAMA_KV_CACHE_DEBUG) : 0;
 }
@@ -1639,13 +1660,19 @@ void llama_kv_cache::set_input_pos_bucket(ggml_tensor * dst, const llama_ubatch 
 void llama_kv_cache::set_input_k_rot(ggml_tensor * dst) const {
     GGML_ASSERT(ggml_backend_buffer_is_host(dst->buffer));
 
-    ggml_gen_hadamard(dst);
+    const auto n_rot = dst->ne[0];
+    GGML_ASSERT(attn_rot_hadamard.count(dst->ne[0]));
+
+    memcpy(dst->data, attn_rot_hadamard.at(n_rot).data(), ggml_nbytes(dst));
 }
 
 void llama_kv_cache::set_input_v_rot(ggml_tensor * dst) const {
     GGML_ASSERT(ggml_backend_buffer_is_host(dst->buffer));
 
-    ggml_gen_hadamard(dst);
+    const auto n_rot = dst->ne[0];
+    GGML_ASSERT(attn_rot_hadamard.count(dst->ne[0]));
+
+    memcpy(dst->data, attn_rot_hadamard.at(n_rot).data(), ggml_nbytes(dst));
 }
 
 size_t llama_kv_cache::total_size() const {
--- a/src/llama-graph.cpp
+++ b/src/llama-graph.cpp
@@ -52,6 +52,29 @@ static bool can_reuse_kq_mask(
 
 // impl
 
+static bool is_power_of_2(int n) {
+    return (n & (n - 1)) == 0;
+}
+
+// orthonormal Walsh-Hadamard rotation matrix
+static void set_input_hadamard(int n, float * data) {
+    assert(is_power_of_2(n));
+
+    data[0*n + 0] = 1.0 / sqrtf(n);
+
+    for (int s = 1; s < n; s *= 2) {
+        for (int i = 0; i < s; i++) {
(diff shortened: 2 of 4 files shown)


ubergarm:
This PR does seem to improve PPL as compared to tip of master using `-ctk q4_0 -ctv q4_0`.

I ran everything CPU-only backend so as to also compare https://github.com/Aaryan-Kapoor/llama.cpp/tree/turboquant-tq3_0 from @Aaryan-Kapoor to see how that specific `tq3_0` implementation compares.

For reference:
* f16 = 16 bpw
* q4_0 = 4.5 bpw
* tq3_0 = 3.5 bpw

#### mainline llama.cpp PPL wiki.test.raw

## Data Results

#### mainline llama.cpp PPL wiki.test.raw

| k-cache | v-cache | *baseline* master@6861f6509 | *PR* gg/attn-rot@e5aa067d6 | Aaryan-Kapoor/turboquant-tq3_0@1fb1fb3ab |
|---------|---------|-----------------------------|----------------------------|--------------------|
| f16     | f16     | 6.5788 +/- 0.04196, 688.81 tok/sec     | 6.5788 +/- 0.04196, 721.01 tok/sec      |                    |
| q4_0    | q4_0    | 6.6148 +/- 0.04216, 703.91 tok/sec     | 6.5962 +/- 0.04208, 540.63 tok/sec   |                    |
| tq3_0   | tq3_0   |                             |                            | 6.6911 +/- 0.04273, 481 tok/sec |

<details>

<summary>👈 Details</summary>

## Experiment
Measure perplexity against wiki.test.raw varying kv-cache quantization.

## Test Quant
* [ubergarm/Qwen3.5-35B-A3B-GGUF Q4_0 19.776 GiB (4.901 BPW)](https://huggingface.co/ubergarm/Qwen3.5-35B-A3B-GGUF?show_file_info=Qwen3.5-35B-A3B-Q4_0.gguf#q4_0-19776-gib-4901-bpw)

## Test Rig
* AMD EPYC 9975 128-Core w/ 12x64GiB DDR5@6400MT/s NPS1 Single Socket (~538 GB/s via mlc)
* Compiled CPU-only backend
```bash
cmake -B build -DCMAKE_BUILD_TYPE=Release -DGGML_CUDA=OFF -DGGML_RPC=OFF -DGGML_BLAS=OFF -DGGML_VULKAN=OFF
cmake --build build --config Release -j $(nproc)
```

## Command
```
# seed does nothing as no sampling here, but is fun
model=/ubergarm/Qwen3.5-35B-A3B-GGUF/Qwen3.5-35B-A3B-Q4_0.gguf
    #-ctk f16 -ctv f16 \
numactl -N "$SOCKET" -m "$SOCKET" \
./build/bin/llama-perplexity \
    -m "$model" \
    -f wiki.test.raw \
    -ctk q4_0 -ctv q4_0 \
    --seed 1337 \
    --ctx-size 512 \
    -ub 512 -b 2048 \
    --no-mmap \
    --numa numactl \
    --threads 96 \
    --threads-batch 128
```

</details>

AesSedai:
I've run a pretty big permutation matrix for the Qwen3.5-9B model against the master branch and this PR branch.

The following images are the ratios vs F16/F16:

Master branch PPL and KLD
image
image

and this PR PPL and KLD
image
image


<details>

<summary>Here's the table of data comparing master vs this PR</summary>

| k-cache | v-cache | master PPL | master mean KLD | attn-rot PPL | attn-rot KLD
|---------|---------|-----|----------|----------|----------|
| f16 | f16 | 8.195983 ± 0.055532 | 0.000782 ± 0.000040 | 8.195983 ± 0.055532 | 0.000782 ± 0.000040 |
| f16 | q4_0 | 8.208920 ± 0.055630 | 0.003747 ± 0.000182 | 8.208920 ± 0.055630 | 0.003747 ± 0.000182 |
| f16 | q4_1 | 8.212052 ± 0.055711 | 0.002985 ± 0.000140 | 8.212052 ± 0.055711 | 0.002985 ± 0.000140 |
| f16 | q5_0 | 8.206032 ± 0.055617 | 0.001752 ± 0.000111 | 8.206032 ± 0.055617 | 0.001752 ± 0.000111 |
| f16 | q5_1 | 8.195687 ± 0.055524 | 0.001393 ± 0.000111 | 8.195687 ± 0.055524 | 0.001393 ± 0.000111 |
| f16 | q8_0 | 8.196533 ± 0.055536 | 0.000804 ± 0.000037 | 8.196533 ± 0.055536 | 0.000804 ± 0.000037 |
| q4_0 | f16 | 8.212304 ± 0.055635 | 0.005426 ± 0.000271 | 8.212304 ± 0.055635 | 0.005426 ± 0.000271 |
| q4_0 | q4_0 | 8.227048 ± 0.055766 | 0.007979 ± 0.000352 | 8.211705 ± 0.055631 | 0.005059 ± 0.000224 |
| q4_0 | q4_1 | 8.223944 ± 0.055773 | 0.007163 ± 0.000298 | 8.209285 ± 0.055635 | 0.004664 ± 0.000196 |
| q4_0 | q5_0 | 8.217618 ± 0.055691 | 0.006165 ± 0.000307 | 8.201787 ± 0.055561 | 0.003708 ± 0.000187 |
| q4_0 | q5_1 | 8.210686 ± 0.055628 | 0.006011 ± 0.000332 | 8.201859 ± 0.055564 | 0.003656 ± 0.000203 |
| q4_0 | q8_0 | 8.209199 ± 0.055617 | 0.005336 ± 0.000272 | 8.198971 ± 0.055537 | 0.003334 ± 0.000160 |
| q4_1 | f16 | 8.219674 ± 0.055730 | 0.004174 ± 0.000239 | 8.219674 ± 0.055730 | 0.004174 ± 0.000239 |
| q4_1 | q4_0 | 8.232653 ± 0.055833 | 0.006612 ± 0.000263 | 8.213936 ± 0.055656 | 0.004702 ± 0.000223 |
| q4_1 | q4_1 | 8.235108 ± 0.055910 | 0.006014 ± 0.000266 | 8.210796 ± 0.055641 | 0.004252 ± 0.000183 |
| q4_1 | q5_0 | 8.229521 ± 0.055827 | 0.004827 ± 0.000253 | 8.204419 ± 0.055595 | 0.003456 ± 0.000191 |
| q4_1 | q5_1 | 8.221232 ± 0.055750 | 0.004739 ± 0.000252 | 8.203722 ± 0.055583 | 0.003297 ± 0.000171 |
| q4_1 | q8_0 | 8.219816 ± 0.055743 | 0.004005 ± 0.000209 | 8.201606 ± 0.055565 | 0.002968 ± 0.000156 |
| q5_0 | f16 | 8.202763 ± 0.055572 | 0.002223 ± 0.000157 | 8.202763 ± 0.055572 | 0.002223 ± 0.000157 |
| q5_0 | q4_0 | 8.213422 ± 0.055674 | 0.005025 ± 0.000243 | 8.211162 ± 0.055646 | 0.003654 ± 0.000244 |
| q5_0 | q4_1 | 8.217085 ± 0.055749 | 0.004181 ± 0.000185 | 8.206240 ± 0.055612 | 0.003055 ± 0.000165 |
| q5_0 | q5_0 | 8.214116 ± 0.055692 | 0.002849 ± 0.000143 | 8.200742 ± 0.055571 | 0.002062 ± 0.000124 |
| q5_0 | q5_1 | 8.202209 ± 0.055575 | 0.002657 ± 0.000170 | 8.200578 ± 0.055565 | 0.002028 ± 0.000141 |
| q5_0 | q8_0 | 8.201038 ± 0.055579 | 0.002310 ± 0.000168 | 8.197975 ± 0.055550 | 0.001553 ± 0.000095 |
| q5_1 | f16 | 8.195862 ± 0.055514 | 0.001843 ± 0.000165 | 8.195862 ± 0.055514 | 0.001843 ± 0.000165 |
| q5_1 | q4_0 | 8.208856 ± 0.055624 | 0.004611 ± 0.000216 | 8.210008 ± 0.055633 | 0.003729 ± 0.000258 |
| q5_1 | q4_1 | 8.210093 ± 0.055686 | 0.003773 ± 0.000176 | 8.206406 ± 0.055621 | 0.002889 ± 0.000152 |
| q5_1 | q5_0 | 8.207489 ± 0.055626 | 0.002754 ± 0.000220 | 8.200513 ± 0.055562 | 0.001825 ± 0.000103 |
| q5_1 | q5_1 | 8.195263 ± 0.055514 | 0.002626 ± 0.000226 | 8.200874 ± 0.055575 | 0.001821 ± 0.000121 |
| q5_1 | q8_0 | 8.194897 ± 0.055520 | 0.001957 ± 0.000178 | 8.198389 ± 0.055547 | 0.001750 ± 0.000182 |
| q8_0 | f16 | 8.197374 ± 0.055531 | 0.000923 ± 0.000075 | 8.197374 ± 0.055531 | 0.000923 ± 0.000075 |
| q8_0 | q4_0 | 8.209297 ± 0.055632 | 0.003642 ± 0.000159 | 8.208469 ± 0.055623 | 0.002935 ± 0.000204 |
| q8_0 | q4_1 | 8.210675 ± 0.055697 | 0.002920 ± 0.000116 | 8.202587 ± 0.055587 | 0.002519 ± 0.000149 |
| q8_0 | q5_0 | 8.207077 ± 0.055629 | 0.001715 ± 0.000095 | 8.197825 ± 0.055538 | 0.001263 ± 0.000053 |
| q8_0 | q5_1 | 8.196398 ± 0.055535 | 0.001484 ± 0.000113 | 8.200065 ± 0.055566 | 0.001187 ± 0.000055 |
| q8_0 | q8_0 | 8.196226 ± 0.055531 | 0.000866 ± 0.000048 | 8.196227 ± 0.055532 | 0.000757 ± 0.000039 |

</details>


And attached here is the full raw run data for every invocation:
[analysis-results.zip](https://github.com/user-attachments/files/26298566/analysis-results.zip)

Dampfinchen:
> This PR does seem to improve PPL as compared to tip of master using `-ctk q4_0 -ctv q4_0`.
> 
> I ran everything CPU-only backend so as to also compare https://github.com/Aaryan-Kapoor/llama.cpp/tree/turboquant-tq3_0 from @Aaryan-Kapoor to see how that specific `tq3_0` implementation compares.
> 
> For reference:
> 
> * f16 = 16 bpw
> * q4_0 = 4.5 bpw
> * tq3_0 = 3.5 bpw
> 
> #### mainline llama.cpp PPL wiki.test.raw
> ## Data Results
> #### mainline llama.cpp PPL wiki.test.raw
> k-cache	v-cache	_baseline_ master@6861f6509	_PR_ gg/attn-rot@e5aa067d6	Aaryan-Kapoor/turboquant-tq3_0@1fb1fb3ab
> f16	f16	6.5788 +/- 0.04196, 688.81 tok/sec	6.5788 +/- 0.04196, 721.01 tok/sec	
> q4_0	q4_0	6.6148 +/- 0.04216, 703.91 tok/sec	6.5962 +/- 0.04208, 540.63 tok/sec	
> tq3_0	tq3_0			6.6911 +/- 0.04273, 481 tok/sec
> 👈 Details

Interesting results, which indicate that Q4_0, even before this PR, is superior to TQ_3. However as a word of caution, this is likely due to a very experimental and early implementation of this specific fork and not indicative of the actual performance of TurboQuant, as it is supposed to be effectively lossless compared to fp16 kv cache. That is not the case here.

ggerganov:
I _think_ we can actually rotate the V tensor using smaller matrices (64 x 64) which should result in better quality of the V cache. We cannot do that for the Q and K because it would not preserve the dot product.

Pushed a change to do that [832e326](https://github.com/ggml-org/llama.cpp/pull/21038/commits/832e32639f600197ad2bfbb2e5bf9c1479243a49) and just from a quick PPL sanity check it looks slightly better.

Note, using 64 instead of 32 because the Metal matrix multiplication kernels require `ne00` to be at least 64.

Edit:

> We cannot do that for the Q and K because it would not preserve the dot product.

On second thought, I think it does preserve it. Something to try too. Here is the patch:

<details>
<summary>More rotations for Q and K</summary>

```diff
diff --git a/src/llama-graph.cpp b/src/llama-graph.cpp
index 8dfc92b71..84bcf26be 100644
--- a/src/llama-graph.cpp
+++ b/src/llama-graph.cpp
@@ -2096,8 +2096,9 @@ static std::unique_ptr<llm_graph_input_attn_kv> build_attn_inp_kv_impl(
         const bool can_rot =
             !hparams.is_n_embd_k_gqa_variable() &&
             !hparams.is_n_embd_v_gqa_variable() &&
-            ggml_is_power_of_2(hparams.n_embd_head_k()) &&
+          //ggml_is_power_of_2(hparams.n_embd_head_k()) &&
           //ggml_is_power_of_2(hparams.n_embd_head_v()) &&
+            hparams.n_embd_head_k() % 64 == 0 &&
             hparams.n_embd_head_v() % 64 == 0 &&
             hparams.n_embd_head_k() >= 64 &&
             hparams.n_embd_head_v() >= 64 &&
@@ -2105,7 +2106,7 @@ static std::unique_ptr<llm_graph_input_attn_kv> build_attn_inp_kv_impl(
             ggml_is_quantized(mctx_cur->type_v());
 
         if (can_rot) {
-            const auto nk = hparams.n_embd_head_k();
+            const auto nk = 64;
             const auto nv = 64;
 
             inp->self_rotk = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, nk, nk);
@@ -2453,8 +2454,9 @@ llm_graph_input_attn_kv_iswa * llm_graph_context::build_attn_inp_kv_iswa() const
         const bool can_rot =
             !hparams.is_n_embd_k_gqa_variable() &&
             !hparams.is_n_embd_v_gqa_variable() &&
-            ggml_is_power_of_2(hparams.n_embd_head_k()) &&
+          //ggml_is_power_of_2(hparams.n_embd_head_k()) &&
           //ggml_is_power_of_2(hparams.n_embd_head_v()) &&
+            hparams.n_embd_head_k() % 64 == 0 &&
             hparams.n_embd_head_v() % 64 == 0 &&
             hparams.n_embd_head_k() >= 64 &&
             hparams.n_embd_head_v() >= 64 &&
@@ -2462,7 +2464,7 @@ llm_graph_input_attn_kv_iswa * llm_graph_context::build_attn_inp_kv_iswa() const
             ggml_is_quantized(mctx_cur->get_base()->type_v());
 
         if (can_rot) {
-            const auto nk = hparams.n_embd_head_k();
+            const auto nk = 64;
             const auto nv = 64;
 
             inp->self_rotk = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, nk, nk);
```
</details>

Rotatingxenomorph:
> > This PR does seem to improve PPL as compared to tip of master using `-ctk q4_0 -ctv q4_0`.
> > I ran everything CPU-only backend so as to also compare https://github.com/Aaryan-Kapoor/llama.cpp/tree/turboquant-tq3_0 from @Aaryan-Kapoor to see how that specific `tq3_0` implementation compares.
> > For reference:
> > 
> > * f16 = 16 bpw
> > * q4_0 = 4.5 bpw
> > * tq3_0 = 3.5 bpw
> > 
> > #### mainline llama.cpp PPL wiki.test.raw
> > ## Data Results
> > #### mainline llama.cpp PPL wiki.test.raw
> > k-cache	v-cache	_baseline_ master@6861f6509	_PR_ gg/attn-rot@e5aa067d6	Aaryan-Kapoor/turboquant-tq3_0@1fb1fb3ab
> > f16	f16	6.5788 +/- 0.04196, 688.81 tok/sec	6.5788 +/- 0.04196, 721.01 tok/sec
> > q4_0	q4_0	6.6148 +/- 0.04216, 703.91 tok/sec	6.5962 +/- 0.04208, 540.63 tok/sec
> > tq3_0	tq3_0			6.6911 +/- 0.04273, 481 tok/sec
> > 👈 Details
> 
> Interesting results, which indicate that Q4_0, even before this PR, is superior to TQ_3. However as a word of caution, this is likely due to a very experimental and early implementation of this specific fork and not indicative of the actual performance of TurboQuant, as it is supposed to be effectively lossless compared to fp16 kv cache. That is not the case here.

It can't approach turboquant's performance because it's missing the QJL part of turboquant. (I think?)

AesSedai:
I re-ran the suite against the updated new commit, looks like the KLD has generally improved a little bit across the board!

Heatmap results:

image

image

<details>

<summary>Details data table</summary>

| k-cache | v-cache | PPL | mean KLD |
|---------|---------|-----|----------|
| f16 | f16 | 8.195983 ± 0.055532 | 0.000782 ± 0.000040 |
| f16 | q4_0 | 8.205849 ± 0.055607 | 0.002518 ± 0.000094 |
| f16 | q4_1 | 8.206960 ± 0.055627 | 0.002308 ± 0.000088 |
| f16 | q5_0 | 8.197224 ± 0.055540 | 0.001275 ± 0.000052 |
| f16 | q5_1 | 8.198303 ± 0.055550 | 0.001126 ± 0.000043 |
| f16 | q8_0 | 8.196163 ± 0.055536 | 0.000838 ± 0.000058 |
| q4_0 | f16 | 8.202777 ± 0.055567 | 0.003475 ± 0.000182 |
| q4_0 | q4_0 | 8.208411 ± 0.055610 | 0.004782 ± 0.000194 |
| q4_0 | q4_1 | 8.210197 ± 0.055641 | 0.004592 ± 0.000184 |
| q4_0 | q5_0 | 8.199608 ± 0.055539 | 0.003880 ± 0.000205 |
| q4_0 | q5_1 | 8.200301 ± 0.055547 | 0.003644 ± 0.000190 |
| q4_0 | q8_0 | 8.199085 ± 0.055545 | 0.003346 ± 0.000183 |
| q4_1 | f16 | 8.202827 ± 0.055570 | 0.003155 ± 0.000182 |
| q4_1 | q4_0 | 8.214026 ± 0.055670 | 0.004670 ± 0.000214 |
| q4_1 | q4_1 | 8.213453 ± 0.055686 | 0.004174 ± 0.000173 |
| q4_1 | q5_0 | 8.200692 ± 0.055565 | 0.003446 ± 0.000189 |
| q4_1 | q5_1 | 8.199583 ± 0.055558 | 0.003457 ± 0.000206 |
| q4_1 | q8_0 | 8.200708 ± 0.055569 | 0.002958 ± 0.000168 |
| q5_0 | f16 | 8.199265 ± 0.055551 | 0.001553 ± 0.000099 |
| q5_0 | q4_0 | 8.207457 ± 0.055617 | 0.003247 ± 0.000161 |
| q5_0 | q4_1 | 8.208826 ± 0.055647 | 0.002947 ± 0.000144 |
| q5_0 | q5_0 | 8.200150 ± 0.055558 | 0.002054 ± 0.000158 |
| q5_0 | q5_1 | 8.199639 ± 0.055568 | 0.001954 ± 0.000120 |
| q5_0 | q8_0 | 8.197875 ± 0.055554 | 0.001521 ± 0.000093 |
| q5_1 | f16 | 8.200392 ± 0.055557 | 0.001585 ± 0.000137 |
| q5_1 | q4_0 | 8.210025 ± 0.055650 | 0.003052 ± 0.000151 |
| q5_1 | q4_1 | 8.207874 ± 0.055632 | 0.002765 ± 0.000128 |
| q5_1 | q5_0 | 8.200399 ± 0.055565 | 0.001849 ± 0.000126 |
| q5_1 | q5_1 | 8.199001 ± 0.055561 | 0.001839 ± 0.000146 |
| q5_1 | q8_0 | 8.199596 ± 0.055562 | 0.001544 ± 0.000125 |
| q8_0 | f16 | 8.196933 ± 0.055526 | 0.000837 ± 0.000047 |
| q8_0 | q4_0 | 8.205234 ± 0.055600 | 0.002568 ± 0.000141 |
| q8_0 | q4_1 | 8.206657 ± 0.055621 | 0.002230 ± 0.000084 |
| q8_0 | q5_0 | 8.197552 ± 0.055547 | 0.001229 ± 0.000046 |
| q8_0 | q5_1 | 8.197543 ± 0.055546 | 0.001300 ± 0.000117 |
| q8_0 | q8_0 | 8.196556 ± 0.055535 | 0.000799 ± 0.000042 |

</details>


Full run archives:
[attn-rot-7711b3a36.zip](https://github.com/user-attachments/files/26318587/attn-rot-7711b3a36.zip)

Dampfinchen:
> I re-ran the suite against the updated new commit, looks like the KLD has generally improved a little bit across the board!
> 
> Heatmap results:
> 
> image image
> Details data table
> k-cache	v-cache	PPL	mean KLD
> f16	f16	8.195983 ± 0.055532	0.000782 ± 0.000040
> f16	q4_0	8.205849 ± 0.055607	0.002518 ± 0.000094
> f16	q4_1	8.206960 ± 0.055627	0.002308 ± 0.000088
> f16	q5_0	8.197224 ± 0.055540	0.001275 ± 0.000052
> f16	q5_1	8.198303 ± 0.055550	0.001126 ± 0.000043
> f16	q8_0	8.196163 ± 0.055536	0.000838 ± 0.000058
> q4_0	f16	8.202777 ± 0.055567	0.003475 ± 0.000182
> q4_0	q4_0	8.208411 ± 0.055610	0.004782 ± 0.000194
> q4_0	q4_1	8.210197 ± 0.055641	0.004592 ± 0.000184
> q4_0	q5_0	8.199608 ± 0.055539	0.003880 ± 0.000205
> q4_0	q5_1	8.200301 ± 0.055547	0.003644 ± 0.000190
> q4_0	q8_0	8.199085 ± 0.055545	0.003346 ± 0.000183
> q4_1	f16	8.202827 ± 0.055570	0.003155 ± 0.000182
> q4_1	q4_0	8.214026 ± 0.055670	0.004670 ± 0.000214
> q4_1	q4_1	8.213453 ± 0.055686	0.004174 ± 0.000173
> q4_1	q5_0	8.200692 ± 0.055565	0.003446 ± 0.000189
> q4_1	q5_1	8.199583 ± 0.055558	0.003457 ± 0.000206
> q4_1	q8_0	8.200708 ± 0.055569	0.002958 ± 0.000168
> q5_0	f16	8.199265 ± 0.055551	0.001553 ± 0.000099
> q5_0	q4_0	8.207457 ± 0.055617	0.003247 ± 0.000161
> q5_0	q4_1	8.208826 ± 0.055647	0.002947 ± 0.000144
> q5_0	q5_0	8.200150 ± 0.055558	0.002054 ± 0.000158
> q5_0	q5_1	8.199639 ± 0.055568	0.001954 ± 0.000120
> q5_0	q8_0	8.197875 ± 0.055554	0.001521 ± 0.000093
> q5_1	f16	8.200392 ± 0.055557	0.001585 ± 0.000137
> q5_1	q4_0	8.210025 ± 0.055650	0.003052 ± 0.000151
> q5_1	q4_1	8.207874 ± 0.055632	0.002765 ± 0.000128
> q5_1	q5_0	8.200399 ± 0.055565	0.001849 ± 0.000126
> q5_1	q5_1	8.199001 ± 0.055561	0.001839 ± 0.000146
> q5_1	q8_0	8.199596 ± 0.055562	0.001544 ± 0.000125
> q8_0	f16	8.196933 ± 0.055526	0.000837 ± 0.000047
> q8_0	q4_0	8.205234 ± 0.055600	0.002568 ± 0.000141
> q8_0	q4_1	8.206657 ± 0.055621	0.002230 ± 0.000084
> q8_0	q5_0	8.197552 ± 0.055547	0.001229 ± 0.000046
> q8_0	q5_1	8.197543 ± 0.055546	0.001300 ± 0.000117
> q8_0	q8_0	8.196556 ± 0.055535	0.000799 ± 0.000042
> Full run archives: [attn-rot-7711b3a36.zip](https://github.com/user-attachments/files/26318587/attn-rot-7711b3a36.zip)

What about higher context, like for example 100K? I believe that's where kv cache quantization actually harms the model from multiplications errors that accumulated.

ggerganov:
@AesSedai Thanks for the results. It seems important to track the KLD rather than PPL (maybe more significant for Qwen3.5).

I did some additional tests with randomized Hadamard matrices (as suggested by @sashkboos in https://github.com/ggml-org/llama.cpp/issues/6444#issuecomment-2042194785). Will follow-up in next PRs.

src/llama-kv-cache.cpp
```diff
@@ -13,6 +13,54 @@
...
+    const int n = tensor->ne[0];
+
+    assert(ggml_is_power_of_2(n));
+    assert(tensor->ne[1] == n);
+    assert(tensor->ne[2] == 1);
+    assert(tensor->ne[3] == 1);
+
+    auto * data = (float *) tensor->data;
+
+    data[0*n + 0] = 1.0 / sqrtf(n);
```
CISC, line 41: This always gets me, at least this one makes esthetic sense. :)

CISC (approved):

am17an:
There is a slowdown which is expected, however probably we should have a flag to opt out?

` GGML_CUDA=ON ./scripts/compare-commits.sh upstream/master gg/attn-rot llama-bench -ctk q4_0 -ctv q4_0 -m /opt/models/qwen_3-30b3a-q4_0.gguf -fa 1 -d 0,2048,4096,8192,16384,32768`

| Model                 | Test         |   t/s 51a84efc5 |   t/s gg/attn-rot |   Speedup |
|:----------------------|:-------------|----------------:|------------------:|----------:|
| qwen3moe 30B.A3B Q4_0 | pp512        |         7603.68 |           7409.35 |      0.97 |
| qwen3moe 30B.A3B Q4_0 | pp512@d2048  |         7085.71 |           6912.28 |      0.98 |
| qwen3moe 30B.A3B Q4_0 | pp512@d4096  |         6586.79 |           6449.26 |      0.98 |
| qwen3moe 30B.A3B Q4_0 | pp512@d8192  |         5844.15 |           5694.56 |      0.97 |
| qwen3moe 30B.A3B Q4_0 | pp512@d16384 |         4704.12 |           4612.89 |      0.98 |
| qwen3moe 30B.A3B Q4_0 | pp512@d32768 |         3398.99 |           3346.66 |      0.98 |
| qwen3moe 30B.A3B Q4_0 | tg128        |          229.42 |            202.37 |      0.88 |
| qwen3moe 30B.A3B Q4_0 | tg128@d2048  |          210.24 |            188.03 |      0.89 |
| qwen3moe 30B.A3B Q4_0 | tg128@d4096  |          197.10 |            177.06 |      0.90 |
| qwen3moe 30B.A3B Q4_0 | tg128@d8192  |          175.60 |            159.45 |      0.91 |
| qwen3moe 30B.A3B Q4_0 | tg128@d16384 |          145.26 |            134.28 |      0.92 |
| qwen3moe 30B.A3B Q4_0 | tg128@d32768 |          101.33 |             94.13 |      0.93 |

handpickencounter:
> I don't know what is the impact of the remaining techniques explained in TurboQuant (PolarQuant, QJL, etc.). They could be important and can potentially improve further on top of this.

PolarQuant is just a rotation by a random matrix to spread the energy of any outliers across all dimensions prior to quantization. A normalized Hadamard matrix may or may not produce similar results.

QJL is extremely important however, PolarQuant (and your Hadamard) gives biased attention scores (despite the good seeming MSE) which results in rank flips. QJL uses the 'unreasonable effectiveness of random projections' to give unbiased results. The theory (more accurately, the lemma) is that randomly projecting a vector to lower dimensions will preserve pairwise distances (and therefore dot products) in expectation. TurboQuant quantizes the error residual all the way to 1-bit (the sign) which somehow works well enough.

The best technical yet intuitive explanation I saw on the subject so far is by some AI Researcher from Amazon: https://darshanfofadiya.com/research-papers/turboquant/

CISC:
> There is a slowdown which is expected, however probably we should have a flag to opt out?

I don't think we should, the only reason to quantize kv-cache is to save memory, if that comes at the cost of speed, so be it, an option to reduce quality does not make sense (unless you think this is not a general quality improvement).

am17an:
My opinion is that any new breaking change should have an option to turn it off, in case of any unforeseen issues. This change changes outputs of the LLM (even if they are materially better), so just as an example someone having tests downstream would see tests break. 

> the only reason to quantize kv-cache is to save memory

I think when using `-nkvo`, it will be faster to transfer a quantized cache to the GPU when offloading, so it can be used for a performance reason also. Also the CUDA fattn vec kernel uses `dp4a` for q8_0 which is 2x the throughput vs f16. So that statement is just wrong IMO.

Rotatingxenomorph:
> > I don't know what is the impact of the remaining techniques explained in TurboQuant (PolarQuant, QJL, etc.). They could be important and can potentially improve further on top of this.
> 
> PolarQuant is just a rotation by a random matrix to spread the energy of any outliers across all dimensions prior to quantization. A normalized Hadamard matrix may or may not produce similar results.
> 
> QJL is extremely important however, PolarQuant (and your Hadamard) gives biased attention scores (despite the good seeming MSE) which results in rank flips. QJL uses the 'unreasonable effectiveness of random projections' to give unbiased results. The theory (more accurately, the lemma) is that randomly projecting a vector to lower dimensions will preserve pairwise distances (and therefore dot products) in expectation. TurboQuant quantizes the error residual all the way to 1-bit (the sign) which somehow works well enough.
> 
> The best technical yet intuitive explanation I saw on the subject so far is by some AI Researcher from Amazon: https://darshanfofadiya.com/research-papers/turboquant/

I have a bad feeling about this PR now. How do we know there's not a regression compared to naive quantization as context grows to large(r) sizes?

CISC:
> My opinion is that any new breaking change should have an option to turn it off, in case of any unforeseen issues.

You could say we are in the business of making breaking changes and that `git bisect` is our option, but I digress. :)

If anything, I guess adding an env-var for some grace period would be sufficient.

> This change changes outputs of the LLM (even if they are materially better), so just as an example someone having tests downstream would see tests break.

Well, until you pointed out the below I would have said that no such test were likely to exist. :)

> I think when using `-nkvo`, it will be faster to transfer a quantized cache to the GPU when offloading, so it can be used for a performance reason also. Also the CUDA fattn vec kernel uses `dp4a` for q8_0 which is 2x the throughput vs f16. So that statement is just wrong IMO.

Dampfinchen:
> > There is a slowdown which is expected, however probably we should have a flag to opt out?
> 
> I don't think we should, the only reason to quantize kv-cache is to save memory, if that comes at the cost of speed, so be it, an option to reduce quality does not make sense (unless you think this is not a general quality improvement).

I disagree. You could use quantized KV Cache to save memory in order to put more layers on the GPU. In that case, the purpose of it would be to increase speed and this change would be counterproductive to that goal.

CISC:
> > > There is a slowdown which is expected, however probably we should have a flag to opt out?
> > 
> > 
> > I don't think we should, the only reason to quantize kv-cache is to save memory, if that comes at the cost of speed, so be it, an option to reduce quality does not make sense (unless you think this is not a general quality improvement).
> 
> I disagree. You could use quantized KV Cache to save memory in order to put more layers on the GPU. In that case, the purpose of it would be to increase speed and this change would be counterproductive to that goal.

You'd most likely gain much more than you lose, so calling it counterproductive is perhaps a bit far fetched.

am17an:
> If anything, I guess adding an env-var for some grace period would be sufficient.

Actually I'm not sure it would be worth it to be on by default for q8_0 since the benefits are much lesser compared to q4 and the performance loss is significant

<details>

| Model                 | Test         |   t/s 51a84efc5 |   t/s gg/attn-rot |   Speedup |
|:----------------------|:-------------|----------------:|------------------:|----------:|
| qwen3moe 30B.A3B Q4_0 | pp512        |         7576.02 |           7372.24 |      0.97 |
| qwen3moe 30B.A3B Q4_0 | pp512@d2048  |         7045.27 |           6838.27 |      0.97 |
| qwen3moe 30B.A3B Q4_0 | pp512@d4096  |         6580.31 |           6435.85 |      0.98 |
| qwen3moe 30B.A3B Q4_0 | pp512@d8192  |         5792.41 |           5661.65 |      0.98 |
| qwen3moe 30B.A3B Q4_0 | pp512@d16384 |         4660.49 |           4584.87 |      0.98 |
| qwen3moe 30B.A3B Q4_0 | pp512@d32768 |         3349.27 |           3300.92 |      0.99 |
| qwen3moe 30B.A3B Q4_0 | tg128        |          230.59 |            203.59 |      0.88 |
| qwen3moe 30B.A3B Q4_0 | tg128@d2048  |          212.31 |            189.41 |      0.89 |
| qwen3moe 30B.A3B Q4_0 | tg128@d4096  |          199.79 |            179.40 |      0.90 |
| qwen3moe 30B.A3B Q4_0 | tg128@d8192  |          179.48 |            162.80 |      0.91 |
| qwen3moe 30B.A3B Q4_0 | tg128@d16384 |          149.03 |            137.47 |      0.92 |
| qwen3moe 30B.A3B Q4_0 | tg128@d32768 |          105.50 |             98.55 |      0.93 | 

</details>

CISC:
> > If anything, I guess adding an env-var for some grace period would be sufficient.
> 
> Actually I'm not sure it would be worth it to be on by default for q8_0 since the benefits are much lesser compared to q4 and the performance loss is significant

Granted, it's probably pointless for q8_0, so perhaps add a check.

ggerganov:
It seems that evaluating AIME25 could be another sensitivity test for confirming the improvement of rotating the activations. I did a few runs today with `gpt-oss-20b` on low reasoning (for faster evaluation):

Here is the table with the hyperlinks preserved for each score:

| eval  | KV type | score (no rot) | score (rot) |
| ---  | ---  | --- | --- |
| AIME25 x8 | F16  | [37.9%](https://github.com/user-attachments/files/26332591/aime2025-gpt-oss-20b-low-x8-kv_f16.json.html) | — |
| AIME25 x8 | Q8_0 | [31.7%](https://github.com/user-attachments/files/26332631/aime2025-gpt-oss-20b-low-x8-kv_q8_0.json.html) | [37.1%](https://github.com/user-attachments/files/26332632/aime2025-gpt-oss-20b-low-x8-kv_q8_0-rot.json.html) |
| AIME25 x8 | Q5_1 | [30.8%](https://github.com/user-attachments/files/26335073/aime2025-gpt-oss-20b-low-x8-kv_q5_1.json.html) | [32.5%](https://github.com/user-attachments/files/26335081/aime2025-gpt-oss-20b-low-x8-kv_q5_1-rot.json.html) |
| AIME25 x8 | Q5_0 | [25.4%](https://github.com/user-attachments/files/26339183/aime2025-gpt-oss-20b-low-x8-kv_q5_0.json.html) | [32.5%](https://github.com/user-attachments/files/26339184/aime2025-gpt-oss-20b-low-x8-kv_q5_0-rot.json.html) |
| AIME25 x8 | Q4_1 | [18.3%](https://github.com/user-attachments/files/26339188/aime2025-gpt-oss-20b-low-x8-kv_q4_1.json.html) | [28.3%](https://github.com/user-attachments/files/26339136/aime2025-gpt-oss-20b-low-x8-kv_q4_1-rot.json.html) |
| AIME25 x8 | Q4_0 |  [2.0%](https://github.com/user-attachments/files/26334313/aime2025-gpt-oss-20b-low-x8-kv_q4_0.json.html) | [21.7%](https://github.com/user-attachments/files/26332635/aime2025-gpt-oss-20b-low-x8-kv_q4_0-rot.json.html) |

Used the new `llama-eval` script to perform these evals: https://github.com/ggml-org/llama.cpp/pull/21152

It's interesting that with `Q4_0` KV type without applying rotations, the model completely breaks down - it cannot solve even 1 problem from the 240 attempts. Note that it is still reasoning (i.e. it's not producing garbage outputs). The final answers and logic are just incorrect.

<details>
<summary>Example of reasoning trace with KV type `Q4_0` without rotations</summary>

image
</details>

Regarding `Q8_0` - I think there is some non-negligible quality benefit of applying the rotation here too (31.7% -> 37.1%). Though we need a bit more stats to confirm that. Atm, I am not convinced that disabling the rotations for `Q8_0` is the better option, but I'm still considering it.

For reference, the expected score of this eval per the `gpt-oss` model card is `37.1%`:

image

https://arxiv.org/pdf/2508.10925

Dampfinchen:
> It seems that evaluating AIME25 could be another sensitivity test for confirming the improvement of rotating the activations. I did a few runs today with `gpt-oss-20b` on low reasoning (for faster evaluation):
> 
> eval	KV type	rot	score	results (HTML)
> AIME25 x8	F16	no	37.9%	[aime2025-gpt-oss-20b-low-x8-kv_f16.json.html](https://github.com/user-attachments/files/26332591/aime2025-gpt-oss-20b-low-x8-kv_f16.json.html)
> AIME25 x8	Q8_0	no	31.7%	[aime2025-gpt-oss-20b-low-x8-kv_q8_0.json.html](https://github.com/user-attachments/files/26332631/aime2025-gpt-oss-20b-low-x8-kv_q8_0.json.html)
> AIME25 x8	Q8_0	yes	37.1%	[aime2025-gpt-oss-20b-low-x8-kv_q8_0-rot.json.html](https://github.com/user-attachments/files/26332632/aime2025-gpt-oss-20b-low-x8-kv_q8_0-rot.json.html)
> AIME25 x8	Q4_0	no	0.0%	[aime2025-gpt-oss-20b-low-x8-kv_q4_0.json.html](https://github.com/user-attachments/files/26332633/aime2025-gpt-oss-20b-low-x8-kv_q4_0.json.html)
> AIME25 x8	Q4_0	yes	21.7%	[aime2025-gpt-oss-20b-low-x8-kv_q4_0-rot.json.html](https://github.com/user-attachments/files/26332635/aime2025-gpt-oss-20b-low-x8-kv_q4_0-rot.json.html)
> Used the new `llama-eval` script to perform these evals: #21152
> 
> It's interesting that with `Q4_0` KV type without applying rotations, the model completely breaks down - it cannot solve even 1 problem from the 240 attempts. Note that it is still reasoning (i.e. it's not producing garbage outputs). The final answers and logic are just incorrect.
> 
> Example of reasoning trace with KV type `Q4_0` without rotations
> image
> Regarding `Q8_0` - I think there is some non-negligible quality benefit of applying the rotation here too (31.7% -> 37.1%). Though we need a bit more stats to confirm that. Atm, I am not convinced that disabling the rotations for `Q8_0` is the better option, but I'm still considering it.
> 
> For reference, the expected score of this eval per the `gpt-oss` model card is `37.1%`:
> 
> image
> https://arxiv.org/pdf/2508.10925

Yeah, based on the score, there is a clear benefit from rotating activations for q8_0, and I think it makes sense to have them on by default as the purpose of Q8_0 is to save memory for KV Cache for the same quality and with attn-rot it is noticeably closer to that goal. Furthermore, I'm not sure if AIME25 tests at long context as well. Quality differences might become even more noticeable at a longer context.

Performance is a concern, however. In the end, I think the best course of action would be to make attn-rots an opt-out/opt-in option with a simple flag.

am17an:
I think there are faster ways of doing the rotations, maybe with a dedicated op it would be faster.

ggerganov:
@am17an Adding a dedicated op for this does not sound like a good idea - it's a very narrow use case. I would rather have specialized matrix-matrix multiplications implemented in the backends for small `src[0]` that fits in shared memory. This is much more generic and allows broader usage (different rotations) and applications at similar performance.

There is also the option of extending `ggml_mul_mat` with an `op_param` for "special-case" multiplication (similar to how we set the precision), which can again optionally be considered by each backend when needed. Both options are much better than a dedicated ggml operator.

For example:

```c
cur = ggml_mul_mat(ctx, ...);
ggml_mul_mat_set_hint(cur, GGML_HINT_SRC0_IS_HADAMARD);
```

am17an:
@ggerganov while I agree it would be better as fusion or some other detection in the backend, it's not a narrow use-case (which was my point earlier about it being default true), it affects all quantized caches, so it's not really model specific, it will run for all models using a kv cache. So IMO it's important for it to have good performance.

ggerganov:
Yes, just to clarify by "narrow use case" I mean not for `llama.cpp` users, but from the PoV of the `ggml` API. We have to keep the API surface small. Adding operators is one of the things that increases maintenance significantly because it requires implementations in all backends. It is narrow use case because rotating using normalized Haramard matrix is a subcase of matrix multiplication. If we want to rotate with some other matrix, we don't want to write a new operator for that matrix. And yet, rotation is a very common operation in machine learning, so it's good to have it optimized in a generic way.

Performance is of course important. But note that KV cache quantization has always been marked as an experimental feature. So technically, it is not super important to preserve the performance on `master`. Plus, from what I am seeing in the latest evaluations, I don't think it has ever been clear how much it degrades the quality. How many people are probably using `Q4_0` KV cache and thinking "oh, PPL is not that bad". And yet, `gpt-oss-20b` cannot solve correctly even one AIME problem. IMO it's more important to have a better quality baseline and optimize on top of that, rather than trying to preserve what is currently on `master`.

erazortt:
> How many people are probably using `Q4_0` KV cache and thinking "oh, PPL is not that bad".

I think that is true for Q8 as well: How many people are using Q8_0 KV cache and are thinking that it would have no significant impact. However 37.1 vs 31.7 appears quite significant to me. And isn't the primary scope for KV quantisation memory usage while performance is only the secondary? So for me this sounds very much like rotations should be enabled even for Q8. However, perhaps there should be done some speed evaluation for Q8 as well, since as I see that has been done here only for Q4.

ryrAiy:
Georgi, hi first time posting here at llama.cpp, and thank you for all you and the crew do to keep the little guys in the game🫡 Also, I am not the biggest math guy, but I have a fair command of the technicals, and infinite curiosity with AI, and I have done a lot of AB testing with BF16 kv cache recently. I am getting not only better performance tps wise on my 4090, but better ability from different models too? Done a lot of vison and short and intermediate context window comparisons with old Cuda 12 builds and fp16 cache, and 13.2 builds with BF16 cache, and seems enough smoke to the fire to keep researching. Can you look at the KLD numbers and stuff with BF16 and the rotation modifications and see how that compares?

pwilkin:
These KV cache results are pretty interesting - I've done a lot benchmark testing on programming benchmarks on Q8_0 KV cache and I haven't observed that degradation.

Is it possible that this is specific to GPT-OSS? I'll maybe check with Qwen3.5 to compare.

mirek190:
> These KV cache results are pretty interesting - I've done a lot benchmark testing on programming benchmarks on Q8_0 KV cache and I haven't observed that degradation.
> 
> Is it possible that this is specific to GPT-OSS? I'll maybe check with Qwen3.5 to compare.

I have no idea what kind of programming you are doing but even for story writing degradation is noticable.

I noticed that Kv Q8 degradation on any model for story writing many months ago and posting on reddit even comparison but mostly people didn't believe me. 
Kv Q4 was making a total mess story writing.

Kv Q8 - is making stores hmm  ...  more flat , simpler stores 
Kv Q4 - broken logic in writing stores , flat, much simpler language, etc

Only Kv fp16 was / is useful 

If ROTATION will be implemented even for Q8 then be finally useful.

Meltedd:
@ggerganov re:

> Next PRs: Add randomized Hadamard matrices

When testing against your branch, I found that applying random +/-1 sign flips to the Hadamard matrix improves quantization quality at negligible additional cost on models with QK RMSNorm (e.g. Qwen3), where the plain Hadamard doesn't fully decorrelate the data.

PPL results with q4_0 KV cache (ctx=512, 4 chunks, wikitext-2):

| Model | Plain Hadamard | Randomized D\*H\*D |
|-------|---------------:|-------------------:|
| Qwen3-4B-Q8_0 | 13.08 | 12.79 (-2.3%) |

The sign flips need to be applied as D\*H\*D (same signs on rows and columns) to preserve H^2 = I.

---

I also have a butterfly Walsh-Hadamard implementation that brings the rotation cost from O(d^2) to O(d log d), which is pretty significant given the rotation runs per-head and per-token, but implementing that needs a native ggml op so probably a separate PR based on my understanding of the contribution guidelines.

I can submit either or both as PRs against your branch, just let me know what works best, and I am open to hearing your thoughts :)

erazortt:
I just wanted to provide some measurements on current hardware:  Running Qwen 3.5 35B on wiki.test.raw with 512 context and with bf16, q8_0 and q4_0 KV caches I am getting the following results:

master:
| KV quant | mean KLD | 99% KLD | same top p |
|-----------|-------------|-----------|--------------|
| bf16 | - | - | - |
| q8_0 | 0.003778 | 0.035869 | 97.303 |
| q4_0 | 0.010338 | 0.078723 | 95.331 |

| type_k | type_v |            test |                  t/s |
| -----: | -----: | --------------: | -------------------: |
|   bf16 |   bf16 |           pp512 |      5263.78 ± 23.30 |
|   bf16 |   bf16 |           tg128 |        173.58 ± 0.46 |
|   q8_0 |   q8_0 |           pp512 |     5210.77 ± 124.88 |
|   q8_0 |   q8_0 |           tg128 |        172.11 ± 0.50 |
|   q4_0 |   q4_0 |           pp512 |      5263.64 ± 15.16 |
|   q4_0 |   q4_0 |           tg128 |        171.63 ± 0.66 |

attn-rot branch:
| KV quant | mean KLD | 99% KLD | same top p |
|-----------|-------------|-----------|--------------|
| bf16 | - | - | - | 
| q8_0 | 0.003702 | 0.035608 | 97.355 |
| q4_0 | 0.007657 | 0.062180 | 96.070 |

| type_k | type_v |            test |                  t/s |
| -----: | -----: | --------------: | -------------------: |
|   bf16 |   bf16 |           pp512 |      5270.17 ± 25.16 |
|   bf16 |   bf16 |           tg128 |        173.47 ± 0.19 |
|   q8_0 |   q8_0 |           pp512 |      5231.55 ± 29.73 |
|   q8_0 |   q8_0 |           tg128 |        167.07 ± 0.75 |
|   q4_0 |   q4_0 |           pp512 |      5245.99 ± 21.93 |
|   q4_0 |   q4_0 |           tg128 |        166.47 ± 0.72 |

I am running this on a Blackwell with everything in VRAM.

segmond:
> These KV cache results are pretty interesting - I've done a lot benchmark testing on programming benchmarks on Q8_0 KV cache and I haven't observed that degradation.
> 
> Is it possible that this is specific to GPT-OSS? I'll maybe check with Qwen3.5 to compare.

I absolutely agree, I think it's worth checking out in other models.  I recall gpt-oss-20b being terrible, granted I still ran it in q8, so there might be issues with it.  I'm very curious to see if this applies to other models, we should definitely throw in gpt-oss-120b, qwen3.5 models and a few older dense models like gemma3, mistral-small into the mix.

pwilkin (approved):

vektorprime:
Just wanted to leave this image here since this post is referenced quite a bit:

The following results come from the official SLANG website, but they show that GPTQ and AIME to be sensitive to KV quant. It's NVFP4 but still very relevant, I think. @ggerganov maybe you can incorporate that into your future testing suite like you did with the python-based AIME testing.

image

PR #21040, opened 2026-03-26
hip: use fnuz fp8 for conversion on CDNA3

IMbackK:
HIP supports the fp8 types e4m3_fnuz and e4m3_ocp, the difference being that fnuz dosent support inf. GFX942 (uniquely) supports only e4m3_fnuz in hardware, due to what looks like an oversight in rocm, the combination of e4m3_ocp on devices with native fp8 support but no ocp support is not implemented.

Use native fnuz here to avoid this.

# Requirements

- AI usage disclosure: No

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -803,6 +803,8 @@ static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
 #ifdef FP8_AVAILABLE
     const uint32_t bits = x * (x != 0x7F && x != 0xFF); // Convert NaN to 0.0f to match CPU implementation.
 #if defined(GGML_USE_HIP) && defined(CDNA3)
+    // ROCm dose not support fp8 in software on devices with fp8 hardware,
+    // but CDNA3 supports only e4m3_fnuz (no inf).
     const __hip_fp8_e4m3_fnuz xf = *reinterpret_cast<const __hip_fp8_e4m3_fnuz *>(&bits);
 #else
     const __nv_fp8_e4m3 xf = *reinterpret_cast<const __nv_fp8_e4m3 *>(&bits);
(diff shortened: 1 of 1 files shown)


IMbackK:
@CISC the only way to avoid this kind of release failure would be to build for all target architectures in the quality check build, if that is worth the significant time required is something i leave to your judgement.

angt:
I think that's related to my issue: https://github.com/angt/installama.sh/actions/runs/23597760326/job/68760227462

CISC (approved):

CISC:
> @CISC the only way to avoid this kind of release failure would be to build for all target architectures in the quality check build, if that is worth the significant time required is something i leave to your judgement.

Sigh, yeah, that won't do...

CISC:
> I think that's related to my issue: https://github.com/angt/installama.sh/actions/runs/23597760326/job/68760227462

It is, it's also failing our release. :)

ggerganov (approved):

IMbackK:
> Sigh, yeah, that won't do...

We could at least improve coverage for free by building the various tests for different devices, Right now both builds in the build workflow are built for gfx906, and the quality check builds for gfx908 twice. We could build for gfx906 (linux) and gfx1100 (windows) in the build workflow and the quality check could build for gfx942 (werror) and gfx908 (vgpr)

CISC:
> > Sigh, yeah, that won't do...
> 
> We could at least improve coverage for free by building the various tests for different devices,

Whichever combination you think gives the best coverage...

JohannesGaessler (commented):
Is there a reason why we wouldn't want to change the `typedef` in `vendors/hip.h` instead?

IMbackK:
> Is there a reason why we wouldn't want to change the `typedef` in `vendors/hip.h` instead?

then you have the same problem in reverse on gfx12 and gfx955 (whenever we support those) they support e4m3_ocp only in hardware and software support is not available when any hardware support for fp8 is present

IMbackK:
as for changing the typedef on a per device basis, that would be fine with me but then we introduce an implicit difference between __nv_fp8_e4m3 on cuda and hip. I dont think the presence/absence of inf is going to cause a real problem but it dosent sit right with me, this way at least you have to think about if its a problem at the point of use.

I guess if we add a lot of code using fp8 it might get unwieldy

EDIT: also at the point of the typdef we cant know what to use in host code.

JohannesGaessler (approved):
In any case, this PR as it is is also fine with me. As of right now the amount of FP8 in our code is miniscule anyways so it's not like it makes a meaningful difference which way we do it and we should just fix the CI.

Please add a brief comment explaining the change though.

CISC (approved):
Merge on EditorConfig green.

PR #21045, opened 2026-03-26
model: add Falcon OCR support

avirajBevli:
## Overview
Add support for the FalconOCR model, a VLM designed for document OCR. 
Key features of the model: 
- 3D spatial RoPE: 1D temporal + 2D golden-ratio spatial positioning for image patches
- Attention sinks: learnable per-layer sink vectors prepended to KV
- Squared ReLU gated FFN with sqrt(2) gate scaling
- QK layer normalization before attention
- Combined non-causal batch for prefix tokens + image patches + suffix token
- Image preprocessing: aspect-ratio-preserving resize 

🤗 Blogpost: https://huggingface.co/tiiuae/Falcon-OCR
📄 Paper: https://arxiv.org/pdf/2603.27365
💻 Code: https://github.com/tiiuae/falcon-perception
🎮 Playground: https://vision.falcon.aidrc.tii.ae/

# Requirements
- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) : YES
- AI usage disclosure: YES - AI (Cursor) was used in an assistive capacity for integrating the model into the llama.cpp codebase.


tools/mtmd/mtmd.h
```diff
@@ -119,6 +119,14 @@ MTMD_API bool mtmd_decode_use_non_causal(mtmd_context * ctx);
 // whether the current model use M-RoPE for llama_decode
 MTMD_API bool mtmd_decode_use_mrope(mtmd_context * ctx);
 
+// whether the current model uses spatial 3D RoPE (temporal + 2D continuous spatial positions)
+MTMD_API bool mtmd_decode_use_spatial_3d_rope(mtmd_context * ctx);
```
ngxson, line 123: is this mrope?

tools/mtmd/mtmd.h
```diff
@@ -119,6 +119,14 @@ MTMD_API bool mtmd_decode_use_non_causal(mtmd_context * ctx);
...
+
+// whether the current model needs inter-chunk coordination where prefix text
+// embeddings (e.g. cls+regs) are merged with image patch embeddings into a
+// single non-causal batch, rather than processing chunks independently
+MTMD_API bool mtmd_decode_use_interchunk_merge(mtmd_context * ctx);
```
ngxson: IMO there is too much code added just to support this mode. it's exactly the same story with paligemma

unless more models using the same style, I don't think it worth adding this much code just to support one or two models

ngxson:
> Add support for the FalconOCR model, a VLM designed for document OCR.

I cannot find the model anywhere, an important reminder that we do NOT support closed-weight models.

if this is a model to be released, I honestly suggest removing the "interchunk_merge" paligemma-style. otherwise, we won't allow this much code added just to support a single model

tools/mtmd/mtmd.h
avirajBevli: Thanks for the feedback. 
I can implement an alternate strategy which should be slightly less invasive to the existing codebase: 

Move the prefix-token merging into the existing image chunk processing path (inside mtmd_helper_decode_image_chunk / mtmd_helper_eval_chunk_single) rather than creating a separate eval loop. The prefix token IDs would be stored as metadata on the image chunk, and prepended to image embeddings during the standard image decode. So no separate interchunk_merge API needed. 

Would this approach be acceptable?

ggml/src/ggml-metal/ggml-metal-ops.cpp
avirajBevli, line 1: This guard prevents mul_mm (which uses FP16 intermediates) from being selected when GGML_PREC_F32 is explicitly requested and both inputs are F32. Without this, Falcon OCR's Squared ReLU activation produces values exceeding 65504 (FP16 max), causing overflow in the FFN down-projection on Metal. 

Note that the GGML_PREC_F32 flag is already set by llama-graph.cpp for affected architectures (GLM4, JAIS2, Falcon OCR), but the issue is that Metal's mul_mm path was ignoring it.

Note: this is a ggml-level fix. Please let me know if this should be submitted as a separate PR to ggml, or if it's fine to include here.

include/llama.h
```diff
@@ -572,6 +572,10 @@ extern "C" {
     // Returns label of classifier output by index (<n_cls_out). Returns nullptr if no label provided
     LLAMA_API const char * llama_model_cls_label(const struct llama_model * model, uint32_t i);
 
+    // Copy the embedding vector for the given token into output (must be at least n_embd floats)
+    // Returns false if the token is out of range or if the model has no token embeddings
+    LLAMA_API bool llama_model_token_to_embd(const struct llama_model * model, llama_token token, float * output);
+
```
ggerganov: This should be moved to `llama-ext.h` for the time being. This is the "staging" header that we will be using for "new"/"experimental" `libllama` API until we are sure it is formulated correctly.

Also, I am planning some similar changes for supporting Eagle3 and MTP and will need to consider those functionalities when introducing this API.

ggml/src/ggml-metal/ggml-metal-ops.cpp
```diff
@@ -2143,7 +2143,9 @@ int ggml_metal_op_mul_mat(ggml_metal_op_t ctx, int idx) {
         !ggml_is_transposed(op->src[1]) &&
         // for now the matrix-matrix multiplication kernel only works on A14+/M1+ SoCs
         // AMD GPU and older A-chips will reuse matrix-vector multiplication kernel
-        props_dev->has_simdgroup_mm && ne00 >= 64 && ne11 > ne11_mm_min) {
+        props_dev->has_simdgroup_mm && ne00 >= 64 && ne11 > ne11_mm_min &&
+        !(ggml_get_op_params_i32(op, 0) == GGML_PREC_F32 &&
+          op->src[0]->type == GGML_TYPE_F32 && op->src[1]->type == GGML_TYPE_F32)) {
```
ggerganov, line 2148: What is the reason to exclude the F32 path here? Did you observe a slowdown, or?

avirajBevli: We observed that our model produces gibberish results on metal. The reason for this is Falcon OCR's Squared ReLU activation produces values exceeding 65504 (FP16 max), causing overflow. 

This guard prevents mul_mm (which uses FP16 intermediates) from being selected when GGML_PREC_F32 is explicitly requested and both inputs are F32. With this guard, the matmul in metal happens in FP32.

Note that the GGML_PREC_F32 flag is already set by llama-graph.cpp for affected architectures (GLM4, JAIS2, Falcon OCR), but the issue is that Metal's mul_mm path was ignoring it (because there is casting to FP16 intermediates).

tools/mtmd/mtmd.h
avirajBevli: No, this is not standard mrope. 
Although our model reuses the M-RoPE 4D position layout for convenience, our spatial_3d_rope is different. 

- The spatial rotation uses learned per-head frequencies (stored as a rope_freqs_golden tensor) rather than the fixed geometric schedule. 
- The first half of head_dim gets standard 1D RoPE via ggml_rope_ext, and the second half gets a custom 2D rotation with the learned frequencies. 

This is why it can't go through the existing M-RoPE code path.

tools/mtmd/mtmd.h
ngxson: > The prefix token IDs would be stored as metadata on the image chunk

you still need to add another API to add metadata to chunk, so I don't think it will be less invasive.

and downstream code will need to be updated in that case too, even more complicated changes.

---

the main question is: because this model is not yet to be released, **is it possible for you to just remove this "interchunk_merge" when training the model?**

the same thing happen to google's gemma: paligemma processes text+image in the same non-causal batch, it was not a good idea. they then remove it on gemma 3, and only image is non-causal; text prompt still causal.

avirajBevli: 1) I just pushed a new commit that addresses this. Rather than adding a new API, I took a simpler approach:
**Removed** mtmd_decode_use_interchunk_merge, eval_chunks_falcon_ocr, and falcon_ocr_eval_image_with_prefix entirely (~135 lines deleted from mtmd-helper.cpp).
I added the field prefix_tokens inside mtmd_image_tokens (an internal struct, no new public API needed). So I am now storing the prefix token IDs inside mtmd_image_tokens and prepending their embeddings during encoding in mtmd_encode(). Note that this prepending token IDs step does not affect any other models because falcon OCR is the only model with non-empty prefix_tokens.


2) The model has already been trained with cls+reg tokens sharing bidirectional attention with image patches. Retraining would require significant compute and is not something we can do quickly. 
That said, I understand the concern. With the latest commit, I tried to address this and now, there is just 15 lines of model-specific code in mtmd.cpp. Please let me know if that is too much for a single model


3) Regarding model availability: the weights are planned for public release soon. I understand this is a prerequisite for merging.

ngxson: honestly say, if you just push AI-generated code here, this discussion is not going any where

I refuse to review this further if my point above cannot be discussed properly.

ngxson:
A general note that most llama.cpp's maintainers already had AI platforms subscriptions and we can generate the same code that you are pushing here easily.

For that reason, we value more discussions about technical aspects and API design. It is not productive for us to read a predominantly AI-generated PR.

tools/mtmd/mtmd.h
avirajBevli: Apologies if I didn't explain myself well in my previous replies. 

Just to make sure we follow each other, this is my understanding of our dicussion till now:

1) Your question was: "because this model is not yet to be released, is it possible for you to just remove this "interchunk_merge" when training the model?".
To address the question concretely, we have already trained the model using text+image non-causal. It will be difficult for us to retrain using text causal + image non-causal at this moment.


2) The original pain point highlighted by you was that there was too mcuh code added to support the falcon OCR model (about 150 lines of model-specific-code to handle the interchunk merging logic).
Note that this interchunk merging logic was required because there was a separate text chunk for the image registration tokens, which was a part of non-causal attention with the image chunk.
I tried to address this by changing the image chunk to now also contain information about the image registration text tokens. So now there is no separate text chunk for the image registration tokens, and hence we don't need to have any inter-chunk merging logic. I think this is an elegant solution to address the original issue you raised and hence, I pushed my latest commit.

Do you think that the current code still has too mcuh model-specific code that you want to avoid? :
- If yes, then other than retraining the model, are there any next steps I can follow?
- If no, then what is the current pain point that I should be addressing for a successful PR?


{
Additional note: 
Please note that the fundamental reason we can not naiively use the current infrastructure of llama.cpp is because of our model has been trained with non-causal on image reg tokens + image tokens (like paligemma). llama.cpp currently only has support for models with causal on image reg tokens and non-causal on image tokens (eg: gemma3). 
This is why we have to make changes to the llama.cpp codebase outside of our own model's code to integrate our model into llama.cpp. This is unavoidable
}

include/llama.h
avirajBevli: llama-ext.h currently does not exist on the master branch. Should I create the file and move llama_model_token_to_embd to that file as a part of this PR?

YasserdahouML:
@ngxson thanks for the feedback, i get it. once this lands it becomes your maintenance issues, and right now it’s a lot of new code / new modes for a model that isn't yet introduced. We will release over the next few days and will come back to you with all the required materials to find the best way to support our arch within the best practices of llama.cpp if possible, thanks

ggml/src/ggml-metal/ggml-metal-ops.cpp
ggerganov: Just to confirm, this is the multiplication that is problematic?

https://github.com/ggml-org/llama.cpp/blob/d0a856895f553f099303823cc7da3f368b88130e/src/llama-graph.cpp#L1135-L1141

include/llama.h
ggerganov: The header is in `src/llama-ext.h`. We can get back to this change later on, don't worry about it for now.

ggml/src/ggml-metal/ggml-metal-ops.cpp
avirajBevli: yes. this one

include/llama.h
avirajBevli: okay sure

tools/mtmd/mtmd.h
ngxson: > geometric schedule

wow, new tech, what is that?

> The first half of head_dim gets standard 1D RoPE via ggml_rope_ext, and the second half gets a custom 2D rotation with the learned frequencies.

it is M-RoPE...

src/models/falcon_ocr.cpp
```diff
@@ -0,0 +1,258 @@
+#include "models.h"
+
+#include <vector>
+
+// Graph input for 2D spatial positions used by the golden RoPE.
+// With rope_3d, ubatch carries 3 position dimensions per token:
+//   pos[i]              = temporal (used by standard 1D RoPE)
+//   pos[i + n_tokens]   = h spatial pos (fixed-point, scale 1e6)
+//   pos[i + 2*n_tokens] = w spatial pos (fixed-point, scale 1e6)
+// For text tokens, hw_data stays zero -> identity rotation.
```
ngxson, line 10: m-rope also uses `[t, h, w]`, the whole image takes one single position `t`

tools/mtmd/mtmd.cpp
```diff
@@ -637,8 +647,18 @@ struct mtmd_tokenizer {
...
+            std::vector<llama_token> img_beg_prefix_tokens;
             if (!ctx->img_beg.empty()) {
-                add_text(ctx->img_beg, true); // add image begin token
+                if (ctx->proj_type_v() == PROJECTOR_TYPE_FALCON_OCR) {
+                    img_beg_prefix_tokens = mtmd_tokenize_text_internal(
+                        vocab, ctx->img_beg, false, true);
+                } else {
+                    add_text(ctx->img_beg, true);
+                }
             }
```
ngxson: you are just over-complicate the whole thing here.

if the `img_beg_prefix_tokens` is fix, why don't you extract these tokens and saved them as a fixed tensor? and better, do it upon conversion, so only add less than 10 lines of C++ to handle that

src/models/falcon_ocr.cpp
ngxson: scale to `1e6` can be done via `freq_scales`

tools/mtmd/mtmd.h
ngxson: > If no, then what is the current pain point that I should be addressing for a successful PR?

Now that you agree you reinvented the paligemma style, look around to be sure if you are not re-inventing the same thing that other models already had. I'm not convinced that your 3D rope and the "golden" things are new.

src/models/falcon_ocr.cpp
```diff
@@ -0,0 +1,258 @@
...
+    sin_t = ggml_reshape_4d(ctx0, sin_t, 1, F, n_head, n_tokens);
+
+    // out_r = x_r*cos - x_i*sin, out_i = x_r*sin + x_i*cos
+    ggml_tensor * out_r = ggml_sub(ctx0,
+            ggml_mul(ctx0, x_real, cos_t),
+            ggml_mul(ctx0, x_imag, sin_t));
+    ggml_tensor * out_i = ggml_add(ctx0,
+            ggml_mul(ctx0, x_real, sin_t),
+            ggml_mul(ctx0, x_imag, cos_t));
+
```
ngxson, line 97: I can already smell some over-engineer here. In many cases, GGML RoPE kernel can already do this under the hood. It's just matter of converting / swapping the tensor to the right ordering that we supports, can be done at conversion time

tools/mtmd/mtmd.cpp
avirajBevli: oh yes, this is a much simpler approach. I will implement this fix. Thanks

tools/mtmd/mtmd.h
avirajBevli: Not exactly. We have the following differences which are making it difficult for me to use existing MRope infrastructure in llama.cpp codebase:
 
1) The spatial frequencies are **learned** rather than calculated using the standard theta = base(-2i/d) formula. So if I were to use the ggml_rope_multi function, I would have to pass in the frequencies tensor to the function, which is not supported currently. Also, freq_factors only divides the geometric theta; it doesn't replace it.

2) The learned spatial frequencies are different for each head

3) MRope uses sections[] to assign each dim pair to one position. Here each section operates on position independently. In our model, this is different because we compute theta = freq_h * pos_h + freq_w * pos_w. i.e. In our model, each dim pair sees a weighted combination of both h and w

4) ggml_rope_multi function expects input positions to be GGML_TYPE_I32. However our spatial positions are continuous floating point values because we apply normalization to coordinates.

It is possible that there are gaps in my understanding. Please let me know if that is the case! 
Happy to discuss if there's a way to restructure this that I'm not seeing.

src/models/falcon_ocr.cpp
avirajBevli: What you are saying is true. But the issue is that the ggml_rope_multi function (responsible for MRope) seems not usable for our model because of the reasons mentioned here: https://github.com/ggml-org/llama.cpp/pull/21045#discussion_r3004950320

tools/mtmd/mtmd.cpp
avirajBevli: Fixed this in the latest commit. 

Original flow: 
Prefix tokens were a field in the mtmd_image_tokens
During runtime, Image registration tokens tokenized-> Embeddings calculated for them during image encoding (mtmd_encode) 

New flow: 
During conversion to .gguf : Calculate image registration token embeddings and store them as a tensor 
During runtime : Just retrieve the embedding tensor and prepend to image patch embeddings.

So now no need for llama_model_token_to_embd. This addresses https://github.com/ggml-org/llama.cpp/pull/21045#discussion_r2999315656

Please let me know if this is okay

tools/mtmd/mtmd.h
avirajBevli: @ngxson the model is released

Please refer to the following links for more details: 
https://huggingface.co/tiiuae/Falcon-OCR
https://github.com/tiiuae/falcon-perception

YasserdahouML:
@ngxson the model has been released, will you look at this PR?

ngxson:
Sorry @avirajBevli for being quite hard on the review. I've been quite busy during the gemma 4 release. Will have a look on in the next week or the week after

tools/mtmd/mtmd.h
ngxson: @avirajBevli I had a deeper look into your impl and I kinda disagree with what you said above. Would be appreciated if you can confirm what I'm saying here is correct:

> Also, freq_factors only divides the geometric theta; it doesn't replace it.

Yes, it can. Currently, you calculate the theta by:

```cpp
ggml_tensor * theta = ggml_mul_mat(ctx0, freqs_flat, pos_hw);
// pos_hw is the scaled position data
```

If you look into `rope.cu`, theta is calculated in GGML by:

```cpp
const float theta_base = pos[i2]*powf(theta_scale, i0/2.0f);
const float freq_factor = has_ff ? freq_factors[i0/2] : 1.0f;
// final theta = theta_base/freq_factor
```

So what that means is that if `theta_scale` (aka `freq_base`) is 1.0f, then `powf(theta_scale, i0/2.0f)` will also be 1.0f; you can then provide the **inverse** of your desired freq via `freq_factors`

And you don't even need to pre-scale the position value via `POS_SCALE_INV = 1.0f / 1000000.0f;`, simply pre-scale your `freq_factors`

And since mrope kernel works the same way, you can just reuse mrope with `freq_base = 1.0f` and your custom `freq_factors`

ngxson: > MRope uses sections[] to assign each dim pair to one position. Here each section operates on position independently. In our model, this is different because we compute theta = freq_h * pos_h + freq_w * pos_w. i.e. In our model, each dim pair sees a weighted combination of both h and w

Would it possible to firstly rotate by `theta0 = freq_h * pos_h` , then rotate again with `theta1 = freq_w * pos_w` ?

In other words, that means calling `ggml_rope_ext` twice, something like:

```cpp
x = ggml_rope_ext(x, pos_h, freqs_h);
x = ggml_rope_ext(x, pos_w, freqs_w);
```

avirajBevli: @ngxson thank you for having an deeper look. 

> Would it possible to firstly rotate by theta0 = freq_h * pos_h , then rotate again with theta1 = freq_w * pos_w 

This should be doable: mathematically equivalent.

> So what that means is that if theta_scale (aka freq_base) is 1.0f, then powf(theta_scale, i0/2.0f) will also be 1.0f; you can then provide the inverse of your desired freq via freq_factors

This simple trick let's me provide my learned freqs via freq_factors. This works!

Thanks for both the insights!


So now in my understanding, the only blocker to not being able to directly use the existing rope infra in llama.cpp for falcon ocr model is the fact that our model has separate learned frequency for each head. 
In `falcon_ocr.cpp`, `rope_freqs_golden` tensor is `[2, head_dim/4, n_head]` — each head has its own unique set of learned frequencies

While the current rope infrastructure in llama.cpp expects the same freq_factor for all heads. This is because in rope.cu
`const float freq_factor = has_ff ? freq_factors[i0/2] : 1.0f;`

freq_factors is indexed only by dimension pair (i0/2), with no head dimension (i1) — so the same value must apply to all heads

Can you please confirm this?

ngxson: > While the current rope infrastructure in llama.cpp expects the same freq_factor for all heads. This is because in rope.cu
> `const float freq_factor = has_ff ? freq_factors[i0/2] : 1.0f;`

Yes that is correct. However, one more idea is that you can flatten the head, for example:

[head_dim, n_head, n_token] --> [hidden_size, 1, n_token]

Then apply `ggml_rope_ext` with the flatten `freq_cis_golden`. That should be mathematically equivalent to apply per-head frequency.

I'm working on the implementation of falcon-ocr on my fork, with some extra clean up, the link is here: https://github.com/ngxson/llama.cpp/pull/100

Still something is still wrong and the model still generates `<file_sep>` tokens non-stop. So, it would be appreciated if you have a look on the mentioned PR. Feel free to push a PR to my fork (base branch `xsn/falcon-ocr`) if you found the fix, thanks in advance!

tools/mtmd/mtmd-image.cpp
```diff
@@ -1110,6 +1110,71 @@ bool mtmd_image_preprocessor_deepseekocr::preprocess(const clip_image_u8 & img,
     return true;
 }
 
+//
+// mtmd_image_preprocessor_falcon_ocr
+//
+
+bool mtmd_image_preprocessor_falcon_ocr::preprocess(const clip_image_u8 & img, clip_image_f32_batch & output) {
```
ngxson, line 1117: Note that this was replaced by `mtmd_image_preprocessor_dyn_size` on https://github.com/ngxson/llama.cpp/pull/100 , because it does the same thing: resize the image to a fixed pixel budget and align to patch size

Please let me know if my replacement is correct.

tools/mtmd/mtmd-helper.cpp
```diff
@@ -174,6 +174,49 @@ struct decode_embd_batch {
...
+                const float w_pos = (nx > 1)
+                    ? -xlim + x * (2.0f * xlim) / (nx - 1)
+                    : 0.0f;
+                int i = n_prefix + y * nx + x;
+                pos[i            ] = pos_0;
+                pos[i + n_total  ] = (llama_pos)(h_pos * SPATIAL_3D_POS_SCALE);
+                pos[i + n_total*2] = (llama_pos)(w_pos * SPATIAL_3D_POS_SCALE);
+                pos[i + n_total*3] = 0;
+            }
+        }
```
ngxson, line 212: Question here: `pos_0` never seems to be increased, I suspect there should be a bug in this code.

If I understand correctly, if we have 3 text tokens followed by 3 image tokens, then the temporal should be linear for text (0, 1, 2) and then fixed for image (4, 4, 4)

I implemented the mentioned fix on https://github.com/ngxson/llama.cpp/pull/100 , inside `mtmd_image_tokens_get_decoder_pos`

tools/mtmd/mtmd.cpp
ngxson, line 1: note that `mtmd_image_tokens_get_n_pos()` may also need to be modified (done on https://github.com/ngxson/llama.cpp/pull/100)

on M-RoPE models, this is to count the number of **temporal** positions (not the number of tokens or KV slots) for a given image. for M-RoPE, an image takes `max(nx,ny)` positions

for falcon-ocr, it seems like the positions is equal to `n_boi + 1`, with `n_boi` is the `n_prefix` in your version here, and plus 1 because the whole image takes one single temporal position. please correct if I'm wrong here

src/llama-hparams.h
```diff
@@ -218,6 +218,10 @@ struct llama_hparams {
     enum llama_rope_type         rope_type               = LLAMA_ROPE_TYPE_NONE;
     enum llama_rope_scaling_type rope_scaling_type_train = LLAMA_ROPE_SCALING_TYPE_NONE;
 
+    bool rope_3d = false;
+
+    // Original KV head count before GQA expansion (e.g. Falcon OCR)
+    uint32_t n_head_kv_orig = 0;
```
ngxson, line 224: probably we should assume the model to have `n_head == n_head_kv` as the `n_head_kv` is mostly used by KV cache implementation (and we do want to have K dim == Q dim here because of the golden rope)

I implemented the solution above on my fork, which makes the code a bit less hacky

include/llama.h
ngxson: (resolving this thread because we no longer need this API, yay)

tools/mtmd/mtmd.h
avirajBevli: > Yes that is correct. However, one more idea is that you can flatten the head, for example:
> 
> [head_dim, n_head, n_token] --> [hidden_size, 1, n_token]
> 
> Then apply ggml_rope_ext with the flatten freq_cis_golden. That should be mathematically equivalent to apply per-head frequency.

Yes. this makes sense!

> I'm working on the implementation of falcon-ocr on my fork, with some extra clean up, the link is here: https://github.com/ngxson/llama.cpp/pull/100
Thank you for creating the fork. Really excited to get this going forward!

> Still something is still wrong and the model still generates <file_sep> tokens non-stop. So, it would be appreciated if you have a look on the mentioned PR. Feel free to push a PR to my fork (base branch xsn/falcon-ocr) if you found the fix, thanks in advance!

Yeah, when I run inference on your branch in your fork, I get garbage inference results. In my branch, the results are correct. 

Let me find the root cause and update here when I find the fix! Thanks

tools/mtmd/mtmd-image.cpp
avirajBevli: Please note that this is not fully correct. 

Falcon ocr does preprocessing in 2 steps: 
1) Fit image into [min_edge_dim, max_edge_dim] while preserving aspect ratio 
2) Fit image into [min_pixel_count, max_pixel_count] while preserving aspect ratio and align to patch size 

In contrast, seems like `mtmd_image_preprocessor_dyn_size` does only the second step. 

Essentially, step 1 is an extra filtering to add a per-dimension clamping on images with extreme aspect ratios

tools/mtmd/mtmd-helper.cpp
avirajBevli: > Question here: pos_0 never seems to be increased, I suspect there should be a bug in this code.

I implemented this way by design because consider a sample prompt, which is tokenized to produce the following tokens (in order): 
<image_CLS> <image_reg1> ... <image_reg4>.... list of image patch tokens... <EOI> <text1> <text2>... <text n>

Following are the intended positions for this sequence:
image

Please refer to this for the temporal position calculation in HF:
https://huggingface.co/tiiuae/Falcon-Perception/blob/main/processing_falcon_perception.py#L310-L312

Please note that this is different from the temporal positions you mention in:
https://github.com/ggml-org/llama.cpp/pull/21851

This might be a source of misunderstanding

tools/mtmd/mtmd.cpp
avirajBevli: This is also related to the fact that the 5 image cls + reg tokens share the same temporal position (as mentioned in the previous comment). 

Please refer to this for the temporal position calculation in HF:
https://huggingface.co/tiiuae/Falcon-Perception/blob/main/processing_falcon_perception.py#L310-L312

Please note that this is different from the temporal positions you mention in:
https://github.com/ggml-org/llama.cpp/pull/21851

src/llama-hparams.h
avirajBevli: Makes sense!

blap:
.

PR #21049, opened 2026-03-26
completion: fix segfault on model load failure

mtmcp:
## Overview

When common_init_from_params fails to load a model (e.g. invalid path),
  the returned common_init_result contains null members. Calling
  context() and model() safely return null, but sampler() dereferences
  an internal unique_ptr that is null, causing a segfault.

  Move the sampler access after null checks for ctx and model so we
  exit cleanly instead of crashing.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/tools/completion/completion.cpp
+++ b/tools/completion/completion.cpp
@@ -146,12 +146,18 @@ int main(int argc, char ** argv) {
 
     ctx   = llama_init->context();
     model = llama_init->model();
-    smpl  = llama_init->sampler(0);
 
     if (ctx == NULL) {
         LOG_ERR("%s: error: unable to create context\n", __func__);
         return 1;
     }
+
+    if (model == NULL) {
+        LOG_ERR("%s: error: unable to load model\n", __func__);
+        return 1;
+    }
+
+    smpl = llama_init->sampler(0);
 
     llama_memory_t mem = llama_get_memory(ctx);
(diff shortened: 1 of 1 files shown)


mtmcp:
Hi @ggerganov I renamed my GitHub handle from @bandoti (to promote it to organization) and my collab status was preserved but getting blocked from multiple PRs. Happy to wait it out if it just requires a single merge. I have these two for-the-most-part one-liners for review.

CISC (approved):

ggerganov (approved):

CISC:
> Calling context() and model() safely return null, but sampler() dereferences an internal unique_ptr that is null, causing a segfault.

That should probably also be fixed to fail cleanly.

ggerganov:
> getting blocked from multiple PRs

Could you clarify?

CISC:
> > getting blocked from multiple PRs
> 
> Could you clarify?

I think just referring to this https://github.com/ggml-org/llama.cpp/pull/21049#issuecomment-4139088852

mtmcp:
> That should probably also be fixed to fail cleanly.

@CISC Yeah good point. I will step through a bit deeper and see if there's another part to the fix.

> Could you clarify?

@ggerganov Sorry I should clarify that the PR bot didn't recognize my prior contributions due to the name change from bandoti to mtmcp, and so it limited me to a single PR.

PR #21051, opened 2026-03-27
Enable testing on Snapdragon devices

shreyajn:
## Overview

Run basic tests for android hexagon when a change is made in the code path.

# Requirements

- AI usage disclosure: Some use of AI to plan the change and figure out the CI issue.


max-krasnyansky (commented):
@shreyajn 
Should we create `build-snapdragon.yml` instead of updating the `build-android`?
We're going to cover Windows and eventually Linux here as well, right?

Also I'd rename hexagon to snapdragon here since we're covering all backends.

shreyajn:
> @shreyajn Should we create `build-snapdragon.yml` instead of updating the `build-android`? We're going to cover Windows and eventually Linux here as well, right?
> 
> Also I'd rename hexagon to snapdragon here since we're covering all backends.

Sounds good

max-krasnyansky (approved):
Looks good to me. 
It's been on our TODO list to enable this for some time now!

@CISC 
Please take a look and let us know if you have any suggestions/concerns.

scripts/snapdragon/qdc/run_qdc_jobs.py
```diff
@@ -0,0 +1,405 @@
+# ---------------------------------------------------------------------
+# Copyright (c) 2025 Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+# ---------------------------------------------------------------------
```
CISC: Can we get rid of these (were they inherited from something or just added as part of the PR)?

CISC (commented):
Generally looks ok, guess @ggerganov needs to weigh in on if it's ok to dispatch cloud jobs like this (and adding the API key).

scripts/snapdragon/qdc/run_qdc_jobs.py
max-krasnyansky: Ah. Yeah, sorry I missed that one in my review. We'll remove it.

max-krasnyansky:
> Generally looks ok, guess @ggerganov needs to weigh in on if it's ok to dispatch cloud jobs like this (and adding the API key).

He already gave me an OK on the API KEY (I already added it a month or so back).
Otherwise yep, would be good to confirm. @ggerganov when you get a minute.

scripts/snapdragon/qdc/run_qdc_jobs.py
shreyajn: I'll remove it. I internally tested the file and it was added as part of that. Sorry!

ggerganov (approved):

.github/workflows/build-and-test-snapdragon.yml
```diff
@@ -0,0 +1,113 @@
...
+      has-key: ${{ steps.check.outputs.has-key }}
+    steps:
+      - id: check
+        run: echo "has-key=${{ secrets.QDC_API_KEY != '' }}" >> "$GITHUB_OUTPUT"
+
+  test-snapdragon-qdc:
+    name: Test on QDC Android Device (${{ matrix.device }})
+    needs: [android-ndk-snapdragon, check-secret]
+    if: needs.check-secret.outputs.has-key == 'true'
+    runs-on: ubuntu-latest
```
CISC, line 73: I wonder, can this be run on `ubuntu-slim`?

CISC (approved):

.github/workflows/build-and-test-snapdragon.yml
max-krasnyansky: That should work. We'll give that a go and do a follow up.
The overall CI is quite slow these days, hopefully slim images help reducing the runner load.
(i.e I see that the slim image gets a single core / reduced ram machine
https://docs.github.com/en/actions/reference/runners/github-hosted-runners)

@shreyajn when you get a minute take a look at the ubuntu-slim image 
https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md

CISC:
Hmmm, what is this?
https://github.com/ggml-org/llama.cpp/actions/runs/24856337776/job/72770062720

shreyajn:
> Hmmm, what is this? https://github.com/ggml-org/llama.cpp/actions/runs/24856337776/job/72770062720

I can't find this in our GitHub workflow. Is it a new Dependabot?

max-krasnyansky:
> > Hmmm, what is this? https://github.com/ggml-org/llama.cpp/actions/runs/24856337776/job/72770062720
> 
> I can't find this in our GitHub workflow. Is it a new Dependabot?

Yep, triggered by dependabot. I'm not familiar with its config.

> Triggered via dynamic 1 hour ago
> @dependabot[dependabot[bot]](https://github.com/apps/dependabot)
> ⁠
> [ 187a456](https://github.com/ggml-org/llama.cpp/commit/187a45637054881ecacf17f8e2f6f8f2ba7df1c7)
> [master](https://github.com/ggml-org/llama.cpp/tree/refs/heads/master)

CISC:
I guess @ggerganov enabled it? Never seen it here before.

CISC:
Oh, this explains it. :)
https://github.blog/changelog/2026-04-23-dependabot-graphs-for-python/

ggerganov:
I didn't enable it - probably auto-enabled. Do we need this, or should I look how to disable it?

CISC:
> I didn't enable it - probably auto-enabled. Do we need this, or should I look how to disable it?

No, it looks useful, just a little jumpy when some unknown CI workflow appears out of nothing. :)

ggerganov:
@max-krasnyansky This workflow seems to be inactive: https://github.com/ggml-org/llama.cpp/actions/workflows/build-and-test-hexagon.yml

image

PR #21052, opened 2026-03-27
ggml-webgpu: fix compiler warnings and refactor FlashAttention encoding

reeselevine:
## Overview

This PR doesn't add new functionality, but does the following:
* Removes compiler warnings due to usage of C++20 initializers and potentially unsafe casting, which cleans up the compilation and is a step towards enabling CI on the ggml NVIDIA machine
* Refactors flashattention encoding to avoid custom structs and be more in-line with encoding of the rest of the operations
* I also added a block in `build-self-hosted.yml` for running the CI on the ggml NVIDIA machine. Unfortunately it doesn't pass right now due to numerical precision issues in `EXP/EXPM1/DIV` (even when using `f32` for the division and then downcasting to `f16`), and `SET_ROWS` (even though this should be a straightforward cast from `f32` to `f16`). My plan is to investigate those imprecisions a little more closely to see if it's something I can fix, if it's an issue with the Dawn WebGPU implementation, or if we might want to relax the NMSE checks slightly for these operations. I'd like to get this PR merged without enabling the CI though so that the refactoring changes are in. See errors here:
  * [EXP/EXPM1](https://github.com/ggml-org/llama.cpp/actions/runs/24368468752/job/71166201498#step:4:1594): appears to be an issue where the infinite f32 values are getting cast to the max f16 value by WebGPU
  * [DIV](https://github.com/ggml-org/llama.cpp/actions/runs/24368468752/job/71166201498#step:4:7207): slightly over NMSE threshold for some DIV operations
  * [SET_ROWS](https://github.com/ggml-org/llama.cpp/actions/runs/24368468752/job/71166201498#step:4:2052): very slightly over the threshold on NMSE for some SET_ROWS operations

# Requirements

- AI usage disclosure: no


reeselevine:
The failing `ubuntu-24-webgpu` on the [ec783c1](https://github.com/ggml-org/llama.cpp/pull/21052/commits/ec783c15137a82748906d40be4b802bb6b758f85) commit (and the same failures on the master branch currently) were due to a bad ccache. Not sure if there's a way to avoid that or clear it, I manually ssh'ed into the Github runner and cleared it.

ggerganov (approved):

CISC (approved):

CISC:
> The failing `ubuntu-24-webgpu` on the [ec783c1](https://github.com/ggml-org/llama.cpp/pull/21052/commits/ec783c15137a82748906d40be4b802bb6b758f85) commit (and the same failures on the master branch currently) were due to a bad ccache. Not sure if there's a way to avoid that or clear it, I manually ssh'ed into the Github runner and cleared it.

You can find and delete the bad cache here: https://github.com/ggml-org/llama.cpp/actions/caches

Unfortunately this happens from time to time, I have yet to find any specific cause for this, and just clean out the cache whenever it happens...

PR #21054, opened 2026-03-27
cuda: fall back to VEC attention when quantized K/V F16 scratch exceeds free VRAM

lucaspirola:
## Problem

The MMA flash attention path on ADA Lovelace+ GPUs pre-converts quantized K/V tensors (`q4_0`, `q8_0`, `bf16`) to full-precision F16 scratch buffers before running the kernel (`K_f16` + `V_f16` in `launch_fattn`). The scratch size is proportional to the KV sequence length:

```
scratch = 2 × ggml_nelements(K) × sizeof(half)
        = 2 × (head_dim × n_kv × n_kv_heads) × 2 bytes
```

On VRAM-constrained setups this grows unboundedly during prefill and eventually exhausts device memory, producing a hard crash inside the VMM pool allocator.

## How to reproduce

1. Load a large model leaving ≤ ~50 MiB of free device VRAM after model + KV cache + compute buffer (e.g. a ~14 GiB model on a 16 GiB GPU with a 1 GiB quantized KV cache)
2. Use a quantized KV cache: `-ctk q4_0 -ctv q4_0` (or `q8_0` / `bf16`)
3. Run a long prefill (threshold depends on model's GQA ratio and head dim, typically > 8 000 tokens)
4. **Before fix:** `CUDA error: out of memory` inside `ggml_cuda_pool_vmm::alloc` during flash attention

Concrete example: RTX 5080 (16 GiB), Q4\_K\_M 30B MoE model (~14.2 GiB), `-ctk q4_0 -ctv q4_0 -c 40960`. Only ~34 MiB of VRAM remains free after load. The MMA scratch grows 2 MiB per 2048-token KV step; around n\_kv ≈ 8 000 the scratch exceeds remaining VRAM and `cuMemCreate` fails.

## Fix

Check free device memory via `cudaMemGetInfo` before selecting the MMA path. If `K_f16 + V_f16` would not fit in available VRAM, fall back to `BEST_FATTN_KERNEL_VEC`, which reads quantized K/V data natively without any temporary buffer.

The check is only reached for quantized K/V types on ADA Lovelace+, so the hot path for F16/F32 KV is unaffected. The `cudaMemGetInfo` call is inexpensive relative to the attention kernel launch overhead.

## Effect

- VRAM-tight setups: adaptive MMA→VEC fallback as context grows; no crash
- Normal setups with plenty of free VRAM: MMA path used throughout (no behaviour change)
- Generation (single token, `Q->ne[1] == 1`): already uses VEC via the existing `Q->ne[1] <= 2` check; unaffected

### DIFF
--- a/ggml/src/ggml-cuda/fattn.cu
+++ b/ggml/src/ggml-cuda/fattn.cu
@@ -395,6 +395,18 @@ static best_fattn_kernel ggml_cuda_get_best_fattn_kernel(const int device, const
                     if (Q->ne[1] <= 2) {
                         return BEST_FATTN_KERNEL_VEC;
                     }
+                    // The MMA path must convert quantized K/V to F16 in VRAM before processing.
+                    // For large contexts this scratch space can exhaust available VRAM. Fall back
+                    // to the VEC path which reads quantized K/V directly without any temporary buffer.
+                    {
+                        const size_t kv_f16_bytes = (size_t)ggml_nelements(K) * sizeof(half)
+                                                  + (size_t)ggml_nelements(V) * sizeof(half);
+                        size_t free_vram = 0, total_vram_unused = 0;
+                        CUDA_CHECK(cudaMemGetInfo(&free_vram, &total_vram_unused));
+                        if (kv_f16_bytes > free_vram) {
+                            return BEST_FATTN_KERNEL_VEC;
+                        }
+                    }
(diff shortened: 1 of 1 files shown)


am17an:
Create an issue rather than fixing it yourself if you faced a crash. I don't think this is the real reason

PR #21055, opened 2026-03-27
cuda: add MXFP4 KV cache support for flash attention

lucaspirola:
> **Note:** I'm not a programmer — this was developed and fixed by Claude Code (Anthropic's AI coding assistant). I'm submitting it because it adds something I believe is genuinely useful and fixes a real crash I encountered.

## Summary

Adds CUDA flash attention VEC kernel support for `GGML_TYPE_MXFP4` key/value caches, enabling `-ctk mxfp4 -ctv mxfp4` on CUDA backends.

MXFP4 is a block-floating-point format (4 bits/element, 1 E8M0 scale per 32 elements) that's already implemented in ggml's quantize/dequantize, but had no CUDA flash attention path. This adds it.

### Changes

**`fattn-common.cuh`**
- `vec_dot_fattn_vec_KQ_mxfp4<D, nthreads>`: DP4A-based KQ dot product reading `block_mxfp4` blocks via 4-bit nibble table lookup
- `dequantize_V_mxfp4<T, ne>`: V dequantization for the VEC accumulator loop
- Both wired into `get_vec_dot_KQ` / `get_dequantize_V` constexpr selectors

**`fattn.cu`**
- `ggml_cuda_get_best_fattn_kernel`: returns `BEST_FATTN_KERNEL_VEC` immediately if either K or V is MXFP4 (no MMA pre-conversion path exists for MXFP4)
- Adds `GGML_TYPE_MXFP4` to the valid-type switch so the VEC path is reachable
- Also includes the OOM guard from #21054 (VEC fallback when F16 scratch would exceed free VRAM — needed for non-MXFP4 quantized KV on ADA+)
- `ggml_cuda_flash_attn_ext_vec`: added `FATTN_VEC_CASES_ALL_D` for all MXFP4 K/V type combinations

**`cpy-utils.cuh` / `cpy.cu`**
- `quantize_f32_mxfp4_block`: GPU kernel for F32→MXFP4 (brute-force nearest-value in 16-entry table)
- `ggml_cpy_f32_mxfp4_cuda` wired into `ggml_cuda_cpy` dispatch

**`set-rows.cu`**
- MXFP4 case wired into `set_rows_cuda` (needed for KV cache prefill)

**`common.cuh`**
- `ggml_backend_cuda_context` constructor now eagerly allocates the cuBLAS handle and a fixed 4 MiB workspace before model/KV weights fill VRAM, preventing lazy OOM on tight VRAM budgets

**`common/arg.cpp`**
- `GGML_TYPE_MXFP4` added to `kv_cache_types`

**15 new template-instance `.cu` files** for all MXFP4 K/V type combinations.

## How to reproduce (the crash this fixes)

On an RTX 5080 (16 GiB) with a model that leaves ~35 MiB free VRAM:

```bash
# Without this PR: crashes with cuMemCreate OOM at n_kv ~ 12k tokens
./llama-server -m model.gguf -ngl 99 -c 40960 -ctk mxfp4 -ctv mxfp4
```

```
GGML_ASSERT: ggml_cuda_pool_vmm.alloc: failed to reserve 2097152 bytes of virtual memory
```

With this PR, mxfp4 KV uses the VEC kernel which reads quantized blocks directly with no temporary F16 buffer, so prefill at 40k context completes without OOM.

## Test plan

- [x] Tested on RTX 5080 (Ada Lovelace / SM 89), Qwen3-Coder-REAP-25B model, 40k context, `-ctk mxfp4 -ctv mxfp4`
- [x] 20k-token prefill completes without crash (previously crashed at ~12k tokens)
- [x] Output quality appears consistent with q8_0 KV cache baseline
- [ ] CI build with `GGML_CUDA_FA_ALL_QUANTS=ON` to verify all template instances compile

## Related

- Depends on / includes fix from #21054 (OOM guard for quantized KV on ADA+)

🤖 Developed with [Claude Code](https://claude.ai/claude-code)


am17an:
Don't open more than 1 PR at a time

PR #21066, opened 2026-03-27
[HIP] Bump ROCm version to 7.2.1

slojosic-amd:
### Summary

Bumps the ROCm version from **7.2** to **7.2.1** across all CI/CD workflows and the ROCm Dockerfile, and adds the missing `gfx1102` GPU target to the fat-build architecture list.

#### ROCm 7.2 → 7.2.1

| File | What changed |
|------|-------------|
| `.devops/rocm.Dockerfile` | `ROCM_VERSION` / `AMDGPU_VERSION` args |
| `.github/workflows/build.yml` | rocWMMA package URL (`70200-43` → `70201-81`), include path |
| `.github/workflows/release.yml` | Matrix version, package URL, include path, artifact label |
| `.github/workflows/hip-quality-check.yml` | Container image tag |

#### Added GPU target

- `gfx1102` added to the fat-build arch list in `.devops/rocm.Dockerfile` and `release.yml`
  - Covers: **RX 7600 class GPUs

### Testing

- [ ] Linux ROCm build passes CI
- [ ] HIP quality check passes
- [ ] Release workflow produces `rocm-7.2.1` artifact

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -924,7 +924,7 @@ jobs:
       - name: Grab rocWMMA package
         id: grab_rocwmma
         run: |
-          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70200-43~24.04_amd64.deb"
+          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2.1/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70201-81~24.04_amd64.deb "
           7z x rocwmma.deb
           7z x data.tar
 
@@ -967,7 +967,7 @@ jobs:
           cmake -G "Unix Makefiles" -B build -S . `
             -DCMAKE_C_COMPILER="${env:HIP_PATH}\bin\clang.exe" `
             -DCMAKE_CXX_COMPILER="${env:HIP_PATH}\bin\clang++.exe" `
-            -DCMAKE_CXX_FLAGS="-I$($PWD.Path.Replace('\', '/'))/opt/rocm-7.2.0/include/" `
+            -DCMAKE_CXX_FLAGS="-I$($PWD.Path.Replace('\', '/'))/opt/rocm-7.2.1/include/" `
             -DCMAKE_BUILD_TYPE=Release `
             -DLLAMA_BUILD_BORINGSSL=ON `
             -DROCM_DIR="${env:HIP_PATH}" `
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -924,7 +924,7 @@ jobs:
       - name: Grab rocWMMA package
         id: grab_rocwmma
         run: |
-          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2.1/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70201-81~24.04_amd64.deb "
+          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2.1/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70201-81~24.04_amd64.deb"
           7z x rocwmma.deb
           7z x data.tar
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -619,7 +619,6 @@ jobs:
       matrix:
         include:
           - ROCM_VERSION: "7.2.1"
-            ROCM_VERSION_SHORT: "7.2"
             gpu_targets: "gfx908;gfx90a;gfx942;gfx1030;gfx1100;gfx1101;gfx1102;gfx1151;gfx1150;gfx1200;gfx1201"
             build: 'x64'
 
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -924,7 +924,7 @@ jobs:
       - name: Grab rocWMMA package
         id: grab_rocwmma
         run: |
-          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70200-43~24.04_amd64.deb"
+          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2.1/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70201-81~24.04_amd64.deb "
```
CISC:
```suggestion
          curl -o rocwmma.deb "https://repo.radeon.com/rocm/apt/7.2.1/pool/main/r/rocwmma-dev/rocwmma-dev_2.2.0.70201-81~24.04_amd64.deb"
```

.github/workflows/release.yml
```diff
@@ -1062,7 +1062,7 @@ jobs:
             **Linux:**
             - [Ubuntu x64 (CPU)](https://github.com/ggml-org/llama.cpp/releases/download/${{ steps.tag.outputs.name }}/llama-${{ steps.tag.outputs.name }}-bin-ubuntu-x64.tar.gz)
             - [Ubuntu x64 (Vulkan)](https://github.com/ggml-org/llama.cpp/releases/download/${{ steps.tag.outputs.name }}/llama-${{ steps.tag.outputs.name }}-bin-ubuntu-vulkan-x64.tar.gz)
-            - [Ubuntu x64 (ROCm 7.2)](https://github.com/ggml-org/llama.cpp/releases/download/${{ steps.tag.outputs.name }}/llama-${{ steps.tag.outputs.name }}-bin-ubuntu-rocm-7.2-x64.tar.gz)
+            - [Ubuntu x64 (ROCm 7.2.1)](https://github.com/ggml-org/llama.cpp/releases/download/${{ steps.tag.outputs.name }}/llama-${{ steps.tag.outputs.name }}-bin-ubuntu-rocm-7.2.1-x64.tar.gz)
```
CISC: I guess this is somewhat sub-optimal, ideally this part of the filename should stay the same for a longer period of time, until some breaking change bump, but I don't know what that would be?

slojosic-amd: I confirmed that ROCm 7.2 and 7.2.1 are ABI-compatible in both directions — artifacts built against 7.2 work with a 7.2.1 stack and vice versa. Also, ROCm 7.2.1 has fix for LLVM related perf regression with ROCm 7.2: https://github.com/ROCm/rocm-systems/issues/2865 We used this additional CMake flag as workaround: `-DCMAKE_HIP_FLAGS="-mllvm --amdgpu-unroll-threshold-local=600" \` I will delete it from build command.

Given that, I can revert the artifact filename back to rocm-7.2 and keep it stable at the minor version level, while still building internally against 7.2.1. That way the download URL stays the same for users across patch bumps. Would that be preferable?

CISC: Yes, for now that's preferable, we may need to revisit the naming in the future.

slojosic-amd (commented):
@CISC I've applied suggested changes. Thank you for your review.

CISC:
> @CISC I've applied suggested changes. Thank you for your review.

You need to remove (or strip minor version from) `matrix.ROCM_VERSION` where it's used in the file name.

.github/workflows/release.yml
```diff
@@ -618,8 +618,9 @@ jobs:
     strategy:
       matrix:
         include:
-          - ROCM_VERSION: "7.2"
-            gpu_targets: "gfx908;gfx90a;gfx942;gfx1030;gfx1100;gfx1101;gfx1151;gfx1150;gfx1200;gfx1201"
+          - ROCM_VERSION: "7.2.1"
+            ROCM_VERSION_SHORT: "7.2"
```
IMbackK: Instead of this i think it would be better to just `cut -d '.' -f 1,2` where its used.

IMbackK (changes requested):
Otherwise its fine i guess, but looking at the [updated components](https://rocm.docs.amd.com/en/latest/about/release-notes.html#rocm-components) in 7.2.1, none of these affect us at compile time so this is a pointless exercise for all builds besides the docker build.

.github/workflows/release.yml
```diff
@@ -678,7 +679,6 @@ jobs:
         run: |
           cmake -B build -S . \
             -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" \
-            -DCMAKE_HIP_FLAGS="-mllvm --amdgpu-unroll-threshold-local=600" \
```
IMbackK, line 681: llvm-project was not changed in 7.2.1, as i understand it this issue in in this package, so this issue this workaround is for still is in 7.2.1

superm1: Yes it was; here is the commit:
https://github.com/ROCm/llvm-project/commit/f58b06dce1f9c15707c5f808fd002e18c2accf7e

IMbackK: OK then the release notes are wrong and llvm was changed, not very helpfull

.github/workflows/release.yml
slojosic-amd: I investigated a little bit and it looks like the `with: path:` and `with: name:` fields in GitHub Actions only support `${{ }}` expressions, not shell commands like `$(...)` or pipes — so `cut` cannot be used there directly. I added workaround with exporting to `$GITHUB_ENV` so the value is computed once in a shell context and then available as `${{ env.ROCM_VERSION_SHORT }}` in any subsequent `with:` field.

.github/workflows/release.yml
```diff
@@ -696,17 +695,20 @@ jobs:
...
-          tar -czvf llama-${{ steps.tag.outputs.name }}-bin-ubuntu-rocm-${{ matrix.ROCM_VERSION }}-${{ matrix.build }}.tar.gz --transform "s,./,llama-${{ steps.tag.outputs.name }}/," -C ./build/bin .
+          tar -czvf llama-${{ steps.tag.outputs.name }}-bin-ubuntu-rocm-${{ env.ROCM_VERSION_SHORT }}-${{ matrix.build }}.tar.gz --transform "s,./,llama-${{ steps.tag.outputs.name }}/," -C ./build/bin .
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v6
         with:
-          path: llama-${{ steps.tag.outputs.name }}-bin-ubuntu-rocm-${{ matrix.ROCM_VERSION }}-${{ matrix.build }}.tar.gz
-          name: llama-bin-ubuntu-rocm-${{ matrix.ROCM_VERSION }}-${{ matrix.build }}.tar.gz
+          path: llama-${{ steps.tag.outputs.name }}-bin-ubuntu-rocm-${{ env.ROCM_VERSION_SHORT }}-${{ matrix.build }}.tar.gz
+          name: llama-${{ steps.tag.outputs.name }}-bin-ubuntu-rocm-${{ env.ROCM_VERSION_SHORT }}-${{ matrix.build }}.tar.gz
```
```suggestion
          name: llama-bin-ubuntu-rocm-${{ env.ROCM_VERSION_SHORT }}-${{ matrix.build }}.tar.gz
```
`path` and `name` are not the same.

IMbackK (approved):

CISC (approved):

taronaeo (approved):

PR #15346, opened 2025-08-15
sched : copy only the used experts when offloading prompt processing

slaren:
When offloading prompt processing to GPU with MoE weights on CPU, all the experts are copied to VRAM. With this change, only the experts that are actually used are copied. With models where some experts are often unused, such as gpt-oss, this can reduce significantly the amount of data that needs to be copied to VRAM.

GGML_CUDA=1 scripts/compare-commits.sh master sl/sched-used-experts llama-bench -m gpt-oss-20b-mxfp4.gguf -ot "exps=CPU" -fa 1 -n 0 -p 2048 -ub "128-2048*2"

| Model                |   Microbatch size | Test   |   t/s master |   t/s sl/sched-used-experts |   Speedup |
|:---------------------|------------------:|:-------|-------------:|----------------------------:|----------:|
| gpt-oss 20B MXFP4 MoE |               128 | pp2048 |       124.24 |                      202.92 |      1.63 |
| gpt-oss 20B MXFP4 MoE |               256 | pp2048 |       253.18 |                      372.75 |      1.47 |
| gpt-oss 20B MXFP4 MoE |               512 | pp2048 |       502.73 |                      685.31 |      1.36 |
| gpt-oss 20B MXFP4 MoE |              1024 | pp2048 |      1015.20 |                     1178.50 |      1.16 |
| gpt-oss 20B MXFP4 MoE |              2048 | pp2048 |      1629.92 |                     2089.53 |      1.28 |

GGML_CUDA=1 scripts/compare-commits.sh master sl/sched-used-experts llama-bench -m gpt-oss-120b-mxfp4-00001-of-00003.gguf -ot "exps=CPU" -fa 1 -n 0 -p 2048 -ub "128-2048*2" -r 1

| Model                |   Microbatch size | Test   |   t/s master |   t/s sl/sched-used-experts |   Speedup |
|:---------------------|------------------:|:-------|-------------:|----------------------------:|----------:|
| gpt-oss 120B MXFP4 MoE |               128 | pp2048 |        22.37 |                       59.09 |      2.64 |
| gpt-oss 120B MXFP4 MoE |               256 | pp2048 |        41.80 |                       99.83 |      2.39 |
| gpt-oss 120B MXFP4 MoE |               512 | pp2048 |        87.93 |                      182.89 |      2.08 |
| gpt-oss 120B MXFP4 MoE |              1024 | pp2048 |       168.17 |                      255.84 |      1.52 |
[...]

PR #20793, opened 2026-03-20
Sched: Reintroduce less synchronizations between token, with fixed pipeline parallelism.

aendk:
Follow up to https://github.com/ggml-org/llama.cpp/pull/20463#issuecomment-4091342946.

https://github.com/ggml-org/llama.cpp/pull/17795 improved performance in the single GPU setting on CUDA, but it was rolled back due to a bug surfacing in multi-GPU pipeline parallel settings.

For the single GPU setting, it moved the scheduling from `sassassasg` to the more efficient `saaasg` pattern, where `s`= sync, `a`= async copy, `g`= graph execution.
 Each asynchronous copy was enclosed in two synchronizations. Removing some superfluous synchronizations improved performance, especially on windows. The change was to only do a single synchronization between memory copies and graph execution.

However in multi-GPU settings, we saw `llama-perplexity` regressions indicating incorrect scheduling (https://github.com/ggml-org/llama.cpp/pull/20463).

I found that the event-based pipeline parallelism scheduling mechanism very likely implicitly relies on synchronous copies, as  (i) in my testing `copy_from_host` worked as intended, and (ii) disabling it and therefore introducing synchronous copies fixed the bug, `llama-perplexity` perplexity was then identical to master.

The proposed fix here is therefore to enroll pipeline parallelism into the same synchronization between async copies and graph execution as the single GPU case already has.
I think this can be a good solution as it keeps scheduling similar between single GPU and multi GPU, and because it is simpler and safer than reworking the event-driven pipeline parallelism logic.

In my testing, this proposal has same performance benefits as the initial PR, and it yields correct perplexity scores both in single and multi-GPU.
**As this bug surfaced in the community with their more diverse hardware setups and usage scenarios, it would be awesome if you could test-drive this change both with `llama-bench` and `llama-perplexity` with your usual model and launch-options usage!** 
@mxxm-t @slavap @superbobo75 @thejacer

If you can, check out this branch and compare this against its master (`git checkout HEAD~2`). Let me know if you run into performance or accuracy issues!

PR #21067, opened 2026-03-27
ggml: allow prefetching tensor overrides

am17an:
## Overview

This PR adds support to prefetch tensor overrides for each layer, overlapping with the compute for the current layer. Only the CUDA implementation is provided, guarded by the `--prefetch-weights` flag. At this stage I would consider this a PoC, so keeping as draft for now for comments and further tests. 

> Note that `--no-mmap` is necessary for this work, otherwise the operations are implicitly serialized due to the weights not being pinned. 

# Performance Analysis

## Dense Models

For dense models, the optimization is relatively straightforward. Let's call $C_{n}$ the time taken for computing the $n^{th}$ layer on the GPU, and $T_{n+1}$ the time to transfer layer $n^{th}+1$ weights from the CPU to the GPU. 

If you choose to override weights on the CPU, then this overlaps $T_{n+1}$ with $C_{n}$, whereas currently everything happens sequentially. So if $C_{n} > T_{n+1}$ we can "hide" the transfer latency and have it behave like a GPU. Although when $C_{n} >> T_{n+1}$, $C_{n}$ dominates so this is less useful. On the other hand, decreasing $T_{n+1}$ is only possible using newer hardware like PCIe Gen5 or using lower bpw.

There are two natural dimensions where we can increase $C_{n}$ without increasing $T_{n+1}$, those are `ubatch` size and the kv-cache `depth`.  Here is a graph using a relatively recent model, since this is a linear attention model, the compute doesn't go up as fast a quadratic attention model with increasing depth. We override `ffn_(gate|up|down).*` to the CPU, which are the bulk of the weights in each layer. We can see it benefits at all batch sizes, but the gap is lesser at higher batch sizes (since $C_{n}$ starts to dominate)

`llama-bench -m /opt/models/Qwen3.5-27B-Q4_K_M.gguf -fa 1 -p 2048 -ub 512,1024,2048 -d 0,10000,20000,30000,40000,50000 -n 0 -ot "ffn_(gate|up|down).*=CPU" -pw 0,1 --mmap 0`

png

 
## MoE models

For MoE models, the situation is different because of selective copying of experts (added in #15346). This is a massive improvement for smaller ubatch sizes, naturally prefetching cannot do this as it does not know which experts will be selected in the next layer. The situation is worse for larger MoE models with more experts and larger inner dims, so likely this will be slower for large MoE models unless they are deep (more layers) and not wide (larger expert dims).
 
The equation for prefetching to be beneficial becomes $C_{n} + T_{c}' > T_{c+1}$, where $T_{c}'$ is the time to transfer the selected experts for the $n^{th}$ layer. The same stuff applies, if we scale $C_{n}$ it becomes more beneficial, but we can also scale $T_{c}'$ by increasing the ubatch size, hence increasing expected number of used experts in a `ubatch`. In the graph below we see comparable performance at `ubatch`=512, but much better at 1024,2048. Since 50k context fully fits on this GPU, I also the added the theoretical maximum of fully offloading to the GPU

`llama-bench -m /opt/models/gpt_oss-20b-mxfp4.gguf -fa 1 -p 2048 -ub 512,1024,2048 -d 0,10000,20000,30000,40000,50000 -n 0 -ncmoe 999 -pw 0,1 --mmap 0`. 

gptoss

## Requirements

- AI usage disclosure:  YES, I used AI to write parts of the code, and I used it for debugging the ggml-backend code while struggling with the graph allocator. I have reviewed and tested it myself.


pandruszkow:
What sort of work needs to be done to get this merged?

am17an:
@pandruszkow someone needs to review it

mholtgraewe:
Are there any plans to merge this PR any time soon? It's been in a limbo for a few months now, but it's sorely needed...

ddh0:
I agree, it would be very nice if this was merged when ready. :)

am17an:
@ggerganov is there some version of this which could be merged? Seems like it would a nice and unique feature for hybrid inference.

JohannesGaessler (commented):
Why are per-backend changes necessary?

am17an:
It's to advertise the async stream availability, it can be done without it though

pwilkin (approved):
This would really help with mixed CPU/GPU inference, from my profiling, the data copy cost is currently the absolutely dominant cost of the entire process when running etc. on `--cpu-moe`.

pwilkin:
@am17an BTW can you make `--prefetch-weights` automatically enable `--no-mmap` with a warning message? I'm absolutely sure most people won't read the fine print at the top :)

ggml/include/ggml-backend.h
```diff
@@ -148,6 +148,8 @@ extern "C" {
         bool buffer_from_host_ptr;
         // event synchronization
         bool events;
+        // dedicated copy stream for compute/transfer overlap
+        bool copy_stream;
```
0cc4m, line 158: Why does this need a new flag and not just a check that events and tensor_set_async is available? You enabled it for CUDA only, but made no changes in the CUDA backend otherwise, so I assume the capability was already available.

am17an: Yes I can do that. It's already available in CUDA

ORippler (commented):
1. To me, this looks like a formalization of how asynchonous copy/compute should behave on ggml backends. I would appreciate a write-up/simple sketch of this, especially for the cases where we do a copy from backend a -> backend b.
    * Linking a possibly related PR that is blocked from the currently missing, shared understanding of how async copy and compute should work on a ggml backend: https://github.com/ggml-org/llama.cpp/pull/20793. CC @aendk
2. Event-based synchronization is currently bugged afaik, so we should ideally resolve this first before merging.

ggml/include/ggml-backend.h
0cc4m: So is there another requirement besides events and async set? copy_stream sounds like a separate transfer-only cuda stream, which e.g. Vulkan also provides, we call it transfer queue. But whether the transfer queue or the compute queue is used is decided by the backend, and currently I think we usually use the compute queue for async operations, because otherwise we have to synchronize the queues.

am17an: At this point there is no distinction between these things. From the ggml side it just creates another stream for the same device.

pwilkin:
Bump, any updates on this? :)

ggml/src/ggml-backend.cpp
```diff
@@ -1617,12 +1761,18 @@ static enum ggml_status ggml_backend_sched_compute_splits(ggml_backend_sched_t s
...
+            ggml_backend_event_record(sched->compute_events[split_backend_id], split_backend);
+        }
+
         // record the event of this copy
         if (split->n_inputs > 0) {
             if (sched->events[split_backend_id][sched->cur_copy] != NULL) {
                 ggml_backend_event_record(sched->events[split_backend_id][sched->cur_copy], split_backend);
             }
         }
+
```
aendk, line 1880:
```suggestion
```

ggml/src/ggml-backend.cpp
```diff
@@ -1450,17 +1550,62 @@ static enum ggml_status ggml_backend_sched_compute_splits(ggml_backend_sched_t s
...
     std::vector<ggml_bitset_t> used_ids;
 
+    bool next_weights_prefetched = false;
+
     for (int split_id = 0; split_id < sched->n_splits; split_id++) {
         struct ggml_backend_sched_split * split = &splits[split_id];
         int split_backend_id = split->backend_id;
         ggml_backend_t split_backend = sched->backends[split_backend_id];
 
+        bool weights_prefetched = next_weights_prefetched;
```
aendk, line 1665: I think, from a readability perspective, this should be moved to a separate function.
(The same goes for the long-standing MoE weight case, which puts ~80 LoC between the input copy logic and the activation copy logic between L1481 and L1564)
IMO, `ggml_backend_sched_compute_splits()` is barely maintainable as-is, so we should be careful when adding features onto it.

0cc4m:
IMO this would need changes to the backend API to explicitely ask for a separate (transfer) stream/queue, otherwise it's just gonna end up in the main queue alongside other async calls. This is exactly the kind of thing where copy/DMA hardware should be used explicitely to handle the copies completely separated from the actual compute.

am17an:
On CUDA it doesn't need that distinction, but perhaps we could instantiate a backend using a copy or compute enum, just thinking out loud.

0cc4m:
Ah, right, I had missed the separate backend instance. Then it might not be important, if the GPU scheduler handles it well-enough. Though maybe some kind of separate data streaming abstraction could still help.

noonghunna:
Some measurements from running this on a large CPU-offloaded MoE, plus a small gate that addresses the regression you predicted in the PR description.

**Setup:** Laguna-S-2.1 (117.6B, 256 experts, top-10, 48 layers) at IQ4_NL, 2× RTX 3090 (PCIe, no NVLink), 28 expert layers offloaded to CPU via `--override-tensor`, `--no-mmap`, 262K context. Within-binary A/B (`-pw 0` reproduces the unpatched engine exactly), 2 boots per arm.

## Prefill improves, first-token latency regresses hard

| metric | `-pw 0` | `-pw 1` | delta |
|---|--:|--:|--:|
| prefill 10K | 902.5 t/s | 1035.3 t/s | +14.7% |
| prefill 90K | 734.2 t/s | 835.7 t/s | +13.8% |
| **TTFT, short prompt** | **1085 ms** | **1604 ms** | **+47.8%** |
| decode | 38.40 t/s | 36.55 t/s | −4.8% |

## Mechanism: it's bytes, and it's the case you called out

Prefetch runs one split ahead of the router, so it cannot know which experts will be selected and must copy **whole** expert tensors — forfeiting the selective-expert copy from #15346 that the non-prefetch path uses. Bytes moved on the short-prompt graph: **13,668 MiB → 28,128 MiB (2.06×)**.

The control that settles it: disabling the selective copy on the `-pw 0` path gives **30,774 MiB and TTFT 1725 ms** — the regression reproduces with no prefetch machinery in the graph at all. So this is not event overhead, allocation, or stream sync.

Why it is asymmetric between short and long prompts: at 256 experts / top-10 the selective copy moves **44%** of the weight set at 34 tokens but **84%** at ubatch 2048. Prefetch therefore forfeits a 56% saving on short prompts and only 16% on long ones, while the overlap gain stays roughly constant.

## A MoE-batch gate recovers TTFT and keeps the prefill win

Dense graphs are never gated — there is no selective copy to lose there, so prefetch is unconditionally profitable.

```c
sched->prefetch_active = sched->prefetch_weights;
if (sched->prefetch_weights && min_batch > 0) {
    bool    has_moe   = false;
    int64_t moe_batch = 0;
    for (int i = 0; i < graph->n_nodes; i++) {
        const struct ggml_tensor * node = graph->nodes[i];
        if (node->op == GGML_OP_MUL_MAT_ID) {
            has_moe = true;
            if (node->ne[2] > moe_batch) {
                moe_batch = node->ne[2];
            }
        }
    }
    if (has_moe && moe_batch < min_batch) {
        sched->prefetch_active = false;
    }
}
```

Then consult `prefetch_active` at the runtime sites instead of `prefetch_weights`. One detail worth keeping: leave the `nodes_per_input` allocation bound on `prefetch_weights`, so buffer sizing does not vary per graph.

Measured crossover is between ~250 and ~1040 MoE batch; 512 works as a default.

| | `-pw 0` | `-pw 1` ungated | `-pw 1` gated |
|---|--:|--:|--:|
| TTFT | 1084 ms | 1613 ms | **1096 ms** |
| prefill ~5.2K | 856 t/s | 1026–1030 | **1026–1030** |
| decode | 38.34 t/s | 36.60 | 37.00 |

## One caveat the gate does *not* fix

The VRAM cost. The scheduler reserves for the worst-case graph at `n_ubatch`, where prefetch is active regardless of the gate, so the double-buffer claims that memory for the life of the process. On this setup it permanently costs ~15% of the VRAM that would otherwise hold resident expert weights — which is why decode only partially recovers (37.00 vs 38.34) rather than returning to baseline. Anyone tight on VRAM pays that whether the gate is in or not.

That is a property of worst-case graph reservation rather than anything in this PR, but it is worth knowing: on a memory-constrained MoE setup the steady-state decode cost can outweigh the prefill gain, so `--prefetch-weights` is probably best left opt-in rather than defaulted on for MoE.

Happy to run other configurations if that would help. The gate is yours to take or leave — it is a small change to your design, not a competing approach.

yalun753:
Good to see this direction getting formalized. I can add measured data on the MoE side, since that's where you (correctly, I think) expect prefetching to struggle.

We run a 157B MoE (DeepSeek-V4-Flash, 256 experts/layer) the same way — experts mmap'd in host RAM, computed on GPU. Key observation for the prefetch equation: expert selection barely overlaps across adjacent layers. A 30-turn session touched ~29 GB of distinct experts and cross-layer expert-set overlap (Jaccard) is low — so "guess the next layer's experts" has very little prediction headroom on large scattered-routing models.

We also tested prefetching at the OS level (posix_madvise(WILLNEED) after DONTNEED, +5 ms wait): cudaHostRegister still took ~3 ms. Prefetch did not hide the cold-expert fault.

Where this landed for us: instead of prefetching, we pin the mmap'd expert tensors (cudaHostRegister, whole-tensor merged registration) so per-token transfers go DMA directly, and keep hot experts in a GPU-side LRU cache. Decode on Qwen3.6-35B-A3B: 46.8 -> 50.2 t/s at 2.9 GB VRAM (4090). On the 157B model the cold-expert first touch (~2.1 ms) can't be hidden and routing is scattered, so ~4-5 t/s there is the physical decode ceiling.

One data point that supports your C_n + T_c' > T_{c+1} analysis: at batch size 1 there's nothing to overlap — a single expert is ~1 MB (Gen4 copy ~60 us vs MMVQ kernel <10 us), and a double-buffered H2D pipeline measured no gain (5.0 vs 5.5 t/s). The overlap headroom only exists at large ubatch (prefill), which matches your dense-model graphs.

Shipped open-source if useful as reference: https://github.com/yalun753/moe-l2

crusaderky:
@am17an   https://github.com/am17an/llama.cpp/pull/25 merges a recent master branch into this PR
Working on benchmarks of my own...

EJainDev:
Hi, is there a way to control how many layers are prefetched at a time as some GPUs may be significantly faster with PCIe being a bigger bottleneck?

yalun753:
@EJainDev Great question — and for MoE, I'd argue the control knob you want isn't "how many layers ahead" but "which experts". The reason prefetch regresses on MoE (2.06× bytes moved, +47.8% TTFT in @noonghunna's measurements) is that it can't know the next layer's expert selection, so it copies whole expert tensors. In our setup we solve exactly this with a routing table (top-N domains × top-K experts per layer, learned from activation history): the table tells us ahead of time which experts each layer will likely activate, so we prefetch/pin only those experts instead of the whole set — per-token transfers then go DMA-direct and hot experts stay in a GPU-side LRU cache. Decode on Qwen3.6-35B-A3B: 46.8 → 50.2 t/s at 2.9 GB VRAM (4090). So: a "prefetch depth" parameter is useful for dense layers, but for MoE the more powerful control is routing-table-guided expert prefetch. Reference: https://github.com/yalun753/moe-l2

EJainDev:
I definitely agree, for MoE a cache with an algorithm like EMA to maintain the topmost accessed experts would be amazing. However, for prompt processing with larger batch sizes, that could prove to provide inferior performance because of how many experts can be accessed in a single batch.

As for the dense model, I'm not noticing a measurable speed up. With upstream llama.cpp and this PR on Qwen3.8 27B I am achieving 4tps with remarkably low GPU utilization. I also do have PCIe 5.0x16 so it seems incorrect.

yalun753:
@EJainDev That's exactly what our routing table is — an EMA over per-layer activation counts, plus a domain dimension on top (top-N domains × top-K experts per layer). On the prompt-processing side, your intuition matches our data: with the cache on, DS-V2-Lite prefill went 99 → 308 t/s (+211%), and the gain concentrates exactly where large batches touch many distinct experts. Happy to share more details if useful.

EJainDev:
@yalun753 I think this could be a useful feature as a separate PR, and with quantized MoE experts, the cached # of experts could drastically increase. I've recently begun to try and experiment with something like this but writing the kernels and researching the algorithm takes some time.

yalun753:
@EJainDev Thanks! Agreed on the quantized point — expert size directly caps how many fit (our Q4 experts are ~24 MB/slot vs ~12 MB for IQ2, so the same VRAM holds 2× the experts). We're evaluating an upstream PR; our diff is the routing-table dimension (domain-aware top-K, learned from activation history) on top of a plain LRU. Happy to share our kernel/benchmark work — good luck with the experiments, happy to compare notes.

crusaderky:
Data point for `unsloth/Qwen3.8-Flash-Next-GGUF:UD-IQ4_XS` running on RTX3090, PCIe 3.0 x16 (not a typo) bus, and 64GB DDR4 RAM: I'm getting worse performance at ub=2048.

Without prefetch: 150 tok/s
With prefetch: 95 tok/s

I guess the key here is that the OP's benchmarks are on PCIe 5.0, which is 4x faster than mine.

Volunteer-1:
@am17an Please verify whether this PR has the same problem as was found in the bot's review of a similar PR: https://github.com/ggml-org/llama.cpp/pull/28414#issuecomment-5546348326 : The prefetch silently corrupts multi GPU inference with -sm layer. I'd also like to know whether it corrupts multi GPU inference with -sm tensor.

rintic-13:
Have you tried forming alternating layers to achieve  overlap the H2D?

        // Pattern: 4 GPU layers : 1 CPU layer
        // For the last 25 layers:
        //   GPU GPU GPU GPU CPU
        // Only tensors >= 10 MiB are moved to CPU/host memory.

        if (!buft &&
            info.layer == LLM_TENSOR_LAYER_REPEATING &&
            tn.bid >= 0) {

            const int n_layers = hparams.n_layer();

            const int gpu_layers_per_cpu = 3;
            const int n_cpu_layers_total = 6;

            const int block_size = gpu_layers_per_cpu + 1; // 5
            const int total_layers = n_cpu_layers_total * block_size; // 25

            const int interleave_start =
                std::max(0, n_layers - total_layers);

            const int relative_layer =
                tn.bid - interleave_start;

            const bool is_cpu_layer =
                relative_layer >= 0 &&
                (relative_layer % block_size == gpu_layers_per_cpu);

            if (is_cpu_layer &&
                ggml_nbytes(t_meta) >= 30 * 1024 * 1024) {

                buft = select_weight_buft(
                    hparams,
                    t_meta,
                    op,
                    buft_list_cpu);

                if (buft) {
                    LLAMA_LOG_INFO(
                        "[AUTO CPU RATIO] layer=%d tensor=%s size=%zu MiB\n",
                        (int) tn.bid,
                        tn.str().c_str(),
                        ggml_nbytes(t_meta) / 1024 / 1024);
                }
            }
        }
        // === END CUSTOM CPU/GPU LAYER PLACEMENT ===

PR #21073, opened 2026-03-27
server: (webui) no more gzip compression

ngxson:
## Overview

Gzip compression was added to reduce the size of built binary. However, it blows up the disk space usage for git because git cannot efficiently run diff against 2 versions of the gzip-ed file.

To verify my theory, I vibe-coded this script which logs all objects for the past 500 commits as well as its packed size on disk: https://gist.github.com/ngxson/d5cf485020767f1ed0a4678b888df1a7 ; it turns out that `index.html.gz` is almost always stored as-is, without any diff.

This PR removes the gzip, the side effect is that it will now add ~3MB size to the `llama-server` build, but I think this is a worth trade-off because (1) the web UI already became matured / very functional and (2) a new flag is added to exclude webui from the build, https://github.com/ggml-org/llama.cpp/pull/20158

## Additional information

To test this, I added a commit https://github.com/ggml-org/llama.cpp/commit/0d7a3531593d87b724d404c8727a96becab3ab07 that changes a single line. Verify the packed size of that particular commit:

```sh
COMMIT=0d7a3531593d87b724d404c8727a96becab3ab07
FILE=tools/server/public/index.html
BLOB=$(git ls-tree $COMMIT -- $FILE | awk '{print $3}')
PARENT_BLOB=$(git ls-tree ${COMMIT}^ -- $FILE | awk '{print $3}')

printf "$PARENT_BLOB\n$BLOB\n" | git pack-objects --delta-base-offset /tmp/pack-show
git verify-pack -v /tmp/pack-show-*.idx
rm -f /tmp/pack-show-*.idx /tmp/pack-show-*.pack
```

Output:

```
e7e836b706b36b08705e78f2a6a29f65d41d178e blob   4913418 1592851 12
5767e2045b6c3ce3de2e6c8617f960f6c9c56f6e blob   555 337 1592863 1 e7e836b706b36b08705e78f2a6a29f65d41d178e
non delta: 1 object
chain length = 1: 1 object
/tmp/pack-show-cf55442797ff18841bc46cdd1d6c29d68b5d8aba.pack: ok
```

Note that column format is `hash  type  raw_size  packed_size`

The object `e7e836b706b36b08705e78f2a6a29f65d41d178e` is the one that contains the newly-created `index.html`, its size if now about 1.5MB.

We can surely improve this further by breaking `index.html` into small files. But that is to be done in another PR. 

# Requirements

- AI usage disclosure: YES, to help with writing debugging scripts


ServeurpersoCom:
Exactly! Git uses zlib internally to compress objects, but it relies on delta encoding between successive versions of a file. The problem is that two files compressed with gzip that differ by only one line produce completely different blobs at the binary level. Therefore, Git stores each complete version.
The only minor practical issue is that git diffs touching HTML will be cluttered with minified code instead of a simple line indicating that the binary has changed.
A possible improvement would be to use terser's max_line_len for JS and html-minifier's maxLineLength for HTML, so that minified output still has line breaks at regular intervals, giving git proper delta boundaries.

ggerganov (approved):

ServeurpersoCom (approved):

ServeurpersoCom:
In addition, we're increasing the security level a bit: the index.html.gz file could hide a potentially unwanted npm update due to a Node version mismatch on one of the contributors' computers or any other unintentional manipulation. (node_modules ignored but code packed inside the archive). Later, if desired, the Trivy security scanner is open source and can be used to scan dependencies on a CI machine or pod.

ngxson:
The web UI is re-built on CI for checking anyway, so unlikely that someone can tamper with it.

Anw, I pushed my last commit to split the file into html/js/css instead of one big file. This is mostly to optimize the speed on low-bandwidth network.

ngxson:
Hmm, seems like server CI fails for unrelated reason (HF hub is down?)

tools/server/webui/scripts/install-git-hooks.sh
allozaur, line 1: @ngxson this PR is a good opportunity to move the build of `index.html` to CI in order to avoid the merge conflicts with static build output for most of the webui PRs that don't have recent `master` as base (or don't have it merged in).

tools/server/webui/svelte.config.js
```diff
@@ -21,7 +21,7 @@ const config = {
 			strict: true
 		}),
 		output: {
-			bundleStrategy: 'inline'
+			bundleStrategy: 'single'
```
allozaur, line 24: @ngxson are u sure we want to move away from a single HTML file with all JS and CSS inlined to have the HTML + CSS bundle + JS bundle separated?

ServeurpersoCom:
Let's try it :)
```
(root|~/llama.cpp.pascal/tools/server/webui) ls -l ../public/
total 4808
-rw-r--r-- 1 root root  499641 28 mars  13:53 bundle.css
-rw-r--r-- 1 root root 4409035 28 mars  13:53 bundle.js
-rw-r--r-- 1 root root    6928 28 mars  13:53 index.html
-rw-r--r-- 1 root root     268 28 mars  13:53 loading.html
(root|~/llama.cpp.pascal/tools/server/webui)
```

And git status say :
```
        modifié :         tools/server/public/bundle.js
        modifié :         tools/server/public/index.html
        modifié :         tools/server/webui/package-lock.json <- removed fflate
```

tools/server/webui/vite.config.ts
```diff
@@ -61,26 +57,35 @@ function llamaCppBuildPlugin() {
...
+						}
+					}
+					// Copy bundle.*.css -> ../public/bundle.css
+					if (existsSync(bundleDir)) {
+						const cssFiles = readdirSync(bundleDir).filter((f) => f.match(/^bundle\..+\.css$/));
+						if (cssFiles.length > 0) {
+							copyFileSync(resolve(bundleDir, cssFiles[0]), resolve('../public/bundle.css'));
+							console.log(`✓ Copied ${cssFiles[0]} -> bundle.css`);
+						}
 					}
```
allozaur, line 86: It just feels that we unnecessarily add complexity when w have the CSS/JS bundles separated. But if there are some solid arguments to have it separate, then ofc im open to introducing it.

.gitattributes
allozaur, line 1: Oh, will this actualy solve the problem with the merge conflicts that `index.html.gz` causes atm?

allozaur (changes requested):

tools/server/webui/vite.config.ts
allozaur: > Anw, I pushed my last commit to split the file into html/js/css instead of one big file. This is mostly to optimize the speed on low-bandwidth network.

Okay i see.

tools/server/webui/svelte.config.js
ServeurpersoCom: I think the simplest solution is an all-in-one html, but with lines that aren't too long so that git can deduplicate them.

.gitattributes
ServeurpersoCom: I think not for now. A git binary file cannot do deltas, a one-line file will do the same, current files are minified, on one line.

ServeurpersoCom: I try on my side :

esbuild: {
  lineLimit: 500
},

ServeurpersoCom:
I tryed :
```
(root|~/llama.cpp.pascal) git diff tools/server/webui/vite.config.ts
diff --git a/tools/server/webui/vite.config.ts b/tools/server/webui/vite.config.ts
index ca478668e..26e94d581 100644
--- a/tools/server/webui/vite.config.ts
+++ b/tools/server/webui/vite.config.ts
@@ -98,6 +98,9 @@ export default defineConfig({
                        'katex-fonts': resolve('node_modules/katex/dist/fonts')
                }
        },
+       esbuild: {
+               lineLimit: 500
+       },
        build: {
                assetsInlineLimit: MAX_ASSET_SIZE,
                chunkSizeWarningLimit: 3072,
(root|~/llama.cpp.pascal)
```

lines
Now Git can "compress" using delta!
I'm curious to see how much the minified code changes when the source code is modified. Interesting!

ServeurpersoCom:
The line length trick isn't exactly revolutionary either. Okay, it'll compress things a bit in the repo, but creating a bunch of bloat diffs for every little change, this is what it looks like:

diff

Furthermore, for index.html, git gives up and keeps it as a binary file... but small !
```
(root|~/llama.cpp.pascal) ls -lha tools/server/public/index.html tools/server/public/bundle.js
-rw-r--r-- 1 root root 4,3M 28 mars  14:19 tools/server/public/bundle.js
-rw-r--r-- 1 root root 6,8K 28 mars  14:19 tools/server/public/index.html
(root|~/llama.cpp.pascal)
```
OK, that could probably slow down the rate at which the Git repository grows… but I only modified one file and the difference is huge… Maybe there's a way to control the entropy of the minimizer lol

ServeurpersoCom:
With this setup, the reduction in the growth of the git repo is the best + git merging capability !!!

```
esbuild: {
  lineLimit: 500,
  minifyIdentifiers: false,
  minifySyntax: true,
  minifyWhitespace: true,
},
```

Small diff !
diff-v2!

If that's the intended goal, it's fine, but at the cost of slightly increasing the size of the code sent to browsers in its current state.
I'm sure we could implement on-the-fly compression in the binary, but that's overkill?

-> 1MB more of variable name :
```
(root|~/llama.cpp.pascal) ls -lha tools/server/public/bundle.js tools/server/public/index.html
-rw-r--r-- 1 root root 5,4M 28 mars  14:36 tools/server/public/bundle.js
-rw-r--r-- 1 root root 6,8K 28 mars  14:36 tools/server/public/index.html
(root|~/llama.cpp.pascal)
```

WebUI Work on my production pod OK :
```
(root|~/llama.cpp.pascal) ls -lha /var/www/ia/webui/
total 5,9M
drwxr-xr-x  2 root root 4,0K 28 mars  14:45 .
drwxr-xr-x 16 root root 4,0K 23 mars  22:36 ..
-rw-r--r--  1 root root 488K 28 mars  14:45 bundle.css
-rw-r--r--  1 root root 5,4M 28 mars  14:45 bundle.js
-rw-r--r--  1 root root 6,8K 28 mars  14:45 index.html
-rw-r--r--  1 root root  268 28 mars  14:45 loading.html
```

.gitattributes
ServeurpersoCom: The good one, disable identifier/var name mangling, with this, git will be able to merge automatically in many cases, and the repo will grow less over time.

esbuild: {
  lineLimit: 500,
  minifyIdentifiers: false,
  minifySyntax: true,
  minifyWhitespace: true,
},

ServeurpersoCom (approved):

ngxson:
> I'm sure we could implement on-the-fly compression in the binary, but that's overkill?

@ServeurpersoCom we can also bring back the gzip compressing, but it can be run on cmake, at the same step as converting assets --> hpp file

.gitattributes
ngxson: > Oh, will this actualy solve the problem with the merge conflicts that `index.html.gz` causes atm?

@allozaur  It can reduce the conflict, but technically in most case the auto-merge will produce a non-functioning `index.html` file because the merged changes is far too complex. It's recommended to rebuild for every new commit / new PR

> lineLimit: 500,

@ServeurpersoCom just wondering, if a line is long and something at the beginning of the line is changed, will it make the whole thing shift? example:

```
var a = "123456"; console.log(...);
if (a == 78901234) { do_something() }
```

and if I change to `a = ""`, now the whole thing shift:

```
var a = ""; console.log(...); if (
a == 78901234) { do_something() }
```

ServeurpersoCom: That's interesting. A successful, silent 3-way merge that actually break code, I've never seen that before... But when it's minimized...
Yes, reflow will come with lineLimit. But it's still infinitely better than gzip, where the slightest change to the source code results in a completely different blob.

ServeurpersoCom: We would need to be able to run simulations, a script that replays the entire git history of the webui locally to experiment!

allozaur (approved):

ServeurpersoCom (approved):

ServeurpersoCom (approved):

PR #21074, opened 2026-03-27
ggml-cuda: Add generic NVFP4 MMQ kernel

michaelw9999:
Here is the next stage of CUDA NVFP4's roll-out, with a generic NVFP4 MMQ leveraging the existing `vec_dot_q8_0_16_q8_1_mma` and `vec_dot_q8_0_16_q8_1_dp4a` kernels and re-using them for NVFP4 with same tile size as Q3_K.   This brings a significant increase in prefill speed vs the baseline MMVQ/dp4a kernel, and was kept small intentionally.  A Blackwell specific MMA kernel will be introduced later.

`ncu` was used to help optimize and tune for performance however I struggled to find a good balance using the existing generic kernels, and despite high register pressure and only 16.7% occupancy / 1sm/block, no other combination tried gave as significant prefill increases without also reducing tg.  This could be investigated and tuned further on other GPUs, with the focus for Blackwell optimizations kept to the future Blackwell NVFP4xNVFP4 kernels.

`FP8_AVAILABLE` checks were added to hopefully avoid any ci issues.

**Llama-bench results (SM120 RTX 5090)**:

| NVFP4 Model | Test | MMVQ | MMQ | Increase | Speedup |
|:---|:---|:---|:---|---:|---|
| Qwen3-0.6B | pp512 | 33870.66 | 42941.79 | +26.8% | 1.27x |
| Qwen3-0.6B | tg128 | 570.82 | 565.80 | -0.9% | 0.99x |
| Qwen3.5-0.8B | pp512 | 23799.93 | 32858.97 | +38.1% | 1.38x |
| Qwen3.5-0.8B | tg128 | 392.92 | 389.75 | -0.8% | 0.99x |
| Qwen3.5-4B | pp512 | 5984.10 | 13064.37 | +118.3% | 2.18x |
| Qwen3.5-4B | tg128 | 205.00 | 203.58 | -0.7% | 0.99x |
| Qwen3.5-27B | pp512 | 1026.28 | 2988.58 | +191.2% | 2.91x |
| Qwen3.5-27B | tg128 | 62.94 | 62.77 | -0.3% | 1.00x |
| Nemotron-3-Nano-30B-A3B | pp512 | 1828.75 | 7116.02 | +289.1% | 3.89x |
| Nemotron-3-Nano-30B-A3B | tg128 | 230.32 | 228.94 | -0.6% | 0.99x |
| [Nemotron-Cascade-2-30B-A3B](https://huggingface.co/michaelw9999/Nemotron-Cascade-2-30B-A3B-NVFP4-GGUF)  | pp512 | 1995.84 | 6591.44 | +230.3% | 3.30x |
| [Nemotron-Cascade-2-30B-A3B](https://huggingface.co/michaelw9999/Nemotron-Cascade-2-30B-A3B-NVFP4-GGUF)  | tg128 | 210.94 | 209.88 | -0.5% | 0.99x ||

AI assistance was used to assist in creating this code.


am17an:
Why is FP8 required? We should think of a fallback for cards which don't have FP8. I don't think even Amperes have FP8

michaelw9999:
> Why is FP8 required? We should think of a fallback for cards which don't have FP8. I don't think even Amperes have FP8

It was added in as a guard when the first PR was failing for ROCm repeatedly.  I can't say I know anything at all about ROCm. @IMbackK ?


```
static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
#ifdef FP8_AVAILABLE
    const uint32_t bits = x * (x != 0x7F && x != 0xFF); // Convert NaN to 0.0f to match CPU implementation.
#if defined(GGML_USE_HIP) && defined(CDNA3)
    // ROCm dose not support fp8 in software on devices with fp8 hardware,
    // but CDNA3 supports only e4m3_fnuz (no inf).
    const __hip_fp8_e4m3_fnuz xf = *reinterpret_cast<const __hip_fp8_e4m3_fnuz *>(&bits);
#else
    const __nv_fp8_e4m3 xf = *reinterpret_cast<const __nv_fp8_e4m3 *>(&bits);
#endif // defined(GGML_USE_HIP) && defined(GGML_USE_HIP)
    return static_cast<float>(xf) / 2;
#else
    NO_DEVICE_CODE;
#endif // FP8_AVAILABLE
```
But, we do have a fallback ready now we could use instead of `NO_DEVICE_CODE`, the generic `ggml_ue4m3_to_fp32` should work fine and is already what was working with the CPU-only NVFP4.   Could just bring in the same code as that and put it in place of `NO_DEVICE_CODE`.

IMbackK:
FP8_AVAILABLE on hip is about platform availability of fp8, as that was added in rocm 6.2, not hardware availability. On devices without native fp8, bitwise operations are emitted to convert for those conversion functions.

I dont think not supporting nvfp4 on rocm 6.1 (the only version <6.2 we support) is a big deal.

am17an:
The name FP8_AVAILABLE is quite confusing as it gives the impression it's about fp8 tensor cores. We should just use the fallback in the conversion function and not pollute the code with it everywhere.

michaelw9999:
> The name FP8_AVAILABLE is quite confusing as it gives the impression it's about fp8 tensor cores. We should just use the fallback in the conversion function and not pollute the code with it everywhere.

I agree with this, it is also messy with `BLACKWELL_MMA_AVAILABLE`  and `use_native_mxfp4` since after this PR, the SM120 MMA kernel is already ready with matching `use_native_nvfp4` adding more separation, since it can support either Q8 or NVFP4 on src1 and in theory that doesn't have to be restricted to Blackwell. Post Blackwell with future firmware support for microscales, quant recipes, options, etc, the #ifdefs everywhere for what specific feature goes with what hardware (on which firmware) makes it hard to keep neat.  I could create some kind of compatibility matrix API or an expanded traits/template generically to try to clean all of this.

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -826,6 +830,65 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+        const int kqs = 16 * kbx;
+        const int ksc = 4 * kbx;
+
+#pragma unroll
+        for (int sub = 0; sub < QK_NVFP4 / QK_NVFP4_SUB; ++sub) {
+            const int2 q0 = get_int_from_table_16(src_qs[2 * sub + 0], kvalues_mxfp4);
+            const int2 q1 = get_int_from_table_16(src_qs[2 * sub + 1], kvalues_mxfp4);
+
+#if defined(AMD_MFMA_AVAILABLE) || defined(TURING_MMA_AVAILABLE) || defined(AMD_WMMA_AVAILABLE)
+            x_qs[i * MMQ_MMA_TILE_X_K_Q3_K + kqs + 4 * sub + 0] = q0.x;
```
am17an: Please define `MMQ_MMA_TILE_X_K_NVFP4` rather than using this

am17an (commented):
Can you compare the speeds vs a mxfp4 model (on a non-blackwell)?

michaelw9999:
> Can you compare the speeds vs a mxfp4 model (on a non-blackwell)?

Every computer I've got at the moment is either Blackwell or slow iGPU laptops so can't easily do this, but I can try anyway with this generic MMQ

xkmire:
I have been coding with CLINE all morning using NVFP4 of Qwen 3.5 122B using this patch on blackwell.
The performance improvement on large contexts is impressive. I will most likely move to this instead of Q5 quant.
Nice work!

michaelw9999:
> I have been coding with CLINE all morning using NVFP4 of Qwen 3.5 122B using this patch on blackwell. The performance improvement on large contexts is impressive. I will most likely move to this instead of Q5 quant. Nice work!

Thank you :))  Glad you like it!!

ggml/src/ggml-cuda/mmq.cuh
michaelw9999: For consistency, the kernels (IQ2_S, IQ2_XS, vec_dot_q8_0_16_q8_1_mma()) retain it as `MMQ_MMA_TILE_X_K_Q3_K` .

I was using `MMQ_MMA_TILE_X_K_NVFP4` for the future NVFP4 x NVFP4 tile.  
Should I still change this perhaps to `MMQ_MMA_TILE_X_K_NVFP4_Q8` and make the other one `MMQ_MMA_TILE_X_K_NVFP4_NVFP4`?

am17an: See how MXFP4 tiles are defined, follow that convention

ggml/src/ggml-cuda/common.cuh
```diff
@@ -800,19 +800,32 @@ static __device__ __forceinline__ float ggml_cuda_e8m0_to_fp32(uint8_t x) {
...
     return static_cast<float>(xf) / 2;
-#else
-    NO_DEVICE_CODE;
-#endif // FP8_AVAILABLE
+#else 
+    if (x == 0 || (x == 0x7F && x != 0xFF)) { // Convert NaN to 0.0f
+        return 0.0f;
+    }
+    int   exp = (x >> 3) & 0xF;
+    int   man = x & 0x7;
```
am17an:
```suggestion
    const int exp = (x >> 3) & 0xF;
    const int man = x & 0x7;
```

ggml/src/ggml-cuda/mmq.cu
```diff
@@ -23,6 +23,11 @@ static void ggml_cuda_mul_mat_q_switch_type(ggml_backend_cuda_context & ctx, con
         case GGML_TYPE_MXFP4:
             mul_mat_q_case<GGML_TYPE_MXFP4>(ctx, args, stream);
             break;
+#ifdef FP8_AVAILABLE
```
am17an: they should be removed now?

ggml/src/ggml-cuda/mmq.cu
```diff
@@ -364,3 +368,5 @@ bool ggml_cuda_should_use_mmq(enum ggml_type type, int cc, int64_t ne11, int64_t
     return (!GGML_CUDA_CC_IS_CDNA(cc)) || ne11 < MMQ_DP4A_MAX_BATCH_SIZE;
 
 }
+
+DECL_MMQ_CASE(GGML_TYPE_NVFP4);
```
am17an: I don't think this should be here, belongs in mmq.cuh

ggml/src/ggml-cuda/mmq.cu
michaelw9999: I think with the fallback added and extra guard for HIP_FP8_AVAILABLE that I put into the conversion function, it should stop complaining.  I hope.

ggml/src/ggml-cuda/common.cuh
```diff
@@ -800,19 +800,32 @@ static __device__ __forceinline__ float ggml_cuda_e8m0_to_fp32(uint8_t x) {
...
     return static_cast<float>(xf) / 2;
-#else
-    NO_DEVICE_CODE;
-#endif // FP8_AVAILABLE
+#else 
+    if (x == 0 || (x == 0x7F && x != 0xFF)) { // Convert NaN to 0.0f
+        return 0.0f;
+    }
+    const int   exp = (x >> 3) & 0xF;
+    const int   man = x & 0x7;
```
am17an: There's an extra space here, please remove it

ggml/src/ggml-cuda/mmq.cu
```diff
@@ -363,4 +367,4 @@ bool ggml_cuda_should_use_mmq(enum ggml_type type, int cc, int64_t ne11, int64_t
 
     return (!GGML_CUDA_CC_IS_CDNA(cc)) || ne11 < MMQ_DP4A_MAX_BATCH_SIZE;
 
-}
+}
```
am17an: Add a newline at the end

ggml/src/ggml-cuda/common.cuh
am17an:
```suggestion
    const int  exp = (x >> 3) & 0xF;
    const int  man = x & 0x7;
```

This is what I meant

ggml/src/ggml-cuda/vendors/hip.h
```diff
@@ -238,7 +238,7 @@ typedef __hip_bfloat162 nv_bfloat162;
 #if HIP_VERSION >= 60200000
 #include <hip/hip_fp8.h>
 typedef __hip_fp8_e4m3 __nv_fp8_e4m3;
-#define FP8_AVAILABLE
+#define HIP_FP8_AVAILABLE
```
IMbackK: This is not correct, cuda also support fp8 in the runtime when above some version and this change breaks older versions see https://github.com/ggml-org/llama.cpp/blob/e6f2ec01ff808b60f80ffcb46c748f871576c8f3/ggml/src/ggml-cuda/vendors/cuda.h#L11

ggml/src/ggml-cuda/mmq.cu
CISC: Still missing newline.

michaelw9999: Got confused and thought "No newline at end of file" meant not to put it in. Doing that in a moment!

michaelw9999:
@am17an I think this last push should actually the root cause of the ci fail. NVFP4 was not in the list of quants in `generate_cu_files.py` script at all, so there was no `mmq-instance-nvfp4.cu` being generated.  Made the divisions between HIP and CUDA and fallback stricter, too just in case.

am17an:
This is failing compilation

michaelw9999:
> This is failing compilation

Argh! Frustrating.
```
[ 26%] Building HIP object ggml/src/ggml-hip/CMakeFiles/ggml-hip.dir/__/ggml-cuda/template-instances/mmq-instance-iq4_xs.cu.o
[ 27%] Building HIP object ggml/src/ggml-hip/CMakeFiles/ggml-hip.dir/__/ggml-cuda/template-instances/mmq-instance-mxfp4.cu.o
[ 27%] Building HIP object ggml/src/ggml-hip/CMakeFiles/ggml-hip.dir/__/ggml-cuda/template-instances/mmq-instance-q2_k.cu.o
```
It still was missing mmq-instance-nvfp4.cu and skipping past it.  I guess it doesn't run the py script.  I just included that file now and added in.  I hope this works now!

michaelw9999:
For HIP Quality check failure, VGPR=262 over limit of 256.  
Can add the MMQ for NVFP4 to the ignore list in `scripts/hip/gcn-cdna-vgpr-check.py`:

```
'_ZL9mul_mat_qIL9ggml_type40ELi112ELb0EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
'_ZL9mul_mat_qIL9ggml_type40ELi112ELb1EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
'_ZL9mul_mat_qIL9ggml_type40ELi128ELb0EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
'_ZL9mul_mat_qIL9ggml_type40ELi128ELb1EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
```

am17an (approved):

IMbackK:
> For HIP Quality check failure, VGPR=262 over limit of 256. Can add the MMQ for NVFP4 to the ignore list in `scripts/hip/gcn-cdna-vgpr-check.py`:
> 
> ```
> '_ZL9mul_mat_qIL9ggml_type40ELi112ELb0EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
> '_ZL9mul_mat_qIL9ggml_type40ELi112ELb1EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
> '_ZL9mul_mat_qIL9ggml_type40ELi128ELb0EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
> '_ZL9mul_mat_qIL9ggml_type40ELi128ELb1EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
> ```

Sure, please do. I mean ideally we would reduce register pressure, the kernel is not going to perform great as is.

michaelw9999:
> > For HIP Quality check failure, VGPR=262 over limit of 256. Can add the MMQ for NVFP4 to the ignore list in `scripts/hip/gcn-cdna-vgpr-check.py`:
> > ```
> > '_ZL9mul_mat_qIL9ggml_type40ELi112ELb0EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
> > '_ZL9mul_mat_qIL9ggml_type40ELi112ELb1EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
> > '_ZL9mul_mat_qIL9ggml_type40ELi128ELb0EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
> > '_ZL9mul_mat_qIL9ggml_type40ELi128ELb1EEvPKcPKiS4_S4_PfS5_iiiiiiiiiiiiiiiii',
> > ```
> 
> Sure, please do. I mean ideally we would reduce register pressure, the kernel is not going to perform great as is.

Added that in.  I'd be curious to see how it actually benches on one but now it should work on all versions.

am17an:
@JohannesGaessler this is good to merge from my side

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -230,6 +237,8 @@ static constexpr __host__ __device__ int mmq_get_mma_tile_x_k(ggml_type type) {
         case GGML_TYPE_Q8_0:    return MMQ_MMA_TILE_X_K_Q8_0;
         // tile sizes are the same for Q8_1 and FP4 for blackwell
         case GGML_TYPE_MXFP4:   return MMQ_MMA_TILE_X_K_Q8_1;
+        // tile sizes are the same for Q3_K and NVFP4 for generic MMQ
+        case GGML_TYPE_NVFP4:   return MMQ_MMA_TILE_X_K_Q3_K;
```
JohannesGaessler:
```suggestion
        case GGML_TYPE_NVFP4:   return MMQ_MMA_TILE_X_K_NVFP4;
```

I don't see a reason to return the q3_k value when there is one for NVFP4.

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -220,6 +224,9 @@ static_assert(MMQ_MMA_TILE_X_K_Q3_K % 8 == 4, "Wrong padding.");
 static_assert(MMQ_MMA_TILE_X_K_Q6_K % 8 == 4, "Wrong padding.");
 static_assert(MMQ_MMA_TILE_X_K_FP4  % 8 == 4, "Wrong padding.");
 static_assert(MMQ_MMA_TILE_X_K_FP4 == MMQ_MMA_TILE_X_K_Q8_1, "Wrong tile size for MXFP4");
+static_assert(MMQ_MMA_TILE_X_K_NVFP4 % 8 == 4, "Wrong padding.");
+static_assert(MMQ_MMA_TILE_X_K_NVFP4 == MMQ_MMA_TILE_X_K_Q3_K, "Wrong tile size for NVFP4");
```
JohannesGaessler:
```suggestion
```

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -826,6 +835,65 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+    constexpr int warp_size = ggml_cuda_get_physical_warp_size();
+
+#if defined(AMD_MFMA_AVAILABLE) || defined(TURING_MMA_AVAILABLE) || defined(AMD_WMMA_AVAILABLE)
+    int   * x_qs = (int   *) x_tile;
+    float * x_df = (float *) (x_qs + MMQ_TILE_NE_K*2);
+#else
+    constexpr tile_x_sizes txs = mmq_get_dp4a_tile_x_sizes(GGML_TYPE_NVFP4, mmq_y);
+    int   * x_qs = (int   *) x_tile;
+    float * x_df = (float *) (x_qs + txs.qs);
+#endif
```
JohannesGaessler:
```suggestion
#endif // defined(AMD_MFMA_AVAILABLE) || defined(TURING_MMA_AVAILABLE) || defined(AMD_WMMA_AVAILABLE)
```

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -826,6 +835,65 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+            x_qs[i * MMQ_MMA_TILE_X_K_NVFP4 + kqs + 4 * sub + 2] = q0.y;
+            x_qs[i * MMQ_MMA_TILE_X_K_NVFP4 + kqs + 4 * sub + 3] = q1.y;
+            x_df[i * MMQ_MMA_TILE_X_K_NVFP4 + ksc + sub] = ggml_cuda_ue4m3_to_fp32(bxi->d[sub]);
+#else
+            x_qs[i * (2 * MMQ_TILE_NE_K + 1) + kqs + 4 * sub + 0] = q0.x;
+            x_qs[i * (2 * MMQ_TILE_NE_K + 1) + kqs + 4 * sub + 1] = q1.x;
+            x_qs[i * (2 * MMQ_TILE_NE_K + 1) + kqs + 4 * sub + 2] = q0.y;
+            x_qs[i * (2 * MMQ_TILE_NE_K + 1) + kqs + 4 * sub + 3] = q1.y;
+            x_df[i * (2 * MMQ_TILE_NE_K * 2 / QI_NVFP4) + i / (QK_NVFP4_SUB / QI_NVFP4) + ksc + sub] = ggml_cuda_ue4m3_to_fp32(bxi->d[sub]);
+#endif
```
JohannesGaessler:
```suggestion
#endif // defined(AMD_MFMA_AVAILABLE) || defined(TURING_MMA_AVAILABLE) || defined(AMD_WMMA_AVAILABLE)
```

michaelw9999:
@JohannesGaessler I put those commits in.  I originally was saving `MMQ_MMA_TILE_X_K_NVFP4` for the Blackwell kernels (future PR) but that's just going to use `MMQ_MMA_TILE_X_K_NVFP4_BW`.

JohannesGaessler:
That value should only ever be used in device code I think, so just define it differently depending on the CUDA architecture.

JohannesGaessler (approved):

ggml/src/ggml-cuda/common.cuh
```diff
@@ -800,19 +800,32 @@ static __device__ __forceinline__ float ggml_cuda_e8m0_to_fp32(uint8_t x) {
...
 #else
+#if defined(FP8_AVAILABLE) && !defined(GGML_USE_HIP)
+    const uint32_t bits = x * (x != 0x7F && x != 0xFF); // Convert NaN to 0.0f to match CPU implementation.
     const __nv_fp8_e4m3 xf = *reinterpret_cast<const __nv_fp8_e4m3 *>(&bits);
-#endif // defined(GGML_USE_HIP) && defined(GGML_USE_HIP)
     return static_cast<float>(xf) / 2;
 #else
-    NO_DEVICE_CODE;
-#endif // FP8_AVAILABLE
+    if (x == 0 || (x == 0x7F && x != 0xFF)) { // Convert NaN to 0.0f
```
ggerganov, line 815: @michaelw9999 This condition looks incorrect. Is it supposed to be:


```suggestion
    if (x == 0 || x == 0x7F || x == 0xFF) { // Convert NaN to 0.0f
```

michaelw9999: Yes, that is incorrect and it is a mistake, all 3 of those conditions separately should resolve to NaN.

PR #21075, opened 2026-03-27
fix cmake problem to exclude CCAN

sunqingn7:
## Additional information

# Requirements

- AI usage disclosure:


taronaeo:
What problem are you trying to fix? Do you have a documented issue?

sunqingn7:
oh, this is to fix the problem compiling with cuda on, cann module throw
exception expecting huawei acsend lib installed. Should bypass cann.

taronaeo:
> oh, this is to fix the problem compiling with cuda on, cann module throw exception expecting huawei acsend lib installed. Should bypass cann.

So are you compiling with both the CUDA backend and the CANN backend turned on? I'm not sure if I am understanding this correctly because I don't know of an instance where someone would need both CUDA and CANN backend in one binary.

If this is a legitimate issue, it would be better if an issue ticket can be filed with the full reproduce and logs for the maintainers to take a look into and have a better understanding of where these problems are coming from. Otherwise, I think the easier method is to just `-DGGML_CUDA=OFF`?

sunqingn7:
Ok
Basically I have -DGGML_CUDA=ON however cmake failed because cann had been
checked (and failed of course since I don't have it in my system).
Any way will try latest code base when get time and if reproducible (not
fixed) I will file a ticket instead.
Thanks.

PR #3552, opened 2023-10-09
Enable llama.cpp on s390x big endian platform

chenqiny:
This pull request aims to enable the execution of llama.cpp on the s390x big endian platform. It includes the following changes:

- Added a conditional check to determine if the platform is s390x. If this condition is met, the immintrin.h header file will not be imported, as it is not compatible with the s390x architecture.
- Introduced the --bigendian option to the conversion script for s390x, ensuring gguf model compatibility with the big endian byte order of the platform.

Verification:
To ensure the changes made in this pull request, the following verifications have been performed:

- Tested baichuan7b with float16.
- Tested baichuan7b with float16.
- Tested chinese-alpaca-2-13b with float16.

Please review this pull request and consider merging it into the main repository. Thank you!

Fixes #3298

PR #21078, opened 2026-03-27
gguf: use reversed magic "FUGG" for big-endian files

### LINKED ISSUE #3957: GGUF endianness cannot be determined from GGUF itself

Scottcjn:
Fixes #3957

## Summary

GGUF files currently have no way to self-describe their endianness from the magic bytes alone. This adds a reversed magic `"FUGG"` (`0x47475546`) for big-endian files so endianness is deterministic from the first 4 bytes, as proposed in the original issue by @philpax.

## Changes

| File | Change |
|------|--------|
| `gguf.h` | Add `GGUF_MAGIC_BE "FUGG"` constant |
| `gguf.cpp` | Read: accept both `"GGUF"` (LE) and `"FUGG"` (BE) magic. Write: use `__BYTE_ORDER__` to select correct magic |
| `constants.py` | Add `GGUF_MAGIC_BE = 0x47475546` |
| `gguf_reader.py` | Detect both magics, set `byte_order` accordingly |
| `gguf_writer.py` | Write `GGUF_MAGIC_BE` when `endianess == BIG` |
| `gguf_convert_endian.py` | Swap magic bytes when converting |
| `test-gguf.cpp` | Add BE magic test case, update bad magic `"FUGG"` → `"BAD!"` |

**72 insertions, 7 deletions across 7 files.**

## Backward Compatibility

- Existing little-endian files with `"GGUF"` magic load unchanged
- The read path accepts both magics and sets the endianness flag accordingly
- The version-field heuristic from #3552 still works as a fallback

## Testing

- **x86_64** (little-endian): reads both LE and BE files, writes `"GGUF"`
- **POWER8 ppc64** (big-endian): writes `"FUGG"`, reads both formats
- We run the only active big-endian llama.cpp deployment (IBM POWER8 S824, 512GB RAM, 147 tok/s on TinyLlama)

## Design Decision

Using reversed magic (`"FUGG"`) rather than a metadata field because:
1. Magic is the first thing read — endianness must be known before parsing any metadata
2. Existing LE readers will immediately reject BE files (unknown magic) instead of silently misinterpreting fields
3. Matches the approach used by ELF (different `EI_DATA` byte), TIFF (different byte order marker), and other binary formats

### DIFF
--- a/ggml/include/gguf.h
+++ b/ggml/include/gguf.h
@@ -38,7 +38,8 @@
 #include <stdbool.h>
 #include <stdint.h>
 
-#define GGUF_MAGIC   "GGUF"
+#define GGUF_MAGIC    "GGUF"
+#define GGUF_MAGIC_BE "FUGG"  // Reversed magic for big-endian GGUF files (issue #3957)
 #define GGUF_VERSION 3
 
 #define GGUF_KEY_GENERAL_ALIGNMENT "general.alignment"
--- a/ggml/src/gguf.cpp
+++ b/ggml/src/gguf.cpp
@@ -323,6 +323,9 @@ struct gguf_context * gguf_init_from_file_impl(FILE * file, struct gguf_init_par
     bool ok = true;
 
     // file magic
+    // Accepts both "GGUF" (little-endian) and "FUGG" (big-endian, reversed magic).
+    // Big-endian GGUF files use reversed magic so endianness is self-describing (issue #3957).
+    bool file_is_be = false;
     {
         std::vector<char> magic;
         ok = ok && gr.read(magic, 4);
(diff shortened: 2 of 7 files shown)


Scottcjn:
**Test results on POWER8 ppc64le (IBM S824, 64 threads):**

```
$ cd build-be-test && cmake .. -DLLAMA_BUILD_TESTS=ON && make test-gguf -j32
$ ./bin/test-gguf

70/70 tests passed
OK
```

Includes the new `HANDCRAFTED_HEADER_BE_MAGIC` test case — C++ read path correctly accepts both `"GGUF"` (LE) and `"FUGG"` (BE) magic.

Python constants verified:
```python
>>> GGUF_MAGIC    = 0x46554747  # b'GGUF'
>>> GGUF_MAGIC_BE = 0x47475546  # b'FUGG'
```

Note: Our POWER8 runs ppc64**le** (little-endian), so the write path produces `"GGUF"` natively. The `"FUGG"` write path is exercised through the Python writer's `GGUFEndian.BIG` mode and the endian conversion script. A true big-endian host (ppc64 BE or s390x) would exercise the C write path's `__BYTE_ORDER__` branch natively.

JohannesGaessler:
Reading the contributing guidelines is not optional, this is your one warning.

PR #21081, opened 2026-03-27
vulkan: add noncontiguous GLU support

0cc4m:
This was missing for Mistral-Small-4 to work fully on Vulkan.

Qwen3.5 122B Q4_0 wrote the code, fully locally. I edited some minor style problems and fixed a compile issue that I ran into. Quite impressive for a local model.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -1112,6 +1112,16 @@ struct vk_op_glu_push_constants {
     uint32_t mode;  // 0: default, 1: swapped, 2: split
     float alpha; // for swiglu_oai
     float limit;
+    uint32_t nb01;
+    uint32_t nb02;
+    uint32_t nb03;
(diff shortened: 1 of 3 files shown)


CISC (approved):

CISC:
Cool, I thought Vulkan already supported this (or at least contiguous_1). :)

I think Qwen3.5 has been extensively trained on llama.cpp, it seems to know a lot about it.

ggerganov:
> Qwen3.5 122B Q4_0 wrote the code, fully locally. 

Which client do you use?

0cc4m:
This was OpenCode, but frankly the default config of that project is really sketchy. Every edit or command is allowed, and it autoformats entire files when they get edited. I don't know what they are thinking. I'll look around for something saner, otherwise I'll have to spend some time checking and configuring everything.

ggerganov (approved):
Yeah. With native tools in `llama-server` added soon, we'll just be able to use the built-in WebUI for such tasks.

PR #21082, opened 2026-03-27
common: add bounds check in common_init_result::sampler to prevent segfault on failed model load

mtmcp:
## Overview

This fix supercedes #21049 to fix the segfault when non-existent models are loaded. It reverts the previous fix because it is no longer necessary with this more correct fix in place. In addition a regression test was added to expose the issue and showcase the fix.

_Before:_
```
./build/bin/test-init-fail
build: 8557 (43d7135f7) with GNU 13.3.0 for Linux x86_64
gguf_init_from_file: failed to open GGUF file '/non_existent_model.gguf' (No such file or directory)
llama_model_load: error loading model: llama_model_loader: failed to load model from /non_existent_model.gguf
llama_params_fit: encountered an error while trying to fit params to free device memory: failed to load model
llama_params_fit: fitting params to free memory took 0.00 seconds
gguf_init_from_file: failed to open GGUF file '/non_existent_model.gguf' (No such file or directory)
llama_model_load: error loading model: llama_model_loader: failed to load model from /non_existent_model.gguf
common_init_from_params: failed to load model '/non_existent_model.gguf'
```

_After:_
```
./build/bin/test-init-fail
build: 8557 (43d7135f7) with GNU 13.3.0 for Linux x86_64
gguf_init_from_file: failed to open GGUF file '/non_existent_model.gguf' (No such file or directory)
llama_model_load: error loading model: llama_model_loader: failed to load model from /non_existent_model.gguf
llama_params_fit: encountered an error while trying to fit params to free device memory: failed to load model
llama_params_fit: fitting params to free memory took 0.00 seconds
gguf_init_from_file: failed to open GGUF file '/non_existent_model.gguf' (No such file or directory)
llama_model_load: error loading model: llama_model_loader: failed to load model from /non_existent_model.gguf
main: OK
common_init_from_params: failed to load model '/non_existent_model.gguf'
```

# Requirements

- AI usage disclosure: YES AI was used to write the unit test


ggerganov (approved):

tests/test-init-fail.cpp
aldehir, line 1: I think this is a pretty simple fix, does it really warrant a dedicated test file?

aldehir (approved):

tests/test-init-fail.cpp
mtmcp: Nothing else stood out to me as a place to put the test but the file could be renamed something more general for testing other aspects of common library as well.

ggerganov: Yeah, I agree with @aldehir, let's remove the test. It's not really worth adding one more binary to the build.

mtmcp: Sounds good.

pwilkin (approved):

PR #21084, opened 2026-03-27
cli : add /glob command

CISC:
## Overview

Add /glob command to be able to batch load text files.

# Requirements

- AI usage disclosure: nah

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -656,6 +656,38 @@ bool string_parse_kv_override(const char * data, std::vector<llama_model_kv_over
     return true;
 }
 
+// simple glob: * matches non-/ chars, ** matches anything including /
+static inline bool glob_match(const char * pattern, const char * str) {
+    if (*pattern == '\0') {
+        return *str == '\0';
+    }
+    if (pattern[0] == '*' && pattern[1] == '*') {
+        const char * p = pattern + 2;
+        if (*p == '/') p++;
+        if (glob_match(p, str)) return true;
+        if (*str != '\0') return glob_match(pattern, str + 1);
+        return false;
+    }
+    if (*pattern == '*') {
+        const char * p = pattern + 1;
+        for (; *str != '\0' && *str != '/'; str++) {
+            if (glob_match(p, str)) return true;
+        }
+        return glob_match(p, str);
+    }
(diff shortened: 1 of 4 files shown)


pwilkin (changes requested):
Either `add_text_to_file` should have a size limit or we should approve the files to read - there's way too much risk of reading some binary or other huge file by accident and poisoning the context IMO.

CISC:
> Either `add_text_to_file` should have a size limit or we should approve the files to read - there's way too much risk of reading some binary or other huge file by accident and poisoning the context IMO.

If you're a dumbass and use `*`, sure. :)

Only way to prevent that is to somehow inspect and verify files, that gets way too complicated, and limiting on file size is too arbitrary. It does list the files though, so you can catch the mistake and use `/clear` before actually submitting them.

pwilkin (approved):

common/common.cpp
```diff
@@ -656,6 +656,38 @@ bool string_parse_kv_override(const char * data, std::vector<llama_model_kv_over
...
+    if (*pattern == '\0') {
+        return *str == '\0';
+    }
+    if (pattern[0] == '*' && pattern[1] == '*') {
+        const char * p = pattern + 2;
+        if (*p == '/') p++;
+        if (glob_match(p, str)) return true;
+        if (*str != '\0') return glob_match(pattern, str + 1);
+        return false;
+    }
```
ngxson, line 670: claude gives this suggestion, not sure if that makes sense (feel free to ignore it if it's irrelevant)

```cpp
if (pattern[0] == '*' && pattern[1] == '*') {
    const char * p = pattern + 2;
    if (*p == '/') p++;
    // try matching the rest of pattern at every position in str
    do {
        if (glob_match(p, str)) return true;
    } while (*str != '\0' && str++);
    return false;
}
```

ngxson (approved):
Re. risk of invisibly / accidentally reading more files than needed, I think we can improve the UX by first listing a list of files, then ask Y/N to confirm read. But this can be a follow-up improvement.

common/common.cpp
CISC: I thought Claude already wrote this in your PR? :)

CISC: It really makes no sense, I think Claude got confused at the intention.

PR #21090, opened 2026-03-28
server/webui: cleanup dual representation approach, simplify to openai-compat

pwilkin:
## Overview

Removes the double-representation from WebUI, bases everything on OpenAI-compat message history

## Additional information

Since the current dual representation approach seems to run into a lot of problems (see eg. #21087 ) and will probably run into more when we add the MCP proxy server, I reckoned I'd try to clean it up by removing the double representation and making everything derive from simple chat history - and it seems to be working out pretty well from my tests (tested both normal reasoning and reasoning with tool calling).

# Requirements

- AI usage disclosure: Yes, no way I rewrite all of that by hand :stuck_out_tongue: 

@allozaur @ServeurpersoCom could you take a look and see if it makes sense to you?


ServeurpersoCom:
This will be a major refactor with many edge cases!

pwilkin:
@ServeurpersoCom I thought so too, but it went surprisingly smoothly, all tests pass and nothing seems broken at least with cursory testing, so maybe it's not as bad as it looks?

ServeurpersoCom:
Technically it only touches the storage/rendering side, I'll pull the branch tomorrow and give it a proper test run, because honestly this is how it was originally supposed to be done! This is really the final piece of the MCP client integration, which started as a simple OAI-compat proxy outside the WebUI.

That said, there are a lot of subtle edge cases we've been polishing over the past weeks (streaming interruptions mid-tool-call, partial marker handling, continue generation with open reasoning blocks, tool result truncation, etc.), so I want to make sure nothing regresses before giving the thumbs up!

ServeurpersoCom:
I'm currently testing it: it's encouraging, it just works. This layer of abstraction was indeed the final blemish on our implementation. And it was much less disruptive than I thought because we naturally converged towards the same structure.

Some obsolete Settings options to remove:
- showRawOutputSwitch
- alwaysShowAgenticTurns
- maxToolPreviewLines

I'm keeping your commit on my everyday server; after a few days of use, I'll be able to spot any edge cases I might have missed.

allozaur:
I don't thikn `showRawOutputSwitch` is obsolete, we still might want to use it to debug Markdown rendering

ServeurpersoCom:
> I don't thikn `showRawOutputSwitch` is obsolete, we still might want to use it to debug Markdown rendering

Ah yes, it needs to be simplified/reworked (broken on the commit for now). We need it for markdown.

ServeurpersoCom (approved):
Honestly, we can merge quickly and fix what we find, we'll add commits on top and we'll make progress! It works.

allozaur:
> Honestly, we can merge quickly and fix what we find, we'll add commits on top and we'll make progress! It works.

Not exactly 😅 Let's first address any regressions that this PR introduces and then let's merge.

tools/server/webui/src/lib/stores/conversations.svelte.ts
```diff
@@ -122,6 +122,10 @@ class ConversationsStore {
 		if (this.isInitialized) return;
 
 		try {
+			// Run one-time migration for old marker-based messages
+			const { runLegacyMigration } = await import('$lib/utils/legacy-migration');
+			await runLegacyMigration();
```
allozaur, line 133: let's add some `@deprecated` markers in the code so that we can flag this and easily find in the future to remove this code

tools/server/webui/src/lib/types/chat.d.ts
```diff
@@ -99,21 +99,28 @@ export interface ChatMessageToolCallTiming {
...
 /**
- * Callbacks for streaming chat responses
+ * Callbacks for streaming chat responses (used by both agentic and non-agentic paths)
  */
 export interface ChatStreamCallbacks {
 	onChunk?: (chunk: string) => void;
 	onReasoningChunk?: (chunk: string) => void;
-	onToolCallChunk?: (chunk: string) => void;
-	onAttachments?: (extras: DatabaseMessageExtra[]) => void;
+	onToolCallsStreaming?: (toolCalls: import('./api').ApiChatCompletionToolCall[]) => void;
```
allozaur: Let's not use dynamic imports unless it's necessary.

tools/server/webui/src/lib/utils/legacy-migration.ts
allozaur, line 1: yeah, let's add some `@deprecated` JSDocs comments

ServeurpersoCom:
> Not exactly 😅 Let's first address any regressions that this PR introduces and then let's merge.

We're going to work together to eradicate them 😅

pwilkin:
Can you tell me what the regressions found are?

allozaur:
> Can you tell me what the regressions found are?

I am yet to finish reviewing and testing this PR tomorrow, I will let u know from my side.

allozaur:
@pwilkin i've fixed the agentic turns UI regression + added few more tests. This is a middle-step to what we've internally discussed about having separate structure for chat messages data that is 100% OpenAI compatible, but on the UI end we want to show the assistant + tool messages as a joint agentic chain of messages

ServeurpersoCom:
Cool, it's definitely more compact this way while retaining the advantages of clean, structured data, good thinking Alek!

Sans titre

allozaur (approved):

ServeurpersoCom (approved):

allozaur:
> All that remains is to correct the agent turn display options (visually useful only if there are multiple CoT + Toolcall pairs), and fix the raw Markdown display switch. I've rebased the pull request on my server to try more:)

these should already be fixed by now

PR #21092, opened 2026-03-28
gguf: add big-endian magic for self-describing endianness

### LINKED ISSUE #3957: GGUF endianness cannot be determined from GGUF itself

Scottcjn:
Fixes #3957

Hey — I run llama.cpp on an IBM POWER8 server (ppc64le, 512GB RAM) and hit the endianness detection issue firsthand. The current GGUF reader has to guess endianness from the version field which is fragile. philpax suggested using reversed magic back in 2023 and it never got done, so here it is.

The idea is simple: big-endian GGUF files start with "FUGG" instead of "GGUF". Reader sees "GGUF" = little-endian, "FUGG" = big-endian, anything else = reject. Same approach as ELF and TIFF.

What I changed:
- Added GGUF_MAGIC_BE in the header and Python constants
- C++ read path checks for both magics, sets a flag
- C++ write path uses __BYTE_ORDER__ to pick the right magic on the host
- Python reader detects both and sets byte_order
- Python writer respects the endianness setting when writing magic
- Convert script swaps magic when converting between endiannesses
- Added a test case for BE magic in test-gguf (changed the old "FUGG" bad-magic test to use "BAD!" since FUGG is now valid)

72 insertions, 7 deletions. Existing LE files still work exactly the same.

I built and ran the tests on our POWER8 — 70/70 passed including the new BE magic case.

AI disclosure: I used Claude as a coding assistant to help locate all the relevant code paths across the C++ and Python codebases. The approach (reversed magic) was designed by me based on philpax's original suggestion in #3957. I can explain every line.

### DIFF
--- a/ggml/include/gguf.h
+++ b/ggml/include/gguf.h
@@ -38,7 +38,8 @@
 #include <stdbool.h>
 #include <stdint.h>
 
-#define GGUF_MAGIC   "GGUF"
+#define GGUF_MAGIC    "GGUF"
+#define GGUF_MAGIC_BE "FUGG"  // Reversed magic for big-endian GGUF files (issue #3957)
 #define GGUF_VERSION 3
 
 #define GGUF_KEY_GENERAL_ALIGNMENT "general.alignment"
--- a/ggml/src/gguf.cpp
+++ b/ggml/src/gguf.cpp
@@ -323,6 +323,9 @@ struct gguf_context * gguf_init_from_file_impl(FILE * file, struct gguf_init_par
     bool ok = true;
 
     // file magic
+    // Accepts both "GGUF" (little-endian) and "FUGG" (big-endian, reversed magic).
+    // Big-endian GGUF files use reversed magic so endianness is self-describing (issue #3957).
+    bool file_is_be = false;
     {
         std::vector<char> magic;
         ok = ok && gr.read(magic, 4);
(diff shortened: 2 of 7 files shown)


Scottcjn:
Re: the AI flag — I use Claude Code as a coding assistant and disclosed this in the PR description. This is real hardware-tested code from my lab. Here's the proof:

**Test system:** IBM POWER8 S824 (ppc64le, 502GB RAM, 64 threads)

```
sophia@s822-control:~/llama.cpp/build-be-test$ uname -a
Linux s822-control 6.1.159-powerelyan-hsa #1 SMP Sat Jan 10 16:44:18 UTC 2026 ppc64le ppc64le ppc64le GNU/Linux

sophia@s822-control:~/llama.cpp/build-be-test$ ./bin/test-gguf

test_handcrafted_file: handcrafted_file_type=HEADER_BAD_MAGIC
test_handcrafted_file:   - context_null: OK

test_handcrafted_file: handcrafted_file_type=HEADER_BE_MAGIC
test_handcrafted_file:   - context_not_null: OK
test_handcrafted_file:   - check_header: OK

test_handcrafted_file: handcrafted_file_type=HEADER_EMPTY
test_handcrafted_file:   - context_not_null: OK
test_handcrafted_file:   - check_header: OK

[... 60 more tests ...]

70/70 tests passed
OK
```

The HEADER_BE_MAGIC test is the new one — C++ read path correctly accepts both \GGUF\ (LE) and \FUGG\ (BE). All existing tests still pass. I can explain every line of this PR.

— Scott

am17an:
You have 13 closed PRs because of AI slop https://github.com/ggml-org/llama.cpp/pulls?q=is%3Apr+author%3AScottcjn+is%3Aclosed - please stop submitting more or you will be banned

JohannesGaessler:

Why did you choose to add a bunch of EM dashes to your posts in this PR when this makes it look like you're just copypasting language model outputs?

Scottcjn:
Understood. Closing. Will never provide work here again over semantic writing style of AI "slop". The big-endian GGUF work is available in my fork if anyone wants it. - Scott

PR #21093, opened 2026-03-28
[SYCL] Enhance build script to use half cores to build, avoid OS hang

arthw:
After FA is supported, the number of source files become more.
When build the source code by SYCL default build script with all CPU cores on PC with small memory (like 16GB), it will lead to the memory exhausted。
Enhance build script to use half cores to build, avoid OS hang.

### DIFF
--- a/examples/sycl/build.sh
+++ b/examples/sycl/build.sh
@@ -20,4 +20,4 @@ cmake .. -DGGML_SYCL=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx -DLLAMA
 #cmake --build . --config Release --target llama-bench
 
 #build all binary
-cmake --build . --config Release -j -v
+cmake --build . --config Release -j$(($(nproc)/2)) -v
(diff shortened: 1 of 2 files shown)


ggerganov (approved):

examples/sycl/build.sh
```diff
@@ -20,4 +20,4 @@ cmake .. -DGGML_SYCL=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx -DLLAMA
 #cmake --build . --config Release --target llama-bench
 
 #build all binary
-cmake --build . --config Release -j -v
+cmake --build . --config Release -j$(($(nproc)/2)) -v
```
CISC:
```suggestion
cmake --build . --config Release -j$((($(nproc)+1)/2)) -v
```

CISC (changes requested):

examples/sycl/build.sh
arthw: Yes, update as the suggestion!
Thank you!

CISC (approved):

PR #21097, opened 2026-03-28
kv-cache: bounded budget with residual-stream recomputation (KV Direct)

malibujack:
Adds optional bounded KV cache that doesn't permanently lose evicted tokens. Instead, it saves layer-0 embeddings to a host-memory ring buffer and recomputes K/V on demand when evicted positions are needed again.

The key insight is that the residual stream at layer 0 encodes everything needed to reconstruct K/V — so saving ~5 KB of residual per token lets you recover hundreds of KB of cached state. This is independently validated by [arXiv:2603.19664](https://arxiv.org/abs/2603.19664) ("The Residual Stream Is All You Need"), which shows 100% token match at every budget level vs permanent-loss baselines like H2O, StreamingLLM, and SnapKV.

**What it does:**
- `--kv-budget-tokens N` or `--kv-budget-mb N` caps the cache at a fixed size
- LRU eviction kicks in when over budget
- Evicted positions get recomputed from saved residuals (batched, capped at 64 per cycle)
- Budget=0 (default) disables everything — zero overhead, identical to current behavior

**API additions:**
- `llama_kv_direct_evict(ctx)` — enforce budget, evict LRU positions
- `llama_kv_direct_recompute_misses(ctx)` — restore evicted positions from residual pool

I've been running this in production for a few weeks on Qwen3 models with a 3090 Ti. Memory stays flat during long multi-turn conversations while output quality matches unbounded cache.

**Disclosure:** AI tools (Claude) assisted with code development and the initial PR description.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -1239,6 +1239,20 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_env("LLAMA_ARG_CTX_SIZE"));
     add_opt(common_arg(
+        {"--kv-budget-mb"}, "N",
+        "cap KV cache at N megabytes, 0 = unbounded (default: 0)",
+        [](common_params & params, int value) {
+            params.kv_budget_bytes = (uint64_t)value * 1024 * 1024;
+        }
+    ).set_env("LLAMA_ARG_KV_BUDGET_MB").set_examples({LLAMA_EXAMPLE_SERVER, LLAMA_EXAMPLE_CLI}));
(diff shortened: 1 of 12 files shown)


malibujack:
Hopefully that's okay :)

ggerganov:
Nope, too much slop

malibujack:
Fair enough, your loss.  Take care.

malibujack:
Looks like it made it eventually through a different interpretation by someone "known".  Go figure

Issue #19858, opened 2026-02-24
Eval bug: Qwen3.5 always re-processes the full prompt


NickM-27:
### Name and Version

build: 8140 (https://github.com/ggml-org/llama.cpp/commit/39fb81f8756208d79dd12ad89359f3b31675e20d) with GNU 15.2.0 for Linux x86_64

### Operating systems

Linux

### GGML backends

Vulkan

### Hardware

RX 7900XTX

### Models

Qwen3.5 Q4_K_XL Unsloth GGUF

### Problem description & steps to reproduce

I am running llama-server in docker in router mode. When I load Qwen3.5 I see the below logs which show it is not re-using the context. I set the below config so there should be plenty of context for it

```
; Context Size
ctx-size = 100000
parallel = 4 
slot-prompt-similarity = 0.25
```

### Relevant log output

<details>
<summary>Logs</summary>

```console
srv  proxy_reques: proxying request to model Qwen3.5 on port 53409
[53409] srv  params_from_: Chat format: peg-constructed
[53409] slot get_availabl: id  3 | task -1 | selected slot by LCP similarity, sim_best = 0.988 (> 0.250 thold), f_keep = 0.940
[53409] slot launch_slot_: id  3 | task -1 | sampler chain: logits -> penalties -> ?dry -> ?top-n-sigma -> top-k -> ?typical -> top-p -> ?min-p -> ?xtc -> ?temp-ext -> dist 
[53409] slot launch_slot_: id  3 | task 2400 | processing task, is_child = 0
[53409] slot update_slots: id  3 | task 2400 | new prompt, n_ctx_slot = 25088, n_keep = 0, task.n_tokens = 7238
[53409] slot update_slots: id  3 | task 2400 | erased invalidated context checkpoint (pos_min = 7536, pos_max = 7536, n_swa = 1, size = 62.813 MiB)
[53409] slot update_slots: id  3 | task 2400 | n_tokens = 7153, memory_seq_rm [7153, end)
[53409] slot update_slots: id  3 | task 2400 | failed to truncate tokens with position >= 7153 - clearing the memory
[53409] slot prompt_clear: id  3 | task 2400 | clearing prompt with 7153 tokens
[53409] slot update_slots: id  3 | task 2400 | prompt processing progress, n_tokens = 2048, batch.n_tokens = 2048, progress = 0.282951
[53409] slot update_slots: id  3 | task 2400 | n_tokens = 2048, memory_seq_rm [2048, end)
[...]

PR #21099, opened 2026-03-28
server: fix prompt cache reuse for hybrid/recurrent models

### LINKED ISSUE #20225: Eval bug: Qwen 3.5 Full prompt re-processing on every conversation turn

hanxiao:
## Problem

Hybrid/recurrent models (Qwen3.5, Falcon-H1, Jamba, Nemotron, etc.) trigger full prompt re-processing at long context lengths, even when the prompt prefix is unchanged. This makes memory-store use cases (large static context + small varying query) impractical.

### Observed behavior with upstream Docker image

Tested with `ghcr.io/ggml-org/llama.cpp:server-cuda` (latest, 2026-03-27 build, includes #16382 + #19045):

| Context size | Cold prefill | Cache reuse (Q2) | Speedup | Status |
|-------------|-------------|-------------------|---------|--------|
| 23K tokens | 20.9s | 0.8s | 24.9x | ✅ Works |
| 112K tokens | 136.9s | 1.5s | 91.5x | ✅ Works |
| **258K tokens** | **437.6s** | **437.6s (full re-process)** | **1x** | ❌ **Broken** |

At 258K tokens, the server log shows all 32 checkpoints being erased and full re-processing:
```
slot update_slots: id 0 | task 259 | n_past = 11, pos_min = 258048, n_swa = 1
slot update_slots: id 0 | task 259 | Checking checkpoint with [258036, 258036] against 10...
...  (all 32 checkpoints checked, none match)
slot update_slots: id 0 | task 259 | erased invalidated context checkpoint (pos_min = 225279, ...)
slot update_slots: id 0 | task 259 | erased invalidated context checkpoint (pos_min = 227327, ...)
...  (all 32 checkpoints erased)
slot update_slots: id 0 | task 259 | n_tokens = 0, memory_seq_rm [0, end)
```

The bug is scale-dependent: cache reuse works at shorter contexts but breaks at longer ones because the checkpoint position thresholds grow beyond any checkpoint's range.

---

## Root cause

For recurrent layers, `llama_memory_seq_pos_min()` returns the full sequence length (not the SWA window position). The existing threshold logic `pos_min_thold = max(0, pos_next - n_swa)` evaluates to `pos_next` when `n_swa == 0` (Qwen3.5 has no SWA), so:

1. `pos_min >= pos_min_thold` is always true → enters checkpoint search
2. `cur.pos_min < pos_min_thold` finds no match (threshold too large at long contexts)
3. `do_reset` zeros `n_past` → full re-processing
4. All checkpoints with `pos_max > pos_next` get erased

At shorter contexts, the prompt cache similarity matching (`sim_best`) finds a usable cached prompt before the checkpoint logic triggers, masking the bug. At 250K+, this fallback fails.

## Fix

Four changes, all gated behind `llama_model_is_recurrent()`:

1. **`pos_min_thold = 0`** for recurrent models (SWA threshold not meaningful for recurrent state)
2. **Checkpoint search** uses position-matching semantics (`cur.pos_max <= n_past`) instead of SWA window logic
3. **`do_reset` path** preserves `n_past` instead of zeroing it (avoids discarding valid cache)
4. **Log cleanup**: checkpoint erasure message no longer references `n_swa`

Pure transformer / SWA paths are completely unchanged.

## Testing

Tested on **Qwen3.5-35B-A3B** (30x GDN + 10x Gated Attention) with 250K token context on NVIDIA L4 (24GB):

| Metric | Before (upstream) | After (this PR) |
|--------|------------------|-----------------|
| Cold prefill (240K tokens) | 408s | 408s |
| Subsequent query (same prefix) | **408s** (full re-process) | **2.6s** (cache reuse) |
| Speedup | 1x | **157x** |
| NIAH retrieval accuracy | 8/8 | 8/8 |

8 different queries tested with 250K static context (needle-in-a-haystack at various depths), all correctly answered in 2.4-2.8s after initial prefill.

Server log after fix (no reprocessing warning):
```
slot update_slots: id 0 | task 79 | new prompt, n_ctx_slot = 262144
(prompt processing proceeds from cached position, only new tokens processed)
```

Also verified at 100K context: 115s cold → 1.7s warm (68x speedup).

## Related issues

Fixes #20225
Related: #19858, #19690, #19794, #18497

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -2307,11 +2307,14 @@ private:
 
                             llama_pos pos_next = slot.prompt.tokens.pos_next(n_past);
 
+                            const bool is_recurrent = llama_model_is_recurrent(model);
+
                             // note: when n_swa == 0, the model does not use SWA
                             const auto n_swa = std::max(0, llama_model_n_swa(model));
 
-                            // the largest pos_min required for a checkpoint to be useful
-                            const auto pos_min_thold = std::max(0, pos_next - n_swa);
+                            // For hybrid/recurrent: SWA threshold not meaningful, set to 0.
+                            // For pure transformer/SWA: preserve existing behavior.
+                            const auto pos_min_thold = is_recurrent ? (llama_pos) 0 : (llama_pos) std::max(0, pos_next - n_swa);
 
                             if (n_past > 0 && n_past < slot.prompt.n_tokens()) {
                                 const auto pos_min = llama_memory_seq_pos_min(llama_get_memory(ctx), slot.id);
@@ -2371,6 +2374,10 @@ private:
                                         slot.prompt.checkpoints.rbegin(),
                                         slot.prompt.checkpoints.rend(),
                                         [&, func_name = __func__](const auto & cur) {
+                                            if (is_recurrent) {
+                                                // For hybrid/recurrent: use position-matching semantics
+                                                return cur.pos_max <= n_past && cur.pos_max < pos_next;
+                                            }
                                             // guarantee that a checkpoint will result in at least one token being processed [TAG_PROMPT_LOGITS]
                                             LOG_INF("slot %12.*s: id %2d | task %d | Checking checkpoint with [%d, %d] against %d...\n", 12,
                                                 func_name, (slot).id, ((slot).task ? (slot).task->id : -1), cur.pos_min, cur.pos_max, pos_min_thold);
@@ -2397,10 +2404,15 @@ private:
                                     }
 
                                     if (do_reset) {
-                                        SLT_WRN(slot, "forcing full prompt re-processing due to lack of cache data (likely due to SWA or hybrid/recurrent memory, see %s)\n",
-                                                "https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055");
-                                        pos_next = 0;
-                                        n_past = 0;
+                                        if (is_recurrent) {
+                                            // For hybrid/recurrent: preserve current state, do not zero n_past
+                                            SLT_WRN(slot, "no matching recurrent checkpoint; preserving prompt state (n_past = %d)\n", n_past);
+                                        } else {
+                                            SLT_WRN(slot, "forcing full prompt re-processing due to lack of cache data (likely due to SWA or hybrid/recurrent memory, see %s)\n",
+                                                    "https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055");
+                                            pos_next = 0;
+                                            n_past = 0;
+                                        }
                                     }
                                 }
                             }
@@ -2410,7 +2422,7 @@ private:
                                 for (auto it = slot.prompt.checkpoints.begin(); it != slot.prompt.checkpoints.end();) {
                                     const auto & cur = *it;
                                     if (cur.pos_max > pos_next) {
-                                        SLT_WRN(slot, "erased invalidated context checkpoint (pos_min = %d, pos_max = %d, n_tokens = %" PRId64 ", n_swa = %d, pos_next = %d, size = %.3f MiB)\n", cur.pos_min, cur.pos_max, cur.n_tokens, n_swa, pos_next, (float) cur.data.size() / 1024 / 1024);
+                                        SLT_WRN(slot, "erased invalidated context checkpoint (pos_min = %d, pos_max = %d, n_tokens = %" PRId64 ", pos_next = %d, size = %.3f MiB)\n", cur.pos_min, cur.pos_max, cur.n_tokens, pos_next, (float) cur.data.size() / 1024 / 1024);
                                         it = slot.prompt.checkpoints.erase(it);
                                     } else {
                                         ++it;


nssatlantis:
Would be extremely helpful to get in, long-context becomes a bit of a pain.

nssatlantis:
> > Would be extremely helpful to get in, long-context becomes a bit of a pain.
> 
> @nssatlantis Please provide a **minimal reproducible example** of what you call "a bit of a pain". This pull request claims to fix the already closed issue #20225. However, that issue describes a bug that does not actually exist, because in that case, the cache is invalidated correctly invalidated by OpenCode mutating the prefix. So if there is still a bug, please open a new issue with an example with which the bug can be reproduced and with which a potential fix can be verified whether it is working or not.
> 
> Note also that some parts of the proposed code change in this pull request seems odd. For example, it removes still relevant information from being logged, rather than extending it with the newly introduced `is_recurrent` flag.

From my quick perspective, it seemed like something that would definitely help the cache not invalidate itself, however, I haven't read through the code, just what it supposedly would fix.

I also learned that the cache does work to a greater extent, and half of the cache invalidations was due to injecting dynamic text into a prompt at the top, rather than the bottom.

So in short, hoped/thought this would help, but as is often the case... User error :)

PR #21056, opened 2026-03-27
server : add custom socket options to disable SO_REUSEPORT

angt:
## Overview

cpp-httplib uses `SO_REUSEPORT` by default, see #20963

## Additional information

https://github.com/yhirose/cpp-httplib/pull/2409
https://github.com/yhirose/cpp-httplib/pull/2411

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: NO

PR #21109, opened 2026-03-28
ci : add --reuse-port

angt:
## Overview

see #21056

## Additional information

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/tools/server/tests/utils.py
+++ b/tools/server/tests/utils.py
@@ -103,6 +103,7 @@ class ServerProcess:
     media_path: str | None = None
     sleep_idle_seconds: int | None = None
     webui_mcp_proxy: bool = False
+    reuse_port: bool = True
 
     # session variables
     process: subprocess.Popen | None = None
@@ -239,6 +240,8 @@ class ServerProcess:
             server_args.extend(["--sleep-idle-seconds", self.sleep_idle_seconds])
         if self.webui_mcp_proxy:
             server_args.append("--webui-mcp-proxy")
+        if self.reuse_port:
+            server_args.append("--reuse-port")
 
         args = [str(arg) for arg in [server_path, *server_args]]
         print(f"tests: starting server with: {' '.join(args)}")


ServeurpersoCom (approved):

angt:
Really not sure we should use `--reuse-port` at all. I’m highly skeptical about what is being done to be honest... 

This is quick fix (i hope) to unlock the CI...

ServeurpersoCom:
Yes, there must be a real underlying cause in the test scripts, but technically, with this option, it should work as before!

ServeurpersoCom:
On a personal project I was developing, I accidentally forgot about an old process->the kernel does TCP load balancing, and you can end up with the old process responding randomly! I had the old version of my WebUI without the modifications.  This will improve the reliability of the tests and clarify this issue.

I'm digging around with my agent to see where this process is dragging on; if I find anything, I'll put it here for you.

Perhaps it's just a matter of a synchronous wait that's missing!

ServeurpersoCom:
Add process.wait() after kill() to ensure the OS has fully released the socket before the next test tries to bind :

```
--- a/tools/server/tests/utils.py
+++ b/tools/server/tests/utils.py
@@ -289,6 +289,7 @@ class ServerProcess:
         if self.process:
             print(f"Stopping server with pid={self.process.pid}")
             self.process.kill()
+            self.process.wait()
             self.process = None

     def make_request(
root@pod:/mnt/workspace/llama.cpp#
```

angt:
> Add process.wait() after kill() to ensure the OS has fully released the socket before the next test tries to bind :
> 
> ```
> --- a/tools/server/tests/utils.py
> +++ b/tools/server/tests/utils.py
> @@ -289,6 +289,7 @@ class ServerProcess:
>          if self.process:
>              print(f"Stopping server with pid={self.process.pid}")
>              self.process.kill()
> +            self.process.wait()
>              self.process = None
> 
>      def make_request(
> root@pod:/mnt/workspace/llama.cpp#
> ```

Yes, see #21110 :)

PR #21111, opened 2026-03-28
common : add character class support to glob_match

CISC:
## Overview

Adds character class support to `glob_match`.

## Additional information

Allows matches like
- `[abc]`
- `[a-z]`
- `[!0-9]`

# Requirements

- AI usage disclosure: YES, asked Qwen3.5-27B to implement it

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -656,7 +656,47 @@ bool string_parse_kv_override(const char * data, std::vector<llama_model_kv_over
     return true;
 }
 
-// simple glob: * matches non-/ chars, ** matches anything including /
+static inline bool glob_class_match(const char & c, const char * pattern, const char * class_end) {
+    const char * class_start = pattern;
+    bool negated = false;
+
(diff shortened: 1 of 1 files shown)


common/common.cpp
```diff
@@ -656,7 +656,47 @@ bool string_parse_kv_override(const char * data, std::vector<llama_model_kv_over
     return true;
 }
 
-// simple glob: * matches non-/ chars, ** matches anything including /
+static inline bool glob_class_match(const char & c, const char * pattern, const char * class_end) {
```
aldehir: Const ref to a char seems pointless.

aldehir (approved):

pwilkin (approved):

common/common.cpp
CISC: LOL, true, wonder what that was about, thanks for catching it. :)

aldehir (approved):

PR #20929, opened 2026-03-24
CI: fix ARM64 image build error & enable compilation

Ts-sound:
## Overview

Fix ARM64 build failures caused by:
1. QEMU ARM64 random segmentation fault bug in older versions
2. Incorrect `armv9.2-a` detection in CMake (compiler only supports up to `armv8.6-a`)

Changes:
- Updated QEMU to `tonistiigi/binfmt:qemu-v10.2.1`
- Temporarily disabled armv9.2 detection in CMake (requires further review by maintainers)

## Additional Information

Related Issues:

- https://github.com/ggml-org/llama.cpp/issues/19177
- https://github.com/ggml-org/llama.cpp/issues/11888

Verified Build: https://github.com/Ts-sound/llama.cpp/actions/runs/23467958136

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: YES
    - AI assisted in summarizing my test notes

PR #21122, opened 2026-03-28
CI: Enable CUDA and Vulkan ARM64 runners and fix CI/CD

### LINKED ISSUE #11888: [Tracker] Docker build fails on CI for arm64
Tracker for upstream issue: https://github.com/docker/build-push-action/issues/1309

At the time of writing this issue, this has been failing for the past 3 days: https://github.com/ggml-org/llama.cpp/actions/workflows/docker.yml

It was fixed in https://github.com/ggml-org/llama.cpp/pull/11472 but now the problem came back.

CC @ggerganov @slaren for visibility

ehfd:
## Overview

Fixes #21123 

Follow-up of #20929 for enabling CUDA/Vulkan ARM64 runners and fixing ARM64 CUDA Dockerfiles (for GH200, DGX Spark, etc.), as well as vastly accelerating build time for ARM64 due to changing to native runners.

AMD ROCm is only built for x86_64, as AMD doesn't provide ARM64 base images.

CC @CISC @taronaeo 

## Additional information

- Enables gcc/g++ 14 in CUDA Dockerfile images and updates images to Ubuntu 24.04 and CUDA 12.5.1 for the CUDA 12 image, and build ARM64 Vulkan Dockerfile images

- Fixes a few aesthetic improvements (`libgomp1 curl\` -> `libgomp1 curl \`) in other Dockerfiles.

- Corresponding changes to `docker.yml` similar to #20929.

- Enable `ubuntu-24.04-arm` native architecture workers for ARM64 for vastly accelerated build times.

- **Properly handle multi-architecture container image manifests through digest-based upload, which was previously broken after #20929.**

# Requirements

- AI usage disclosure: Gemini 3.1 Pro, GPT-5.3-Codex, purely for review and minor assistance.


ehfd:
https://github.com/ehfd/llama.cpp/actions/runs/23704816053

@CISC @taronaeo Perfectly works (s390x was excluded due to the IBM runner not being available in the fork).

CISC:
This seems to overcomplicate things, why not just use multiplatform instead of this elaborate (and possibly flawed) scheme?

ehfd:
@CISC

Multiplatform build in one runner forces emulation in ARM64 because both x86_64 and arm64 builds happen in an x86_64 runner, taking possibly 5-6 times more per build (50 minutes to 1 hours and 15 minutes vs 4.5 - 5 hours in my experience). This will quickly lead to reaching the 6-hour limit of GitHub Runners in case the codebase expands and more things need to be compiled.

By using digest upload and subsequent manifest creation, different workers, native x86_64, arm64, and s390x runners can build different images that will point to the same tag through the manifest (https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

https://medium.com/@shrinidhi.atmakur/the-importance-of-docker-manifest-in-multi-platform-image-management-b64a91b6e060

This is standard in multi-architecture, multi-platform scenarios.

.github/workflows/docker.yml
```diff
@@ -25,27 +25,72 @@ permissions:
...
+            { "tag": "musa", "dockerfile": ".devops/musa.Dockerfile", "platforms": "linux/amd64", "full": true, "light": true, "server": true, "free_disk_space": true, "runs_on": "ubuntu-24.04" },
+            { "tag": "intel", "dockerfile": ".devops/intel.Dockerfile", "platforms": "linux/amd64", "full": true, "light": true, "server": true, "free_disk_space": true, "runs_on": "ubuntu-24.04" },
+            { "tag": "vulkan", "dockerfile": ".devops/vulkan.Dockerfile", "platforms": "linux/arm64", "full": true, "light": true, "server": true, "free_disk_space": false, "runs_on": "ubuntu-24.04" },
+            { "tag": "vulkan", "dockerfile": ".devops/vulkan.Dockerfile", "platforms": "linux/amd64", "full": true, "light": true, "server": true, "free_disk_space": false, "runs_on": "ubuntu-24.04" },
+            { "tag": "s390x", "dockerfile": ".devops/s390x.Dockerfile", "platforms": "linux/s390x", "full": true, "light": true, "server": true, "free_disk_space": false, "runs_on": "ubuntu-24.04-s390x" },
```
CISC:
```suggestion
            { "tag": "cpu", "dockerfile": ".devops/s390x.Dockerfile", "platforms": "linux/s390x", "full": true, "light": true, "server": true, "free_disk_space": false, "runs_on": "ubuntu-24.04-s390x" },
```
Can we not do this now if handled appropriately elsewhere? If necessary, keep the the `s390x` tag as well for a little while...

ehfd: Okay!

.github/workflows/docker.yml
```diff
@@ -54,33 +99,32 @@ jobs:
 
       - name: Set up QEMU
         if: ${{ matrix.config.tag != 's390x' }}
-        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3
+        uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4
         with:
           image: tonistiigi/binfmt:qemu-v10.2.1
```
CISC, line 144: Is this still needed?

ehfd: No, but I believe that it is useful for builder environments on `x86_64`, just in case there are other architectures that need to be emulated as a template in the future (it only takes three seconds, so keep it on `x86_64` in case it's needed in the future). I changed to match runners containing `linux/amd64` with `${{ contains(matrix.config.platforms, 'linux/amd64') }}`.

.github/workflows/docker.yml
ehfd: All solved and backward compatibility sections are commented as such.

ehfd:
https://github.com/ehfd/llama.cpp/actions/runs/23708426244

Builds are starting to pass (s390x is not included because the runner is not available in the fork).

CISC:
> https://github.com/ehfd/llama.cpp/actions/runs/23708426244
> 
> Builds are starting to pass (s390x is not included because the runner is not available in the fork).

The images don't seem to get tagged though:
https://github.com/ehfd/llama.cpp/pkgs/container/llama.cpp/versions?filters%5Bversion_type%5D=tagged

Edit: Ah, probably not until all pushes are done.

ehfd:
@CISC Correct. Does the final outcome look good to you?

.github/workflows/docker.yml
```diff
@@ -195,12 +246,178 @@ jobs:
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Determine source tag name
+        id: srctag
+        uses: ./.github/actions/get-tag-name
+        env:
+          BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
```
CISC: This should be retrieved from `push_to_registry`, otherwise we might get out of sync by the time `merge_arch_tags` runs.

CISC:
> @CISC Correct. Does the final outcome look good to you?

Yes, very nice, just the small nit above.

ehfd:
@CISC I obtained the value from `create_tag`, as I realized the same block was there. It's the earliest possible point.

Anyways, fixed.

.github/workflows/docker.yml
ehfd: The same block was present in `create_tag`, so obtained from there (done within the first 10-30 seconds).

.github/workflows/docker.yml
```diff
@@ -25,62 +25,147 @@ permissions:
...
+          echo "merge_matrix=$MERGE_MATRIX" >> "$GITHUB_OUTPUT"
+
   push_to_registry:
-    name: Push Docker image to Docker Hub
+    name: Push Docker image to Docker Registry
+    needs: prepare_matrices
 
     runs-on: ${{ matrix.config.runs_on }}
     env:
       COMMIT_SHA: ${{ github.sha }}
```
CISC:
```suggestion
```
This is some legacy leftover.

.github/workflows/docker.yml
```diff
@@ -25,62 +25,147 @@ permissions:
...
+            | [.[] | .arches = (.arches | unique | sort | join(" "))]
+          ' build-matrix.json)"
+
+          echo "build_matrix=$BUILD_MATRIX" >> "$GITHUB_OUTPUT"
+          echo "merge_matrix=$MERGE_MATRIX" >> "$GITHUB_OUTPUT"
+
   push_to_registry:
-    name: Push Docker image to Docker Hub
+    name: Push Docker image to Docker Registry
+    needs: prepare_matrices
```
CISC:
```suggestion
    needs: [prepare_matrices, create_tag]
```

.github/workflows/docker.yml
```diff
@@ -25,62 +25,147 @@ permissions:
...
-          - { tag: "vulkan", dockerfile: ".devops/vulkan.Dockerfile", platforms: "linux/amd64", full: true, light: true, server: true, free_disk_space: false, runs_on: "ubuntu-24.04" }
-          - { tag: "s390x",  dockerfile: ".devops/s390x.Dockerfile",  platforms: "linux/s390x", full: true, light: true, server: true, free_disk_space: false, runs_on: "ubuntu-24.04-s390x" }
-          - { tag: "rocm",   dockerfile: ".devops/rocm.Dockerfile",   platforms: "linux/amd64", full: true, light: true, server: true, free_disk_space: true,  runs_on: "ubuntu-24.04" }
-          - { tag: "openvino", dockerfile: ".devops/openvino.Dockerfile", platforms: "linux/amd64", full: true, light: true, server: true, free_disk_space: false, runs_on: "ubuntu-24.04" }
+        config: ${{ fromJSON(needs.prepare_matrices.outputs.build_matrix) }}
     steps:
       - name: Check out the repo
         uses: actions/checkout@v6
         with:
           fetch-depth: 0 # preserve git history, so we can determine the build number
```
CISC:
```suggestion
          fetch-depth: 0
          ref: ${{ needs.create_tag.outputs.source_tag }}
```

ehfd:
I just applied the suggestions from the web interface; you could squash-merge at the end (you can also direct-edit if you want if you have anything further).

CISC:
> I just applied the suggestions from the web interface; you could squash-merge at the end (you can also direct-edit if you want if you have anything further).

The merge to master will be squashed anyway, it doesn't matter.

Mind re-testing with latest changes?

CISC (approved):

Ts-sound:
Buidling on the native architecture and then merging the archives at the end ,  seems more reasonable than using QEMU.

.github/workflows/docker.yml
```diff
@@ -25,62 +25,146 @@ permissions:
...
+            { "tag": "intel", "dockerfile": ".devops/intel.Dockerfile", "platforms": "linux/amd64", "full": true, "light": true, "server": true, "free_disk_space": true, "runs_on": "ubuntu-24.04" },
+            { "tag": "vulkan", "dockerfile": ".devops/vulkan.Dockerfile", "platforms": "linux/amd64", "full": true, "light": true, "server": true, "free_disk_space": false, "runs_on": "ubuntu-24.04" },
+            { "tag": "vulkan", "dockerfile": ".devops/vulkan.Dockerfile", "platforms": "linux/arm64", "full": true, "light": true, "server": true, "free_disk_space": false, "runs_on": "ubuntu-24.04-arm" },
+            { "tag": "rocm", "dockerfile": ".devops/rocm.Dockerfile", "platforms": "linux/amd64", "full": true, "light": true, "server": true, "free_disk_space": true, "runs_on": "ubuntu-24.04" },
+            { "tag": "openvino", "dockerfile": ".devops/openvino.Dockerfile", "platforms": "linux/amd64", "full": true, "light": true, "server": true, "free_disk_space": false, "runs_on": "ubuntu-24.04" }
+          ]
+          JSON
```
taronaeo, line 87: Isn't this easier to manage if we created a dedicated file for it in the repository instead of managing it via a bash command in a YAML file?

ehfd:
https://github.com/ehfd/llama.cpp/actions/runs/23731725922

@CISC @taronaeo Had to fix the Vulkan image because Ubuntu 26.04 Beta removed Python 3.13 overnight (replaced with `uv` to respect https://github.com/ggml-org/llama.cpp/pull/20530). The Ubuntu packages page removed Python 3.13 for 26.04 between yesterday and today, and the minimum is 3.14.

Edit: Vulkan now passes.

CISC:
> @CISC @taronaeo Had to fix the Vulkan image because Ubuntu 26.04 Beta removed Python 3.13 overnight (replaced with `uv` to respect #20530). The Ubuntu packages page removed Python 3.13 for 26.04 between yesterday and today, and the minimum is 3.14.

This image is becoming annoying. :)

taronaeo (commented):
I'm not familiar with the multi-architecture, multi-platform setup that you've proposed here and will need to read the Medium article linked before I can give an approval.

Same as @CISC's comment [here](https://github.com/ggml-org/llama.cpp/pull/21122#issuecomment-4149910174), I think this is very elaborate and IMO hard to maintain (especially the `jq` commands).

If another maintainer understands this well enough, they can give the approval in the meantime :)

.github/workflows/docker.yml
ehfd: Well, I thought that it would be harder for people to find and update it if it's in a separate file.

The purpose was to keep the behavior as similar as possible to before, where that list is in the YAML.

ehfd: All I could say is that the `jq` behavior in manifest is the standard way when we have to run the same image tag in different runners.

Sticking to multi-platform build in the x86_64 runner will be very problematic and likely cross the 6-hour threshold and fail at worst due to emulator overhead.

ehfd:
Well, the CI/CD passed.

taronaeo:
I'm reading this blog from Docker themselves: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

It appears that the solution proposed here is slightly outdated and as I thought, [harder than it actually should be](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/#:~:text=We%E2%80%99ll%20start%20by%20doing%20it%20the%20hard%20way%20with%20%60docker%20manifest%60%20because%20it%E2%80%99s%20the%20oldest%20tool%20made%20by%20Docker%20to%20build%20multiarch%20images).

Can't we just use `docker buildx` to simplify this process? Am I missing something? https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/#:~:text=The%20simple%20way%20with%20docker%20buildx

Edit: More documentation: https://docs.docker.com/build/building/multi-platform/#build-multi-platform-images

ehfd:
@taronaeo The simple way doesn't work because we are running in GitHub Actions and pushing to GitHub Registry, and such a method would easily lead to 6-hour build times (which timeout) vs 1-hour due to emulating arm64 and s390x within x86_64 instead of using different native runners.

There is one other way not in that article to use that syntax you proposed, and at the same time run in native build runners, only when we have control of the configurations for build nodes of each architecture, but not in GitHub Actions, because we have no such control of dispatching builds to various remote runners.

In such a situation, using `docker manifest create` and the more modern equivalent `docker buildx imagetools create` is standard.

ehfd:
A more to-the-point AI explanation:

The short answer is: No, you cannot specify multiple platforms in a single build step and expect them all to run natively on different runners.

When you set platforms: "linux/amd64,linux/arm64" in a single docker/build-push-action step, that specific runner tries to build both. Since a runner only has one native architecture, it will use QEMU to emulate the other one, which is significantly slower.

However, your current workflow is actually already using the "industry standard" best practice for achieving what you want. Here is the breakdown of why your current approach is the right one and how to think about the platforms field.

Why your current "Matrix + Merge" is better
Your workflow uses a distributed build pattern. This is the most efficient way to handle multi-arch builds in GitHub Actions.

Native Speed: By splitting the matrix so linux/amd64 runs on ubuntu-24.04 and linux/arm64 runs on ubuntu-24.04-arm, every instruction in your Dockerfile runs at native hardware speed.

Avoids QEMU: Emulating ARM on AMD64 (or vice versa) can be 10x–20x slower, especially for compiled languages (C++, Rust, Go).

The Manifest Merge: The merge_arch_tags job using imagetools create is the "glue." It doesn't rebuild anything; it just points a single tag (like my-image:latest) to the different architecture-specific digests you already pushed.

What happens if you change "platforms" to both?
If you modified your JSON to look like this:
{ "platforms": "linux/amd64,linux/arm64", "runs_on": "ubuntu-24.04" }

The ubuntu-24.04 (AMD64) runner would start.

It would build the AMD64 version natively.

It would see it also needs an ARM64 version and start a QEMU emulator.

The ARM64 build would take much longer and likely time out or exhaust resources.

The "Native" Alternative: Remote Build Nodes
If you truly want to use a single platforms: "linux/amd64,linux/arm64" line in one job but have it run natively, you would need to set up a remote Buildx driver.

This involves:

Starting an ARM64 runner and an AMD64 runner.

Connecting them via SSH.

Configuring Buildx to use both as "nodes" in a single builder instance.

Why you shouldn't do this: It is significantly harder to maintain in GitHub Actions because of networking/SSH complexities between ephemeral runners. Your current "Push by Digest & Merge" strategy is much more robust.

taronaeo:
Okay I am starting to get the full picture of this approach now, and can understand why. In this case, this PR is perfectly acceptable.

ehfd:
@taronaeo Okay, then! Thank you. Can't wait to get it merged.

taronaeo (approved):

ehfd:
@CISC Two approvals!

nalf3in:
Hi, this broke the llamacpp install (using a container) on my debian 13 server as the latest nvidia driver is 555.58.02-3 which only supports up to cuda 12.4 and it looks like this PR bring the minimum up to 12.8.1 (possibly 12.8 from what I understand from the bug mentioned earlier). 

Would it be possible to bring the minimum back to 12.4 or make it obvious to users that cuda 12.4 is no longer supported using containers ?

ehfd:
@nalf3in 

Here is where the currently supported CUDA version is located: https://github.com/ggml-org/llama.cpp/blob/master/docs/docker.md#building-docker-locally

Unfortunately, Debian is frequently problematic combined with NVIDIA GPUs because of not provide adequate driver versions required. After discussion with the maintainer of this code, we decided to retain CUDA 12.8 due to the support of Blackwell GPUs.

But, I suggest reading this to install the latest driver version available in Debian: https://docs.nvidia.com/datacenter/tesla/driver-installation-guide/debian.html

mediouni-m:
@nalf3in CUDA 12.9 binaries run under older CUDA 12.x releases if they do not go through the runtime PTX path but have prebuilt SASS binaries for the given architecture. But that might come with a binary size penalty...

nalf3in:
@ehfd Thank you for the clarifications. In my case I will stick with building my own containers but it's nice to know there's an alternative. This is a bit pedantic, but the CUDA version is only mentioned in the "Building Docker locally" section, not in the "Usage" part where the prebuilt image is discussed. Please let me know if you think a PR to address this would be a good idea.

@mediouni-m This is interesting. I personally wouldn’t mind a larger binary size for the sake of wider compatibility, but I know that’s not everyone’s view. What do you all think? Is that a good trade-off?

ehfd:
PR is always appreciated. @nalf3in And I do think SASS is a good tradeoff for especially llama.cpp.

PR #21124, opened 2026-03-28
fix: handle non-capturing groups (?:...) in JSON schema pattern converter

XciD:
## Summary

- Fix SIGSEGV in `_visit_pattern()` when a JSON schema `pattern` contains non-capturing groups `(?:...)`
- Skip `?:` after `(` to treat non-capturing groups as regular groups
- Safely handle unsupported lookahead/lookbehind syntax by skipping to matching `)`

## Root cause

In `common/json-schema-to-grammar.cpp`, `_visit_pattern()` line ~420: when the parser sees `(` followed by `?`, it pushes a warning but does not advance `i` past `?:`. The recursive `transform()` call then interprets `?` as a quantifier and calls `seq.back()` on an empty vector (undefined behavior, SIGSEGV).

## Reproducer

Start `llama-server` with any model, then:

```bash
curl -X POST http://127.0.0.1:8080/v1/messages \
  -H "x-api-key: dummy" \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"any","max_tokens":50,"messages":[{"role":"user","content":"hi"}],
  "tools":[{"name":"t","description":"t","input_schema":{"type":"object",
  "properties":{"arr":{"type":"array","items":{"anyOf":[{"type":"object",
  "properties":{"d":{"type":"string","pattern":"^(?:foo)$"}},
  "required":["d"]}]}}}}}],"tool_choice":{"type":"auto"}}'
```

The server crashes with SIGSEGV before returning a response. With this fix, it returns HTTP 200 correctly.

This affects real-world clients (e.g. Claude Code with Notion MCP tools) that send tool schemas containing date validation patterns like `^(?:(?:\d\d[2468][048]|...)-02-29|...)$`.

### DIFF
--- a/common/json-schema-to-grammar.cpp
+++ b/common/json-schema-to-grammar.cpp
@@ -416,15 +416,30 @@ private:
                     i++;
                 } else if (c == '(') {
                     i++;
-                    if (i < length) {
-                        if (sub_pattern[i] == '?') {
+                    if (i < length && sub_pattern[i] == '?') {
+                        if (i + 1 < length && sub_pattern[i + 1] == ':') {
+                            i += 2; // skip "?:" for non-capturing group, treat as regular group
+                        } else {
+                            // lookahead/lookbehind (?=, ?!, ?<=, ?<!) - not supported
                             _warnings.push_back("Unsupported pattern syntax");
+                            // skip to matching ')' to avoid UB on empty seq
+                            int depth = 1;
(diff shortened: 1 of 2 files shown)


aldehir:
Can you add a test to `tests/test-json-schema-to-grammar.cpp`?

pwilkin (approved):

aldehir (approved):

PR #21147, opened 2026-03-29
ggml-webgpu: Add the support of `MUL_MAT_ID`

yomaytk:
## Overview

This PR adds the support of `MUL_MAT_ID`. This op is used by general MoE model such as DeepSeek-V2/V3 and Mixtral.
- The computation logic is based on the fast path of `MUL_MAT` (without subgroup matrix for now), and the supported types are same as the fast path of `MUL_MAT`. I confirmed that the test-backend-ops passes for all supported types.

# Requirements

- AI usage disclosure:  YES - I used AI to help fix a bug in the shader code.


ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id.wgsl
```diff
@@ -0,0 +1,214 @@
...
+const TOTAL_WORKGROUP_SIZE = WORKGROUP_SIZE_M * WORKGROUP_SIZE_N;
+const TILE_SRC0_SHMEM = TILE_K * WORKGROUP_SIZE_M * TILE_M;
+const TILE_SRC1_SHMEM = TILE_K * WORKGROUP_SIZE_N * TILE_N;
+const MAX_EXPERT = 512;
+
+var<workgroup> shmem: array<f16, TILE_SRC0_SHMEM + TILE_SRC1_SHMEM>;
+var<workgroup> gathered_count_ids: array<u32, MAX_EXPERT>;
+
+#ifdef INIT_SRC1_SHMEM_FLOAT
+fn init_shmem_id_src1(thread_id: u32, offset_src1: u32, rest_token_n: u32, gathered_expert_used: array<u32, WORKGROUP_SIZE_N * TILE_N>, gathered_tokens: array<u32, WORKGROUP_SIZE_N * TILE_N>, k_outer: u32) {
```
yomaytk: The `src1` tile is selected based on `gathered_expert_used` and `gathered_tokens` (the array of expert_used or token for every expert).

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id.wgsl
```diff
@@ -0,0 +1,214 @@
...
+@compute @workgroup_size(TOTAL_WORKGROUP_SIZE)
+fn main(@builtin(workgroup_id) wg_id: vec3<u32>,
+        @builtin(local_invocation_id) local_id: vec3<u32>,
+        @builtin(num_workgroups) num_wg: vec3<u32>) {
+
+    let thread_id = local_id.x;
+    let local_m = get_local_m(thread_id);
+    let local_n = get_local_n(thread_id);
+
+    if (thread_id == 0u) {
```
reeselevine: if I understand correctly, this block is doing the gather for the full token batch size. However, there are a couple issues:

* Every workgroup does the exact same calculation (and in fact there are data races here since thread 0 from each workgroup writes out to the global buffers). While this looks like it works in practice, it's technically unsound and redundant. Instead, can we use a multi-kernel approach here? Unfortunately, there is no "global" synchronization within a WGSL shader so we can't rely on having only one workgroup do this calculation.
* The gather isn't parallelized at all. I don't know how much this really matters for performance, since this loop seems mostly pretty small. Maybe a first step would be to just extract it out as is into its own kernel, and then see if parallelizing makes a difference.

Let me know if this makes sense or I'm misunderstanding something.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id.wgsl
```diff
@@ -0,0 +1,214 @@
...
+    return thread_id / WORKGROUP_SIZE_M;
+}
+fn get_local_m(thread_id: u32) -> u32 {
+    return thread_id % WORKGROUP_SIZE_M;
+}
+
+const TOTAL_WORKGROUP_SIZE = WORKGROUP_SIZE_M * WORKGROUP_SIZE_N;
+const TILE_SRC0_SHMEM = TILE_K * WORKGROUP_SIZE_M * TILE_M;
+const TILE_SRC1_SHMEM = TILE_K * WORKGROUP_SIZE_N * TILE_N;
+const MAX_EXPERT = 512;
```
reeselevine: can this be specialized when compiling based on the `src0` tensor dimensions to avoid pessimistically over-provisioning here? Or does it change often enough that we wouldn't want different shaders?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2812,6 +2909,19 @@ static size_t ggml_backend_webgpu_buffer_type_get_alloc_size(ggml_backend_buffer
                 }
             }
             break;
+        case GGML_OP_MUL_MAT_ID:
+            {
+                const ggml_tensor * src0 = tensor->src[0];
+                const ggml_tensor * src1 = tensor->src[1];
+                if (src0 && src1) {
+                    const size_t gatherd_size = sizeof(uint32_t) * tensor->src[0]->ne[2] * tensor->src[1]->ne[2];
```
reeselevine: minor typo: `gathered`

reeselevine (commented):
Thanks for the implementation! A few comments, which I think could help with performance. For some numbers, on my M3 I see 200 GFLOPS for `MUL_MAT_ID(type_a=f16,type_b=f32,n_mats=128,n_used=8,b=0,m=768,n=512,k=2048)`, whereas the Metal implemntation is almost 2 TFLOPS.

The 200 GFLOPS doesn't change when I run the `f32` perf variant, which makes me believe this current implementation is bandwidth-bound instead of compute-bound like it should be. Hopefully this is just due to the overhead of the gather stage and removing redundancy/parallelizing closes some of that gap.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id.wgsl
yomaytk: You're right, and I agree that using a multi-kernel approach makes sense.

This loop iterates over `ids` and gathers `(token, expert_used)` for each expert. Currently, thread 0 from every workgroup performs this calculation. Since all threads process the same `ids` in the same order, it should work correctly.

However, as you pointed out, having multiple threads write to the same buffer at the same time is not good, so the multi-kernel approach looks better. And gathering should be parallelized. I will update it accordingly.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id.wgsl
yomaytk: I had to allocate `gathered_count_ids` as a workgroup buffer to avoid data races between workgroups, so I used a fixed size (`MAX_EXPERT`). However, by using a multi-kernel approach, we can store `gathered_count_ids` directly in the storage buffer and set the array size to `src0->ne[2]`.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id.wgsl
```diff
@@ -0,0 +1,193 @@
...
+#ifdef INIT_SRC1_SHMEM_FLOAT
+fn init_shmem_id_src1(thread_id: u32, offset_src1: u32, rest_token_n: u32, k_outer: u32) {
+    for (var elem_idx = thread_id * VEC_SIZE; elem_idx < TILE_SRC1_SHMEM; elem_idx += TOTAL_WORKGROUP_SIZE * VEC_SIZE) {
+        let tile_n = elem_idx / TILE_K;
+        let tile_k = elem_idx % TILE_K;
+        if (tile_n < rest_token_n) {
+            let global_src10 = k_outer + tile_k;
+            let expert_used_idx = gathered_expert_used[tile_n] % params.b_ne1;
+            let token_idx = gathered_tokens[tile_n];
+            let src1_idx = offset_src1 + token_idx * params.stride_12 + expert_used_idx * params.stride_11 + global_src10;
```
yomaytk, line 69: Each element is accessed only once, but the access pattern is scattered in physical memory compared to `MUL_MAT`, because `ne1` and `ne2` are determined by gathered `(expert_used, token)` for each expert.
So it might be inefficient in terms of cache, but for small `n_tokens`, the overhead is still quite large.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id.wgsl
```diff
@@ -0,0 +1,193 @@
...
+    let local_m = get_local_m(thread_id);
+    let local_n = get_local_n(thread_id);
+
+    var expert_idx:u32 = 0xFFFFFFFFu;
+    var wg_in_batch:u32 = 0;
+    var wg_sum:u32 = 0;
+    let wg_m_count = (params.m + WORKGROUP_SIZE_M * TILE_M - 1u) / (WORKGROUP_SIZE_M * TILE_M);
+    let wg_linear = wg_id.y * num_wg.x + wg_id.x;
+
+    for (var i = 0u;i < params.n_expert;i += 1) {
```
yomaytk, line 97: This loop is not parallelized, but since `params.n_expert` is at most 128 in the benchmark, it should not affect the performance much.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id_gather.wgsl
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        workgroupBarrier();
+
+        for (var i = thread_id;i < params.n_expert_used * params.n_tokens;i += WG_SIZE) {
+            let row = i / params.n_expert_used;
+            let col = i % params.n_expert_used;
+            let expert = u32(ids[params.offset_ids + row * params.stride_ids_1 + col]);
+            if (own_expert == expert) {
+                let pos = atomicAdd(&gathered_count_ids[own_expert], 1u);
```
reeselevine: not sure how much of the overall total time is due to this kernel but this could be optimized by using a workgroup level atomic variable, then having one thread in the workgroup store back to the `gathered_count_ids` at the very end I think.

reeselevine (commented):
Thanks for the updates! I agree there's probably more room for performance, probably using subgroup matrices would help a lot on native, but won't work in the browser yet.

At this point it's up to you whether to merge it now or keep iterating on performance. I think it's in reasonable shape, and having the functionality is great. Maybe more performance optimization can be done in later PRs when it becomes higher priority?

As far as merging, there are some merge conflicts, but maybe wait until https://github.com/ggml-org/llama.cpp/pull/21278 is merged to fix them since that will also cause some more conflicts.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id_gather.wgsl
yomaytk: It looks better so I’ve updated it accordingly, but the performace is almost the same.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -1598,6 +1621,115 @@ class ggml_webgpu_shader_lib {
...
+                break;
+            default:
+                {
+                    std::string type_upper = src0_name;
+                    std::transform(type_upper.begin(), type_upper.end(), type_upper.begin(), ::toupper);
+
+                    defines.push_back("BYTE_HELPERS");
+                    defines.push_back("INIT_SRC0_SHMEM_" + type_upper);
+                    defines.push_back("INIT_SRC1_SHMEM_FLOAT");
+                    defines.push_back("U32_DEQUANT_HELPERS");
```
yomaytk, line 1696: I added this macro based on the recent fix(https://github.com/ggml-org/llama.cpp/pull/21046).

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_id.wgsl
```diff
@@ -0,0 +1,193 @@
...
+    var rest_token_n: u32 = 0;
+    var src0_batch_offset: u32 = 0;
+
+    wg_m = wg_in_batch % wg_m_count;
+    wg_n = wg_in_batch / wg_m_count;
+
+    offset_wg_m = wg_m * WORKGROUP_SIZE_M * TILE_M;
+    offset_wg_n = wg_n * WORKGROUP_SIZE_N * TILE_N;
+
+    if (is_valid) {
```
reeselevine:
```suggestion
    if (is_valid) {
```

reeselevine (approved):

CISC (approved):

PR #21160, opened 2026-03-29
Cross-backend profiler

pwilkin:
## Overview

Moved to main repo, can discuss here :)

Previous discussion: #21138


pwilkin:
@jeffbolznv I'll be sure to add the pure text dump, as for all the other stuff it should already be there, I explicitly kept concurrent mode as an option for Vulkan.

jeffbolznv:
For the concurrent mode it looked like it was splitting the time between the nodes, but it should just report them as a combined operation.

I doubt the command line parameter just works everywhere, for example sd.cpp has separate command line arg parsing. IMO it should be possible to enable it with an env var.

ngxson:
A bit late to the party from https://github.com/ggml-org/llama.cpp/pull/21138 , but my 2c is that this can be beneficial in 2 ways:
1. Providing a better DX when working with multiple backends, since now developer no longer have to re-learn how to debug perf for each backend they are dealing with
2. Having a systematic way to compare perf across different backends

---

I do hear the concern from @JohannesGaessler that at least for CUDA, this can be difficult to maintain. However, at least for my use case of adding support for new models (text / multimodal) where I often need to evaluate the perf in a more "end-to-end" setup, I can imagine this could help saving quite a lot of time. Exporting a graph then manually debugging it could work, but for example with qwen-next, the graph changes depends on batch size, so it could be a bit tedious if I have to go for an extra step of extracting the graph then debug it.

Therefore, in an attempt to mediate this matter, I would like to ask: What do you see as the main maintenance burden(s) of the current implementation?

JohannesGaessler:
The maintenance burden is that with this PR Piotr would be injecting large amounts of machine-generated code into many different areas of the codebase that he is (to my knowledge) not familiar with. I think this is unacceptable, even for a maintainer. The ggml backend code in particular is a critical component and in my experience the one where debugging is the most time-consuming.

As I said in the previous thread, this should be implemented as an external tool with minimal changes to the ggml backend code. The backend scheduler already has code for per-node callbacks. I think this could already be used for profiling ggml ops without any ggml backend changes: for `ask == true`, simply request every node and record the start time, for `ask = false` record the time again and calculate the difference. The only thing you wouldn't be easily getting is proper representation of tensor fusion (since you would want to execute multiple nodes at once for that), I don't know whether it's currently already possible to retrieve this.

pwilkin:
@JohannesGaessler I was first considering doing this with `cb_eval` since that would be backend-agnostic, but there are two problems: (1) from my knowledge `cb_eval` forces synchronous processing, so it basically disables any sort of optimizations and (2) I'm not able to record data transfers if using just `cb_eval`.

I'm not totally unaware of the code :) I've made I think 5 attempts at this so far trying to understand all the caveats and finally ending up with something that does what I think is relatively minimal changes to the API - as in, I believe the profiler changes are completely self-contained and do not influence any execution paths if the profiler is disabled. But as I said above - I'm absolutely fine with keeping this as a separate branch and not merging it in.

JohannesGaessler:
>The only thing you wouldn't be easily getting is proper representation of tensor fusion (since you would want to execute multiple nodes at once for that), I don't know whether it's currently already possible to retrieve this.

Looking at `ggml-impl.h` there seems to already be code for that. So should be possible to implement a profiler in user code via `ggml_backend_sched_eval_callback` by getting the maximum fusable subgraph for `ask == true` and returning false a number of times equal to the number of nodes in the subgraph - 1, then record the time and return true once to initiate the graph execution, then record the time again for `ask = false`.

JohannesGaessler:
>(1) from my knowledge cb_eval forces synchronous processing, so it basically disables any sort of optimizations 

I agree that the measurements would be biased towards larger values but for CUDA this is already the case if you use eager execution rather than CUDA graphs. In my opinion this is not a dealbreaker as long as this is understood by the person looking at the data. 

At least for CUDA the overhead from launching a kernel synchronously should be roughly constant. It would then be possible to execute individual ops and to record the runtime with overhead and to then substract the overhead at the end. It should be possible to determine the overhead by launching 2 ops as 2 individual graphs and also as 1 graph and calculating the difference in runtime (the ops would need to be not inplace so that executing them twice is safe).

>(2) I'm not able to record data transfers if using just cb_eval.

I agree that this could right now not be captured properly and data transfers would just show up as empty spaces in the timeline. I think it would be doable to get the time for data transfers at the beginning and end of a split with relatively minimal changes which I think would be enough.

ngxson:
Actually I think the idea of re-proposing `cb_eval` makes more sense in the context of ggml / llama.cpp, it has certain advantages over adding a completely new profiling infrastructure:
1. Least code to be added to backend if I understand correctly
2. Make it more flexible for downstream code (i.e. where developer wants to add their custom handling logic)
3. Less state management inside ggml --> less bugs; stateful objects are always pain when talking about debugging

An idea could be a new `cb_eval_ext` that is more flexible, like:

```cpp
struct callback_response {
  int flag = GGML_RES_NONE; // a bit field, can be for example: GGML_RES_TIMING | GGML_RES_DATA
  // maybe other things here
};
struct callback_request {
  struct ggml_tensor * t;
  // maybe other things here
};
typedef callback_response (*ggml_backend_sched_eval_callback_ext)(struct callback_request * req, void * user_data);
```

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -14973,10 +15028,43 @@ static ggml_status ggml_backend_vk_graph_compute(ggml_backend_t backend, ggml_cg
...
+                    rec.start_ns   = cpu_ts;
+                    rec.end_ns     = cpu_ts + duration_ns;
+                    rec.bytes      = total_bytes;
+                    rec.extra      = names[0];  // fusion name of first op, or NULL
+                    ggml_profile_record_from_tensor(&rec, node);
+                    ctx->profiler_state->records.push_back(rec);
+                }
+            }
+        }
+    }
```
0cc4m: This doesn't look right.

pwilkin: Did my wonderful assistant miscount the braces again? :D

Eh, I have to do a Vulkan build each time I upload those.

0cc4m: You can build Vulkan and CUDA together. It will still prefer CUDA on devices that support both, but allows to easily do tests with Vulkan with `--device Vulkan0`. Quite useful.

fairydreaming:
@pwilkin Tried this today while playing with DeepSeek V4 PR. Overall seems useful, but I don't get it how I'm supposed to match the profiler records with GGML graph nodes. There's the following info in the output json:

```
    {"type": 0, "name": "MUL_MAT", "backend_id": 0, "split_id": 1, "start_ns": 42569833650557, "duration_ns": 9780384, "bytes": 192, "extra": null, "ne": [24, 2, 1, 1], "out_type": 0, "n_src": 2, "ne_src": [[16384, 24, 1, 1], [16384, 2, 1, 1]], "nb_src": [[4, 65536, 1572864, 1572864], [4, 65536, 131072, 131072]], "type_src": [0, 0], "op_params": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], "sub_op": -1},
```

I suppose I can do some wild guessing from the source tensor shapes and the overall graph flow, but maybe at least the tensor name should be there?

Also it looks like the PR needs some update as it no longer merges cleanly with the current master.

pwilkin:
@fairydreaming thanks for reminding me to do the occasional rebase on master ;)

Added tensor name to the profiler output.

fairydreaming:
@pwilkin 😭

```
$ git apply 21160.diff
error: patch failed: ggml/src/ggml-vulkan/ggml-vulkan.cpp:15826
error: ggml/src/ggml-vulkan/ggml-vulkan.cpp: patch does not apply
error: tests/export-graph-ops.cpp: No such file or directory
error: patch failed: tests/test-backend-ops.cpp:9813
error: tests/test-backend-ops.cpp: patch does not apply
```

Edit: needs only minor fixes

```
$ diff 21160.diff.1 21160.diff
3191c3191
<      ctx->last_total_mul_mat_bytes = total_mul_mat_bytes;
---
>      ctx->last_total_flops = total_flops;
3478c3478
< diff --git a/tests/export-graph-ops.cpp b/tests/export-graph-ops.cpp
---
> diff --git a/tests/test-export-graph-ops.cpp b/tests/test-export-graph-ops.cpp
3480,3481c3480,3481
< --- a/tests/export-graph-ops.cpp
< +++ b/tests/export-graph-ops.cpp
---
> --- a/tests/test-export-graph-ops.cpp
> +++ b/tests/test-export-graph-ops.cpp
3632,3633c3632,3633
< -    printf("    --test-file reads test operators from a test file generated by llama-export-graph-ops\n");
< +    printf("    --test-file reads test operators from a test file generated by llama-export-graph-ops or the profiler\n");
---
> -    printf("    --test-file reads test operators from a test file generated by test-export-graph-ops\n");
> +    printf("    --test-file reads test operators from a test file generated by test-export-graph-ops or the profiler\n");
```

pwilkin:
@fairydreaming rebased :)

ggml/src/ggml-backend.cpp
jeffbolznv: Tried using this today, it didn't dump because the constructor wasn't getting run.

```suggestion
    struct ggml_backend_sched * sched = new ggml_backend_sched{};
```

jeffbolznv:
```suggestion
    delete sched;
```

jeffbolznv (commented):
codex found these fixes, they worked in my testing

pwilkin:
@jeffbolznv thanks!

jeffbolznv:
FWIW, I tried to use this to measure where ggml-cuda spends its time, and it seemed like some fused ops aren't reported? I didn't attempt to debug.

pwilkin:
@jeffbolznv yes, I noticed as well there's some problem with fusions, I'll debug it.

PR #21163, opened 2026-03-29
non-compliant sycl mamba support

pramukta:
## Overview

Adds SYCL SSM_SCAN implementation to keep computation on the GPU.  This frees up my CPU while inference is running.


pramukta:
sorry about this, meant to be working on my fork as this does not comply with the project requirements.

PR #21149, opened 2026-03-29
Support for DeepseekV32ForCausalLM with DeepSeek Sparse Attention (DSA)

fairydreaming:
## Overview

This PR adds support for DeepseekV32ForCausalLM (DeepSeek V3.2 Exp, DeepSeek V3.2, DeepSeek V3.2 Speciale) models. It contains implementation of the lightning indexer and DeepSeek Sparse Attention (DSA) - both implemented in the simplest possible way as a proof of concept. So far only CPU and CUDA backends are supported.

~~Due to the way it's currently implemented it doesn't improve long context performance yet, more work is needed for this.~~ Long context performance was improved by using sparse top_k indices in CUDA flash attention MMA kernel.

Some GGUFs for testing are available [here (-light models)](https://huggingface.co/sszymczyk), I uploaded Q8_0/Q4_K_M quants, so you need over 700GB/400GB of RAM/VRAM to run them.

I also created a [16GB baby DeepSeek V3.2 GGUF](https://huggingface.co/sszymczyk/DeepSeek-V3.2-4Layers-GGUF) for VRAM-deprived people. It outputs incoherent gibberish, but should be useful for testing and optimizing this implementation even with limited resources.

**I really could use some help with verifying the implementation correctness. If you have large GPU cluster and can run some benchmarks to compare results with official reported benchmark results for DeepSeek V3.2 models then go for it.** More details in #21183.

Fixes #16331, #20363

## Additional information

Decisions I made when implementing this:
- new model arch `DEEPSEEK32` was added (mostly a copy of existing `GLM_DSA` arch),
- sparse attention was implemented by masking KQ mask entries corresponding to tokens that are not in the set of top-k tokens selected by the lightning indexer,
- ~~for this purpose I added new GGML op `GGML_OP_SCATTER` that works similar to torch scatter_ operation but is currently limited to setting tensor elements at specified indices to a given scalar value,~~ replaced by SET_ROWS with 1-element rows
- Hadamard transform ~~was added as another new GGML op `GGML_OP_HADAMARD` with implementation borrowed from [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp) (thx @ikawrakow),~~ implementation from #21038 was used in lightning indexer
[...]

PR #19726, opened 2026-02-19
Port IQ*_K quants from ik_llama.cpp

AesSedai:
This PR is an initial effort at porting @ikawrakow's IQ*_K quants from ik_llama.cpp to mainline llama.cpp. Attribution has been provided for the quantization code, and if additional attribution work is required please let me know.

This branch implements the CPU backend for the following quantization types:
- IQ2_K, IQ3_K, IQ4_K, IQ5_K, IQ6_K

Models quantized with these types from ik_llama.cpp should correctly load in this PR, and llama.cpp can now be used to produce those quantizations.

Since this implementation is for the CPU backend only currently, do not expect excellent performance. Further backends will be added in future PRs (CUDA, Vulkan, etc.).

I made a few small quality-of-life tweaks to `test-quantize-fns` and it passes for these new quantization types:
```
$ ./build/bin/test-quantize-fns
...
Testing iq2_k
Testing iq3_k
Testing iq4_k
Testing iq5_k
Testing iq6_k
```

I have also done some initial KLD testing by quantizing a model with ik_llama.cpp, collecting logits for it, then loading that model in llama.cpp and comparing its KLD:
```
# Qwen3-4B-Instruct-2507-IQ6_K.gguf
====== Perplexity statistics ======
Mean PPL(Q)                   :  10.910113 ±   1.495592
Mean PPL(base)                :  10.803911 ±   1.477219
Cor(ln(PPL(Q)), ln(PPL(base))):  99.79%
Mean ln(PPL(Q)/PPL(base))     :   0.009782 ±   0.008832
Mean PPL(Q)/PPL(base)         :   1.009830 ±   0.008919
Mean PPL(Q)-PPL(base)         :   0.106202 ±   0.097559

====== KL divergence statistics ======
Mean    KLD:   0.011825 ±   0.000975
Maximum KLD:   0.270169
99.9%   KLD:   0.243841
99.0%   KLD:   0.099079
95.0%   KLD:   0.032769
90.0%   KLD:   0.026773
Median  KLD:   0.005709
10.0%   KLD:   0.000064
 5.0%   KLD:   0.000014
 1.0%   KLD:   0.000000
 0.1%   KLD:  -0.000004
Minimum KLD:  -0.000004
```

Don't expect the PPL and KLD to match up 100% due to backend differences and general noise. However the results should be reasonable and within ballpark range.

[...]

PR #21165, opened 2026-03-29
contrib : clarify code origin guidelines

ddh0:
## Overview

This PR updates the contributing guidelines to add a new section. At the time of posting, the section reads as follows:

```md
## Code Origin Disclosure Policy

*Contributors* and *collaborators*, as defined above, must disclose the origin of their code, if they did not write it themselves, to avoid potential licensing conflicts or intellectual property right violations.

Contributions that can be shown to originate from one or more of the following sources, in whole or in part, will not be accepted:
- [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp)
- [Iwan Kawrakow](https://github.com/ikawrakow)

Note that maintainers reserve the right to reject code on the basis of its origin, even if not explicitly listed above.
```

Future readers are encouraged to view the diff of this PR, in case anything has changed. After this PR is merged, readers should instead refer to the [current content of CONTRIBUTING.md](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md), and should disregard this PR.

I also edited all Markdown headers except the first one from `#` to `##`, such that the document now has a proper structure with a heading. I consider the new section to be the only substantial change.

## Motivation

In https://github.com/ggml-org/llama.cpp/pull/19726 and https://github.com/ggml-org/llama.cpp/pull/21149 (and probably other instances), code was rejected the basis of its origin. In both cases, the contributors were not necessarily aware of the stance of llama.cpp with respect to what is allowed, and what is not.

I hope that this addition can clarify things for contributors, and reduce the amount of friction in the future. I welcome feedback from any maintainers, in particular @ggerganov and @JohannesGaessler. Thank you.

# Requirements

- AI usage disclosure: NO


fairydreaming:
> Contributions that can be shown to originate from one or more of the following sources, in whole or in part, will not be accepted:
> - [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp)

@ddh0 So let's say I'm implementing support for new model and it's merged in ik_llama.cpp first. Then I can't submit that PR to llama.cpp anymore?

ddh0:
> @ddh0 So let's say I'm implementing support for new model and it's merged in ik_llama.cpp first. Then I can't submit that PR to llama.cpp anymore?

I don't know - hence the PR and ping to core maintainers.

0cc4m:
> > Contributions that can be shown to originate from one or more of the following sources, in whole or in part, will not be accepted:
> > 
> > * [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp)
> 
> @ddh0 So let's say I'm implementing support for new model and it's merged in ik_llama.cpp first. Then I can't submit that PR to llama.cpp anymore?

I don't think that case is a problem at all, you authored both PRs. The question is what to do with ports of ik features to llama.cpp. Those have usually been denied. In your case the hadamard operator, I guess.

JohannesGaessler (approved):
Would be fine with me.

saood06:
> I don't think that case is a problem at all, you authored both PRs.

But this PR would make that a problem or at least a violation of this new policy.

I do agree with @fairydreaming in that this makes me feel put off and confused. Why lump the whole repo in if the whole point is to block contributions originating from a speicific person. There are still things I have considered implementing or fixing that could be useful to both repos (since not everything has diverged yet), but if this policy change becomes enacted than am I supposed to submit here first, and then only after it's either merged or rejected submit it over there? All for what?

I understand the benefit of explicitly having the rule written about not accepting code originating from that person because it won't be accepted (and thus would save anyone the time it takes to make PRs with it) even if I don't like anything about this situation, but also having the rule mention the forked repo doesn't seem to be based on precedence. It seems like a new rule, and one that seems like it punishes people who I feel like are not involved in this conflict.

ddh0:
> > But this PR would make that a problem or at least a violation of this new policy.
> 
> I do agree with @fairydreaming in that this makes me feel put off and confused. Why lump the whole repo in if the whole point is to block contributions originating from a speicific person. There are still things I have considered implementing or fixing that could be useful to both repos (since not everything has diverged yet), but if this policy change becomes enacted than am I supposed to submit here first, and then only after it's either merged or rejected submit it over there? All for what?
> 
> I understand the benefit of explicitly having the rule written about not accepting code originating from that person because it won't be accepted (and thus would save anyone the time it takes to make PRs with it) even if I don't like anything about this situation, but also having the rule mention the forked repo doesn't seem to be based on precedence. It seems like a new rule, and one that seems like it punishes people who I feel like are not involved in this conflict.

My intention with this PR is just to clarify the current stance of the project. I would appreciate some input from @ggerganov concerning the wording / phrasing. I certainly don't want to introduce any _more_ friction for contributors in the future.

ddh0:
Given @JohannesGaessler's previous comment [here](https://github.com/ggml-org/llama.cpp/pull/19726#issuecomment-3946355613) (from #19726), and assuming that that comment still aligns with GGML's stance, I am still interested in merging this PR to clarify the position of the llama.cpp project.

@ggerganov what do you think about this?

saood06:
> Given @JohannesGaessler's previous comment [here](https://github.com/ggml-org/llama.cpp/pull/19726#issuecomment-3946355613) (from #19726), and assuming that that comment still aligns with GGML's stance, I am still interested in merging this PR to clarify the position of the llama.cpp project.

That comment is very clearly about a specific person. I still don't understand why [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp) needs to be mentioned as that would affect other people under this rule. Like I said before, adding the reference to [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp) is not a clarification, it is an additional restriction which affects people who are not the targeted individual who contribute to that repo.

ddh0:
> > Given @JohannesGaessler's previous comment [here](https://github.com/ggml-org/llama.cpp/pull/19726#issuecomment-3946355613) (from #19726), and assuming that that comment still aligns with GGML's stance, I am still interested in merging this PR to clarify the position of the llama.cpp project.
> 
> That comment is very clearly about a specific person. I still don't understand why [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp) needs to be mentioned as that would affect other people under this rule. Like I said before, adding the reference to [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp) is not a clarification, it is an additional restriction which affects people who are not the targeted individual who contribute to that repo.

Your point is fair. I would be absolutely fine with reducing the scope of the wording, but it makes no sense for me to be the one that makes this change - this PR is stuck waiting for input from ggerganov or another core maintainer.

Issue #20733, opened 2026-03-18
Eval bug: Unable to generate parser for this template, System message must be at the beginnin


oleg-kn:
### Name and Version

version: 7803 (b70d25107)

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

2x RTX 3090

### Models

unsloth/Qwen3.5-27B-GGUF:Q8_K_XL

### Problem description & steps to reproduce

using Codex v0.94.0 i receive such message:

```
{"error":{"code":400,"message":"Unable to generate parser for this template. Automatic parser generation failed: \n------------\nWhile executing CallExpression at line 85, column 32
in source:\n...first %}↵            {{- raise_exception('System message must be at the beginnin...\n                                           ^\nError: Jinja Exception: System message
must be at the beginning.","type":"invalid_request_error"}}
```

### Relevant log output

srv    operator(): got exception: {"error":{"code":400,"message":"Unable to generate parser for this template. Automatic parser generation failed: \n------------\nWhile executing CallExpression at line 85, column 32 in source:\n...first %}↵            {{- raise_exception('System message must be at the beginnin...\n                                           ^\nError: Jinja Exception: System message must be at the beginning.","type":"invalid_request_error"}}
srv  log_server_r: done request: POST /chat/completions

Vivida1:
Same issue here using latest llama.cpp:server-cuda image.

```
llama-server -hf unsloth/Qwen3.5-35B-A3B-GGUF:Q8_0 --port 8080 --host 0.0.0.0 -c 0 --temp 0.6 --top-p 0.95 --top-k 20 --min-p 0.0 --chat-template-kwargs '{"enable_thinking":true}'
```

org-tekeli-borisp:
Same issue here with latest llama.cpp (48cda24c1175363dd17925102dbe1da49279940e) compiled with 
```cmake -B build -DGGML_CUDA=ON && cmake --build build --config Release``` on Arch Linux.

```
llama-cli --version
ggml_cuda_init: found 3 CUDA devices (Total VRAM: 44738 MiB):
  Device 0: Tesla T4, compute capability 7.5, VMM: yes, VRAM: 14912 MiB
  Device 1: Tesla T4, compute capability 7.5, VMM: yes, VRAM: 14912 MiB
  Device 2: Tesla T4, compute capability 7.5, VMM: yes, VRAM: 14912 MiB
version: 8555 (48cda24c1)
[...]

Issue #20607, opened 2026-03-15
/v1/responses streaming SSE missing fields required by Vercel AI SDK (output_index, id on function_call items, created_at/model)


MidnightOnMars:
## Summary

The `/v1/responses` streaming SSE implementation (from PR #18486) is missing three fields that the Vercel AI SDK's `@ai-sdk/openai` package requires via Zod schema validation. This prevents tools like OpenCode, Cursor, and other AI SDK-based clients from using the Responses API with llama-server for tool calling.

Chat Completions (`/v1/chat/completions`) works correctly. The model produces perfect structured tool calls. This is specifically about the Responses API streaming event format.

### 1. `output_index` (number) on multiple event types

**Affected events:** `response.output_item.added`, `response.output_item.done`, `response.function_call_arguments.delta`, `response.output_text.delta`

The AI SDK uses `output_index` to track which output item each event belongs to. Without it, the Zod schema validation fails (`chunk.success = false`) and the event is silently dropped.

**Fix:** Add a sequential counter starting at 0, incremented each time a new output item is added.

### 2. `id` (string) on function_call items

**Affected events:** `response.output_item.added` and `response.output_item.done` where `item.type === "function_call"`

llama-server sends `call_id` but not `id` on function_call items. The AI SDK schema requires both fields. When `id` is missing, the Zod discriminated union matches `type === "function_call"` but fails inner validation. The stream handler skips the event, `hasFunctionCall` never becomes true, and `finishReason` is always `"stop"` regardless of the model's actual output.

**Fix:** Copy `call_id` to `id` (or generate a unique `id`) on function_call items.

### 3. `created_at` (number) and `model` (string) on `response.created`

The AI SDK schema requires `response.{id, created_at, model}` but llama-server sends only `response.{id, status}`.

**Fix:** Add `created_at` (unix timestamp) and `model` (the model name from the request) to the response object in `response.created` events.

## What does NOT need fixing

The AI SDK has a catch-all that transforms unknown event types into `{ type: "unknown_chunk" }`. These events are silently ignored and do NOT cause errors:
- `response.reasoning_text.delta` (vs AI SDK's expected `response.reasoning_summary_text.delta`)
- `response.in_progress`
- `response.content_part.added`/`done`
- `response.output_text.done`

## Evidence

[...]

Issue #20156, opened 2026-03-06
Eval bug: Codex, 'type' of tool must be 'function'


chigkim:
### Name and Version

version: 8214 (f7db3f378)
built with AppleClang 17.0.0.17000603 for Darwin arm64

### Operating systems

Mac

### GGML backends

Metal

### Hardware

M3Max

### Models

ggml-org/gpt-oss-20b-GGUF

### Problem description & steps to reproduce

Run llama-server:
./llama-server \
-m gpt-oss-20b-mxfp4.gguf \
-c 64000 \
-np 2 \
-b 4096 \
-ub 4096 \
--jinja \
-fa on \
--host 0.0.0.0%                                                                                                                                                                                                  cgk@CK-MBP llama.cpp %
Configure Codex to work with llama.cpp.
Then run codex and give a prompt.

### First Bad Commit

If I use Codex>=0.88.0 with llama.cpp, I get the error:
`{"error":{"code":400,"message":"'type' of tool must be 'function'","type":"invalid_request_error"}}`
Codex 0.87.0 works fine.
Ollama works fine with latest Codex.

### Relevant log output

<summary>Logs</summary>

```console
`{"error":{"code":400,"message":"'type' of tool must be 'function'","type":"invalid_request_error"}}`<details>
```
</details>

stevenmoto:
Same error as @chigkim with running llamacpp and gpt-oss-120b with latest 0.114.0 codex

mhitza:
codex is finicky around the configuration format. While also being a poorly documented configuration format, I've hit the same error message before due to the configuration file. I fixed it, but I didn't take notes on what I was doing wrong.

Here's my current configuration that works.

```toml
model_reasoning_effort = "high"
web_search = "disabled"

[model_providers.local]
name = "llama.cpp"
base_url = "http://[::1]:6001/v1"
wire_api = "responses"

[profiles.20b]
model_provider = "local"
model = "gpt-oss:20b"
```

`codex --profile 20b`

This still shows a warning in the console "⚠ Model metadata for `gpt-oss:20b` not found. Defaulting to fallback metadata; this can degrade performance and cause issues.". But things seem to work generally (I only use it for small snippet generation and code review).

chigkim:
@mhitza, what version of llama.cpp and codex are you running?
It used to work fine in previous version, but no longer works in latest llama.cpp.

mhitza:
@chigkim 

[...]

PR #19720, opened 2026-02-19
server: add OpenAI Responses API compliance

riskywindow:
Fix Response object schema, streaming event schema, and multi-turn conversation input parsing for /v1/responses endpoint.

- Add 24 missing fields to Response object (tools, truncation, temperature, etc.)
- Add sequence_number, output_index, content_index, item_id, logprobs to all streaming events
- Add annotations and logprobs to content part objects
- Fix completed_at and usage fields on response.created/in_progress events
- Fix function_call output item structure (id vs call_id)
- Fix multi-turn input parsing to handle both OutputMessage and AssistantMessageItemParam

Passes 5/6 compliance tests (Image Input requires mmproj, not a code issue).

Fixes #19138

*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

PR #21174, opened 2026-03-30
server: improve Responses API compliance and Codex CLI compatibility

### LINKED ISSUE #19138: Feature Request: Support OpenAI Responses API (/v1/responses) in llama.cpp server
### Prerequisites

### Feature Description

Description:
The llama.cpp OpenAI-compatible server currently supports endpoints like /v1/chat/completions, but does not support the newer OpenAI Responses API (/v1/responses).

Many OpenAI client SDKs and tools are moving toward the unified Responses API, which supports structured outputs, tool calls, and multimodal responses in a single endpoint. Lack of /v1/responses support makes it harder to use llama.cpp as a drop-in replacement for OpenAI backends without an additional proxy or translation layer.

Requested feature:
Add native support for the /v1/responses endpoint in llama-server, aligned as closely as possible with OpenAI’s Responses API request/response format.

Benefits:

Improved compatibility with modern OpenAI SDKs

Easier migration from OpenAI APIs to llama.cpp

Reduced need for custom proxy or request rewriting layers

Thanks for the great work on llama.cpp!

### Motivation

I want to use ClaudeCode with llama-server offline, but it currently don't support necessary endpoints.

### Possible Implementation

_No response_

krystophny:
## Summary

Improve Responses API (`/v1/responses`) compliance and Codex CLI compatibility.

**Response object (non-streaming and streaming):**
- Add 24 missing fields per OpenAI Responses API spec (tools, truncation, temperature, top_p, metadata, store, service_tier, etc.)
- Fix function_call `id`/`call_id` field mapping (`id` gets unique fc_ ID, `call_id` gets the model's tool_call.id)
- Add `output_text` top-level convenience field
- Add `output_tokens_details` with `reasoning_tokens` to usage
- Restore `refusal` content type handling (was broken in upstream, unreachable code after throw)

**Streaming events:**
- Add `sequence_number` to ALL streaming events (created, in_progress, added, delta, done, completed)
- Add `output_index` to all events referencing output items
- Add `content_index` to content-related events
- Populate full response object in `response.created` and `response.in_progress` (was only `{id, object, status}`)
- Function call item IDs consistent across `output_item.added`, `function_call_arguments.delta`, and `output_item.done`
- Counter state persisted across streaming chunks via `task_result_state`

**Codex CLI compatibility:**
- Skip non-function tool types (web_search, code_interpreter) instead of rejecting with 400
- Merge developer/system role messages into first system message for templates requiring system at position 0 (e.g. Qwen)
- Strip Responses-only request keys (store, include, prompt_cache_key, web_search, text, truncation, metadata)
- Accept `input_text` type alongside `output_text` for EasyInputMessage / AssistantMessageItemParam

Prior art: #19720 by @riskywindow (stale, 500+ commits behind). This PR incorporates applicable ideas adapted to the current codebase.

Fixes #19138. Related: #20156, #20733, #20607.

### Post-merge local verification (2026-06-18)

Focused failures fixed during the merge:

```text
test-llama-archs: ggml/src/ggml-backend-meta.cpp:590: split_states_equal(src_ss[0], src_ss[1])
test-thread-safety: CUDA error: unspecified launch failure during CUDA stream teardown
```

Passing checks after the fixes:

```text
$ cmake -S . -B build -G Ninja -DLLAMA_BUILD_TESTS=ON && cmake --build build -j$(nproc)
[6/6] Linking CXX executable bin/llama
```

```text
$ ctest --test-dir build --output-on-failure -R '^test-llama-archs$'
100% tests passed, 0 tests failed out of 1
```

```text
$ ctest --test-dir build --output-on-failure -R '^test-thread-safety$'
100% tests passed, 0 tests failed out of 2
```

```text
$ ctest --test-dir build --output-on-failure -E '^(test-backend-ops|test-quant-type-selection)$'
100% tests passed, 0 tests failed out of 53
Total Test time (real) = 151.27 sec
```

```text
$ cd tools/server/tests
$ PORT=18080 AIP_HTTP_PORT=18080 LLAMA_SERVER_BIN_PATH=../../../build/bin/llama-server LLAMA_CACHE=$HOME/.cache/llama.cpp uv run --with-requirements requirements.txt ./tests.sh
314 passed, 3 skipped, 199 deselected in 818.91s (0:13:38)
```

```text
$ cd tools/ui
$ npm run test:ui -- --run --testTimeout=60000
Test Files 11 passed (11)
Tests 35 passed (35)
```

```text
$ cd tools/ui
$ npm run test:client -- --run
Test Files 2 passed (2)
Tests 9 passed (9)
```

```text
$ cd tools/ui
$ npm run test:unit -- --run
Test Files 15 passed (15)
Tests 211 passed (211)
```

```text
$ cd tools/ui
$ npm run test:e2e
6 passed (1.4m)
```

Local all-CTest without exclusions was not the final gate: `test-backend-ops` is a long backend matrix, and `test-quant-type-selection` failed because live model metadata no longer matched the upstream snapshots. The quant snapshot files are unchanged against `upstream/master`.

### pytest (14 tests, tinyllama2)

```text
$ LLAMA_SERVER_BIN_PATH=./build/bin/llama-server pytest unit/test_compat_oai_responses.py -v
test_responses_with_openai_library PASSED
test_responses_stream_with_openai_library PASSED
test_responses_schema_fields PASSED
test_responses_stream_schema_fields PASSED
test_responses_non_function_tool_skipped PASSED
test_responses_only_non_function_tools_same_as_no_tools PASSED
test_responses_extra_keys_stripped PASSED
test_responses_developer_role_merging PASSED
test_responses_input_text_type_multi_turn PASSED
test_responses_output_text_matches_content PASSED
test_responses_stream_output_text_consistency PASSED
test_responses_stream_created_event_has_full_response PASSED
test_responses_stream_all_events_have_sequence_number PASSED
test_responses_stream_delta_events_have_indices PASSED
14 passed
```

### E2E with async OpenAI SDK + Qwen3.5-9B Q4_K_M

```text
$ python3 e2e_test.py  # uses AsyncOpenAI against localhost:8080
Test 1: Non-streaming basic          OK: output_text='4', fields=36
Test 2: Streaming basic              OK: 205 events, gathered text matches
Test 3: Non-function tools skipped   OK: status=completed
Test 4: Developer role merging       OK
Test 5: Multi-turn with input_text   OK
Test 7: Streaming seq_nums           OK: 105 events, strictly increasing
Test 6: Concurrent stress (5 req)    OK: 5/5 completed, 0 failures
ALL E2E TESTS PASSED
```

### Codex CLI E2E

```text
$ codex exec -p local "Say hello in one word"
Hello
tokens used: 6,119

$ codex exec -p local "Run 'echo hello world' and show the output"
exec: /bin/zsh -lc 'echo hello world' succeeded
hello world
tokens used: 1,054
```

## Test plan

- [x] 14 pytest tests covering all code paths (schema, streaming, tools, roles, keys)
- [x] Async OpenAI SDK: non-streaming, streaming, concurrent stress
- [x] Streaming events: sequence_number strictly increasing, output_index/content_index present
- [x] response.created has full response object with all fields
- [x] Non-function tools silently skipped (200 not 400)
- [x] Developer/system messages merged correctly
- [x] Codex CLI connects and works (text + tool calling)
- [x] Function call IDs consistent between added/done streaming events

## Requirements

- AI usage disclosure: YES - AI was used to help draft code, tests, and PR text. The submitter reviewed and is responsible for the final changes.

If reviewers prefer, this can be split into smaller PRs for faster review. Let me know.


ngxson:
please add proper testings for it

krystophny:
> please add proper testings for it

Done!

fumlig:
According to the [responses API reference for streaming events](https://developers.openai.com/api/reference/resources/responses/streaming-events#response.created), the `response.created` and `response.in_progress` streaming events should also contain `created_at`, `model`  etc.

It seems like the current implementation just returns a minimal response object in those events. This causes issues with certain spec-compliant client libraries like [async-openai](https://github.com/64bit/async-openai). Would it be possible to add the missing streaming event fields here as well?

krystophny:
@fumlig Thanks for the feedback. The streaming events now include the full response object in `response.created` and `response.in_progress` events (with all 24+ required fields, `status: "in_progress"`, `completed_at: null`, zeroed usage). All partial events (`output_item.added`, `content_part.added`, all deltas) now have `sequence_number`, `output_index`, and `content_index` per spec.

Tested with the async OpenAI Python SDK (which validates event schemas similarly to async-openai on the Rust side).

@ngxson Tests added: 14 pytest tests covering schema fields, streaming compliance, tool type skipping, developer role merging, key stripping, multi-turn input, and output_text consistency. Plus E2E tests with async OpenAI SDK against Qwen3.5-9B.

If you'd prefer this split into smaller PRs for faster review, happy to do so.

krystophny:
Additional E2E testing (async OpenAI SDK, Codex CLI, concurrent stress tests, multiple Qwen3.5 models) is documented in the companion meta-repo: https://github.com/krystophny/llama.cpp-dev

The meta-repo includes a Nix flake for reproducible test environments and scripted test harnesses under `scripts/`.

ngxson:
I'm ok with the current PR, but could you let us know when you are finally ready for review? I have been re-running the CI each time you pushed a new PR, and without CI passed, I cannot merge it

michaelw9999:
Nice work, thank you!  I tried using it with the Codex IDE extension and almost everything works, but still had a few issues causing errors that would freeze up the entire thread, making it unusable to continue from where it was left off.  I'll share some small suggested changes soon.

tools/server/server-common.cpp
```diff
@@ -1255,45 +1255,59 @@ json convert_responses_to_chatcmpl(const json & response_body) {
...
-                                {"type", "text"},
-                            });
                         }
+                        chatcmpl_content.push_back({
+                            {"text", output_text.at("text")},
+                            {"type", "text"},
+                        });
                     } else if (type == "refusal") {
                         if (!exists_and_is_string(output_text, "refusal")) {
                             throw std::invalid_argument("'Refusal' requires 'refusal'");
```
michaelw9999:
```suggestion
```
The throw errors make the thread crash, and unable to resume from there.  It would never reach the pushback below

tools/server/server-common.cpp
```diff
@@ -1255,45 +1255,59 @@ json convert_responses_to_chatcmpl(const json & response_body) {
...
-                            // Ignore annotations and logprobs for now
-                            chatcmpl_content.push_back({
-                                {"refusal", output_text.at("refusal")},
-                                {"type", "refusal"},
-                            });
                         }
```
michaelw9999:
```suggestion
                        
```

tools/server/server-common.cpp
michaelw9999:
```suggestion
                            chatcmpl_content.push_back({{"text", "[image input - no URL provided]"}, {"type", "text"}});
                            continue;
```
The throw errors cause the thread to crash and not able to continue

michaelw9999:
```suggestion
                            chatcmpl_content.push_back({{"text", "[text content missing - input_text is not supported yet!"}, {"type", "text"}});
                            continue;
```

michaelw9999:
```suggestion
                        // Show the attached file as text
                        if (input_item.contains("filename")) {
                            chatcmpl_content.push_back({{"text", "[file: " + input_item.at("filename").get<std::string>() + "]"}, {"type", "text"}});
                        } else {
                            chatcmpl_content.push_back({{"text", "[file data]"}, {"type", "text"}});
                        }
                    } else {
                        // If not text then give json representation instead of the entire item
                        chatcmpl_content.push_back({{"text", input_item.dump()}, {"type", "text"}});
```
This outputs the file to the LLM

michaelw9999:
```suggestion
                        chatcmpl_content.push_back({{"text", input_item.dump()}, {"type", "text"}});
```

tools/server/server-common.cpp
```diff
@@ -1255,45 +1255,59 @@ json convert_responses_to_chatcmpl(const json & response_body) {
...
-                auto chatcmpl_content = json::array();
+                // Also handles AssistantMessageItemParam / EasyInputMessage with role "assistant"
+                std::vector<json> chatcmpl_content;
 
                 for (const auto & output_text : item.at("content")) {
                     const std::string type = json_value(output_text, "type", std::string());
-                    if (type == "output_text") {
+                    if (type == "output_text" || type == "input_text") {
                         if (!exists_and_is_string(output_text, "text")) {
                             throw std::invalid_argument("'Output text' requires 'text'");
```
michaelw9999:
```suggestion
```

tools/server/server-common.cpp
```diff
@@ -1255,45 +1255,59 @@ json convert_responses_to_chatcmpl(const json & response_body) {
...
-                            // Ignore annotations and logprobs for now
-                            chatcmpl_content.push_back({
-                                {"text", output_text.at("text")},
-                                {"type", "text"},
-                            });
                         }
```
michaelw9999:
```suggestion

```

tools/server/server-common.cpp
```diff
@@ -1255,45 +1255,59 @@ json convert_responses_to_chatcmpl(const json & response_body) {
...
-                                {"type", "refusal"},
-                            });
                         }
+                        chatcmpl_content.push_back({
+                            {"refusal", output_text.at("refusal")},
+                            {"type", "refusal"},
+                        });
                     } else {
-                        throw std::invalid_argument("'type' must be one of 'output_text' or 'refusal'");
+                        throw std::invalid_argument("'type' must be 'output_text', 'input_text', or 'refusal'");
```
michaelw9999:
```suggestion
                        chatcmpl_content.push_back({{"text", "[Incorrect output: type must be 'output_text','input_text', or 'refusal'"}, {"type", "text"}});
```

tools/server/server-common.cpp
krystophny: Thanks for the report. The throw here doesn't actually crash the thread -- it's caught by `ex_wrapper` in server.cpp (line 38-69) which converts it to a clean HTTP 400 response. The server keeps running.

This throw fires when `input_text` is missing its required `text` field, which is a genuine client bug. Returning 400 is correct here -- it tells the client to fix the malformed request. Kept as-is.

krystophny: Same situation as the input_text case above -- the throw is caught by `ex_wrapper` and returns a proper 400, not a crash. An `input_image` without `image_url` is malformed input, so 400 is the right response. Kept as-is.

krystophny: Good call on this one. I've reworked `input_file` handling in edb7a79: instead of rejecting, it now renders the file content as text. If `file_data` is present, it gets injected into the prompt (with a `[file: name]` label when filename is available). If only `filename` is present, it shows a placeholder. Also removed the dead commented-out code.

The approach differs slightly from your suggestion -- I didn't add a raw `input_item.dump()` catch-all since injecting arbitrary JSON into the prompt can confuse models. Instead, unknown types are just skipped with a server warning log.

krystophny: Agreed that rejecting the whole request for an unknown type is too harsh. Changed to skip with a warning log in edb7a79. I went with a silent skip rather than `input_item.dump()` to avoid injecting raw JSON into the prompt, which tends to confuse models.

tools/server/server-common.cpp
krystophny: This guard needs to stay. If you remove the check but keep the `.at("text")` call below, it throws a `json::exception` when `text` is missing -- which `ex_wrapper` catches as a 500 (server error) instead of the current 400 (client error). So removing the guard actually makes things worse. The throw here is correctly caught and returned as a 400 response, not a crash.

tools/server/server-common.cpp
krystophny: Same as above -- part of the same validation guard that protects against a 500 from the `.at()` call. Kept as-is.

tools/server/server-common.cpp
krystophny: Same reasoning as the output_text guard -- this protects against a 500 from `.at("refusal")` when the field is missing. The throw is caught by `ex_wrapper` and returns a clean 400. Removing it would downgrade a proper 400 to a confusing 500. Kept as-is.

tools/server/server-common.cpp
krystophny: Part of the same guard, see above.

tools/server/server-common.cpp
krystophny: Changed in edb7a79 -- unknown assistant content types are now skipped with a warning log instead of rejecting. I went with a clean skip rather than injecting placeholder text into the prompt, since the model doesn't benefit from seeing error messages in its conversation history.

tools/server/server-common.cpp
krystophny: Updated in fa2ca43 -- you were right about this one. With Codex IDE replaying the full conversation history, any 400 here permanently kills the thread. Now skips the malformed item with a server warning instead of rejecting.

krystophny: Updated in fa2ca43 -- same reasoning as above. If an image attachment ends up missing the URL in the replayed history, the thread shouldn't die. Now skips with a warning.

tools/server/server-common.cpp
krystophny: Updated in fa2ca43 -- reconsidered after your clarification about the Codex IDE thread loop. The guard now skips the item with a warning + continue instead of throwing. This way a malformed item in replayed history won't kill the thread.

tools/server/server-common.cpp
krystophny: Updated in fa2ca43 -- same treatment. Skips with warning instead of throwing.

krystophny:
@michaelw9999 after giving it some thought, I'm patching this now to move usual mistakes to debug logging and reply to the model with some hint so it may recover. If you could play with it a bit in your setup this would be great to get some more feedback. In case only minor issues remain and the rest seems solid to you I would be happy if you could merge the PR sooner than later and file the remaining issues and I will deal with them then individually in smaller follow up PRs. What do you think?

TimexPeachtree (approved):

SharkWipf:
What's the status on this? Just waiting for merge? I seem to remember there were already follow-ups waiting for this to be merged for better codex image handling I think, I would love to see this + that follow-up merged.

SharkWipf:
> You are welcome. I have more optional nice-to-haves for later after this gets merged (web_search and image handling if the model supports it, automatic compaction). Right now sending an image as an attachment or having it look at one, it just eats up tons of tokens and sends it as base64, then the blind-model has no idea what to do with it. Thanks again for getting this up!

@michaelw9999 do you happen to have these follow-up changes published somewhere? I don't see them on your fork. I've been manually merging this PR for the past few llama.cpp versions and it's been working great, but I'd really like "proper" compaction and image support a lot too.

michaelw9999:
Ahh  yes, I had assumed this was going to get merged and I would add them later.  I will get to working on putting the rest of it up.

krystophny:
> Ahh yes, I had assumed this was going to get merged and I would add them later. I will get to working on putting the rest of it up.

Thx! I rebase once more

michaelw9999:
@krystophny just about done, do you want to rebase again? Then I'll post a commit and base it back on the PR.  The update changes quite a lot.  Been testing it heavily and so far it seems pretty much fully compatible with the VSCode extension with everything I can throw at it.  See below...
Screenshot 2026-04-30 144054
Screenshot 2026-04-30 144134
Screenshot 2026-04-30 144207
Screenshot 2026-04-30 144322
Screenshot 2026-04-30 144421
Screenshot 2026-04-30 144919
Screenshot 2026-04-30 144919
Screenshot 2026-04-30 145305
Screenshot 2026-04-30 151325
Screenshot 2026-04-30 151412
Screenshot 2026-04-30 152214
Screenshot 2026-04-30 152347
Screenshot 2026-04-30 152644
Screenshot 2026-04-30 152710
Screenshot 2026-04-30 152956
Screenshot 2026-04-30 153306
Screenshot 2026-04-30 153612

krystophny:
@michaelw9999 Done!

therealkenc:
> @krystophny just about done, do you want to rebase again? Then I'll post a commit and base it back on the PR. 

I have a [small patch](https://gist.github.com/therealkenc/e90347f1e298c60492f229349f949124) to enable image tool call responses. Working for me with Qwen/Qwen3.6-35B-A3B and my local MCP.

michaelw9999:
Sorry for the extra wait @krystophny , I had found some bugs but I believe I fixed them.     Near full Codex VScode support is finally [Here](https://github.com/ggml-org/llama.cpp/commit/fb76f8f3cfd1eee43de43e15dcd76fd2d5a300a3) 
@ngxson @pwilkin sorry this is large, I tried to make it as small as I can.  I am not super familiar with this part of the codebase yet, and hope all the changes are OK.  When using with Codex on Vscode, everything is working now.  Had to decode the .vsix extension and get into the codex-cli src to get the full semantics, and some subtle edits to push smaller models.  It's quite incredible but Qwen3.5-4B works really well with it and can get a lot done, and it's super fast.  But if the model gets into a thinking loop, you might have to interrupt it to get it back again.  The reasoning budget and mapping of Intelligence levels probably still need some work.
Things brings in automatic compaction, `web_search` and `file_search` and is super easy to configure, for example:

```
model = "qwen3.5-4B-NVFP4"
model_provider = "llamacpp"
personality = "friendly"
model_context_window = 128000
model_auto_compact_token_limit = 100000
model_supports_reasoning_summaries = true
model_reasoning_summary = "auto"
model_reasoning_effort = "medium"

[model_providers.llamacpp]
name = "Local llama.cpp"
model = "Qwen3.5-4B-NVFP4.gguf"
base_url = "http://192.168.50.50:43901/v1"
supports_websockets = false

[model_providers.llamacpp.http_headers]
X-Llama-Responses-Web-Search-Wrapper = "tvly"
X-Llama-Responses-File-Search-Wrapper = "rg"
X-Llama-Responses-Reasoning-Budget-Tokens = "minimal=2048,low=4096,medium=8192,high=16384,xhigh=32768"
```
For the automatic compaction to work, you must set `model_context_window` and `model_auto_compact_token_limit`.  Summary boxes and clickable diffs with the undo button ususally need `model_supports_reasoning_summaries = true` and `model_reasoning_summary = "auto"`.
Just install tavily (but shell command is `tvly`) and `rg` or any other preferred web search MCP or file search/locator tool, it will wrap it through the shell and integrate it more natively and intuitively.  If left out, it will hide these tools from the model.

therealkenc:
> Sorry for the extra wait @krystophny , I had found some bugs but I believe I fixed them. Near full Codex VScode support is finally [Here](https://github.com/ggml-org/llama.cpp/commit/fb76f8f3cfd1eee43de43e15dcd76fd2d5a300a3) 

Thanks for this. Mostly worked for me, although I hit a problem on multi-turn Responses + tools + image content (e.g. `image_url` in a `function_call_output`). Second turn returns `400: unsupported content[].type with type='media_marker'`.

Cause is that `oaicompat_chat_params_parse` takes `json & body` and mutates content parts in place. For every `image_url` and `input_audio` part it pulls the bytes out into `out_files` and rewrites the part to `{type: "media_marker", text: <marker>}`.

Proposed fix, snapshot before the first parse fires.

```
diff --git a/tools/server/server-context.cpp b/tools/server/server-context.cpp
index d4f10e59f..c82d0b77c 100644
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -3893,14 +3893,27 @@ void server_routes::init_routes() {
             }
         }
         SRV_DBG("converted request: %s\n", debug_body.dump().c_str());
+        // Pre-build the parser variant before the first parse:
+        // oaicompat_chat_params_parse mutates `body` content parts in
+        // place — every `image_url` / `input_audio` part is rewritten
+        // to `media_marker` after the bytes are extracted. A second
+        // parse over the mutated body then throws
+        // "unsupported content[].type='media_marker'". Snapshot the
+        // body up front so the parser pass starts from the original
+        // request shape.
+        const bool needs_parser_pass =
+            body.contains("tools") && body.at("tools").is_array() && !body.at("tools").empty();
+        json parser_body;
+        if (needs_parser_pass) {
+            parser_body = responses_relax_tool_required_for_parser(body);
+        }
         json body_parsed = oaicompat_chat_params_parse(
             body,
             meta->chat_params,
             files,
             /* no_prefill_assistant */ true);
-        if (body.contains("tools") && body.at("tools").is_array() && !body.at("tools").empty()) {
+        if (needs_parser_pass) {
             std::vector<raw_buffer> parser_files;
-            json parser_body = responses_relax_tool_required_for_parser(body);
             json parser_parsed = oaicompat_chat_params_parse(
                 parser_body,
                 meta->chat_params,
```

michaelw9999:
What model did you use, I'll try to test it out.  I never actually tried
doing images or using a model that actually does the image.  Just tried
calling up the files/urls and seeing the images on the chat and the
model describing the file name.

therealkenc:
> What model did you use, I'll try to test it out. 

[Qwen/Qwen3.6-35B-A3B](https://huggingface.co/unsloth/Qwen3.6-35B-A3B-GGUF). I haven't tried Codex out (your use-case), but I'll probably play around later today for fun. I am using it against a local MCP that does vision inference against webpage snapshots as PNG arrays.

SharkWipf:
(FWIW, I believe you can PR directly to https://github.com/krystophny/llama.cpp/tree/responses-api-codex-compat (the branch backing this PR) @michaelw9999 and then @krystophny can merge it and it'll show up directly in this PR, instead of working on separate trees. I think 😅)

michaelw9999:
> (FWIW, I believe you can PR directly to https://github.com/krystophny/llama.cpp/tree/responses-api-codex-compat (the branch backing this PR) @michaelw9999 and then @krystophny can merge it and it'll show up directly in this PR, instead of working on separate trees. I think 😅)

I've been trying to do that but all I could manage was getting it done in my own fork, it won't let me in @krystophny 's and to do so on llama.cpp itself with the other base commits, it gives the error "Pull request creation failed. Validation failed: must be a collaborator"... hm,m.  Maybe @krystophny you have to update your llama.cpp fork...

krystophny:
On it ...

krystophny:
Done: rebased on current master, included @michaelw9999’s commit, added the image/tool parser fix and regression test. Local server tests pass (`315 passed, 3 skipped, 231 deselected`).

SharkWipf:
Nice work. There were still a few bugs with MCP tool call handling that made Codex CLI unable to run any MCP calls (MCP calls weren't being exposed correctly).  
I have a branch [https://github.com/SharkWipf/llama.cpp/tree/responses-api-codex-compat](here) that is not in any reasonably upstreamable state and entirely vibe coded (so wouldn't be eliglible for llama.cpp upstreaming anyway), but has MCP fixed, and has a most likely very poorly integrated and not fully tested but seemingly surprisingly functional compaction endpoint enabled (which Codex CLI only seems to use if you configure llama.cpp as an Azure endpoint).  

This is written for my personal use, I was not and am not planning on upstreaming it as I cannot vouch for the quality in any reasonable capacity, but if anyone wants to take anything from it, by all means, go ahead.

therealkenc:
Same [here](https://github.com/therealkenc/llama.cpp/commits/pr-21174-codex/). 

- Instead of the snapshot work-around from earlier, the caller's body is left alone. `oaicompat_chat_params_parse()` takes a `const json& body`.
- Added an LRU cache for `item_reference`s. Vercel's `@ai-sdk/openai`  defaults to `openai: { store: true }` because that is what real OpenAI `v1/responses` expects. This matters for multi-turn multi-modal responses.
- Remove the `item.dump()` injection in the `server_chat_convert_responses_to_chatcmpl()` fallthough. That can be a wall of base64 the LLM can't read.
- Some clang-tidy of files touched.

michaelw9999:
This PR is already huge.. I think we should stop adding anything for now and then more could be added in different PRs or this will never get merged.

TimexPeachtree:
> This PR is already huge.. I think we should stop adding anything for now and then more could be added in different PRs or this will never get merged.

Yes will like to get this merged for using Codex CLI with local models using llama.cpp for higher performance.

krystophny:
Rebased onto current master again and force-pushed. Still ready to merge from my side.

krystophny:
@pwilkin could you take a look as second reviewer when you have time?

This is rebased on current master, CI is green, and I’m not planning further scope changes before merge. The remaining question is whether the current Responses/Codex compatibility shape is acceptable for `tools/server`, with follow-up work kept separate.

krystophny:
@ngxson @michaelw9999 Merged upstream master to resolve conflicts and resolved the outdated review threads. Could you trigger CI and merge if it passes? Thanks!

samuelololol:
I’m looking forward to this repair. I really need it. I don‘t know when it will merge 🤔

krystophny:
Fixed a small issue that caused tests to fail. Please approve the workflows once more and provide some final review approval if everything is fine @michaelw9999 .

michaelw9999:
> Fixed a small issue that caused tests to fail. Please approve the workflows once more and provide some final review approval if everything is fine @michaelw9999 .

@krystophny I am actually using the PR and this is great (plus I've made some other tricks too!) but but I am just a user and contributor, not an admin, so I cannot approve anything...

krystophny:
OK sorry for the mess. Now CI checks should pass.

krystophny:
Update once more from upstream/master . This should fix last remaining real CI failure and re-run of checks hopefully one flaky download.

KyleSanderson (approved):

ngxson:
> If reviewers prefer, this can be split into smaller PRs for faster review. Let me know.

yes, I don't think any of the maintainers can review this PR as-is

also, there seems to be some unrelated changes included here. please review your PR again before requesting a maintainer's review

krystophny:
> > If reviewers prefer, this can be split into smaller PRs for faster review. Let me know.
> 
> yes, I don't think any of the maintainers can review this PR as-is
> 
> also, there seems to be some unrelated changes included here. please review your PR again before requesting a maintainer's review

Moved two unrelated changes to #24935 and #24936 . Once those land, I will continue slicing the PR to smaller pieces.

pwilkin (changes requested):
I can't approve the PR as it is because it has a lot of content that I find questionable. While I have no problems with adding fields that are needed for Responses API compliance, I don't understand the behavior here with reporting a lot of native tools, then providing implementations that assume a specific Unix shell environment? I'm not even sure how this is supposed to work because it's completely uncommented, but that's certainly not something that we can merge.

krystophny:
> I can't approve the PR as it is because it has a lot of content that I find questionable. While I have no problems with adding fields that are needed for Responses API compliance, I don't understand the behavior here with reporting a lot of native tools, then providing implementations that assume a specific Unix shell environment? I'm not even sure how this is supposed to work because it's completely uncommented, but that's certainly not something that we can merge.

Thank you for taking a look into this. This was indeed nonsense. Sorry about it. I remove it now.

tools/server/server-chat.cpp
```diff
@@ -1,19 +1,513 @@
...
+    return json {
+        {"name", name}, {"description", description},
+        {"strict", strict}, {"parameters", parameters},
+    };
+}
+
+static std::string responses_custom_tool_description(const json & resp_tool, const std::string & name) {
+    std::string description = json_value(resp_tool, "description", std::string("Freeform Codex tool input."));
+
+    if (name == "apply_patch") {
```
pwilkin: Still see some hardcoded stuff that I'm not sure belongs here. Serving tool descriptions should *always* be the client's responsibility, not the server's (unless they're server-side tools).

tools/server/server-chat.cpp
```diff
@@ -1,19 +1,513 @@
...
+        fn = responses_function_tool(
+                name,
+                responses_custom_tool_description(resp_tool, name),
+                responses_custom_tool_parameters(name));
+    } else if (tool_type == "local_shell") {
+        // Declared so the model can call it; the client runs the command. No server-side execution.
+        name = sanitize_tool_name(json_value(resp_tool, "name", std::string("local_shell")), "local_shell");
+        fn = responses_function_tool(
+                name,
+                "Run a local shell command.",
```
pwilkin: These ones as well, doesn't look like something a server should be handling at all.

tools/server/server-chat.cpp
```diff
@@ -1,19 +1,513 @@
...
+                    : responses_object_schema(
+                            json {
+                                {"query", json {{"type", "string"}}},
+                            },
+                            json::array({"query"})));
+    } else if (tool_type == "task_list" || tool_type == "todo_list") {
+        name = sanitize_tool_name(json_value(resp_tool, "name", std::string("update_plan")), "update_plan");
+        fn = responses_function_tool(
+                name,
+                json_value(resp_tool, "description", std::string("Updates the task plan.")),
```
pwilkin: Ditto.

tools/server/server-chat.cpp
```diff
@@ -1,19 +1,513 @@
...
+                    {"command", json {{"type", "array"}, {"items", json {{"type", "string"}}}}},
+                    {"cmd", json {{"type", "string"}}},
+                    {"working_directory", json {{"type", "string"}}},
+                    {"timeout_ms", json {{"type", "integer"}}},
+                }));
+    } else if (tool_type == "tool_search") {
+        name = sanitize_tool_name(json_value(resp_tool, "name", std::string("tool_search")), "tool_search");
+        fn = responses_function_tool(
+                name,
+                json_value(resp_tool, "description", std::string("Search for tools available to this Codex session.")),
```
pwilkin: Ditto.

tools/server/server-chat.cpp
```diff
@@ -1,19 +1,513 @@
...
+        if (!expose_hosted_file_search) {
+            return nullptr;
+        }
+        if (has_function_tool_named(all_tools, name)) {
+            remember_responses_tool(metadata, name, resp_tool, tool_type);
+            return nullptr;
+        }
+        fn = responses_function_tool(
+                name,
+                "Search files in the current workspace. Use mode='files' to find paths by name, or mode='content' to search file contents. Prefer this over shell commands such as find.",
```
pwilkin: Ditto.

tools/server/server-chat.cpp
```diff
@@ -1,19 +1,513 @@
...
+                    {"path", json {{"type", "string"}, {"description", "Optional relative workspace path to search."}}},
+                }, json::array({"query"})));
+    } else if (tool_type == "image_generation") {
+        name = sanitize_tool_name(json_value(resp_tool, "name", std::string("image_generation")), "image_generation");
+        if (!expose_hosted_image_generation) {
+            return nullptr;
+        }
+        fn = responses_function_tool(
+                name,
+                "Generate an image from a prompt.",
```
pwilkin: Ditto.

tools/server/server-chat.cpp
```diff
@@ -167,122 +699,208 @@ json server_chat_convert_responses_to_chatcmpl(const json & response_body) {
...
-                    }},
-                    {"id",   item.at("call_id")},
-                    {"type", "function"},
-                };
+                append_assistant_tool_call(chatcmpl_messages, make_tool_call(
+                    item.at("name").get<std::string>(),
+                    item.at("call_id").get<std::string>(),
+                    parse_arguments_best_effort(item.at("arguments"))));
+            } else if (exists_and_is_string(item, "type") &&
+                (item.at("type") == "function_call" ||
```
pwilkin, line 447: I take it those are actually separate calls, but we should treat them just like all `function_call` instances, i.e. pass them on without modifications.

tools/server/server-task.cpp
```diff
@@ -549,6 +551,402 @@ json server_task_result_cmpl_final::to_json_oaicompat_chat_stream() {
...
+    if (!get_custom_tool_input_value_if_present(parsed_args, current_input)) {
+        return std::string();
+    }
+    if (current_input.size() < previous_input.size()) {
+        return std::string();
+    }
+    return current_input.substr(previous_input.size());
+}
+
+static json build_local_shell_action(const json & args, const std::string & raw_arguments) {
```
pwilkin: This has got to go.

pwilkin (changes requested):

krystophny:
@pwilkin thanks for the remarks. I removed the tool-related code you flagged:

1. The hardcoded tool descriptions and parameter schemas for the non-function tool types (`local_shell`, `apply_patch`, `web_search`, `file_search`, `image_generation`, `tool_search`, `task_list`). The server no longer serves any of these — non-function tools are simply skipped instead of translated.
2. The shell-command synthesis (`build_local_shell_action` and the `bash -lc` wrapping) is gone, so nothing assumes a Unix shell anymore.
3. The per-type reconstruction of tool calls is gone: `*_call` items are relayed as plain `function_call` on both the input and output side, instead of being rebuilt per tool type.

So the converter only relays `function` tools and their calls now. It's in the latest three commits.

pwilkin:
Yeah, this is starting to resemble something merge'able, let's see the test results :)

whyiug:
Independent black-box A/B result (disclosure: I maintain AgentAPI Doctor).

I tested these exact commits:

- current master at test time: `4193ea697f0406e2ccbc16f84251ca5d76a83914`
- this PR head: `8f8009edc9a6bf5e939ea737af52ec403999983b`

Environment: Linux x86_64, CPU-only `llama-server` bound to `127.0.0.1`, `ggml-org/test-model-stories260K`, OpenAI Python 2.14.0. No GPU or external provider endpoint was used.

Results:

- master's existing non-streaming + streaming OpenAI SDK tests: **2 passed**
- the same selected 9 core Responses tests from this PR, executed against master via an external loopback server: **1 passed, 8 failed**
- those 9 tests against the PR head: **9 passed**
- AgentAPI Doctor v0.1.0 raw Responses Quick Check against master: **3 PASS, 1 FAIL** (exit 1)
- the same Doctor check against the PR head: **4 PASS** (exit 0)

The independent raw-wire failure on master was in `openai-responses-http-030-terminal-exactly-once`, which stopped on an invalid output item before accepting the lifecycle:

```text
expected: nonempty item ID/kind and nonnegative index
observed: item kind="message", output_index=-1
finding: invalid_output_item / STREAM_STATE_MACHINE
```

The relevant event shape was:

```json
// master
{"type":"response.output_item.added","item":{"id":"<redacted>","type":"message", "...":"..."}}

// PR head
{"type":"response.output_item.added","sequence_number":2,"output_index":0,
 "item":{"id":"<redacted>","type":"message", "...":"..."}}
```

Reproduction command (port changed between the two local servers):

```sh
doctor test \
  --base-url http://127.0.0.1:<port>/v1 \
  --protocol openai-responses \
  --model tinyllama-2 \
  --allow-plain-http
```

Limits: this evidence is bound to the exact commits above. The PR head has substantially diverged from current master and GitHub currently reports the PR as dirty/non-mergeable, so this does **not** establish mergeability or attribute every difference solely to one patch. Doctor's built-in interpretations are also still candidate interpretations; the result is an observation for these four bounded checks, not compatibility certification.

Tool version used: [AgentAPI Doctor v0.1.0](https://github.com/whyiug/agentapi-doctor/releases/tag/v0.1.0) (`ab281c269bc28b3732cdae8aa52da2201fc87001`).

krystophny:
Thanks, this was useful independent evidence. I pushed a28a6d324, merging current master and resolving the two conflicts in server-task.cpp and the Responses tests. The resolution was AI-assisted and locally reviewed. The merged llama-server builds, and the created-response schema, sequence-number, and telemetry streaming tests pass (3 passed).

whyiug:
Thanks for updating the branch. I reran the same local A/B against `a28a6d324e5e36e593c625985025027c7395adc2`, using AgentAPI Doctor v0.1.1 and the same selected Responses tests.

- master at test time, `f955e394bf94e01e5e36186d13c985727e5ef5b5`: the existing OpenAI SDK smoke tests passed 2/2; the selected 9 PR tests passed 1/9; Doctor reported 3 PASS + 1 FAIL (exit 1)
- PR head, `a28a6d324e5e36e593c625985025027c7395adc2`: the selected tests passed 9/9; Doctor reported 4 PASS (exit 0)

The remaining Doctor failure on master was `openai-responses-http-030-terminal-exactly-once`: the output item had a valid ID and kind, but no nonnegative output index (`index=-1`). That check passed on the PR head.

Environment: Linux x86_64, CPU-only loopback servers, `ggml-org/test-model-stories260K`, Python 3.12.7, OpenAI Python 2.14.0. No GPU or external provider endpoint was used.

For precision, this head merged `33a75f41`; the test-time master was one later, unrelated GGML commit. These results support only the bounded checks above, not full Responses compatibility or merge readiness. The current PR head is unchanged since the run. I appreciate the conflict resolution and the clear update.

PrimalPimmy:
Hey folks. We use llama cpp under our project, and would really appreciate if you could tell me the status of this PR!

If there is anything pending, I would love to contribute in case there is no bandwidth. Looking forward to this :)

ServeurpersoCom:
> Hey folks. We use llama cpp under our project, and would really appreciate if you could tell me the status of this PR!
> 
> If there is anything pending, I would love to contribute in case there is no bandwidth. Looking forward to this :)

Hi, the PR is currently blocked: ngxson asked for it to be split into smaller PRs, and pwilkin's review is still at "changes requested". It also conflicts with master in three server files.

The streaming compliance part (sequence_number, output_index, content_index, full response.created) is the most self-contained piece and could land first on its own.

You can either check with @krystophny or do the split yourself. We can only review small, focused PRs.

krystophny:
@PrimalPimmy please go ahead and pick up the pieces and split this PR. While I fully understand the volume of PRs you guys are dealing with, I gave up to track the status after a few months. Closing now and happy if some of this makes it into a set of smaller PRs.

PR #21176, opened 2026-03-30
common : init in params parser, add Windows UTF-8 support

angt:
## Overview

Simplify the common init process and fix windows logging

## Additional information

We already call `common_params_parse()` first, then `common_init()`, so this adjusts the code to call `common_init()` first (directly from `common_params_parse`).

The build info is now only for debug, so we avoid the duplicate with `--version`.

The UTF-8 setup at the beginning is needed to avoid logging garbage on Windows.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -359,6 +359,11 @@ bool parse_cpu_mask(const std::string & mask, bool (&boolmask)[GGML_MAX_N_THREAD
 }
 
 void common_init() {
+#if defined(_WIN32)
+    SetConsoleOutputCP(CP_UTF8);
+    SetConsoleCP(CP_UTF8);
+#endif
+
     llama_log_set(common_log_default_callback, NULL);
 
 #ifdef NDEBUG
@@ -367,7 +372,7 @@ void common_init() {
     const char * build_type = " (debug)";
 #endif
 
-    LOG_INF("build: %d (%s) with %s for %s%s\n", LLAMA_BUILD_NUMBER, LLAMA_COMMIT, LLAMA_COMPILER, LLAMA_BUILD_TARGET, build_type);
+    LOG_DBG("build: %d (%s) with %s for %s%s\n", LLAMA_BUILD_NUMBER, LLAMA_COMMIT, LLAMA_COMPILER, LLAMA_BUILD_TARGET, build_type);
 }
 
 std::string common_params_get_system_info(const common_params & params) {
--- a/examples/batched/batched.cpp
+++ b/examples/batched/batched.cpp
@@ -24,11 +24,11 @@ int main(int argc, char ** argv) {
     params.prompt = "Hello my name is";
     params.n_predict = 32;
 
+    common_init();
+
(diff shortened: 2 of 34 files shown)


angt:
On windows, without the patch:
```
PS Z:\>  C:\Users\angt\installama\server.exe -hf angt/modelutf8
common_download_file_single_online: no previous model file found C:\Users\angt\AppData\Local\llama.cpp\angt_modelutf8_preset.ini
common_download_file_single_online: HEAD failed, status: 404
no remote preset found, skipping
common_download_file_single_online: no previous model file found C:\Users\angt\.cache\huggingface\hub\models--angt--modelutf8\blobs\bd258782e35f7f458f8aced1adc053e6e92e89bc735ba3be89d38a06121dc517
common_download_file_single_online: downloading from https://huggingface.co/angt/modelutf8/resolve/85ca32e076fd0d1f3817a2cbc78cb34ebae46459/M├╢d├¿le_≡ƒÜÇ_µ╡ïΦ»ò_╪º╪«╪¬╪¿╪º╪▒_╨ó╨╡╤ü╤é-Q4_K_M.gguf to 
```

common/arg.cpp
```diff
@@ -876,6 +876,8 @@ bool common_params_to_map(int argc, char ** argv, llama_example ex, std::map<com
 }
 
 bool common_params_parse(int argc, char ** argv, common_params & params, llama_example ex, void(*print_usage)(int, char **)) {
+    common_init();
+
```
ngxson: I'm not quite sure about this, but on master, `common_init()` is called after `common_params_parse()`, and here it will make the order to be reversed

angt: The only issue I managed to find was the duplication of the build log when using `--version`.

But having a `common_init()` called after parsing params isn't clear at all, and I think we should rename `common_params_parse()` to something like `common_init_with_params()` at some point, but not now.

Also, `common_params_parse()` already setup tty, logs, ggml backends, ...
A bigger refactoring will be needed, to be able to split `common` from `llama`.

And putting the UTF-8 setup alone in `common_params_parse()` and `common_init()` would be going in the wrong direction.

ngxson: IIRC correctly, there are also examples using `common_init()` but doesn't use the arg parsing system from libcommon, so just doubt is it will break these cases.

I think it's probably better to explicitly call `common_init()` at the start of `main()`, similar to `convert-llama2c-to-ggml.cpp`. Explicit init calls like this can be better for clarity.

ngxson: Otherwise, we can also have a static global `is_common_init = false` to avoid potential double call to `common_init`

angt: The code is already in a bad shape IMHO, let's not making it worst

angt: I removed `common_init()` from `common_params_parse()`

ngxson (approved):

ggerganov (approved):

PR #21181, opened 2026-03-30
CUDA: Fix CUB's argsort when nrows % block_size == 0 CCCL < 3.1

ORippler:
## Overview

We wrongly calculated offset_grid as `ceildiv(nrows, block_size)`, while it must be `ceildiv(nrows + 1, block_size)`. As a consequence, we had uninitialized values in `offset_iterator[nrows]` for the case when `nrows % block_size == 0`.

This bug affected CCCL < 3.1 only, as newer CCCL versions take the strided_iterator path 

## Additional information

Fixes #21162

# Requirements

- AI usage disclosure: Yes, to ask some clarifying questions during implementation.

### DIFF
--- a/ggml/src/ggml-cuda/argsort.cu
+++ b/ggml/src/ggml-cuda/argsort.cu
@@ -47,9 +47,11 @@ void argsort_f32_i32_cuda_cub(ggml_cuda_pool & pool,
 #ifdef STRIDED_ITERATOR_AVAILABLE
     auto offset_iterator = cuda::make_strided_iterator(cuda::make_counting_iterator(0), ncols);
 #else
-    ggml_cuda_pool_alloc<int> offsets_alloc(pool, nrows + 1);
+    // offset_iterator needs to populate nrows + 1 elements, so we also have to ceildiv nrows + 1 by block_size
+    const int                 nrows_offset = nrows + 1;
+    ggml_cuda_pool_alloc<int> offsets_alloc(pool, nrows_offset);
     int *                     offset_iterator = offsets_alloc.get();
-    const dim3                offset_grid((nrows + block_size - 1) / block_size);
+    const dim3                offset_grid((nrows_offset + block_size - 1) / block_size);
     init_offsets<<<offset_grid, block_size, 0, stream>>>(offset_iterator, ncols, nrows);
 #endif
     CUDA_CHECK(cudaMemcpyAsync(temp_keys, x, ncols * nrows * sizeof(float), cudaMemcpyDeviceToDevice, stream));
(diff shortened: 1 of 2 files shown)


JohannesGaessler (approved):

CISC:
Maybe I misunderstood, but didn't @fairydreaming say that it crashed with CUB disabled?

ORippler:
> Maybe I misunderstood, but didn't @fairydreaming say that it crashed with CUB disabled?

Quoting from the issue
>> [@fairydreaming](https://github.com/fairydreaming) How exactly did you configure your CUDA build? On CTK 13.1 and CCCL 3.1, tests pass for me.

>@ORippler I tested:
>CUDA 12.8, no CCCL -> crash
>CUDA 13.2 and CCCL 13.2.27 -> all OK

CUDA Toolkit (CTK) >= 11.0 always comes [bundled with CUB](https://docs.nvidia.com/cuda/archive/11.0/cuda-toolkit-release-notes/#cuda-whats-new), which was later on integrated into CCCL.

So in CTK 12.8 we have CCCL's CUB components available (see [here](https://github.com/ggml-org/llama.cpp/blob/389c7d4955ba55c7418afaebf7c23d9ed64ef707/ggml/src/ggml-cuda/common.cuh#L104-L106) for the compile-time check which limits CUB availability to CTK >= 11.7). I see no way to disable this via cmake.

Moreover, we correctly limit the support surface to bitonic sort in the case where CUB is unavailable in [`ggml_backend_cuda_device_supports_op`](https://github.com/ggml-org/llama.cpp/blob/389c7d4955ba55c7418afaebf7c23d9ed64ef707/ggml/src/ggml-cuda/ggml-cuda.cu#L5020-L5026).

CTK 12.8 comes bundled with CUB 2.7, which does not yet have `strided_iterator` available (this was added in CCCL 3.1, [bundled with CTK 13.1](https://docs.nvidia.com/cuda/archive/13.1.0/cuda-toolkit-release-notes/index.html#id6)). CTK 12.8 thus entered the explicit offset_index path, which was bugged exactly for nrows % block_size == 0 (where block_size = 256 currently).

CISC:
@ORippler My confusion may stem from the statement in the related PR:
> Due to limitations of the current CUDA ggml_top_k() implementation NVIDIA CUDA CCCL library (version >3.2) and enabling GGML_CUDA_USE_CUB during CUDA backend compilation is needed, otherwise the CUDA implementation will crash for context sizes larger than (I think) 1024 tokens. I use it with CUDA 13.2 and CCCL 13.2.27.

Either way, if @fairydreaming can confirm this completely fixes the issue we're good.

fairydreaming:
> Either way, if @fairydreaming can confirm this completely fixes the issue we're good.

@CISC I confirm that it's all OK with this PR, no more crashes observed, tested up to `TOP_K(type=f32,ne=[164352,2048,1,1],k=2048,ties=0)`.

CISC (approved):

PR #21182, opened 2026-03-30
tests: allow exporting graph ops from HF file without downloading weights

0cc4m:
## Overview

This expands the `export-graph-ops` binary to also allow using `--hf-repo` instead of `--model`. It uses the HF metadata loader from #19796 to set up a dummy model graph without loading weights and parses the cgraph from that, which allows running test-backend-ops on tensors from models without downloading them. That should make checking if a backend works correctly for a specific model/quant much easier, and also allows performance benchmark comparisons without downloads.

I tried to keep the changes to disable actually downloading the model minimal, but let me know if you can see a better way to do this.

# Requirements

- AI usage disclosure: Yes, Claude was used for assistance. Code is written by me.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -537,9 +537,11 @@ static bool common_params_parse_ex(int argc, char ** argv, common_params_context
     } catch (const std::exception & e) {
         LOG_WRN("HF cache migration failed: %s\n", e.what());
     }
+    // export_graph_ops loads only metadata
+    const bool skip_model_download = ctx_arg.ex == LLAMA_EXAMPLE_EXPORT_GRAPH_OPS;
 
     // maybe handle remote preset
-    if (!params.model.hf_repo.empty()) {
+    if (!params.model.hf_repo.empty() && !skip_model_download) {
         std::string cli_hf_repo = params.model.hf_repo;
         bool has_preset = common_params_handle_remote_preset(params, ctx_arg.ex);
 
@@ -570,7 +572,7 @@ static bool common_params_parse_ex(int argc, char ** argv, common_params_context
     }
 
     // handle model and download
-    {
+    if (!skip_model_download) {
         auto res = common_params_handle_model(params.model, params.hf_token, params.offline);
         if (params.no_mmproj) {
             params.mmproj = {};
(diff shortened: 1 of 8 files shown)


tests/export-graph-ops.cpp
```diff
@@ -132,9 +143,50 @@ int main(int argc, char ** argv) {
 
     params.warmup = false;
 
-    auto init_result = common_init_from_params(params);
+    llama_context * ctx;
```
ngxson, line 146: maybe better to use `llama_context_ptr` here?

0cc4m: The problem with that is that if I use `common_init_from_params` then the init_result owns the context. I'd have to disable the constructor of the unique_ptr. That didn't really seem better than just using a pointer.

ngxson: what I mean is:

```cpp
llama_context * ctx;
common_init_result_ptr init_result;
llama_model_ptr model;
llama_context_ptr ctx2; // not using init_result

if (params.model.hf_repo.empty()) {
    ctx = init_result->context();
} else {
    // ...
    ctx2.reset(llama_init_from_model(model.get(), ctx_params));
    ctx = ctx2.get();
}

// ctx2 is freed when going out of scope
```

0cc4m: Good idea, fixed.

tests/gguf-model-data.cpp
```diff
@@ -611,3 +615,93 @@ std::optional<gguf_remote_model> gguf_fetch_model_meta(
...
+        return nullptr;
+    }
+
+    auto & model = model_opt.value();
+
+    const std::string cache_path = get_cache_file_path(cdir, repo_part, filename);
+
+    ggml_context * ggml_ctx;
+    gguf_init_params params{true, &ggml_ctx};
+    gguf_context * ctx = gguf_init_from_file(cache_path.c_str(), params);
```
ngxson: Note that we also have `gguf_context_ptr` and `ggml_context_ptr` from `ggml-cpp.h` header that allow freeing when going out of context. It's recommended to use them when possible

0cc4m: Thank you, using that now.

ngxson (approved):

pwilkin (approved):

common/arg.cpp
```diff
@@ -591,7 +593,7 @@ static bool common_params_parse_ex(int argc, char ** argv, common_params_context
 
     // model is required (except for server)
     // TODO @ngxson : maybe show a list of available models in CLI in this case
-    if (params.model.path.empty() && ctx_arg.ex != LLAMA_EXAMPLE_SERVER && !params.usage && !params.completion) {
+    if (params.model.path.empty() && ctx_arg.ex != LLAMA_EXAMPLE_SERVER && !skip_model_download&& !params.usage && !params.completion) {
```
ggerganov:
```suggestion
    if (params.model.path.empty() && ctx_arg.ex != LLAMA_EXAMPLE_SERVER && !skip_model_download && !params.usage && !params.completion) {
```

ggerganov (approved):

CISC (approved):

PR #21201, opened 2026-03-30
Fix undefined timing measurement errors in server context

thedanhoffman:
## Overview
Fix UB issue reported by Valgrind involving timing measurements for prompt processing and eval

## Additional information

# Requirements

- AI usage disclosure: I threw an AI at this, but the fix is simple enough that there's no real downside risk (independently verified this solves the issue)


ggerganov (approved):

ngxson (approved):

thedanhoffman:
:eyes:

PR #21203, opened 2026-03-31
server: respect the ignore eos flag

ykhrustalev:
## Overview

ignore_eos=true in /completion requests has no effect. server_routes captures a reference to common_params before model loading, but logit_bias_eog is populated after model loading. The captured copy is always empty, so no biases are applied.

Regression test added: sends `ignore_eos=true` and verifies `logit_bias` contains EOG entries with `-inf` bias.

```
pushd tools/server/tests
LLAMA_SERVER_BIN_PATH=../../../build/bin/llama-server \
  uv run --with-requirements requirements.txt \
    pytest unit/test_ignore_eos.py -v -x
popd
```
# Requirements

- AI usage disclosure: YES - AI was used to draft the regression test and review the fix


ngxson (approved):
Can merge when CI passes

ggerganov (approved):

PR #21204, opened 2026-03-31
CANN: Add support for Qwen35 ops

hipudding:
## Overview

 This PR adds support for several missing operators in the CANN (Ascend NPU) backend for qwen3.5
                                                                                                                                                         
 New operators:
- GGML_OP_SET: implement via aclnnInplaceCopy on target region
- GGML_OP_CUMSUM: implement via aclnnCumsum
- GGML_OP_FILL: implement via aclnnInplaceFillScalar
- GGML_OP_DIAG: implement via aclnnInplaceCopy on diagonal strides
- GGML_OP_TRI (lower/lower_diag/upper_diag/upper): implement via
  aclnnTril(-1/0) and aclnnTriu(0/1) with appropriate diagonal offsets
- GGML_OP_SOLVE_TRI: implement via aclnnTriangularSolve
- GGML_UNARY_OP_SOFTPLUS: implement via aclnnSoftplus

Optimizations:
- GLU (SwiGLU/GeGLU/GeGLU_ERF/GeGLU_QUICK): fuse with aclnnSwiGlu /
  aclnnGeGluV3 when applicable; fallback conditions now checked inside
  each function rather than at the call site
- CROSS_ENTROPY_LOSS: replace 5-kernel sequence (LogSoftmax→Mul→
  ReduceSum×2→Muls) with single aclnnSoftmaxCrossEntropyWithLogits call
- L2_NORM: fix in-place ClampMin on norm result (was clamping wrong
  tensor); add eps clamping before division to avoid divide-by-zero
- PAD_REFLECT_1D: eliminate per-ne[3] loop; assert contiguity and call
  ReflectionPad1d once on the full 4-D view; remove redundant nb copies
- GET_ROWS: replace IndexSelect with GatherV2 per batch slice; refactor
  helper into gather_batched lambda with batch loop inlined
- SET_ROWS: replace IndexCopy with InplaceIndexCopy per batch slice;
  refactor helper into scatter_batched lambda with batch loop inlined
- OUT_PROD: replace O(ne[3]*ne[2]*ne[1]) Ger+InplaceAdd loop with
  per-slice Matmul loop (src0 @ src1^T); handles strided-broadcast
  batch dims where ne02/ne03 may differ from ne2/ne3
- backend memset_tensor: implement via aclrtMemset (was NULL)

Bug fixes:
- COUNT_EQUAL: use non-inplace EqTensor into a same-type temporary
  buffer instead of InplaceEqTensor, avoiding corruption of src0
- ACL graph cache (USE_ACL_GRAPH): restore node_type and src_type[]
  fields in ggml_graph_node_properties; has_matching_properties() was
  missing type checks, causing F16 and BF16 tensors (same nb[0]=2) to
  incorrectly share cached graphs and produce wrong results (ERR≈679)
- graph cache op_params matching: compare full GGML_MAX_OP_PARAMS
  bytes so that ops differing only in parameters are not incorrectly
  replayed from cache                

# Requirements

- AI usage disclosure: 
YES — AI tools (Claude code) were used in an assistive capacity only. Specifically, AI was used to analyze problems, suggest implementation approaches, and provide explanations of relevant CANN/ACL APIs. All code was written, reviewed line-by-line, and validated by the human contributor, who takes full responsibility for the correctness and design of the changes.  

**Known issue**: Due to kernel caching, in models like qwen3.5 where each layer is not entirely identical, a "kernel cache not found" error occurs during multi-device inference. Workaround: Disable kernel caching by setting environment variables. ACLNN_CACHE_LIMIT=0

### DIFF
--- a/ggml/src/ggml-cann/aclnn_ops.cpp
+++ b/ggml/src/ggml-cann/aclnn_ops.cpp
@@ -25,6 +25,7 @@
 #include "ggml-impl.h"
 #include "ggml.h"
 
+
 #include <aclnnop/aclnn_add.h>
 #include <aclnnop/aclnn_add_rms_norm.h>
 #include <aclnnop/aclnn_addcdiv.h>
@@ -45,7 +46,9 @@
 #include <aclnnop/aclnn_fused_infer_attention_score_v2.h>
 #include <aclnnop/aclnn_ger.h>
 #include <aclnnop/aclnn_group_norm.h>
+#include <aclnnop/aclnn_gather_v2.h>
 #include <aclnnop/aclnn_grouped_matmul_v3.h>
+#include <aclnnop/aclnn_scatter.h>
 #include <aclnnop/aclnn_gt_scalar.h>
 #include <aclnnop/aclnn_im2col.h>
 #include <aclnnop/aclnn_index_copy.h>
@@ -62,6 +65,7 @@
 #include <aclnnop/aclnn_permute.h>
 #include <aclnnop/aclnn_pow.h>
 #include <aclnnop/aclnn_pow_tensor_tensor.h>
+#include <aclnnop/aclnn_recurrent_gated_delta_rule.h>
 #include <aclnnop/aclnn_reduce_sum.h>
 #include <aclnnop/aclnn_reflection_pad1d.h>
 #include <aclnnop/aclnn_repeat.h>
@@ -69,11 +73,15 @@
 #include <aclnnop/aclnn_rms_norm.h>
 #include <aclnnop/aclnn_roll.h>
 #include <aclnnop/aclnn_softmax.h>
+#include <aclnnop/aclnn_softmax_cross_entropy_with_logits.h>
 #include <aclnnop/aclnn_sub.h>
 #include <aclnnop/aclnn_sum.h>
 #include <aclnnop/aclnn_threshold.h>
 #include <aclnnop/aclnn_tril.h>
+#include <aclnnop/aclnn_triangular_solve.h>
 #include <aclnnop/aclnn_triu.h>
+#include <aclnnop/aclnn_logical_not.h>
+#include <aclnnop/aclnn_masked_fill_scalar.h>
 #include <aclnnop/aclnn_upsample_nearest_2d.h>
 #include <aclnnop/aclnn_weight_quant_batch_matmul_v2.h>
 #include <aclnnop/aclnn_zero.h>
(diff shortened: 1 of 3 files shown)


tests/test-backend-ops.cpp
```diff
@@ -3689,6 +3689,20 @@ struct test_gated_delta_net : public test_case {
         : type(type), head_count(head_count), head_size(head_size), n_seq_tokens(n_seq_tokens), n_seqs(n_seqs),
           v_repeat(v_repeat), permuted(permuted), kda(kda) {}
 
+    double max_nmse_err() override {
```
hipudding: revert temporary debug changes.

ggml/src/ggml-cann/ggml-cann.cpp
```diff
@@ -1835,6 +1748,20 @@ static bool ggml_cann_compute_forward(ggml_backend_cann_context & ctx, struct gg
                 case GGML_UNARY_OP_STEP:
                     ggml_cann_step(ctx, dst);
                     break;
+                case GGML_UNARY_OP_SOFTPLUS:
+                    {
+                        auto lambda = [](ggml_backend_cann_context & ctx, aclTensor * acl_src, aclTensor * acl_dst) {
+                            float beta_val = 1.0f;
```
hipudding: Use smart pointer

ggml/src/ggml-cann/aclnn_ops.cpp
```diff
@@ -151,6 +169,92 @@ void ggml_cann_op_unary_gated(std::function<void(ggml_backend_cann_context &, ac
...
+// SiLU to left half, multiplies by right half. Only used when src1==nullptr
+// and swapped==0 (single-tensor, standard order).
+//
+// CANN tiling for SwiGlu requires (storageShapeDim + viewDims) to be even.
+// aclCreateTensor always uses storageShapeDim=1, so viewDims must be odd.
+// We use a 3D view (1+3=4, even) to satisfy this constraint while preserving
+// correct split semantics along the innermost (ne[0]) dimension.
+void ggml_cann_swiglu(ggml_backend_cann_context & ctx, ggml_tensor * dst) {
+    // aclnnSwiGlu requires the split dim (src->ne[0]) to be even; fall back otherwise.
+    if (dst->src[0]->ne[0] % 2 != 0) {
```
hipudding, line 185: It's better to check support in function, not in support_op function.

ggml/src/ggml-cann/aclnn_ops.cpp
```diff
@@ -151,6 +169,92 @@ void ggml_cann_op_unary_gated(std::function<void(ggml_backend_cann_context &, ac
...
+}
+
+// Fused GeGLU using aclnnGeGluV3: splits input along ne[0] (CANN last dim),
+// activates the LEFT half with GELU, multiplies by right half.
+// approximate: 0=tanh, 1=none(erf). activateLeft=true matches GGML convention.
+// outGelu is a required-but-discard output buffer.
+// Only used when src1==nullptr and swapped==0.
+void ggml_cann_geglu(ggml_backend_cann_context & ctx, ggml_tensor * dst, int64_t approximate) {
+    // aclnnGeGlu requires the split dim (src->ne[0]) to be even; fall back otherwise.
+    if (dst->src[0]->ne[0] % 2 != 0) {
```
hipudding, line 234: Same above.

ggml/src/ggml-cann/aclnn_ops.cpp
```diff
@@ -151,6 +169,92 @@ void ggml_cann_op_unary_gated(std::function<void(ggml_backend_cann_context &, ac
...
+    size_t  elem_size    = ggml_element_size(dst);
+    int64_t ne[GGML_MAX_DIMS] = { dst->ne[0], dst->ne[1], dst->ne[2], dst->ne[3] };
+    size_t  nb[GGML_MAX_DIMS];
+    nb[0] = elem_size;
+    for (int i = 1; i < GGML_MAX_DIMS; i++) {
+        nb[i] = nb[i - 1] * ne[i - 1];
+    }
+    size_t gelu_out_size = nb[GGML_MAX_DIMS - 1] * ne[GGML_MAX_DIMS - 1];
+    void * gelu_out_data = nullptr;
+    ACL_CHECK(aclrtMalloc(&gelu_out_data, gelu_out_size, ACL_MEM_MALLOC_HUGE_FIRST));
```
hipudding: raw memory alloc and free is not suggest to use here. Use memory pool instead.

ggml/src/ggml-cann/aclnn_ops.cpp
```diff
@@ -1695,152 +1966,99 @@ void ggml_cann_softmax(ggml_backend_cann_context & ctx, ggml_tensor * dst) {
...
+    };
+
     switch (src0->type) {
         case GGML_TYPE_BF16:
         case GGML_TYPE_F16:
         case GGML_TYPE_F32:
             if (src0->type == dst->type) {
-                aclnn_index_select_4d(ctx, src0->data, src0->ne, src0->nb, dst->data, dst->ne, dst->nb, src1,
-                                      dst->type);
+                for (int64_t i3 = 0; i3 < src0->ne[3]; i3++) {
```
hipudding: Put these two for loop in gather_2d lambda, and reduce parameter count.

ggml/src/ggml-cann/aclnn_ops.cpp
```diff
@@ -652,6 +763,166 @@ void ggml_cann_sum(ggml_backend_cann_context & ctx, ggml_tensor * dst) {
...
+    void * mask_inv_buf = mask_inv_alloc.get();
+    acl_tensor_ptr acl_mask_inv = ggml_cann_create_tensor(mask_inv_buf, ACL_BOOL, nb_bool, ne3d, nb3d_bool, 3);
+
+    aclTensor * fill_mask = nullptr;
+    switch (ttype) {
+        case GGML_TRI_TYPE_LOWER_DIAG:
+            GGML_CANN_CALL_ACLNN_OP(ctx, LogicalNot, acl_mask_b.get(), acl_mask_inv.get());
+            fill_mask = acl_mask_inv.get();
+            break;
+        case GGML_TRI_TYPE_UPPER_DIAG:
```
hipudding, line 861: use case fallthrough.

ggml/src/ggml-cann/aclnn_ops.cpp
```diff
@@ -1883,31 +2111,79 @@ void ggml_cann_get_rows(ggml_backend_cann_context & ctx, ggml_tensor * dst) {
...
+        acl_tensor_ptr acl_src = ggml_cann_create_tensor(src_ptr, acl_type, type_size, s_ne, s_nb, 2);
+        GGML_CANN_CALL_ACLNN_OP(ctx, InplaceIndexCopy, acl_dst.get(), 0, acl_idx.get(), acl_src.get());
+    };
 
     switch (dst->type) {
         case GGML_TYPE_F32:
-            {
-                aclnn_index_copy_4d(ctx, src0->data, src0->ne, src0->nb, dst->data, dst->ne, dst->nb, src1, dst->type);
-                break;
+            for (int64_t i3 = 0; i3 < dst->ne[3]; i3++) {
```
hipudding: Same above.

ggml/src/ggml-cann/aclnn_ops.cpp
```diff
@@ -3268,29 +3544,62 @@ void ggml_cann_pad_reflect_1d(ggml_backend_cann_context & ctx, ggml_tensor * dst
...
+    // is contiguous with respect to dim2 in both src and dst.
+    GGML_ASSERT(src0->nb[3] == src0->nb[2] * src0->ne[2]);
+    GGML_ASSERT(dst->nb[3]  == dst->nb[2]  * dst->ne[2]);
 
-        acl_tensor_ptr acl_dst =
-            ggml_cann_create_tensor((char *) dst->data + i * src0->ne[3], ggml_cann_type_mapping(dst->type),
-                                    ggml_element_size(dst), dst->ne, dst->nb, 3);
+    int64_t src_ne_3d[3] = { src0->ne[0], src0->ne[1], src0->ne[2] * src0->ne[3] };
+    size_t  src_nb_3d[3];
+    src_nb_3d[0] = src0->nb[0];
```
hipudding: just use src0->nb, not necessary to use a new value.

ggml/src/ggml-cann/ggml-cann.cpp
```diff
@@ -1845,12 +1772,41 @@ static bool ggml_cann_compute_forward(ggml_backend_cann_context & ctx, struct gg
                     GGML_CANN_CALL_OP_UNARY_GATED(Relu);
                     break;
                 case GGML_GLU_OP_GEGLU:
+                    {
+                        // aclnnGeGluV3 only accepts a single interleaved input tensor
+                        // (split internally along dim). When src[1] != nullptr the two
+                        // halves are independent tensors, and swapped!=0 reverses which
+                        // half gets activated — neither case is expressible with the
+                        // fused op, so fall back to the generic two-kernel path.
+                        const int32_t swapped_glu = ggml_get_op_params_i32(dst, 1);
```
hipudding: put this check in ggml_cann_geglu function. same for others.

ggml/src/ggml-cann/aclnn_ops.cpp
```diff
@@ -652,6 +763,166 @@ void ggml_cann_sum(ggml_backend_cann_context & ctx, ggml_tensor * dst) {
...
+    {
+        float          one_val = 1.0f;
+        acl_scalar_ptr acl_one = ggml_cann_create_scalar(&one_val, ACL_FLOAT);
+        GGML_CANN_CALL_ACLNN_OP(ctx, InplaceFillScalar, acl_ones.get(), acl_one.get());
+    }
+
+    ggml_cann_pool_alloc mask_f_alloc(ctx.pool(), buf_sz);
+    void * mask_f_buf = mask_f_alloc.get();
+    acl_tensor_ptr acl_mask_f = ggml_cann_create_tensor(mask_f_buf, ACL_FLOAT, nb_f32, ne3d, nb3d, 3);
+    GGML_CANN_CALL_ACLNN_OP(ctx, Tril, acl_ones.get(), (int64_t)-1, acl_mask_f.get());
```
hipudding: Tril and Triu can be used for this OP.

ggml/src/ggml-cann/aclnn_ops.cpp
hipudding: This OP has optimized, switch case is not used.

noemotiovon:
Thanks for your contribution! Overall it looks fine, but please make sure all precision tests pass.

noemotiovon:
Additionally, it would be better to optimize one operator at a time.

hipudding:
> Additionally, it would be better to optimize one operator at a time.

Thank you for your review. This is because cann backend for qwen3.5 lacked support for several operators. To ensure high-performance inference in qwen3.5, it is necessary to add all the missing and optimizable operators.

hipudding:
> Thanks for your contribution! Overall it looks fine, but please make sure all precision tests pass.

image

KokerZhou:
> **Known issue**: Due to kernel caching, in models like qwen3.5 where each layer is not entirely identical, a "kernel cache not found" error occurs during multi-device inference. Workaround: Disable kernel caching by setting environment variables. ACLNN_CACHE_LIMIT=0

So the proper fix should be to use separate threads for each device so the cache is managed per device and invalidated correctly, rather than disabling kernel caching globally as a workaround?

hipudding:
> > **Known issue**: Due to kernel caching, in models like qwen3.5 where each layer is not entirely identical, a "kernel cache not found" error occurs during multi-device inference. Workaround: Disable kernel caching by setting environment variables. ACLNN_CACHE_LIMIT=0
> 
> So the proper fix should be to use separate threads for each device so the cache is managed per device and invalidated correctly, rather than disabling kernel caching globally as a workaround?

Yes, you're right. But I didn't find a way to manage cache key, do you know any cache key's api to manage the key by llama.cpp? I thought it's a interface only in CANN.

hipudding:
@KokerZhou Sorry, I misunderstood at first. The revised version, using multi-threaded inference (one thread per device), involves many modifications; I think this feature can be considered for later implementation. The best approach would be to add device information to the cache key, but I haven't found a suitable interface for that.

hipudding:
@ggerganov  Friendly ping!  Looking forward to any feedback when you have a chance. Thanks!

KokerZhou:
@hipudding The multithread issue has been fixed upstream in commit `f37e88e`:
https://gitcode.com/cann/opbase/tree/f37e88e9525fac93214acd6fc652d9c69e191e02

No further action is required.

hipudding:
> https://gitcode.com/cann/opbase/tree/f37e88e9525fac93214acd6fc652d9c69e191e02

Wow, Good news, Thanks for your information.

PR #21206, opened 2026-03-31
webui: fix syntax highlighting lost after streaming for non-common languages

hmblair:
## Overview

Code blocks for languages like Haskell, Elixir, Dart, Scala, etc. show
proper syntax highlighting while streaming but lose all color once the
closing fence arrives.

The streaming path uses `highlight.js` directly, which supports 192
languages. The completed-block path uses `rehype-highlight`, which
internally uses `lowlight` — and lowlight's default `common` bundle only
includes 37 languages. For unrecognized languages, lowlight silently
outputs plain unformatted text.

The fix passes lowlight's full language set to `rehype-highlight` so both
rendering paths support the same languages. `lowlight` is already a
transitive dependency, so no new packages are added.

# Requirements

- AI usage disclosure: Yes, AI was used to help diagnose the root cause and draft the PR.

### DIFF
--- a/tools/server/public/index.html
+++ b/tools/server/public/index.html
@@ -18,7 +18,7 @@
 		<div style="display: contents">
 			<script>
 				{
-					__sveltekit_10avopp = {
+					__sveltekit_1ppa22i = {
 						base: new URL('.', location).pathname.slice(0, -1)
 					};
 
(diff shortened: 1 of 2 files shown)


allozaur (changes requested):
Please run `npm run build` and push the updated static files

allozaur (approved):

ServeurpersoCom (approved):

PR #21213, opened 2026-03-31
common : gpt-oss handle builtins and unsolicited tool calls

aldehir:
## Overview

Add parsing rules for builtin functions and unsolicited tool calls. This will prevent the parsing from failing and instead returns empty content since there is no good way to handle this edge case.

ref: https://github.com/ggml-org/llama.cpp/issues/20650#issuecomment-4150001519

## Additional information

Occasionally, gpt-oss will try to invoke a builtin function even if it isn't enabled in the system prompt. An easy test is to enable the builtin python tool and confirm that it returns an empty string:

```json
{
  "messages": [
    {
      "role": "user",
      "content": "Can you run print('hello') in python."
    }
  ],
  "stream": false,
  "chat_template_kwargs": {"builtin_tools": ["python"]}
}
```
```
{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "",
        "reasoning_content": "User asks: \"Can you run print('hello') in python.\" They want to see output. As ChatGPT, we can execute python code using tool. So we should run print('hello') and show output. Provide result."
      }
    }
  ],
  "model": "gpt-oss-120b-F16.gguf",
  "id": "chatcmpl-VtCPiEeOxStFwd66nQiR9YWAXj6nfbOE",
  "__verbose": {
    "content": "<|channel|>analysis<|message|>User asks: \"Can you run print('hello') in python.\" They want to see output. As ChatGPT, we can execute python code using tool. So we should run print('hello') and show output. Provide result.<|end|><|start|>assistant<|channel|>commentary to=python code<|message|>print('hello')",
  }
}
```

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -989,6 +989,10 @@ static common_chat_params common_chat_params_init_gpt_oss(const common_chat_temp
         auto analysis = p.ref("analysis");
         auto preamble = p.rule("preamble", p.literal("<|channel|>commentary<|message|>") + p.content(content) + end);
         auto final_msg = p.rule("final", p.literal("<|channel|>final<|message|>") + p.content(content));
+
+        // Consume any unsolicited tool calls, e.g. builtin functions
+        auto unsolicited = p.rule("unsolicited", p.atomic(p.optional(channel) + p.literal(" to=") + content + end));
+
         auto any = p.rule("any", preamble | analysis);
 
         if (has_response_format) {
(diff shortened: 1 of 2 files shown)


ggerganov (approved):
With this patch, I am not seeing anymore errors in the AIME25 evals.

pwilkin (approved):

PR #16382, opened 2025-10-02
implement context checkpointing for hybrid and recurrent models

ddh0:
This PR generalizes the SWA checkpointing logic (ref #15293) to also create checkpoints for recurrent and hybrid models such as Mamba, Jamba, etc.

- SWA-specific parts of the code are generalized:
  - the `--swa-checkpoints` CLI arg is renamed to `--ctx-checkpoints`
  - the internal `LLAMA_STATE_SEQ_FLAGS_SWA_ONLY` flag is renamed to `LLAMA_STATE_SEQ_FLAGS_CHECKPOINT_ONLY`
- adds `llama_model_is_hybrid` to `llama-model.cpp` and `llama.h`

This removes the need to re-process the entire context in the majority of cases.

Would resolve #15677 and #14625

*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

PR #21221, opened 2026-03-31
kv-cache: add LLAMA_STATE_SEQ_FLAGS_APPEND for incremental state restore

sammshen:
## Overview

`state_read_meta` unconditionally calls `seq_rm(dest_seq_id, -1, -1)` before restoring so every `state_seq_set_data_ext` wipes the sequence first, making chunk by chunk restore impossible (e.g. restoring chunk 1 destroys chunk 0) 

This PR adds a new flag `LLAMA_STATE_SEQ_FLAGS_APPEND` (value `2`) that skips the `seq_rm` call, allowing repeated `state_seq_set_data_ext` calls to append to the same sequence, using the existing `llama_state_seq_flags` mechanism (precedent`LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY` is from #16382).    

Exact changes:                                                                                                                                                                                  
  - `include/llama.h`: add `#define LLAMA_STATE_SEQ_FLAGS_APPEND 2`
  - `src/llama-kv-cache.h`: add `flags` param to `state_read_meta`                                                                                                                          
  - `src/llama-kv-cache.cpp`: thread `flags` through `state_read` → `state_read_meta`, gate `seq_rm` on `!(flags & APPEND)`   

default behavior (`flags=0`) is unchanged.                                                                                                                                                

 **motivation:** this enables external KV cache systems (e.g. [LMCache](https://github.com/LMCache/LMCache)) to store and restore opaque state blobs one chunk at a time, without reassembling into a single monolithic blob or parsing the internal serialization format.
                  

# Requirements

- AI usage disclosure:

### DIFF
--- a/src/llama-kv-cache.h
+++ b/src/llama-kv-cache.h
@@ -280,7 +280,7 @@ private:
     void state_write_meta(llama_io_write_i & io, const cell_ranges_t & cr, llama_seq_id seq_id = -1) const;
     void state_write_data(llama_io_write_i & io, const cell_ranges_t & cr) const;
 
-    bool state_read_meta(llama_io_read_i & io, uint32_t strm, uint32_t cell_count,       slot_info & sinfo, llama_seq_id dest_seq_id = -1);
+    bool state_read_meta(llama_io_read_i & io, uint32_t strm, uint32_t cell_count,       slot_info & sinfo, llama_seq_id dest_seq_id = -1, llama_state_seq_flags flags = 0);
     bool state_read_data(llama_io_read_i & io, uint32_t strm, uint32_t cell_count, const slot_info & sinfo);
 };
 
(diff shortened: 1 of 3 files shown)


src/llama-kv-cache.h
```diff
@@ -280,7 +280,7 @@ class llama_kv_cache : public llama_memory_i {
     void state_write_meta(llama_io_write_i & io, const cell_ranges_t & cr, llama_seq_id seq_id = -1) const;
     void state_write_data(llama_io_write_i & io, const cell_ranges_t & cr) const;
 
-    bool state_read_meta(llama_io_read_i & io, uint32_t strm, uint32_t cell_count,       slot_info & sinfo, llama_seq_id dest_seq_id = -1);
+    bool state_read_meta(llama_io_read_i & io, uint32_t strm, uint32_t cell_count,       slot_info & sinfo, llama_seq_id dest_seq_id = -1, llama_state_seq_flags flags = 0);
```
sammshen, line 283: default no change

PR #20595, opened 2026-03-15
ggml: avoid creating CUDA context during device init

ServeurpersoCom:
*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

ggml_cuda_init() calls cudaSetDevice() on every GPU just to query free VRAM for logging. This triggers the creation of a CUDA primary context (120-550 MB depending on GPU), which is irreversible for the lifetime of the process. Every process that loads the backend pays this cost, even if it never uses the GPU (router mode).

This PR removes cudaSetDevice + cudaMemGetInfo from device init. The log loses the free VRAM part but still shows total VRAM via cudaGetDeviceProperties (no context needed). Free VRAM is queried later by FIT through its own cudaSetDevice path, so the context creation is simply deferred to first real use.

Tested on RTX PRO 6000 Blackwell. Router process not in nvidia-smi anymore. Model loading unchanged.

Fixes #20582

PR #21231, opened 2026-03-31
server: add router device memory margin parameter for dynamic unloading

0cc4m:
## Overview

I have a server running which serves small embedding models and large text models. Currently there's no way to allow multiple small models to coexist in memory, but to unload them when a large model needs space. This PR solves that by adding a `--models-memory-margin` parameter which works similarly to the autofit feature memory margin. The router keeps track of memory requirements per model and uses it to unload models in the same order as previously defined for the max number of models, once that margin is exceeded on any device.

I've been running this on my server and it solves the issue I had.

# Requirements

- AI usage disclosure: YES, Qwen3.5 122B and Claude were used for assistance.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -3237,6 +3237,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_MODELS_MAX"));
     add_opt(common_arg(
+        {"--models-memory-margin"}, "N",
+        string_format("for router server, MiB of memory to leave free, per device (default: %d, 0 = unlimited)", params.models_memory_margin),
+        [](common_params & params, int value) {
+            params.models_memory_margin = value;
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_MODELS_MEMORY_MARGIN"));
+    add_opt(common_arg(
         {"--models-autoload"},
         {"--no-models-autoload"},
         string_format("for router server, whether to automatically load models (default: %s)", params.models_autoload ? "enabled" : "disabled"),
@@ -3486,6 +3493,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.offline = true;
         }
     ).set_examples({LLAMA_EXAMPLE_COMMON, LLAMA_EXAMPLE_DOWNLOAD}).set_env("LLAMA_ARG_OFFLINE"));
+    add_opt(common_arg(
+        {"--measure-only"},
+        "Load the model to measure memory requirements, print to stdout, then exit",
+        [](common_params & params) {
+            params.measure_only = true;
+        }
+    ));
     add_opt(common_arg(
         {"-lv", "--verbosity", "--log-verbosity"}, "N",
         string_format("Set the verbosity threshold. Messages with a higher verbosity will be ignored. Values:\n"
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -3544,6 +3544,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_MODELS_MAX"));
     add_opt(common_arg(
+        {"--models-memory-margin"}, "N",
+        string_format("for router server, MiB of memory to leave free, per device (default: %d, 0 = unlimited)", params.models_memory_margin),
+        [](common_params & params, int value) {
+            params.models_memory_margin = value;
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_MODELS_MEMORY_MARGIN"));
+    add_opt(common_arg(
         {"--models-autoload"},
         {"--no-models-autoload"},
         string_format("for router server, whether to automatically load models (default: %s)", params.models_autoload ? "enabled" : "disabled"),
@@ -3798,6 +3805,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.offline = true;
         }
     ).set_examples({LLAMA_EXAMPLE_COMMON, LLAMA_EXAMPLE_DOWNLOAD, LLAMA_EXAMPLE_TOKENIZE}).set_env("LLAMA_ARG_OFFLINE"));
+    add_opt(common_arg(
+        {"--measure-only"},
+        "Load the model to measure memory requirements, print to stdout, then exit",
+        [](common_params & params) {
+            params.measure_only = true;
+        }
+    ));
     add_opt(common_arg(
         {"-lv", "--verbosity", "--log-verbosity"}, "N",
         string_format("Set the verbosity threshold. Messages with a higher verbosity will be ignored. Values:\n"
--- a/src/llama-ext.h
+++ b/src/llama-ext.h
@@ -90,7 +90,17 @@ LLAMA_API ggml_backend_dev_t llama_model_get_device(const struct llama_model * m
 
 LLAMA_API llama_memory_breakdown llama_get_memory_breakdown(const struct llama_context * ctx);
 
-// Set whether the context outputs nextn embeddings or not
+// Returns the projected memory use (model + context + compute) in bytes
+// for the given device within this context. Returns 0 if the device is not used.
+LLAMA_API uint64_t llama_context_device_memory(
+        const struct llama_context * ctx,
+        ggml_backend_dev_t           device);
+
+//
+// pre-norm embeddings (hidden state before the final output norm)
+//
+
+// Set whether the context outputs pre-norm embeddings or not
 // If masked == true,  output the embeddings only for the tokens with batch.logits != 0
 // If masked == false, output the embeddings for all tokens in the batch regardless of batch.logits
 LLAMA_API void llama_set_embeddings_nextn(struct llama_context * ctx, bool value, bool masked);
--- a/tools/server/server-models.cpp
+++ b/tools/server/server-models.cpp
@@ -145,6 +145,7 @@ static void unset_reserved_args(common_preset & preset, bool unset_model_args) {
     preset.unset_option("LLAMA_API_KEY");
     preset.unset_option("LLAMA_ARG_MODELS_DIR");
     preset.unset_option("LLAMA_ARG_MODELS_MAX");
+    preset.unset_option("LLAMA_ARG_MODELS_MEMORY_MARGIN");
     preset.unset_option("LLAMA_ARG_MODELS_PRESET");
     preset.unset_option("LLAMA_ARG_MODELS_AUTOLOAD");
     if (unset_model_args) {
@@ -260,9 +261,39 @@ server_models::server_models(
         bin_path = get_server_exec_path().string();
     } catch (const std::exception & e) {
         bin_path = argv[0];
-        LOG_WRN("failed to get server executable path: %s\n", e.what());
-        LOG_WRN("using original argv[0] as fallback: %s\n", argv[0]);
+        SRV_WRN("failed to get server executable path: %s\n", e.what());
+        SRV_WRN("using original argv[0] as fallback: %s\n", argv[0]);
     }
+
+    const size_t memory_margin = (size_t) base_params.models_memory_margin * 1024 * 1024;
+
+    if (memory_margin > 0) {
+        ggml_backend_dev_t cpu_dev = ggml_backend_dev_by_type(GGML_BACKEND_DEVICE_TYPE_CPU);
+        ggml_backend_buffer_type_t cpu_buft = cpu_dev ? ggml_backend_dev_buffer_type(cpu_dev) : nullptr;
+
+        const size_t n_devs = ggml_backend_dev_count();
+        for (size_t i = 0; i < n_devs; i++) {
+            ggml_backend_dev_t dev = ggml_backend_dev_get(i);
+            ggml_backend_buffer_type_t dev_buft = ggml_backend_dev_buffer_type(dev);
+            if (dev_buft) {
+                buft_by_name[ggml_backend_buft_name(dev_buft)] = dev_buft;
+            }
+            ggml_backend_buffer_type_t host_buft = ggml_backend_dev_host_buffer_type(dev);
+            if (host_buft && cpu_buft) {
+                buft_by_name[ggml_backend_buft_name(host_buft)] = cpu_buft;
+            }
+
+            size_t free, total;
+            ggml_backend_dev_memory(dev, &free, &total);
+            if (total > 0 && dev_buft) {
+                const size_t available = (free > memory_margin) ? free - memory_margin : 0;
+                bmm_available[dev_buft] = available;
+                SRV_DBG("buft %s: available memory after margin=%zu MiB\n",
+                    ggml_backend_buft_name(dev_buft), available / (1024 * 1024));
+            }
+        }
+    }
+
     load_models();
 }
 
@@ -456,6 +487,7 @@ void server_models::load_models() {
                 /* port          */ 0,
                 /* status        */ SERVER_MODEL_STATUS_UNLOADED,
                 /* last_used     */ 0,
+                /* bmm_req       */ {},
                 /* args          */ std::vector<std::string>(),
                 /* loaded_info   */ {},
                 /* progress      */ {},
@@ -623,6 +655,7 @@ void server_models::load_models() {
                     /* port          */ 0,
                     /* status        */ SERVER_MODEL_STATUS_UNLOADED,
                     /* last_used     */ 0,
+                    /* bmm_req       */ {},
                     /* args          */ std::vector<std::string>(),
                     /* loaded_info   */ {},
                     /* progress      */ {},
@@ -797,30 +830,87 @@ std::vector<server_model_meta> server_models::get_all_meta() {
     return result;
 }
 
-void server_models::unload_lru() {
-    if (base_params.models_max <= 0) {
-        return; // no limit
-    }
-    // remove one of the servers if we passed the models_max (least recently used - LRU)
-    std::string lru_model_name = "";
-    int64_t lru_last_used = ggml_time_ms();
-    size_t count_active = 0;
-    {
-        std::unique_lock<std::mutex> lk(mutex);
-        for (const auto & m : mapping) {
-            if (m.second.meta.is_running()) {
-                count_active++;
-                if (m.second.meta.last_used < lru_last_used) {
-                    lru_model_name = m.first;
-                    lru_last_used = m.second.meta.last_used;
-                }
+int server_models::can_fit(const buft_memory_map & bmm_req) const {
+    buft_memory_map bmm_total;
+    for (const auto & m : mapping) {
+        if (m.second.meta.is_running()) {
+            for (const auto & [buft, mem] : m.second.meta.bmm_req) {
+                bmm_total[buft] += mem;
             }
         }
     }
-    if (!lru_model_name.empty() && count_active >= (size_t)base_params.models_max) {
-        SRV_INF("models_max limit reached, removing LRU name=%s\n", lru_model_name.c_str());
+
+    auto get = [](const buft_memory_map & dmm, ggml_backend_buffer_type_t buft) -> size_t {
+        auto it = dmm.find(buft);
+        return it != dmm.end() ? it->second : 0;
+    };
+
+    int res = 0;
+
+    for (const auto & [buft, limit] : bmm_available) {
+        const size_t mem_total = get(bmm_total, buft);
+        const size_t mem_new   = get(bmm_req,   buft);
+
+        SRV_DBG("buft %s: total=%zu MiB, new=%zu MiB, limit=%zu MiB\n",
+            ggml_backend_buft_name(buft),
+            mem_total / (1024 * 1024), mem_new / (1024 * 1024), limit / (1024 * 1024));
+
+        if (mem_total + mem_new > limit) {
+            res++;
+        }
+    }
+
+    return res;
+}
+
+bool server_models::limits_exceeded(const buft_memory_map & bmm_req) const {
+    const bool check_active = base_params.models_max > 0;
+    const bool check_memory = base_params.models_memory_margin > 0;
+
+    if (!check_active && !check_memory) {
+        return false;
+    }
+
+    int count_active = 0;
+    for (const auto & m : mapping) {
+        if (m.second.meta.is_running()) {
+            count_active++;
+        }
+    }
+
+    const bool active_exceeded = check_active && count_active >= base_params.models_max;
+    const bool memory_exceeded = check_memory && can_fit(bmm_req) > 0;
+
+    return active_exceeded || memory_exceeded;
+}
+
+void server_models::unload_lru(const buft_memory_map & bmm_req) {
+    if (base_params.models_memory_margin > 0) {
+        GGML_ASSERT(!bmm_available.empty());
+    }
+
+    while (true) {
+        std::string lru_model_name;
+        {
+            std::unique_lock<std::mutex> lk(mutex);
+            if (!limits_exceeded(bmm_req)) {
+                break;
+            }
+            int64_t lru_last_used = ggml_time_ms();
+            for (const auto & m : mapping) {
+                if (m.second.meta.is_running() && m.second.meta.last_used < lru_last_used) {
+                    lru_model_name = m.first;
+                    lru_last_used  = m.second.meta.last_used;
+                }
+            }
+        }
+
+        if (lru_model_name.empty()) {
+            break;
+        }
+
+        SRV_INF("limits exceeded, removing LRU name=%s\n", lru_model_name.c_str());
         unload(lru_model_name);
-        // wait for unload to complete
         {
             std::unique_lock<std::mutex> lk(mutex);
             cv.wait(lk, [this, &lru_model_name]() {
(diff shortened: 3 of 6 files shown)


tools/server/server-models.cpp
```diff
@@ -493,44 +494,111 @@ std::vector<server_model_meta> server_models::get_all_meta() {
...
+                    return mapping[lru_model_name].meta.status == SERVER_MODEL_STATUS_UNLOADED;
+                });
+            }
+        } else {
+            break;
         }
     }
 }
 
+static uint64_t get_model_memory_mb(const common_preset& preset) {
```
ngxson: IIRC this is mostly the same logic with `-fit`, right? If so, is it possible to merge them into a new function called `common_fit_get_model_memory()`

Btw, I think it can be useful to return more details about mem usage, for example: mem usage by context and by weight. In the future, we can also return usage per backend.

common/arg.cpp
```diff
@@ -3042,6 +3042,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.models_max = value;
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_MODELS_MAX"));
+    add_opt(common_arg(
+        {"--models-memory-max"}, "N",
+        string_format("for router server, maximum memory usage in MB (default: %d, 0 = unlimited)", params.models_memory_max),
+        [](common_params & params, int value) {
+            params.models_memory_max = value;
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_MODELS_MEMORY_MAX"));
```
ngxson: UX-wise, I would prefer specifying the reverse like what `--fit-target` does. Instead of specifying max used memory, it's more intuitive to specify margin. But I'm not quite sure how hard it is to implement such logic here

0cc4m: Good idea, this router feature can likely share both the memory calculation and the defaults with the autofit code.

include/llama.h
```diff
@@ -1536,6 +1536,12 @@ extern "C" {
     // print a breakdown of per-device memory use via LLAMA_LOG:
     LLAMA_API void llama_memory_breakdown_print(const struct llama_context * ctx);
 
+    // Returns the projected memory use (model + context + compute) in bytes
+    // for the given device within this context. Returns 0 if the device is not used.
+    LLAMA_API uint64_t llama_context_device_memory(
+            const struct llama_context * ctx,
+            ggml_backend_dev_t           device);
+
```
ggerganov: Most likely the device querying API should be more generic, so this signature is likely to be obsoleted. Let's move to `llama-ext.h` for now.

tools/server/server-models.h
```diff
@@ -107,14 +110,20 @@ struct server_models {
...
 
     // unload least recently used models if the limit is reached
-    void unload_lru();
+    void unload_lru(const model_memory_map& new_model_memory_per_device);
 
     // not thread-safe, caller must hold mutex
     void add_model(server_model_meta && meta);
 
+    // not thread-safe, caller must hold mutex
+    uint64_t get_memory_exceeded(const model_memory_map& new_model_memory_per_device) const;
```
ggerganov:
```suggestion
    uint64_t get_memory_exceeded(const model_memory_map & new_model_memory_per_device) const;
```

tools/server/server-models.h
```diff
@@ -107,14 +110,20 @@ struct server_models {
...
     common_preset base_preset; // base preset from llama-server CLI args
 
+    // available memory per device
+    std::map<ggml_backend_dev_t, uint64_t> memory_per_device;
+
     void update_meta(const std::string & name, const server_model_meta & meta);
 
     // unload least recently used models if the limit is reached
-    void unload_lru();
+    void unload_lru(const model_memory_map& new_model_memory_per_device);
```
ggerganov:
```suggestion
    void unload_lru(const model_memory_map & new_model_memory_per_device);
```

common/arg.cpp
```diff
@@ -3044,6 +3044,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.models_max = value;
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_MODELS_MAX"));
+    add_opt(common_arg(
+        {"--models-memory-margin"}, "N",
+        string_format("for router server, MB of memory to leave free, per device (default: %d, 0 = unlimited)", params.models_memory_margin),
```
ggerganov:
```suggestion
        string_format("for router server, MiB of memory to leave free, per device (default: %d, 0 = unlimited)", params.models_memory_margin),
```

tools/server/server-models.cpp
```diff
@@ -178,6 +180,25 @@ server_models::server_models(
...
+    if (memory_margin > 0) {
+        const size_t n_devs = ggml_backend_dev_count();
+        for (size_t i = 0; i < n_devs; i++) {
+            ggml_backend_dev_t dev = ggml_backend_dev_get(i);
+            size_t free, total;
+            ggml_backend_dev_memory(dev, &free, &total);
+            if (total > 0) {
+                const uint64_t available = (free > memory_margin) ? free - memory_margin : 0;
+                memory_per_device[dev] = available;
+                SRV_DBG("device %s: available memory after margin=%lu MB\n",
```
ggerganov:
```suggestion
                SRV_DBG("device %s: available memory after margin=%lu MiB\n",
```

tools/server/server-models.cpp
```diff
@@ -493,44 +515,159 @@ std::vector<server_model_meta> server_models::get_all_meta() {
     return result;
 }
 
-void server_models::unload_lru() {
-    if (base_params.models_max <= 0) {
+uint64_t server_models::get_memory_exceeded(const model_memory_map& new_model_memory_per_device) const {
+    model_memory_map total_memory_per_device;
+    for (const auto & m : mapping) {
+        if (m.second.meta.is_running()) {
+            for (const auto& [key, value] : m.second.meta.memory_per_device) {
```
ggerganov:
```suggestion
            for (const auto & [key, value] : m.second.meta.memory_per_device) {
```

tools/server/server-models.cpp
```diff
@@ -493,44 +515,159 @@ std::vector<server_model_meta> server_models::get_all_meta() {
...
+    }
+
+    auto get = [](const model_memory_map & m, ggml_backend_dev_t k) {
+        auto it = m.find(k);
+        return it != m.end() ? it->second : 0;
+    };
+
+    uint64_t memory_exceeded = 0;
+
+    for (const auto& [key, limit] : memory_per_device) {
```
ggerganov:
```suggestion
    for (const auto & [key, limit] : memory_per_device) {
```

tools/server/server-models.cpp
```diff
@@ -493,44 +515,159 @@ std::vector<server_model_meta> server_models::get_all_meta() {
     return result;
 }
 
-void server_models::unload_lru() {
-    if (base_params.models_max <= 0) {
+uint64_t server_models::get_memory_exceeded(const model_memory_map& new_model_memory_per_device) const {
```
ggerganov:
```suggestion
uint64_t server_models::get_memory_exceeded(const model_memory_map & new_model_memory_per_device) const {
```

tools/server/server-models.h
```diff
@@ -107,14 +110,20 @@ struct server_models {
     std::vector<std::string> base_env;
     common_preset base_preset; // base preset from llama-server CLI args
 
+    // available memory per device
+    std::map<ggml_backend_dev_t, uint64_t> memory_per_device;
```
ggerganov:
```suggestion
    model_memory_map memory_per_device;
```

tools/server/server-models.cpp
```diff
@@ -493,44 +515,159 @@ std::vector<server_model_meta> server_models::get_all_meta() {
...
+            }
+        }
+    }
+
+    auto get = [](const model_memory_map & m, ggml_backend_dev_t k) {
+        auto it = m.find(k);
+        return it != m.end() ? it->second : 0;
+    };
+
+    uint64_t memory_exceeded = 0;
```
ggerganov: Let's keep all memory sizes in `size_t` type for consistency.

tools/server/server-models.h
```diff
@@ -62,6 +64,7 @@ struct server_model_meta {
     int port = 0;
     server_model_status status = SERVER_MODEL_STATUS_UNLOADED;
     int64_t last_used = 0; // for LRU unloading
+    model_memory_map memory_per_device; // projected bytes per device
```
ggerganov: I am still not sure why we have to keep a memory map both in `server_model_meta` and `server_models`. Isn't one map in `struct server_models` going to be enough?

0cc4m: The idea here was to create a map of how much memory is available within the margin for each device when the server is started. That's what is stored in the `server_models` struct. It should at least get a clearer name.

`server_model_meta` then stores the requirement per device for that model. That way it can be summed up and compared to the values in `server_models`.

ggerganov: I see. Yes, better names are needed.

ggerganov:
I'll take a look next week.

tools/server/server-models.cpp
```diff
@@ -494,44 +516,168 @@ std::vector<server_model_meta> server_models::get_all_meta() {
...
-            });
+}
+
+static device_memory_map get_model_memory_per_device(const common_preset & preset) {
+    common_params params;
+    preset.apply_to_params(params);
+
+    if (params.model.path.empty()) {
+        return {};
+    }
```
ggerganov, line 612: @0cc4m This check prevents using the functionality for models downloaded with `-hf` because they don't have a `path`. Should be fixed before merging.

0cc4m: You're right, I hadn't considered that. Currently it will let it through, but estimate it as having no size, which is not ideal. I could fill the map on second load, but that risks an OOM when it's loading for the first time. I could try to estimate with file size before downloading, but that's unreliable and hard to map to devices. Or I could trigger the download before actually loading with `common_download_model`, estimate and then load. Not sure if that would cause issues in other places, I'd have to try it. What do you think?

ggerganov: Hm not sure. Seems complicated.

Maybe the right way is to have a `--download-only` CLI argument. If set, the `llama-server` (and any other tool) would just download the model and exit.

This way, we can split the model loading in stages:

- `llama-server ... --download-only`
- Perform memory size checks
- `llama-server ... --offline`

0cc4m: I implemented this, please take a look when you find some time.

MGAndreasen:
This is an interesting thing, I use both small and large models as well, on different systems.
if I set my max models to something like 4, and loads a relative big model, and then 3 smaller. then larger model might get evicted partial from vram, and there for gets very slow.
if i then unloads all but the large model, i will be looking at a vram consumption of only a few GB, the rest of the layers have been moved to system memory, but they dont come back into gpu automaticaly.  at least this way it get completely unloaded.

0cc4m:
@ggerganov This is still waiting for your review.

src/llama-ext.h
```diff
@@ -88,3 +88,9 @@ LLAMA_API int32_t llama_model_n_devices(const struct llama_model * model);
 LLAMA_API ggml_backend_dev_t llama_model_get_device(const struct llama_model * model, int i);
 
 LLAMA_API llama_memory_breakdown llama_get_memory_breakdown(const struct llama_context * ctx);
+
+// Returns the projected memory use (model + context + compute) in bytes
+// for the given device within this context. Returns 0 if the device is not used.
+LLAMA_API uint64_t llama_context_device_memory(
+        const struct llama_context * ctx,
+        ggml_backend_dev_t           device);
```
ggerganov, line 97: Instead of adding this new function, can you reuse the `llama_get_memory_breakdown` that we added recently (https://github.com/ggml-org/llama.cpp/pull/22171)?

tools/server/server-models.cpp
```diff
@@ -497,49 +519,287 @@ std::vector<server_model_meta> server_models::get_all_meta() {
...
+                device_memory_map mem;
+                if (base_params.models_memory_margin > 0) {
+                    std::lock_guard<std::mutex> lk(mutex);
+                    auto & meta = mapping[name].meta;
+                    meta.dmm_req = get_model_memory_per_device(meta.preset);
+                    if (meta.dmm_req.empty()) {
+                        SRV_WRN("failed to estimate memory for model %s, memory limits will not apply\n", name.c_str());
+                    }
+                    mem = meta.dmm_req;
+                }
```
ggerganov: Should this chunk of code be moved at the start of `_load()` so you can deduplicate it here?

tools/server/server-models.cpp
```diff
@@ -497,49 +519,287 @@ std::vector<server_model_meta> server_models::get_all_meta() {
...
+                try {
+                    _load(name, mem);
+                } catch (const std::exception & e) {
+                    SRV_ERR("failed to load model %s after download: %s\n", name.c_str(), e.what());
+                    update_status(name, SERVER_MODEL_STATUS_UNLOADED, 1);
+                }
+            }).detach();
```
ggerganov: In general, I consider detaching threads an anti-pattern. Add a TODO to keep track of the threads in `server_models` and join them periodically (maybe on each `load()` and/or `update_status()`).

tools/server/server-models.cpp
```diff
@@ -497,49 +519,287 @@ std::vector<server_model_meta> server_models::get_all_meta() {
...
+        common_download_opts opts;
+        opts.offline = true;
+        auto result = common_download_model(params.model, opts);
+        return result.model_path;
+    }
+
+    return "";
+}
+
+static device_memory_map get_model_memory_per_device(const common_preset & preset) {
```
ngxson: Re-thinking about this, I think it's better to make a new mode on `llama-server` that simply measure the memory usage then exit (without loading model weights). Probably adding a new env variable `LLAMA_SERVER_MEASURE_ONLY=1` when launching the child process.

The problem with calling `llama_model_load_from_file` here is that a broken GGUF can trigger GGML_ASSERT which crash the whole router process.

0cc4m:
It is now using the existing `llama_get_memory_breakdown` to map buft to memory requirements instead of devices, and also using the "thread manages subprocess" pattern for measuring memory requirements and for downloads.

I'm running this on my Spark to test it now, one issue to look into is host buffer types:
```
estimate_mod: unknown buft name 'CUDA_Host' from measure child for model name=Qwen3.5-122B-A10B
```

0cc4m:
I added a map from name to buft and mapped device host buffer types to the CPU buft, that should fix it. It's working in my tests, one concern is that the memory estimation output is printed to the log from the subprocess, which is not ideal. I can try to cover that up.

ggerganov:
@0cc4m Could you rebase the branch again on latest `master`? Thanks.

0cc4m:
Done

tools/server/server.cpp
```diff
@@ -118,6 +125,47 @@ int llama_server(int argc, char ** argv) {
     // struct that contains llama context and inference
     server_context ctx_server;
 
+    llama_backend_init();
```
danbev: I think the `llama_backend_init` and `llama_luma_init` a bit [further down](https://github.com/ggml-org/llama.cpp/blob/6adf9643ee32666504f46c824b21bf824b65cd9b/tools/server/server.cpp#L128-L129) could be removed now.

ServeurpersoCom:
I'm testing the PR on my setup.
A CUDA context is initialized by the router, which eats up 550MB of VRAM even though the parent never loads a model. It comes from models_memory_margin defaulting to 1024.

Sans titre

ServeurpersoCom:
Maybe probe VRAM in the measure child, not the router parent ?

0cc4m:
> Maybe probe VRAM in the measure child, not the router parent ?

I guess that's necessary then, I wasn't aware CUDA had to initialize so much just to read memory state. Is there no way around that?

ServeurpersoCom:
> I guess that's necessary then, I wasn't aware CUDA had to initialize so much just to read memory state. Is there no way around that?

Yeah, I hit this exact wall twice already #20595, #23137: once anything touches a CUDA device the primary context is materialized for the process lifetime, and cudaMemGetInfo needs a context, so reading free VRAM from the backend always pays it.
The measure child already creates a context to estimate the model, so it can probe VRAM there too at no extra cost and keep the router parent off the GPU entirely.
Only genuinely context-free read is NVML (driver-side, like nvidia-smi), but that's a GGML change in ggml_backend_cuda_device_get_memory, not a server one. Separate PR, out of scope here.

ServeurpersoCom:
~~So I'd lean towards keeping the probe in the child for this PR. NVML would be the only context-free option and I'm not sure it fits GGML's generic backend API anyway.~~ (I'm not sure)

Worth noting Vulkan and Metal already read this context-free (physical device / MTLDevice props), CUDA is the odd one out since cudaMemGetInfo needs a context. NVML would bring it in line, and then the router could just read VRAM directly with no child probe needed....

0cc4m:
I can look into it, but does ROCm have an equivalent?

ServeurpersoCom:
That's exactly what I'm wondering, because I've already done a quick POC with Opus, under Linux and Windows.
ggml-cuda.cu is shared with HIP/ROCm I check....

ServeurpersoCom:
ROCm has the equivalent via rocm-smi-lib (rsmi_dev_memory_*) or the newer amdsmi, querying the driver without a HIP context the same way! 


I'll leave it to you, and I can test it under Linux and Windows if needed. It's the cleanest way to definitively close this issue.
```
read GGML  : free 95970 MiB  total 97887 MiB
nvidia-smi : free 95971 MiB  total 97887 MiB
```

Here is my tested POC :
```
root@pod:/mnt/workspace/llama.cpp# git diff
index e7ac73a55..eee4a6e3b 100644
@@ -4959,23 +4959,112 @@ static bool ggml_backend_cuda_get_available_uma_memory(long * available_memory_k
 }
 #endif // defined(__linux__)

+// NVML reads free VRAM driver side the way nvidia-smi does, without the
+// cudaSetDevice plus cudaMemGetInfo pair that materializes a CUDA primary context
+// for the lifetime of the process. ggml-cuda.cu is shared with the HIP and MUSA
+// builds, so this genuine CUDA only path is gated out of those backends
+#if !defined(GGML_USE_HIP) && !defined(GGML_USE_MUSA)
+#define GGML_CUDA_NVML
+
+#if defined(_WIN32)
+#include <windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
+// minimal NVML surface resolved at runtime, no link dependency and no nvml.h
+struct ggml_cuda_nvml_loader {
+    typedef int (*init_fn)(void);
+    typedef int (*handle_fn)(const char *, void **);
+    typedef int (*mem_fn)(void *, void *);
+
+    handle_fn handle = nullptr;
+    mem_fn    mem    = nullptr;
+    bool      ready  = false;
+
+    ggml_cuda_nvml_loader() {
+#if defined(_WIN32)
+        // nvml.dll ships with the driver and lives in System32
+        HMODULE lib = LoadLibraryA("nvml.dll");
+        if (lib == nullptr) {
+            return;
+        }
+        init_fn init = (init_fn) GetProcAddress(lib, "nvmlInit_v2");
+        handle = (handle_fn) GetProcAddress(lib, "nvmlDeviceGetHandleByPciBusId_v2");
+        mem    = (mem_fn)    GetProcAddress(lib, "nvmlDeviceGetMemoryInfo");
+#else
+        void * lib = dlopen("libnvidia-ml.so.1", RTLD_NOW | RTLD_LOCAL);
+        if (lib == nullptr) {
+            lib = dlopen("libnvidia-ml.so", RTLD_NOW | RTLD_LOCAL);
+        }
+        if (lib == nullptr) {
+            return;
+        }
+        init_fn init = (init_fn) dlsym(lib, "nvmlInit_v2");
+        handle = (handle_fn) dlsym(lib, "nvmlDeviceGetHandleByPciBusId_v2");
+        mem    = (mem_fn)    dlsym(lib, "nvmlDeviceGetMemoryInfo");
+#endif
+        if (init == nullptr || handle == nullptr || mem == nullptr) {
+            return;
+        }
+        // NVML_SUCCESS is 0
+        if (init() != 0) {
+            return;
+        }
+        ready = true;
+    }
+};
+
+// fills free and total for the device matching pci_bus_id, returns false when NVML
+// is unavailable so the caller can fall back to the CUDA path
+static bool ggml_cuda_nvml_get_memory(const std::string & pci_bus_id, size_t * free, size_t * total) {
+    static ggml_cuda_nvml_loader nvml;
+    if (!nvml.ready || pci_bus_id.empty()) {
+        return false;
+    }
+    void * dev = nullptr;
+    if (nvml.handle(pci_bus_id.c_str(), &dev) != 0) {
+        return false;
+    }
+    // matches the layout of nvmlMemory_t : total, free, used in bytes
+    struct { unsigned long long total; unsigned long long free; unsigned long long used; } info = {};
+    if (nvml.mem(dev, &info) != 0) {
+        return false;
+    }
+    *total = (size_t) info.total;
+    *free  = (size_t) info.free;
+    return true;
+}
+#endif // !defined(GGML_USE_HIP) && !defined(GGML_USE_MUSA)
+
 static void ggml_backend_cuda_device_get_memory(ggml_backend_dev_t dev, size_t * free, size_t * total) {
     ggml_backend_cuda_device_context * ctx = (ggml_backend_cuda_device_context *)dev->context;
-    ggml_cuda_set_device(ctx->device);
-    CUDA_CHECK(cudaMemGetInfo(free, total));

-// ref: https://github.com/ggml-org/llama.cpp/pull/17368
 #if defined(__linux__)
-    // Check if this is a UMA (Unified Memory Architecture) system
+    // detect UMA up front, cudaGetDeviceProperties does not materialize a primary context
     cudaDeviceProp prop;
     CUDA_CHECK(cudaGetDeviceProperties(&prop, ctx->device));
-
-    // Check if UMA is explicitly enabled via environment variable
     bool uma_env = getenv("GGML_CUDA_ENABLE_UNIFIED_MEMORY") != nullptr;
     bool is_uma = prop.integrated > 0 || uma_env;
+#elif defined(GGML_CUDA_NVML)
+    bool is_uma = false;
+#endif
+
+#if defined(GGML_CUDA_NVML)
+    // discrete GPU context free read through NVML, addressed by PCI bus id so it tracks
+    // the CUDA device regardless of enumeration order. UMA keeps the cudaMemGetInfo path
+    if (!is_uma && ggml_cuda_nvml_get_memory(ctx->pci_bus_id, free, total)) {
+        return;
+    }
+#endif // defined(GGML_CUDA_NVML)

+    ggml_cuda_set_device(ctx->device);
+    CUDA_CHECK(cudaMemGetInfo(free, total));
+
+// ref: https://github.com/ggml-org/llama.cpp/pull/17368
+#if defined(__linux__)
     if (is_uma) {
-        // For UMA systems (like DGX Spark), use system memory info
+        // for UMA systems like DGX Spark, use system memory info
         long available_memory_kb = 0;
         long free_swap_kb = 0;

root@pod:/mnt/workspace/llama.cpp#
```

ORippler:
https://github.com/ggml-org/llama.cpp/pull/23604#issuecomment-4548032543 cross-posting my thoughts here for viz.

0cc4m:
The CUDA memory issue should be resolved now.

tools/server/server.cpp
```diff
@@ -120,6 +127,44 @@ int llama_server(int argc, char ** argv) {
...
+
+        common_log_pause(common_log_main());
+        for (const auto & [buft, data] : llama_get_memory_breakdown(ctx.get())) {
+            size_t total = data.total();
+            if (total > 0) {
+                fprintf(stdout, "measure:%s %zu\n", ggml_backend_buft_name(buft), total);
+            }
+        }
+        fflush(stdout);
+        common_log_resume(common_log_main());
```
ggerganov, line 201: Wrap this logic in a dedicated function in `common/fit.h/.cpp` to avoid including `llama-ext.h` here.

Also, probably need to have more unique string prefix than `measure:`. For example something like: `"__TAG_FIT_MEMORY_TOTAL__:%s %zu\n"`

0cc4m:
The CUDA issue is not yet resolved, it caused some unforeseen issues. (#24715) Maybe there is no good way but to read the GPU memory state from a different process as well? Any other ideas @ggerganov @JohannesGaessler ?

I'm not sure if there's a safe way to reset the CUDA context, and it'd likely still have to be within a backend context if we go that way (so backend init -> memory read -> backend free, for the router case), which is unnecessary overhead for any other backend, though maybe not terribly much?

ngxson:
FYI @0cc4m , this PR could be a bit related to your works here: https://github.com/ggml-org/llama.cpp/pull/24821

I'm planning to make it such that the router code stays as clean as possible. With the recent additions of model management API, downloading thread management is added to the router itself and turns out it's not a good idea. The code complexity blows up too much.

My new plan would be that: anything model-related will be handled in child process, including:
- Downloading model
- Memory estimation (your current PR)

The way it works is: router spawns a child process with a special flag, child instance calculates memory requirements (--fit), then exit without loading the model

So I will go back to this after I finish https://github.com/ggml-org/llama.cpp/issues/24822 ; no actions are needed from your side now, but I will ping you when I get there

JohannesGaessler:
@0cc4m sorry for the late reply, my opinion is that the issues with the previous solution for CUDA are in principle fixable. However, I think that I underestimated the potential for unforeseen and difficult to debug issues. 

In general, I think that we do not actually need a 100% accurate estimate of the memory use, some amount of inprecision if fine as long as it's sufficiently small vs. the safety margin that we are already using anyways. How about this: in the ggml backend device interface it is possible to retrieve PCI bus IDs. In the llama.cpp user code, if both CUDA and Vulkan are available, fetch free memory for "Vulkan devices" first and use those values for "other devices" with the same PCI bus IDs (+ maybe some rough estimate of how much memory the CUDA context would need). This would not fix persistent memory use after unloading a model but it would fix CUDA stealing VRAM from other backends.

0cc4m:
That would just point back at the NVML memory query implementation. But the problem with that is that CUDA uses unusually much context memory compared to other backends, so the "underestimation" through these is worse, unusably so judging by the feedback we got.

ngxson:
I will have a look on new changes here this week

ngxson:
ok so I just done a review pass on this PR and roughly understand what it does:
- when a model is load(), the model tells router how much memory it's using
- router do bookkeeping based on this info, keeping track of memory occupied by loaded/active models
- when a new model is load(), if we don't have enough mem, router unload() active models until it has enough mem again

that's actually a cleaner approach than my initial proposal of having every model to report their mem usage (no matter if they are loaded). however, I still want to see if we can push it further:

as explained in the issue, it can be useful for downstream (like for example pi-agent) to know exactly the context length that it can have if it starts a given model, and this number can changed dynamically based on how much system memory is available. a quick check around the code base suggests that it could be possible to use `llama_get_memory_breakdown` twice to get a breakdown of memory per token, so I will see if it's possible to do this way

also, I will see if it's possible to directly use `llama_get_memory_breakdown` as Georgi suggested in https://github.com/ggml-org/llama.cpp/pull/21231#discussion_r3179609465

I will push my changes directly to this PR

0cc4m:
Why not just do that in a follow-up PR?

ngxson:
my suggestion and the current PR are too overlapped so it's better to be done here, rather than a follow-up

for example, if memory usage is measured once at server boot up instead of per-instance launch, then the `--measure-only` mode will be handled differently

will see if I can focus on this later this week. I already had the details in mind, will not take too long to implement all the changes

ngxson:
I'll go back to this right after https://github.com/ggml-org/llama.cpp/pull/28555

0cc4m:
@ngxson Any ETA for that?

PR #21238, opened 2026-03-31
CUDA/HIP: Fix kernel slection for mmvq mmid kernel to align host selection with device launch bounds

IMbackK:
The conditions `cc == GGML_CUDA_CC_VOLTA || cc >= GGML_CUDA_CC_ADA_LOVELACE` and `cc >= GGML_CUDA_CC_TURING` match all non-nvidia devices. This causes us to attempt to launch the kernel for batch sizes with larger configurations than our launch bounds on HIP devices. This pr fixes the conditionals in get_mmvq_mmid_max_batch.

Fixes #21191

# Requirements

- AI usage disclosure: NO

@ngxson we should remove the `disclosure` from the template "AI usage disclosure: NO" do sent read as what is intended here.


CISC (approved):

am17an (approved):

JohannesGaessler (commented):
Thanks for the fix.

PR #20251, opened 2026-03-08
PEG parser for LFM2

pwilkin:
Dedicated parser for LFM2, this is the only model using Python-style tools for now, so I won't integrate it into the autoparser yet until more templates in this style show up.

Fixes #20245

PR #21242, opened 2026-03-31
fix: tool call parsing for LFM2 and LFM2.5 models

jbuchananr:
## Overview

Currently, LFM2 & LFM2.5 tool calling is broken in `llama.cpp`, issue  https://github.com/ggml-org/llama.cpp/issues/20245,  this commit https://github.com/ggml-org/llama.cpp/pull/20251 introduced a dedicated parser for LFM2; however, LFM2.5 and LFM2 have different tool calling jinja templates

This PR fixes the tool calling parser to catch the expected formats for both cases
 - LFM2: tool list as List of tools: <|tool_list_start|>[...]<|tool_list_end|>, tool calls as
  <|tool_call_start|>[name(arg="val")]<|tool_call_end|>
 - LFM2.5: tool list as List of tools: [...], tool calls as bare [name(arg="val")] with no wrapper tokens

Added `common_chat_params_init_lfm2_5`
### Testing
- Added unit test for lfm2.5 in test-chat.cpp, lfm2 was already covered
- Tested tool calling use case locally with both `LFM2.5-1.2B-Instruct-BF16.gguf` and ` LFM2-8B-A1B-Q4_0.gguf` 

# Requirements

- AI usage disclosure: 
- Used claude code for assistance in tracing code, assistance in where to make fix, generating local test scripts

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1274,19 +1274,93 @@ static common_chat_params common_chat_params_init_kimi_k2(const common_chat_temp
     return data;
 }
 
-// LFM2 / LFM2.5 format:
-// - Reasoning: <think>{reasoning}</think> (optional, only if enable_thinking is true)
+// LFM2 format: uses <|tool_list_start|>[...]<|tool_list_end|> in system prompt
+// and <|tool_call_start|>[name(arg="val")]<|tool_call_end|> for tool calls.
+// - Reasoning: <think>{reasoning}</think> (optional)
 // - Content: text before a tool call (optional)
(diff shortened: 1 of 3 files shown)


aldehir:
Thanks for the PR. Can you add tests for these models?

pwilkin (approved):
Thanks for this, I've been wondering what the deal was :)

jbuchananr:
Happy to help! It's about time tool calling works for LFM 2.5 models 😄  

@aldehir added unit tests for lfm2.5, lfm2 was already covered. Let me know if anything else is needed

aldehir (approved):

PR #21257, opened 2026-04-01
unicode : add custom Qwen2 regex handler to fix segfault on long input

nhs000:
Closes: https://github.com/ggml-org/llama.cpp/issues/21113

## Overview

`std::regex` uses recursive backtracking internally, which causes a stack overflow (segfault) when tokenizing long sequences of repeated characters (e.g. 43K 'A's). The Qwen2 tokenizer regex differs from Llama3 only in the digit pattern `(\p{N}` vs `\p{N}{1,3})`, so it was falling through to the `std::regex fallback` path instead of using a custom handler.

Add `unicode_regex_split_custom_qwen2()` following the established pattern used by gpt2, llama3, kimi_k2, and afmoe custom handlers.

**Note**: Any model whose regex lacks a custom handler is still vulnerable to the same `std::regex` stack overflow on long inputs. A longer-term fix would be replacing `std::regex` with a non-backtracking engine (e.g. RE2). See the TODO comment added in the code.

# Requirements
- AI usage disclosure: Use the Opus 4.6 model to discuss the a general solution but couldn't find a better one.

### DIFF
--- a/src/unicode.cpp
+++ b/src/unicode.cpp
@@ -1134,10 +1134,6 @@ std::vector<std::string> unicode_regex_split(const std::string & text, const std
         }
 
         // fallback to general-purpose std::regex / std::wregex
-        // TODO: std::regex uses recursive backtracking which can cause stack overflow (segfault)
-        //       on long inputs (e.g. 43K repeated characters). Any regex that lacks a custom
-        //       handler above is vulnerable. Consider replacing with a non-backtracking engine
-        //       such as RE2. ref: https://github.com/ggml-org/llama.cpp/issues/21113
         try {
             // if a unicode category is used in the regex, we use the collapsed text and replace the unicode category
             // with the corresponding collapsed representation
--- a/src/unicode.cpp
+++ b/src/unicode.cpp
@@ -470,6 +470,141 @@ static std::vector<size_t> unicode_regex_split_custom_llama3(const std::string &
     return bpe_offsets;
 }
 
+// Qwen2 system regex: "(?:'[sS]|'[tT]|'[rR][eE]|'[vV][eE]|'[mM]|'[lL][lL]|'[dD])|[^\\r\\n\\p{L}\\p{N}]?\\p{L}+|\\p{N}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+"
+static std::vector<size_t> unicode_regex_split_custom_qwen2(const std::string & text, const std::vector<size_t> & offsets) {
+    std::vector<size_t> bpe_offsets; // store the offset of each word
+    bpe_offsets.reserve(offsets.size()); // Reserve memory for the approximate size
+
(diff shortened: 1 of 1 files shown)


src/unicode.cpp
```diff
@@ -997,6 +1134,10 @@ std::vector<std::string> unicode_regex_split(const std::string & text, const std
         }
 
         // fallback to general-purpose std::regex / std::wregex
+        // TODO: std::regex uses recursive backtracking which can cause stack overflow (segfault)
+        //       on long inputs (e.g. 43K repeated characters). Any regex that lacks a custom
+        //       handler above is vulnerable. Consider replacing with a non-backtracking engine
+        //       such as RE2. ref: https://github.com/ggml-org/llama.cpp/issues/21113
```
aldehir:
```suggestion
```

src/unicode.cpp
```diff
@@ -470,6 +470,141 @@ static std::vector<size_t> unicode_regex_split_custom_llama3(const std::string &
     return bpe_offsets;
 }
 
+// Qwen2 system regex: "(?:'[sS]|'[tT]|'[rR][eE]|'[vV][eE]|'[mM]|'[lL][lL]|'[dD])|[^\\r\\n\\p{L}\\p{N}]?\\p{L}+|\\p{N}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+"
```
aldehir:
```suggestion
// Qwen2 system regex: "(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\\r\\n\\p{L}\\p{N}]?[\\p{L}\\p{M}]+|\\p{N}| ?[^\\s\\p{L}\\p{M}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+"
```

aldehir (approved):
Looks good. Passes my comparisons tests against PCRE2 on the Wikitext large dataset.

src/unicode.cpp
```diff
@@ -470,6 +470,141 @@ static std::vector<size_t> unicode_regex_split_custom_llama3(const std::string &
     return bpe_offsets;
 }
 
+// Qwen2 system regex: "(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\\r\\n\\p{L}\\p{N}]?[\\p{L}\\p{M}]+|\\p{N}| ?[^\\s\\p{L}\\p{M}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+"
```
aldehir:
```suggestion
// Qwen2 system regex: "(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\\r\\n\\p{L}\\p{N}]?\\p{L}+|\\p{N}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+"
```

ggerganov (approved):

PR #21259, opened 2026-04-01
kleidiai: provide KleidiAI-Enabled Arm Release Artifact

martin-klacer-arm:
## Overview

This PR adds a KleidiAI-enabled MacOS Arm release artifact definition to the release.yml workflow.

## Additional information

The PR updates the existing MacOS jobs in the release.yml file in an attempt for the KleidiAI-enabled addition to be concise and in line with the rest of the file. This is achieved using a matrix strategy, similarly to other jobs in this file. Using the matrix strategy allows adding a KleidiAI-enabled artifact job without a large amount of duplicate code.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,8 +36,26 @@ env:
   CMAKE_ARGS: "-DLLAMA_BUILD_EXAMPLES=OFF -DLLAMA_BUILD_TESTS=OFF -DLLAMA_BUILD_TOOLS=ON -DLLAMA_BUILD_SERVER=ON -DGGML_RPC=ON"
 
 jobs:
-  macOS-arm64:
-    runs-on: macos-14
+  macOS-cpu:
+    strategy:
+      matrix:
+        include:
+          - build: 'arm64'
+            arch: 'arm64'
+            os: macos-14
+            defines: "-DGGML_METAL_USE_BF16=ON -DGGML_METAL_EMBED_LIBRARY=ON"
+          - build: 'arm64-kleidiai'
+            arch: 'arm64'
+            os: macos-14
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -96,23 +67,20 @@ jobs:
...
-          # Metal is disabled due to intermittent failures with Github runners not having a GPU:
-          # https://github.com/ggml-org/llama.cpp/actions/runs/8635935781/job/23674807267#step:5:2313
           cmake -B build \
+            ${{ matrix.defines }} \
             -DCMAKE_INSTALL_RPATH='@loader_path' \
             -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
             -DLLAMA_FATAL_WARNINGS=ON \
             -DLLAMA_BUILD_BORINGSSL=ON \
-            -DGGML_METAL=OFF \
-            -DGGML_RPC=ON \
```
CISC, line 114: Disabling RPC was perhaps in error?

martin-klacer-arm: Thanks for the comment! 

I excluded the explicit RPC=ON flag from the additional CMake args as RPC appears to be enabled globally in env CMAKE_ARGS on line 36, which are appended to the end of the CMake flags anyway, so I believe enabling it here is redundant.

If there's a reason I missed to keep the RPC on flag here as well, do let me know and I can add it back. Thank you!

CISC (approved):

.github/workflows/release.yml
CISC: Ah, thanks, missed that.

taronaeo (approved):

PR #21263, opened 2026-04-01
Use default RISE RISC-V Runners

luhenry:
## Overview

Switching to default `runs-on: ubuntu-24.04-riscv`. It currently works with `runs-on: RISCV64` because there is a workaround in the app, but I would rather converge.

## Additional information

That's following the adoption of the [RISE RISC-V Runners](https://riseproject-dev.github.io/riscv-runner/) for linux-riscv64 runners. See https://github.com/ggml-org/llama.cpp/pull/16682#issuecomment-4161638130

# Requirements

N/A

### DIFF
--- a/.github/workflows/build-riscv.yml
+++ b/.github/workflows/build-riscv.yml
@@ -35,7 +35,7 @@ env:
 
 jobs:
   ubuntu-riscv64-native-sanitizer:
-    runs-on: RISCV64
+    runs-on: ubuntu-24.04-riscv
 
     continue-on-error: true
 
@@ -50,17 +50,18 @@ jobs:
           sudo apt-get update
 
           # Install necessary packages
-          sudo apt-get install -y libatomic1 libtsan2 gcc-14 g++-14 rustup cmake build-essential wget ccache git-lfs
+          sudo apt-get install -y libatomic1 libtsan2 gcc-14 g++-14 cmake build-essential wget git-lfs
 
           # Set gcc-14 and g++-14 as the default compilers
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
           sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 100
-          sudo ln -sf /usr/bin/gcc-14 /usr/bin/gcc
-          sudo ln -sf /usr/bin/g++-14 /usr/bin/g++
 
-          # Install Rust stable version
-          rustup install stable
-          rustup default stable
+          if ! which rustc; then
+            # Install Rust stable version
+            sudo apt-get install -y rustup
+            rustup install stable
+            rustup default stable
+          fi
 
           git lfs install
 
@@ -73,23 +74,12 @@ jobs:
         id: checkout
         uses: actions/checkout@v6
 
-      - name: Setup ccache
-        run: |
-          # Unique cache directory per matrix combination
-          export CCACHE_DIR="$HOME/.ccache/sanitizer-${{ matrix.sanitizer }}-${{ matrix.build_type }}"
-          mkdir -p "$CCACHE_DIR"
-
-          # Configure ccache
-          ccache --set-config=max_size=5G
-          ccache --set-config=compression=true
-          ccache --set-config=compression_level=6
-          ccache --set-config=cache_dir="$CCACHE_DIR"
-          ccache --set-config=sloppiness=file_macro,time_macros,include_file_mtime,include_file_ctime
-          ccache --set-config=hash_dir=false
-
-          # Export for subsequent steps
-          echo "CCACHE_DIR=$CCACHE_DIR" >> $GITHUB_ENV
-          echo "PATH=/usr/lib/ccache:$PATH" >> $GITHUB_ENV
+      # FIXME: Enable when ggml-org/ccache-action works on riscv64
+      # - name: ccache
+      #   uses: ggml-org/ccache-action@v1.2.21
+      #   with:
+      #     key: ubuntu-riscv64-native-sanitizer-${{ matrix.sanytizer }}-${{ matrix.build_type }}
+      #     save: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
 
       - name: Build
         id: cmake_build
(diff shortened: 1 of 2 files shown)


CISC (approved):

taronaeo (approved):

CISC:
@luhenry Hmmm, this is writing lots of `sccache` entries to our already strained GitHub Cache, can you make it store the cache locally?

Edit: At the very least only store it when run on the default branch, like the other jobs with `ccache`.

.github/workflows/build-riscv.yml
```diff
@@ -50,17 +50,20 @@ jobs:
...
 
           # Install necessary packages
-          sudo apt-get install -y libatomic1 libtsan2 gcc-14 g++-14 rustup cmake build-essential wget ccache git-lfs
+          sudo apt-get install -y libatomic1 libtsan2 gcc-14 g++-14 cmake build-essential wget git-lfs
 
           # Set gcc-14 and g++-14 as the default compilers
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
           sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 100
           sudo ln -sf /usr/bin/gcc-14 /usr/bin/gcc
           sudo ln -sf /usr/bin/g++-14 /usr/bin/g++
```
taronaeo: I think these aren't required right? `--install` would already create the symlink for you.

.github/workflows/build.yml
```diff
@@ -985,43 +986,41 @@ jobs:
...
 
           # Install necessary packages
-          sudo apt-get install -y libatomic1 libtsan2 gcc-14 g++-14 rustup cmake build-essential libssl-dev wget ccache git-lfs
+          sudo apt-get install -y libatomic1 libtsan2 gcc-14 g++-14 cmake build-essential libssl-dev wget git-lfs
 
           # Set gcc-14 and g++-14 as the default compilers
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
           sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 100
           sudo ln -sf /usr/bin/gcc-14 /usr/bin/gcc
           sudo ln -sf /usr/bin/g++-14 /usr/bin/g++
```
taronaeo: Likewise for here

.github/workflows/build-riscv.yml
luhenry: It will install a `gcc-14` and `g++-14`, but the configure step will have to specify them explicitly for them to be picked up. With the synlink, this “gcc” will then be gcc-14

CISC: The whole point of `update-alternatives` is to update the symlinks?

luhenry: Indeed, I missed these update-alternatives right above. I’ll verify locally .

luhenry: Yep, you're right. I'll remove the `ln -sf ...`

taronaeo (approved):

CISC:
@luhenry Are we ready to merge?

CISC:
@luhenry We've had some frequent runner failures:
https://github.com/ggml-org/llama.cpp/actions/runs/24877391114/job/72837070282
```
E: Could not open lock file /var/lib/dpkg/lock-frontend - open (2: No such file or directory)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
```

luhenry:
> @luhenry We've had some frequent runner failures: https://github.com/ggml-org/llama.cpp/actions/runs/24877391114/job/72837070282
> 
> ```
> E: Could not open lock file /var/lib/dpkg/lock-frontend - open (2: No such file or directory)
> E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
> ```

Fixed with  https://github.com/riseproject-dev/riscv-runner-images/commit/09d207a6e06a3c7e587a2c097568f4912d0b6fba. It’s deploying right now. And I’ll make sure to add a test for that as well, as it wasn’t covered previously. Sorry about the breakage!

alitariq4589:
@CISC I can see that CI tests are refactored now. Previously, there was a `build.yaml` file, which is not present anymore. Are all the tests for RISC-V added in the `build-riscv.yaml`? If I would like to add tests related to, say, `build-rpc.yml` for RISC-V, should I add them to `build-riscv.yml` or add riscv related changes in `build-rpc.yaml` and add `ubuntu-24.04-riscv` in it?

Also, it will be great if you point me towards the PR that removed the `build.yml` so I can read the context behind it

CISC:
> @CISC I can see that CI tests are refactored now. Previously, there was a `build.yaml` file, which is not present anymore. Are all the tests for RISC-V added in the `build-riscv.yaml`? If I would like to add tests related to, say, `build-rpc.yml` for RISC-V, should I add them to `build-riscv.yml` or add riscv related changes in `build-rpc.yaml` and add `ubuntu-24.04-riscv` in it?

Ideally all RISC-V related workflows should be in `build-riscv.yml`. Would you actually need to test RPC changes on RISC-V, or was it just an example?

> Also, it will be great if you point me towards the PR that removed the `build.yml` so I can read the context behind it

See #23789, it's only one of several refactors related to #20446.

alitariq4589:
> Ideally all RISC-V related workflows should be in build-riscv.yml. Would you actually need to test RPC changes on RISC-V, or was it just an example?

Some of the tests were previously not added due to the absence of transformers and PyTorch's dependency on RISC-V. But now we may be able to build them separately (or set up a nightly build which would build these dependencies and publish them to a separate place in Cloud-V repos). So I will be enabling those workflows for Cloud-V. I agree with you that we should place them in `build-riscv.yml` because we may need to tweak them for RISC-V, and adding RISC-V-related changes in the general workflows may break the functionality for other ISAs.

> See https://github.com/ggml-org/llama.cpp/pull/23789, it's only one of several refactors related to https://github.com/ggml-org/llama.cpp/discussions/20446.

Thanks. Will look into them

PR #21269, opened 2026-04-01
server: Bypass API Key validation for WebUI static bundle assets

allozaur:
Close #21229

# Requirements

- AI usage disclosure: YES (LLM wrote C++ syntax)

### DIFF
--- a/tools/server/server-http.cpp
+++ b/tools/server/server-http.cpp
@@ -143,7 +143,11 @@ bool server_http_context::init(const common_params & params) {
             "/v1/health",
             "/models",
             "/v1/models",
-            "/api/tags"
+            "/api/tags",
+            "/",
+            "/index.html",
+            "/bundle.js",
+            "/bundle.css",
         };
 
         // If API key is not set, skip validation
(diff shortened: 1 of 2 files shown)


tools/server/server-http.cpp
```diff
@@ -151,11 +151,22 @@ bool server_http_context::init(const common_params & params) {
...
-        // If path is public or is static file, skip validation
+        // If path is public, skip validation
         if (public_endpoints.find(req.path) != public_endpoints.end() || req.path == "/") {
             return true;
         }
 
+        // Skip validation for web UI static assets (they don't require API key)
+        {
+            static const std::unordered_set<std::string> static_assets = {
+                "/bundle.js", "/bundle.css", "/index.html",
```
ngxson: maybe it's easier to add these to the list of `public_endpoints` above ?

allozaur: ofc, ur right

tools/server/tests/unit/test_security.py
```diff
@@ -22,6 +22,16 @@ def test_access_public_endpoint(endpoint: str):
...
 
 
+@pytest.mark.parametrize("endpoint", ["/", "/bundle.js", "/bundle.css"])
+def test_access_static_assets_without_api_key(endpoint: str):
+    """Static web UI assets should not require API key authentication (issue #21229)"""
+    global server
+    server.start()
+    res = server.make_request("GET", endpoint)
+    assert res.status_code == 200
+    assert "error" not in res.body
```
ngxson: nits, but it's recommended to move the list `["/", "/bundle.js", "/bundle.css"]` inside the function so that we reuse the same server instance to test each endpoint (faster)

smth like:

```py
global server
server.start()
paths = ["/", "/bundle.js", "/bundle.css"]
for p in paths:
   # make request then assert
```

ngxson: in general, we use `parametrize` only when we need to change server params (i.e. reboot the server)

ngxson (approved):

ggerganov (approved):

PR #21270, opened 2026-04-01
contrib : rewrite AGENTS.md, make it more clear about project values

ngxson:
## Overview

Alternative to https://github.com/ggml-org/llama.cpp/pull/21007

This version make it more clear about the values of the project:
- We value human interactions
- Human contributor must "own" the code
- Contributor should be aware that a contribution is not just code, but also the planning and maintenance costs

# Requirements

- AI usage disclosure: yes, I wrote a broken english version and asked the AI to rewrite it; the idea is still mine

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,78 +5,103 @@
 >
 > Read more: [CONTRIBUTING.md](CONTRIBUTING.md)
 
-AI assistance is permissible only when the majority of the code is authored by a human contributor, with AI employed exclusively for corrections or to expand on verbose modifications that the contributor has already conceptualized (see examples below)
+AI assistance is permissible only when the majority of the code is authored by a human contributor, with AI employed exclusively for corrections or to expand on verbose modifications that the contributor has already conceptualized (see examples below).
 
 ---
 
 ## Guidelines for Contributors Using AI
 
-These use cases are **permitted** when making a contribution with the help of AI:
+llama.cpp is built by humans, for humans. Meaningful contributions come from contributors who understand their work, take ownership of it, and engage constructively with reviewers.
 
-- Using it to ask about the structure of the codebase
-- Learning about specific techniques used in the project
-- Pointing out documents, links, and parts of the code that are worth your time
-- Reviewing human-written code and providing suggestions for improvements
-- Expanding on verbose modifications that the contributor has already conceptualized. For example:
-    - Generating repeated lines with minor variations (this should only be used for short code snippets where deduplication would add more complexity, compared to having almost the same code in multiple places)
-    - Formatting code for consistency and readability
-    - Completing code segments based on established patterns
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -5,78 +5,106 @@
...
+### Prohibited AI Usage
 
-## Related Documentation
+The following will result in immediate PR closure:
 
-For related documentation on building, testing, and guidelines, please refer to:
+- **AI-written PR descriptions or commit messages** - these are typically recognizable and waste reviewer time
+- **AI-generated responses to reviewer comments** - this undermines the human-to-human interaction fundamental to code review
+- **Implementing features without understanding the codebase** - particularly new model support or architectural changes
+- **Automated commits or PR submissions** - this may spam maintainers and can result in contributor bans
```
ngxson, line 61: Writing code is no longer forbidden here. Note that I initially wrote this point as "no wall-of-text" and the AI rephrased it such that it sounds more professional

ServeurpersoCom (approved):

0cc4m (approved):

pwilkin (approved):

AGENTS.md
yychyo: AGENTS.md is loaded as system instructions by agents. The "Guidelines for Contributors" section is project policy for humans that fills agent context without making them more effective (that's peanuts, but still...). Should that part stay in CONTRIBUTING.md, keeping AGENTS.md focused on what agents should/shouldn't do? 🤔

0cc4m: That is a good point. It may be good to keep it shorter and only address the agent itself.

ngxson: The current content takes around ~1200 tokens which is I think is still somewhat acceptable. Ofc it is possible to reduce it to around 500 tokens (will be a follow up PR). I don't think a second file will work, because in most cases there are no reasons for the agent to read it

ServeurpersoCom: It's obviously possible to create a prompt that formats everything perfectly down to the smallest detail, like a human hardcore programmer, but it would be even harder to check. What a paradox

ServeurpersoCom: Why not simply give the assistant a behavior that explains to the user that they need to know precisely what they want to do, that they must ask all the necessary architectural questions, and that the work of others must be respected, and that humans have to proofread everything and that it takes time? The perfect Inception approach…

CISC (approved):

allozaur (approved):

AGENTS.md
ngxson: IMO prompting to change the behavior of the AI is almost like role play, and quite tricky to get right with heavily guarded models like claude or gpt.

Anw, I'm merging this PR as-is because it should already unblock some legitimate use cases for maintainers / contributors. Any improvement to this can be done via follow-up PRs.

PR #21277, opened 2026-04-01
kv-cache : do not quantize SWA KV cache

ggerganov:
## Overview

cont #21038 

We don't need to quantize the SWA part of the cache for iSWA models because it is relatively small. So keep it in F16.

# Requirements

- AI usage disclosure: NO


ggerganov:
This is expected, we need extra space beyond the `n_swa`. Here is the exact formula:

https://github.com/ggml-org/llama.cpp/blob/57ace0d612a11133ac86edcc7af1b323bf05f12f/src/llama-kv-cache-iswa.cpp#L47-L51

PR #21278, opened 2026-04-01
ggml-webgpu: move from parameter buffer pool to single buffer with offsets

reeselevine:
## Overview

Continuing some work to simplify and make the WebGPU backend scheduling more asynchronous, I realized that we don't actually need a pool of parameter buffers. Instead we can use a single buffer with multiple offset slots, and cycle through them on a batch of submissions. This PR replaces a pool with a `webgpu_param_arena`, and moves all operations to use it. Memset is special because it lives in the global context, but because it is now asynchronous it uses a single parameter buffer.

In this PR I also updated GPU submissions to be batched into a single `CommandBuffer`, instead of having a `CommandBuffer`per operation. This increases efficiency/speed a bit on larger systems and should help with stability on mobile devices.

# Requirements

- AI usage disclosure: yes, to help refactor to use the arena and simplify gpu profile future handling


nikhilJain17 (approved):
Nice! This removes a lot of the footguns we were previously dealing with in kernel submission and waiting by getting rid of individual future handles and just waiting on the whole queue.

ggerganov (approved):

CISC (approved):
There's a few formatting issues, but I'll keep my OCD in check. :)

reeselevine:
@CISC haha I am trying to use clang-format to make sure it's standardized in some way, but perhaps there are better defaults for it :)

PR #21234, opened 2026-03-31
scripts: add function call test script

ngxson:
## Overview

Adding a test script to allow testing model's tool calling capability with realistic use cases, inspired by https://huggingface.co/datasets/JungHun/Efficient_ToolCalling_train

I decided to add a dedicated script instead of relying on current testing infra because this is mostly for one-time testing when adding a new model.

Quite straight-forward to use:

```sh
# terminal 1
llama-server -m ....

# terminal 2
python scripts/server-test-function-call.py
```

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: YES, this is nice-to-have part of the project, so I didn't want to spend too much time on it

PR #21286, opened 2026-04-02
common : add commentary rules for gpt-oss-20b

aldehir:
## Overview

Add additional `<|channel|>commentary to=assistant` rules for gpt-oss-20b. These are not defined in the Harmony docs, but appears to be a common generation from the 20b variant.

ref: https://github.com/ggml-org/llama.cpp/pull/21234#pullrequestreview-4042554282

## Additional information

Prereq: #21285 (or we can merge this one and close the other)

# Requirements

- AI usage disclosure: NO


ggerganov (approved):
This fixes the tool calling issue with `gpt-oss-20b` on my end.

pwilkin (approved):

PR #21287, opened 2026-04-02
model: support step3-vl-10b

forforever73:
## Overview

Add support for [STEP3-VL-10B](https://huggingface.co/stepfun-ai/Step3-VL-10B) multimodal vision-language model, a 10B VLM from StepFun AI with a PE-lang vision encoder (1.8B) and Qwen3-8B decoder.

feature request in [#19258](https://github.com/ggml-org/llama.cpp/issues/19258)

# Requirements
- AI usage disclosure: YES, codex to assist with automated testing


convert_hf_to_gguf.py
```diff
@@ -4949,6 +4949,136 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+            block_id = int(match.group(1))
+            block_name = match.group(2)
+
+            if block_name.startswith("attn.in_proj_"):
+                suffix = ".weight" if block_name.endswith("_weight") else ".bias"
+                q_tensor, k_tensor, v_tensor = torch.chunk(data_torch, 3, dim=0)
+                yield (self.format_tensor_name(gguf.MODEL_TENSOR.V_ENC_ATTN_Q, block_id, suffix=suffix), q_tensor)
+                yield (self.format_tensor_name(gguf.MODEL_TENSOR.V_ENC_ATTN_K, block_id, suffix=suffix), k_tensor)
+                yield (self.format_tensor_name(gguf.MODEL_TENSOR.V_ENC_ATTN_V, block_id, suffix=suffix), v_tensor)
+                return
```
CISC: Do not split QKV, `build_vit` handles them fused.

convert_hf_to_gguf.py
```diff
@@ -4949,6 +4949,136 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+        assert self.hparams_vision is not None
+
+        self.gguf_writer.add_clip_projector_type(gguf.VisionProjectorType.STEP3VL)
+        self.gguf_writer.add_vision_attention_layernorm_eps(float(self.hparams_vision["layer_norm_eps"]))
+        self.gguf_writer.add_vision_projector_scale_factor(4)
+        # 3024 max resize and 504 local crop come from step3-vl-10b processing_step3.py.
+        self.gguf_writer.add_vision_preproc_image_size(3024)
+        self.gguf_writer.add_vision_image_crop_resolution(504)
```
ngxson: it's recommended to read those from `config.json`, otherwise you can add some `assert` to avoid you releasing a new model in the future, and user may re-use this script, resulting in a broken GGUF

otherwise, it's recommended to just set those params in GGUF, so that you don't need to create a new `image_crop_resolution` metadata right now

convert_hf_to_gguf.py
```diff
@@ -12700,6 +12840,11 @@ def get_model_architecture(hparams: dict[str, Any], model_type: ModelType) -> st
         # For non-hf Mamba and Mamba2 models
         arch = hparams["ssm_cfg"].get("layer", "Mamba") + "ForCausalLM"
 
+    # Step3-VL keeps text config under text_config but uses a custom top-level architecture.
+    # For text conversion we route to a dedicated text-only class.
```
ngxson, line 12998:
```suggestion
    # For text conversion we route to a dedicated text-only class.
    # TODO: refactor this later to avoid adding exception here
```

ngxson (commented):
The mtmd part (C++ code) looks good

convert_hf_to_gguf.py
forforever73: These values aren’t in config.json, so I added an assert for this 10B model. And just kept image_crop_resolution as a runtime default.

gguf-py/gguf/constants.py
```diff
@@ -297,27 +297,27 @@ class Clip:
...
+        BLOCK_COUNT           = "clip.vision.block_count"
+        IMAGE_MEAN            = "clip.vision.image_mean"
+        IMAGE_STD             = "clip.vision.image_std"
+        SPATIAL_MERGE_SIZE    = "clip.vision.spatial_merge_size"
+        USE_GELU              = "clip.use_gelu"
+        USE_SILU              = "clip.use_silu"
+        N_WA_PATTERN          = "clip.vision.n_wa_pattern" # used by qwen2.5vl
+        WA_LAYER_INDEXES      = "clip.vision.wa_layer_indexes" # used by youtuvl
+        IS_DEEPSTACK_LAYERS   = "clip.vision.is_deepstack_layers"
+        WINDOW_SIZE           = "clip.vision.window_size"
```
ngxson: also need to revert this

tools/mtmd/mtmd-image.cpp
```diff
@@ -30,6 +30,59 @@ void mtmd_image_preprocessor::img_u8_to_f32(const clip_image_u8 & src, clip_imag
...
+        const clip_image_u8 & src,
+        clip_image_f32 & dst,
+        int target_width,
+        int target_height,
+        const float mean[3],
+        const float std[3]) {
+    if (src.nx == target_width && src.ny == target_height) {
+        img_u8_to_f32(src, dst, mean, std);
+        return;
+    }
```
ngxson: there is no need to add such function, you can use the existing `bilinear` resize mode then finally `img_u8_to_f32`

ngxson (changes requested):

tools/mtmd/mtmd-image.cpp
forforever73: I did try reusing img_tool::resize followed by img_u8_to_f32 first, but the error was quite large. In https://huggingface.co/stepfun-ai/Step3-VL-10B/blob/main/processing_step3.py the reference path is effectively operating in float space (to_tensor -> normalize -> resize), whereas our current img_tool::resize performs bilinear resize in uint8 space and converts back to uint8 before normalization. That introduces extra quantization error, and the resize behavior is not the same as the original tensor resize path either.

ngxson: in such case, better to put this logic inside `mtmd_image_preprocessor_step3vl` because it's model-specific

forforever73: Sry for delay, this function now in the mtmd_image_preprocessor_step3vl class.

ngxson (approved):
Can be merged after @CISC's review

convert_hf_to_gguf.py
```diff
@@ -4949,6 +4949,93 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+        vision_config["use_ln_pre"] = bool(vision_config.get("use_ln_pre", True))
+        vision_config["use_ln_post"] = bool(vision_config.get("use_ln_post", False))
+        vision_config["use_abs_posemb"] = bool(vision_config.get("use_abs_posemb", True))
+        vision_config["use_rope2d"] = bool(vision_config.get("use_rope2d", True))
+        hparams["vision_config"] = vision_config
+
+        super().__init__(dir_model, *args, hparams=hparams, **kwargs)
+
+        self.preprocessor_config.setdefault("image_mean", list(_MISTRAL_COMMON_DATASET_MEAN))
+        self.preprocessor_config.setdefault("image_std", list(_MISTRAL_COMMON_DATASET_STD))
```
```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        assert self.hparams_vision is not None

        if not self.hparams_vision.get("intermediate_size"):
            hidden_size = self.hparams_vision.get("hidden_size") or self.hparams_vision.get("width") or 0
            assert hidden_size > 0
            mlp_ratio = float(self.hparams_vision.get("mlp_ratio", 8960 / 1536))
            self.hparams_vision["intermediate_size"] = int(round(hidden_size * mlp_ratio))

        self.preprocessor_config.setdefault("image_mean", list(_MISTRAL_COMMON_DATASET_MEAN))
        self.preprocessor_config.setdefault("image_std", list(_MISTRAL_COMMON_DATASET_STD))
```
Add `width`, `layers` and `heads` to `MmprojModel.set_gguf_parameters`.
The rest is unused AFAICT.

convert_hf_to_gguf.py
```diff
@@ -4949,6 +4949,93 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+
+        projector_stride = int(self.global_config.get("understand_projector_stride", -1))
+        hidden_size = int(self.hparams_vision.get("hidden_size", self.hparams_vision.get("width", -1)))
+        num_layers = int(self.hparams_vision.get("num_hidden_layers", self.hparams_vision.get("layers", -1)))
+        assert (projector_stride, int(self.hparams_vision.get("image_size", -1)), hidden_size, num_layers) == (2, 728, 1536, 47), (
+            "current Step3-VL conversion path is only validated for Step3-VL-10B"
+        )
+
+        self.gguf_writer.add_clip_projector_type(gguf.VisionProjectorType.STEP3VL)
+        self.gguf_writer.add_vision_attention_layernorm_eps(float(self.hparams_vision["layer_norm_eps"]))
```
```suggestion
        self.gguf_writer.add_vision_attention_layernorm_eps(float(self.hparams_vision.get("layer_norm_eps", 1e-5)))
```

convert_hf_to_gguf.py
```diff
@@ -4949,6 +4949,93 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+
+    def tensor_force_quant(self, name, new_name, bid, n_dims):
+        if ".position_embd." in new_name:
+            return gguf.GGMLQuantizationType.F32
+        return super().tensor_force_quant(name, new_name, bid, n_dims)
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        del bid
+
+        if name.startswith("model.") or name.startswith("lm_head."):
```
```suggestion
        if name.startswith("model.") or name.startswith("lm_head."):
```

convert_hf_to_gguf.py
```diff
@@ -4949,6 +4949,93 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+            elif name.endswith(".gamma") and ".ls_" in name:
+                name = name.removesuffix(".gamma") + ".weight"
+
+            name = name.replace("attn.in_proj_weight", "attn.in_proj.weight")
+            name = name.replace("attn.in_proj_bias", "attn.in_proj.bias")
+
+            yield (self.map_tensor_name(name), data_torch)
+            return
+
+        return
```
```suggestion
            yield from super().modify_tensors(data_torch, name, bid)
```

CISC (approved):

CISC (approved):

CISC (changes requested):
You forgot to add `width`, ~`layers`~ and `heads` to `MmprojModel.set_gguf_parameters`.

CISC (approved):

PR #21304, opened 2026-04-02
llama-bench: add `-fitc` and `-fitt` to arguments

am17an:
Add `-fitt` and `-fitc` to llama-bench. 

Output example

| model                          |       size |     params | backend    | ngl | fa | mmap |       fitt |        fitc |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -: | ---: | ---------: | ----------: | --------------: | -------------------: |
| qwen35moe 35B.A3B Q4_K - Small |  18.55 GiB |    34.66 B | CUDA       |  99 |  1 |    0 |      10000 |      100000 |           pp512 |      7001.57 ± 20.54 |
| qwen35moe 35B.A3B Q4_K - Small |  18.55 GiB |    34.66 B | CUDA       |  99 |  1 |    0 |      10000 |      200000 |           pp512 |     4481.81 ± 715.27 |
| qwen35moe 35B.A3B Q4_K - Small |  18.55 GiB |    34.66 B | CUDA       |  99 |  1 |    0 |      10000 |      300000 |           pp512 |      3655.35 ± 44.06 |

# Requirements

- AI usage disclosure: Yes, used for writing parts of the code since it's quite repetitive

### DIFF
--- a/tools/llama-bench/llama-bench.cpp
+++ b/tools/llama-bench/llama-bench.cpp
@@ -386,8 +386,8 @@ static const cmd_params cmd_params_defaults = {
     /* embeddings           */ { false },
     /* no_op_offload        */ { false },
     /* no_host              */ { false },
-    /* fit_params_target    */ { 1024 },
-    /* fit_params_min_ctx   */ { 4096 },
+    /* fit_params_target    */ { 0 },
+    /* fit_params_min_ctx   */ { 0 },
     /* numa                 */ GGML_NUMA_STRATEGY_DISABLED,
     /* reps                 */ 5,
     /* prio                 */ GGML_SCHED_PRIO_NORMAL,
@@ -2243,18 +2243,16 @@ int main(int argc, char ** argv) {
 
             std::vector<size_t> margins(llama_max_devices(), inst.fit_target * 1024 * 1024);
 
+            uint32_t n_ctx_needed = inst.n_prompt + inst.n_gen + inst.n_depth;
+            cparams.n_ctx = std::max(cparams.n_ctx, n_ctx_needed);
+
             llama_params_fit(inst.model.c_str(), &mparams, &cparams,
                 fit_tensor_split.data(),
                 fit_overrides.data(),
                 margins.data(),
                 inst.fit_min_ctx,
                 params.verbose ? GGML_LOG_LEVEL_DEBUG : GGML_LOG_LEVEL_ERROR);
-
-            uint32_t n_ctx_needed = inst.n_prompt + inst.n_gen + inst.n_depth;
-            if (cparams.n_ctx < n_ctx_needed) {
-                cparams.n_ctx = n_ctx_needed;
-            }
-        }
+       }
 
         // keep the same model between tests when possible
         if (!lmodel || !prev_inst || !inst.equal_mparams(*prev_inst)) {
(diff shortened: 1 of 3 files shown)


tools/llama-bench/llama-bench.cpp
```diff
@@ -2141,21 +2218,59 @@ int main(int argc, char ** argv) {
         if (params.progress) {
             fprintf(stderr, "llama-bench: benchmark %d/%zu: starting\n", params_idx, params_count);
         }
+        auto mparams = inst.to_llama_mparams();
+        auto cparams = inst.to_llama_cparams();
+
+        bool do_fit = inst.fit_target != cmd_params_defaults.fit_params_target[0] ||
+                      inst.fit_min_ctx != cmd_params_defaults.fit_params_min_ctx[0];
```
JohannesGaessler, line 2225: If I understand the code correctly the intend here is to disable `llama_params_fit` by default. However, if a user manually sets `-fitt 1024 -fitc 4096` I think it will also be disabled.

tools/llama-bench/llama-bench.cpp
```diff
@@ -2141,21 +2218,59 @@ int main(int argc, char ** argv) {
...
+                fit_tensor_split.data(),
+                fit_overrides.data(),
+                margins.data(),
+                inst.fit_min_ctx,
+                params.verbose ? GGML_LOG_LEVEL_DEBUG : GGML_LOG_LEVEL_ERROR);
+
+            uint32_t n_ctx_needed = inst.n_prompt + inst.n_gen + inst.n_depth;
+            if (cparams.n_ctx < n_ctx_needed) {
+                cparams.n_ctx = n_ctx_needed;
+            }
```
JohannesGaessler: I think it makes more sense to do this check before calling `llama_params_fit` by modifying the minimum context so that it will show up correctly in the logs.

tools/llama-bench/llama-bench.cpp
am17an: I can set the defaults to 0 and keep the same check, will that be alright?

JohannesGaessler: Would be fine I think.

JohannesGaessler (approved):
I think it's suboptimal that `llama-bench` and the other binaries are inconsistent when it comes to when `llama_params_fit` is used but I would still consider this PR to be an improvement over master.

ggerganov (approved):

tools/llama-bench/llama-bench.cpp
```diff
@@ -1473,7 +1517,8 @@ struct test {
             "type_k",         "type_v",         "n_gpu_layers",  "n_cpu_moe",      "split_mode",
             "main_gpu",       "no_kv_offload",  "flash_attn",    "devices",        "tensor_split",
             "tensor_buft_overrides",            "use_mmap",      "use_direct_io",  "embeddings",
-            "no_op_offload",  "no_host",        "n_prompt",      "n_gen",          "n_depth",
+            "no_op_offload",  "no_host",        "fit_target",     "fit_min_ctx",
+            "n_prompt",       "n_gen",          "n_depth",
```
CISC, line 1521: Changing this requires you to update `scripts/compare-llama-bench.py`.

JohannesGaessler: Good catch, I forgot about that.

CISC (approved):

PR #21309, opened 2026-04-02
model: support gemma 4 (vision + moe, no audio)

ngxson:
## Overview

Fix a bug where model with both vision/audio cannot be converted properly

# Requirements

- AI usage disclosure: NO


danbev (approved):

ggerganov (approved):

convert_hf_to_gguf.py
```diff
@@ -7425,6 +7436,212 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+        # using a dedicated tokenizer name so that we can fix later without re-converting GGUF
+        self.gguf_writer.add_tokenizer_model("gemma4")
+        self.gguf_writer.add_token_list(tokens)
+        self.gguf_writer.add_token_scores(scores)
+        self.gguf_writer.add_token_types(toktypes)
+
+        special_vocab = gguf.SpecialVocab(self.dir_model, load_merges=True)
+        special_vocab.add_to_gguf(self.gguf_writer)
+        self.gguf_writer.add_add_space_prefix(False)
+        self.gguf_writer.add_add_bos_token(False) # already added via the chat template
```
CISC, line 7478: If already added in the chat template that means it should be `True`, or is it not a single token?

danbev (approved):

ggerganov (approved):

convert_hf_to_gguf.py
ngxson: the chat template already had `{{ bos_token }}`, so `add_bos_token` is not necessary (though even if it set to True here, llama.cpp do have a logic to avoid double BOS)

I'm explicitly setting it to False here for clarity though

CISC: I think you misunderstand the purpose of this field, it's to explicitly signal that a model requires BOS, ie. for correct behavior when not using a chat template (completion or FIM).

ngxson: According to the HF implementation, BOS is defined but not being added automatically when I try `tokenizer("my prompt", return_tensors="pt")`, so I hope this is correct

ngxson: (For context, this is to match the behavior of HF transformers when I compare the activations via `llama-eval-callback`)

CISC: OK, so `False` is probably the correct value then.

Kreijstal:
no audio :(

PR #21326, opened 2026-04-02
Gemma 4 template parser fixes

pwilkin:
## Overview

As in topic

## Additional information

Quick fixes for some observed discrepancies + refactoring of the parser architecture for the dict format

# Requirements

- AI usage disclosure: YES for the refactoring, NO for the fixes

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -7,10 +7,108 @@
 #include "log.h"
 #include "nlohmann/json.hpp"
 
+#include <algorithm>
 #include <stdexcept>
 #include <string>
 
 using json = nlohmann::ordered_json;
+
+namespace {
+
+// Gemma4-specific PEG builder extending the standard chat builder.
+// Adds value type parsers that use <|\"|> as string delimiters
+// instead of JSON's double quotes, and disables json-to-schema
+// conversion for these types.
+class common_peg_gemma4_builder {
+    common_chat_peg_builder & p_;
+    static constexpr const char * QUOTE = "<|\"|>";
+
+public:
+    explicit common_peg_gemma4_builder(common_chat_peg_builder & p) : p_(p) {}
+
+    common_peg_parser gemma4_string() {
+        return p_.rule("gemma4-string", [&]() {
+            return p_.literal(QUOTE) + p_.until(QUOTE) + p_.literal(QUOTE);
+        });
+    }
+
+    common_peg_parser gemma4_number() {
+        return p_.rule("gemma4-number", [&]() {
+            auto digit1_9 = p_.chars("[1-9]", 1, 1);
+            auto digits   = p_.chars("[0-9]");
+            auto int_part = p_.choice({p_.literal("0"), p_.sequence({digit1_9, p_.chars("[0-9]", 0, -1)})});
+            auto frac     = p_.sequence({p_.literal("."), digits});
+            auto exp      = p_.sequence({p_.choice({p_.literal("e"), p_.literal("E")}),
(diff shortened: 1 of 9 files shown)


common/peg-parser.cpp
```diff
@@ -737,6 +737,46 @@ struct parser_executor {
...
+
+            pos += utf8_result.bytes_consumed;
+            last_valid_pos = pos;
+        }
+
+        if (last_valid_pos == ctx.input.size() && ctx.is_lenient()) {
+            return common_peg_parse_result(COMMON_PEG_PARSE_RESULT_NEED_MORE_INPUT, start_pos, last_valid_pos);
+        }
+        return common_peg_parse_result(COMMON_PEG_PARSE_RESULT_NEED_MORE_INPUT, start_pos, last_valid_pos);
+    }
```
aldehir: This is functionally equivalent to `p.until("<|\"|>") + p.literal("<|\"|>")`. There's no need for a new parser.

common/chat.h
```diff
@@ -214,8 +215,8 @@ struct common_chat_params {
     bool                                grammar_lazy         = false;
     std::string                         generation_prompt;
     bool                                supports_thinking    = false;
-    std::string                         thinking_start_tag;  // e.g., "<think>"
-    std::string                         thinking_end_tag;    // e.g., "</think>"
+    std::string                         thinking_start_tag;  // e.g., "💭"
+    std::string                         thinking_end_tag;    // e.g., "_flow"
```
aldehir: ???

aldehir (commented):
We should roll a dedicated parser for this model.

common/chat.h
pwilkin: Beats me, model went cuckoo :P

common/chat-auto-parser-generator.cpp
```diff
@@ -461,34 +561,35 @@ common_peg_parser analyze_tools::build_tool_parser_tag_gemma4_dict(parser_build_
...
+            } else if (type == "null") {
+                value_parser = p.tool_arg_value(g4.gemma4_null());
+            } else if (type == "object") {
+                value_parser = p.tool_arg_value(g4.gemma4_dict());
+            } else if (type == "array") {
+                value_parser = p.tool_arg_value(g4.gemma4_array());
             } else {
-                // Numbers, booleans: raw text up to the next comma or closing brace
-                value_parser = p.tool_arg_value(p.until_one_of({",", "}"}));
+                value_parser = p.tool_arg_value(g4.gemma4_value());
```
aldehir: Should use `gemma4_value_for_type()` here?

common/chat-peg-parser.cpp
```diff
@@ -75,6 +75,84 @@ static std::string escape_json_string_inner(const std::string & s) {
     return escaped;
 }
 
+static const std::string GEMMA4_QUOTE = "<|\"|>";
+
+static std::string normalize_gemma4_to_json(const std::string & input) {
+    std::string result;
+    result.reserve(input.size() * 2);
```
aldehir, line 82: In the previous `chat-peg-parser`, I had a mapper that would build this JSON up incrementally via the AST instead of through a separate pass. Was that removed?

aldehir (approved):
Ok, undo the core peg parser changes, undo unrelated changes, and this is fine as a quick fix for now.

common/chat-peg-parser.cpp
pwilkin: No, it's still there (`void common_chat_peg_mapper::map(const common_peg_ast_node & node)`). Would have to adapt to the funny format, the model I used for the refactoring was too dumb to do it apparently.

ggerganov (approved):

PR #21327, opened 2026-04-02
common : fix tool call type detection for nullable and enum schemas

sacredvoid:
## Overview

Fixes #21316

The Gemma4 dict parser and the tagged parser both only check `type_v.is_string()` when figuring out if a tool argument is a string. This breaks for schemas that use nullable types like `"type": ["string", "null"]` or enum fields without an explicit `"type"` key, both of which are pretty common in OpenAPI/Home Assistant setups.

When the type isn't recognized as `"string"`, the parser falls through to the raw-value path and captures `<|"|>` delimiter tokens as literal text, which is how you end up with output like `"domain": "[<|\"|>light<|\"|>]"` instead of `"domain": "light"`.

The fix is straightforward:
- Extract the effective type from type arrays (grab the first non-null entry)
- Infer string type from `enum` fields that have string values
- Applied to both `build_tool_parser_tag_gemma4_dict` and `build_tool_parser_tag_tagged` since they share the same pattern

All existing tests pass (`test-chat-auto-parser`: 56 tests, 373 assertions, 0 failures; `test-chat`: all passed).

# Requirements

- AI usage disclosure: YES - used Claude Code to explore the codebase, trace through the parser logic, and identify the root cause. The fix itself is a simple type-detection improvement that I reviewed and tested manually.

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -299,12 +299,29 @@ common_peg_parser analyze_tools::build_tool_parser_tag_tagged(parser_build_conte
         for (const auto & [param_name, param_schema] : properties.items()) {
             bool        is_required = required.find(param_name) != required.end();
             std::string type        = "object";
-            auto        type_obj    = param_schema.contains("type") ? param_schema.at("type") : json::object();
-            if (type_obj.is_string()) {
-                type_obj.get_to(type);
-            } else if (type_obj.is_object()) {
-                if (type_obj.contains("type") && type_obj.at("type").is_string()) {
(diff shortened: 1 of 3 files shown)


ngxson:
Yes I'll test it tomorrow

aldehir:
It does infer types if there is no `type` property. I say hold off until we get input from other maintainers.

I might have jumped the gun and perhaps this surgical change might be good for now and we can address the general case later.

pwilkin (approved):
I'd say we utilize this fix now and possibly generalize later.

aldehir (approved):

PR #21343, opened 2026-04-03
fix: Gemma4 tokenizer

pwilkin:
## Overview

A very ugly fix to the Gemma 4 tokenizer

**IMPORTANT NOTE**: new gguf is NOT changed, this is a C++ change only. Re-generating gguf is not necessary.

## Additional information

The Gemma 4 tokenizer had a subtle bug due to using the SPE tokenizer that caused it to incorrectly tokenize stuff, leading to significant errors in longer sessions.

# Requirements

- AI usage disclosure: YES, had Claude murder the tokenizer code

I have no idea what I'm doing, it's 2 AM and I've spent the last 4 hours chasing everything from scale discrepancies to tokenizers, but this seems to actually fix Gemma 4.

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -7464,9 +7464,6 @@ class Gemma4Model(Gemma3Model):
 
         assert len(tokens) == vocab.vocab_size
 
-        # TODO @ngxson : there are some known (rare) issues with the tokenizer during development
-        # but I don't have time to dive into them right now;
-        # using a dedicated tokenizer name so that we can fix later without re-converting GGUF
         self.gguf_writer.add_tokenizer_model("gemma4")
         self.gguf_writer.add_token_list(tokens)
         self.gguf_writer.add_token_scores(scores)
--- a/src/llama-vocab.cpp
+++ b/src/llama-vocab.cpp
@@ -493,6 +493,16 @@ struct llm_tokenizer_bpe : llm_tokenizer {
                     "(?:'[sS]|'[tT]|'[rR][eE]|'[vV][eE]|'[mM]|'[lL][lL]|'[dD])|[^\\r\\n\\p{L}\\p{N}]?(?:\\p{L}\\p{M}*(?: \\p{L}\\p{M}*)*)+|\\p{N}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n/]?|\\s*[\\r\\n]|\\s+(?!\\S)|\\s+",
                 };
                 break;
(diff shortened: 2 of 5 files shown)


sayap:
Can you look into some findings at https://github.com/ikawrakow/ik_llama.cpp/issues/1572#issuecomment-4181866837? With this PR, it still doesn't tokenize multiple consecutive \n in the prompt correctly.

ngxson (approved):
Very nice fix, thanks. Indeed @danbev came up with a somewhat same fix during development, but I was struggled to get it work on server, so the fix was delayed to post-release (hence the new `gemma4` tokenizer name)

Can you also remove the comment here?

https://github.com/ggml-org/llama.cpp/blob/57ace0d612a11133ac86edcc7af1b323bf05f12f/convert_hf_to_gguf.py#L7467-L7473

danielhanchen:
Had to ask CC / Codex to test more:
1. Regex should match consecutive newlines as a group:
```
// Before:
"[^\\n]+|\\n",
// After:
"[^\\n]+|[\\n]+",
```
Without this, \n\n gets split into two separate \n tokens (107, 107) instead of one \n\n token (108). The model was trained with multi-newline tokens.

2. Newline-only chunks need a direct vocab lookup bypass before BPE merging:
```diff
             if (vocab.get_ignore_merges() && vocab.text_to_token(word) != LLAMA_TOKEN_NULL) {
                 symbols.emplace_back(llm_symbol{-1, -1, word.c_str(), word.size()});
                 offset = word.size();
+            } else if (word.find_first_not_of('\n') == std::string::npos) {
+                auto tok = vocab.text_to_token(word);
+                if (tok != LLAMA_TOKEN_NULL) {
+                    symbols.emplace_back(llm_symbol{-1, -1, word.c_str(), word.size()});
+                    offset = word.size();
+                }
             }
```
Even with [\n]+ grouping, the BPE merge loop cannot process newline-only chunks (silent failure, 0 tokens output). This bypasses BPE and does a direct vocab lookup for \n\n -> 108, \n\n\n -> 109, etc.

With both fixes applied, llama-tokenize matches transformers AutoTokenizer on all test cases. Credit to @sayap for identifying both issues.

danielhanchen (approved):

ggerganov (approved):

danielhanchen:
Yay!

multi-vitamin:
`llama.cpp/src/llama-vocab.cpp:3715: GGML_ASSERT(token_left.find('\n') == std::string::npos) failed`

Using **fetch_url** feature in Open WebUI triggers llama.cpp to terminate.

PR #21345, opened 2026-04-03
ci : add AMD ZenDNN label to PR labeler

z-vishal:
This PR adds a new "AMD ZenDNN" label to automatically tag pull requests that modify ZenDNN backend files, making it easier to track changes to AMD's EPYC CPU acceleration support.

cc: @amukho @avinashcpandey @taronaeo @danbev

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -27,6 +27,11 @@ IBM zDNN:
         - any-glob-to-any-file:
             - ggml/include/ggml-zdnn.h
             - ggml/src/ggml-zdnn/**
+AMD CPU:
+    - changed-files:
+        - any-glob-to-any-file:
+            - ggml/include/ggml-zendnn.h
+            - ggml/src/ggml-zendnn/**
 documentation:
     - changed-files:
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -27,6 +27,11 @@ IBM zDNN:
         - any-glob-to-any-file:
             - ggml/include/ggml-zdnn.h
             - ggml/src/ggml-zdnn/**
+AMD CPU:
```
taronaeo:
```suggestion
AMD ZenDNN:
```

We already have a label created for you under `AMD ZenDNN`. Would you prefer `AMD CPU` or `AMD ZenDNN`?

taronaeo (approved):

.github/labeler.yml
z-vishal: Agreed, AMD ZenDNN is more specific and consistent with the existing label. Applying the suggestion!

taronaeo (approved):

am17an (approved):

PR #21355, opened 2026-04-03
fix: add openssl to nix dependencies (#21353)

Tillerino:
## Overview

Closes #21353. Error was:

`llama-cli -hf unsloth/gemma-4-26B-A4B-it-GGUF:UD-Q4_K_XL` ->

```
HTTPS is not supported. Please rebuild with one of:
  -DLLAMA_BUILD_BORINGSSL=ON
  -DLLAMA_BUILD_LIBRESSL=ON
  -DLLAMA_OPENSSL=ON (default, requires OpenSSL dev files installed)
```

## Additional information

This just slaps in openssl as a dependency. llama.cpp supports other libraries, but openssl is the default and thus no further configuration is needed.

I guess you could make the nix package configurable to use other ssl libraries, but I am not sure that folks would even care.

# Requirements

- AI usage disclosure: no

### DIFF
--- a/.devops/nix/package.nix
+++ b/.devops/nix/package.nix
@@ -16,7 +16,7 @@
   rocmPackages,
   vulkan-headers,
   vulkan-loader,
-  curl,
+  openssl,
   shaderc,
   useBlas ?
     builtins.all (x: !x) [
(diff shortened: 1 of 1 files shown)


.devops/nix/package.nix
```diff
@@ -16,7 +16,7 @@
   rocmPackages,
   vulkan-headers,
   vulkan-loader,
-  curl,
+  openssl,
```
Tillerino, line 19: `curl` was not referenced, so I removed it.

CISC (approved):

ggerganov (approved):

PR #1360, opened 2023-05-07
Implement backward passes for llama with small training llama from scratch example

xaedes:
Training a llama directly with ggml would be really nice.

For this I implemented the backward passes required for the llama model, tested them with test_grad0 from ggml repo and trained a small llama from scratch to output a sinus wave.

Also see the more detailed discussion in https://github.com/ggerganov/ggml/issues/8#issuecomment-1518465663

List of all new operations that I had to add:
- GGML_OP_ADD1 : Could be replaced with add(X,repeat(Y,X)) but it is faster when repeat can be avoided.
- GGML_OP_ACC : Necessary for view backward pass. This adds src1 to view(src0, nb, offset) and returns tensor of shape src0. 
- GGML_OP_SET : **(this is new)** Necessary for propagating gradients through kv cache. Instead of copying to kv cache this function sets the values in the kv cache viewed with offsets and strides and returns a tensor representing the modified kv cache. This can also be inplace returning a view to the modified kv cache.
- GGML_OP_LOG : **(this is new)** Necessary for cross entropy loss
- GGML_OP_SUM_ROWS : Necessary for repeat backward pass: Reduces rows by summing them. shape[a,b,c,d] -> shape[1,b,c,d]
- GGML_OP_SILU_BACK : Necessary for silu backward pass
- GGML_OP_RMS_NORM_BACK : Could also be implemented using primitives, at the cost of performance.
- GGML_OP_GET_ROWS_BACK : Necessary for get_rows backward pass: Adds src0[i] rows to opt0[src1[i]] rows, returning a tensor of shape opt0. 
- GGML_OP_DIAG : Necessary for softmax backward pass, alternative would have been to implement SOFTMAX_BACK directly, but DIAG is at least usable for other stuff. It turns rows into diagonal matrices.
- GGML_OP_DIAG_MASK_ZERO : Necessary for diag_mask_inf backward pass
- GGML_OP_ROPE_BACK : Necessary for rope backward pass.

Notable other changes:
- add inplace and non-inplace variants for scale, diag_mask_inf, soft_max and rope
- fix sub, mul and div functions to work correctly with transposed tensor, uses the same logic as in add:
- fix ggml_forward_add functions to work correctly with transposed tensors. uses the same logic as in ggml_compute_forward_add_q_f32, but make it consistent across all ggml_compute_forward_add_... functions. this also slightly changes the mem access pattern of the different threads to work as in ggml_compute_forward_add_q_f32. 
[...]

PR #21363, opened 2026-04-03
ggml : deprecate GGML_OP_ADD1

ggerganov:
## Overview

The `GGML_OP_ADD1` was added back in #1360. However, the op is a subclass of the generic `GGML_OP_ADD` and in favor of simplicity, it's better to remove it. Deprecating for now.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -16274,9 +16274,8 @@ static void ggml_vk_check_results_0(ggml_backend_vk_context * ctx, ggml_cgraph *
         } else if (tensor->op == GGML_OP_SCALE) {
             const float * params = (const float *)tensor->op_params;
             tensor_clone = ggml_scale_bias(ggml_ctx, src_clone[0], params[0], params[1]);
-        // NOTE: this is deprecated
-        //} else if (tensor->op == GGML_OP_ADD1) {
-        //    tensor_clone = ggml_add1(ggml_ctx, src_clone[0], src_clone[1]);
+        } else if (tensor->op == GGML_OP_ADD1) {
+            tensor_clone = ggml_add1(ggml_ctx, src_clone[0], src_clone[1]);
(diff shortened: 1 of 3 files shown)


ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -16274,8 +16274,9 @@ static void ggml_vk_check_results_0(ggml_backend_vk_context * ctx, ggml_cgraph *
         } else if (tensor->op == GGML_OP_SCALE) {
             const float * params = (const float *)tensor->op_params;
             tensor_clone = ggml_scale_bias(ggml_ctx, src_clone[0], params[0], params[1]);
-        } else if (tensor->op == GGML_OP_ADD1) {
-            tensor_clone = ggml_add1(ggml_ctx, src_clone[0], src_clone[1]);
+        // NOTE: this is deprecated
+        //} else if (tensor->op == GGML_OP_ADD1) {
+        //    tensor_clone = ggml_add1(ggml_ctx, src_clone[0], src_clone[1]);
```
0cc4m: This is debug code, not really required to be kept up to date. It's not included in a regular build. I assume the function will eventually get removed? We can remove this code at the same time. Until then it's probably best to keep it.

ggerganov: I commented out the code to avoid deprecation warnings such as this:

https://github.com/ggml-org/llama.cpp/actions/runs/23941987181/job/69830107172?pr=21363#step:3:1034

If we keep it, I think it will trip the Vulkan CI in a similar way?

0cc4m: No, it is not intended to be run on the CI. It exists to allow manually figuring out why a specific model run has issues. To enable you build with `GGML_VULKAN_CHECK_RESULTS=1` and it will run each operation in the compute graph on CPU and on GPU and compare the results.

0cc4m (approved):

PR #21390, opened 2026-04-03
llama-model: read final_logit_softcapping for Gemma 4

ssam18:
The `LLM_ARCH_GEMMA4` block in `llama-model.cpp` was never reading `final_logit_softcapping` from the GGUF, so the value was always stuck at the hardcoded default of `30.0f`. This meant editing the GGUF key or using `--override-kv gemma4.final_logit_softcapping=float:X` had no effect on inference.

Adding the missing `ml.get_key` call (optional, so older GGUFs without the key fall back gracefully to `30.0f`) is all that's needed, the softcapping logic in `gemma4-iswa.cpp` is already correct.

Fix for the issue #21388.


BugReporterZ:
I tried this one-line change, but if I don't override manually the final logit soft cap e.g. with `--override-kv gemma4.final_logit_softcapping=float:30.0` (default value for Gemma 4) I get gibberish.

BugReporterZ:
My bad, you're right. I mistakenly targeted the wrong GGUF file (the one I had originally modified for the bug report) in that test. It seems to work correctly with a quantization from Unsloth.

CISC (approved):

ngxson (approved):

PR #21405, opened 2026-04-04
vendor : update cpp-httplib to 0.41.0

cabelo:
## Additional information

# Requirements

- AI usage disclosure: NO 

## What's Changed

- Replace httplib::any / std::map<std::string, httplib::any> based res.user_data with a 
  new type-safe UserData class that provides set() / get<T>() API
- Add parse_url utility function with UrlComponents struct for parsing URLs into 
  scheme, host, port, path, and query components
- [cmake] Allow using pre-existing zstd::libzstd target if it already exists, useful for
  projects that bundle their own zstd
- Add header parser and URL parser fuzzers for OSS-Fuzz coverage improvement

### DIFF
--- a/scripts/sync_vendor.py
+++ b/scripts/sync_vendor.py
@@ -5,7 +5,7 @@ import os
 import sys
 import subprocess
 
-HTTPLIB_VERSION = "refs/tags/v0.40.0"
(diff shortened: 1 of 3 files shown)


cabelo:
New PR https://github.com/ggml-org/llama.cpp/pull/21781

PR #21407, opened 2026-04-04
cmake: add flag to use system httplib

WhyNotHugo:
## Overview

Allow building using the system cpp-httplib. The default behaviour remains the same.

## Additional information

Downstream distributions prefer to use a single copy of libraries like this one, instead of rebuilding it for each project.

# Requirements

- [x] AI usage disclosure.


WhyNotHugo:
Kind ping, maybe this slipped through the cracks?

WhyNotHugo:
Rebased to latest master.

PR #21411, opened 2026-04-04
fix(openvino): explicit ov::Tensor frees in ggml_backend_openvino_free

thedanhoffman:
## Overview

The following throws a segmentation fault on exit.

```
export GGML_OPENVINO_DEVICE=NPU
./bin/llama-cli -m Llama-3.2-1B-Instruct-Q4_0.gguf -ngl 99 -np 1 -c 512 --simple-io --single-turn -n 1 -p hi
```

Root-caused to a use-after-free involving OpenVINO's `ZeroMemPool`. Goes like this
1. `ZeroMemPool::allocate_zero_memory()` returns `std::shared_ptr<ZeroMem>`
2. `std::shared_ptr<ZeroMem>` has a custom destructor method of the form `[this](ZeroMem *ptr) { delete_pool_entry(ptr); }` (where `delete_pool_entry` is a non-static member function of `ZeroMemPool`)
3. `this` is invalid by the time the destructor runs because RAII does not clean up `std::shared_ptr<ZeroMem>` consistently

This is downstream of `std::shared_ptr<void>` being used for `runtime_context` not propagating the destructor call to class members.

Went ahead and defined a `clear_caches` method that does this destruction explicitly (but not using `std::shared_ptr<void>` seems like a better approach considering the failure mode here, will leave that to the maintainers unless they suggest I change that here).

## Additional information

This PR (and all other PRs I'll file in this space) use a recent from-source build of OpenVINO. I do work at Intel but not on the team owning this.

# Requirements

- AI usage disclosure: I threw an AI at this, but I have a solid understanding of the mechanics at play and the fix is targeted enough to not cause collateral issues.

### DIFF
--- a/ggml/src/ggml-openvino/ggml-openvino.cpp
+++ b/ggml/src/ggml-openvino/ggml-openvino.cpp
@@ -598,6 +598,15 @@ bool ggml_backend_buft_is_openvino_host(ggml_backend_buffer_type_t buft) {
 
 static void ggml_backend_openvino_free(ggml_backend_t backend) {
     ggml_backend_openvino_context * ctx = (ggml_backend_openvino_context *) backend->context;
+
+    if (ctx->runtime_context) {
+        auto r_ctx = std::static_pointer_cast<ov_runtime_context>(ctx->runtime_context);
+        r_ctx->backend_count--;
+        if (r_ctx->backend_count == 0) {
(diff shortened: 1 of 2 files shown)


thedanhoffman:
Closing to PR against Ravi's branch directly

PR #21421, opened 2026-04-04
mtmd: add Gemma 4 audio conformer encoder support

stephencox-ict:
> [!IMPORTANT]
> It is recommended to use BF16 mmproj. Other quantizations are known to have degraded performance; ref comment: https://github.com/ggml-org/llama.cpp/pull/21421#issuecomment-4230306463

## Overview

Add audio processing support for Gemma 4 models via a USM-style Conformer encoder.

**Architecture:**
- 12-layer Conformer: FFN -> Self-Attention -> Causal Conv1D -> FFN -> Norm
- Subsampling Conv Projection: 2x Conv2D(stride=2) with LayerNorm
- Chunked local attention with sinusoidal RPE (chunk_size=12, context_size=24)
- Logit softcapping at 50.0, ClippableLinear with per-tensor clamping
- Output projection -> RMSNorm -> multimodal embedder

**Chunked local attention** (matching PyTorch reference):
- Q split into non-overlapping blocks of 12
- K/V extracted as overlapping context windows of 24 via `ggml_view_4d` with stride 12
- Per-block causal mask matching PyTorch's `dist < left_window_size` condition
- Blocked relative position shift (Transformer-XL appendix B)
- RPE: 13 sinusoidal position embeddings [12, 11, ..., 0]

**Mel preprocessing** (dedicated `mtmd_audio_preprocessor_gemma4a`):
- HTK mel scale, 128 bins, magnitude STFT, mel_floor=1e-3
- Standard periodic Hann window (320 samples), zero-padded to FFT size
- 30-second chunking (splits long audio into 30s segments)
- Mel cosine similarity vs PyTorch: **1.0**

**Fixes (beyond the initial encoder):**

1. **Contiguous sigmoid input** (`gemma4a.cpp`): Wrap GLU gate view in `ggml_cont()` before `ggml_sigmoid()`. The non-contiguous view caused CUDA/Vulkan to fall back to CPU for sigmoid, creating 25 graph splits and numerical divergence on longer audio.

2. **Conv norm swap at load site** (`clip.cpp`): The upstream `tensor_mapping.py` maps the gemma4 audio lconv1d norms with swapped GGUF names (`conv_norm` ↔ `norm_conv`). The loader now loads tensors in reverse order at the load site to correct this, rather than swapping post-load. Verified by element-wise comparison against Python transformers safetensors weights.

3. **Replace `ggml_roll` with `ggml_view` + `ggml_concat`** (`gemma4a.cpp`): The `ggml_roll` op has no Metal kernel, causing 73 graph splits and CPU fallbacks on Apple Silicon. Replaced with equivalent view+concat sequences that are supported on all backends. Audio encoder graph splits reduced from 73 to 1 on Metal.

**Usage:**
```bash
llama-mtmd-cli \
    -m gemma-4-E2B-it-Q6_K.gguf \
    --mmproj mmproj-BF16.gguf \
    --audio sample.wav \
    -p "Transcribe this audio exactly." \
    --temp 1.0 --top-k 64 --top-p 0.95 \
    -ngl 99 --jinja
```

**Audio transcription results (E2B, best across CPU/Vulkan/CUDA):**

Short audio (5.9s LibriSpeech):
```
Ground truth: "MISTER QUILTER IS THE APOSTLE OF THE MIDDLE CLASSES AND WE ARE GLAD TO WELCOME HIS GOSPEL"
Output:       "Mr. Quilter is the apostle of the middle classes and we are glad to welcome his gospel."
```

Long audio (17.4s, moon landing narration):
```
Output: "This is the moment when humanity's oldest dream became reality.
         This is the New York Times's boldest headline across America's newspaper record.
         The New York Times has documented our nation's most pivotal moments, but rarely
         has any story matched the cosmic significance of..."
```

| Quant | Short (5.9s) | Long (17.4s) |
|-------|-------------|-------------|
| BF16 | ✅ | ✅ |
| Q8_0 | ✅ | ✅ |
| UD-Q8_K_XL | ✅ | ✅ |
| Q6_K | ✅ | ✅ |
| Q5_K_M | ✅ | ✅ |
| Q5_K_S | ✅ | ✅ |
| Q4_0 | ✅ | ✅ |
| IQ4_NL | ✅ | ✅ |
| Q4_K_M | ✅ | ⏳ |
| Q4_K_S | ✅ | ⏳ |
| Q4_1 | ✅ | ⏳ |
| IQ4_XS | ✅ | ⏳ |
| Q3_K_M | ✅ | 🔶 |
| Q3_K_S | ✅ | ❌ |

**E2B short audio: 14/14 PASS.** All quantizations correctly transcribe across all backends.

⏳ = model's thinking block consumed all tokens before outputting the transcription. Higher `-n` values resolve this.

**E4B also tested: 19/21 short PASS, 20/21 long PASS/PARTIAL.** Only ultra-low 2-bit quants (UD-IQ2_M, UD-IQ3_XXS) fail. See [PR comment](https://github.com/ggml-org/llama.cpp/pull/21421#issuecomment-4222421146) for full E4B matrix.

**Resampling note:** For best audio quality, provide input already at 16kHz. Audio at other sample rates will be downsampled using miniaudio's linear resampler regardless of format (WAV, MP3, FLAC).

**Generation parameters** (from model's `generation_config.json`):
`--temp 1.0 --top-k 64 --top-p 0.95`

## Additional information

**Test plan:**
- [x] `test-mtmd-c-api` passes
- [x] `test-llama-archs` passes
- [x] Full ctest suite (40/40 tests pass)
- [x] E2B all 14 quants: short audio PASS on CPU, Vulkan, CUDA
- [x] E2B BF16/Q6_K/Q8_0/Q5_K_M: long audio PASS on CPU, Vulkan, CUDA
- [x] E4B 19/21 quants: short audio PASS, 20/21 long PASS/PARTIAL
- [x] Mel values verified against PyTorch (cosine 1.0)
- [x] Encoder output verified against Python transformers (element-wise)
- [x] Conv norm mapping verified against safetensors weights
- [x] Tested with BF16 mmproj (recommended; F16/Q8_0 cause repetitions due to ClippableLinear sensitivity)

**Dependency:** #21625 (per-layer embedding scale for multimodal path) improves longer audio transcription reliability on longer audio (~17s+).

Ref: #21325
Related: #21599 (quantization fix: force Q6_K minimum for Gemma4 tied embeddings)
Related: #21612 (merged: perform per-layer projections in the first layer)
Related: #21625 (dependency: per-layer embedding scale for multimodal path)

# Requirements

- AI usage disclosure: YES - Claude Code and Gemini were used in an assistive capacity for iterative debugging (tensor comparison, mel spectrogram verification, conformer layer tracing, quantization analysis) and code review. All architecture decisions, algorithm implementations, and code were manually reviewed and verified against the PyTorch reference.

### DIFF
--- a/gguf-py/gguf/tensor_mapping.py
+++ b/gguf-py/gguf/tensor_mapping.py
@@ -2066,22 +2066,22 @@ class TensorNameMap:
 
         MODEL_TENSOR.A_ENC_CONV_NORM: (
             "conformer.layers.{bid}.conv.batch_norm", # lfm2
-            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma3n
+            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma4
         ),
 
         MODEL_TENSOR.A_ENC_CONV_PW1: (
             "conformer.layers.{bid}.conv.pointwise_conv1", # lfm2
-            "conformer.layers.{bid}.lconv1d.linear_start", # gemma3n
+            "conformer.layers.{bid}.lconv1d.linear_start", # gemma4
         ),
 
         MODEL_TENSOR.A_ENC_CONV_PW2: (
             "conformer.layers.{bid}.conv.pointwise_conv2", # lfm2
-            "conformer.layers.{bid}.lconv1d.linear_end", # gemma3n
+            "conformer.layers.{bid}.lconv1d.linear_end", # gemma4
         ),
 
         MODEL_TENSOR.A_ENC_NORM_CONV: (
             "conformer.layers.{bid}.norm_conv", # lfm2
-            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma3n
+            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma4
         ),
 
         MODEL_TENSOR.A_PER_DIM_K_SCALE: (
--- a/gguf-py/gguf/tensor_mapping.py
+++ b/gguf-py/gguf/tensor_mapping.py
@@ -2066,22 +2066,22 @@ class TensorNameMap:
 
         MODEL_TENSOR.A_ENC_CONV_NORM: (
             "conformer.layers.{bid}.conv.batch_norm", # lfm2
-            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma4
+            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma3n, gemma4
         ),
 
         MODEL_TENSOR.A_ENC_CONV_PW1: (
             "conformer.layers.{bid}.conv.pointwise_conv1", # lfm2
-            "conformer.layers.{bid}.lconv1d.linear_start", # gemma4
+            "conformer.layers.{bid}.lconv1d.linear_start", # gemma3n, gemma4
         ),
 
         MODEL_TENSOR.A_ENC_CONV_PW2: (
             "conformer.layers.{bid}.conv.pointwise_conv2", # lfm2
-            "conformer.layers.{bid}.lconv1d.linear_end", # gemma4
+            "conformer.layers.{bid}.lconv1d.linear_end", # gemma3n, gemma4
         ),
 
         MODEL_TENSOR.A_ENC_NORM_CONV: (
             "conformer.layers.{bid}.norm_conv", # lfm2
-            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma4
+            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma3n, gemma4
         ),
 
         MODEL_TENSOR.A_PER_DIM_K_SCALE: (
--- a/gguf-py/gguf/tensor_mapping.py
+++ b/gguf-py/gguf/tensor_mapping.py
@@ -2066,7 +2066,7 @@ class TensorNameMap:
 
         MODEL_TENSOR.A_ENC_CONV_NORM: (
             "conformer.layers.{bid}.conv.batch_norm", # lfm2
-            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma3n, gemma4
+            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma3n, gemma4
         ),
 
         MODEL_TENSOR.A_ENC_CONV_PW1: (
@@ -2081,7 +2081,7 @@ class TensorNameMap:
 
         MODEL_TENSOR.A_ENC_NORM_CONV: (
             "conformer.layers.{bid}.norm_conv", # lfm2
-            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma3n, gemma4
+            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma3n, gemma4
         ),
 
         MODEL_TENSOR.A_PER_DIM_K_SCALE: (
--- a/gguf-py/gguf/tensor_mapping.py
+++ b/gguf-py/gguf/tensor_mapping.py
@@ -2066,22 +2066,22 @@ class TensorNameMap:
 
         MODEL_TENSOR.A_ENC_CONV_NORM: (
             "conformer.layers.{bid}.conv.batch_norm", # lfm2
-            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma3n, gemma4
+            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma3n
         ),
 
         MODEL_TENSOR.A_ENC_CONV_PW1: (
             "conformer.layers.{bid}.conv.pointwise_conv1", # lfm2
-            "conformer.layers.{bid}.lconv1d.linear_start", # gemma3n, gemma4
+            "conformer.layers.{bid}.lconv1d.linear_start", # gemma3n
         ),
 
         MODEL_TENSOR.A_ENC_CONV_PW2: (
             "conformer.layers.{bid}.conv.pointwise_conv2", # lfm2
-            "conformer.layers.{bid}.lconv1d.linear_end", # gemma3n, gemma4
+            "conformer.layers.{bid}.lconv1d.linear_end", # gemma3n
         ),
 
         MODEL_TENSOR.A_ENC_NORM_CONV: (
             "conformer.layers.{bid}.norm_conv", # lfm2
-            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma3n, gemma4
+            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma3n
         ),
 
         MODEL_TENSOR.A_PER_DIM_K_SCALE: (
--- a/src/llama-model.cpp
+++ b/src/llama-model.cpp
@@ -1517,14 +1517,16 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                 uint32_t swa_period = 2;
                 ml.get_key_or_arr(LLM_KV_ATTENTION_SLIDING_WINDOW_PATTERN, swa_period, false);
                 hparams.set_swa_pattern(swa_period);
-                hparams.attn_soft_cap = true;
                 hparams.rope_freq_base_train_swa  = hparams.rope_freq_base_train;
                 hparams.rope_freq_scale_train_swa = hparams.rope_freq_scale_train;
 
                 ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA,          hparams.rope_freq_base_train_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW,    hparams.n_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS, hparams.f_norm_rms_eps);
-                ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING,      hparams.f_attn_logit_softcapping, false);
+                // Gemma4 does NOT use attention logit softcapping (unlike Gemma2)
+                hparams.f_attn_logit_softcapping = 0.0f;
+                ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING, hparams.f_attn_logit_softcapping, false);
+                hparams.attn_soft_cap = (hparams.f_attn_logit_softcapping > 0.0f);
                 ml.get_key(LLM_KV_FINAL_LOGIT_SOFTCAPPING,     hparams.f_final_logit_softcapping, false);
 
                 switch (hparams.n_layer) {
--- a/src/llama-model.cpp
+++ b/src/llama-model.cpp
@@ -1517,13 +1517,14 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                 uint32_t swa_period = 2;
                 ml.get_key_or_arr(LLM_KV_ATTENTION_SLIDING_WINDOW_PATTERN, swa_period, false);
                 hparams.set_swa_pattern(swa_period);
+                hparams.attn_soft_cap = true;
                 hparams.rope_freq_base_train_swa  = hparams.rope_freq_base_train;
                 hparams.rope_freq_scale_train_swa = hparams.rope_freq_scale_train;
 
                 ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA,          hparams.rope_freq_base_train_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW,    hparams.n_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS, hparams.f_norm_rms_eps);
-                hparams.attn_soft_cap = ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING, hparams.f_attn_logit_softcapping, false);
+                ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING,      hparams.f_attn_logit_softcapping, false);
                 ml.get_key(LLM_KV_FINAL_LOGIT_SOFTCAPPING,     hparams.f_final_logit_softcapping, false);
 
                 switch (hparams.n_layer) {
--- a/src/llama-model.cpp
+++ b/src/llama-model.cpp
@@ -1523,10 +1523,7 @@ void llama_model::load_hparams(llama_model_loader & ml) {
                 ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA,          hparams.rope_freq_base_train_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW,    hparams.n_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS, hparams.f_norm_rms_eps);
-                // Gemma4 does NOT use attention logit softcapping (unlike Gemma2)
-                hparams.f_attn_logit_softcapping = 0.0f;
-                ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING, hparams.f_attn_logit_softcapping, false);
-                hparams.attn_soft_cap = (hparams.f_attn_logit_softcapping > 0.0f);
+                hparams.attn_soft_cap = ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING, hparams.f_attn_logit_softcapping, false);
                 ml.get_key(LLM_KV_FINAL_LOGIT_SOFTCAPPING,     hparams.f_final_logit_softcapping, false);
 
                 switch (hparams.n_layer) {
--- a/src/models/gemma4-iswa.cpp
+++ b/src/models/gemma4-iswa.cpp
@@ -17,8 +17,7 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
     inpL = build_inp_embd(model.tok_embd);
 
     // important: do not normalize weights for raw embeddings input (i.e. encoded image emdeddings)
-    // use BF16-rounded scale to match PyTorch's native BF16 training precision (ref: PR #21451)
-    inpL = ggml_scale(ctx0, inpL, ubatch.token ? ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf(n_embd))) : 1.0f);
+    inpL = ggml_scale(ctx0, inpL, ubatch.token ? sqrtf(n_embd) : 1.0f);
     cb(inpL, "inp_scaled", -1);
 
     // inp_pos - contains the positions
@@ -151,7 +150,7 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
 
             // custom MoE logits calculation (router operates on attn_out, not cur)
             ggml_tensor * tmp = ggml_rms_norm(ctx0, attn_out, hparams.f_norm_rms_eps);
-            tmp = ggml_scale(ctx0, tmp, 1.0f / ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf((float) n_embd))));
+            tmp = ggml_scale(ctx0, tmp, 1.0f / sqrtf((float) n_embd));
             tmp = ggml_mul(ctx0, tmp, model.layers[il].ffn_gate_inp_s);
             ggml_tensor * logits = build_lora_mm(model.layers[il].ffn_gate_inp, tmp); // [n_expert, n_tokens]
             cb(logits, "ffn_moe_logits", il);
--- a/src/models/gemma4-iswa.cpp
+++ b/src/models/gemma4-iswa.cpp
@@ -150,7 +150,6 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
             cb(cur_moe, "ffn_norm_2", il);
 
             // custom MoE logits calculation (router operates on attn_out, not cur)
-            // use BF16-rounded scale to match PyTorch's native BF16 training precision (ref: PR #21451)
             ggml_tensor * tmp = ggml_rms_norm(ctx0, attn_out, hparams.f_norm_rms_eps);
             tmp = ggml_scale(ctx0, tmp, 1.0f / ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf((float) n_embd))));
             tmp = ggml_mul(ctx0, tmp, model.layers[il].ffn_gate_inp_s);
--- a/src/models/gemma4-iswa.cpp
+++ b/src/models/gemma4-iswa.cpp
@@ -17,7 +17,8 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
     inpL = build_inp_embd(model.tok_embd);
 
     // important: do not normalize weights for raw embeddings input (i.e. encoded image emdeddings)
-    inpL = ggml_scale(ctx0, inpL, ubatch.token ? sqrtf(n_embd) : 1.0f);
+    // use BF16-rounded scale to match PyTorch's native BF16 training precision (ref: PR #21451)
+    inpL = ggml_scale(ctx0, inpL, ubatch.token ? ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf(n_embd))) : 1.0f);
     cb(inpL, "inp_scaled", -1);
 
     // inp_pos - contains the positions
@@ -149,8 +150,9 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
             cb(cur_moe, "ffn_norm_2", il);
 
             // custom MoE logits calculation (router operates on attn_out, not cur)
+            // use BF16-rounded scale to match PyTorch's native BF16 training precision (ref: PR #21451)
             ggml_tensor * tmp = ggml_rms_norm(ctx0, attn_out, hparams.f_norm_rms_eps);
-            tmp = ggml_scale(ctx0, tmp, 1.0f / sqrtf((float) n_embd));
+            tmp = ggml_scale(ctx0, tmp, 1.0f / ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf((float) n_embd))));
             tmp = ggml_mul(ctx0, tmp, model.layers[il].ffn_gate_inp_s);
             ggml_tensor * logits = build_lora_mm(model.layers[il].ffn_gate_inp, tmp); // [n_expert, n_tokens]
             cb(logits, "ffn_moe_logits", il);
--- a/src/models/gemma4-iswa.cpp
+++ b/src/models/gemma4-iswa.cpp
@@ -17,8 +17,7 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
     inpL = build_inp_embd(model.tok_embd);
 
     // important: do not normalize weights for raw embeddings input (i.e. encoded image emdeddings)
-    // use BF16-rounded scale to match PyTorch's native BF16 training precision
-    inpL = ggml_scale(ctx0, inpL, ubatch.token ? ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf(n_embd))) : 1.0f);
+    inpL = ggml_scale(ctx0, inpL, ubatch.token ? sqrtf(n_embd) : 1.0f);
     cb(inpL, "inp_scaled", -1);
 
     // inp_pos - contains the positions
@@ -150,9 +149,8 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
             cb(cur_moe, "ffn_norm_2", il);
 
             // custom MoE logits calculation (router operates on attn_out, not cur)
-            // use BF16-rounded scale to match PyTorch's native BF16 training precision
             ggml_tensor * tmp = ggml_rms_norm(ctx0, attn_out, hparams.f_norm_rms_eps);
-            tmp = ggml_scale(ctx0, tmp, 1.0f / ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf((float) n_embd))));
+            tmp = ggml_scale(ctx0, tmp, 1.0f / sqrtf((float) n_embd));
             tmp = ggml_mul(ctx0, tmp, model.layers[il].ffn_gate_inp_s);
             ggml_tensor * logits = build_lora_mm(model.layers[il].ffn_gate_inp, tmp); // [n_expert, n_tokens]
             cb(logits, "ffn_moe_logits", il);
@@ -284,9 +282,7 @@ ggml_tensor * llm_build_gemma4_iswa::build_inp_per_layer() {
         // TODO: verify if this is the correct behavior in transformers implementation
         const int64_t embd_size = model.per_layer_tok_embd->ne[0];  // n_embd_per_layer * n_layer
 
-        // Extract and dequantize padding token embedding (row 0).
-        // PyTorch replaces multimodal IDs with pad_token_id before lookup,
-        // then ScaledWordEmbedding scales by sqrt(n_embd_per_layer).
+        // Extract and dequantize padding token embedding (row 0)
         ggml_tensor * padding = ggml_view_1d(ctx0, model.per_layer_tok_embd, embd_size, 0);
         inp_per_layer = ggml_cast (ctx0, padding, GGML_TYPE_F32);
         inp_per_layer = ggml_scale(ctx0, inp_per_layer, tok_embd_scale);
--- a/src/models/gemma4-iswa.cpp
+++ b/src/models/gemma4-iswa.cpp
@@ -17,7 +17,8 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
     inpL = build_inp_embd(model.tok_embd);
 
     // important: do not normalize weights for raw embeddings input (i.e. encoded image emdeddings)
-    inpL = ggml_scale(ctx0, inpL, ubatch.token ? sqrtf(n_embd) : 1.0f);
+    // use BF16-rounded scale to match PyTorch's native BF16 training precision
+    inpL = ggml_scale(ctx0, inpL, ubatch.token ? ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf(n_embd))) : 1.0f);
     cb(inpL, "inp_scaled", -1);
 
     // inp_pos - contains the positions
@@ -149,8 +150,9 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
             cb(cur_moe, "ffn_norm_2", il);
 
             // custom MoE logits calculation (router operates on attn_out, not cur)
+            // use BF16-rounded scale to match PyTorch's native BF16 training precision
             ggml_tensor * tmp = ggml_rms_norm(ctx0, attn_out, hparams.f_norm_rms_eps);
-            tmp = ggml_scale(ctx0, tmp, 1.0f / sqrtf((float) n_embd));
+            tmp = ggml_scale(ctx0, tmp, 1.0f / ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf((float) n_embd))));
             tmp = ggml_mul(ctx0, tmp, model.layers[il].ffn_gate_inp_s);
             ggml_tensor * logits = build_lora_mm(model.layers[il].ffn_gate_inp, tmp); // [n_expert, n_tokens]
             cb(logits, "ffn_moe_logits", il);
@@ -282,7 +284,9 @@ ggml_tensor * llm_build_gemma4_iswa::build_inp_per_layer() {
         // TODO: verify if this is the correct behavior in transformers implementation
         const int64_t embd_size = model.per_layer_tok_embd->ne[0];  // n_embd_per_layer * n_layer
 
-        // Extract and dequantize padding token embedding (row 0)
+        // Extract and dequantize padding token embedding (row 0).
+        // PyTorch replaces multimodal IDs with pad_token_id before lookup,
+        // then ScaledWordEmbedding scales by sqrt(n_embd_per_layer).
         ggml_tensor * padding = ggml_view_1d(ctx0, model.per_layer_tok_embd, embd_size, 0);
         inp_per_layer = ggml_cast (ctx0, padding, GGML_TYPE_F32);
         inp_per_layer = ggml_scale(ctx0, inp_per_layer, tok_embd_scale);
--- a/tests/test-llama-archs.cpp
+++ b/tests/test-llama-archs.cpp
@@ -88,6 +88,11 @@ static gguf_context_ptr get_gguf_ctx(const llm_arch arch, const bool moe) {
     uint32_t n_layer = 2;
     if (arch == LLM_ARCH_LLAMA4) {
         n_layer = 4; // hparams.n_no_rope_layer_step is hard-coded to 4
+    } else if (arch == LLM_ARCH_GEMMA4) {
+        n_embd = 128;
+        n_head = 2;
+        n_ff   = 192;
+        n_layer = 5; // need at least 5 for swa_pattern (every 5th is full_attention)
     } else if (arch == LLM_ARCH_GEMMA3N) {
         n_embd = 64;
         n_head = 1;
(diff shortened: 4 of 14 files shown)


tools/mtmd/mtmd-audio.cpp
```diff
@@ -464,8 +492,8 @@ static bool log_mel_spectrogram(
                 out.data[i * out.n_len + j] = 0.0;
             }
         }
-    } else {
-        // clamping and normalization
+    } else if (!params.skip_global_norm) {
+        // Whisper-style clamping and normalization (NOT used by Gemma4)
```
stephencox-ict, line 491: Cleanup comment

tools/mtmd/mtmd-audio.cpp
```diff
@@ -623,6 +655,17 @@ bool mtmd_audio_preprocessor_conformer::preprocess(const float *
         return false;
     }
 
+    // Debug: print mel stats
```
stephencox-ict: Remove debug code

README.md
```diff
@@ -157,6 +157,7 @@ Instructions for adding support for new models: [HOWTO-add-model.md](docs/develo
 - [x] [GLM-EDGE](https://huggingface.co/models?search=glm-edge)
 - [x] [Qwen2-VL](https://huggingface.co/collections/Qwen/qwen2-vl-66cee7455501d7126940800d)
 - [x] [LFM2-VL](https://huggingface.co/collections/LiquidAI/lfm2-vl-68963bbc84a610f7638d5ffa)
+- [x] [Gemma 4](https://huggingface.co/google/gemma-4-E4B-it) (vision + audio)
```
stephencox-ict: Update link to https://huggingface.co/collections/google/gemma-4

tools/mtmd/mtmd.cpp
```diff
@@ -1247,6 +1264,20 @@ mtmd_input_chunks * mtmd_test_create_input_chunks() {
     };
     chunks->entries.emplace_back(std::move(chunk_image));
 
+    // create an audio chunk
+    mtmd_audio_tokens_ptr audio_tokens(new mtmd_audio_tokens);
+    audio_tokens->n_tokens = 750;
```
stephencox-ict: This looks odd

tools/mtmd/mtmd-audio.h
```diff
@@ -73,6 +74,17 @@ struct mtmd_audio_preprocessor_conformer : mtmd_audio_preprocessor {
     void initialize() override;
     bool preprocess(const float * samples, size_t n_samples, std::vector<mtmd_audio_mel> & output) override;
 
+    // Configurable params (set before initialize())
```
ngxson: instead of adding to `mtmd_audio_preprocessor_conformer`, it's better to create a dedicated `mtmd_audio_preprocessor_gemma4`

docs/multimodal/gemma4.md
```diff
@@ -0,0 +1,50 @@
+# Gemma 4 Multimodal (Vision + Audio)
+
+Gemma 4 is a family of multimodal models from Google. The **E2B** and **E4B** variants support text, vision, and audio inputs. The audio encoder is a USM-style Conformer.
+
+## Quick Start
+
+```sh
+# E4B (larger, better quality)
```
stephencox-ict: Need to document the other models too

ngxson: IMO these documentations are unrelated changes and better to be in a dedicated PR.

Please keep the current PR as simple to review as possible

tools/mtmd/mtmd-audio.h
ngxson: you may want to look at my version of the preprocessor too (better matching the original implementation)

https://gist.github.com/ngxson/dccf4784494454cba9a5cbebeaf95a4b

stephencox-ict: Have a look now

docs/multimodal/gemma4.md
stephencox-ict: Removed

tools/mtmd/clip.cpp
```diff
@@ -1519,16 +1534,22 @@ struct clip_model_loader {
         }
 
         // helper function
+        std::set<std::string> loaded_tensor_names;
         auto get_tensor = [&](const std::string & name, bool required = true) {
+            // Prevent double-loading the same tensor
+            if (loaded_tensor_names.count(name)) {
+                ggml_tensor * existing = ggml_get_tensor(ctx_clip.ctx_data.get(), name.c_str());
+                return existing;
+            }
```
ngxson, line 1580: better to throw an error here instead, and modify your code so that it doesn't load a tensor twice.

allow double-loading will make it more error-prone in the future.

stephencox-ict: Fixed

stephencox-ict: Resolving as I believe this is dealt with. @ngxson please reopen if not.

tools/mtmd/mtmd-audio.h
stephencox-ict: Resolving as I believe this is dealt with. @ngxson please reopen if not.

tests/test-llama-archs.cpp
```diff
@@ -167,7 +172,15 @@ static gguf_context_ptr get_gguf_ctx(const llm_arch arch, const bool moe) {
     ms.add_kv(LLM_KV_ATTENTION_RELATIVE_BUCKETS_COUNT, uint32_t(8));
     ms.add_kv(LLM_KV_ATTENTION_SLIDING_WINDOW,         n_ctx/8);
 
-    if (arch == LLM_ARCH_MIMO2 || arch == LLM_ARCH_STEP35) {
+    if (arch == LLM_ARCH_GEMMA4) {
+        ms.add_kv(LLM_KV_EMBEDDING_LENGTH_PER_LAYER,      n_embd/2);
+        ms.add_kv(LLM_KV_ATTENTION_SHARED_KV_LAYERS,      uint32_t(0));
+        ms.add_kv(LLM_KV_ATTENTION_KEY_LENGTH_SWA,        n_embd_head);
+        ms.add_kv(LLM_KV_ATTENTION_VALUE_LENGTH_SWA,      n_embd_head);
+        ms.add_kv(LLM_KV_ROPE_FREQ_BASE_SWA,             10000.0f);
```
JohannesGaessler:
```suggestion
        ms.add_kv(LLM_KV_ROPE_FREQ_BASE_SWA,              10000.0f);
```

JohannesGaessler (commented):
The changes to `test-llama-archs.cpp` LGTM otherwise. For some of the other files I'm seeing though that you are adding code comments with EM dashes. Please stick to ASCII unless there is a good reason not to.

tools/mtmd/models/gemma4a.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+            // Softcap
+            scores = ggml_scale(ctx0, scores, 1.0f / softcap);
+            scores = ggml_tanh(ctx0, scores);
+            scores = ggml_scale(ctx0, scores, softcap);
+
+            // Block-local attention mask
+            ggml_tensor * kq_mask = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, n_pos, n_pos);
+            ggml_set_name(kq_mask, "kq_mask");
+            ggml_set_input(kq_mask);
```
ngxson: move any input tensors to the top of graph builder

tools/mtmd/clip-impl.h
```diff
@@ -176,6 +176,21 @@
...
+#define TN_A_CONV1D         "a.conv1d.%d.%s"
+#define TN_A_CONV1D_NORM    "a.conv1d.%d.norm.%s"
+#define TN_A_OUT_PROJ       "a.pre_encode.out.%s"
+#define TN_A_ATTN_PRE_NORM  "%s.blk.%d.attn_pre_norm.%s"
+#define TN_A_ATTN_POST_NORM "%s.blk.%d.attn_post_norm.%s"
+#define TN_A_ATTN_K_REL     "%s.blk.%d.attn_k_rel.%s"
+#define TN_A_PER_DIM_SCALE  "%s.blk.%d.per_dim_scale.%s"
+#define TN_A_PER_DIM_K_SCALE "%s.blk.%d.per_dim_k_scale.%s"
+#define TN_A_FFN_POST_NORM  "%s.blk.%d.ffn_post_norm.%s"
+#define TN_A_FFN_POST_NORM_1 "%s.blk.%d.ffn_post_norm_1.%s"
```
ngxson: vertical alignment

tools/mtmd/clip-model.h
```diff
@@ -218,6 +218,13 @@ struct clip_layer {
     ggml_tensor * conv_pw2_w    = nullptr;
     ggml_tensor * conv_pw2_b    = nullptr;
 
+    // gemma4 audio conformer per-layer
+    ggml_tensor * attn_pre_norm_w  = nullptr;
+    ggml_tensor * attn_k_rel_w     = nullptr;
+    ggml_tensor * per_dim_scale_w  = nullptr;
+    ggml_tensor * per_dim_k_scale_w = nullptr;
+    ggml_tensor * ff_post_norm_1_w = nullptr;
```
ngxson: vertical align

tools/mtmd/clip.cpp
```diff
@@ -2145,7 +2233,16 @@ struct clip_model_loader {
...
+                if (!cur) {
+                    LOG_ERR("TENSOR NOT FOUND in ctx_data: %s\n", t->name);
+                    continue;
+                }
+                auto it_off = tensor_offset.find(t->name);
+                if (it_off == tensor_offset.end()) {
+                    LOG_ERR("NO OFFSET for tensor: %s\n", t->name);
+                    continue;
+                }
+                const size_t offset = it_off->second;
```
ngxson: better to either throw or GGML_ASSERT here, any missing data is not expected

tools/mtmd/clip.cpp
```diff
@@ -3195,6 +3301,52 @@ bool clip_image_batch_encode(clip_ctx * ctx, const int n_threads, const clip_ima
...
+            {
+                GGML_ASSERT(imgs.entries.size() == 1);
+                const auto & img0 = imgs.entries.front();
+                // Compute n_pos matching SSCP output: two stride-2 convs
+                int n_pos = img0->nx;
+                for (int i = 0; i < 2; i++) { n_pos = (n_pos - 1) / 2 + 1; }
+
+                // Sliding window attention mask: each query sees context_size positions
+                // context_size = chunk_size + context_left - 1 = 12 + 12 = 24
+                if (ggml_graph_get_tensor(gf, "kq_mask")) {
```
ngxson: use `set_input_u/f32` like other models

input tensor must always present in cgraph, there is no need to check

tools/mtmd/clip.cpp
```diff
@@ -3195,6 +3301,52 @@ bool clip_image_batch_encode(clip_ctx * ctx, const int n_threads, const clip_ima
...
+                    for (int i = 0; i < num_timescales; i++) {
+                        float inv_ts = expf(-(float)i * log_timescale_increment);
+                        float scaled = position * inv_ts;
+                        pos_emb[p * n_embd + i]                 = sinf(scaled);
+                        pos_emb[p * n_embd + i + num_timescales] = cosf(scaled);
+                    }
+                }
+                if (ggml_graph_get_tensor(gf, "pos_emb")) {
+                    set_input_f32("pos_emb", pos_emb);
+                }
```
ngxson: same

tools/mtmd/mtmd-audio.cpp
```diff
@@ -627,6 +650,90 @@ bool mtmd_audio_preprocessor_conformer::preprocess(const float *
...
+    }
+
+    // HTK mel scale, no Slaney area normalization
+    cache.fill_mel_filterbank_matrix(
+        hparams.n_mel_bins, hparams.audio_n_fft, hparams.audio_sample_rate,
+        0.0f, hparams.audio_sample_rate / 2.0f,
+        /*slaney_area_norm=*/false,
+        /*scale=*/1.0f,
+        /*use_htk=*/true
+    );
```
ngxson:
```suggestion
        /*slaney_area_norm=*/ false,
        /*scale=*/ 1.0f,
        /*use_htk=*/ true
    );
```

tools/mtmd/mtmd-audio.cpp
```diff
@@ -627,6 +650,90 @@ bool mtmd_audio_preprocessor_conformer::preprocess(const float *
...
+}
+
+bool mtmd_audio_preprocessor_gemma4a::preprocess(const float *                 samples,
+                                                  size_t                        n_samples,
+                                                  std::vector<mtmd_audio_mel> & output) {
+    if (n_samples == 0) {
+        return false;
+    }
+
+    // Enforce 30-second maximum (model trained context limit, ~750 tokens)
```
ngxson: there should be a chunking logic like whisper, `output` contains chunks of 30s with silence padding

ngxson (changes requested):

tools/mtmd/clip.cpp
```diff
@@ -17,6 +17,7 @@
 #include <cstring>
 #include <fstream>
 #include <map>
+#include <set>
```
ngxson:
```suggestion
```

tools/mtmd/clip.cpp
```diff
@@ -1519,16 +1534,21 @@ struct clip_model_loader {
         }
 
         // helper function
+        std::set<std::string> loaded_tensor_names;
```
ngxson:
```suggestion
        std::unordered_set<std::string> loaded_tensor_names;
```

ngxson (commented):
Is there any other problems with the implementation? I ran the test again, but it still giving repeated output:

> 4.  **Perform the Transcription (Line-by-Line/Segment-by-Segment):**
> 
>     *   *(Initial segment - check punctuation and phrasing)*
>     *   "The New York Times from July 20th, 1969."
>     *   "This isn't just a newsprint, this is the dawn of a new era."
>     *   "Men walk on the moon."
>     *   "The New York Times from July 20th, 1969." (The text repeats this phrase)
>     *   "This isn't just a newsprint, this is the moment when humanity's oldest dream became a reality."
>     *   "Men walk on the moon."
>     *   "The New York Times from July 20th, 1969." (Again, the repetition)
>     *   "This isn't just a newsprint, this is the moment when humanity's oldest dream became a reality." (The repetition continues)

tools/mtmd/clip.cpp
stephencox-ict: Fixed

tools/mtmd/clip.cpp
stephencox-ict: Fixed

angelsu:
## Bug Found: `softplus_pds` input tensors never populated

I have been working on getting this PR to produce correct embeddings and found the root cause of the quality issues.

### The Bug

In `gemma4a.cpp`, the graph declares input tensors `softplus_pds_{il}` for each conformer layer:

```cpp
auto * pds = ggml_new_tensor_1d(ctx0, GGML_TYPE_F32, d_head);
ggml_set_name(pds, pds_name);
ggml_set_input(pds);
Qcur = ggml_mul(ctx0, Qcur, ggml_reshape_3d(ctx0, pds, d_head, 1, 1));
```

But in `clip.cpp`, the `set_input` section for `PROJECTOR_TYPE_GEMMA4A` only populates `kq_mask` and `pos_emb` — **`softplus_pds_*` tensors are never filled with data**. They remain zero-initialized, so all Q vectors in the attention get multiplied by ~0.

### The Fix

Add this after the `pos_emb` set_input block in clip.cpp:

```cpp
// Pre-compute softplus(per_dim_scale) for each conformer layer
{
    const int n_layer = ctx->model.hparams.n_layer;
    const int d_head  = ctx->model.hparams.n_embd / ctx->model.hparams.n_head;
    for (int il = 0; il < n_layer; il++) {
        const auto & layer = ctx->model.layers[il];
        if (!layer.per_dim_scale_w) continue;
        std::vector<float> pds_data(d_head);
        ggml_backend_tensor_get(layer.per_dim_scale_w, pds_data.data(), 0, d_head * sizeof(float));
        for (int i = 0; i < d_head; i++) {
            pds_data[i] = logf(1.0f + expf(pds_data[i]));
        }
        char pds_name[64];
        snprintf(pds_name, sizeof(pds_name), "softplus_pds_%d", il);
        set_input_f32(pds_name, pds_data);
    }
}
```

### Secondary Fix: NaN from -INFINITY mask

The attention mask uses `-INFINITY` as the masked value. For padded positions (where all scores are masked), `softmax(-inf, -inf, ...) = NaN`, which propagates through the network. Changing to `-1e9f` produces the same ~0 softmax output without NaN:

```cpp
std::vector<float> mask(context_size * chunk_size * num_blocks, -1e9f);
```

### Results

With both fixes applied (tested with E2B Q4_K_M):
- Conformer embeddings cosine similarity vs PyTorch reference: **0.87 → 0.9946**
- All 436 tokens now have cos > 0.90 (previously only 34%)
- Model can correctly transcribe audio and answer questions about it

Verified by building a PyTorch reference conformer using GGUF weights and comparing layer-by-layer outputs.

gguf-py/gguf/tensor_mapping.py
```diff
@@ -2045,22 +2045,22 @@ class TensorNameMap:
 
         MODEL_TENSOR.A_ENC_CONV_NORM: (
             "conformer.layers.{bid}.conv.batch_norm", # lfm2
-            "conformer.layers.{bid}.lconv1d.pre_layer_norm", # gemma3n
+            "conformer.layers.{bid}.lconv1d.conv_norm", # gemma4
```
ORippler: Does this mean we will drop support for converting gemma3n?

gguf-py/gguf/tensor_mapping.py
ngxson, line 1: Is it possible to fix the mapping in C++ code instead? i.e. only re-assign the tensor when loading the model.

Updating conversion script is not very desirable because user will have to re-generate gguf, and old gguf will break when this PR is merged

src/models/gemma4-iswa.cpp
```diff
@@ -138,6 +138,7 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
                     LLM_NORM_RMS, il);
             cb(cur_moe, "ffn_norm_2", il);
 
+            // custom MoE logits calculation (router operates on attn_out, not cur)
```
ngxson, line 151: duplicated line

src/models/gemma4-iswa.cpp
```diff
@@ -274,13 +275,15 @@ ggml_tensor * llm_build_gemma4_iswa::get_per_layer_inputs() {
         // TODO: verify if this is the correct behavior in transformers implementation
         const int64_t embd_size = model.tok_embd_per_layer->ne[0];  // n_embd_per_layer * n_layer
 
-        // Extract and dequantize padding token embedding (row 0)
+        // Extract and dequantize padding token embedding (row 0).
+        // PyTorch uses Gemma4TextScaledWordEmbedding which scales by sqrt(n_embd_per_layer).
         ggml_tensor * padding = ggml_view_1d(ctx0, model.tok_embd_per_layer, embd_size, 0);
         inp_per_layer = ggml_cast(ctx0, padding, GGML_TYPE_F32);
+        inp_per_layer = ggml_scale(ctx0, inp_per_layer, sqrtf((float) n_embd_per_layer));
```
ngxson: this is incorrect for vision. any scaling should be applied from projector before going to this

src/llama-model.cpp
```diff
@@ -1186,14 +1186,16 @@ void llama_model::load_hparams(llama_model_loader & ml) {
...
                 hparams.rope_freq_scale_train_swa = hparams.rope_freq_scale_train;
 
                 ml.get_key(LLM_KV_ROPE_FREQ_BASE_SWA,          hparams.rope_freq_base_train_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_SLIDING_WINDOW,    hparams.n_swa, false);
                 ml.get_key(LLM_KV_ATTENTION_LAYERNORM_RMS_EPS, hparams.f_norm_rms_eps);
-                ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING,      hparams.f_attn_logit_softcapping, false);
+                // Gemma4 does NOT use attention logit softcapping (unlike Gemma2)
+                hparams.f_attn_logit_softcapping = 0.0f;
+                ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING, hparams.f_attn_logit_softcapping, false);
+                hparams.attn_soft_cap = (hparams.f_attn_logit_softcapping > 0.0f);
```
ngxson: this should be in a dedicated PR

ngxson (changes requested):

gguf-py/gguf/tensor_mapping.py
stephencox-ict: Done. Reverted tensor_mapping.py back to upstream mappings (also fixed a missing comma that caused silent string concatenation in the V_ENC_ATTN_QKV tuple). The conv_norm <-> norm_conv fix is now in clip.cpp — the GEMMA4A loader cross-loads the swapped GGUF tensor names so existing GGUFs stay compatible.

src/models/gemma4-iswa.cpp
stephencox-ict: Fixed.

src/models/gemma4-iswa.cpp
stephencox-ict: Removed. Reverted the vision path back to upstream (no scaling, no comment changes).

src/llama-model.cpp
stephencox-ict: Reverted back to upstream.

ngxson:
Can you run transcription `tools/mtmd/test-2.mp3` an report the result here?

src/models/gemma4-iswa.cpp
stephencox-ict: Investigated this. The scaling must stay in the text model — it can't be moved to the projector.

In the PyTorch reference (`Gemma4Model.forward()`), multimodal token IDs are replaced with `pad_token_id` before calling `get_per_layer_inputs()`:
```python
llm_input_ids[multimodal_mask] = self.config.text_config.pad_token_id
```

Then `Gemma4TextScaledWordEmbedding` looks up the padding token in `per_layer_token_embeddings` and applies `sqrt(hidden_size_per_layer_input)` automatically — it's a property of the text model's embedding layer, not something the projector controls.

Our implementation matches: we use padding token (row 0) from `tok_embd_per_layer` and scale by `sqrtf(n_embd_per_layer)`. The token path (line 269) does the exact same scaling. The projector can't do this because:
1. `tok_embd_per_layer` is a text model weight — the projector has no access
2. The batch API has no mechanism to pass per-layer auxiliary data from the projector

Ref: https://github.com/huggingface/transformers/issues/45206

Removed the duplicated comment on line 141 as requested.

src/llama-model.cpp
stephencox-ict: Kept in this PR — the hardcoded `attn_soft_cap = true` causes incorrect attention softcapping for Gemma 4 (which has no `attn_logit_softcapping` key in its GGUF, but the struct default is 50.0). This breaks audio transcription on longer audio.

Using `ml.get_key()` return value: `hparams.attn_soft_cap = ml.get_key(LLM_KV_ATTN_LOGIT_SOFTCAPPING, ...)`. Gemma 2 GGUFs have the key (returns true), Gemma 4 GGUFs don't (returns false). Also submitted as standalone PR #21613.

tests/test-llama-archs.cpp
stephencox-ict: Already applied, thanks.

tools/mtmd/clip.cpp
```diff
@@ -2283,6 +2283,46 @@ struct clip_model_loader {
...
+
+            // In correctly-mapped GGUFs, conv_norm (post-conv) has lower magnitude
+            // than norm_conv (pre-conv/pre_layer_norm). If conv_norm has higher
+            // magnitude, the mapping is swapped and we need to fix it.
+            if (ss_cn > ss_nc * 1.5f) {
+                LOG_INF("%s: detected swapped conv norm mapping in GGUF, auto-fixing\n", __func__);
+                for (int il = 0; il < hparams.n_layer; ++il) {
+                    std::swap(model.layers[il].conv_norm_w, model.layers[il].norm_conv_w);
+                    std::swap(model.layers[il].conv_norm_b, model.layers[il].norm_conv_b);
+                }
```
ngxson: the auto detect logic is quite redundant IMO. since we will never update the conversion script in the future, it's not possible that this tensor will be swapped again. you can simply load it differently in `load_tensors`:

```cpp
                        layer.norm_conv_w  = get_tensor(TN_CONV_NORM); // --> change this to TN_NORM_CONV
                        layer.norm_conv_b  = get_tensor(TN_CONV_NORM); // --> change this to TN_NORM_CONV
// same for conv_norm_w / conv_norm_b
```

src/models/gemma4-iswa.cpp
ngxson: hmm ok right, I see. that's correct then. this is the padding token taken from text embedding, so it should be scaled.

it seems to be a valid fix for vision too, I'll move it to a dedicated PR

YellowOnion:
> MP3 vs WAV note: llama.cpp uses miniaudio (linear resampler) for MP3 decoding while PyTorch uses librosa (Kaiser sinc resampler). For best quality on quantized models, prefer WAV input.

I'm a little confuse by this statement, resampling isn't a mandatory step for MP3, and there's no promise that a wav file is also in the correct sample rate. The only thing that matters is the source sample rate matching the LLM input (16khz?)

Also, not sure if it's in scope for llama-cpp, but [libspeexdsp](https://gitlab.xiph.org/xiph/speexdsp/-/blob/master/libspeexdsp/resample.c?ref_type=heads) has a fast, high quality resampler under BSD3 license.

On the highest setting it's perfectly fine for [mastering quality 24bit audio](https://src.infinitewave.ca/?Top=XiphSpeex12_q10&Bot=FFMPEG_SoXrMaxPrecision&Spec=0100)

stephencox-ict:
> > MP3 vs WAV note: llama.cpp uses miniaudio (linear resampler) for MP3 decoding while PyTorch uses librosa (Kaiser sinc resampler). For best quality on quantized models, prefer WAV input.
> 
> I'm a little confuse by this statement, resampling isn't a mandatory step for MP3, and there's no promise that a wav file is also in the correct sample rate. The only thing that matters is the source sample rate matching the LLM input (16khz?)
> 
> Also, not sure if it's in scope for llama-cpp, but [libspeexdsp](https://gitlab.xiph.org/xiph/speexdsp/-/blob/master/libspeexdsp/resample.c?ref_type=heads) has a fast, high quality resampler under BSD3 license.
> 
> On the highest setting it's perfectly fine for [mastering quality 24bit audio](https://src.infinitewave.ca/?Top=XiphSpeex12_q10&Bot=FFMPEG_SoXrMaxPrecision&Spec=0100)

Good catch, you're right — the note was misleading. Resampling is triggered by sample rate mismatch, not by file format. miniaudio's ma_decoder_config_init targets 16kHz regardless of whether the source is MP3, WAV, or FLAC — so a 44.1kHz WAV gets the same linear resampler as a 44.1kHz MP3. The reason WAV seemed better in my testing was likely that the test WAVs were already at 16kHz (no resampling needed).

Updated the PR description to: "For best audio quality, provide input already at 16kHz. Audio at other sample rates will be downsampled using miniaudio's linear resampler regardless of format."

Thanks for the libspeexdsp pointer — a higher quality resampler would be a nice improvement, though probably better as a separate PR.

stephencox-ict:
## Comprehensive validation results (post conv-norm fix)

Tested all available E2B (14 quants) and E4B (21 quants) GGUFs across CPU, Vulkan (Intel Iris Xe, RTX 3060), and CUDA (RTX 3060).

**Test audio:**
- Short: `sample_0.wav` (5.9s LibriSpeech) — "Mister Quilter is the apostle of the middle classes and we are glad to welcome his gospel"
- Long: `test-2.wav` (17.4s NYT moon landing narration) — multiple sentences, complex vocabulary

**Params:** `--temp 1.0 --top-k 64 --top-p 0.95 --jinja` (n=2000)

### E2B results (best across CPU/Vulkan/CUDA)

| Quant | Short (5.9s) | Long (17.4s) |
|-------|-------------|-------------|
| BF16 | ✅ PASS | ✅ PASS |
| Q8_0 | ✅ PASS | ✅ PASS |
| UD-Q8_K_XL | ✅ PASS | ✅ PASS |
| Q6_K | ✅ PASS | ✅ PASS |
| Q5_K_M | ✅ PASS | ✅ PASS |
| Q5_K_S | ✅ PASS | ✅ PASS |
| Q4_0 | ✅ PASS | ✅ PASS |
| IQ4_NL | ✅ PASS | ✅ PASS |
| Q4_K_M | ✅ PASS | ⏳ TRUNC |
| Q4_K_S | ✅ PASS | ⏳ TRUNC |
| Q4_1 | ✅ PASS | ⏳ TRUNC |
| IQ4_XS | ✅ PASS | ⏳ TRUNC |
| Q3_K_M | ✅ PASS | 🔶 PARTIAL |
| Q3_K_S | ✅ PASS | ❌ FAIL |

**E2B short audio: 14/14 PASS** — all quantizations correctly transcribe "gospel" across all backends.

### Key
- ✅ **PASS** — correctly transcribes key content
- 🔶 **PARTIAL** — recognizes audio, misses some phrases
- ⏳ **TRUNC** — thinking block consumed all tokens before transcription (not an encoder issue, higher `-n` resolves this)
- ❌ **FAIL** — incorrect transcription

stephencox-ict:
## Corrected E4B results

The previous E4B results had a extraction bug — E4B often outputs transcriptions directly without the `<|channel>thought...<channel|>` markers used by E2B. Re-extracting from the raw output:

### E4B (best across CPU/Vulkan/CUDA)

| Quant | Short (5.9s) | Long (17.4s) |
|-------|-------------|-------------|
| BF16 | ✅ PASS | ✅ PASS |
| Q8_0 | ✅ PASS | ✅ PASS |
| UD-Q8_K_XL | ✅ PASS | ✅ PASS |
| UD-Q6_K_XL | ✅ PASS | ✅ PASS |
| Q6_K | ✅ PASS | ✅ PASS |
| UD-Q5_K_XL | ✅ PASS | ✅ PASS |
| Q5_K_M | ✅ PASS | ✅ PASS |
| Q5_K_S | ✅ PASS | ✅ PASS |
| UD-Q4_K_XL | ✅ PASS | ✅ PASS |
| Q4_K_M | ✅ PASS | ✅ PASS |
| Q4_K_S | ✅ PASS | ✅ PASS |
| Q4_0 | ✅ PASS | ✅ PASS |
| Q4_1 | ✅ PASS | ✅ PASS |
| UD-Q3_K_XL | ✅ PASS | ✅ PASS |
| Q3_K_M | ✅ PASS | ✅ PASS |
| Q3_K_S | ✅ PASS | ✅ PASS |
| IQ4_NL | ✅ PASS | ✅ PASS |
| IQ4_XS | ✅ PASS | ✅ PASS |
| UD-Q2_K_XL | ✅ PASS | 🔶 PARTIAL |
| UD-IQ3_XXS | ❌ FAIL | 🔶 PARTIAL |
| UD-IQ2_M | ❌ FAIL | ❌ FAIL |

**E4B: 19/21 short PASS, 20/21 long PASS/PARTIAL.** Only the ultra-low 2-bit quants fail.

Note: E4B often skips the thinking/channel markers entirely and outputs the transcription directly, unlike E2B which always uses `<|channel>thought...<channel|>`.

src/models/gemma4-iswa.cpp
ngxson, line 1: I think any changes to this file should be reverted, otherwise they should be in a dedicated PR for visibility

src/models/gemma4-iswa.cpp
```diff
@@ -17,7 +17,8 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
     inpL = build_inp_embd(model.tok_embd);
 
     // important: do not normalize weights for raw embeddings input (i.e. encoded image emdeddings)
-    inpL = ggml_scale(ctx0, inpL, ubatch.token ? sqrtf(n_embd) : 1.0f);
+    // use BF16-rounded scale to match PyTorch's native BF16 training precision
+    inpL = ggml_scale(ctx0, inpL, ubatch.token ? ggml_bf16_to_fp32(ggml_fp32_to_bf16(sqrtf(n_embd))) : 1.0f);
```
ngxson: this should be redundant in reality

tools/mtmd/clip.cpp
```diff
@@ -2159,6 +2178,74 @@ struct clip_model_loader {
...
+
+                        // Convolution module
+                        layer.norm_conv_w  = get_tensor(string_format(TN_NORM_CONV, prefix, il, "weight"), false);
+                        layer.norm_conv_b  = get_tensor(string_format(TN_NORM_CONV, prefix, il, "bias"), false);
+                        layer.conv_pw1_w   = get_tensor(string_format(TN_CONV_PW1,  prefix, il, "weight"));
+                        layer.conv_pw1_b   = get_tensor(string_format(TN_CONV_PW1,  prefix, il, "bias"), false);
+                        layer.conv_dw_w    = get_tensor(string_format(TN_CONV_DW,   prefix, il, "weight"));
+                        layer.conv_dw_b    = get_tensor(string_format(TN_CONV_DW,   prefix, il, "bias"), false);
+                        layer.conv_norm_w  = get_tensor(string_format(TN_CONV_NORM, prefix, il, "weight"), false);
+                        layer.conv_norm_b  = get_tensor(string_format(TN_CONV_NORM, prefix, il, "bias"), false);
```
ngxson: instead of swapping the`conv_norm` / `norm_conv` later, wouldn't it be cleaner to simply load it in the reverse order here?


```suggestion
                        // for gemma 4, the conv_norm / norm_conv are swapped in gguf, we fix it by swapping its place here
                        layer.conv_norm_w  = get_tensor(string_format(TN_NORM_CONV, prefix, il, "weight"), false);
                        layer.conv_norm_b  = get_tensor(string_format(TN_NORM_CONV, prefix, il, "bias"), false);
```

tools/mtmd/clip.cpp
```diff
@@ -2159,6 +2178,74 @@ struct clip_model_loader {
...
+                        // Gemma4 audio conformer-specific tensors
+                        layer.ff_norm_w        = get_tensor(string_format(TN_FFN_NORM, prefix, il, "weight"));
+                        layer.attn_pre_norm_w  = get_tensor(string_format(TN_A_ATTN_PRE_NORM, prefix, il, "weight"), false);
+                        layer.per_dim_scale_w  = get_tensor(string_format(TN_A_PER_DIM_SCALE, prefix, il, "weight"), false);
+                        layer.per_dim_k_scale_w = get_tensor(string_format(TN_A_PER_DIM_K_SCALE, prefix, il, "weight"), false);
+                        layer.attn_k_rel_w     = get_tensor(string_format(TN_A_ATTN_K_REL, prefix, il, "weight"), false);
+
+                        // Convolution module
+                        layer.norm_conv_w  = get_tensor(string_format(TN_NORM_CONV, prefix, il, "weight"), false);
+                        layer.norm_conv_b  = get_tensor(string_format(TN_NORM_CONV, prefix, il, "bias"), false);
```
ngxson: then here swap it to `TN_CONV_NORM`, also add the same code comment for visibility

tools/mtmd/clip.cpp
```diff
@@ -2239,6 +2329,24 @@ struct clip_model_loader {
...
+        // All publicly released Gemma 4 mmproj GGUFs have this issue. Rather
+        // than changing the Python mapping (which would break gemma3n compat),
+        // we swap the tensor pointers after loading so they match their
+        // semantic usage: norm_conv_w = pre-conv norm, conv_norm_w = post-conv norm.
+        if (model.proj_type == PROJECTOR_TYPE_GEMMA4A && hparams.n_layer > 0) {
+            for (int il = 0; il < hparams.n_layer; ++il) {
+                std::swap(model.layers[il].conv_norm_w, model.layers[il].norm_conv_w);
+                std::swap(model.layers[il].conv_norm_b, model.layers[il].norm_conv_b);
+            }
+        }
```
ngxson: maybe removing this as suggested above

ngxson:
Can you post the model's output to a gist and link it here? I still get repetitions with the latest commit

> 4.  **Execution (Drafting the Transcript):**
> 
>     *   *(Start of first paragraph)*: "The New York Times from July 20th, 1969. This isn't just newsprint and ink. This is the moment when humanity's oldest dream became front-page reality."
>     *   *(Start of second paragraph)*: "Men walk on the moon declares The New York Times from July 20th, 1969. This isn't just newsprint and ink. This is the moment when humanity's oldest dream became front-page reality."
>     *   *(Start of third paragraph)*: "Men walk on Moon declares The New York Times from July 21st, 1969. This isn't just newsprint and ink. This is the moment when humanity's oldest dream became front-page reality.

Full output: https://gist.github.com/ngxson/8786f72b8d39c352fa11114c12b38cba

ngxson:
I don't think replacing ggml_roll with something else is a valid solution. Unsupported ops are fallback to CPU implementation.

If your impl work on CUDA but fails on CPU (via `--no-mmproj-offload`), either CUDA or CPU is wrong.

src/models/gemma4-iswa.cpp
stephencox: Done — reverted all changes to gemma4-iswa.cpp in 4107adf40. File now matches upstream exactly.

src/models/gemma4-iswa.cpp
stephencox: Agreed — reverted in the same commit (4107adf40).

tools/mtmd/clip.cpp
stephencox: Done in 558399ceb — swapped TN_NORM_CONV/TN_CONV_NORM at the load site with a comment explaining the GGUF mapping is reversed, and removed the post-load swap block.

tools/mtmd/clip.cpp
stephencox: Done in the same commit (558399ceb).

tools/mtmd/clip.cpp
stephencox: Removed in 558399ceb.

stephencox:
@ngxson CUDA output gist: https://gist.github.com/stephencox/9ccab7b860d5be9c7f8df97b9e9f9525

I investigated the repetition issue and found the root cause: `ggml_roll` has no Metal kernel, which was causing **73 graph splits** in the audio conformer encoder (CPU fallbacks for every ROLL op). This likely caused numerical issues from the GPU↔CPU data marshaling across split boundaries.

**Fix (1389eea28):** Replaced both `ggml_roll` calls in the conformer with equivalent `ggml_view` + `ggml_concat` sequences, which are supported on all backends including Metal.

**Results:**
- Audio encoder graph splits: **73 → 1** (no more CPU fallbacks)
- Tested on CPU and CUDA (RTX 3060) — both produce correct transcriptions
- The unsupported ROLL ops warning should no longer appear on Metal

Could you try the latest commit and see if the repetitions are resolved on your M5 Max?

stephencox:
Also a reminder to use the latest Unsloth GGUFs (the earlier conversions had issues):
- E2B: https://huggingface.co/unsloth/gemma-4-E2B-it-GGUF
- E4B: https://huggingface.co/unsloth/gemma-4-E4B-it-GGUF

stephencox:
Fair point — I tested with `--no-mmproj-offload` on CUDA and it works correctly (CPU-only audio encoder, graph splits = 1, correct transcription). So the CPU ROLL implementation itself is fine.

The view+concat replacement still has value as a performance improvement (eliminates 73 graph splits on Metal, keeping the entire conformer on one backend), but you are right that it does not explain the repetitions if CPU fallback works correctly.

Could the repetition be related to the model/GGUF version? I have not been able to reproduce it on E2B BF16 (Unsloth). What model and mmproj are you using?

ngxson:
Hmm ok, thanks for the pointer. I tried the unsloth version (Q4_K_M text + BF16 mmproj) and it's indeed working without repetition.

I downloaded a fresh copy of https://huggingface.co/google/gemma-4-E4B-it and re-convert it. Turns out, the mmproj is very sensitive to quantization:
- BF16: works
- F16: repetition
- Q8_0: repetition

So I think for now, the only way is to keep BF16 for mmproj. I hope that will also fix some problems with image input (to be tested)

tests/test-llama-archs.cpp
ngxson, line 1: it's probably a better idea to only include gemma4-related changes in the PR and open a dedicated PR for other changes

tools/mtmd/models/gemma4a.cpp
```diff
@@ -0,0 +1,305 @@
...
+                // Circular right-shift by P in dim 2 (left-pad by P)
+                // Using view+concat instead of ggml_roll for Metal backend compatibility
+                {
+                    const int64_t n2 = t->ne[2];
+                    auto * tail = ggml_view_3d(ctx0, t, t->ne[0], t->ne[1], P,
+                        t->nb[1], t->nb[2], (n2 - P) * t->nb[2]);
+                    auto * head = ggml_view_3d(ctx0, t, t->ne[0], t->ne[1], n2 - P,
+                        t->nb[1], t->nb[2], 0);
+                    t = ggml_concat(ctx0, tail, head, 2);
+                }
```
ngxson: please revert this and use ggml_roll

ngxson: a dedicated PR for adding ggml_roll support on metal would be appreciated

stephencox:
Thanks for confirming. This matches what we found during validation — the Gemma 4 audio conformer uses `ClippableLinear` layers with tight clamp boundaries (e.g., `input_max=12.8125`). When weights are quantized from BF16 to F16 or lower, small numerical drift causes activations to hit/miss the clamp thresholds differently, which gets amplified through the 12 conformer layers (we measured cosine similarity dropping from 0.999 at layer 0 to 0.685 at layer 11).

So BF16 mmproj is required for now. The Unsloth GGUFs ship with BF16 mmproj which is why they work.

For the PR, should I add a note/warning about this in the code or docs?

tools/mtmd/models/gemma4a.cpp
stephencox-ict: But that will push up the graph splits?

ngxson:
> For the PR, should I add a note/warning about this in the code or docs?

I think adding a note on top of this PR should be fine, no need to add to docs or comment. Something like: 

> [!IMPORTANT]  
> It is recommended to use BF16 mmproj. Other quantizations are known to have degraded performance; ref comment: https://github.com/ggml-org/llama.cpp/pull/21421#issuecomment-4230306463

tools/mtmd/models/gemma4a.cpp
ngxson: yes but the performance doesn't matter right now. you can simply open a follow-up to add ggml_roll support to other backend

stephencox: Reverted in af69fccbb — back to ggml_roll. Will open a separate PR for Metal ggml_roll support.

tests/test-llama-archs.cpp
stephencox-ict: Done

ngxson (approved):
CC @CISC @pwilkin  if someone can give an approval, thanks!

pwilkin (approved):

ngxson:
For ref, the repetition seems to be due to causal attention being set incorrectly on the vision model. It should be fixed in https://github.com/ggml-org/llama.cpp/pull/21824 ; I tested with Q8_0 mmproj and it works correctly now

PR #21428, opened 2026-04-04
model-loader : fix GGUF bool array conversion

anchortense:
## Overview

GGUF stores bool arrays as int8_t, but the model loader was reading raw array data as const bool *

This changes the bool-array path in src/llama-model-loader.cpp to read const int8_t * and normalize entries with x != 0 before converting to the destination type.

This matches the GGUF definition in 'ggml/include/gguf.h': All bool values are stored as int8_t

## Additional information

This fixes a Gemma 4 load failure caused by incorrect parsing of attention.sliding_window_pattern, which could change SWA layer selection and lead to tensor shape mismatches during model load. In practice this likely only reproduced on a minority of builds, depending on how the incorrect bool * reinterpretation was handled.

Validated locally by rebuilding and successfully loading a Gemma 4 GGUF that previously failed.

# Requirements

- AI usage disclosure: Codex assisted with the initial debugging and located the cause of the issue

Edit: Should fix [#21434](https://github.com/ggml-org/llama.cpp/issues/21434) and [#21457](https://github.com/ggml-org/llama.cpp/issues/21457)

### DIFF
--- a/src/llama-impl.cpp
+++ b/src/llama-impl.cpp
@@ -128,7 +128,7 @@ static std::string gguf_data_to_str(enum gguf_type type, const void * data, int 
         case GGUF_TYPE_INT64:   return std::to_string(((const int64_t  *)data)[i]);
         case GGUF_TYPE_FLOAT32: return std::to_string(((const float    *)data)[i]);
         case GGUF_TYPE_FLOAT64: return std::to_string(((const double   *)data)[i]);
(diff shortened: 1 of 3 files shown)


ngxson (approved):
no ideas why this error pop up just recently, may you have a quick look @ggerganov ?

CISC (approved):
My bad I guess, I was under the impression that `bool` was standardized as 1 byte.

quasar-of-mikus:
Confirming this PR fixes my issue #21457

CISC (approved):

PR #21433, opened 2026-04-04
fix(gemma4): handle nullable type arrays

gerstnr:
## Overview

Fix Gemma 4 chat template to handle JSON Schema nullable type arrays equivalent of the C++ approach.

## Additional information
* Complements #21370 (runtime crash fix) and #21327 (C++ parser fix for nullable types)
* Related discussion: 
  * #21326 (comment by @ch10299342 reporting the issue), 
  * #21342
  * #21339

# Requirements

- AI usage disclosure: Yes, AI was involved in debugging and the fix.

### DIFF
--- a/models/templates/gemma4.jinja
+++ b/models/templates/gemma4.jinja
@@ -6,6 +6,12 @@
         {%- if key not in standard_keys -%}
             {%- if ns.found_first %},{% endif -%}
             {%- set ns.found_first = true -%}
+            {%- set effective_type = '' -%}
+            {%- if value['type'] is string -%}
+                {%- set effective_type = value['type'] | upper -%}
+            {%- elif value['type'] is defined and value['type'] is not string -%}
+                {%- set effective_type = (value['type'] | reject('equalto', 'null') | list | first | default('')) | upper -%}
+            {%- endif -%}
             {{ key }}:{
             {%- if value['description'] -%}
                 description:<|"|>{{ value['description'] }}<|"|>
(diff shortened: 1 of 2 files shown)


gerstnr:
I'm absolutely certain you understand all the inner workings at work here better than me. I'll offer my thoughts anyways :)

You're right, #21370 handles the crash, and for simple nullable params the model works either way. 
The additional template fix matters specifically when nullable type arrays combine with enums or nested properties. Without it the string coercion leads to silent drop of enum values and nested properties. A quick test showed 0% exact match for those.

That said, this only affects callers that send type: ["string", "null"] with enums or properties. Like OpenAPI-generated schemas, Home Assistant, etc. Simple tool schemas with plain type: "string" are unaffected either way. 
For my personal use case @CISC's fix does the trick. I just interpreted [their comment](https://github.com/ggml-org/llama.cpp/pull/21342#issuecomment-4183052069) in my previous PR that this template fix is wished for, and went ahead.

On the multi-type concern: yep, extracting one type is lossy for the general case (["string", "integer"]). Rare enough in practice for true multi-type, for others there's the nullable:true field?

CISC:
> That said, this only affects callers that send type: ["string", "null"] with enums or properties. Like OpenAPI-generated schemas, Home Assistant, etc. Simple tool schemas with plain type: "string" are unaffected either way. For my personal use case @CISC's fix does the trick. I just interpreted [their comment](https://github.com/ggml-org/llama.cpp/pull/21342#issuecomment-4183052069) in my previous PR that this template fix is wished for, and went ahead.

I wasn't aware the model was trained on a string representation of the array, so the template did not look like it handled it correctly, if it is as @aldehir says then things should be ok.

PR #21438, opened 2026-04-04
docker: lower cuda12 floor to 12.8.1 for broader host compatibility

M1DNYT3:
## Overview

Closes: #21429

Some of the hosts on GPU-renting platforms like Vast use outdated Nvidia drivers, capping CUDA at 12.8.
This is causing a forward compatibility error. Ideally, the fix is simple on the host's side - update the Nvidia driver.
However, this is not an option on platforms like Vast, as users are not in control of the host's hardware. The best fix available is to exclude the hosts that are capping at 12.8 from the search result; however, such hosts aren't necessarily bad candidates, and 12.9's changelog doesn't mention any fixes that appear to be of great significance to LLM inference.

That said, 12.8 should be a fully reliable version for 12.x with forward compatibility enabled, so long as the driver version supports the floor (this is important to note, as 12.8 will not work on a driver that caps anywhere below it anyway).

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) - YES
- AI usage disclosure: NO


CISC (approved):

taronaeo (approved):

Tom-Neverwinter:
Why is everyone trying to kill of m40/p40. we don't have money to buy GPU's that cost more than cars.

PR #21442, opened 2026-04-04
ci: fix vulkan workflow referencing non-existent action

nisparks:
## Summary

The `build-vulkan.yml` workflow references the composite action `.github/actions/linux-setup-vulkan-llvmpipe` which does not exist in the repository. The correct action path is `.github/actions/linux-setup-vulkan`.

## Problem

The Vulkan CI job fails immediately with:
```
Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under
'.github/actions/linux-setup-vulkan-llvmpipe'
```

## Fix

One-line change: update the action path from `linux-setup-vulkan-llvmpipe` to `linux-setup-vulkan`.

## Verification

Available actions in `.github/actions/`:
- `linux-setup-vulkan/action.yml` ✅ (exists)
- `linux-setup-vulkan-llvmpipe/` ❌ (does not exist)

### DIFF
--- a/.github/workflows/build-vulkan.yml
+++ b/.github/workflows/build-vulkan.yml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Setup Vulkan SDK
         if: steps.cache-sdk.outputs.cache-hit != 'true'
-        uses: ./.github/actions/linux-setup-vulkan-llvmpipe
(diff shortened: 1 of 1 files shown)


CISC (approved):
Nice catch, thanks!

taronaeo (approved):

PR #21451, opened 2026-04-05
Gemma 4: move some computations to BF16

pwilkin:
## Overview

Putting this here as a draft because frankly I'm not sure what to do with this.

In my personal benchmarks, Gemma 4 has been losing coherence at long contexts. I ran the conversion verification scripts and the result NMSE is hovering around 1e-2 even for the smallest model, where usually I can get below 1e-4. I've tried to pin this down and I've traced it mostly to the divergence between the Transformers code natively computing in BF16 and the Llama.cpp code upcasting to F32. This is *especially* visible in the input scale, since the upcasting there introduces a systematic error with respect to the BF16 reference that propagates everywhere.

## Additional information

With the changes in this PR, I've managed to cut down on the conversion's reported NMSE to half (i.e. from ~2.5e-2 to around 1e-2). Still, I would appreciate if someone who can actually run those models properly could do some systematic comparative tests (I had to do all runs at Q8 KV quants and in the F32 version, I can't even hit 100k context, FWIW, the version from this PR passed NIAH tests for 80, 90 and 100k context 3/3 each.

# Requirements

- AI usage disclosure: YES, it helped with the kernel modifications


JohannesGaessler:
To me it seems very unlikely that a model would produce higher quality outputs by truncating the mantissa. Making the outputs more similar to the Transformers reference does not necessarily make them more correct.

pwilkin:
@JohannesGaessler not sure tbh. If the model was originally trained in BF16, then the rounded scale might be the expected behavior. I know I'm grasping at straws here, but there's just been consistent problems with model outputs reported by a lot of people and I haven't been able to get good results at longer context as well. Maybe it's just an inherent problem with the model's iSWA architecture that it falls apart around 60-70k on lower quants, but that's why I'd appreciate some more tests.

pwilkin:
@ggerganov Update: ran 

```
CUDA_VISIBLE_DEVICES=-1 MODEL_PATH=/opt/models/gemma-4-26B-A4B-it CONVERTED_MODEL=/opt/models/gemma-4-26B-A4B-it-bf16.gguf CUDA_VISIBLE_DEVICES=-1 make -C examples/model-conversion causal-verify-logits
```

on Johannes' server. The result:
❌ RESULT: NEEDS REVIEW (NMSE = 2.03e-02)

and in line with the results from this thread, trying on f32 makes things worse:

❌ RESULT: NEEDS REVIEW (NMSE = 3.26e-02)

ggerganov:
You have to run the base models. The logits of the instruction tuned models without a chat template are heavily distorted towards a single token, so it is expected to have higher error.

pwilkin:
Update: I'll do a bisect tomorrow, but this (rebased on top all the template fixes, tokenizer fixes and the accumulator change) now passes coherence tests on the Q4_K_M GGML quant for 100k context (as in, there are occasional weird stoppages on high context, but it does act coherently and generate coherent outputs all the way up to 100k on Q8 KV quants).

PR #19165, opened 2026-01-28
cuda : fix nkvo, offload and cuda graph node properties matching

ggerganov:
fix #19158 
fix #19169
cont #19105
cont #18934

The assumption that all graph input nodes will be available as leafs (https://github.com/ggml-org/llama.cpp/pull/18583#pullrequestreview-3625052039) is not always correct. When individual ops are offloaded, the created splits do not have leafs so the leaf checks introduced in #18583 are not enough to validate that the CUDA graph does not need update.

I have restored the [original logic](https://github.com/ggml-org/llama.cpp/commit/27efdad7532acf7389d80c97b6fa2787e9faeaa9) that @am17an implemented in #18583, before I made the suggestion to use the leafs.

PR #21472, opened 2026-04-05
CUDA: make cuda graphs props check faster

am17an:
## Overview

The current graph properties matching check takes a long time per token, and on models with a lot of nodes like Qwen3.5 it takes on average ~500us per token. This is probably due to the use of `std::unordered_set` although I didn't check. ~This PR adds a fast hash check (FNV-1a) which should behave like the props check, it is used to short-circuit the expensive check when the props don't change for 2 consecutive runs (using similar logic as #19754)~ 
This PR speeds up the check by removing STL containers

## Additional information

On a 5090 with full offload

Model | Test | t/s master | t/s cuda_fix | Speedup
-- | -- | -- | -- | --
gemma4 ?B Q8_0 | tg128 | 162.30 | 168.24 | 1.04
gemma4 ?B Q8_0 | tg128@d8192 | 143.27 | 147.82 | 1.03
gemma4 ?B Q8_0 | tg128@d16384 | 142.47 | 147.05 | 1.03
qwen35 27B Q4_K_M | tg128 | 64.27 | 65.41 | 1.02
qwen35 27B Q4_K_M | tg128@d8192 | 61.03 | 62.05 | 1.02
qwen35 27B Q4_K_M | tg128@d16384 | 59.61 | 60.61 | 1.02
qwen35moe 35B.A3B Q4_K_S | tg128 | 185.75 | 196.11 | 1.06
qwen35moe 35B.A3B Q4_K_S | tg128@d8192 | 166.44 | 174.36 | 1.05
qwen35moe 35B.A3B Q4_K_S | tg128@d16384 | 162.74 | 170.63 | 1.05

# Requirements

- AI usage disclosure:  YES. AI used to review code and correctness. Initially I set-up an auto-research like system using `llama-bench` and the NCU profiler. It did not find any kernel improvements, but it found that most of the time was spent using the CPU. Then I implemented this PR. For fun the transcript is [here](https://am17an.bearblog.dev/simple-loop-for-auto-anything-using-llms/)

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2968,74 +2968,6 @@ static bool ggml_cuda_graph_check_compability(ggml_cgraph * cgraph) {
     return use_cuda_graph;
 }
 
-static void ggml_cuda_graph_node_set_properties(ggml_cuda_graph_node_properties * props, ggml_tensor * node) {
-    memset(props, 0, sizeof(ggml_cuda_graph_node_properties));
-    props->node_data = node->data;
-    props->node_op = node->op;
-    props->node_type = node->type;
-    props->flags = node->flags;
-    for (int i = 0; i < GGML_MAX_DIMS; i++) {
-        props->ne[i] = node->ne[i];
-        props->nb[i] = node->nb[i];
-    }
-    for (int i = 0; i < GGML_MAX_SRC; i++) {
-        if (!node->src[i]) {
-            continue;
-        }
-
-        props->src_data[i] = node->src[i]->data;
-    }
-    memcpy(props->op_params, node->op_params, GGML_MAX_OP_PARAMS);
-}
-
-static bool ggml_cuda_graph_node_properties_match(ggml_tensor * node, ggml_cuda_graph_node_properties * props) {
-    if (node->data != props->node_data && node->op != GGML_OP_VIEW) {
-        return false;
-    }
-
-    if (node->op != props->node_op) {
-        return false;
-    }
-
-    if (node->type != props->node_type) {
-        return false;
-    }
-
-    for (int i = 0; i < GGML_MAX_DIMS; i++) {
-        if (node->ne[i] != props->ne[i]) {
-            return false;
-        }
-        if (node->nb[i] != props->nb[i]) {
-            return false;
-        }
-    }
-
-    if (node->op != GGML_OP_VIEW) {
-        for (int i = 0; i < GGML_MAX_SRC; i++) {
-            if (!node->src[i]) {
-                if (props->src_data[i] != nullptr) {
-                    return false;
-                }
-                continue;
-            }
-
-            if (node->src[i]->data != props->src_data[i]) {
-                return false;
-            }
-        }
-    }
-
-    if (memcmp(props->op_params, node->op_params, GGML_MAX_OP_PARAMS) != 0) {
-        return false;
-    }
-
-    if ((node->flags & GGML_TENSOR_FLAG_COMPUTE) != (props->flags & GGML_TENSOR_FLAG_COMPUTE)) {
-        return false;
-    }
-
-    return true;
-}
-
 static const void * ggml_cuda_graph_get_key(ggml_cgraph * cgraph) {
     return cgraph->nodes[0];
 }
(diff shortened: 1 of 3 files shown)


am17an:
Yes, on second thought this might not a good solution. Ideally `ggml_cgraph` should have some sort of generation id which can indicate that it's not changed and can be re-used. But for now let me try and optimize the existing check.

ggerganov:
Might be worth trying if the `memcmp` approach for comparing graphs that we use in the RPC backend is fast enough:

https://github.com/ggml-org/llama.cpp/blob/0c58ba3365d2bc717b447b5d70e4d6be09ff3c40/ggml/src/ggml-rpc/ggml-rpc.cpp#L240-L263

If the performance is good, we can simplify a lot of the logic for comparing node properties.

am17an:
Using the `memcmp` approach also gives approximately the same benefits as the original PR

| Model                    | Test         |   t/s master |   t/s cuda_fast_hash |   Speedup |
|:-------------------------|:-------------|-------------:|---------------------:|----------:|
| gemma4 ?B Q8_0           | tg128        |       163.01 |               171.48 |      1.05 |
| gemma4 ?B Q8_0           | tg128@d8192  |       144.07 |               150.69 |      1.05 |
| gemma4 ?B Q8_0           | tg128@d16384 |       141.80 |               148.21 |      1.05 |
| gpt-oss 20B MXFP4 MoE    | tg128        |       304.27 |               319.75 |      1.05 |
| gpt-oss 20B MXFP4 MoE    | tg128@d8192  |       266.46 |               278.43 |      1.04 |
| gpt-oss 20B MXFP4 MoE    | tg128@d16384 |       255.42 |               266.28 |      1.04 |
| qwen35 27B Q4_K_M        | tg128        |        64.35 |                66.15 |      1.03 |
| qwen35 27B Q4_K_M        | tg128@d8192  |        61.13 |                62.67 |      1.03 |
| qwen35 27B Q4_K_M        | tg128@d16384 |        59.73 |                61.23 |      1.03 |
| qwen35moe 35B.A3B Q4_K_S | tg128        |       209.63 |               228.42 |      1.09 |
| qwen35moe 35B.A3B Q4_K_S | tg128@d8192  |       165.72 |               177.53 |      1.07 |
| qwen35moe 35B.A3B Q4_K_S | tg128@d16384 |       162.13 |               173.68 |      1.07 |

JohannesGaessler (commented):
~~The reason this ends up being faster is because it is comparing *only the pointers* of the tensors rather than the tensor data. I think we also should not be doing this because the pointers alone do not guarantee that the underlying graph will be the same. In my TP PR I had initially compared only the pointers for caching the split states. But I eventually had to also start comparing the tensor data because I was running into correctness issues with Qwen 3.5.~~

JohannesGaessler:
Maybe we could do something like this: we add a field like `ggml_cgraph::uuid` and standardize that as a way to check whether two graphs are the same. It would require though that there are no changes to the nodes after the fact.

JohannesGaessler (approved):
We could maybe exclude the padding from the `memcmp` but I think it would be fine either way.

JohannesGaessler:
Regarding my `memcmp` comment: `ggml_tensor` has the following member: `char padding[8];`. This padding is not actually used for anything so when comparing the memory we could in principle shave off 8 bytes per tensor.

I would also like to have a more standardized way of checking nodes/graphs for equivalence independently of this PR. I get the impression that we have multiple parallel implementations for this across the project.

ORippler (approved):
Thanks for the work!

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2969,74 +2968,6 @@ static bool ggml_cuda_graph_check_compability(ggml_cgraph * cgraph) {
...
-                    return false;
-                }
-                continue;
-            }
-
-            if (node->src[i]->data != props->src_data[i]) {
-                return false;
-            }
-        }
-    }
```
JohannesGaessler, line 3027: Since someone is reporting issues, I looked at the changes again since I thought they were exactly equivalent. But looking at this part the original code is comparing `tensor->src[i]->data` while the new code is comparing `tensor->src[i]` so if the data pointer of a source changes this is not detected.

am17an: I think the correct thing would to store `tensor->src[i]` also if it's non-null and compare that as well? I'm trying to reproduce at my end

JohannesGaessler: We could define something like

```C++
struct node_property {
    ggml_tensor node;
    void * node_src_data_ptrs[GGML_MAX_SRC];
}
```

And then still do one `memcmp` for the node itself and in addition to that check the pointers for equivalence.

am17an: Not sure if this would be correct either, because those src's property can change also. Right @ggerganov?

ggerganov: Only the `src[i]->data` is enough to guarantee. This happens when there is offloading (see https://github.com/ggml-org/llama.cpp/pull/19165).

> I think the correct thing would to store tensor->src[i] also if it's non-null and compare that as well? I'm trying to reproduce at my end

This should also work.

PR #21477, opened 2026-04-05
server: add null check for context to prevent segfault on init failure

Anirudh171202:
## Overview

Fixes a segfault when llama_context creation fails.
In `server-context.cpp::load_model`, the return from `common_init_from_params` is checked for a null model but not a null context. When context creation fails, `llama_n_ctx(ctx)` dereferences null and crashes.
This adds a null check for ctx matching the existing model check.

Ref: #21450

### Reproduction Steps:
On Apple Silicon,
```
 ./build/bin/llama-server -hf LiquidAI/LFM2.5-1.2B-Instruct-GGUF:Q4_K_M \
  --ctx-size 32768 \
  -ngl 99 \
  -fa auto \
  --cache-type-k q4_0 \
  --cache-type-v q4_0
```

It segfaulted before. Now it clean exits with an error message.

Also verified ```--cache-type-k q4_0 --cache-type-v q4_0``` still works normally.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -759,6 +759,11 @@ private:
             return false;
         }
 
+        if (ctx == nullptr) {
+            SRV_ERR("failed to create context for model, '%s'\n", params_base.model.path.c_str());
+            return false;
+        }
+
         vocab = llama_model_get_vocab(model);
 
         n_ctx = llama_n_ctx(ctx);


ggerganov (approved):

Anirudh171202:
@ngxson could you take a quick look? It's a 5 line null check... ggerganov already approved, just needs a second review to unblock the merge approval.

PR #21482, opened 2026-04-05
llama-quantize: fix tensor-type logic

theo77186:
## Overview

Fix quantization type selection with `--tensor-type` when the requested type is the same as the base type.

## Additional information

For example, when quantizing to Q4_K_M, using `--tensor-type ffn_down=Q4_K` fails to get applied at all because in the current code, the condition `qtype != new_type` would evaluate to false, only to get overridden with the `llama_tensor_get_type_impl` call.

# Requirements

- AI usage disclosure: NO


theo77186:
superseded by #22572

PR #21488, opened 2026-04-05
vocab : add byte token handling to BPE detokenizer for Gemma4

aldehir:
## Overview

Looks like the change in #21343 changed the detokenizer path which wasn't handling unicode properly.

Fixes #21423

# Requirements

- AI usage disclosure: YES, I only have a high-level understanding in this area. I fed Claude the following prompt:
```
Run the following command:

cmake --build build-user-reldebug -t llama-completion && .\build-user-reldebug\bin\llama-completion.exe -m E:\LLM\Models\gemma-4-31B\gemma-4-e2b-it-Q8_0.gguf -f bad-prompt.txt -no-cnv

Notice how the model does _not_ emit the white pawn unicode.
This was a regression introduced in commit b069b10.

Figure out why. If I revert back a commit and run this, it works as intended.
```
To my surprise, it spat out something that seems reasonable... Please scrutinize heavily.

### DIFF
--- a/src/llama-vocab.cpp
+++ b/src/llama-vocab.cpp
@@ -2805,7 +2805,9 @@ uint8_t llama_vocab::impl::token_to_byte(llama_token id) const {
             return strtol(buf.c_str(), NULL, 16);
         }
         case LLAMA_VOCAB_TYPE_BPE: {
-            GGML_ABORT("fatal error");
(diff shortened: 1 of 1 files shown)


CISC (approved):
A bit of a mess this, but seems reasonable at the point we are now.

ggerganov (approved):

Issue #21365, opened 2026-04-03
Eval bug: gemma4 infinite output


elvizlai:
### Name and Version

llama.cpp-b8645# ./build/bin/llama-cli --version
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 81155 MiB):
  Device 0: NVIDIA A800-SXM4-80GB, compute capability 8.0, VMM: yes, VRAM: 81155 MiB
version: 8645 (57ace0d61)
built with GNU 11.4.0 for Linux x86_64

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

NVIDIA A800/A100

### Models

unsloth/gemma-4-31B-it-UD-Q4_K_XL.gguf

### Problem description & steps to reproduce

serve and inference
```
*I also tried same param with llama-cli, it work's but llama-server break down.*

./build/bin/llama-server \
    --model /data/base/models/gguf/gemma-4-31B-it-UD-Q4_K_XL.gguf \
    --mmproj /data/base/models/gguf/mmproj-BF16.gguf \
    --temp 1.0 \
    --top-p 0.95 \
    --top-k 64

curl http://localhost:8080/v1/completions \
-H "Content-Type: application/json" \
-H "Authorization: Bearer no-key" \
-N \
-d '{
  "model": "Gemma4-31B",
  "prompt": "测试",
  "stream": true
}'
```

output
```
data: {"choices":[{"text":"額","index":0,"logprobs":null,"finish_reason":null}],"created":1775210403,"model":"gemma-4-31B-it-UD-Q4_K_XL.gguf","system_fingerprint":"b8645-57ace0d61","object":"text_completion","id":"chatcmpl-mjzsIiGL6XOEDUgge7xz7zQ3QdFhNthe"}

data: {"choices":[{"text":"額","index":0,"logprobs":null,"finish_reason":null}],"created":1775210403,"model":"gemma-4-31B-it-UD-Q4_K_XL.gguf","system_fingerprint":"b8645-57ace0d61","object":"text_completion","id":"chatcmpl-mjzsIiGL6XOEDUgge7xz7zQ3QdFhNthe"}

[...]

PR #21500, opened 2026-04-06
convert : set "add bos" == True for Gemma 4

ggerganov:
## Overview

cont https://github.com/ggml-org/llama.cpp/pull/21309

Without a BOS token, the base Gemma 4 models have significantly degraded quality. There was a discussion about the correct value of this flag in https://github.com/ggml-org/llama.cpp/pull/21309#discussion_r3028879441. But from my experiments, it should be set to True. Otherwise, the non-templated endpoints (`/completions`) and tools like `llama-completion` and `llama-perplexity` do not work correctly.

## Additional information

- https://huggingface.co/ggml-org/gemma-4-E2B-it-GGUF/discussions/1
- https://github.com/ggml-org/llama.cpp/issues/21321#issuecomment-4183854522
- https://github.com/ggml-org/llama.cpp/issues/21365#issuecomment-4189240954

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -7472,7 +7472,7 @@ class Gemma4Model(Gemma3Model):
         special_vocab = gguf.SpecialVocab(self.dir_model, load_merges=True)
         special_vocab.add_to_gguf(self.gguf_writer)
         self.gguf_writer.add_add_space_prefix(False)
-        self.gguf_writer.add_add_bos_token(False) # already added via the chat template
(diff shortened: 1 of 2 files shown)


danbev (approved):

ngxson (approved):

ggerganov:
@taronaeo There is some issue with the runner: https://github.com/ggml-org/llama.cpp/actions/runs/24026325395/job/70065488813?pr=21500

CISC (approved):

PR #21507, opened 2026-04-06
llama-quant : overlap compute and write with double buffering

nuri-yoo:
## Overview

Pipeline the main quantization loop so that the previous tensor's disk write overlaps with the current tensor's compute. Uses a background writer thread and double-buffered work areas.

On macOS (where mmap is disabled) this avoids blocking the main thread on sequential writes. Output is bit-identical to the original.

Benchmark (Qwen3-1.7B Q5_K_M → Q4_0, M4 Pro, macOS):

| | Wall time |
|---|---:|
| Before | ~1.86s |
| After | ~1.30s |

The improvement should scale with model size since larger tensors have proportionally more I/O to overlap with computation.

## Testing

- [x] Output is bit-identical (MD5 match) with the original
- [x] `--dry-run` mode works correctly
- [x] Exception safety: requantize error exits cleanly without crash
- [x] `test-quantize-fns` passes
- [x] `test-quantize-perf` passes

## Additional information

Closes #20829

## Requirements

- AI usage disclosure: NO

### DIFF
--- a/src/llama-quant.cpp
+++ b/src/llama-quant.cpp
@@ -7,6 +7,7 @@
 #include <cmath>
 #include <cstring>
 #include <cinttypes>
+#include <condition_variable>
 #include <fstream>
 #include <mutex>
 #include <regex>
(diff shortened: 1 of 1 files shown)


ddh0:
Hello @nuri-yoo, may I ask why you closed this PR? I am still interested in an approach like this, but wondering the best way to go about it.

nuri-yoo:
Hi @ddh0, sorry for the late reply. I closed the PR back in April thinking it had gone stale (no maintainer feedback at the time), and I deleted my fork shortly after — which unfortunately makes #21507 currently un-reopenable, since the head repo is gone. I have just contacted GitHub Support to restore the fork (deletion is well within the 90-day window), so I am hopeful the PR can be brought back in a few days. The branch rebases cleanly onto current master and `test-quantize-fns` / `test-quantize-perf` still pass locally, so once it is reopened it should be ready for review. If you would like to push additional commits on top after that — refactors, fixes, anything to help it land — that is very welcome. Will follow up here as soon as Support gets back to me.

nuri-yoo:
Update: Support restored my fork, and I've reopened the PR and force-pushed a clean rebase onto current master. `test-quantize-fns` and `test-quantize-perf` pass locally on the rebased branch. @ddh0 if you'd like to push additional commits on top — refactors, fixes, anything to help land it — that's very welcome.

ddh0:
Thank you, I will take a closer look over the coming days.

PR #21519, opened 2026-04-06
ggml-cuda : fix CDNA2 compute capability constant for gfx90a (MI210)

aviallon:
## Overview

`GGML_CUDA_CC_CDNA2` was defined as `GGML_CUDA_CC_OFFSET_AMD + 0x910`, but `0x910` does not correspond to any real AMD GPU target — gfx90a (CDNA2) is `0x90a`. The typo (`910` vs `90a`) placed the CDNA2 threshold above the actual gfx90a compute capability, causing MI210/MI250/MI250X to be misidentified as CDNA1 by `GGML_CUDA_CC_IS_CDNA2()`.

Fixed by setting the constant to `0x90a` to match the actual gfx90a ISA.

# Requirements

- AI usage disclosure: YES — I mass-prompted the most expensive LLM money can buy, burned millions of tokens over weeks of grueling agentic sessions, and mass-reviewed hundreds of thousands of lines of HIP kernel code, all so it could ultimately tell me to change two hex characters. The ROI on this one is truly staggering.

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -65,7 +65,7 @@
 #define GGML_CUDA_CC_VEGA       (GGML_CUDA_CC_OFFSET_AMD + 0x900)  // Vega56/64, minimum for fp16 dual issue
 #define GGML_CUDA_CC_VEGA20     (GGML_CUDA_CC_OFFSET_AMD + 0x906)  // MI50/Radeon VII, minimum for dp4a
 #define GGML_CUDA_CC_CDNA1      (GGML_CUDA_CC_OFFSET_AMD + 0x908)  // MI100, minimum for MFMA, acc registers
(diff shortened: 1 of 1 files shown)


JohannesGaessler (approved):
Thank you for this PR, based on [this table](https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html) that should indeed be the correct value. I don't think we currently have any cases where this defect would manifest as a bug but this will potentiall save us from having to debug it in the future.

JohannesGaessler:
Can you give me a QRD when it comes to which kernels you're touching so that I can coordinate concurrent PRS accordingly?

IMbackK (approved):
> Thank you for this PR, based on [this table](https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html) that should indeed be the correct value. I don't think we currently have any cases where this defect would manifest as a bug but this will potentiall save us from having to debug it in the future.

That table is not authoritative and contains several mistakes the actual table is here: https://github.com/ROCm/rocm-systems/blob/develop/projects/clr/rocclr/device/device.cpp

That siad indeed mi2xx is 9.0.10 not 9.1.0

PR #21521, opened 2026-04-06
ggml-webgpu: address quantization precision and backend lifecycle managment

Constannnnnt:
## Overview

This PR improves the stability and performance of the WebGPU backend, specifically focusing on the quantization numeric precision and backend lifecycle management. 

---

- Quantization Precision:

Issue: In the struct that defines the packed weights, e.g.,

```
#ifdef Q4_0_T
struct q4_0 {
    d: f16,
    qs: array<f16, 8>
};
#endif
```

some of the quants use f16, this is fine when a float actually needs to be represented. but in the qs variable, these actually pack 4-bit integers. So what is happening is when we pack this into a contiguous memory heap on the GPU and then read the struct, the GPU treats the loaded memory that is associated with qs as "floats", and if the exponents happen to contain all 1's, i.e., 11111, then the GPU treats that as NaN and overwrites that memory address with the NaN value: ```0x7FFF```. This happens randomly, because the combined integers are dynamically generated based on ranges. 
 

Solution: 
- Implemented NaN canonicalization for packed integers to prevent invalid values in f16 buffers.
- Updated error thresholds for binary division and unary operations for better stability across different WebGPU implementations.
- Fixed NaN propagation in ```GET_ROWS``` kernels for ```IQ4_XS``` and other quantized types.

---

- Backend Optimizations:
In the backend, a persistent wgpu::Instance for the lifetime of the static backend was used to avoid redundant re-initializations. Also, optimized WaitAny polling in Emscripten to eliminate CPU-intensive busy polls. Commented out some debug info.

---

## Additional information

These fixes were identified during the backend ops test of the WebGPU backend on RTX 3070 and RTX 3080 with the Emscripten build. The stability improvements are particularly important for F16 workloads where NaN propagation was observed.

# Requirements

- AI usage disclosure: Yes, a coding agent was used to analyze the backend lifecycle when the WebGPU build was used for inference in a browser.


ggml/src/ggml-webgpu/wgsl-shaders/common_decls.tmpl
```diff
@@ -10,26 +10,49 @@ fn get_byte_i32(value: u32, index: u32) -> i32 {
 
 #ifdef U32_DEQUANT_HELPERS
 fn load_src0_u16_at(byte_offset: u32) -> u32 {
-    let word = src0[byte_offset / 4u];
-    let shift = (byte_offset & 2u) * 8u;
```
reeselevine, line 14: I don't have much of an opinion either way about whether to include the `u` suffix, but is there a reason for this change?

ggml/src/ggml-webgpu/wgsl-shaders/common_decls.tmpl
```diff
@@ -10,26 +10,49 @@ fn get_byte_i32(value: u32, index: u32) -> i32 {
...
+fn load_src0_f16_as_f32_at(byte_offset: u32) -> f32 {
+    let word = src0[byte_offset / 4];
+    let shift = (byte_offset & 0x2) * 8;
+    let d_bits = (word >> shift) & 0xFFFF;
+    return unpack2x16float(d_bits)[0];
+}
+#endif
+
+#ifdef U32_DEQUANT_HELPERS_SRC
+fn load_src_u32_at(byte_offset: u32) -> u32 {
```
reeselevine: to avoid duplicating this and the `src0` helpers, can we pass in the buffer as a pointer instead? Not sure if this will work with the strict type requirements.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat.wgsl
```diff
@@ -20,11 +20,12 @@ fn multiply_add(src0_idx_base: u32, src1_idx_base: u32, offset: u32) -> f32 {
 
 #ifdef Q4_0
 fn multiply_add(src0_idx_base: u32, src1_idx_base: u32, offset: u32) -> f32 {
```
reeselevine, line 22: At this point some of these quantization types here are completely deprecated on all systems, we might just get rid of them instead.

ggml/src/ggml-webgpu/wgsl-shaders/unary.wgsl
```diff
@@ -107,7 +107,8 @@ fn main(@builtin(global_invocation_id) gid: vec3<u32>) {
     let res = src[params.offset_src + src_idx] / (1.0 + exp(-src[params.offset_src + src_idx]));
 #endif
 #ifdef EXP
-    let res = exp(src[params.offset_src + src_idx]);
+    let src_f32 = f32(src[params.offset_src + src_idx]);
+    let res = TYPE(exp(src_f32));
```
reeselevine, line 111: I do know that the increased precision is necessary to pass tests on some systems, but this hasn't been required so far on any of the CI nodes here. Can you explain in a little more detail why these operations need to have higher precision? Does it lead to degraded/incoherent model output?

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -1087,6 +1087,126 @@ class ggml_webgpu_shader_lib {
                 defines.push_back("BLOCK_SIZE=1u");
                 variant += "_i32";
                 break;
+            case GGML_TYPE_Q4_0:
+                defines.push_back("BYTE_HELPERS");
+                defines.push_back("U32_DEQUANT_HELPERS_SRC");
+                defines.push_back("SRC_TYPE=u32");
```
reeselevine: if you're moving the src type to u32 (which is great btw, since it's much more portable), do we need the struct types still?

reeselevine: Also there's probably a way to avoid all the duplication in these defines

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -1568,9 +1688,109 @@ class ggml_webgpu_shader_lib {
                 defines.push_back("FLOAT");
                 variant += "_f16";
                 break;
+            // Types with f16 fields storing packed integers — use raw u32 access
```
reeselevine: this seems to be a classic LLM comment, talking about old functionality that just ends up confusing anyone reading this now.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -1568,9 +1688,109 @@ class ggml_webgpu_shader_lib {
...
+                defines.push_back("BYTE_HELPERS");
+                defines.push_back("U32_DEQUANT_HELPERS");
+                defines.push_back("IQ4_NL");
+                defines.push_back("IQ4_NL_GRID");
+                variant += "_iq4_nl";
+                break;
             default:
                 {
-                    // quantized types
+                    // Safe struct-based types (all u32 fields, no NaN risk):
```
reeselevine: let's say "more portable" instead of NaN risk

tests/test-backend-ops.cpp
```diff
@@ -3065,6 +3065,11 @@ struct test_bin_bcast : public test_case {
     double max_maa_err() override {
         return op == ggml_add ? 1e-4 : 1e-3;
     }
+
+    // For op DIV and F16: due to hardware ULP error during divsion, need to allow a higher nmse error for the gradients.
+    double max_nmse_err() override {
```
reeselevine: hmm I don't think we want to modify test-backend-ops, can you explain more why this was added?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -99,6 +99,27 @@ static inline void compute_2d_workgroups(uint32_t total_wg, uint32_t max_per_dim
 
 /* End Constants */
 
+static inline wgpu::CallbackMode ggml_webgpu_callback_mode() {
```
reeselevine, line 100: Some of the changes here are going to conflict with https://github.com/ggml-org/llama.cpp/pull/21533, and generally some of the changes that busy-wait on the CPU have already been removed.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -258,6 +279,9 @@ struct webgpu_global_context_struct {
     wgpu::Buffer         get_tensor_staging_buf;
     // Global mutex for pipeline and staging buffer, will be refactored to exclude pipeline caches.
     std::recursive_mutex mutex;
+    std::mutex           debug_mutex;
```
reeselevine: please don't include extra debugging code unless we decide it's very useful/important for all future contributions.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2701,7 +2799,7 @@ static std::optional<webgpu_encoded_op> ggml_webgpu_encode_node(webgpu_context
 }
 
 static ggml_status ggml_backend_webgpu_graph_compute(ggml_backend_t backend, struct ggml_cgraph * cgraph) {
-    WEBGPU_LOG_DEBUG("ggml_backend_webgpu_graph_compute(" << cgraph->n_nodes << " nodes)");
+    // WEBGPU_LOG_DEBUG("ggml_backend_webgpu_graph_compute(" << cgraph->n_nodes << " nodes)");
```
reeselevine: please don't comment out existing code

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -3751,10 +3857,23 @@ static const struct ggml_backend_reg_i ggml_backend_webgpu_reg_i = {
...
+        /* .context     = */ &ctx,
+    };
 
-    static ggml_backend_webgpu_reg_context ctx;
-    ctx.name         = GGML_WEBGPU_NAME;
-    ctx.device_count = 1;
+    // Keep one Dawn/WebGPU instance alive for the lifetime of the static backend
+    // registry. Recreating it on repeated registry lookups can invalidate
+    // adapter/device references that are still held by the backend/device layer.
+    if (ctx.webgpu_global_ctx != nullptr && ctx.webgpu_global_ctx->instance != nullptr) {
```
reeselevine, line 4064: This change generally seems reasonable, but is it actually a real problem? My understanding is that normally, the reg function is only called once by llama.cpp, and I haven't seen any errors related to this running in different contexts or in the browser.

reeselevine (commented):
Thanks for the PR! Generally, the changes seem reasonable. I left a few comments and it might have to be integrated with some other changes that will be merged first. Other than that, a couple things:

* I think this PR is too big on its own. It's ok at this point, I've reviewed it all, but in the future please try to keep PRs smaller and focused on one feature. Especially as a newer contributor, where I want to be a little more careful making sure there is nothing very wrong or malicious about the contribution. Smaller PRs also help keep the conversation and changes requested easier to talk about.
* When using LLMs for commenting/debugging, please make sure you go through and clean up what they've done before opening the PR, so it doesn't add to the overall review load.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: Thanks, I will restore the commented code later.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: Sure, this was used for the wait-any debug and I forgot to remove it. Will clean this later.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: Sure, I will wait for the merge with #21533 and then clean the edit later.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: The reason I added this was that in the inference system we built around llama.cpp and wasm, we tested a scenario where a new model was loaded to replace the model already in the system, and the system crashed. My understanding of this issue at that time was that resources like the buffer and pipelines are tied to the `wgpu::device` and `wgpu::instance` that created them. When I switched the model, `llama.cpp` created a new `device` and the system tried to use an old buffer with the new `device` or clean up old resources using the new context. Therefore, to prevent this, I think it would be safer to forces `llama.cpp` to reuse the existing `instance` and `device` etc for the entire lifetime of the system.

ggml/src/ggml-webgpu/wgsl-shaders/common_decls.tmpl
Constannnnnt: When removing the if statement to use WGSL select(), @jjhartmann removed them since it seemed that the rest of the wgsl code was not using them and thought it would make things more consistent. He personally don't have an opinion on suffix either!

ggml/src/ggml-webgpu/wgsl-shaders/common_decls.tmpl
Constannnnnt: Yes, this is a good idea. We can update this to be more genralizable across all shaders.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat.wgsl
Constannnnnt: True, should we remove the deprecated structs in common_decls?

ggml/src/ggml-webgpu/wgsl-shaders/unary.wgsl
Constannnnnt: Some of the unary test were failing when testing on Nvidia RTX 3090 locally. Was attempting to solve them, this  is related to the NMSE error too, the division op continued to fail on the machine with values just over the default error.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
Constannnnnt: Yes, we could remove the structs. I wasn't sure if that should be in this PR or a different clean up PR. 

One thing we didn't do is update _all_ the kernels to use u32, but maybe it makes sense jut to conver them all? @jjhartmann was mostly just trying to solve the NaN conicalization that happened when packing integers in f16 arrays. But could do a complete pass and remove the structs too. 


Definitely, could organize the defines a bit better.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
Constannnnnt: Right! Will remove this.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
Constannnnnt: Sounds good.

tests/test-backend-ops.cpp
Constannnnnt: This relates to the Unary changes. Where the division op was failing on my test machine with values just exceeding the default error threshold.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat.wgsl
reeselevine: yeah, we can remove the deprecated structs. And then also this naive form of matrix multiplication is not used for the float paths and some of the quantization types, in favor of the register tiling/subgroup matrix versions. But I think we can leave their removal for a later PR, this just reminded me.

ggml/src/ggml-webgpu/wgsl-shaders/common_decls.tmpl
reeselevine: cool, I'm good with removing them

ggml/src/ggml-webgpu/wgsl-shaders/unary.wgsl
reeselevine: ok sounds good, I think the increased precision is probably good for these operations, even if performance takes a slight hit

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
reeselevine: I don't think we need to move all the types to u32 right now, because these are still legacy paths. The goal is to eventually move all of the matrix/matrix-vector multiplication into the fast-path shaders, but we just haven't had the time yet to work on it for the i-quants (and there's an update I'd like to make to matrix-vector multiplication to make it more performant/in-line with other backends first too).

Similarly, for get rows the operation should be parallelized like it is for the float path now. But I actually haven't seen usecases yet for get_rows for most of the quantization types in the models I've been working with.

tests/test-backend-ops.cpp
reeselevine: this is even if you cast and do the division itself in f32 precision?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
reeselevine: ok thanks for the context, this makes sense!

tests/test-backend-ops.cpp
Constannnnnt: yeah, even with the Unary changes, the test cases are still failing on the 3090 for division ops. there might be another fix, but we agree not to change the test cases.

reeselevine: If you can, would be interesting to test this against Vulkan and/or CUDA, to see if they also fail.

Constannnnnt: Sounds good. We will do that and will share the results with CUDA later.

Constannnnnt: Hi @reeselevine, we run the test with CUDA. It appeared that binary DIV ops behaved differently on RTX 3070 and RTX 3090: in 3090, it exceeded the default error threshold while it passed all tests on RTX 3070.

Constannnnnt:
Hi @reeselevine, we addressed the concerns in the previous code and merged with the latest. Let us know if there are things that we missed and thanks for the review.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -97,6 +97,14 @@ static inline void compute_2d_workgroups(uint32_t total_wg, uint32_t max_per_dim
 
 /* End Constants */
 
+static inline wgpu::CallbackMode ggml_webgpu_callback_mode() {
+#ifdef __EMSCRIPTEN__
+    return wgpu::CallbackMode::AllowProcessEvents;
```
reeselevine, line 102: is there a reason to limit the callback mode? My understanding was that `AllowSpontaneous` encompasses all the semantics of `AllowProcessEvents` while also allowing implementations flexibility to fire them whenever they're ready?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -4055,27 +4074,34 @@ ggml_backend_reg_t ggml_backend_webgpu_reg() {
     instance_descriptor.nextInChain        = &instanceTogglesDesc;
 #endif
 
-    wgpu::Instance inst             = wgpu::CreateInstance(&instance_descriptor);
+    wgpu::Instance inst = wgpu::CreateInstance(&instance_descriptor);
+
+#ifdef __EMSCRIPTEN__
+    if (inst == nullptr) {
+        GGML_LOG_ERROR("ggml_webgpu: Failed to create WebGPU instance. Make sure either -sASYNCIFY or -sJSPI is set\n");
```
reeselevine: I know this was the previous log message but I recently removed it. I guess I originally added it because initially working on the WASM builds I was fighting with the two async modes. But now do you see much utility in it still? 

Also not sure if returning the nullptr matters too much, since it seems like on browsers that don't support WebGPU, it's the probe for the adapter that really fails.

tests/test-backend-ops.cpp
```diff
@@ -3065,6 +3065,7 @@ struct test_bin_bcast : public test_case {
     double max_maa_err() override {
         return op == ggml_add ? 1e-4 : 1e-3;
     }
+
```
reeselevine, line 3064: nit: remove

reeselevine (commented):
Thanks for the updates. Just a couple minor comments, and I'll kick of a CI run now.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: I did not see much utility. I simply handle the merge by comparing our last commit history and the latest one and then merge them. I will clean this.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: Initially, I had this for the use of `WaitAny`  in some tests with blocking concerns; it is simply more conservative from that perspective. In the current implementation, `AllowSpontaneous` should be good. I will use `AllowSpontaneous ` then.

tests/test-backend-ops.cpp
Constannnnnt: I addressed these comments; let me know if you notice any issues!

reeselevine:
looks like that last commit did a bunch of formatting on test-backend-ops. Since I'm not the maintainer of that file, probably better to avoid any formatting on it.

Constannnnnt:
> looks like that last commit did a bunch of formatting on test-backend-ops. Since I'm not the maintainer of that file, probably better to avoid any formatting on it.

oh, I see. I think it is my formatter's issue and I did not notice the git diff for that. I will correct this.

Constannnnnt:
Reverted any changes to the test ops file. It should be clean now.

reeselevine (approved):
looks good now!

CISC (approved):

PR #21418, opened 2026-04-04
common : add gemma 4 specialized parser

aldehir:
## Overview
Specialized Gemma 4 parser with various fixes.

There's a lot here, so I'll do my best to summarize.

- Removed Gemma 4 parsing from the autoparser and composed a dedicated parser. The model is sufficiently different to warrant specialized parsing.
- Rewrote the Gemma 4 normalization to a function that walks the AST and emits JSON directly without a second pass. This is cleaner and leaves the parsing details to the PEG parser.
- Added a few AST walking utilities to the PEG parser implementation.
- Reworked how `tool_responses` are built (details below).
- Added a custom chat template that supports interleaved thinking (more on this below).
- Added `<|tool_response>` as an EOG token. I noticed the model was emitting content after tool calling, and it turns out this is an EOG token in the model's config.
- Added more tests.

### Gemma 4 Tool Responses

The Gemma 4 template does not follow the usual `tool` role. Instead, it accumulates `tool_responses` in a `model` role (mapped from `assistant` in the template).

The original implementation added them to the `user` role. According to the [Prompt Formatting Guide](https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4), they should be collected during the model turn. I refactored this to build up the assistant messages to include consecutive tool responses from `tool` role messages.

### Interleaved Thinking

The current templates **do not** support interleaved thinking. However, the [Prompt Formatting Guide](https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4) mentions the following:

> Function Calling (Exception): If a single model turn involves function or tool calls, thoughts must NOT be removed between the function calls.

Interesting! I created a new template, `models/templates/google-gemma-4-31B-it-interleaved.jinja`, that supports this behavior. I tested it, and it appears to work well. The examples in the guide are sparse, so I went with what I believe is the proper format. That may change as more documentation becomes available.

For anyone doing agentic tasks, I recommend trying the interleaved template.

### Non-goal: JSON Schema Support

JSON schema support for tool calling is not included in this PR. The grammar constrains the model to generate proper tool calls, but the schema is not strictly enforced for tool call arguments. It is enforced for `response_format`, which can use the existing `json-schema-to-grammar` implementation. The reason it's not enforced for tool calling is that the model does not produce JSON in tool calls, so our existing implementation doesn't apply.

There is future work to be done here. Ideally, the schema should have an intermediate representation that we can use to generate grammars for different formats. I'm exploring this further.

## Requirements

[...]

PR #21522, opened 2026-04-06
common : preserve original Gemma 4 tool responses even when JSON-like

kiwixz:
## Overview

Before this PR, llama.cpp Gemma 4 specialized parser (#21418) would attempt to parse the string-typed tool responses as JSON; and convert it to an object if successful.  It would then be minified by the template:
```jinja
{%- if tool_response['response'] is mapping -%}
    {{- 'response:' + tool_response['name'] | default('unknown') + '{' -}}
    {%- for key, value in tool_response['response'] | dictsort -%}
        {{- key -}}:{{- format_argument(value, escape_keys=False) -}}
        {%- if not loop.last %},{% endif -%}
    {%- endfor -%}
    {{- '}' -}}
{%- else -%}
    {{- 'response:' + tool_response['name'] | default('unknown') + '{value:' + format_argument(tool_response['response'], escape_keys=False) + '}' -}}
{%- endif -%}
```

This means that anything that looks like JSON would get unexpectedly altered.
I encountered this with Gemma 4 failing to edit a JSON file; it was getting confused as the inconsistent line numbers and indentations between what it read and what it attempted to write/edit.

## Additional information

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1732,17 +1732,7 @@ struct gemma4_model_turn_builder {
     void collect_result(const json & curr) {
         json response;
         if (curr.contains("content")) {
-            const auto & content = curr.at("content");
-            if (content.is_string()) {
-                // Try to parse the content as JSON; fall back to raw string
-                try {
-                    response = json::parse(content.get<std::string>());
-                } catch (...) {
-                    response = content;
-                }
-            } else {
-                response = content;
-            }
+            response = curr.at("content");
         }
 
         std::string name;


aldehir:
Which coding agent are you using? This might be the best thing to do, but it does have negative implications for clients that send structured tool output.

kiwixz:
Decided to just wrap my tool response in a JSON.  Current behavior doesn't feel right and I'm sure others will trip on it; but this solution is not perfect either.  I have better PRs to open.

aldehir:
Just FYI, the updated Google template will do what you did here.

PR #21531, opened 2026-04-06
llama: remove per-arch tensor name lists

JohannesGaessler:
In https://github.com/ggml-org/llama.cpp/pull/20503 I added a warning that is printed when a tensor name is not properly formatted, which can happen many times suring quantization. However, this only happens when there is an attempt to format a tensor name with a layer id and that tensor is not listed in `llm_get_tensor_names` for that specific model architecture. If a tensor name is not listed for a given architecture the placeholders for e.g. the layer id are not replaced, resulting in broken tensor names. I don't think this function is providing us with any actual utility but it is causing an additional maintenance burden for model architectures. This PR makes it so that the explicit per-architecture tensor name lists are removed and that instead a tensor name is always formatted with the provided parameters.

# Requirements

- AI usage disclosure: NO


CISC (approved):

ggerganov (approved):

fairydreaming:
@JohannesGaessler I suppose this line should be updated now to avoid confusion?

https://github.com/ggml-org/llama.cpp/blob/aa00911d12198f1772905c319c0a073f87984883/docs/development/HOWTO-add-model.md?plain=1#L101

JohannesGaessler:
Yes.

PR #21533, opened 2026-04-06
ggml-webgpu: parameterize submission size and add iOS specific limits

reeselevine:
## Overview

Working on stability of the WebGPU backend on different devices/browsers, I noticed that on iOS 26, the WebGPU backend tends to crash unless the number of operations + submitted command buffers is pretty severely throttled. This PR adds support for parameterizing the number of operations per batch and inflight submissions, which is limited on iOS.

Detecting the platform/device being run on is not the easiest from WebGPU, since browsers don't give out this information easily (for example on an iPhone querying WebGPU information like device name/description just returns "apple"). So this PR adds some JavaScript directly into the WebGPU backend that queries the User-Agent string in the browser to determine if it's running on iOS.

I also plan on submitting an issue/bug report with WebKit to understand if the limitation on inflight command buffers is expected, or if it's a bug/something I'm doing wrong in the WebGPU backend here.

Edit: Issue filed against WebKit here: https://bugs.webkit.org/show_bug.cgi?id=311598

# Requirements

- AI usage disclosure:  Yes, helped with writing new function to check callback status


nikhilJain17 (approved):
Nice! From the testing it seems only iOS Safari has been sensitive but seems there is a straightforward path to tune for other platforms if needed.

PR #21534, opened 2026-04-06
vocab: add gemma4 tokenizer tests, fix edge case

pwilkin:
## Overview

There's been a lot of problems with the tokenizer for Gemma 4, so I decided to actually add tests.

## Additional information

The tests did find one more problem with UTF-8 handling, so there's a fix attached as well.

# Requirements

- AI usage disclosure: Yeah, told the AI to generate the test and it found a bug in the process.


a4lg:
I encountered a bug fixed by this PR and I'm happy to see this!

Due to this bug, some Japanese words are not passed into the model.

Examples (characters enclosed by【】 are not correctly recognized before this fix):
*   【囮】(noun: decoy)
*   蹂【躙】する (verb: overwhelm)
*   同じ穴の【狢】 (Japanese proverb similar to: birds of a feather flock together)
*   水戸光【圀】公 ([person: a beloved historical figure and folk hero](https://en.wikipedia.org/wiki/Tokugawa_Mitsukuni))

ggerganov (approved):
Please use more descriptive names for the commits and the PR titles. Ideally, stick to the format described in the https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md#pull-requests-for-maintainers

taronaeo (approved):

src/llama-vocab.cpp
```diff
@@ -658,9 +658,18 @@ struct llm_tokenizer_bpe_session {
                 const auto token = vocab.text_to_token(str);
 
                 if (token == LLAMA_TOKEN_NULL) {
+                    static const char * hex = "0123456789ABCDEF";
```
ggerganov: Move the string constant inside the branch that uses it.

pwilkin: Done.

ServeurpersoCom (approved):

PR #21019, opened 2026-03-26
common : fix split model migration

angt:
## Overview

Sadly the manifest does not list all required files, i honestly thought it was the case

Without the files listed we don't have the sha256, so if the first file is valid, and all others have the correct size, then we can assume we are good and do the migration...

## Additional information

Here my test:

```
    $ find /home/angt/.cache/llama.cpp
    /home/angt/.cache/llama.cpp
    /home/angt/.cache/llama.cpp/angt_test-split-model-stories260K_stories260K-f32-00002-of-00002.gguf
    /home/angt/.cache/llama.cpp/angt_test-split-model-stories260K_stories260K-f32-00001-of-00002.gguf
    /home/angt/.cache/llama.cpp/angt_test-split-model-stories260K_stories260K-f32-00001-of-00002.gguf.etag
    /home/angt/.cache/llama.cpp/angt_test-split-model-stories260K_stories260K-f32-00002-of-00002.gguf.etag
    /home/angt/.cache/llama.cpp/manifest=angt=test-split-model-stories260K=latest.json

    $ build/bin/llama-server
    ================================================================================
    WARNING: Migrating cache to HuggingFace cache directory
      Old cache: /home/angt/.cache/llama.cpp/
      New cache: /home/angt/.cache/huggingface/hub
    This one-time migration moves models previously downloaded with -hf
    from the legacy llama.cpp cache to the standard HuggingFace cache.
    Models downloaded with --model-url are not affected.
    ================================================================================
    migrate_file: migrated angt_test-split-model-stories260K_stories260K-f32-00001-of-00002.gguf -> /home/angt/.cache/huggingface/hub/models--angt--test-split-model-stories260K/snapshots/68c3ea2061e8c7688455fab07597dde0f4d7f0db/stories260K-f32-00001-of-00002.gguf
    migrate_file: migrated angt_test-split-model-stories260K_stories260K-f32-00002-of-00002.gguf -> /home/angt/.cache/huggingface/hub/models--angt--test-split-model-stories260K/snapshots/68c3ea2061e8c7688455fab07597dde0f4d7f0db/stories260K-f32-00002-of-00002.gguf
    migrate_old_cache_to_hf_cache: migration complete, deleting manifest: /home/angt/.cache/llama.cpp/manifest=angt=test-split-model-stories260K=latest.json

    $ find /home/angt/.cache/llama.cpp /home/angt/.cache/huggingface
    /home/angt/.cache/llama.cpp
    /home/angt/.cache/huggingface
[...]

PR #21535, opened 2026-04-06
common: fix split model loading by sorting file list

brettp:
## Overview

Fix split model loading from cache in offline mode.

Refs: #21019 #21016

## Additional information

File order is not guaranteed when listing dirs. In offline mode, files are not sorted when read from cache, which can result in the wrong part loading first if the files have changed metadata (e.g., by moving, symlinking, etc).

This is a minimal approach to always sort model parts so the correct part is loaded first when downloaded or loaded from cache, in online or offline mode. 

A test is included, but can be relocated if not in the best spot since it introduces additional dependencies in `tests/test-gguf-model-data.cpp`.

# Requirements

- AI usage disclosure: YES. AI was used to familiarize myself with the code base when decided the best injection point, to ask about sorting functions, and to write the test.

### DIFF
--- a/common/download.cpp
+++ b/common/download.cpp
@@ -511,9 +511,14 @@ static hf_cache::hf_files get_split_files(const hf_cache::hf_files & files,
     if (split.count <= 1) {
         return {file};
     }
-    hf_cache::hf_files result;
 
-    for (const auto & f : files) {
+    hf_cache::hf_files sorted_files = files;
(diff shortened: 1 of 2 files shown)


angt:
Can you test with master if it's still an issue ?

ngxson:
IMO it might be more convenient if `libllama` support loading non-first-shard. It should not be too complicated to implement

brettp:
@angt - Confirmed this is fixed by fb38d6f. Thank you!

PR #21539, opened 2026-04-07
vulkan: Support Q1_0

jeffbolznv:
## Overview

Add Q1_0 support to ggml-vulkan. Supports get_rows, set_rows, mul_mat(id). Does not support the q8_1 dp4 path (though this is probably worth adding in a followon), since we get the most benefit with smaller quants.

## Additional information

None.

# Requirements

- AI usage disclosure: I used Claude to write the majority of this change. I reviewed all the code and had it rewrite several things for consistency, and I rewrote the coopmat2 dequant function myself.

### DIFF
--- a/ggml/src/ggml-vulkan/vulkan-shaders/dequant_funcs.glsl
+++ b/ggml/src/ggml-vulkan/vulkan-shaders/dequant_funcs.glsl
@@ -87,6 +87,25 @@ vec4 dequantize4(uint ib, uint iqs, uint a_offset) {
 }
 #endif
 
+#if defined(DATA_A_Q1_0)
+vec2 dequantize(uint ib, uint iqs, uint a_offset) {
+    const float d = float(data_a[a_offset + ib].d);
+    const uint bits = uint(data_a[a_offset + ib].qs[iqs / 8u]) >> (iqs % 8u);
+    return vec2(
+        (bits & 1u) != 0u ? d : -d,
+        (bits & 2u) != 0u ? d : -d);
+}
+vec4 dequantize4(uint ib, uint iqs, uint a_offset) {
+    const float d = float(data_a[a_offset + ib].d);
+    const uint bits = uint(data_a[a_offset + ib].qs[iqs / 8u]) >> (iqs % 8u);
+    return vec4(
+        (bits & 1u) != 0u ? d : -d,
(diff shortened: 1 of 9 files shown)


ggml/src/ggml-vulkan/vulkan-shaders/dequant_funcs.glsl
```diff
@@ -454,6 +473,12 @@ vec2 get_dm(uint ib, uint a_offset) {
 }
 #endif
 
+#if defined(DATA_A_Q1_0)
+vec2 get_dm(uint ib, uint a_offset) {
+    return vec2(1, 0);
```
0cc4m: Same question as with nvfp4, why not put d here? Then it only has to be loaded once.

ggml/src/ggml-vulkan/vulkan-shaders/types.glsl
```diff
@@ -188,6 +188,22 @@ struct block_q8_0_packed16
 #define DATA_A_QUANT_LEGACY
 #endif
 
+#define QUANT_K_Q1_0 128
+#define QUANT_R_Q1_0 1
```
0cc4m, line 192: I can see how that could cause issues, but shouldn't this be 8? It might make sense for each thread to load at least a byte.

0cc4m (commented):
Please resolve the merge conflict.

ggml/src/ggml-vulkan/vulkan-shaders/dequant_funcs.glsl
jeffbolznv: Sure, it works in this case. Done.

ggml/src/ggml-vulkan/vulkan-shaders/types.glsl
jeffbolznv: The current code only really supports R==1 or 2, and this quant doesn't really fit what R==2 wants.

For mul_mm, we currently do load a whole byte per thread (load_vec_quant = "8"). For mul_mat_vec, I think we should do whatever is simplest and then add any perf tuning/complexity to the dp4 path once that is implemented (this quant should be ALU-limited and benefit from dp4).

0cc4m (approved):

ggerganov (approved):

PR #21544, opened 2026-04-07
llama-quant : remove these checks as some arches do not have these tensors

ownia:
## Overview

Extend https://github.com/ggml-org/llama.cpp/pull/21219.

## Additional information

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/src/llama-quant.cpp
+++ b/src/llama-quant.cpp
@@ -285,7 +285,7 @@ static void llama_tensor_dequantize_impl(
 // do we allow this tensor to be quantized?
 //
 
-static bool tensor_allows_quantization(const llama_model_quantize_params * params, llm_arch arch, const ggml_tensor * tensor) {
+static bool tensor_allows_quantization(const llama_model_quantize_params * params, [[maybe_unused]] llm_arch arch, const ggml_tensor * tensor) {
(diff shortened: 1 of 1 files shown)


src/llama-quant.cpp
```diff
@@ -285,7 +285,7 @@ static void llama_tensor_dequantize_impl(
 // do we allow this tensor to be quantized?
 //
 
-static bool tensor_allows_quantization(const llama_model_quantize_params * params, llm_arch arch, const ggml_tensor * tensor) {
+static bool tensor_allows_quantization(const llama_model_quantize_params * params, [[maybe_unused]] llm_arch arch, const ggml_tensor * tensor) {
```
JohannesGaessler, line 288: Wouldn't it make more sense to just remove the argument?

ownia: Yes it makes sense. However, I've just noticed that you submitted the PR https://github.com/ggml-org/llama.cpp/pull/21531, perhaps we could drop my PR if yours can be merged?

JohannesGaessler: I would also be fine with closing this PR if mine is merged.

PR #21562, opened 2026-04-07
ggml-cpu: add Q1_0 AVX2 fast path

elusznik:
## Overview

Adds an AVX2 SIMD fast path for `ggml_vec_dot_q1_0_q8_0()` in `ggml/src/ggml-cpu/quants.c`.

Q1_0 was missing an x86 kernel and fell back to a scalar loop. This patch implements the fast path using `bytes_from_bits_32()` and `mul_sum_i8_pairs_float()` helpers — added in this commit alongside the fast path itself — keeping it minimal and consistent with the q4/q5 kernel style. The scalar fallback remains intact for non-AVX2 builds, and the AVX2 path degrades gracefully to a mul+add sequence when FMA is not available.

Benchmark (AMD Ryzen 7 5800X, Bonsai-8B Q1_0, 16 threads):

`test-quantize-perf --type q1_0 --op vec_dot_q -4`:

| Size    | Baseline (cycles/32) | AVX2 (cycles/32) | Speedup |
|---------|---------------------:|------------------:|--------:|
| 4 KB    |                104.4 |               6.3 |  ~16.5x |
| 64 KB   |                103.6 |               5.6 |  ~18.5x |
| 2.5 MB  |                104.8 |               5.8 |  ~18.1x |
| 250 MB  |                105.6 |               6.0 |  ~17.6x |

`llama-server --threads 16 --ctx-size 512`:

| Metric      | Baseline | AVX2   | Speedup |
|-------------|---------:|-------:|--------:|
| Prompt eval |  1.24/s  | 18.64  |   ~15x  |
| Generation  |  1.13/s  | 18.01  |   ~16x  |

## Additional information

Follow-up to the existing ARM NEON Q1_0 implementation. The x86 AVX2 path uses the same algorithm adapted for x86 intrinsics.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/ggml/src/ggml-cpu/quants.c
+++ b/ggml/src/ggml-cpu/quants.c
@@ -22,6 +22,42 @@
 
 #define UNUSED GGML_UNUSED
 
+#if defined(__AVX2__)
+#include <immintrin.h>
+
+// horizontally add 8 floats
+static inline float hsum_float_8(const __m256 x) {
+    __m128 res = _mm256_extractf128_ps(x, 1);
+    res = _mm_add_ps(res, _mm256_castps256_ps128(x));
+    res = _mm_add_ps(res, _mm_movehl_ps(res, res));
+    res = _mm_add_ss(res, _mm_movehdup_ps(res));
+    return _mm_cvtss_f32(res);
+}
+
+// 32 packed bits -> 32 bytes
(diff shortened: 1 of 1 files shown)


am17an:
You're adding to the wrong file, you need to add in `ggml/src/ggml-cpu/arch/x86/quants.c`, and some of the functions already exist there

Issue #21225, opened 2026-03-31
Feature Request: use of ccache for build in Dockerfiles


t-hofmann:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

In order to speedup repeated builds, add ccache to the Dockerfiles.

Please note: I am not familiar with the project's build environment/needs.

### Motivation

speedup repeated builds

### Possible Implementation

An approach could involve the following changes in the builder-stage:
- install `ccache`
- use a mount in the build step for caching like:

```Dockerfile
RUN --mount=type=cache,target=/root/.cache/ccache ...
```

add ccache directory to .gitignore

t-hofmann:
still relevant

PR #21563, opened 2026-04-07
optionally enable ccache for use in Dockerfiles

t-hofmann:
## Overview

This pull-request addresses #21225, which is about faster recompilation in `Dockerfiles` by enabling a build cache.

## Additional information

The solution introduces a new build-time argument `CCACHE_ENABLED` (default: `false`). Which if `true` leads to installation of `ccache` in the Dockerfiles.

A cache-mount was added to the actual build step `RUN ... make ...`. The cache will only be used if `ccache` is installed, so the cache-mount can unconditionally be added there.

In `s390x.Dockerfile` `ccache` was already installed and always used, - that behaviour is unchanged. - Would be happy to change if need be.

Only `cuda.Dockerfile` is manually tested by compiling, starting and accessing the web-interface.

# Requirements

- AI usage disclosure: NO


mikhail-shevtsov-wiregate:
@t-hofmann Do I understand correctly that `RUN --mount=type=cache,target=/root/.cache/ccache` will use github cache? If yes - I think that it's limited to 10GB and it's already full.

One of the option is to leverage docker image inline cache.

t-hofmann:
The `RUN --mount=type=cache ...` is a docker mechanism, therefore that should be the thing you address via "docker image inline cache".

Regarding the "github cache": Neither do I know how github caches in this context nor am I in the position to test if the change influences its usage.

apropos merge below: please be aware, that I am not the author of the one-API related changes

kyle-basis:
Cache mounts do not use GitHub cache unless https://github.com/reproducible-containers/buildkit-cache-dance is used.

PR #21566, opened 2026-04-07
CUDA: check for buffer overlap before fusing

am17an:
## Overview

When doing GEMV fusion for gate + up + glu, the src buffer can overlap with the dst buffer. This PR adds a check so that fusion can be skipped in case this happens. Saw this happening in Gemma4 f16 models, but it can happen to other models as well. 

## Additional information

# Requirements

- AI usage disclosure:  YES, I identified the issue and asked AI to implement the fix. I reviewed the output and also checked it solved the issue outlined by @ggerganov in #21321

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2164,31 +2164,6 @@ static bool ggml_cuda_should_fuse_mul_mat(const ggml_tensor * ffn_up,
         return false;
     }
 
-    // Check that the fused output does not overlap with source tensors
-    auto overlaps = [](const ggml_tensor * a, const ggml_tensor * b) -> bool {
-        if (!a->data || !b->data) {
-            return false;
-        }
-        const uintptr_t a_start = (uintptr_t) a->data;
-        const uintptr_t a_end   = a_start + ggml_nbytes(a);
-        const uintptr_t b_start = (uintptr_t) b->data;
-        const uintptr_t b_end   = b_start + ggml_nbytes(b);
-        return a_start < b_end && b_start < a_end;
-    };
-
-    if (overlaps(glu, ffn_up->src[0]) || overlaps(glu, ffn_up->src[1]) || overlaps(glu, ffn_gate->src[0])) {
-        return false;
-    }
-
-    if (has_bias) {
-        const ggml_tensor * up_bias_data = ffn_up_bias->src[0] == ffn_up ? ffn_up_bias->src[1] : ffn_up_bias->src[0];
-        const ggml_tensor * gate_bias_data =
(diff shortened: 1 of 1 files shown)


JohannesGaessler (commented):
Don't we already have a function `ggml_cuda_check_fusion_memory_ranges` for this?

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2164,6 +2164,31 @@ static bool ggml_cuda_should_fuse_mul_mat(const ggml_tensor * ffn_up,
...
 
+    // Check that the fused output does not overlap with source tensors
+    auto overlaps = [](const ggml_tensor * a, const ggml_tensor * b) -> bool {
+        if (!a->data || !b->data) {
+            return false;
+        }
+        const uintptr_t a_start = (uintptr_t) a->data;
+        const uintptr_t a_end   = a_start + ggml_nbytes(a);
+        const uintptr_t b_start = (uintptr_t) b->data;
+        const uintptr_t b_end   = b_start + ggml_nbytes(b);
```
ggerganov: For future-proof-ness it's better to use `ggml_backend_buft_get_alloc_size()` instead of `ggml_nbytes()`. Same in `ggml_cuda_check_fusion_memory_ranges()`.

pwilkin:
Building right now together with the tokenizer fix, going to check on my long context cases in a moment.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3308,6 +3308,71 @@ static bool ggml_cuda_topk_moe_fusion(const struct ggml_cgraph * cgraph, int nod
...
+        if ((b_start <= a_start && a_start < b_end) || (a_start <= b_start && b_start < a_end)) {
+            return true;
+        }
+
+        return false;
+    };
+
+    bool is_ok = true;
+    // exception for topk-moe, as each row is read entirely before writing
+    if (ggml_nrows(cgraph->nodes[node_idx]) == 1 && is_topk_moe) {
```
JohannesGaessler, line 3334:
```suggestion
    // for topk-moe each thread only writes back to its own memory so an in-place operation is okay
    if (is_topk_moe && ggml_nrows(cgraph->nodes[node_idx]) == 1) {
```

To me this wording is more clear. Also I thing the code is more intuitive if you put the flag at the beginning of the conditional.

JohannesGaessler (approved):

ggml/src/ggml-cuda/ggml-cuda.cu
am17an: I don't think the suggested comment is correct, because we read `n_experts` and write back `n_expert_used`

ggerganov (approved):

EthanBlazkowicz:
This PR fixed gemma 4 with flash attention on for me!

PR #21568, opened 2026-04-07
llama: fit ctx size for CPU only

JohannesGaessler:
Alternative to https://github.com/ggml-org/llama.cpp/pull/19711#issuecomment-3977588416 .

I think the correct way to reduce context size for CPU-only builds is to accumulate the host buffer types and to compare those vs. total system memory. This PR is currently only partially tested (and thus a draft) because I don't have a convenient combination of model and system memory sizes ready.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/src/llama.cpp
+++ b/src/llama.cpp
@@ -91,12 +91,16 @@ static std::vector<llama_device_memory_data> llama_get_device_memory_data(
         throw std::runtime_error("failed to create llama_context from model");
     }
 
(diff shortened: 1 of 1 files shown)


taronaeo (approved):
Tested on 2 GB and 32 GB memory systems respectively. Both working as intended.

ggerganov (approved):

taronaeo (approved):

PR #21572, opened 2026-04-07
vulkan: Programmatically add RoundingModeRTE to all shaders when the device supports it

jeffbolznv:
## Overview

Replace the current RTE16 handling with something that applies to all shaders.

## Additional information

I don't have easy access to a Turing system where lack of RTE16 would cause a failure, but it should be immediately obvious in CI if this is broken.

I was disappointed to find that there doesn't seem to be a good way to include spirv.hpp. It's in a different location in windows and linux Vulkan SDK installs, and for linux without the Vulkan SDK I don't think it's available with the current packages we assume are installed (though I'm not confident of this). Is there a better way?

# Requirements

- AI usage disclosure: YES, I used Claude to implement most of this change, with some directed bugfix suggestions and manual edits.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -494,7 +494,7 @@ On Debian / Ubuntu, you can install the required dependencies using:
 sudo apt-get install libvulkan-dev glslc spirv-headers
 ```
 
-SPIRV-Headers (`spirv/unified1/spirv.hpp`) are required for the Vulkan backend and are **not** always pulled in by the Vulkan loader dev package alone. Other distros use names such as `spirv-headers` (Arch), `spirv-headers-devel` (Fedora / openSUSE), or `libspirv-headers-dev` (some Debian derivatives). On Windows, the LunarG Vulkan SDK’s `Include` directory already contains these headers.
+SPIRV-Headers (`spirv/unified1/spirv.hpp`) are required for the Vulkan backend and are **not** always pulled in by the Vulkan loader dev package alone. Other distros use names such as `spirv-headers` (Ubuntu / Debian / Arch), or `spirv-headers-devel` (Fedora / openSUSE). On Windows, the LunarG Vulkan SDK’s `Include` directory already contains these headers.
 
 #### Common steps
 
--- a/docs/build.md
+++ b/docs/build.md
@@ -456,7 +456,8 @@ pacman -S git \
     mingw-w64-ucrt-x86_64-gcc \
     mingw-w64-ucrt-x86_64-cmake \
     mingw-w64-ucrt-x86_64-vulkan-devel \
-    mingw-w64-ucrt-x86_64-shaderc
+    mingw-w64-ucrt-x86_64-shaderc \
+    mingw-w64-ucrt-x86_64-spirv-headers
 ```
 
 Switch into the `llama.cpp` directory and build using CMake.
@@ -490,8 +491,10 @@ First, follow the official LunarG instructions for the installation and setup of
 
 On Debian / Ubuntu, you can install the required dependencies using:
 ```sh
-sudo apt-get install libvulkan-dev glslc
+sudo apt-get install libvulkan-dev glslc spirv-headers
 ```
+
+SPIRV-Headers (`spirv/unified1/spirv.hpp`) are required for the Vulkan backend and are **not** always pulled in by the Vulkan loader dev package alone. Other distros use names such as `spirv-headers` (Arch), `spirv-headers-devel` (Fedora / openSUSE), or `libspirv-headers-dev` (some Debian derivatives). On Windows, the LunarG Vulkan SDK’s `Include` directory already contains these headers.
 
 #### Common steps
 
--- a/.devops/vulkan.Dockerfile
+++ b/.devops/vulkan.Dockerfile
@@ -7,7 +7,7 @@ RUN apt update && apt install -y git build-essential cmake wget xz-utils
 
 # Install SSL and Vulkan SDK dependencies
 RUN apt install -y libssl-dev curl \
-    libxcb-xinput0 libxcb-xinerama0 libxcb-cursor-dev libvulkan-dev glslc
+    libxcb-xinput0 libxcb-xinerama0 libxcb-cursor-dev libvulkan-dev glslc spirv-headers
 
 # Build it
 WORKDIR /app
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -318,7 +318,7 @@ jobs:
         id: depends
         run: |
           sudo apt-get update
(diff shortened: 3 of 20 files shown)


jeffbolznv:
I don't know what's going on with the CI failure. Last time I had a failure in llama-save-load-state it was due to different shaders being used in each run (due to fusion issues). I don't see how this change would have triggered a similar issue. I haven't been able to reproduce it locally on a 4070.

0cc4m:
I like it, it's cleaner. I can compile and run it successfully, but I don't know if I have hardware that would fail the tests without rte. This also needs a rebase.

masamaru-san:
This PR works excellectly for llama.cpp 👍, but it causes issues in projects that use GGML as a submodule, such as stable-diffusion.cpp. This is because CMake's `FetchContent` contaminates the **parent** work tree.

Adding only the necessary `spv` definitions to `ggml-vulkan.cpp` avoids the need to use `FetchContent` and prevents an increase in external dependencies. Would you consider doing this?

jeffbolznv:
I've rebased it.

I'm not sure what to do about the fetchcontent concern. Manually defining the spirv enums we need is possible, but there may be other dependencies we want to fetch at some point in the future and I'd like to have a real solution.

I think the other main way of getting dependencies is via submodules, but I'm not sure how that interacts with ggml conditionally including ggml-vulkan, and it adds a manual step to clone/fetch the submodules.

masamaru-san:
To be honest, I find it too difficult to use CMake and Git submodules properly. When I first applied this PR to ggml under stable-diffusion.cpp, I couldn't figure out what was happning in the working tree.
After asking Copilot AI and going through some trial and error, I found that `FetchContent` was the cause, but I still don't understand anything beyond that.

However, I am concerned that the `spirv.hpp` file version obtained via `FetchContent` might not be compatible with the Vulkan SDK or glslc used in each user's environment -- at least for the LunarG Vulkan SDK for Windows.
(The AI also said that `spirv.hpp` is sometimes vendor-specific, e.g. for Android -- is that true?)

jeffbolznv:
What was the actual error you saw? From what I can tell, the main concerns with fetchcontent are in case of conflicting versions being fetched. But stable-diffusion.cpp doesn't fetch spirv-headers, so I don't see why it would cause a problem.

spirv.hpp is generally backward compatible (not sure if it's 100% guaranteed, but pretty close), and I'm not aware of anything else in llama.cpp/ggml/sd depending on it. So IMO, it's more of a theoretical concern for some other project that may include ggml.

masamaru-san:
Oh, thanks for the quick reply!

Here's the specific issue I encountered:
The `deps` directory and its contents, retrieved via `fetchcontent`, are created in the build directory (i.e. `out/build/x64-Release/`).
That directory is supposed to be ignored by `.gitignore`, but for some reason(?), it ends up being added to the working tree.

There might be some kind of mistake in the `CMakeLists.txt` file for the parent `stable-diffusion.cpp`, but I can't tell for sure.

jeffbolznv:
Hmm, it _is_ supposed to be fetched to the build directory. But I don't know why it's not getting ignored on your system.

masamaru-san:
If this is just an issue with my specific environment (as is often the case with Windows), I'll handle it myself within my local-repo., so please don't worry about it.

0cc4m:
> I was disappointed to find that there doesn't seem to be a good way to include spirv.hpp. It's in a different location in windows and linux Vulkan SDK installs, and for linux without the Vulkan SDK I don't think it's available with the current packages we assume are installed (though I'm not confident of this). Is there a better way?

It might be okay to patch the location for Windows and Linux if it's in different places in the SDK, but shouldn't cmake be able to resolve that? I can't easily look at the Windows SDK since it's an exe.

On Fedora I can install `spirv-headers-devel` and get the file in `/usr/include/spirv/unified1/spirv.hpp`. On Arch and Ubuntu I can install `spirv-headers` and get the same file. On the Linux SDK I get it in `x86_64/include/spirv/unified1/spirv.hpp` and it also provides a cmake file for it.

jeffbolznv:
Just to make sure I understand what you're suggesting: Remove fetchcontent, ifdef the include to use a different path for windows vs linux, and update the build guide to tell folks to install spirv-headers or spirv-headers-devel?

0cc4m:
Yes, basically. Or is there a good reason not to? I can't check how it works on Windows, currently.

jeffbolznv:
On windows I have the file in C:\vulkansdk\1.4.341.1\Include\spirv-headers. OK, I'll try this change.

jeffbolznv:
The CI-intel failures are on selfhosted systems and I think these haven't installed the spirv-headers package. So this is what I was concerned about in the original description. We could continue down this path and just install it on that system, but I'm slightly worried about folks filing a ton of issues about build failures.

docs/build.md
```diff
@@ -490,9 +491,11 @@ First, follow the official LunarG instructions for the installation and setup of
 
 On Debian / Ubuntu, you can install the required dependencies using:
 ```sh
-sudo apt-get install libvulkan-dev glslc
+sudo apt-get install libvulkan-dev glslc spirv-headers
 ```
 
+SPIRV-Headers (`spirv/unified1/spirv.hpp`) are required for the Vulkan backend and are **not** always pulled in by the Vulkan loader dev package alone. Other distros use names such as `spirv-headers` (Arch), `spirv-headers-devel` (Fedora / openSUSE), or `libspirv-headers-dev` (some Debian derivatives). On Windows, the LunarG Vulkan SDK’s `Include` directory already contains these headers.
```
0cc4m:
```suggestion
SPIRV-Headers (`spirv/unified1/spirv.hpp`) are required for the Vulkan backend and are **not** always pulled in by the Vulkan loader dev package alone. Other distros use names such as `spirv-headers` (Ubuntu / Debian / Arch), or `spirv-headers-devel` (Fedora / openSUSE). On Windows, the LunarG Vulkan SDK’s `Include` directory already contains these headers.
```

0cc4m (commented):
We could fetch it as a fallback, but building is already an "advanced" activity and I don't think it's a big problem to add a small dependency.

jeffbolznv:
@rillomas I think you own the selfhosted vulkan CI systems, can you install spirv-headers in them?

rillomas:
@jeffbolznv I've installed spirv-headers on both Win/Linux instances. Can you try again?

jeffbolznv:
Thanks, it's passing now.

0cc4m (approved):
LGTM. No CMake change needed at all?

CISC (approved):

jeffbolznv:
Yeah, I didn't find a need for a cmake change. Seems like the include paths are already there (the VulkanSDK on windows, or /usr/include on linux).

Issue #16839, opened 2025-10-29
Feature Request: multiple llama-server WebUI FRs


ABJ4403:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

- [x] "Continue response" button (something with icon that looks like this: [>>]) when inference is stopped
- [ ] WebUI offline caching (similiar to how hexed.it doing it, for example)
- [ ] (somewhat related to 2nd item) When model is loading, instead of showing just "Model is loading" text, load the homepage like usual, but there's phase where it "waits for server" (forgot what its called), we can try send the "error model is still loading" response thing there, and error handler that catches this will show in homepage "loading the model" and of course it gets rechecked every 5 seconds
- [ ] Multiple languages (assuming its not that hard or costly in size to implement)(idk if this was implemented already or not, bcz i dont see language toggle)
- [ ] For mobile users, pressing enter should not send query, for that user can press the submit button instead
- [ ] After done inferencing, make another inference to generate short summary for the chat title

### Motivation

- In case when user stopped inference or output limit (not context) reaches maximum, instead of user having to tell the AI to "continue" response, which likely break, provide "continue response" button (maybe next to "regenerate response" button) which when pressed continues the inference. I've seen this somewhere but i forgot where it was, maybe DeepSeek has it?
- Faster loading times
- Cleaner UI for showing when the model is loading
- Not all people can understand English, and some AI frontends have multiple language support
- For mobile users, the "Shift" key on almost all virtual keyboards can only be used to switch letters/symbols only it cant send literal shift key when pressing enter for example
- makes the chat title readable, but could be slow to generate, especially if long texts are involved, so this should be optional toggle. So far i only see this on ChatGPT

PR #21577, opened 2026-04-07
webui: add "Send message on Enter" setting

mourix:
## Overview

In our current webui, pressing Enter sends a message and pressing Shift+Enter inserts a newline on all platforms. This makes it impossible to format messages on mobile, since touch keyboards don't include a Shift key. A common solution apps tend to use for this is to hardcode Enter to send on web, and Enter for newline on mobile.

**This PR makes "Send message on Enter" behavior a user setting.** 

1. When enabled, the old Enter to send behavior is used. (set to default for web)
2. When disabled, Enter inserts a newline and Ctrl/Cmd+Enter sends a message. (set to default for mobile)

The chat form text helper will also detect Mac/PC in order to show the applicable Ctrl/Cmd modifier key in its text.

## Additional information

This behavior was also asked in feature request #16839.

# Requirements

- AI usage disclosure: Yes, mainly for automated review and Mac Cmd detection as I own no Apple devices to test.

### DIFF
--- a/tools/server/webui/src/lib/stores/settings.svelte.ts
+++ b/tools/server/webui/src/lib/stores/settings.svelte.ts
@@ -34,10 +34,10 @@
 import { browser } from '$app/environment';
 import {
 	CONFIG_LOCALSTORAGE_KEY,
-	DEFAULT_MOBILE_BREAKPOINT,
 	SETTING_CONFIG_DEFAULT,
 	USER_OVERRIDES_LOCALSTORAGE_KEY
 } from '$lib/constants';
+import { IsMobile } from '$lib/hooks/is-mobile.svelte';
 import { ParameterSyncService } from '$lib/services/parameter-sync.service';
 import { serverStore } from '$lib/stores/server.svelte';
 import {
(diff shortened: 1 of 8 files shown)


tools/server/webui/src/lib/stores/settings.svelte.ts
```diff
@@ -122,6 +123,16 @@ class SettingsStore {
 				...savedVal
 			};
 
+			// Default sendOnEnter to false on mobile when the user has no saved preference
+			if (!('sendOnEnter' in savedVal)) {
+				const isMobile = window.matchMedia(
```
allozaur: We actually already have a `isMobile` hook in `$lib/hooks`

allozaur (changes requested):
Please use the existing `isMobile` hook + also run `npm run dev` or `bash tools/server/webui/scripts/install-git-hooks.sh` so that you have the static output built on pre-commit hook. Alternatively you can just run `npm run build` and manually commit the `tools/server/public` changes.

allozaur (approved):

ServeurpersoCom (approved):

CISC:
Failure:
https://github.com/ggml-org/llama.cpp/actions/runs/24185325452/job/70588474059

ServeurpersoCom:
> Failure: https://github.com/ggml-org/llama.cpp/actions/runs/24185325452/job/70588474059

Yes it's prettier formatting issue
Sans titre

mourix:
I'm trying to learn here... is it angry because the line is 101 chars long?

ServeurpersoCom:
> I'm trying to learn here... is it angry because the line is 101 chars long?

He's not happy because there's a discrepancy that the CI checks; this prevents him from "hiding" code different from the repo in the binary!

It's simply a case of forgetting to run "prettier" (npm run format) to ensure the code is clean and standardized. When you're really meticulous, you always run:

npm run format
npm run lint
npm run check
npm run build

Before any commit :)

mourix:
Thank you for the explanation. I will be meticulous from now on too!

PR #21582, opened 2026-04-07
server: hide embedding/rerank models from chat model selector

### LINKED ISSUE #21545: server: Add model capabilities information (instruction, embedding etc.)
### Name and Version

version: 8685 (0988accf8)
built with GNU 13.3.0 for Linux aarch64

### Operating systems

_No response_

llama-server

### Command line

```shell

```

Image

Embedding models show up in the model selection, in this case it's even the default, but the model does not work for chat. It should be hidden or shown differently.

@ggml-org/llama-webui

### First Bad Commit

_No response_

### Relevant log output

_No response_

parasol-aser:
## Summary

Fixes #21545

In router mode (multi-model), the WebUI model selector shows all loaded models including embedding and reranking models. Users can select an embedding model for chat, which doesn't work.

- **Server-side (`server-models.cpp`)**: Add `capabilities` field to the router `/models` endpoint response. Capabilities are derived from preset flags (`--embedding` → `["embedding"]`, `--rerank` → `["embedding", "rerank"]`, otherwise → `["completion"]`).
- **Client-side (`ModelsSelector.svelte`)**: Filter models in the chat selector dropdown — only show models with `"completion"` capability or with no capabilities listed (backwards compatibility with older servers).

## Test plan

- [ ] **Router mode with mixed models**: Start server with one chat model and one embedding model (`--embedding` flag). Verify only the chat model appears in the WebUI dropdown. Verify `/models` returns correct capabilities for each.
- [ ] **Reranking model**: Add a model with `--rerank` flag. Verify it is hidden from the chat selector and has `capabilities: ["embedding", "rerank"]`.
- [ ] **Single model mode (regression)**: Start server with a single model. Verify model selector works as before.
- [ ] **Backwards compatibility**: Verify models with empty capabilities array are still shown (legacy server support).
- [ ] **All embedding models**: If all loaded models are embedding-only, the chat selector should be empty (no chat-capable models available).
- [ ] Run new unit tests: `tools/server/tests/unit/test_model_capabilities.py` and `tools/server/webui/tests/unit/model-capabilities-filter.test.ts`

### DIFF
--- a/tools/server/server-models.cpp
+++ b/tools/server/server-models.cpp
@@ -1004,15 +1004,29 @@ void server_models_routes::init_routes() {
                 status["exit_code"] = meta.exit_code;
                 status["failed"]    = true;
             }
+            // Determine capabilities from preset flags
+            json capabilities = json::array();
+            std::string tmp_val;
+            if (meta.preset.get_option("LLAMA_ARG_EMBEDDINGS", tmp_val) ||
+                    meta.preset.get_option("LLAMA_ARG_RERANKING", tmp_val)) {
+                capabilities.push_back("embedding");
+                std::string rerank_val;
+                if (meta.preset.get_option("LLAMA_ARG_RERANKING", rerank_val)) {
+                    capabilities.push_back("rerank");
+                }
+            } else {
+                capabilities.push_back("completion");
+            }
+
             models_json.push_back(json {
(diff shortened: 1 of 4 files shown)


allozaur:
Potentially duplicate, but definitely related to #21567

PR #16611, opened 2025-10-16
Fix non-ASCII path handling in common argument parser on Windows

kuguma:
## Summary

Fixed the behavior of the llama.cpp general argument parser when given non-ASCII paths.

On Windows, argv strings are received in the system’s ANSI code page (ACP), so conversion to UTF-8 is required.

## Comparison of behavior

Windows 11 Pro 24H2
```
PS > Get-WinSystemLocale
LCID             Name             DisplayName
----             ----             -----------
1041             ja-JP            Japanese (Japan)
PS > [Console]::InputEncoding.EncodingName
Japanese (Shift-JIS)
PS > [Console]::OutputEncoding.EncodingName
Unicode (UTF-8)
```

b6756
```
PS > llama-cli.exe -m "..\モデル\gemma-3-4b-it-Q4_K_M.gguf"
build: 6756 (bc07349a7) with MSVC 19.44.35214.0 for x64
main: llama backend init
main: load the model and apply lora adapter, if any
gguf_init_from_file: failed to open GGUF file '..\���f��\gemma-3-4b-it-Q4_K_M.gguf'
llama_model_load: error loading model: llama_model_loader: failed to load model from ..\���f��\gemma-3-4b-it-Q4_K_M.gguf
llama_model_load_from_file_impl: failed to load model
common_init_from_params: failed to load model '..\���f��\gemma-3-4b-it-Q4_K_M.gguf', try reducing --n-gpu-layers if you're running out of VRAM
main: error: unable to load model
```

This PR
```
PS > llama-cli.exe -m "..\モデル\gemma-3-4b-it-Q4_K_M.gguf"
build: 6758 (7ae5e44d4) with MSVC 19.44.35214.0 for x64
main: llama backend init
main: load the model and apply lora adapter, if any
llama_model_loader: loaded meta data with 40 key-value pairs and 444 tensors from ..\モデル\gemma-3-4b-it-Q4_K_M.gguf (version GGUF V3 (latest))
llama_model_loader: Dumping metadata keys/values. Note: KV overrides do not apply in this output.
~~~
> 
```

## Note

This PR is related to 
- #16609.

Reproducing the issue with non-ASCII .mmproj paths in existing CLI tools requires first applying the changes from this PR. (This doesn’t apply when the tools are called internally.)

PR #16589, opened 2025-10-15
fix(ggml): use safe character conversion in ggml_fopen on Windows

sirus20x6:
This commit fixes an unsafe character conversion in the `ggml_fopen` function on Windows. The file `mode` string was being converted from `char *` to `wchar_t *` by simple casting, which is not safe for multi-byte character sets.

This change replaces the manual conversion with a call to the `ggml_mbstowcs` helper function, which properly handles UTF-8 to wide character conversion. This makes the code more robust and correct, especially for handling file paths on Windows.

PR #21583, opened 2026-04-07
fix: support non-ASCII (Unicode) file paths on Windows

### LINKED ISSUE #18571: Eval bug: Model loading fails on Windows when path contains non-ASCII (Cyrillic) characters
### Name and Version

b7539
On Windows:
Currently, llama.cpp attempts to open GGUF files using `fopen`, which fails when the path contains Cyrillic characters. gguf_init_from_file: failed to open GGUF file
'D:\Загрузки\models\llm\olmOCR-2-7B-1025-Q4_K_M.gguf'.
For example, any ASCII characters are not supported, and the app throws an error. This means that any "Russian," "Chinese," "Korean," or any other characters in the path faild to load model. 

### Operating systems

Windows

### GGML backends

Vulkan

### Hardware

Ryzen 9 5900x + RTX 4090

### Models

olmOCR-2-7B-1025-Q4_K_M
Qwen2.5-VL-7B-Instruct-Q4_K_M

On Windows, create a directory with Cyrillic characters, e.g.:

1. for exempel D:\Загрузки\models\
2. Place a .gguf model file inside.
3. Load the model using llama.cpp
4. Model loading fails with failed to open GGUF file.

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

lekot:
Fixes #18571

Revives and combines the work from #16611 (by @kuguma, approved by @ggerganov) and #16589 (by @LunarTiger), both of which went stale.

## Problem

On Windows, `argv` strings are encoded in the system's ANSI code page, which cannot represent characters outside the current locale — Cyrillic, CJK, Arabic, etc. This causes model loading to fail with `gguf_init_from_file: failed to open GGUF file` when the path contains such characters (e.g. `D:\Загрузки\models\model.gguf`).

### 1. `common/arg.cpp` — UTF-8 argv on Windows

Re-obtain the command line via `CommandLineToArgvW` and convert each argument from UTF-16 to UTF-8 at the entry point of `common_params_parse`. This ensures all file paths are correctly encoded before reaching any file I/O.

### 2. `ggml/src/ggml.c` — fix `ggml_fopen` mode conversion

The `mode` parameter (e.g. `"rb"`) was converted from `char*` to `wchar_t*` by a manual byte-by-byte cast, which is technically unsafe for multi-byte locales. Replaced with the existing `ggml_mbstowcs` helper, consistent with how `fname` is already converted.

## Test Plan

- [x] Place a `.gguf` model in a path with Cyrillic characters (e.g. `D:\Загрузки\`)
- [x] Load the model — should succeed instead of failing with "failed to open"
- [x] Paths with only ASCII characters work as before (no behavior change)

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -17,6 +17,7 @@
 #   define NOMINMAX
 #endif
 #include <windows.h>
+#include <shellapi.h>
 #endif
 
 #define JSON_ASSERT GGML_ASSERT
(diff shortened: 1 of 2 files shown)


ngxson:
superseded by https://github.com/ggml-org/llama.cpp/pull/24779

PR #21584, opened 2026-04-07
Q1_0: port CUDA kernels

pwilkin:
## Overview

CUDA kernels for Q1_0 ported from the original fork.

## Additional information

The CUDA kernels from the original fork's Q1_0_g128.

# Requirements

- AI usage disclosure: Yes, used Kimi 2.5 to compare the branch and port the kernels


am17an:
I think the original authors already had planned CUDA kernels. If so, we should let them add them

pwilkin:
Ah, no worries :)

Obsoleted by #21629

PR #20797, opened 2026-03-20
vulkan: Flash Attention DP4A shader for quantized KV cache

0cc4m:
This PR adds DP4A (integer dot product) support to the scalar FA shader, enabled if the GPU supports DP4A. It's only used for quantized KV cache (both q8_0 or both q4_0), and not for coopmat FA shaders.

I also unified the GLSL vector type name preprocessor macros because we had swapped from FLOAT_TYPE_VECx to FLOAT_TYPEVx in Flash Attention, and the old naming was getting in the way of code reuse here.

Performance graphs for q8_0 kv cache:
<details>
<summary>AMD Radeon Pro VII</summary>
provii_gemma9b
provii_glm4 7flash
provii_gptoss20b
provii_llama8b
provii_qwen3moe

</details>
<details>
<summary>AMD RX 9070 XT (without coopmat)</summary>
9070xt_llama8b
9070xt_gptoss20b
9070xt_qwen3moe
9070xt_gemma9b
9070xt_glm4 7flash

</details>
<details>
<summary>AMD RX 8060S (without coopmat)</summary>
8060s_llama8b
8060s_gptoss20b
8060s_qwen3moe
8060s_gemma9b
8060s_glm4 7flash

</details>
<details>
<summary>Intel A770</summary>
a770_glm4 7flash
a770_gemma9b
a770_qwen3moe
a770_gptoss20b
a770_llama8b

</details>
<details>
<summary>Nvidia RTX 3090 (without coopmat)</summary>
rtx3090_llama8b
rtx3090_gptoss20b
rtx3090_qwen3moe
rtx3090_gemma9b
rtx3090_glm4 7flash

</details>

PR #21605, opened 2026-04-08
vulkan: unify type macros to use Vx instead of _VECx

0cc4m:
## Overview

While working on #20797 I ran into the issue that some shaders use TYPE_VEC4 and some use TYPEV4 for type macros, which makes using code from both hard. This PR changes them to the shorter version.

# Requirements

- AI usage disclosure: NO


jeffbolznv:
The cm1 CI failure in llama-save-load-state is the same thing I saw in another PR, seems intermittent and I don't know when it started.

jeffbolznv (approved):

0cc4m:
@ggml-org/maintainers Can I get another approval?

ggerganov (approved):

PR #21611, opened 2026-04-08
CUDA: use LRU based eviction for cuda graphs

am17an:
## Overview

Since introducing graphs per node to enable multiple splits to have cuda graphs in #18934, there are cases when the node pointers in ggml_cgraph keep changing and it leads to the map being unbounded leading to memory leaks (e.g #20315) 

This PR fixes the memory leaks

## Additional information

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -28,6 +28,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <deque>
 
 #if defined(GGML_USE_HIP)
 #include "vendors/hip.h"
@@ -151,6 +152,7 @@ static int ggml_cuda_highest_compiled_arch(const int arch) {
 #define MATRIX_ROW_PADDING 512 // last row of quant. matrices is a multiple of this to avoid out-of-bounds memory accesses
 
 #define GGML_CUDA_MAX_STREAMS 8
+#define GGML_CUDA_MAX_GRAPHS 64
 
 [[noreturn]]
 void ggml_cuda_error(const char * stmt, const char * func, const char * file, int line, const char * msg);
@@ -1367,11 +1369,17 @@ struct ggml_backend_cuda_context {
     // Map from first_node_ptr to cuda_graph - allows multiple graphs per context
     // when the computation is split across CPU/GPU (e.g., with --n-cpu-moe)
     std::unordered_map<const void *, std::unique_ptr<ggml_cuda_graph>> cuda_graphs;
+    std::deque<const void *> graph_roots;
 
     ggml_cuda_graph * cuda_graph(const void * first_node_ptr) {
         auto it = cuda_graphs.find(first_node_ptr);
         if (it == cuda_graphs.end()) {
+            if (graph_roots.size() >= GGML_CUDA_MAX_GRAPHS) {
+                cuda_graphs.erase(graph_roots.front());
+                graph_roots.pop_front();
+            }
             cuda_graphs[first_node_ptr] = std::make_unique<ggml_cuda_graph>();
+            graph_roots.push_back(first_node_ptr);
             return cuda_graphs[first_node_ptr].get();
         }
         return it->second.get();
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -28,7 +28,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
-#include <deque>
+#include <queue>
 
 #if defined(GGML_USE_HIP)
 #include "vendors/hip.h"
@@ -152,7 +152,6 @@ static int ggml_cuda_highest_compiled_arch(const int arch) {
 #define MATRIX_ROW_PADDING 512 // last row of quant. matrices is a multiple of this to avoid out-of-bounds memory accesses
 
 #define GGML_CUDA_MAX_STREAMS 8
-#define GGML_CUDA_MAX_GRAPHS  128
 
 [[noreturn]]
 void ggml_cuda_error(const char * stmt, const char * func, const char * file, int line, const char * msg);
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
I think we probably need some time-based LRU eviction logic. But let's merge this now as a stopgap.

ggml/src/ggml-cuda/common.cuh
```diff
@@ -151,6 +152,7 @@ static int ggml_cuda_highest_compiled_arch(const int arch) {
 #define MATRIX_ROW_PADDING 512 // last row of quant. matrices is a multiple of this to avoid out-of-bounds memory accesses
 
 #define GGML_CUDA_MAX_STREAMS 8
+#define GGML_CUDA_MAX_GRAPHS  128
```
gaugarg-nv: I think 128 might be too low.

The TP implementation breaks each decoder layer into 2-sub graphs. So, for larger models, it will easily hit this limit.

am17an: How many layers do large models have? Is 256 a better limit?

gaugarg-nv: I have seen up to 80 layers. So, yes, 256 will be a better limit.

MaxKruse: GLM4.7 notably has 92 layers https://huggingface.co/zai-org/GLM-4.7/blob/main/config.json

JohannesGaessler: Without NCCL the number of ggml graphs is currently even higher. However, these graphs only have a single node so using CUDA graphs may not be worthwhile in the first place.

ggerganov: Probably directly implement the timestamp-based eviction:

- Keep track of when a cuda graph was most recently used
- On each compute, evict all cuda graphs that haven't been used in the last 10 seconds
- Remove the `GGML_CUDA_MAX_GRAPHS` limit

ORippler:
> I think we probably need some time-based LRU eviction logic. But let's merge this now as a stopgap.

I feel ring-buffers on their own are not going to reflect well more complex orchestration loops (e.g. libmtmd), where a lot of "never initialized" `ggml_cuda_graph` objects will be generated (cuda graphs are only used for decode).

There is a parallel LRU-based proposal at #21673 that was proposed to reflect this (but is of course more complex). Instead of being time-based, it assigns a VRAM budget.

am17an:
@JohannesGaessler can you share the command for which you saw memory increasing? I can test with the current PR

ggml/src/ggml-cuda/common.cuh
am17an: Okay implemented this. Since `std::priority_queue` doesn't support random access, we need to a do a little bit of book-keeping on the side and keep some stale entries in the queue, but I think it should be okay.

JohannesGaessler:
@am17an just start the llama.cpp server with any model and incrementally fill up the context with generations.

am17an:
I tested with 4x 4090 GPUs, on master memory steadily increases and with this PR it is constant after a point

ggml/src/ggml-cuda/common.cuh
```diff
@@ -1368,11 +1369,41 @@ struct ggml_backend_cuda_context {
...
+    std::priority_queue<
+        std::pair<int64_t, const void *>,
+        std::vector<std::pair<int64_t, const void *>>,
+        std::greater<>
+    > graph_lru_cache;
+
+    std::unordered_map<const void *, int64_t> graph_last_used_time;
+
     ggml_cuda_graph * cuda_graph(const void * first_node_ptr) {
+        int64_t time_now = ggml_time_us();
```
JohannesGaessler:
```suggestion
        const int64_t time_now = ggml_time_us();
```

ggml/src/ggml-cuda/common.cuh
```diff
@@ -1368,11 +1369,41 @@ struct ggml_backend_cuda_context {
...
+        while (!graph_lru_cache.empty() && time_now - graph_lru_cache.top().first >= 10'000'000) {
+            const auto & [ts, node_ptr] = graph_lru_cache.top();
+            graph_lru_cache.pop();
+
+            // lazy delete
+            auto it = graph_last_used_time.find(node_ptr);
+            if (it != graph_last_used_time.end() && ts == it->second) {
+                cuda_graphs.erase(node_ptr);
+                graph_last_used_time.erase(node_ptr);
+            }
```
JohannesGaessler: Unless I am misunderstanding something the contents of `cuda_graphs`, `graph_lru_cache`, and `graph_last_used_time` should always be consistent in terms of which graphs they contain. If so, this should be an assert rather than a check. Ideally we would be using data structures that automatically enforce consistency (does not need to be in this PR).

am17an: Time is in microseconds, so it is possible to call a graph within the same microsecond. I did face a sporadic crash, so changed to std::find.

ggml/src/ggml-cuda/common.cuh
ggerganov: I was thinking about a much simpler implementation:

- Add a timestamp to `ggml_cuda_graph`
- LRU purging is loop over `cuda_graphs` and remove outdated entires
- Update timestamp each time a graph is referenced

I highly doubt the priority queue here has any advantage in terms of performance since we are dealing with very small number of entries, while it makes the logic quite complicated.

am17an: `-sm tensor` on peak was at 800 cuda graphs with 4x 4090 running gpt-oss-120b, so I don't think we can loop over all entries? We have to keep them sorted somehow

ggerganov: We don't need to purge on each iteration. For example, we can purge only if X seconds have passed since the last purge, so even with 800 graphs or more in the container, it should not be a problem to loop over all of them from time to time.

am17an: Ok I simplified this, and did not see any performance degradation

ggerganov (approved):

am17an:
@ggml-org/ggml-cuda can I get another approval?

gaugarg-nv:
Looks good to me.

JohannesGaessler (approved):

PR #21612, opened 2026-04-08
gemma : perform per-layer projections in the first layer

ggerganov:
## Overview

Previously, we performed this computations on the output layer which causes inefficiencies for multi GPU systems and partial offload. Now we perform the per-layer projections on the device assigned to layer 0. The number of graph splits is generally reduced.

Affected models:

- Gemma 3n
- Gemma 4 E2B
- Gemma 4 E4B

# Requirements

- AI usage disclosure: NO


src/models/gemma4-iswa.cpp
```diff
@@ -19,14 +26,14 @@ llm_build_gemma4_iswa::llm_build_gemma4_iswa(const llama_model & model, const ll
...
     auto * inp_attn = build_attn_inp_kv_iswa();
 
-    // inp_per_layer shape: [n_embd_per_layer, n_tokens, n_layer]
+    ggml_tensor * inp_out_ids = build_inp_out_ids();
+
     ggml_tensor * inp_per_layer = nullptr;
-    if (model.tok_embd_per_layer) {
-        inp_per_layer = project_per_layer_inputs(inpL, get_per_layer_inputs());
+    if (model.per_layer_tok_embd) {
+        inp_per_layer = build_inp_per_layer();
```
ngxson, line 33:
```suggestion
        inp_per_layer = build_inp_per_layer();
        // inp_per_layer will be projected inside the loop of layers
```

Otherwise, we can also split this into 2 different variables to make the code more clear: `inp_per_layer`, and after projected: `inp_per_layer_proj`

ngxson (approved):
Just out of curiosity, can we also use the `ggml_build_forward_expand` trick to do the same thing? Similar to:

https://github.com/ggml-org/llama.cpp/blob/3bd9aa1f9250cd15f5371f3622d73d954b68a747/src/llama-graph.cpp#L2020-L2024

Th idea is to keep `project_per_layer_inputs` on top, outside of the loop, and do `ggml_build_forward_expand(gf, inp_per_layer)` right before `build_attn` of the first layer.

src/models/gemma4-iswa.cpp
ggerganov: I moved it back at the top.

ggerganov:
@taronaeo I think I noticed the self-hosted runner failing to compile the CUDA code a couple of times:

https://github.com/ggml-org/llama.cpp/actions/runs/24130554576/job/70405824355?pr=21612#step:3:1216

Any ideas?

ngxson (approved):

PR #21616, opened 2026-04-08
Add support for Reka Edge 2603

kwajiehao:
# Overview

This PR adds support for Reka Edge 2603.

Model overview:
  - LLM: Llama-style, 32 layers, GQA with 32/8 heads, RoPE, SiLU                                      
  - Vision Encoder: ConvNeXt V2-based (4-stage, depths [3, 3, 27, 3], with GRN)                                   
  - Projector: 2-layer MLP with GELU (2816 → 4096 → 4096), adaptive 8×8 pooling → 64 visual tokens per tile 

Note that while the template contains reasoning chunks, the Reka Edge 2603 model does not support reasoning.

The code changes in this PR are purely additive.

Useful links:
- [HuggingFace](https://huggingface.co/RekaAI/reka-edge-2603)
- [Script to convert Reka Edge to GGUF](https://huggingface.co/RekaAI/reka-edge-2603/blob/main/convert_reka_vlm_to_gguf.py)
- [Script to quantize LLM text decoder](https://huggingface.co/RekaAI/reka-edge-2603/blob/main/quantize_reka_q4_last8_q8.sh) 

## Other changes

I made a few additions to the Jinja engine to get Reka Edge working with llama.cpp: string repetition, ensure_ascii in tojson, and int() as a no-op on integers. This is needed for Reka Edge because of how our chat template is constructed.

I'm happy to split this PR up so that they can be reviewed separately. I just wanted to include them up front first because if I made a PR to add Jinja engine features that wouldn't make sense without the context of Reka Edge support.

## Tests
Beyond basic image query vibe checks, I ran VQA v2, RefCoco A & B, mobile-actions benchmarks against Reka Edge running on the llama-server.

## Contributing Guidelines
I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md). I've run the CPU-only build of the full CI on my local machine.

**AI usage disclosure**: Claude Code and Codex were used to explore the codebase and generate boilerplate and drafts for the functions and tests introduced. These changes were then substantively edited, fleshed out, and cleaned up by me, and tested extensively against our chosen benchmarks to ensure correctness.


tools/mtmd/models/yasa2.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+    // Keep GRN math in fp32 for stability; fp16/bf16 accumulation can drift.
+    ggml_tensor * inp_f = ggml_cast(ctx0, inp, GGML_TYPE_F32);
+    ggml_tensor * sq = ggml_mul(ctx0, inp_f, inp_f);
+    ggml_tensor * sq_flat = ggml_reshape_4d(ctx0, sq, wdim * hdim, cdim, 1, bdim);   // [WH,C,1,B]
+    ggml_tensor * gx = ggml_sum_rows(ctx0, sq_flat);                                   // [1,C,1,B]
+    gx = ggml_sqrt(ctx0, gx);                                                           // [1,C,1,B]
+
+    ggml_tensor * gx_ch_first = ggml_permute(ctx0, gx, 1, 0, 2, 3);                    // [C,1,1,B]
+    gx_ch_first = ggml_cont(ctx0, gx_ch_first);
```
ngxson, line 67: some `ggml_cont` seem to be redundant, please verify if it still works after removing these `ggml_cont`

tools/mtmd/models/yasa2.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include "models.h"
+
+static ggml_tensor * add_channel_bias(
+        ggml_context * ctx0,
+        ggml_tensor * x_whcb,
+        ggml_tensor * b_c) {
+    if (!b_c) {
+        return x_whcb;
+    }
+    ggml_tensor * bc = ggml_cast(ctx0, b_c, x_whcb->type);
```
ngxson: normally you shouldn't need `ggml_cast`. if this is a weight tensor, enforce the type upon conversion (python script)

tools/mtmd/models/yasa2.cpp
```diff
@@ -0,0 +1,217 @@
...
+    ggml_tensor * out = ggml_add(ctx0, inp_f, xnx);
+    return ggml_cast(ctx0, out, inp->type);
+}
+
+ggml_cgraph * clip_graph_yasa2::build() {
+    ggml_tensor * cur = build_inp_raw();
+
+    // Patch embedding Conv2d(kernel=4, stride=4)
+    ggml_tensor * patch_w = ggml_cast(ctx0, model.yasa_patch_w, GGML_TYPE_F32);
+    ggml_tensor * patch_inp = ggml_cast(ctx0, cur, GGML_TYPE_F32);
```
ngxson: enforce the type of `patch_w` upon conversion

for, `patch_inp`, it's always be F32, there is no need to cast

tools/mtmd/models/yasa2.cpp
```diff
@@ -0,0 +1,217 @@
...
+    // Exact ConvNeXtV2 GRN:
+    // Gx = ||x||_2 over spatial dims (W,H), Nx = Gx / (mean_c(Gx) + eps)
+    // y  = w * (x * Nx) + b + x
+    const int64_t wdim = inp->ne[0];
+    const int64_t hdim = inp->ne[1];
+    const int64_t cdim = inp->ne[2];
+    const int64_t bdim = inp->ne[3];
+
+    // Keep GRN math in fp32 for stability; fp16/bf16 accumulation can drift.
+    ggml_tensor * inp_f = ggml_cast(ctx0, inp, GGML_TYPE_F32);
```
ngxson: likely a redundant cast

tools/mtmd/models/yasa2.cpp
```diff
@@ -0,0 +1,217 @@
...
+    const int kw = std::max(1, (int) cur->ne[0] / pooled_w);
+    const int kh = std::max(1, (int) cur->ne[1] / pooled_h);
+    cur = ggml_pool_2d(ctx0, cur, GGML_OP_POOL_AVG, kw, kh, kw, kh, 0, 0);
+
+    // [W,H,C,B] -> [C,T,B]
+    ggml_tensor * tokens = ggml_reshape_3d(ctx0, cur, cur->ne[0] * cur->ne[1], cur->ne[2], cur->ne[3]);
+    tokens = ggml_permute(ctx0, tokens, 1, 0, 2, 3);
+    tokens = ggml_cont(ctx0, tokens);
+    cb(tokens, "yasa2_tokens", -1);
+    cb(tokens, "yasa2_tokens_pos", -1);
```
ngxson: redundant cb

tools/mtmd/clip.cpp
```diff
@@ -2640,6 +2701,8 @@ int clip_n_output_tokens_x(const struct clip_ctx * ctx, struct clip_image_f32 *
             return (img->nx / params.patch_size) / 2;
         case PROJECTOR_TYPE_STEP3VL:
             return img->nx / (params.patch_size * params.n_merge);
+        case PROJECTOR_TYPE_YASA2:
+            return 8;
```
ngxson: `tokens_x/y` is only valid for m-rope models, please revert these line

tools/mtmd/models/yasa2.cpp
```diff
@@ -0,0 +1,217 @@
...
+                b2 = ggml_reshape_3d(ctx0, b2, b2->ne[0], 1, 1);             // [C,1,1]
+                b2 = ggml_repeat(ctx0, b2, tok);
+                tok = ggml_add(ctx0, tok, b2);
+            }
+            x = ggml_permute(ctx0, tok, 1, 0, 2, 3);                         // [T,C,B]
+            x = ggml_cont(ctx0, x);
+            x = ggml_reshape_4d(ctx0, x, w, h, tok->ne[0], b);               // [W,H,C,B]
+
+            cur = ggml_add(ctx0, res, x);
+            ggml_format_name(cur, "yasa2_stage%zu_blk%zu_out", s, bi);
```
ngxson: %zu has no effect here

tools/mtmd/models/yasa2.cpp
```diff
@@ -0,0 +1,217 @@
...
+ggml_tensor * clip_graph_yasa2::layer_norm_channels(ggml_tensor * inp, ggml_tensor * w, ggml_tensor * b, float eps) {
+    // Match HF ConvNextLayerNorm(channels_first):
+    // u = mean_c(x), s = mean_c((x-u)^2), x = (x-u)/sqrt(s+eps)
+    // cast back to input dtype before affine.
+    ggml_tensor * cur = ggml_cast(ctx0, inp, GGML_TYPE_F32);
+    cur = ggml_permute(ctx0, cur, 2, 1, 0, 3); // [W,H,C,B] -> [C,H,W,B]
+    cur = ggml_cont(ctx0, cur);
+
+    ggml_tensor * u = ggml_mean(ctx0, cur);                 // [1,H,W,B]
+    u = ggml_repeat(ctx0, u, cur);                          // [C,H,W,B]
```
ngxson: some `ggml_repeat` seem to be redundant because ggml supports broadcasting internally

common/jinja/runtime.cpp
ngxson, line 1: it's recommended to move jinja-related changes to a dedicated PR

kwajiehao: Thanks @ngxson for the quick reply

Will close this PR, create a separate one for the Jinja changes, and address your comments

ngxson: I would suggest reusing this PR and simply remove jinja-related changes from it.

Creating a new PR will discard all my comments above

kwajiehao: Thanks! Jinja PR over here  https://github.com/ggml-org/llama.cpp/pull/21623

tools/mtmd/models/yasa2.cpp
kwajiehao: Addressed in https://github.com/ggml-org/llama.cpp/pull/21616/commits/91956bf3617dd278551de08b52942209fe2e04ef

tools/mtmd/models/yasa2.cpp
kwajiehao: Addressed in https://github.com/ggml-org/llama.cpp/pull/21616/commits/91956bf3617dd278551de08b52942209fe2e04ef

tools/mtmd/models/yasa2.cpp
kwajiehao: Thanks for the pointer, I've addressed it in https://github.com/ggml-org/llama.cpp/pull/21616/commits/91956bf3617dd278551de08b52942209fe2e04ef and re-ran my benchmarks

tools/mtmd/clip.cpp
kwajiehao: Addressed in https://github.com/ggml-org/llama.cpp/pull/21616/commits/d2a30b5db1842f4429d60baca13fac9efd1a4e95

tools/mtmd/models/yasa2.cpp
kwajiehao: Addressed in https://github.com/ggml-org/llama.cpp/pull/21616/commits/d2a30b5db1842f4429d60baca13fac9efd1a4e95

tools/mtmd/models/yasa2.cpp
kwajiehao: Thanks, addressed in https://github.com/ggml-org/llama.cpp/pull/21616/commits/b165badc49c995d626095491c762229eaf2e2fa0

Added some of them back after doing tests on CPU https://github.com/ggml-org/llama.cpp/pull/21616/commits/5e48fb343eee3a296fac190f2057f03f68f93416

tools/mtmd/models/yasa2.cpp
kwajiehao: Addressed in https://github.com/ggml-org/llama.cpp/pull/21616/commits/9f8eaf4d05e64a1170bf1dc9dc552355fae37f3e

common/chat-reka.cpp
ngxson, line 1: a general note that we don't add a dedicated file for each new model, please follow the existing pattern from other models

ngxson (approved):
the mtmd part looks OK, but chat template part requires additional review from @pwilkin @aldehir

pwilkin:
I'm puzzled - the autoparser seems to detect the template just fine, why would we need a specialized one for this model?

```
--- Reasoning & Content Structure ---
reasoning_mode: TAG_BASED
reasoning_start: '<think>
'
reasoning_end: '</think>'
content_mode: PLAIN
content_start: ''
content_end: ''

--- Tool Call Structure ---
tool_mode: JSON_NATIVE
supports_tools: true
supports_parallel_calls: true
tool_section_start: '<tool_call>
'
tool_section_end: '</tool_call>'
per_call_start: ''
per_call_end: ''
func_name_prefix: ''
func_name_suffix: ''
func_close: ''
call_id_prefix: ''
call_id_suffix: ''
call_id_pos: 'NONE'
args_start: ''
args_end: ''
arg_name_prefix: ''
arg_name_suffix: ''
arg_value_prefix: ''
arg_value_suffix: ''
name_field: 'name'
args_field: 'arguments'
id_field: ''
gen_id_field: ''
parameter_order: 'name, arguments'
```

common/chat-reka.cpp
kwajiehao: Sorry about that, fixed in https://github.com/ggml-org/llama.cpp/pull/21616/commits/f1c241800edbfcfaebd187970f0e51dc53729847!

kwajiehao:
> I'm puzzled - the autoparser seems to detect the template just fine, why would we need a specialized one for this model?
> 
> ```
> --- Reasoning & Content Structure ---
> reasoning_mode: TAG_BASED
> reasoning_start: '<think>
> '
> reasoning_end: '</think>'
> content_mode: PLAIN
> content_start: ''
> content_end: ''
> 
> --- Tool Call Structure ---
> tool_mode: JSON_NATIVE
> supports_tools: true
> supports_parallel_calls: true
> tool_section_start: '<tool_call>
> '
> tool_section_end: '</tool_call>'
> per_call_start: ''
> per_call_end: ''
> func_name_prefix: ''
> func_name_suffix: ''
> func_close: ''
> call_id_prefix: ''
> call_id_suffix: ''
> call_id_pos: 'NONE'
> args_start: ''
> args_end: ''
> arg_name_prefix: ''
> arg_name_suffix: ''
> arg_value_prefix: ''
> arg_value_suffix: ''
> name_field: 'name'
> args_field: 'arguments'
> id_field: ''
> gen_id_field: ''
> parameter_order: 'name, arguments'
> ```

@pwilkin when I tested with the autoparser I found that it failed for parallel tool calls. This is because of a difference in the definition of parallel tool calls. The autoparser defines parallel tool calls = multiple JSON objects comma-separated inside a single section. But for the Reka chat template, each tool call gets its own `<tool_call>...</tool_call>` wrapper. Specifically, standard_json_tools (chat-peg-parser.cpp:855-866) puts multiple comma-separated JSONs inside one `<tool_call>` block, but the Reka template (Reka-Edge.jinja:119-136) loops and emits a separate `<tool_call>...</tool_call>` per call.

pwilkin:
@kwajiehao ah, okay, in that case, don't make a separate parser, instead, add a workaround at the top of chat-diff-analyzer.cpp with:

```c++
analysis.tools.format.section_start = "";
analysis.tools.format.section_end = "";
analysis.tools.format.per_call_start = "<tool_call>";
analysis.tools.format.per_call_end = "</tool_call>";
```

That should be enough.

kwajiehao:
@pwilkin thanks for the suggestion! I tried this and found that the autoparser has a few gaps for Reka's chat template format:
1.   The Reka chat template uses JSON for tool calls, so it uses `build_tool_parser_json_native`. However, `build_tool_parser_json_native` ignores per_call_start/end and only calls `standard_json_tools` with only section_start/end i.e. JSON native + per-call wrapping isn't supported
2. In lazy mode (common/peg-parser.cpp:1796-1810), the grammar is built from the single tool-call trigger rule even if the PEG parser understands the repetition, so it only permits one call even when parallel is requested

Let me know if I'm missing anything in my analysis. Happy to either keep the custom parser for Reka, or to extend the behavior of the autoparser if needed

pwilkin:
Yeah, I thought about it right after I suggested the approach, sorry :) I'll incorporate the per-call approach for JSON-native in autoparser.

kwajiehao:
@pwilkin just to confirm, should i wait for you to ship the per-call approach before we can merge this in? We would also need the peg parser to be fixed for lazy mode

tools/mtmd/clip.cpp
```diff
@@ -1389,6 +1393,14 @@ struct clip_model_loader {
                         hparams.set_limit_image_tokens(1, 62500);
                         hparams.set_warmup_n_tokens(16*16); // avoid OOM on warmup
                     } break;
+                case PROJECTOR_TYPE_YASA2:
+                    {
+                        hparams.ffn_op = FFN_GELU_ERF;
+                        log_ffn_op = "gelu_erf";
+                        hparams.image_resize_algo = RESIZE_ALGO_BICUBIC;
+                        hparams.image_resize_pad = false;
+                        hparams.set_warmup_n_tokens(64);
```
ngxson:
```suggestion
```

since the model only support fixed size image (not dynamic size like qwen-vl), the number of warmup tokens will be calculated automatically

ngxson: btw, you're setting `image_resize_pad = false`, is that expected?

that means non-square image will be center-cropped, resulting in loss of information

pwilkin:
@kwajiehao https://github.com/ggml-org/llama.cpp/pull/21892 should automatically add support for Reka-Edge (added your template as an example as well as a small battery tests, no needs for workarounds anymore, should be correctly detected out-of-the-box).

tools/mtmd/clip.cpp
kwajiehao: Thanks @ngxson, i removed the warmup tokens call in https://github.com/ggml-org/llama.cpp/pull/21616/commits/b5aa85d4bc9845df8cccdb282c6d2fabcda0555e

`image_resize_pad = false` is intentional. From from the benchmarks I ran, our model performs better with `image_resize_pad = false`. We use the `resize_bicubic` algorithm and looking at the function it distorts the aspect ratio of the image instead of cropping it (I verified this by running the function on a test image). My guess is that our dataset does not contain many images with padding so it performs worse on them, and performs better on stretched images

pwilkin:
@kwajiehao aight, autoparser changes are merged now, you can rebase on master and check.

tools/mtmd/clip.cpp
ngxson: Hmm ok, in this case, I think it's better to add a comment so that maintainers don't accidentally remove this behavior in the future:

// stretching image to fit fixed square size if expected for this model

kwajiehao: Makes sense, added in https://github.com/ggml-org/llama.cpp/pull/21616/commits/e6ec92e607249d905726b16211c9001456d6410b

kwajiehao:
Thanks @pwilkin - i've rebased on master, removed the custom parser code in https://github.com/ggml-org/llama.cpp/pull/21616/commits/04743ad42a897caba846b7a984bb000fd8131007, and verified that everything looks good!

kwajiehao:
@ngxson @pwilkin thanks for the detailed comments and review. Do you have any more feedback for this PR? Or would we be good to merge this in

pwilkin (approved):

common/chat.cpp
```diff
@@ -2396,4 +2396,3 @@ std::map<std::string, bool> common_chat_templates_get_caps(const common_chat_tem
     GGML_ASSERT(chat_templates->template_default != nullptr);
     return chat_templates->template_default->caps.to_map();
 }
-
```
pwilkin, line 2399: Revert chat.cpp edits, this removes the trailing newline.

pwilkin (approved):

ngxson (approved):
can merge after @pwilkin 's approval

common/chat.cpp
kwajiehao: Fixed by checking out `common/chat.cpp` from `upstream/master`

pwilkin (approved):

pwilkin (approved):

ngxson (approved):

ngxson:
I'm re-runing one of the server CI, will merge when it passes

ngxson:
Hmm seems like server's CI didn't pass because it misses a fix from upstream, but it's unrelated to the current PR. Should be ok for merge.

sebagallo:
made my own quants but it looks like something is broken in the parsing:
image
The mmproj works, but I noticed all responses being wrapped inside the `<answer>` or `<result>` tags, and tool calls don't seem to work.

For quanting, I used the script at the top to produce BF16 quants, then ran llama-quantize following the suggestion to keep the last 8 blocks at Q8_0. The Resulting Q8_0 is producing these results.

kwajiehao:
Hi @sebagallo, first of all thank you for your interest in Reka Edge 2603. Unfortunately this model does not support reasoning so you will need to run `llama-server` with `--reasoning off`

Beyond that, I found an error with the gguf conversion script (due to a reliance on an older version of `convert_hf_to_gguf`) and I've [replaced it](https://huggingface.co/RekaAI/reka-edge-2603/blob/main/convert_reka_vlm_to_gguf.py). With `--reasoning off` my tool call tests pass and I can't reproduce any tool call errors. 

Can you give it a shot with `--reasoning off`?

sebagallo:
> Hi @sebagallo, first of all thank you for your interest in Reka Edge 2603. Unfortunately this model does not support reasoning so you will need to run `llama-server` with `--reasoning off`
> 
> Beyond that, I found an error with the gguf conversion script (due to a reliance on an older version of `convert_hf_to_gguf`) and I've [replaced it](https://huggingface.co/RekaAI/reka-edge-2603/blob/main/convert_reka_vlm_to_gguf.py). With `--reasoning off` my tool call tests pass and I can't reproduce any tool call errors.
> 
> Can you give it a shot with `--reasoning off`?

reasoning off seems to fix the issues! my impression was that thinking was supported as I saw in the chat template `enable_thinking` and the `<think>` tag.

Tried tool calling and the mmproj again and everything works!.
Regarding the conversion, at first I had problems with the tokenizer, but momentarily downgrading to transformers 4.57.3 allowed to produce the GUFFs. I didn't try your latest fix, but the files I produced with the mentioned transormers version right now *work*.

Thanks for this new model!

PR #21527, opened 2026-04-06
[SYCL] Add Q8_0 reorder optimization for Intel GPUs (~3x token generation speedup)

PMZFX:
## Summary
- Extends the existing SYCL reorder optimization (currently Q4_0/Q4_K/Q6_K) to support Q8_0
- Q8_0 token generation on Intel Arc Pro B70 (Xe2/Battlemage): 4.88 t/s → 15.24 t/s (3.1x faster)
- Memory bandwidth utilization improves from 21% to 66% of theoretical maximum
- Q8_0 is now faster than Q6_K (15.24 vs 13.83 t/s) while providing higher quality

## Motivation

Q8_0 quantization was ~4-5x slower than Q4_K_M on Intel Xe2 GPUs despite only having 1.7x more data per weight. The root cause was that Q8_0's 34-byte blocks (not power-of-2) cause poor cache line utilization and non-coalesced memory access patterns. The reorder optimization — which separates scale factors from weight data for coalesced reads — was already implemented for Q4_0, Q4_K, and Q6_K but never added for Q8_0.

Related issue: #21517

## Changes

- `quants.hpp`: Add `block_q_t<GGML_TYPE_Q8_0>` reorder block traits
- `dequantize.hpp`: Add `dequantize_q8_0_reorder()` for separated d/qs layout
- `dmmv.cpp`: Add Q8_0 DMMV reorder kernel + dispatch
- `vecdotq.hpp`: Add `reorder_vec_dot_q_sycl<GGML_TYPE_Q8_0>`
- `mmvq.cpp`: Add Q8_0 MMVQ reorder kernel launcher + dispatch
- `ggml-sycl.cpp`: Add `reorder_qw_q8_0()`, update dispatch + extra allocation

## Benchmark Results

Qwen3.5-27B (26.9B params, dense), Intel Arc Pro B70, single GPU:

- Q8_0 before: pp512 295 t/s, tg128 4.88 t/s (21% BW)
- Q8_0 after: pp512 698 t/s, tg128 15.24 t/s (66% BW)
- Q4_K_M: pp512 729 t/s, tg128 20.12 t/s (52% BW)
- Q6_K: tg128 13.83 t/s (48% BW, no reorder)

Qwen2.5-1.5B (1.78B params), Intel Arc Pro B70, single GPU:

- Q8_0 before: tg128 ~64 t/s (19% BW)
- Q8_0 after: tg128 180 t/s (52% BW)
- Q4_K_M: tg128 208 t/s (36% BW)

## Testing
[...]

PR #21618, opened 2026-04-08
SYCL: fix reorder crash when device memory is full

PMZFX:
## Summary

- The reorder optimization (AoS to SoA weight layout, added in #21527) allocates a temp buffer the size of the entire weight tensor. On GPUs where the model fills most of VRAM, this allocation fails and crashes with a NULL pointer memcpy.
- Fix adds a host memory fallback: if device allocation fails, the temp buffer is allocated in system RAM instead. The reorder kernel reads from host memory over PCIe, which is slower for the one-time transform but preserves the optimization for all subsequent tokens.
- Also fixes a bug where `opt_for_reorder()` unconditionally set the reorder flag even when the reorder was skipped. This caused the DMMV/MMVQ kernels to interpret AoS data as SoA, producing garbage output.

## Details

Reported in #20478. Root cause is that `sycl_ext_malloc_device` returns NULL when VRAM is too full for the tensor-sized temp buffer. The crash happens on the first token generation (DMMV/MMVQ path), not during prompt processing (GEMM path, which doesn't reorder).

The fix affects all four reorder functions (Q4_0, Q8_0, Q4_K, Q6_K). Two new helpers (`sycl_ext_malloc_with_fallback`, `sycl_ext_free_fallback`) keep the per-function changes minimal.

## Test plan

- [x] Q8_0 9B, single GPU, device reorder path (37.7 t/s tg)
- [x] Q8_0 9B, forced host fallback path (21.3 t/s tg, correct output)
- [x] Q4_K_M 27B, single GPU (16.0 t/s tg)
- [x] Q8_0 9B, llama-bench pp512/tg32 (2554/48 t/s, GEMM path unaffected)
- [x] Q8_0 9B with GGML_SYCL_DISABLE_OPT=1 (16.9 t/s tg, reorder skipped correctly)

Tested on Intel Arc Pro B70 (32GB). AI-assisted development (Claude). Code reviewed and tested on my hardware.

### DIFF
--- a/ggml/src/ggml-sycl/ggml-sycl.cpp
+++ b/ggml/src/ggml-sycl/ggml-sycl.cpp
@@ -3333,9 +3333,37 @@ static inline void sycl_ext_free(dpct::queue_ptr stream, void * ptr) {
     sycl::free(ptr, *stream);
 }
 
-static void reorder_qw_q4_0(uint8_t * data_device, const int ncols, const int nrows, size_t size, size_t offset,
+// Try device allocation first; if VRAM is full, fall back to host memory so the
+// reorder kernel can still run (reading over PCIe instead of device-local).
+static inline void * sycl_ext_malloc_with_fallback(dpct::queue_ptr stream, size_t size, bool & host_fallback) {
+    host_fallback = false;
+    void * ptr = sycl_ext_malloc_device(stream, size);
+    if (!ptr) {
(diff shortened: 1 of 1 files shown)


PMZFX:
Folded into #21638 along with the GEMM dequantize fix. One PR for both Q8_0 reorder issues.

PR #21632, opened 2026-04-08
Enable ccache on riscv64

luhenry:
## Overview

This is made possible with https://github.com/ggml-org/ccache-action/pull/1

## Additional information

# Requirements

- AI usage disclosure:

### DIFF
--- a/.github/workflows/build-riscv.yml
+++ b/.github/workflows/build-riscv.yml
@@ -47,21 +47,9 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
-          sudo apt-get update
-
-          # Install necessary packages
-          sudo apt-get install -y libatomic1 libtsan2 gcc-14 g++-14 cmake build-essential wget git-lfs
-
           # Set gcc-14 and g++-14 as the default compilers
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
           sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 100
-
-          if ! which rustc; then
-            # Install Rust stable version
-            sudo apt-get install -y rustup
-            rustup install stable
-            rustup default stable
-          fi
 
           git lfs install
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-riscv.yml
```diff
@@ -74,12 +74,13 @@ jobs:
...
-      # - name: ccache
-      #   uses: ggml-org/ccache-action@v1.2.21
-      #   with:
-      #     key: ubuntu-riscv64-native-sanitizer-${{ matrix.sanytizer }}-${{ matrix.build_type }}
-      #     save: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+      - name: ccache
+        # uses: ggml-org/ccache-action@v1.2.21
+        uses: luhenry/ccache-action@main
+        with:
+          key: ubuntu-riscv64-native-sanitizer-${{ matrix.sanytizer }}-${{ matrix.build_type }}
```
CISC:
```suggestion
          key: ubuntu-riscv64-native-sanitizer-${{ matrix.sanitizer }}-${{ matrix.build_type }}
```

CISC (commented):
Unsure how we'll proceed, I think maybe wait for upstream to not pull in additional binaries...

.github/workflows/build-riscv.yml
luhenry: So turns out, because ccache is already available on the machine, it won’t pull the binary. Adding “riscv64” along the code is still necessary because it first check if it’s “supported” (if there is an entry in the code), then it checks if it’s already installed…

luhenry: Once it’s merged in ggml-org/ccache-actions, I’ll of course switch this PR to use it rather than luhenry/ccache-action.

luhenry: I'm testing without the change to pull from `riseproject-dev/ccache` releases, let's see if it works. I would rather have that so that we don't have to maintain ccache releases until upstream is merged.

.github/workflows/build-riscv.yml
```diff
@@ -47,22 +47,10 @@ jobs:
...
           # Set gcc-14 and g++-14 as the default compilers
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
           sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 100
 
-          if ! which rustc; then
-            # Install Rust stable version
-            sudo apt-get install -y rustup
-            rustup install stable
-            rustup default stable
-          fi
```
CISC, line 64: This should be removed from `build.yml` as well if it's no longer necessary.

CISC (approved):

ggerganov:
https://github.com/ggml-org/ccache-action/tree/riscv64 is merged now.

CISC:
@ggerganov I think this PR is fine to merge as-is, it targets a commit based off 1.2.21 instead of current unreleased main branch.

luhenry:
Let me bump to https://github.com/hendrikmuhs/ccache-action/commit/480829ac41148baea5c49d6e79cb14455c210d84 for `riscv64`, unless you want to add a tag?

CISC:
> Let me bump to [hendrikmuhs/ccache-action@480829a](https://github.com/hendrikmuhs/ccache-action/commit/480829ac41148baea5c49d6e79cb14455c210d84) for `riscv64`, unless you want to add a tag?

I'd rather not target intermediate changes, best keep it as close as possible to all the other CIs.

luhenry:
I’ll work on unifying Ubuntu-cpu-riscv64 into the matrix of other Ubuntu-cpu jobs next, making sure there are all the same dependencies as other platforms out of the box

Issue #20938, opened 2026-03-24
Misc. bug: OpenVINO backend performance abysmally bad


paoletto:
### Name and Version

using ghcr.io/ggml-org/llama.cpp:server-openvino	2026-03-23 07:02:20 
server-openvino: build 8477 (ec2b787eb)

inference with Qwen3-4B Q4_0  is at about 0.1t/s

performance with openvino own images/model format around 2-3 t/s, possibly more with such small model.

### Operating systems

Linux

### Which llama.cpp modules do you know to be affected?

llama-server

### Command line

```shell
docker run -it --rm --name llama.cpp --network=host --device /dev/dri --device /dev/accel -e GGML_OPENVINO_DEVICE=NPU -v /usr/lib/x86_64-linux-gnu/libze_loader.so.1:/usr/lib/x86_64-linux-gnu/libze_loader.so.1:ro -v /usr/lib/x86_64-linux-gnu/libze_intel_npu.so.1:/usr/lib/x86_64-linux-gnu/libze_intel_npu.so.1:ro -v /usr/lib/x86_64-linux-gnu/libnpu_driver_compiler.so:/usr/lib/x86_64-linux-gnu/libnpu_driver_compiler.so:ro -e GGML_OPENVINO_CACHE_DIR=/models/.openvino_cache -v /models/llama.cpp:/models ghcr.io/ggml-org/llama.cpp:server-openvino --host 0.0.0.0 --port 8080 -m /models/Qwen3-4B-Instruct-2507.Q4_0.gguf -ngl 99 -np 1 -c 8192 -ctk f16 -ctv f16 -ctkd f16 -ctvd f16 --flash-attn on --jinja --reasoning-format deepseek
```

### Problem description & steps to reproduce

open localhost:8080, say hi

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

thedanhoffman:
I don't have enough RAM on my laptop to fully reproduce this, but in general the NPU works best with a very short context (for now at least). Try explicitly setting something like `--ctx-size 256` and let me know if that solves things for you

EDIT: `--ctx-size 16 --batch-size 16 --ubatch-size 16` and using a recent from-source build of OpenVINO gave me 5.1t/s prompt processing and 5.5t/s sequential (this is with zram/swapping to disk on Linux, so that might be lower than what other people may see)

PR #21637, opened 2026-04-08
fix(openvino): define PartialShape bounds for tensors

thedanhoffman:
## Overview

Bound `PartialShape` dimensions to be within some reasonable range. This is sufficient to fix an OpenCL allocation issue observed by running https://github.com/ggml-org/llama.cpp/issues/20938 with a GPU.

Even in the case with no functional problems, these shape bounds help inform the OpenVINO graph compiler.

## Additional information

Discussed offline with @cavusmustafa. The underlying issue will probably be root-caused to an OpenVINO upstream issue, but if "best practices" on the API user end solves this issue, then I don't see much of a problem with enabling this to work around the issue.

```
export GGML_OPENVINO_STATEFUL_EXECUTION=1

/home/dhoffman/llama_openvino/llama.cpp/build/bin/llama-cli --verbose -m /home/dhoffman/models/Qwen3-4B-Instruct-2507-Q4_0.gguf -ngl 99 -np 1 -c 8192 -ctk f16 -ctv f16 -ctkd f16 -ctvd f16 --flash-attn on --jinja --reasoning-format deepseek --simple-io --single-turn -n 16 -p hello --no-warmup -c 16
```

# Requirements

- AI usage disclosure: All code written was done by myself personally, but I did throw an AI at the debug process. Again, I work at Intel and have discussed this with the owners in-person, so the changes are not unreasonable

### DIFF
--- a/ggml/src/ggml-openvino/ggml-decoder.cpp
+++ b/ggml/src/ggml-openvino/ggml-decoder.cpp
@@ -332,28 +332,49 @@ void GgmlOvDecoder::validate_cgraph() const {
 
 ov::PartialShape GgmlOvDecoder::get_graph_input_shape(const ggml_tensor * op, const ggml_tensor * input) const {
     if (m_naive) {
-        return input!= nullptr ? ov::PartialShape{get_shape(input)} : ov::PartialShape{get_shape(op)};
+        return input != nullptr ? ov::PartialShape{get_shape(input)} : ov::PartialShape{get_shape(op)};
     }
     auto name = std::string(input->name);
     ov::PartialShape input_shape;
 
+    // ggml_tensor gives us exact measurements in all cases, so none of those should be -1 (as that's an
+    // OpenVINO native convention). All of those are passed thorugh directly.
+    //
+    // Cases where a tensor dimension size varies are handled case-by-case below. We provide a PartialShape to
+    // communicate the worst-case scenario: a PartialShape has a lower and upper bound on the dimension,
+    // used to inform OpenVINO optimizations. An issue was observed with OpenCL remote buffers not allocating
+    // unless such a range was provided (considerations with remote memory). Although that's not the responsibility
+    // of llama.cpp to solve, providing dimension bounds is useful nonetheless.
+
+    const auto prefill_upper = m_is_prefill ? m_prefill_chunk_size : 1;
+    const auto dim_span_ctx = ov::Dimension(1, m_model_params.ctx);
+
     if (is_inp_tok(input, op) || is_inp_pos(input, op)) {
         // tokens or positions
-        int len = m_is_static ? (m_is_prefill ? m_prefill_chunk_size : 1) : -1;
-        input_shape = ov::PartialShape{1, 1, 1, len};
+        if (m_is_static) {
+            input_shape = ov::PartialShape{1, 1, 1, prefill_upper};
+        } else {
+            // NOTE: AFAICT PartialShape with min_dim == max_dim is not valid, Shape must be used.
+            // Updating callsites to return some Shape optional to allow 1,1,1,1
+            // might materially improve things
+            input_shape = ov::PartialShape{1, 1, 1, ov::Dimension(1, m_prefill_chunk_size)};
+        }
 
     } else if (is_output_idx(input, op)) {
         // output index
-        input_shape = ov::PartialShape{1, 1, 1, m_is_static ? m_compute_params.output_len : -1};
-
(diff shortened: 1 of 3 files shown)


ggml/src/ggml-openvino/ggml-decoder.cpp
```diff
@@ -332,51 +332,76 @@ void GgmlOvDecoder::validate_cgraph() const {
...
+        }
 
     } else if (is_output_idx(input, op)) {
         // output index
-        input_shape = ov::PartialShape{1, 1, 1, m_is_static ? m_compute_params.output_len : -1};
-
+        if (m_is_static) {
+            input_shape = ov::PartialShape{1, 1, 1, m_compute_params.output_len};
+        } else {
+            input_shape = ov::PartialShape{1, 1, 1, ov::Dimension(1, m_compute_params.output_len)};
```
cavusmustafa, line 368: The changes look good to me but why do we need to remove one liners for this?

For example, can we still use something like this?

```
input_shape = ov::PartialShape{1, 1, 1, m_is_static ?
                        m_compute_params.output_len : ov::Dimension(1, m_compute_params.output_len)};
```

ggml/src/ggml-openvino/ggml-decoder.cpp
```diff
@@ -332,51 +332,76 @@ void GgmlOvDecoder::validate_cgraph() const {
...
+            input_shape = ov::PartialShape{dim_span_ctx, 1, dim_span_ctx, dim_span_ctx};
         }
 
     } else if (is_kvcache(input, op)) {
         // kvcache
         input_shape = ov::PartialShape{get_shape(input)};
         if (!m_is_static) {
             // do not fix ctx size to make llama-bench work across test params
-            input_shape[2] = -1;
+            input_shape[2] = dim_span_ctx;
```
cavusmustafa, line 385: llama-bench failed on my side for larger context sizes and for all stateful executions.
tested with llama3.2 1B q4_0

stateless: -d 512,1024 fails
stateful: all ctx sizes fail

cavusmustafa (changes requested):

ggml/src/ggml-openvino/ggml-decoder.cpp
thedanhoffman: yea i can change it back to how it was

ggml/src/ggml-openvino/ggml-decoder.cpp
thedanhoffman: investigating this now

thedanhoffman: cache wasn't being properly invalidated and there's no easy way (AFAICT) to get the max possible batch size

thedanhoffman:
cache wasn't being invalidated for cases where the batch size changes. adjusted `ComputeParams` to accept the max observed prefill size and invalidate the cache in cases where the maximum increases

I'm hoping for most "real" workloads, there will be some sort of system prompt which exceeds whatever batch size is configured, so there shouldn't be a performance hit in this case (except maybe post-warmup, but that's a one-time cost, so not a big deal)

thedanhoffman:
also went ahead and moved this to draft since how we go about invalidating the cache might be a broader architectural question

ggml/src/ggml-openvino/ggml-decoder.cpp
thedanhoffman: actually it looks like your problem also exists without these changes(?)

thedanhoffman:
```
set_n_threads: n_threads = 6, n_threads_batch = 6
GGML OpenVINO backend std::exception: might be outdated RESHAPE case
graph_compute: ggml_backend_sched_graph_compute_async failed with error -1
process_ubatch: failed to compute graph, compute status: -1
decode: removing memory module entries for seq_id = 0, pos = [0, +inf)
llama_decode: failed to decode, ret = -3
test_prompt: failed to decode prompt batch, res = -3
main: error: failed to run prompt warmup
```

is what I see when I run `./bin/llama-bench --batch-size 128 --ubatch-size 128 -m [the llama 3.2 1b instruct q4_0 GGUF] --verbose`

I'll try and fix the CPU native cases on my end and file some PRs for that if I can. For now I'm assuming these failures are unrelated and we can sync on Monday to understand the details here.

thedanhoffman:
Closing to PR against Ravi's branch directly

PR #21654, opened 2026-04-09
sycl : add flash-attn support for head size 512

qnixsynapse:
## Overview

This patch extends the SYCL Flash Attention implementation to support head sizes (DKQ/DV) of 512(Gemma 4).

Changes:
- Added DKQ/DV 512 cases to both tile and vector Flash Attention kernels.
- Updated kernel selection logic to allow vector kernels for head sizes up to 512 (previously 256).
- Added necessary template instances for the new 512 head size across various quantization types.

Also refactored the reorder buffer init in `ggml_backend_sycl_buffer_init_tensor` to use a switch case instead of if conditions. We may need to move it to a separate function as new types support is added.

Also, removed tile config functions for AMD. Those were unused.

test-backend-ops is passing with this change.

# Requirements

- AI usage disclosure: Used it for understanding the existing flash attention code in SYCL and for getting the recommended tile and vec configs for Arc GPU's(alchemist) slm size.


arthw (approved):
It's good job!
Welcome back to maintain SYCL backend! :)

qnixsynapse:
@arthw Thank you!

PR #21658, opened 2026-04-09
Fix/lfm2 vl tiling v2

alay2shah:
The current LFM2-VL tiling threshold uses a per-dim check (width > 1024 || height > 1024) which doesn't match the HF implementation. This causes slightly different tiling behavior in the 721-1024px range causing early downscaling or upscaling. 

- Fixes the threshold to use an area-based check matching HF: rounded_area > image_max_pixels * tolerance
- Adds --image-max-pixels-tolerance flag to control tiling thresh at runtime
- Adds --image-resize-bicubic flag to opt into bicubic interpolation (training uses bicubic, llama.cpp defaults to bilinear)

Tested with LFM2-VL-450M on Jetson AGX Orin. Tiling now matches HF transformers 5.2.0 exactly (Across 6 different resolution tests). Existing test-mtmd-c-api passes.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2197,6 +2197,20 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.image_max_tokens = value;
         }
     ).set_examples(mmproj_examples).set_env("LLAMA_ARG_IMAGE_MAX_TOKENS"));
+    add_opt(common_arg(
+        {"--image-max-pixels-tolerance"}, "N",
+        "tolerance multiplier for tiling threshold in vision models (default: read from model, typically 2.0)",
+        [](common_params & params, const std::string & value) {
+            params.image_max_pixels_tolerance = std::stof(value);
+        }
+    ).set_examples(mmproj_examples).set_env("LLAMA_ARG_IMAGE_MAX_PIXELS_TOLERANCE"));
+    add_opt(common_arg(
+        {"--image-resize-bicubic"},
+        "use bicubic interpolation for image resizing (default: bilinear). bicubic is slower but preserves fine detail",
+        [](common_params & params) {
+            params.image_resize_bicubic = true;
+        }
+    ).set_examples(mmproj_examples).set_env("LLAMA_ARG_IMAGE_RESIZE_BICUBIC"));
     if (llama_supports_rpc()) {
         add_opt(common_arg(
             {"--rpc"}, "SERVERS",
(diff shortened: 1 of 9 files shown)


common/arg.cpp
```diff
@@ -2197,6 +2197,20 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
...
+            params.image_max_pixels_tolerance = std::stof(value);
+        }
+    ).set_examples(mmproj_examples).set_env("LLAMA_ARG_IMAGE_MAX_PIXELS_TOLERANCE"));
+    add_opt(common_arg(
+        {"--image-resize-bicubic"},
+        "use bicubic interpolation for image resizing (default: bilinear). bicubic is slower but preserves fine detail",
+        [](common_params & params) {
+            params.image_resize_bicubic = true;
+        }
+    ).set_examples(mmproj_examples).set_env("LLAMA_ARG_IMAGE_RESIZE_BICUBIC"));
```
ngxson, line 2213: unfortunately I don't think we can accept these params. these are specific to model upon training.

you should implement this logic in your application code instead. for example, if image is already resized to match the grid size, they won't be resized again inside llama.cpp code

tools/mtmd/mtmd-image.cpp
```diff
@@ -908,8 +908,16 @@ mtmd_image_preprocessor_llava_uhd::slice_instructions mtmd_image_preprocessor_lf
...
+    // tile if pixel area exceeds max_pixels budget with tolerance
+    // matches HuggingFace transformers reference implementation
+    const float tolerance = hparams.custom_max_pixels_tolerance > 0
+        ? hparams.custom_max_pixels_tolerance
+        : max_pixels_tolerance;
+    const int max_pixels_budget = static_cast<int>(hparams.image_max_pixels * tolerance);
+    // round dimensions to align_size before area check (banker's rounding to match Python/HF)
+    const int h_bar = std::max(align_size, static_cast<int>(std::nearbyint(static_cast<float>(original_size.height) / align_size) * align_size));
+    const int w_bar = std::max(align_size, static_cast<int>(std::nearbyint(static_cast<float>(original_size.width)  / align_size) * align_size));
+    const bool needs_tiling = h_bar * w_bar > max_pixels_budget;
```
tdakhran, line 920: @alay2shah these are the only changes needed, constants can be hardcoded with the comment that they are LFM2.5-VL specific.

tdakhran: lfm2 image preprocessor already has some [constants ](https://github.com/ggml-org/llama.cpp/blob/master/tools/mtmd/mtmd-image.h#L115-L118) hardcoded, suggest extending it

Issue #19639, opened 2026-02-15
Misc. bug: CUDA memory leak with MTMD requests loop with Gemma3


Cyrille37:
Hi.

Thanks a lot for sharing this so great ggml work 💌.

### Name and Version

./build/bin/llama-cli --version

ggml_cuda_init: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 3060, compute capability 8.6, VMM: yes
version: 7971 (5fa1c190d)
built with GNU 13.3.0 for Linux x86_64

### Operating systems

Linux NS5x-NS7xAU.localdomain 6.17.0-14-generic #14~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jan 15 15:52:10 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### Which llama.cpp modules do you know to be affected?

llama-server

```
$ git log -1
commit 5fa1c190d9fc86c02698b730a2cb933195e19d96 (HEAD -> master, origin/master, origin/HEAD)
Author: Adrien Gallouët <angt@huggingface.co>
Date:   Sun Feb 8 09:06:45 2026 +0100
    rpc : update from common.cpp (#19400)    
    Signed-off-by: Adrien Gallouët <angt@huggingface.co>

$ export CUDA_VERSION=12.9 && cmake -B build -DGGML_CUDA=ON \
 -DCMAKE_CUDA_ARCHITECTURES="86;120" \
 -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
 -DCMAKE_CUDA_COMPILER=/usr/local/cuda-${CUDA_VERSION}/bin/nvcc \
 -DCMAKE_INSTALL_RPATH="/usr/local/cuda-${CUDA_VERSION}/lib64;\$ORIGIN"

$ cmake --build build --config Release -j 10
```

### Command line

```shell
./llama.cpp/build/bin/llama-server --port 8012 \
 -m ~/Data/AI_ModelsVision/gemma-3-4b-it-Q4_K_M.gguf \
 --mmproj ~/Data/AI_ModelsVision/mmproj-model-f16.gguf \
 -c 0 --parallel 1
```

`--parallel 1` does not change the count of requests before fail.

### Problem description & steps to reproduce

The use case is object detection (MTMD) in small jpeg images (1.7Ko to 19ko) and llama-server fails with `CUDA out of memory` after ~700 requests.

[...]

PR #21673, opened 2026-04-09
Bug-Fix sets an upper VRAM limit for cached ggml_cuda graphs to prevent VRAM memory leaks

kmorennv:
## Overview
Bug fix for: https://github.com/ggml-org/llama.cpp/issues/19639 The fix in cuda-backend prevents the unbound growth of cuda graphs for Gemma3 models.

## Additional information

The fix is local in the cuda-backend. It adds new data-type that manages the cached cuda graphs. With this PR the bug is fixed, but there is still a code refactoring necessary. 
- For instance ggml_cuda_graph is open for misuse of CudaGraph_t. 
- There is also the problem with current design of the graph-cache, many ggml_cuda_graph insatances are non initialized (contains cuda-graph == nullptr) but still consume the host-memory.

# Requirements

- AI usage disclosure: No


kmorennv:
1. This was a general problem that has been fixed (and gemma3n was only affected in the multimodal case).
2. The change in llama_context::set_causal_attn is not strictly speaking necessary -> can be isolated as a separate pull request.
3. I want to further improve memory safety with additional refactoring (since we currently have pollution/host-side memory because we're not cleaning up empty graphs, unused instances of ggml_cuda_graph).

am17an:
I'm not sure if this is the correct way to do this, seems a bit over-engineered with the LRU cache. I would much prefer a simple ring buffer, for example in #21611. A memory limit is fine I guess, but it would be preferable to have a dynamic upper bound to do this correctly. I don't have a good solution yet, still thinking about it.

JohannesGaessler:
I agree with @am17an that a simple solution would be much preferred.

kmorennv:
> I'm not sure if this is the correct way to do this, seems a bit over-engineered with the LRU cache. I would much prefer a simple ring buffer, for example in #21611. A memory limit is fine I guess, but it would be preferable to have a dynamic upper bound to do this correctly. I don't have a good solution yet, still thinking about it.

- ok, to fix only the upper-bound bug this might be too complex but there are other issues. 
- The solution proposed in #21611 fix only the unbound cache , but still the caching of many empty ggm_cuda_graph instances is not addressed -> ( but to be fair this another related issue ). 

> A memory limit is fine I guess, but it would be preferable to have a dynamic upper bound to do this correctly. I don't have a good solution yet, still thinking about it.
- Dynamic upper bound for instance is more complex than the static upper-bound/heuristic. Dynamic means dependent, on what ? It is clear, it is possible to make it dependent on the micro-arch. but than is a more complex check ....

Long story short , I am agree, to fix the bug only a few code changes are sufficient in the current code base.  To prevent other issues the separate PR is preferred ?

PR #21674, opened 2026-04-09
cli: add option to connect to server via http(s)

pwilkin:
## Overview

Adds an `--endpoint` option to connect to an existing server instance.

## Additional information

In many cases, people want to run a `llama-server` for various uses but also might want a quick test UI in cases where they cannot access the WebUI (i.e. pure console / terminal environments). Since `llama-cli` spawns a separate server instance, you cannot run both in VRAM-constrained environments, so having the option to run `llama-cli` with a `llama-server` endpoint seems desirable.

# Requirements

- AI usage disclosure: Yes, although GLM 5.1 generated code with a `goto` in it, so I had to double-check.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -635,14 +635,17 @@ static bool common_params_parse_ex(int argc, char ** argv, common_params_context
         throw std::invalid_argument("error: --prompt-cache-all not supported in interactive mode yet\n");
     }
 
+    // with --server-base, the model is provided by the external server
+    const bool has_remote_server = !params.server_base.empty();
+
     // handle model and download
-    if (!skip_model_download) {
+    if (!skip_model_download && !has_remote_server) {
         common_params_handle_models(params, ctx_arg.ex);
     }
 
-    // model is required (except for server)
+    // model is required (except for server, or when an external server provides it)
     // TODO @ngxson : maybe show a list of available models in CLI in this case
-    if (params.model.path.empty() && ctx_arg.ex != LLAMA_EXAMPLE_SERVER && !skip_model_download && !params.usage && !params.completion) {
+    if (params.model.path.empty() && ctx_arg.ex != LLAMA_EXAMPLE_SERVER && !skip_model_download && !params.usage && !params.completion && !has_remote_server) {
         throw std::invalid_argument("error: --model is required\n");
     }
 
@@ -1547,6 +1550,15 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.single_turn = true;
         }
     ).set_examples({LLAMA_EXAMPLE_COMPLETION, LLAMA_EXAMPLE_CLI}));
+    add_opt(common_arg(
+        {"--server-base"}, "URL",
+        "base url of an external llama-server instance to connect to, e.g. http://localhost:8080\n"
+        "when set, llama-cli does not spawn a local server and model args are ignored\n"
+        "(default: unset, spawn a local llama-server)",
+        [](common_params & params, const std::string & value) {
+            params.server_base = value;
+        }
+    ).set_examples({LLAMA_EXAMPLE_CLI}).set_env("LLAMA_ARG_SERVER_BASE"));
     add_opt(common_arg(
         {"-i", "--interactive"},
         string_format("run in interactive mode (default: %s)", params.interactive ? "true" : "false"),
--- a/common/common.h
+++ b/common/common.h
@@ -563,6 +563,8 @@ struct common_params {
 
     bool single_turn       = false; // single turn chat conversation
 
+    std::string server_base = ""; // base url of an external llama-server to connect to (CLI) // NOLINT
+
     ggml_type cache_type_k = GGML_TYPE_F16; // KV cache data type for the K
     ggml_type cache_type_v = GGML_TYPE_F16; // KV cache data type for the V
 
--- a/tools/cli/CMakeLists.txt
+++ b/tools/cli/CMakeLists.txt
@@ -2,11 +2,21 @@
 
 set(TARGET llama-cli-impl)
 
-add_library(${TARGET} cli.cpp)
+add_library(${TARGET}
+    cli.cpp
+    cli-client.cpp
+    cli-client.h
+    cli-context.cpp
+    cli-context.h
+    cli-server.cpp
+    cli-server.h
+    cli-view.cpp
+    cli-view.h
+)
 set_target_properties(${TARGET} PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
 
-target_include_directories(${TARGET} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ../server)
-target_link_libraries(${TARGET} PUBLIC server-context llama-common ${CMAKE_THREAD_LIBS_INIT})
+target_include_directories(${TARGET} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+target_link_libraries(${TARGET} PUBLIC llama-common cpp-httplib ${CMAKE_THREAD_LIBS_INIT})
 
 if(LLAMA_TOOLS_INSTALL)
     install(TARGETS ${TARGET} LIBRARY)
--- a/tools/cli/cli-client.cpp
+++ b/tools/cli/cli-client.cpp
@@ -1,0 +1,141 @@
+#include "cli-client.h"
+
+#include "http.h"
+
+#include <algorithm>
+#include <chrono>
+#include <thread>
+
+// generation can stall for a long time during prompt processing, so the
+// read timeout must be generous
+static constexpr time_t CLI_HTTP_READ_TIMEOUT_SEC = 3600;
+
+// upper bound for the accumulated response body kept for error reporting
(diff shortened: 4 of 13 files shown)


ngxson (changes requested):
IMO I'm not quite comfortable with this change. This adds too much for a feature that no one ever asked (via an issue)

If you really need to do this, just code your own CLI thing in higher-level languages like python or nodejs

tools/cli/cli-remote.h
```diff
@@ -0,0 +1,69 @@
...
+    // load a local text file, return its contents (empty string on failure)
+    virtual std::string load_text_file(const std::string & fname) = 0;
+
+    // load a local media file, return the OAI content part JSON for it
+    // returns empty JSON object on failure
+    virtual json load_media_file(const std::string & fname) = 0;
+
+    // cleanup
+    virtual void terminate() = 0;
+};
```
ngxson: I imagine this will add double the effort each time someone adds a new feature to the CLI

Not a wise choice for long-term maintenance. The CLI should either support native API or remote API, but not both

pwilkin: To be honest, I really do feel like having the remote API as the only one would be the better option. As in: it would add interoperability, it would make it simpler to implement the MCP / command execution stuff and it would remove the need to keep a separate track for accessing the server. And all it would take to retain the current functionality of launching the client and the server at the same time would be a simple wrapper.

Putting this up for consideration and converting this to draft for now.

pwilkin:
@ngxson since we don't want double APIs, what do you think of a prototype here that does the following:
* removes the cli-specific path
* migrates everything to the http path
* if run without `--endpoint`, launches a server on a random port that is closed as soon as the cli exits to mimic the previous `cli` functionality
?

ngxson:
Honestly I don't have a strong opinion on whether the CLI should use native API, HTTP API or another IPC mechanism like unix socket. However, since most LLM CLI uses HTTP API under the hood, I agree that it may be better in the longer term to go with that for llama-cli.

I do have 2 concerns though:
1. Currently, the CLI acts as an example on how easy it is to use llama.cpp as an external library (via binding; without being a HTTP server). If we choose to move CLI away from this, we must still need to add an example of doing so (though, can be much more basic than CLI)
2. If we choose to use HTTP for CLI, we should no longer link CLI against `libserver`. The consequence is that CLI must either spawn an `llama-server` instance, or `llama-server` should be a daemon.

For the point (1), no actions is needed from your side, I will eventually implement it (which goes back to the idea of `llamax` library), there are many people already asking for a easy-to-use native API that accepts multimodal. However, for the point (2), I think we need to consider it more carefully.

pwilkin:
@ngxson cross-platform daemon management can get really tricky, so I'd prefer not to go that route. I'd say spawning a `llama-server` instance that gets shut down at `cli` close would be the preferred way to go.

ngxson:
hmm ok then, I think there is an extra edge case is where the server runs in router mode, but I think it can be handled in the future. would be nice when we finally have hot-swappable models on CLI though

in the longer term, I think having server as a daemon can be quite nice, but for now an `--endpoint` option can be a good start

common/arg.cpp
```diff
@@ -1380,6 +1380,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.show_timings = value;
         }
     ).set_examples({LLAMA_EXAMPLE_CLI}).set_env("LLAMA_ARG_SHOW_TIMINGS"));
+    add_opt(common_arg(
+        {"--endpoint"}, "URL",
```
ngxson, line 1384: IMO this should be a bit more specific, smth like `--server-base`

and in the future when server becomes a daemon, maybe with a fixed port like `8800`, we can move `--server-base` to default mode and use the mentioned port

pwilkin:
@ngxson so, experimental findings:

-> the cli works pretty nicely with llama-server spawning
-> but properly handling Ctrl+C is a PITA
-> needed to patch the library to allow detaching the server, otherwise the SIGINT would propagate and kill the server

When you're free take a look and tell me if this path is fine. Looked at the code and exorcised any stupid-stuff Kimi wrote. Parameter passing seems to work correctly (i.e. `llama-cli -m <qwen> --reasoning off` actually results in the model not reasoning).

pwilkin:
@ngxson rebased it on master, I guess we should go through with this, as this would unblock easy `llama-cli` tools integration + the MCP server-side server.

vendor/sheredom/subprocess.h
ngxson, line 1: this is a vendor lib, it should be use as-is, no modifications

vendor/sheredom/subprocess.h
```diff
@@ -89,7 +89,12 @@ enum subprocess_option_e {
   // Search for program names in the PATH variable. Always enabled on Windows.
   // Note: this will **not** search for paths in any provided custom environment
   // and instead uses the PATH of the spawning process.
-  subprocess_option_search_user_path = 0x10
+  subprocess_option_search_user_path = 0x10,
+
+  // Spawn the child in a new session/process group so that terminal signals
+  // (e.g. SIGINT from Ctrl+C) are not forwarded to the child process.
+  // On Unix this uses setsid(); on Windows it uses CREATE_NEW_PROCESS_GROUP.
+  subprocess_option_new_session = 0x20
```
ngxson: this is quite overkill IMO, just look at the server router mode, it simply monitor stdin/stdout between 2 processes and if one of the pipe breaks, exits itself

so even router instance crashes, it leaves no abandon children behind

tools/cli/cli-backend.h
ngxson, line 1: from high-level view, I'm not very comfortable about this design because this notion of "backend" is not well-defined (e.g. no clear functional boundary)
- first of, I don't think this should be called a "backend", it's better to be a "client", `cli_client` for example, refers to "API client"
- secondly, `cli_client` should only provide the data, it should not modify the view, i.e. should never call `console::` functions

`cli_client` should simply expose something like `create_chat_completions` and anything else should be provided as callback

process management (i.e. `spawn_...`) should be in its own class, smth like `std::optional<cli_server>` that will be, as the name suggested, optional as user might want to use a remote server

ngxson (commented):
the idea is ok but I don't have a very strong feeling about the current implementation.

IMO the CLI can now be viewed as a frontend project (similar to the svelte web ui), so it should mirror the same frontend paradigm, like for example MVC or centralized state management. many CLI agents (if not all of them) are designed that way and I think that will be the proper impl.

since I already had a good picture in mind, probably the best is that I can vibe-code a half-working version with the proper architecture, and you can build upon that. does his way sound better to you?

pwilkin:
@ngxson yeah, that sounds good, I saw that there's been a strong push towards the new architecture so if you could sketch the intended direction I'll gladly build on that.

ngxson:
I drafted this gist to show the direction: https://gist.github.com/ngxson/ccc774a2441cbfc98ce082c31d94075a

for `cli_server`, I think create a whole new process is quite overkill; running server inside a thread should be enough, my impl is just 50 LOC to manage that

for `cli_client`:
- it should be somewhat equivalent to [openai python package](https://pypi.org/project/openai/0.26.5/)
- contains only HTTP-related functions
- any real-time responses must be returned via callback (for example with SSE, a callback can be invoked multiple times)
- it should not call anything related to `console::`
- it's roughly equivalent to the "model" in MVC

the final part missing from my code above is the `cli_context`, but it will have something like:

```cpp
struct cli_context {
    std::optional<cli_server> server; // only set if user doesn't specify remote server
    cli_client client; // always initialized
    // ...
};
```

the `cli_context` can be put into a new file `cli-context.cpp|h` if you like, it should manage the state of the whole CLI app and responsible to make calls to `console::` to render the view. in other words, the `cli_context` is the "controller" in MVC

we don't have `cli_view` for now as the `console::` is already the view layer. in the future, we can also have a TUI lib that is responsible for the view, but that will be a future improvement

lmk if that sounds good to you

pwilkin:
Aight, I'll take a look and let you know (I think I get the general idea).

pwilkin:
@ngxson Mkay, experimenting with remote CC a bit, told it to implement it according to the spec then yelled at it for not respecting MVC abstractions, think it looks decent now.

ngxson:
to say honestly, I think this is now quite over-complicated / slop than I initially expected (for example, my points about spawning thread and point about not having view.cpp are not respected)

given I also have CC subscription paid by HF (plus pi agents with HF inference endpoints), probably better for me to just do it on my side. I can push a new PR tmr if you are ok with it.

the only part that I can likely salvage is the cli-client.cpp from this PR, I will put you in the Co-Author

ngxson:
Ok so I decided to take over this task, I will push another PR that clean things up, most notably:
- Spawn server in a thread, not a dedicated subprocess (for simplicity)
- Make view layer more generic. The view layer in this PR is a thin layer on top of `console::`, that works but quite redundant. Instead, I want a generic view layer that is designed to (maybe) replaced by a dialog-based TUI in the future; or at least, support render-loop style UI implementation

ngxson:
@pwilkin PTAL https://github.com/ggml-org/llama.cpp/pull/24948

PR #21691, opened 2026-04-09
debug: functionality to dump full tensors and compare

pwilkin:
## Overview

Extracted from my Gemma BF16 experiments, changes to the debug / compare tensors functionality.

## Additional information

This adds the following functionality:
* support for (selective) full tensor dump for both `llama-debug` and the `run-org-model.sh` script
* a script that compares the tensors pairwise using a "matching tensors" blueprint in YAML format

# Requirements

- AI usage disclosure: Yes, mostly for generating the python scripts


am17an:
@pwilkin what about creating a "skill" which is about debugging tensors from HF vs llama.cpp and identifying the issue? Saw some skills integrated into sglang https://github.com/sgl-project/sglang/tree/main/.claude/skills

pwilkin:
@am17an interesting idea, will take a look :)

pwilkin:
Bump, could use some approvals here :)

@ngxson this might be useful for your newest model testing.

pwilkin:
@danbev could you take a quick look at this? It's an old PR but I think it's still useful.

examples/debug/debug.cpp
```diff
@@ -3,15 +3,107 @@
 #include "common.h"
 #include "log.h"
 #include "llama.h"
-
 #include <cstdlib>
+#include <optional>
```
danbev, line 7: There is a optional include a few lines down on line 12.

examples/model-conversion/scripts/causal/validate-conversion.py
```diff
@@ -0,0 +1,324 @@
...
+    sys.exit(1)
+
+
+# ---------------------------------------------------------------------------
+# Main
+# ---------------------------------------------------------------------------
+
+def main():
+    args  = parse_args()
+    script_dir = Path(__file__).parent
```
danbev, line 160: Currently this will cause the `data` directory used to be in `scripts/causal`. I think it would be nice to use the current working directory instead, so if this is run from examples/model-conversion all files would end up in the data directory there.

danbev (approved):
examples/debug/debug.cpp does not compile at the moment, I'll push a suggestion for it shortly.

PR #21693, opened 2026-04-09
server : improve cache reuse diagnostics for SWA and hybrid models

1oridevs:
## Overview

This PR improves diagnostics when prompt cache reuse falls back to full prompt re-processing.

It adds additional logging around SWA and hybrid/recurrent memory cases so it is easier to understand why cache reuse fails. This change does not modify any cache reuse logic — it only improves observability.

## Additional information

The change is limited to `tools/server/server-context.cpp` and only affects logging behavior.

# Requirements

- AI usage disclosure: YES — I used AI for brainstorming and wording help, but I wrote and verified the changes myself and take full responsibility for this PR.

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -1985,6 +1985,25 @@ private:
                 } break;
         }
     }
+    static const char * cache_reuse_reject_reason_str(
+        const llama_model * model,
+        const common_params & params_base) {
+    const bool has_swa = llama_model_n_swa(model) > 0;
+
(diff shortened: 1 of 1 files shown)


1oridevs:
Thanks for the note. I updated the PR description with an explicit AI usage disclosure.

I used AI for brainstorming and wording assistance only. I made the code changes myself, reviewed the diff, built the project locally, and take full responsibility for this submission.

If any further clarification is needed, I’m happy to provide it.

PR #21695, opened 2026-04-09
server: fix SWA prompt reuse boundary condition

1oridevs:
This fixes an SWA prompt-reuse edge case in the server.
At the sliding-window boundary, pos_min == pos_next - n_swa was treated as invalid for reuse, which can cause unnecessary full prompt reprocessing.

This patch adjusts the boundary handling so reuse remains valid at that edge and aligns checkpoint lookup with the same behavior.

Tested locally with long prompts on an SWA-enabled server. Before this change, some boundary cases reset and reprocessed the full prompt; after this change, reuse is preserved in those cases.

No intended behavior change outside SWA boundary handling.


1oridevs:
Ive used cursor AI to help me write this pl description

ORippler (commented):
Please remove unrelated hip changes from the PR prior to review (note I did not take a look at the actual changes in server-context.cpp, as I'm unfamiliar with that part of the code-base)

1oridevs:
Done

PR #21697, opened 2026-04-09
common : enable reasoning budget sampler for gemma4

berkidem:
## Overview

As #21487 also reports, gemma4 thinking budget doesn't work. I noticed that `common_chat_params_init_gemma4()` sets `supports_thinking = true` but never populates `thinking_start_tag` / `thinking_end_tag`. The budget sampler in `server-common.cpp` works conditional on `thinking_end_tag` being non-empty, so it skips gemma4 entirely.

So I added the missing tags. The main fix is just two lines (chat.cpp:1087-1088). The rest of the diff is about making budget=0 work cleanly: while testing for my personal use (see the details of the local testing environment below), I found that budget=0 causes a PEG parse error because the sampler forces the end tag before the model emits a newline after "thought". Even though `--reasoning off` already handles the no-thinking case, I didn't want to introduce a parse error at that edge case. I made the newline optional in the parser, and added a test case for it.

Fixes #21487

## Changes

- Set `thinking_start_tag = "<|channel>thought"` and `thinking_end_tag = "<channel|>"` in `common_chat_params_init_gemma4()`
- Make `\n` optional after `thought` in both PEG parser rules (extract and non-extract paths) so the parser handles empty thinking blocks
- Add test case for empty thinking block (budget=0 scenario)

## Testing

Unit tests: `test-reasoning-budget` (5/5), `test-chat` (all passed).

Integration tested on RTX 5090 with `gemma-4-26B-A4B-it-Q4_K_M.gguf` (ggml-org), llama.cpp b8718, CUDA 13.0:

| Config | Thinking tokens |
|--------|----------------|
| `--reasoning-budget -1` (unlimited) | 1,447 |
| `--reasoning-budget 1024` | 1,024 |
| `--reasoning-budget 0` | 0 |
| Per-request `thinking_budget_tokens: 256` | 256 |

I didn't run perplexity/bench since my PR simply allows the existing sampler that already works for other models to also apply to gemma4. Default behavior (`--reasoning-budget -1`) is unchanged so the only people affected are those who explicitly set a reasoning budget (and don't get it). (Also the CI test uses Qwen3 so my diff wouldn't be on the code path that gets exercised.)

Note on the start tag: the tag omits the trailing `\n` because gemma4 generates a double-newline token (`\n\n`, token 108) after "thought", not the single-newline token (`\n`, token 107) that the tokenizer produces for `"thought\n"`. Without this, the sampler's token matcher fails on the third token and never activates.

# Requirements

- AI usage disclosure: AI assisted with locating the root cause in the source code and suggested a fix. Fix was reviewed and tested manually, and extensively both for this PR and for my own local projects as well.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1104,9 +1104,9 @@ static common_chat_params common_chat_params_init_gemma4(const common_chat_templ
         auto start = p.rule("start", p.prefix(inputs.generation_prompt, "<|channel>"));
 
         if (extract_reasoning) {
-            p.rule("thought", p.literal("<|channel>thought") + p.optional(p.literal("\n")) + p.reasoning(p.until("<channel|>")) + p.literal("<channel|>"));
+            p.rule("thought", p.literal("<|channel>thought") + p.space() + p.reasoning(p.until("<channel|>")) + p.literal("<channel|>"));
         } else {
-            p.rule("thought", p.content(p.literal("<|channel>thought") + p.optional(p.literal("\n")) + p.until("<channel|>") + p.literal("<channel|>")));
+            p.rule("thought", p.content(p.literal("<|channel>thought") + p.space() + p.until("<channel|>") + p.literal("<channel|>")));
         }
 
         auto thought = (p.peek(p.literal("<|channel>")) + p.ref("thought")) | p.negate(p.literal("<|channel>"));
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1083,7 +1083,9 @@ static common_chat_params common_chat_params_init_gemma4(const common_chat_templ
 
     data.prompt            = common_chat_template_direct_apply_impl(tmpl, inputs);
     data.format            = COMMON_CHAT_FORMAT_PEG_GEMMA4;
-    data.supports_thinking = true;
+    data.supports_thinking  = true;
+    data.thinking_start_tag = "<|channel>thought";
+    data.thinking_end_tag   = "<channel|>";
 
     data.preserved_tokens = {
         "<|channel>",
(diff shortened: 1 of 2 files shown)


pwilkin:
Any reason `p.optional(p.literal("\n"))` can't be simply replaced with `p.space()`?

pwilkin (approved):
LGTM, but would feel safe if @aldehir took a peek at it too.

berkidem:
> Any reason `p.optional(p.literal("\n"))` can't be simply replaced with `p.space()`?

Good point, no particular reason; I was overfitting to the exact pattern I saw (it was the same in all experiments). Changed to p.space().

aldehir (approved):
Looks good, thank you.

CMay:
This merge caused a 30% drop in token generation speed on my hardware, from 98t/s down to 70t/s.  Can you take a second look at this and see if there are any obvious culprits?

Reported it here: https://github.com/ggml-org/llama.cpp/issues/21784

aldehir:
@CMay, do you see your t/s increase if you comment out these two lines?
https://github.com/ggml-org/llama.cpp/blob/920b3e78cba8c2ae440691bdea02c96a20a51a54/common/chat.cpp#L1096-L1097

CMay:
@aldehir I'm not setup to compile Vulkan llama.cpp, but what I can say is that using --chat-template to specify any other template gives me normal performance.  It's just that, when I specify any external template (the latest Google one, or any others) the model spits out the template and acts like it's part of the prompt.  So that's not an option for me, even if it is faster.

For what it's worth, I did just try --reasoning off to see if that made any difference in the code path, but the performance didn't improve.

From what I can tell in the code, there isn't any platform specific magic happening, so the performance drop seems like it would be testable on any platform even if it's more or less pronounced depending on the hardware.

CMay:
If there isn't a more performant way to do it, maybe it would be better if that entire process is bypassed unless a reasoning budget parameter is used?

berkidem:
I think it should be easy to skip the sampler when the budget is unlimited; I just need to add something like `params.reasoning_budget_tokens >= 0)` but I need to check if that works.

I'll test this today.

PR #21704, opened 2026-04-10
common : better align to the updated official gemma4 template

aldehir:
## Overview

Google has pushed an update to their chat template: https://huggingface.co/google/gemma-4-31B-it/commit/e51e7dcdb6febd74c182fe0cb41c236363ae2ac5

This update includes everything within our internal workarounds, as well as the custom modifications in the `models/templates/google-gemma-31B-it-interleaved.jinja` template. Add support by detecting it and forgoing the workarounds. Additionally, emit a warning message so users are aware there is an update.

The existing template within GGUFs, as well as the custom interleaved template, will continue to function. I even added some of the formatting changes to the bos and think tokens.

**NOTE**: The new template will work without this PR. I checked and even after building the model turn to use `tool_responses`, the template formats it properly. This PR better aligns to the template since it now handles OpenAI chat completions style messages natively.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/models/templates/google-gemma-4-31B-it-interleaved.jinja
+++ b/models/templates/google-gemma-4-31B-it-interleaved.jinja
@@ -152,7 +152,7 @@
 
 {%- set ns = namespace(prev_message_type=None, last_user_message=-1) -%}
 {%- set loop_messages = messages -%}
-{{- bos_token }}
+{{- bos_token -}}
 {#- Handle System/Tool Definitions Block -#}
 {%- if (enable_thinking is defined and enable_thinking) or tools or messages[0]['role'] in ['system', 'developer'] -%}
     {{- '<|turn>system\n' -}}
@@ -255,13 +255,13 @@
                             {{- item['text'] | trim -}}
                         {%- endif -%}
                     {%- elif item['type'] == 'image' -%}
-                        {{- '\n\n<|image|>\n\n' -}}
+                        {{- '<|image|>' -}}
                         {%- set ns.prev_message_type = 'image' -%}
                     {%- elif item['type'] == 'audio' -%}
                         {{- '<|audio|>' -}}
                         {%- set ns.prev_message_type = 'audio' -%}
                     {%- elif item['type'] == 'video' -%}
-                        {{- '\n\n<|video|>\n\n' -}}
+                        {{- '<|video|>' -}}
                         {%- set ns.prev_message_type = 'video' -%}
                     {%- endif -%}
                 {%- endfor -%}
--- a/models/templates/google-gemma-4-31B-it-interleaved.jinja
+++ b/models/templates/google-gemma-4-31B-it-interleaved.jinja
@@ -152,14 +152,14 @@
 
 {%- set ns = namespace(prev_message_type=None, last_user_message=-1) -%}
 {%- set loop_messages = messages -%}
-{{ bos_token }}
+{{- bos_token }}
 {#- Handle System/Tool Definitions Block -#}
 {%- if (enable_thinking is defined and enable_thinking) or tools or messages[0]['role'] in ['system', 'developer'] -%}
     {{- '<|turn>system\n' -}}
 
     {#- Inject Thinking token at the very top of the FIRST system turn -#}
     {%- if enable_thinking is defined and enable_thinking -%}
-        {{- '<|think|>' -}}
+        {{- '<|think|>\n' -}}
         {%- set ns.prev_message_type = 'think' -%}
     {%- endif -%}
 
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1914,7 +1914,12 @@ std::optional<common_chat_params> common_chat_try_specialized_template(
 
     // Gemma4 format detection
     if (src.find("'<|tool_call>call:'") != std::string::npos) {
-        workaround::convert_tool_responses_gemma4(params.messages);
+        if (src.find("{#- OpenAI Chat Completions:") == std::string::npos) {
+            // apply workarounds if using the older gemma4 templates
+            LOG_WRN("%s: detected an outdated gemma4 chat template, applying compatibility workarounds. "
+                    "Consider updating to the official template.\n", __func__);
+            workaround::convert_tool_responses_gemma4(params.messages);
+        }
         return common_chat_params_init_gemma4(tmpl, params);
     }
 
--- a/models/templates/google-gemma-4-31B-it.jinja
+++ b/models/templates/google-gemma-4-31B-it.jinja
@@ -11,34 +11,15 @@
                 description:<|"|>{{ value['description'] }}<|"|>
                 {%- set add_comma = true -%}
             {%- endif -%}
-            {%- if value['nullable'] %}
-                {%- if add_comma %},{%- else -%} {%- set add_comma = true -%} {% endif -%}
-                nullable:true
-            {%- endif -%}
             {%- if value['type'] | upper == 'STRING' -%}
                 {%- if value['enum'] -%}
                     {%- if add_comma %},{%- else -%} {%- set add_comma = true -%} {% endif -%}
                     enum:{{ format_argument(value['enum']) }}
                 {%- endif -%}
-            {%- elif value['type'] | upper == 'OBJECT' -%}
-                ,properties:{
-                {%- if value['properties'] is defined and value['properties'] is mapping -%}
-                    {{- format_parameters(value['properties'], value['required'] | default([])) -}}
-                {%- elif value is mapping -%}
-                    {{- format_parameters(value, value['required'] | default([])) -}}
-                {%- endif -%}
-                }
-                {%- if value['required'] -%}
-                    ,required:[
-                    {%- for item in value['required'] | default([]) -%}
-                        <|"|>{{- item -}}<|"|>
-                        {%- if not loop.last %},{% endif -%}
-                    {%- endfor -%}
-                    ]
-                {%- endif -%}
             {%- elif value['type'] | upper == 'ARRAY' -%}
                 {%- if value['items'] is mapping and value['items'] -%}
-                    ,items:{
+                    {%- if add_comma %},{%- else -%} {%- set add_comma = true -%} {% endif -%}
+                    items:{
                     {%- set ns_items = namespace(found_first=false) -%}
                     {%- for item_key, item_value in value['items'] | dictsort -%}
                         {%- if item_value is not none -%}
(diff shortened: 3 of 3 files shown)


TomTheWise:
Hi just a question to clarify.
The template is named to the 31B dense.
Is it exactly the same and perfectly aligned for the other versions like the MoE 26B-A4B?

if it’s the same for every model it could help to name the template just Gemma 4 (without parameter count)

thank you for your work!

aldehir:
@TomTheWise, 31B and 26B are the same. E2B and E4B are slightly different but you can pick up the updated templates from their respective hf repos:

E4B: https://huggingface.co/google/gemma-4-E4B-it/blob/main/chat_template.

Re: naming. With the updated official templates properly formatting reasoning traces, the guidance going forward is to use those.

models/templates/google-gemma-4-31B-it-interleaved.jinja
```diff
@@ -152,14 +152,14 @@
 
 {%- set ns = namespace(prev_message_type=None, last_user_message=-1) -%}
 {%- set loop_messages = messages -%}
-{{ bos_token }}
+{{- bos_token }}
```
kurnevsky: Should be `{{- bos_token -}}`

CISC: Not really, someone has just been overcautious with adding `-%}/-}}` everywhere, as long as the next entry starts with `{%-/{{-` you don't need it.

pwilkin (approved):
A lot of people have been requesting this: can you please rename the templates from "google-gemma-4-31B-it.jinja" to just "google-gemma-4-it.jinja"? People are asking questions whether it can be used for the 26B MoE model and if there are any differences.

ngxson (approved):

aldehir:
> A lot of people have been requesting this: can you please rename the templates from "google-gemma-4-31B-it.jinja" to just "google-gemma-4-it.jinja"? People are asking questions whether it can be used for the 26B MoE model and if there are any differences.

I think we just direct them to the official templates now.

models/templates/google-gemma-4-31B-it-interleaved.jinja
aldehir: I'll change the whitespace logic to match. I'm only keeping the interleaved template so it doesn't break users that use this. Hopefully the warning will prompt them to update and we can remove it in the future.

CISC (approved):

Dampfinchen:
On a side note, I really appreciate how many of these fixes work without having to re-download the quants. This is what the gguf version 3 format promised from the start.

emansom:
Instead of trying to make Gemma 4 conform and teach it how OpenAI tools work, shouldn't llama.cpp try to emit and parse native Gemma 4 format instead?

https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4

aldehir:
@emansom Everything fed to Gemma is in its native format, since day one. There is no "teaching" involved.

emansom:
> @emansom Everything fed to Gemma is in its native format, since day one. There is no "teaching" involved.

llama.cpp is still converting from OpenAI style input and vice-versa in the server afaict.

aldehir:
Sure, to support clients. It's a lossless conversion. The prompt sent to the model matches with the prompting guide.

emansom:
> Sure, to support clients. It's a lossless conversion. The prompt sent to the model matches with the prompting guide.

I don't think it is lossless. All the examples in LiteRT-LM assume simple flat call-stack argument tools, not ones with deeply nested structures that can get emitted now.

Making the Gemma 4 model confused, and sometimes emit tool call stacks that are not correct.

aldehir:
I'm not understanding. Please provide an actual example.

emansom:
> I'm not understanding. Please provide an actual example.

Example:
https://github.com/google-ai-edge/LiteRT-LM/blob/1428cd7ecaf59c9e9699e0b06d094bae7f8d4fe4/docs/api/cpp/tool-use.md#how-to-use

Problems that arise when apps in the ecosystem use nested tool call arguments:
https://github.com/anomalyco/opencode/issues/21034
https://github.com/anomalyco/opencode/issues/20669

aldehir:
> Example:
> https://github.com/google-ai-edge/LiteRT-LM/blob/1428cd7ecaf59c9e9699e0b06d094bae7f8d4fe4/docs/api/cpp/tool-use.md#how-to-use

Nothing in that documentation indicates the tool arguments must be flattened, only that they are defined as a JSON schema, which may be of arbitrary depth. Beyond that, neither of the linked issues actually involves tools with nested arguments.

> https://github.com/anomalyco/opencode/issues/21034
> https://github.com/anomalyco/opencode/issues/20669

These predate the many fixes that have landed in llama.cpp since release. The first one isn't even llama.cpp, it's LM Studio, so the comparison doesn't hold up.

It's also likely that neither of those issues includes the interleaved template usage I provided in #21418. That template was a stop-gap until Google fixed their official one, precisely because it didn't align with their own prompt guide. Google has since updated it, and the purpose of this PR is to shift prompt generation over to their corrected template.

Many of the tool call errors at release were caused by two things: inference issues in the early days, and an official chat template that didn't match the prompt guide. Most critically, reasoning traces were omitted between tool calls, which dramatically degrades performance in agentic workflows.

If you have a counter-example of a malformed prompt produced via the `/apply-template` endpoint, let me know.

emansom:
@ngxson can the GGUF models on HF be updated to include the new template too? much obliged!

https://huggingface.co/collections/ggml-org/gemma-4

https://huggingface.co/google/gemma-4-31B-it/commit/e51e7dcdb6febd74c182fe0cb41c236363ae2ac5

emansom:
> Nothing in that documentation indicates the tool arguments must be flattened, only that they are defined as a JSON schema, which may be of arbitrary depth. Beyond that, neither of the linked issues actually involves tools with nested arguments.

I have no concrete data, as I lack the hardware to run a full test suite to conform this hypothesis.

However, logically it makes sense that the smaller (E2B, E4B) models would have trouble comprehending larger object depths in tool calls?

Flattening the tool arguments exposed to the model, would help there?

aldehir:
> However, logically it makes sense that the smaller (E2B, E4B) models would have trouble comprehending larger call-stack depths in tool calls?
>
> Flattening the tool arguments exposed to the model, would help there?

I'd classify this as a client-side issue.

Worth noting that Gemma 4's tool calling format (`{key1: <|"|>string value<|"|>, key2:{key3:4}}`) is not [regular](https://en.wikipedia.org/wiki/Regular_language) and can represent structures of arbitrary depth. If it were regular, I'd expect a different tool calling notation that enforces that limitation.

For completeness, I tested E2B against a tool with heavily nested structures and it handled them.

<details>
<summary>Request</summary>

```json
{
  "model": "gemma-4",
  "messages": [
    {
      "role": "user",
      "content": "Deploy the service called 'frontend' to the 'staging' environment. For the config, enable auto-scaling with these limits: CPU should request 250m with a max of 1000m, and enable burst mode with a ceiling of 2000m for 30 seconds. Set memory to 512Mi and replicas between 2 and 8. Add a healthcheck on /healthz every 10 seconds."
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "deploy",
        "parameters": {
          "type": "object",
          "properties": {
            "service": { "type": "string" },
            "environment": { "type": "string" },
            "config": {
              "type": "object",
              "properties": {
                "scaling": {
                  "type": "object",
                  "properties": {
                    "auto": { "type": "boolean" },
                    "limits": {
                      "type": "object",
                      "properties": {
                        "cpu": {
                          "type": "object",
                          "properties": {
                            "request": { "type": "string" },
                            "max": { "type": "string" },
                            "burst": {
                              "type": "object",
                              "properties": {
                                "enabled": { "type": "boolean" },
                                "ceiling": { "type": "string" },
                                "duration_seconds": { "type": "integer" }
                              },
                              "required": ["enabled"]
                            }
                          },
                          "required": ["request", "max"]
                        },
                        "memory": { "type": "string" },
                        "replicas": {
                          "type": "object",
                          "properties": {
                            "min": { "type": "integer" },
                            "max": { "type": "integer" }
                          },
                          "required": ["min", "max"]
                        }
                      },
                      "required": ["cpu", "memory", "replicas"]
                    }
                  },
                  "required": ["auto", "limits"]
                },
                "healthcheck": {
                  "type": "object",
                  "properties": {
                    "path": { "type": "string" },
                    "interval_seconds": { "type": "integer" }
                  },
                  "required": ["path"]
                }
              },
              "required": ["scaling"]
            }
          },
          "required": ["service", "environment", "config"]
        }
      }
    }
  ],
  "tool_choice": "required",
  "stream": false
}
```
</details>

<details>
<summary>Response</summary>

```json
{
  "choices": [
    {
      "finish_reason": "tool_calls",
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "",
        "reasoning_content": "1. **Analyze the Request:** The user wants to deploy a service named 'frontend' to the 'staging' environment with specific configuration details:\n    *   **Auto-scaling:** Enabled (`auto: true`).\n    *   **CPU Limits:** Request `250m`, Max `1000m`, Burst mode enabled with a ceiling of `2000m` for `30` seconds.\n    *   **Memory:** `512Mi`.\n    *   **Replicas:** Min `2`, Max `8`.\n    *   **Healthcheck:** Path `/healthz`, interval `10` seconds.\n\n2. **Examine Available Tools:** The available tool is `deploy`.\n\n3. **Map Request to Tool Parameters:**\n    *   `service`: 'frontend'\n    *   `environment`: 'staging'\n    *   `config`: This needs to be structured according to the `deploy` function definition.\n        *   `scaling`: Must be an object.\n            *   `auto`: `true` (inferred from \"enable auto-scaling\").\n            *   `limits`: Must be an object.\n                *   `cpu`: Must be an object.\n                    *   `request`: \"250m\"\n                    *   `max`: \"1000m\"\n                    *   `burst`: Must be an object.\n                        *   `enabled`: `true`\n                        *   `ceiling`: \"2000m\"\n                        *   `duration_seconds`: 30\n                *   `memory`: \"512Mi\"\n                *   `replicas`: Must be an object.\n                    *   `min`: 2\n                    *   `max`: 8\n            *   `healthcheck`: Must be an object.\n                *   `path`: \"/healthz\"\n                *   `interval_seconds`: 10\n\n4. **Construct the Tool Call (Drafting):** Assemble the parameters into the required JSON structure for the `deploy` tool.\n\n    *   `service`: \"frontend\"\n    *   `environment`: \"staging\"\n    *   `config`:\n        *   `scaling`:\n            *   `auto`: true\n            *   `limits`:\n                *   `cpu`: { `request`: \"250m\", `max`: \"1000m\", `burst`: { `enabled`: true, `ceiling`: \"2000m\", `duration_seconds`: 30 } }\n                *   `memory`: \"512Mi\"\n                *   `replicas`: { `min`: 2, `max`: 8 }\n        *   `healthcheck`: { `path`: \"/healthz\", `interval_seconds`: 10 }\n\n5. **Final Review:** Check if all constraints from the user request have been met by the generated tool call.\n    *   Service: 'frontend' (Yes)\n    *   Environment: 'staging' (Yes)\n    *   Auto-scaling enabled (Yes)\n    *   CPU request 250m, max 1000m (Yes)\n    *   CPU burst enabled (ceiling 2000m, duration 30s) (Yes)\n    *   Memory 512Mi (Yes)\n    *   Replicas 2-8 (Yes)\n    *   Healthcheck /healthz every 10s (Yes)\n\n6. **Generate Output:** Produce the final tool call in the required format.",
        "tool_calls": [
          {
            "type": "function",
            "function": {
              "name": "deploy",
              "arguments": "{\"config\":{\"healthcheck\":{\"interval_seconds\":10,\"path\":\"/healthz\"},\"scaling\":{\"auto\":true,\"limits\":{\"cpu\":{\"burst\":{\"ceiling\":\"2000m\",\"duration_seconds\":30,\"enabled\":true},\"max\":\"1000m\",\"request\":\"250m\"},\"memory\":\"512Mi\",\"replicas\":{\"max\":8,\"min\":2}}}}}"
            },
            "id": "QvyumYFBDE2OChEg5lsoQoYyOtfBusWd"
          }
        ]
      }
    }
  ]
}
```

image
</details>

That said, I do agree this isn't ideal, and clients should limit the depth of their tool call arguments. The `response_format` feature is better suited for deeply nested structures (e.g., data extraction).

emansom:
> For completeness, I tested E2B against a tool with heavily nested structures and it handled them.

This test does not reflect real world usage with a large context, bloated system prompts; while expecting the model to interpret and understand that it needs to call with a small difference at a certain depth.

> That said, I do agree this isn't ideal, and clients should limit the depth of their tool call arguments. The `response_format` feature is better suited for deeply nested structures (e.g., data extraction).

Working on a few patches towards OpenCode to better the situation.

PR #21714, opened 2026-04-10
codeowners: add team members

0cc4m:
This reverts commit cf45437d35eecd3f05a21968a4958a50e8038fb6.

## Overview

The problem is that Github organization teams are invisible to anyone who is not a member of the organization. This unnecessarily obscures who is responsible for what. We can still use teams manually for pings or review requests.

I manually went through the changes and fixed the cases where something else was edited in the file.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,21 @@
 # collaborators can optionally add themselves here to indicate their availability for reviewing related PRs
-# multiplie collaborators per item can be specified
+# multiple collaborators per item can be specified
+#
+# ggml-org/ci               : CISC, danbev, ggerganov, netrunnereve, ngxson, taronaeo
+# ggml-org/ggml-cann        : hipudding
+# ggml-org/ggml-cuda        : JohannesGaessler, am17an, IMbackK, ORippler
+# ggml-org/ggml-hexagon     : lhez, max-krasnyansky
+# ggml-org/ggml-metal       : ggerganov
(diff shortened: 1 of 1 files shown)


CISC:
Hmmm, that's a bit silly, it's really useful to have teams automatically assigned for reviews, not to mention not having to update `CODEOWNERS` whenever a team is updated...

Anyway, we should not revert it back to what it was, but rather expand the teams into its members if we're going to do this.

0cc4m:
We can do that, I wasn't sure which changes were only due to teams. Maybe it's possible to keep the team + the maintainer members of it in the file?

CISC:
> We can do that, I wasn't sure which changes were only due to teams. Maybe it's possible to keep the team + the maintainer members of it in the file?

Perhaps a compromise; keep the teams, but add one main responsible member? If that works...

am17an:
Can we change visibility of some teams which do code reviews? Like the backend teams

0cc4m:
They are already set to visible, but sadly that just means "visible within the organization". From the outside not much is shown:
image

ggerganov:
Yes, it's a limitation of Github to not shown the teams outside the organization.

We can add comments in the CODEOWNERS file that mirrors the team members. F.ex.:

```bash
# ggml-org/team-abc : member1, member2, ...
```

0cc4m:
I have added comments as suggested.

ggerganov (approved):

PR #21717, opened 2026-04-10
TP: fix arbitrary -ot

### LINKED ISSUE #21686: Eval bug: tensor parallelism failing with -nkvo (Gemma 31B)
### Name and Version

pwilkin@SYN-PC-11:/devel/models$ llama-cli --version
load_backend: loaded BLAS backend from /devel/tools/llama.cpp/build/bin/libggml-blas.so
ggml_cuda_init: found 2 CUDA devices (Total VRAM: 31679 MiB):
  Device 0: NVIDIA GeForce RTX 5070 Ti, compute capability 12.0, VMM: yes, VRAM: 15839 MiB
  Device 1: NVIDIA GeForce RTX 5070 Ti, compute capability 12.0, VMM: yes, VRAM: 15839 MiB
load_backend: loaded CUDA backend from /devel/tools/llama.cpp/build/bin/libggml-cuda.so
load_backend: loaded CPU backend from /devel/tools/llama.cpp/build/bin/libggml-cpu-alderlake.so
version: 8738 (d6f303004)

pwilkin@SYN-PC-11:/devel/models$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Wed_Aug_20_01:58:59_PM_PDT_2025
Cuda compilation tools, release 13.0, V13.0.88
Build cuda_13.0.r13.0/compiler.36424714_0

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

2x 5070 Ti 16 GB
CPU: Intel(R) Core(TM) i7-14700KF

### Models

Gemma 31B Q5_K_S (bartowski)

Running 

```console
llama-server -m google_gemma-4-31B-it-Q5_K_M.gguf -c 150000 -a syndatis --mmproj mmproj-google_gemma-4-31B-it-q8_0.gguf --chat-template-file /devel/tools/llama.cpp/models/templates/google-gemma-4-31B-it-interleaved.jinja --host 0.0.0.0 --cache-ram 4096 -ctxcp 4 -np 1 -sm tensor -nkvo
```

fails with an assertion error. The same model runs fine on smaller context without `-nkvo`.

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console
/devel/tools/llama.cpp/ggml/src/ggml-backend-meta.cpp:729: GGML_ASSERT(src_ss[1].axis == GGML_BACKEND_SPLIT_AXIS_2) failed
```
</details>

JohannesGaessler:
Fixes https://github.com/ggml-org/llama.cpp/issues/21686 .

The problem is that the meta backend always assumes a mirrored layout for the graph inputs created by the backend scheduler. However, with options like `-nkvo` this is not correct because in order to correctly align the KV cache with the split weights surrounding it the KV cache also needs to be split. The meta backend determines split states by going up the chain of `ggml_tensor::src` until it finds statically allocated weights that have a fixed split state. So the issue can be fixed by finding the original tensor that is being copied to the meta backend and propagating the split state from there (since this logic only depends on tensor ops and shapes). In the meta backend this requires

1. a way to identify which tensors are inputs created by the backend scheduler, and
2. a reference to the original tensor that is being copied.

This PR implements 1 by comparing the tensor name and 2 by setting `ggml_tensor::src[0]` of the graph input to the original tensor; since the graph input has `GGML_OP_NONE` this should be safe. I am not happy with this implementation though.

For the split states of weights I am currently comparing tensor names which is I think a hacky and bad way to do it. I've been thinking that it would be better to, when creating the statically allocated tensors for the model, create a map of `ggml_tensor *` -> `enum llm_tensor` and to use that to determine which tensors should receive which split state. But this would not work for the dynamically allocated tensors of the backend scheduler. One solution would be to set a flag for those tensors. But long-term I've been thinking whether it would maybe make sense to add something like a `GGML_OP_BACKEND_COPY` and to do the data copies as part of the ggml graph. That would also give the meta backend a natural way to handle this edge case.

# Requirements

- AI usage disclosure: No

### DIFF
--- a/ggml/src/ggml-backend-meta.cpp
+++ b/ggml/src/ggml-backend-meta.cpp
@@ -16,6 +16,7 @@
 #include <cstring>
 #include <map>
 #include <memory>
+#include <regex>
 #include <string>
 #include <tuple>
 #include <utility>
@@ -452,9 +453,12 @@ static struct ggml_tensor * ggml_backend_meta_buffer_simple_tensor(const struct 
     return it->second[index];
 }
 
-static struct ggml_backend_meta_split_state ggml_backend_meta_get_split_state(const struct ggml_tensor * tensor, bool assume_sync) {
-    const size_t n_bufs = ggml_backend_meta_buffer_n_bufs(tensor->buffer);
-    ggml_backend_meta_buffer_context * buf_ctx = (ggml_backend_meta_buffer_context *) tensor->buffer->context;
+static struct ggml_backend_meta_split_state ggml_backend_meta_get_split_state_impl(
+        const struct ggml_tensor * tensor, bool assume_sync, ggml_backend_meta_buffer_context * buf_ctx,
+        ggml_backend_meta_get_split_state_t get_split_state, void * get_split_state_ud) {
+    const size_t n_bufs = buf_ctx->buf_configs.size();
+    const std::regex pattern_sched_input(".*#.*#.*");
+    const bool tensor_is_sched_input = tensor->op == GGML_OP_NONE && std::regex_match(tensor->name, pattern_sched_input); // FIXME
 
     auto split_states_equal = [&](const ggml_backend_meta_split_state & a, const ggml_backend_meta_split_state & b) -> bool {
         if (a.axis != b.axis) {
@@ -765,9 +769,7 @@ static struct ggml_backend_meta_split_state ggml_backend_meta_get_split_state(co
             return {GGML_BACKEND_SPLIT_AXIS_UNKNOWN, {0}, 1};
         }
         if (ggml_backend_buffer_get_usage(tensor->buffer) != GGML_BACKEND_BUFFER_USAGE_COMPUTE && tensor->view_src == nullptr) {
-            ggml_backend_dev_t dev = ggml_backend_buft_get_device(ggml_backend_buffer_get_type(tensor->buffer));
-            const ggml_backend_meta_device_context * dev_ctx = (const ggml_backend_meta_device_context *) dev->context;
-            ggml_backend_meta_split_state ret = dev_ctx->get_split_state(tensor, dev_ctx->get_split_state_ud);
+            ggml_backend_meta_split_state ret = get_split_state(tensor_is_sched_input ? tensor->src[0] : tensor, get_split_state_ud);
             if (ret.axis >= 0 && ret.axis <= GGML_MAX_DIMS) {
                 const int64_t granularity = ret.axis == GGML_BACKEND_SPLIT_AXIS_0 ? ggml_blck_size(tensor->type) : 1;
                 int64_t ne_sum = 0;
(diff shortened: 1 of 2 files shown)


TheBlueMatt:
Trying to run this + #21808 (on master) currently hits `llama.cpp/ggml/src/ggml-backend.cpp:334: GGML_ASSERT(offset + size <= ggml_nbytes(tensor) && "tensor write out of bounds") failed` when trying to load with `-hf unsloth/Qwen3.5-122B-A10B-GGUF:Q5_K_S -dev Vulkan0,Vulkan1,Vulkan2,Vulkan3 -sm tensor -fa 1 -c 65536` (though there's nothing vulkan-specific in the backtrace - just `llama_model::load_tensors`/`llama_model_loader::load_all_data`/`ggml_backend_meta_buffer_set_tensor`.

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1043,7 +1048,8 @@ static struct ggml_backend_meta_split_state ggml_backend_meta_get_split_state(co
                 if (!srcs_info.empty()) {
                     srcs_info += ", ";
                 }
-                const ggml_backend_meta_split_state split_state = ggml_backend_meta_get_split_state(tensor->src[0], true);
+                const ggml_backend_meta_split_state split_state = ggml_backend_meta_get_split_state_impl(
+                    tensor->src[0], true, buf_ctx, get_split_state, get_split_state_ud);
```
gaugarg-nv, line 1052:
```suggestion
                const ggml_backend_meta_split_state split_state = ggml_backend_meta_get_split_state_impl(
                    tensor->src[i], true, buf_ctx, get_split_state, get_split_state_ud);
```

ggml/src/ggml-backend-meta.cpp
```diff
@@ -452,9 +453,12 @@ static struct ggml_tensor * ggml_backend_meta_buffer_simple_tensor(const struct
...
 
-static struct ggml_backend_meta_split_state ggml_backend_meta_get_split_state(const struct ggml_tensor * tensor, bool assume_sync) {
-    const size_t n_bufs = ggml_backend_meta_buffer_n_bufs(tensor->buffer);
-    ggml_backend_meta_buffer_context * buf_ctx = (ggml_backend_meta_buffer_context *) tensor->buffer->context;
+static struct ggml_backend_meta_split_state ggml_backend_meta_get_split_state_impl(
+        const struct ggml_tensor * tensor, bool assume_sync, ggml_backend_meta_buffer_context * buf_ctx,
+        ggml_backend_meta_get_split_state_t get_split_state, void * get_split_state_ud) {
+    const size_t n_bufs = buf_ctx->buf_configs.size();
+    const std::regex pattern_sched_input(".*#.*#.*");
+    const bool tensor_is_sched_input = tensor->op == GGML_OP_NONE && std::regex_match(tensor->name, pattern_sched_input); // FIXME
```
gaugarg-nv, line 461: Given that this function is called per tensor, `std::regex` can be expensive. Consider replacing this with a simple function like:

```
static bool matches_sched_input_name(const char * name) {
    const char * first = strchr(name, '#');
    return first != nullptr && strchr(first + 1, '#') != nullptr;
}
```

Or, consider making `std::regex` a `static const`.

Stoney49th:
is this a potential fix for https://github.com/ggml-org/llama.cpp/issues/22817 ?

PR #21727, opened 2026-04-10
imatrix: fix division by zero

krampenschiesser:
## Overview

Viewing the imatrix statistics via `./llama-imatrix --in-file imatrix.gguf --show-statistics` does not work in recent llama versions.
It segfaults when opening any imatrix files (unsloth, mraderbacher, local) because the context length in m_params is 0 and then hard-coded in the example.
This is a super basic fix to see the statistics again by using the hardcoded context length from the example itself when its stored as 0.

# Requirements

- AI usage disclosure: NO ai was used

### DIFF
--- a/tools/imatrix/imatrix.cpp
+++ b/tools/imatrix/imatrix.cpp
@@ -38,6 +38,8 @@ static const char * const LLM_KV_IMATRIX_DATASETS    = "imatrix.datasets";
 static const char * const LLM_KV_IMATRIX_CHUNK_COUNT = "imatrix.chunk_count";
 static const char * const LLM_KV_IMATRIX_CHUNK_SIZE  = "imatrix.chunk_size";
 
+static const int32_t CONTEXT_LEN = 512;
+
 struct Stats {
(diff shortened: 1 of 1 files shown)


CISC:
It can be much more basic than this (while also supporting legacy `imatrix.dat`), essentially `m_last_chunk` is not used anyway when using `--show-statistics`, so just force `params.n_ctx = 512` (hardcode it, as this is legacy default, no need for global) here:
https://github.com/ggml-org/llama.cpp/blob/2d46b0f6023b7ee32257bf2c539aa1ef689f99fe/tools/imatrix/imatrix.cpp#L1228

krampenschiesser:
So i can completely remove the m_last_chunk calculation from the imatrix loading?

CISC:
> So i can completely remove the m_last_chunk calculation from the imatrix loading?

No, it's needed when not using `--show-statistics`, the point is you can just force the legacy default then so it won't break; a simple one-liner.

CISC:
Why closing?

PR #21738, opened 2026-04-10
ggml-webgpu: updated matrix-vector multiplication

neha-ha:
## Overview

Improved performance of the matrix-vector multiplication kernel.

# Requirements

- AI usage disclosure:


yomaytk:
I've also been working on optimizing the path of k-quants mat-vec.
Could you share some benchmarks on the changes, if possible?

reeselevine:
Hey @yomaytk that's awesome! I was going to start looking at that this week as well. @neha-ha might have some numbers but she's also developing on an older Intel Mac, so they might not be representative of more modern hardware.

I was thinking of doing a rewrite of the mat-vec shader so that it matches some of the other llama.cpp backends more closely. The general idea is to have one workgroup work on a set of rows of the matrix, and have each thread tile parts of the vector in registers, creating a set of local accumulation values. per-thread. Then, do either a subgroup or a workgroup memory reduction per-output, depending on what the device supports.

I was thinking this would be simpler than the current approach of tiling in shared memory and doing a more complex splitting of a workgroup across rows, while hopefully still maintaining/beating its performance. Not sure what optimizations specifically you're working on for k-quants right now, but does this make sense to you? And do you think your optimizations would fit in/complement these changes?

yomaytk:
@reeselevine Your approach makes sense to me — each thread tiling the vector in registers is simpler and could improve performance. I'm curious to see the results. I've been working on porting the current Q6_K logic to other k-quants types, but since you're already working on a full rewrite, I'm happy to leave the mat-vec work to you and @neha-ha, and focus on other WebGPU tasks :)

reeselevine:
Some graphs with performance comparisons based on this branch vs. current master using `test-backend-ops perf` and this variant for different quantization types:

```
MUL_MAT(type_a=<TYPE>,type_b=f32,m=4096,n=1,k=14336,bs=[1,1],nr=[1,1],per=[0,1,2,3],k_v=0,o=1)
```

Apple M3:
image

NVIDIA RTX 4070:
image

reeselevine (approved):

src/ggml-webgpu.cpp
CISC, line 1: Why this empty file, seems like a mistake?

CISC (approved):

reeselevine (approved):

src/ggml-webgpu.cpp
reeselevine: yep, removed. Can I get a reapproval?

ggerganov (approved):

PR #21744, opened 2026-04-10
ggml-webgpu: Windows D3D12 fallback for ShaderF16-lacking primary ada…

MansfieldPlumbing:
## Overview

On Windows with mixed GPU configs (e.g. primary display on Pascal NVIDIA,
compute on AMD Vega/RDNA), Dawn's HighPerformance adapter selection picks
the display GPU. If that GPU lacks ShaderF16, the backend asserts and crashes.

Adds a generic DXGI enumeration fallback: if the default adapter lacks
ShaderF16, enumerate all adapters and bind to the first that supports it,
logging the adapter name. No hardcoded device strings.

Also fixes DXC loading by resolving dxil.dll and dxcompiler.dll from the
executable directory before Dawn's restricted LoadLibraryEx runs, preventing
silent FXC fallback and associated performance regression.

Tested: AMD Radeon Pro V340 (gfx901, D3D12 FL 12_1) + NVIDIA Quadro P2000
(no ShaderF16), driver 26.20.11016.1, Windows 10 IoT LTSC.
Baseline: 20.40 t/s pp512, Qwen3.5-9B-Q3_K_S.

## Additional information

# Requirements

- AI usage disclosure: YES - used Claude to generalize the adapter enumeration logic from a device-specific prototype to a generic ShaderF16 fallback

### DIFF
--- a/ggml/src/ggml-webgpu/ggml-webgpu.cpp
+++ b/ggml/src/ggml-webgpu/ggml-webgpu.cpp
@@ -15,6 +15,15 @@
 
 #include <webgpu/webgpu_cpp.h>
 
+#ifdef _WIN32
+#  ifndef WIN32_LEAN_AND_MEAN
+#    define WIN32_LEAN_AND_MEAN
+#  endif
+#  include <windows.h>
+#  include <dxgi1_6.h>
+#  pragma comment(lib, "dxgi.lib")
+#endif
+
 #include <atomic>
 #include <cstdint>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


reeselevine:
Thanks for the PR, it does seem like there are valid use-cases where on a system with multiple GPUs you'd want to be able to pick one, beyond the "high-performance/low-power" WebGPU options. Even though WebGPU isn't really built to choose one of several GPUs, I think it's reasonable for the llama.cpp WebGPU backend to be more flexible for people who want to use it natively.

However, I'm not sure this is the best fix. Can't we just use `EnumerateAdapters` and find one that works with the required features? That would also be a more generic approach that works with systems other than Windows.

MansfieldPlumbing:
Good point on EnumerateAdapters — I went the D3D12-direct route because that's the substrate I had hardware to test against (V340L + Radeon Pro driver 19.Q2), but you're right that enumerating through the WebGPU-exposed adapter list would be more portable. I can rework it to iterate adapters via RequestAdapter with feature-availability filtering, falling back only when no ShaderF16-capable primary exists. Will push a revised commit. One question: should the fallback selection log which adapter was chosen and why, so users debugging multi-GPU setups can see the decision?

reeselevine:
Sounds good. I think the adapter chosen already ends up being printed under the `create_webgpu_device` function, is there more that you think should be logged?

MansfieldPlumbing:
You're right, create_webgpu_device already prints the adapter. I'll route the fallback path through the same log so the output is consistent whether we picked the primary or fell back. No additional logging needed.
Reworking the adapter selection fresh against current master and pushing shortly.

reeselevine:
since the comment above seems like it was generated directly by an AI agent, there hasn't been any movement on this in > 1 week, and I am fully capable of prompting agents myself to write code instead of communicating asynchronously with some unknown agent, I am closing this PR.

PR #21752, opened 2026-04-11
common: honor HTTP_PROXY/HTTPS_PROXY env vars in http client

### LINKED ISSUE #21694: Misc. bug: Remote download fails behind proxy with -hf flag
### Name and Version

```
bash-4.4$ ./llama-server --version
version: 8724 (b54cb2e3d)
built with GNU 13.3.1 for Linux x86_64
```

### Operating systems

Linux

llama-server

### Command line

```shell
./llama-server -hf hugging-quants/Llama-3.2-1B-Instruct-Q8_0-GGUF -v
```

### 1. Build

```
cmake -S . -B build \
  -DLLAMA_OPENSSL=ON

cmake --build build --config Release -j 
```
### 2. Behavior
```
bash-4.4$ ./llama-server -hf hugging-quants/Llama-3.2-1B-Instruct-Q8_0-GGUF -v 

common_download_file_single_online: no previous model file found ~/.cache/llama.cpp/hugging-quants_Llama-3.2-1B-Instruct-Q8_0-GGUF_preset.ini 
common_download_file_single_online: HEAD failed, status: -1 
no remote preset found, skipping
```

### 3. Troubleshooting

I ran
```
curl -I https://huggingface.co/
curl -I https://huggingface.co/hugging-quants/Llama-3.2-1B-Instruct-Q8_0-GGUF
```
Both succeed and return:

```
HTTP/1.0 200 Connection Established
HTTP/2 200
```

So the host can reach Hugging Face through the configured proxy.

### 4. Expected behavior

Remote download should work when the host has working HTTPS connectivity, or at minimum fail with a clearer error explaining why the HTTP request could not be completed.

I locally patched `common/http.h` to read HTTPS_PROXY and apply proxy settings to the underlying HTTP client, and that appears to be the right direction for a fix.

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

texasich:
Fixes #21694

## What

`common_http_client()` builds the httplib client but never checks standard proxy environment variables, so any download that goes through this path (e.g. `-hf` flag, model presets) silently fails in environments that require a proxy.

## How

After constructing the client, read `HTTPS_PROXY`/`https_proxy` for `https://` URLs and `HTTP_PROXY`/`http_proxy` for `http://` ones (scheme-specific takes precedence). If set and parseable, pass it to `cli.set_proxy()` and optionally `cli.set_proxy_basic_auth()`. Malformed proxy URLs are caught and silently ignored so nothing breaks for callers without a proxy configured.

Reuses the existing `common_http_parse_url()` to handle auth credentials embedded in the proxy URL (e.g. `http://user:pass@proxy:3128`).

## Notes

- `NO_PROXY`/`no_proxy` bypass list not implemented yet; can be added as follow-up
- `ALL_PROXY` not checked; HTTPS_PROXY and HTTP_PROXY cover the common case
- `std::getenv` is available via `<cstdlib>` already pulled in by httplib.h


texasich:
closing this to get back under the 1-PR limit for new contributors — #21630 is closer to merging so keeping that one active. will reopen this once that lands. branch is preserved on my fork.

texasich:
reopening — #21630 just merged so the new-contributor limit no longer applies. ready for review whenever the maintainers have a minute.

MoRanYue:
Why is this still unmerged?

texasich:
Gentle nudge on this one — been open a while. Happy to rebase if there are conflicts, or make any changes needed. Let me know if anything's blocking review. 🙏

BoneGoat:
Please consider this PR.

nicriverhoo:
+1. Please consider this PR.

MoRanYue:
For a modern software, not respecting to proxy is pretty bad.

ngxson:
I'd assume most maintainers don't use proxy so no one can confidently review / test this PR (me including)

Asking @ServeurpersoCom for your thoughts about this feature / this implementation

ServeurpersoCom:
Need to methodically test the HTTP client on Linux and on Windows with a small Squid as a classic enterprise forward proxy... I'm setting up the platform and will keep you updated.

ServeurpersoCom:
Tested this and the proxy path works fine for -hf, both direct and through the proxy, Linux only. Windows check in progress.

A few non-blocking notes:

- NO_PROXY/no_proxy isn't honored. The client also fetches arbitrary --model-url, docker manifests and runtime image_url, but those are all external targets so it's not urgent, possible follow-up.

- A malformed proxy url is caught silently, so it fails open and falls back to a direct connection with no log. A LOG_WRN in the catch would make that debuggable.

- nit: for http you read uppercase HTTP_PROXY first. Out of habit a lot of code only trusts lowercase http_proxy for the http scheme and ignores the uppercase form, since httpoxy (see: https://httpoxy.org). It doesn't apply here since this isn't CGI, so it's purely cosmetic.

- nit: the comment mentions checking the scheme var then a generic fallback, but both vars are scheme specific and there's no generic one, so it doesn't match the code.

common/http.h
```diff
@@ -91,6 +91,32 @@ static std::pair<httplib::Client, common_http_url> common_http_client(const std:
 
     cli.set_follow_location(true);
 
+    // Honor HTTP_PROXY / HTTPS_PROXY environment variables (both lower- and upper-case).
+    // Convention: check the scheme-specific var first, fall back to the generic one.
```
ServeurpersoCom: nit: I don't understand this "fall back", I think we can delete the second line (// Convention:)

ServeurpersoCom:
Windows OK, quick test with llama-installer.sh Wine + BorringSSL (default) with Zig. If OK, LGTM.

ServeurpersoCom (approved):
Clarify or drop that comment, otherwise LGTM

texasich:
Dropped the misleading fallback comment, the code only has the two scheme-specific vars, no generic one. Good catch.

ServeurpersoCom:
It's a minimalist proxy support that will unlock GGUF downloads for a lot of users, and we can easily do a follow-up PR later if we need to expand it. You could add the env var to the docs, but it's a super standard one for every web app on Linux.

namniav:
Thank you for the patch and it works for me.

May I ask if there is any issue that caused this PR to be closed without being merged?

ServeurpersoCom:
There's quite a bit of work pending on the proxy; I'll try to make progress this weekend.

bsproger:
> +1. Please consider this PR.

+1 from me too

marecabo:
@texasich Thanks for this feature. I addressed @ServeurpersoCom's [review comments](https://github.com/ggml-org/llama.cpp/pull/21752#issuecomment-4564428726) (no_proxy & logging) in https://github.com/texasich/llama.cpp/pull/1.
Note, that since 2026-07-08 [#24948](https://github.com/ggml-org/llama.cpp/pull/24948) no_proxy handling is required to ensure CLI loopback is not sent to proxy.

On top of that, I also merged master back in [`proxy-env-vars-merged`](https://github.com/marecabo/llama.cpp/tree/proxy-env-vars-merged) ([cc576d1](https://github.com/ggml-org/llama.cpp/commit/f61c527ed95e9a4a0be9df67b13719d3a3be29fe)) (just one small readme conflict), tested it on RHEL 9 with ZScaler and can confirm it works, see detailed logs in the PR.

ServeurpersoCom:
Needs a rebase, there's a README conflict with master. I already tested the proxy and network side, Linux and Windows, and it behaves as expected, so once that's sorted the only thing missing is a second approval.

ServeurpersoCom (approved):

common/http.h
```diff
@@ -91,6 +91,31 @@ static std::pair<httplib::Client, common_http_url> common_http_client(const std:
...
+
+    if (!proxy_url.empty()) {
+        try {
+            common_http_url proxy = common_http_parse_url(proxy_url);
+            cli.set_proxy(proxy.host, proxy.port);
+            if (!proxy.user.empty()) {
+                cli.set_proxy_basic_auth(proxy.user, proxy.password);
+            }
+        } catch (...) {
+            // ignore malformed proxy URL — don't break existing callers
```
ngxson: per agents.md: no em-dash

marecabo: @texasich is fixed in https://github.com/texasich/llama.cpp/pull/1/changes#diff-30865ebe44d6724ca6dc550d4eefe93d3ca32abf04dbd8e3869fdfd472c4b632L115

PR #21551, opened 2026-04-07
common : only warn when using unsupported kv cache types

Green-Sky:
## Overview

Add a name (string) to ggml type helper and only warn when kv cache quantization is set to an unsupported type.

## Additional information

Thanks to the new kv rotation if quantized, the lower bound to what seems reasonable for kv quantization has fallen. Also with turboquant prs still coming in, we should just look at existing quants first.

Opening this more as a discussion with numbers with code attached, than an actual code pr.

For unsupported types, you need to either offload the kvcache to cpu (`-nkvo` + `-fa on`) or only partially offload the model (slower `-ngl 1`).

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: NO

---

Some numbers. Some where previously posed here: https://github.com/ggml-org/llama.cpp/pull/21273#issuecomment-4192469220 .

## Perplexity
This metric is ok to use when comparing the same model with itself.

### wikitext2-test 512 !! first 8 chunks only !!
very slow, since its mostly cpu fallback or cuda throwing NaNs.

#### llama2 7b Q4_K_M
| k   | v   | rot | ppl |
|-----|-----|-----|------|
| f16  | f16  | 0 | 6.3123 |
| q4_0 | q4_0 | 0 | 6.4648 |
| q4_0 | q4_0 | 1 | 6.3243 |
| q4_0 | q1_0 | 0 | 207.91 |
| q4_0 | q1_0 | 1 | **9.9382** |
| q1_0 | q1_0 | 1 | 199.33 |

#### gemma3 4b pt q4_0
| k   | v   | rot | ppl | size (base + swa) |
|-----|-----|-----|------|---|
| f16  | f16  | 0 | 7.8551 +/- 0.05036 | 40.00 MiB + 232.00 MiB |
| q8_0 | q8_0 | 1 | 7.8569 +/- 0.05037 | 21.25 MiB + 123.25 MiB |
| q4_0 | q4_0 | 1 | 7.9506 +/- 0.05103 | 11.25 MiB + 65.25 MiB |
| q4_0 | q1_0 | 1 | 9.2384 +/- 0.06022 | 7.03 MiB + 40.78 MiB |
| q3_K | q2_K | 1 | 8.2374 +/- 0.05307 | 7.58 MiB + 43.95 MiB |

### hellaswag (400)
A little better at showing the models performance, but its only using a short context.

#### llama2 7b Q4_K
[...]

PR #21753, opened 2026-04-11
vulkan: Support asymmetric FA in coopmat2 path

jeffbolznv:
## Overview

There has been some recent interest/experimentation with mixed quantization types for FA. I had originally designed the cm2 FA shader with this in mind (because I didn't realize it wasn't supported at the time!), this change adds the missing pieces and enables it.

Also support Q1_0 since people have been trying that out (seems crazy, but who knows).

We should be able to do similar things in the coopmat1/scalar path, but there's another change open against the scalar path and I don't want to conflict.

# Requirements

- AI usage disclosure: YES, I used Cursor (composer-2-fast) for most of this change.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -426,10 +426,12 @@ struct vk_fa_pipeline_state {
     bool f32acc;
     uint32_t flags;
     uint32_t limit_occupancy_shmem;
+    ggml_type k_type;
+    ggml_type v_type;
 
     bool operator<(const vk_fa_pipeline_state &b) const {
-        return std::tie(HSK, HSV, Br, Bc, D_split, row_split, shmem_staging, path, workgroup_size, subgroup_size, aligned, f32acc, flags, limit_occupancy_shmem) <
-               std::tie(b.HSK, b.HSV, b.Br, b.Bc, b.D_split, b.row_split, b.shmem_staging, b.path, b.workgroup_size, b.subgroup_size, b.aligned, b.f32acc, b.flags, b.limit_occupancy_shmem);
+        return std::tie(HSK, HSV, Br, Bc, D_split, row_split, shmem_staging, path, workgroup_size, subgroup_size, aligned, f32acc, flags, limit_occupancy_shmem, k_type, v_type) <
+               std::tie(b.HSK, b.HSV, b.Br, b.Bc, b.D_split, b.row_split, b.shmem_staging, b.path, b.workgroup_size, b.subgroup_size, b.aligned, b.f32acc, b.flags, b.limit_occupancy_shmem, b.k_type, b.v_type);
     }
 };
 
@@ -2967,7 +2969,7 @@ static vk_fa_tuning_params get_fa_tuning_params_coopmat1(const vk_device& device
     return result;
 }
 
-static vk_fa_tuning_params get_fa_tuning_params_coopmat2(const vk_device& device, uint32_t hsk, uint32_t hsv, uint32_t n_rows, uint32_t n_kv, ggml_type kv_type, bool f32acc) {
+static vk_fa_tuning_params get_fa_tuning_params_coopmat2(const vk_device& device, uint32_t hsk, uint32_t hsv, uint32_t n_rows, uint32_t n_kv, ggml_type k_type, ggml_type v_type, bool f32acc) {
     GGML_UNUSED(n_kv);
     GGML_UNUSED(f32acc);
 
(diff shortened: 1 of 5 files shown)


Green-Sky:
I have not run it, but at first glance the code looks like iq4_nl is missing.

---

Thanks for providing this, I will see what testing I can get out of this. But from the testing I did with q1_0 with nkvo, q1_0 hurts the performance of the model too much. It is more or less equivalent to a thirding or quatering of model parameter count.

jeffbolznv:
I took out iq4nl because I wanted to avoid allocating and initializing the shared memory table if it's not used. Is this something people actually use? If so, I can put it back.

ClosetPie107:
This is exactly what I was hoping for, hope it lands on the coopmat1/scalar path eventually as well. Lower V cache quant is pretty bearable now with the hadamard rotation but I still find it tricky to lower K cache quant too much.

Green-Sky:
> This is exactly what I was hoping for, hope it lands on the coopmat1/scalar path eventually as well. Lower V cache quant is pretty bearable now with the hadamard rotation but I still find it tricky to lower K cache quant too much.

I spent some time testing and posted some results in my draft pr: https://github.com/ggml-org/llama.cpp/pull/21551



> I took out iq4nl because I wanted to avoid allocating and initializing the shared memory table if it's not used. Is this something people actually use? If so, I can put it back.

Hard to know. I am once again having trouble getting cm2 to work. In the mean time, I found a test that is more sensitive to kvcache quantization.
https://huggingface.co/datasets/Green-Sky/mmlu-redux-2.0-for-llama.cpp/blob/main/mmlu-redux-2-ok%2Bexpert.bin
mmlu redux is mmlu, sampled 100 tasks from each category and then annotated with errors, in case its wrong. I filtered out the error cases and converted the ok and expert marked tasks and put them into a ABCD answer context. I posted some numbers produced by the cuda backend in the pr mentioned above. You can double parallelism to '-np 8', which should result in a slight speedup. Would be great if you could test some configurations.

```
$ llama-perplexity -m model.gguf -bf mmlu-redux-2-ok+expert.bin --multiple-choice -ctk q8_0 -ctv q8_0 -ngl 999 -fa on -np 8
```

edit: Also the iq4_nl+q3_K combo performed surprisingly well here.

jeffbolznv:
Bumping for review.

0cc4m:
Yeah, sorry, busy week. I'll get to it soon.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -3571,16 +3604,35 @@ static void ggml_vk_load_shaders(vk_device& device) {
     }
 #endif
 #if defined(VK_NV_cooperative_matrix2) && defined(GGML_VULKAN_COOPMAT2_GLSLC_SUPPORT)
+#define CREATE_FA_CM2_MIXED() \
+        for (int fa_k_ty = 0; fa_k_ty < (int)GGML_TYPE_COUNT; ++fa_k_ty) { \
+        for (auto &fa : device->pipeline_flash_attn_f32_f16[fa_k_ty]) { \
```
0cc4m, line 3616: If fa_k_type is part of the pipeline state, I guess technically this array is no longer necessary. We can clean it up when adding mixed support to scalar and CM1 FA.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -15471,38 +15525,27 @@ static bool ggml_backend_vk_device_supports_op(ggml_backend_dev_t dev, const ggm
...
-                default:
+                auto fa_kv_ok = [coopmat2](ggml_type t) {
+                    switch (t) {
+                    case GGML_TYPE_F16:
+                    case GGML_TYPE_F32:
+                    case GGML_TYPE_Q4_0:
+                    case GGML_TYPE_Q8_0:
+                    case GGML_TYPE_Q4_1:
+                    case GGML_TYPE_Q5_0:
+                    case GGML_TYPE_Q5_1:
```
0cc4m, line 15547: Let's use the chance to order them from big to small, and also, what happened to IQ4_NL?

0cc4m (approved):
Looks good overall, the few compatibility steps can be cleaned out once the other FA shaders are updated. I can do that, or do you plan to?

ggml/src/ggml-vulkan/ggml-vulkan.cpp
0cc4m: Ah, this is already answered. I don't know if anyone uses it.

jeffbolznv:
> Looks good overall, the few compatibility steps can be cleaned out once the other FA shaders are updated. I can do that, or do you plan to?

I can do it, but prefer to leave it for a separate change.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
jeffbolznv: Yeah, the other option was to have a separate parallel table and update all code to look at both. As you say, once we support this everywhere it'll get cleaner.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
jeffbolznv: done.

0cc4m (approved):

0cc4m:

pwilkin (approved):

PR #21756, opened 2026-04-11
mtmd : add MERaLiON-2 multimodal audio support

SiruiHe:
This adds support for MERaLiON-2 to mtmd. MERaLiON-2 is a speech-text model developed by I2R, A*STAR Singapore, available in 3B and 10B variants. It uses a Whisper large-v2 encoder paired with a Gemma2 decoder.

New projector type: `PROJECTOR_TYPE_MERALION`

The audio adaptor stacks 15 encoder frames per output token, then runs a layer norm followed by a 4-layer MLP: compression Linear+SiLU, a GLU block (gate and pool projections), and a final out_proj to match the decoder embedding dim. The implementation reuses the existing `linear_{bid}` / `mm_norm_pre` tensor naming so the change to tensor_mapping.py is just a comment update.

The mmproj is generated with `convert_hf_to_gguf.py --mmproj` on the full MERaLiON-2 model directory. The text decoder is a standalone Gemma2 model — weights are stored with a `text_decoder.` prefix that gets stripped before standard Gemma2 conversion.

Tested on ASR benchmarks in EN, ZH, MS, TA. English WER delta vs the bf16 HuggingFace baseline is under 0.002 absolute, confirming the inference graph is correct.

Model: https://huggingface.co/MERaLiON/MERaLiON-2-3B

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -11279,6 +11279,55 @@ class UltravoxWhisperEncoderModel(WhisperEncoderModel):
         self.gguf_writer.add_audio_stack_factor(self.global_config["stack_factor"])
 
 
+@ModelBase.register("MERaLiON2ForConditionalGeneration")
+class MERaLiONWhisperEncoderModel(WhisperEncoderModel):
+    has_vision_encoder = False
+    has_audio_encoder = True
+
+    def get_audio_config(self) -> dict[str, Any] | None:
+        return self.global_config.get("speech_config")
+
+    def set_gguf_parameters(self):
+        super().set_gguf_parameters()
+        self.gguf_writer.add_clip_projector_type(gguf.VisionProjectorType.MERALION)
+        # scale_factor = 15: stacks 15 frames → 1 token (1500 → 100 timesteps)
+        self.gguf_writer.add_audio_stack_factor(self.global_config.get("speech_mlp_scale_factor", 15))
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        # Skip text decoder tensors
+        if name.startswith("text_decoder."):
+            return
+
+        # Map speech_encoder.* → audio_tower.*
+        if name.startswith("speech_encoder."):
+            name = name.replace("speech_encoder.", "audio_tower.")
+
+        # Map ln_speech.* → audio.multi_modal_projector.ln_pre.*
+        if name.startswith("ln_speech."):
+            name = name.replace("ln_speech.", "audio.multi_modal_projector.ln_pre.")
+
+        # Map speech_audio_adapter.* → audio.multi_modal_projector.*
+        # Use indexed linear_{bid} naming to fit A_MMPROJ tensor mapping:
+        #   linear_0 = frame compression (19200 → 6400) + SiLU
+        #   linear_1 = gate_proj (6400 → 6400) for GLU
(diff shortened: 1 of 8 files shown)


SiruiHe:
AI tools were used to help structure the initial implementation, following the Ultravox/Voxtral pattern already in the codebase. The tensor mapping and adaptor graph were worked out by tracing the original model weights and validated on ASR benchmarks, English WER within 0.002 of the bf16 baseline.

The change itself is low impact: it adds one new projector type alongside the existing ones, with no modifications to the core inference path. The motivation is to let users run MERaLiON GGUF files without needing a custom patch on every llama.cpp update.

Also pushed a follow-up commit to fix the comment order (stack happens before LayerNorm, not after) and strip out the verbose per-line annotations, as flagged by Copilot.

tools/mtmd/clip.cpp
```diff
@@ -2017,6 +2020,30 @@ struct clip_model_loader {
...
+                    // Whisper encoder conv layers
+                    model.conv1d_1_w = get_tensor(string_format(TN_CONV1D, 1, "weight"));
+                    model.conv1d_1_b = get_tensor(string_format(TN_CONV1D, 1, "bias"));
+                    model.conv1d_2_w = get_tensor(string_format(TN_CONV1D, 2, "weight"));
+                    model.conv1d_2_b = get_tensor(string_format(TN_CONV1D, 2, "bias"));
+                    // MERaLiON adaptor: 4 linear layers + ln_pre
+                    // linear_0 = frame compression (19200→6400) + SiLU
+                    // linear_1 = gate_proj (6400→6400) for GLU
+                    // linear_2 = pool_proj (6400→6400) for GLU
+                    // linear_3 = out_proj  (6400→3584)
```
ngxson: we prefer ascii arrow whenever possible
```suggestion
                    // linear_0 = frame compression (19200 -> 6400) + SiLU
                    // linear_1 = gate_proj (6400 -> 6400) for GLU
                    // linear_2 = pool_proj (6400 -> 6400) for GLU
                    // linear_3 = out_proj  (6400 -> 3584)
```

convert_hf_to_gguf.py
```diff
@@ -11279,6 +11279,43 @@ def set_gguf_parameters(self):
...
+        if name.startswith("speech_encoder."):
+            name = name.replace("speech_encoder.", "audio_tower.")
+
+        if name.startswith("ln_speech."):
+            name = name.replace("ln_speech.", "audio.multi_modal_projector.ln_pre.")
+
+        if name.startswith("speech_audio_adapter."):
+            name = name.replace("speech_audio_adapter.", "audio.multi_modal_projector.")
+            if ".mlp_adapter.0." in name:
+                name = name.replace(".mlp_adapter.0.", ".linear_0.")
```
ngxson: instead of renaming, better to do:

```py
yield (self.format_tensor_name(gguf.MODEL_TENSOR....., bid), data_torch)
return
```

tools/mtmd/clip.cpp
ngxson: there are also other places you're using unicode arrow, please change them

SiruiHe:
Done, pushed a follow-up commit (ef6cce8). Changed unicode arrows to ASCII in the clip.cpp comments, and replaced the string-rename approach in modify_tensors with format_tensor_name / A_MMPROJ / A_MM_NORM_PRE.

ngxson (approved):
Waiting for an additional review from @CISC before merging

CISC (approved):

Issue #11577, opened 2025-02-01
Feature Request: resize an existing context


giladgd:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggerganov/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggerganov/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

Being able to resize an existing context, either by enlarging it or shrinking it.
When shrinking the context assume we have enough unused slots in the KV cache, and otherwise return an error code.

This would be the ideal API for this:
```cpp
// Change the size of the context.
// When shrinking, ensure that there are enough empty slots in the KV cache to accommodate the new size.
//   0 - success
// < 0 - error. the KV cache does not have enough empty slots to accommodate the new size
LLAMA_API int32_t llama_set_n_ctx(struct llama_context * ctx, uint32_t n_ctx);
```

### Motivation

I want to make the API of [`node-llama-cpp`](https://github.com/withcatai/node-llama-cpp) as simple as possible, and for that I want users to not have to manually specify the context size they need.
At the moment, `node-llama-cpp` calculates the maximum context size that is supported by the model and can fit in the current memory state and constraints and defaults to it. The issue with this approach is that it can reserve a lot of memory, way more that will eventually be used for inference.

I want to be able to implement an incremental memory allocation strategy:
* by default, a small context size will be allocated.
* as the evaluation progresses and needs a larger context, the context will be resized to make room for further evaluation.
* if there's not enough memory available, have `node-llama-cpp` resort to performing a context shift, or unloading/offloading deprioritized things from memory.
* When a sequence is garbage collected or manually disposed (in `node-llama-cpp`), erase its state from the KV cache and shrink the context size to free unused memory.

Being able to resize a context dynamically introduces flexibility for implementing various high-level memory management optimizations.

The current solution for this is to save the state of the context to a file (via `llama_state_load_file`), free it, and then create a larger context and load the state from the file to it, but this approach is inefficient and can be very slow (depending on the context size and the hardware specifics).

ggerganov:
The plan is to refactor the `llama_kv_cache` to become a public object with an API that would allow such kind of operations (#11213).

giladgd:
It's not stale

[...]

PR #21757, opened 2026-04-11
kv : add dynamic KV cache resize (--kv-dynamic)

rockyRunnr:
Add --kv-dynamic flag that starts with a small KV cache (256 cells) and grows on demand via try_resize(). Supports both standalone llama_kv_cache and hybrid (llama_memory_hybrid) architectures.

Growth strategy: doubling for small caches, +1GB linear for large.

## Overview

When a large `-c` is set, llama.cpp allocates the full KV cache upfront. On
Apple Silicon / unified memory this can cause GPU OOM even when actual usage
is small.

`--kv-dynamic` starts the cache at 256 cells and grows as needed:
- `prepare()` fail → `try_resize()` → retry in the same `init_batch()` call
- resize: create new cache → copy per-layer/per-stream → swap internals
- after resize, scheduler reserve is re-triggered in the same decode call
Grow-only for now. Shrink is out of scope for this PR.
Related to #11577.

## Additional information

This is a draft — looking for feedback on direction before iterating further:
- Is grow-only as a first step reasonable?
- Is the create/copy/swap pattern acceptable?
- The current growth heuristics are experimental and based on local testing rather than broad benchmarking.
- Should growth heuristics be configurable rather than hardcoded?

Earlier experiments on Apple M4 (32 GB, Qwen3.5-27B-Q4_K_M, `-c 131072`):

| prompt tokens | vanilla KV | dynamic KV |
|---------------|------------|------------|
| ~100          | 8 GB       | 16 MB      |
| ~6K           | 8 GB       | 512 MB    |
| ~80K          | 8 GB       | 5 GB       |

In earlier local runs, the vanilla path frequently OOMed while the dynamic path completed without OOM.

# Requirements

- AI usage disclosure: YES — used for debugging assistance and code review. PR text written by me.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -1320,6 +1320,13 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_env("LLAMA_ARG_CLEAR_IDLE").set_examples({LLAMA_EXAMPLE_SERVER}));
     add_opt(common_arg(
+        {"--kv-dynamic"},
+        "enable dynamic KV cache resizing (start small, grow on demand)",
+        [](common_params & params) {
+            params.kv_dynamic = true;
+        }
+    ));
+    add_opt(common_arg(
         {"--context-shift"},
         {"--no-context-shift"},
         string_format("whether to use context shift on infinite text generation (default: %s)", params.ctx_shift ? "enabled" : "disabled"),
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -1476,6 +1476,7 @@ struct llama_context_params common_context_params_to_llama(const common_params &
     cparams.op_offload        = !params.no_op_offload;
     cparams.swa_full          = params.swa_full;
     cparams.kv_unified        = params.kv_unified;
+    cparams.kv_dynamic        = params.kv_dynamic;
 
     cparams.type_k = params.cache_type_k;
     cparams.type_v = params.cache_type_v;
--- a/common/common.h
+++ b/common/common.h
@@ -530,6 +530,7 @@ struct common_params {
     bool ctx_shift         = false; // context shift on infinite text generation
     bool swa_full          = false; // use full-size SWA cache (https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055)
     bool kv_unified        = false; // enable unified KV cache
+    bool kv_dynamic        = false; // enable dynamic KV cache resizing
 
     bool input_prefix_bos  = false; // prefix BOS to user inputs, preceding input_prefix
     bool use_mmap          = true;  // enable mmap to use filesystem cache
--- a/include/llama.h
+++ b/include/llama.h
@@ -372,6 +372,7 @@ extern "C" {
         bool kv_unified;  // use a unified buffer across the input sequences when computing the attention
                           // try to disable when n_seq_max > 1 for improved performance when the sequences do not share a large prefix
                           // ref: https://github.com/ggml-org/llama.cpp/pull/14363
+        bool kv_dynamic;  // enable dynamic KV cache resizing (start small, grow on demand)
 
         // [EXPERIMENTAL]
         // backend sampler chain configuration (make sure the caller keeps the sampler chains alive)
--- a/src/llama-context.cpp
+++ b/src/llama-context.cpp
@@ -5,9 +5,31 @@
 #include "llama-batch.h"
 #include "llama-io.h"
 #include "llama-memory.h"
+#include "llama-memory-hybrid.h"
 #include "llama-mmap.h"
 #include "llama-model.h"
 #include "llama-ext.h"
+
+namespace {
+
+bool check_and_clear_resized_kv(llama_memory_i * memory) {
+    if (memory == nullptr) {
+        return false;
+    }
+
+    if (auto * kv = dynamic_cast<llama_kv_cache *>(memory)) {
+        return kv->check_and_clear_resized();
+    }
+
+    if (auto * hybrid = dynamic_cast<llama_memory_hybrid *>(memory)) {
+        auto * mem_attn = hybrid->get_mem_attn();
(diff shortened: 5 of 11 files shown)


jacekpoplawski:
Am I right that this mostly defers OOM rather than eliminating it? If so, what is the advantage over simply limiting the context size at startup?

rockyRunnr:
@jacekpoplawski, Thank you for your comment. 

Limiting the context size at startup forces the user to guess their eventual usage in advance. In many real sessions, the context starts small and only later grows longer. --kv-dynamic keeps the large upper bound available without paying the full KV cost upfront.

mvatafu:
Wondering how would this complement the --parallel flag, since that one is splitting the -c into x pieces. It would mean we would be able to have more parallel small sessions ? That would be really nice.

rockyRunnr:
@mvatafu, thanks for raising this.

Yes, that is part of the goal. It still helps with --parallel by deferring full KV preallocation. I verified this with --parallel 4, where the dynamic KV path is exercised and resize is triggered as expected.

The current limitation is that growth is still cache-wide, not per-sequence. True per-sequence independent growth would require a larger KV layout/resize refactor, so I left that as follow-up work.

michal-zurkowski:
Hmm, to be honest i kinda do not see the scenario where that would be useful, other than you would be using VRAM for something else in the meantime.
We already have `-kvu` that helps with scenario where one slot uses 16k context, and other 100k, where if you have KV set to 1M, you can have 10 slots using 100k context, or 20 slots 50k and. it does not split it like default
```
-kvu,  --kv-unified, -no-kvu, --no-kv-unified
                                        use single unified KV buffer shared across all sequences (default:
                                        enabled if number of slots is auto)
                                        (env: LLAMA_ARG_KV_UNIFIED)
```

It does not help with more slots, it also makes people not aware that they will OOM if they work with that config at some point if they do not know how much KV would take in total. In theory just setting KV to max we can with `-fit` or manually provides a safe environemnt where we guaranetee no OOM and maximize the context size.

ddh0:
I agree, I am not sure this feature fits with the philosophy of GGML / llama.cpp in general.

rockyRunnr:
@michal-zurkowski and @ddh0 , Thanks for the comments. I think the late-OOM concern is fair.

This PR is not trying to reduce the worst-case KV footprint or claim to eliminate OOM once the configured upper bound is actually reached. The goal is narrower: in memory-constrained personal / edge-device setups, avoid paying the full KV cost upfront when actual usage often stays small for much or most of a session. In those environments, deferring commitment can materially reduce initial memory pressure and improve early-session usability, even if the eventual full-KV cost is unchanged.

I also see this as addressing a different problem from `-kvu`. `-kvu` improves how KV capacity is shared across slots / sequences, while `--kv-dynamic` changes when that capacity is committed. So I do not see it as a replacement for `-kvu`, but as an opt-in tradeoff for users who prefer lower upfront allocation pressure.

That said, the late-OOM concern is valid. I am not attached to the current shape as the only acceptable design. If this direction is interesting, I would be happy to move it toward a more explicit / bounded model — for example, lazy growth only within a clear budget, plus better observability such as reporting the estimated full-KV footprint and emitting warnings as the cache grows. Also, this is intended as an opt-in flag only; the default deterministic behavior would remain unchanged.

I opened this PR mainly to discuss whether this tradeoff belongs upstream and, if so, what shape would fit `llama.cpp` best.

cmp-nct:
There are definitely benefits - especially if the KV can shrink again when context resets.

This means you can use llama.cpp for tasks with unknown token requirement (like long reasoning prompts or agentic growth) up to a VRAM size that you know the system can handle barely but you are not comfortable with that usage.
Say you can reserve a sane amount of VRAM and an UPPER bound of VRAM to which the kv-cache can grow, that means your request will not run out of context and once the task is fulfilled you drop back to normal VRAM.

On CUDA+Windows you can even exceed VRAM a bit, the system will start swapping VRAM with RAM (at a massive performance cost) - so llama.cpp could exceed total VRAM temporarily (with performance cost) and then auto-reduce back once the context is not needed anymore.

I've had many situations where this would come very handy.

rockyRunnr:
@cmp-nct, Thanks, I agree that bounded lazy growth is the more interesting direction here, especially for unknown-length workloads. This PR is only a grow-only first step for now. Shrink/reset also seems like a natural part of that broader direction, and I’d definitely be open to discussing it further.

MarkWard0110:
Having this feature would mean you are not loading the model with the worst-case scenario context reserved.  It would enable the system to adjust to the memory requirements based on clients' actual usage.  

Does this feature handle edge cases like the following
1. Multiple GPU, the context starts at an initial size.  Either during the initial prompt request or during generation, the KV cache size is approaching the limits of GPU 1.  Pause generation, reload the model split among two GPUs, then resume generation with the previous cache prefix reloaded into the new cache.   The pause/resume can be a best effort.  It won't drop the client's request because the reload is internal and simply resumes generating the next token.
2. One or more GPUs + System RAM.  If the initial prompt request or during generation, the KV cache size approaches GPU VRAM limits.  Pause generation, reload the model, split among additional GPUs and/or system RAM.  resume generation.

There is a massive delay between tokens when the model is reloaded, but it works.

If the KV cache is less than the model's maximum context size and can't grow due to resource constraints, either perform a silent prompt-fit or return out-of-memory.  

I have prototyped the idea with Ollama, which provides different capabilities because Ollama's host is separate from the runner, allowing it to handle some edge cases differently.

rockyRunnr:
Thanks @MarkWard0110. This PR is a first step toward dynamic KV cache sizing: instead of worst-case upfront allocation, the KV cache grows as needed. Runtime model reload, GPU repartitioning, KV cache migration, and pause/resume generation are interesting directions, but they are out of scope here. This could naturally connect to follow-up work on bounded lazy growth or an OOM fallback policy.

PR #21746, opened 2026-04-10
CUDA: initialize NCCL comms lazily

JohannesGaessler:
Fixes https://github.com/ggml-org/llama.cpp/issues/21692 .
Fixes https://github.com/ggml-org/llama.cpp/issues/21719 .

On master NCCL comms are created for all GPUs eagerly and unconditionally. However, it seems that NCCL comms consume several hundred MB of VRAM per device Also, if only a subset of visible CUDA devices is used this causes the AllReduce to hang indefinitely on master. This PR makes it so that NCCL comms are created lazily and per vector of device ids. The VRAM for them is never freed until the program terminates but this is an issue with the CUDA backend in general.

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: debugging

PR #21761, opened 2026-04-11
CUDA: only init NCCL for setups with multi GPU

### LINKED ISSUE #21759: Misc. bug: Higher VRAM usage after b8738
### Name and Version

b8738 and later

### Operating systems

Linux

libllama (core library), llama-server

### Command line

```shell
./llama-server \                                                                                                                                                                            
    --alias local-ai \                                                                                                                                                                        
    -m /models/google_gemma-4-31B-it-Q4_K_S.gguf \                                                                                                                                            
    --host 0.0.0.0 \                                                                                                                                                                          
    --port 8080 \                                                                                                                                                                             
    -np 1 \                                                                                                                                                                                   
    -ngl 99 \                                                                                                                                                                                 
    -c 131072 \                                                                                                                                                                               
    -n 16000 \                                                                                                                                                                                
    --mmproj /models/mmproj-google_gemma-4-31B-it-bf16.gguf \                                                                                                                                 
    --cache-type-k q4_0 \                                                                                                                                                                     
    --cache-type-v q4_0 \                                                                                                                                                                     
    --cache-ram 2048 \                                                                                                                                                                        
    -ctxcp 2 \                                                                                                                                                                                
    --flash-attn on \                                                                                                                                                                         
    --jinja \                                                                                                                                                                                 
    --chat-template-file /templates/google-gemma-4-31B-it-                                                                                                                                    
  interleaved.jinja \                                                                                                                                                                         
    --no-prefill-assistant \                                                                                                                                                                  
    --chat-template-kwargs '{"enable_thinking":true}' \                                                                                                                                       
    --temp 1.0 \                                                                                                                                                                              
    --top-p 0.95 \                                                                                                                                                                            
    --top-k 64 \                                                                                                                                                                              
    --min-p 0.0 \                                                                                                                                                                             
    --presence-penalty 1.5 \                                                                                                                                                                  
    --repeat-penalty 1.0 \                                                                                                                                                                    
    --metrics
```

On a single RTX 3090, pre b8738 has lower VRAM idle and under load than at b8738+.
Pre b8738 idle VRAM is 22726mb, while in b8738+ idle VRAM increases to 23172mb.
A stress test with tuned full context works pre b8738, and fails on b8738+.

According to investigation by AI, the bad state is that ggml_cuda_init() initiatilzes NCCL communicators even on single GPU runs, which uses uneccessary VRAM. This should be fixed so it only initializes NCCL when there are more than one GPU.

Reproduce by running pre b8738 and measure idle VRAM, and then do the same on b8738+. At least with a single RTX 3090 on this setup.

### First Bad Commit

d6f3030047f85a98b009189e76f441fe818ea44db8738

### Relevant log output

Pre b8738 idle VRAM is 22726mb, while in b8738+ idle VRAM increases to 23172mb.

EldarBorge:
## Overview

Skips NCCL if the setup is single GPU, i.e. only initializes NCCL if GPU count is more than 1.
This reduces VRAM on single GPU setups that was introduced in b8738.

Fixes #21759 

Tested locally and reduces VRAM both idle and under load to pre b8738 levels on my setup.

# Requirements

- AI usage disclosure: Yes, AI was used to bisect the bad commit and to investigate what in that commit caused higher VRAM usage.


JohannesGaessler:
https://github.com/ggml-org/llama.cpp/pull/21746 should be a better fix.

EldarBorge:
You are probably right, your fix is too advanced for me to understand! Let's await @ggerganov , but feel free to close my PR if it's not a good fix. :slightly_smiling_face:

eelgaev:
Is NCCL always initialized now? or is it just for `--split-mode tensor`?

EldarBorge:
> Is NCCL always initialized now? or is it just for `--split-mode tensor`?

always after b8738 is my understanding

eelgaev:
> > Is NCCL always initialized now? or is it just for `--split-mode tensor`?
> 
> always after b8738 is my understanding

But it seems like we don't need NCCL for row/layer splitting, right?
This means we are wasting VRAM on idle NCCL instances?

PR #21764, opened 2026-04-11
ggml: add graph_reused

am17an:
## Overview

Add `reused` member variable to `ggml_cgraph` so backends can take advantage of the graph reuse functionality. Currently when graph_reuse in invoked, the CUDA backend still does the props change check to figure out if the graph has changed or not, where in fact `graph_reuse` (to my understanding) guarantees this to be true. This helps bypass a mildly expensive O(n) check. 

## Additional information

Testing: I tested various combinations like `--n-cpu-moe`, `-nkvo` and `-ngl` and verified it works. Additional testing would be welcome. 

Results on a 5090: 

| Model                    | Test         |   t/s cuda_mul_fused |   t/s ggml_graph_reuse |   Speedup |
|:-------------------------|:-------------|---------------------:|-----------------------:|----------:|
| gemma4 ?B Q4_0           | tg128        |               219.43 |                 231.50 |      1.06 |
| gemma4 ?B Q4_0           | tg128@d16384 |               183.23 |                 191.91 |      1.05 |
| gemma4 ?B Q4_0           | tg128@d32768 |               175.15 |                 182.58 |      1.04 |
| gpt-oss 20B MXFP4 MoE    | tg128        |               313.29 |                 323.01 |      1.03 |
| gpt-oss 20B MXFP4 MoE    | tg128@d16384 |               279.59 |                 288.05 |      1.03 |
| gpt-oss 20B MXFP4 MoE    | tg128@d32768 |               258.80 |                 266.12 |      1.03 |
| qwen35 27B Q4_K_M        | tg128        |                65.04 |                  66.61 |      1.02 |
| qwen35 27B Q4_K_M        | tg128@d16384 |                62.45 |                  63.61 |      1.02 |
| qwen35 27B Q4_K_M        | tg128@d32768 |                59.65 |                  60.70 |      1.02 |
| qwen35moe 35B.A3B Q4_K_S | tg128        |               194.52 |                 206.56 |      1.06 |
| qwen35moe 35B.A3B Q4_K_S | tg128@d16384 |               184.11 |                 196.36 |      1.07 |
| qwen35moe 35B.A3B Q4_K_S | tg128@d32768 |               175.95 |                 187.60 |      1.07 |

# Requirements

- AI usage disclosure: YES, for general understanding of the scheduler


JohannesGaessler (commented):
This would work from a llama.cpp perspective but I'm not convinced it's the right way to handle this from a ggml perspective. In my opinion we should attach a `ggml_guid` and write the code around that. With a GUID the contract for "user code" would be "if you change the graph you must change the GUID", it's not clear to me what the correct way to use a `reused` flag would be.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3060,6 +3060,14 @@ static bool ggml_cuda_graph_update_required(ggml_backend_cuda_context * cuda_ctx
     const void * graph_key = ggml_cuda_graph_get_key(cgraph);
     ggml_cuda_graph * graph = cuda_ctx->cuda_graph(graph_key);
 
+    if (cgraph->version != 0 &&
+        cgraph->version == graph->last_graph_version &&
+        (int)graph->node_props.size() == cgraph->n_nodes) {
```
gaugarg-nv: Will there ever be a case where versions match, but node count doesn't?
If not, maybe we can make this an assert?

am17an:
I added a version number, currently we don't have a guid generator in master. If the current version is acceptable, I can add a guid generator however I think a simple static version counter is also fine imo. The idea of versioning that anything that mutates the cgraph should have increment the version number.

ggml/src/ggml-cuda/ggml-cuda.cu
am17an: Yes will add an assert, it should not happen

ggml/src/ggml-backend.cpp
```diff
@@ -1891,6 +1891,10 @@ enum ggml_status ggml_backend_sched_graph_compute_async(ggml_backend_sched_t sch
         }
     }
 
+    for (int i = 0; i < sched->n_splits; i++) {
+        sched->splits[i].graph.version = graph->version;
```
jeffbolznv: Should all splits really be the same uuid?

ggml/src/ggml-impl.h
```diff
@@ -30,6 +30,11 @@ extern "C" {
 
 void ggml_print_backtrace(void);
 
+static inline uint64_t ggml_graph_next_version(void) {
+    static uint64_t counter = 0;
```
jeffbolznv: This should be atomic

ggml/src/ggml-backend.cpp
am17an: I don't see any additional benefit of having a per split uuid as of now

am17an: I think you're right though, it makes sense to have a per split uuid

JohannesGaessler: Splits should definitely have different identifiers than the main graph.

JohannesGaessler:
I suppose an index that is incremented per graph would also work, even at 1000 new graphs / second it would take like 500 million years until a 64 bit integer is exhausted.

am17an:
> 1000 new graphs / second it would take like 500 million years until a 64 bit integer is exhausted.

I've shortened this time frame by using the top 12 bits for the split index. It would only last ~500 years now. The counter needs to move to a `.c` file though, otherwise each TU will get it's own copy

ggml/src/ggml-impl.h
```diff
@@ -338,6 +340,8 @@ struct ggml_cgraph {
     struct ggml_hash_set visited_hash_set;
 
     enum ggml_cgraph_eval_order order;
+
+    uint64_t version;
```
JohannesGaessler: I would not call it "version" since to me that sounds like something a human programmer would assign based on functionality. My preference would be `id` or `uid`.

ggml/src/ggml.c
```diff
@@ -7125,6 +7132,7 @@ struct ggml_cgraph ggml_graph_view(struct ggml_cgraph * cgraph0, int i0, int i1)
         /*.use_counts       =*/ cgraph0->use_counts,
         /*.visited_hash_set =*/ cgraph0->visited_hash_set,
         /*.order            =*/ cgraph0->order,
+        /*.version          =*/ cgraph0->version,
```
JohannesGaessler: I don't think this is logically correct. If a user were to for example create a view of an existing graph and then execute it with the CUDA backend it would definitely need to be mapped to a different CUDA graph.

JohannesGaessler (commented):
* The function `ggml_backend_graph_optimize` should regenerate the graph identifier. 
* In `ggml-backend.cpp` there are graph manipulations where the sources of nodes are replaced, this should also result in a change in the graph identifier. 
* I think that `ggml-opt.cpp` should be fine with these changes.
* `ggml-backend-meta.cpp` the manually created and re-used graphs `cgraphs_aux` will need to have their identifiers regenerated.

Please let me know if any of these are unclear or if you want me to take over some of these.

am17an:
> In ggml-backend.cpp there are graph manipulations where the sources of nodes are replaced, this should also result in a change in the graph identifier.

Could you point me where? From what I understand bumping the counter only when `sched_reset` is called should be good. Since everything is reset from this point and the graph's id should remain the same till the next `sched_reset`

JohannesGaessler:
Consider [this line](https://github.com/ggml-org/llama.cpp/blob/aa4695c5e5bf0abda8942c08e94cb804a7ea0347/ggml/src/ggml-backend.cpp#L1368). On that line `node` is just a pointer to one of the nodes of the original graph. So before and after calling `ggml_backend_split_graph` the passed `graph` is different and the corresponding identifier should also be different to reflect this.

Also looking at the code again, I think the correct place to set the split graph identifiers would be at the end of `ggml_backend_split_graph` rather than in `ggml_backend_sched_graph_compute_async`.

am17an:
I see, since `ggml_backend_sched_split_graph` is a part of the public API it can be used like you said. So we should also increment the id there (or maybe only there? because I don't see how you can create a graph without calling that function)

JohannesGaessler:
>So we should also increment the id there (or maybe only there?

That would I think be the correct way to do it.

ggml/src/ggml.c
```diff
@@ -53,6 +53,16 @@
 
 #define UNUSED GGML_UNUSED
 
+uint64_t ggml_graph_next_version(void) {
```
JohannesGaessler:
```suggestion
uint64_t ggml_graph_next_uid(void) {
```

Also this function needs an assert to make sure the returned UID is `< 1 << (64 - GGML_SCHED_MAX_SPLIT_BITS)`.

JohannesGaessler (commented):
From my end these changes would be good otherwise.

JohannesGaessler:
>ggml-backend-meta.cpp the manually created and re-used graphs cgraphs_aux will need to have their identifiers regenerated.

Sorry, I had misremembered. At some WIP version I had used `ggml_cgraph` references but at some later point I replaced these with ggml graphs that are properly created from a `ggml_context *` so no changes were necessary after all.

ggml/src/ggml.c
```diff
@@ -53,6 +53,18 @@
...
+    static volatile long long counter = 1;
+    GGML_ASSERT(counter < (1ULL << (64 - GGML_SCHED_MAX_SPLIT_BITS)));
+    return (uint64_t) _InterlockedIncrement64(&counter) - 1;
+#else
+    static uint64_t counter = 1;
+    GGML_ASSERT(counter < (1ULL << (64 - GGML_SCHED_MAX_SPLIT_BITS)));
+    return __atomic_fetch_add(&counter, 1, __ATOMIC_RELAXED);
+#endif
+}
+
```
JohannesGaessler, line 65: From a multithreading perspective this is not quite correct. Conceivably two separate threads could pass the pre-fetch assert, and then separately fetch and increment the counter with one of the threads ending up with a UID above the threshold. I would suggest to just do an assert of the returned value post-fetch.

ggml/src/ggml.c
```diff
@@ -53,6 +53,18 @@
...
+uint64_t ggml_graph_next_uid(void) {
+#ifdef _MSC_VER
+    static volatile long long counter = 1;
+    long long ret = (uint64_t) _InterlockedIncrement64(&counter) - 1;
+    GGML_ASSERT(ret < (1ULL << (64 - GGML_SCHED_MAX_SPLIT_BITS)));
+#else
+    static uint64_t counter = 1;
+    int64_t ret = __atomic_fetch_add(&counter, 1, __ATOMIC_RELAXED);
+    GGML_ASSERT(ret < (1ULL << (64 - GGML_SCHED_MAX_SPLIT_BITS)));
+#endif
```
gaugarg-nv, line 63: no return from the function?

ggml/src/ggml.c
```diff
@@ -7125,6 +7138,7 @@ struct ggml_cgraph ggml_graph_view(struct ggml_cgraph * cgraph0, int i0, int i1)
         /*.use_counts       =*/ cgraph0->use_counts,
         /*.visited_hash_set =*/ cgraph0->visited_hash_set,
         /*.order            =*/ cgraph0->order,
+        /*.version          =*/ 0
```
gaugarg-nv:
```suggestion
        /*.uid              =*/ 0
```

ggml/src/ggml.c
```diff
@@ -7098,6 +7110,7 @@ struct ggml_cgraph * ggml_new_graph_custom(struct ggml_context * ctx, size_t siz
         /*.use_counts   =*/ use_counts_ptr,
         /*.hash_table   =*/ { hash_size, hash_used, hash_keys_ptr },
         /*.order        =*/ GGML_CGRAPH_EVAL_ORDER_LEFT_TO_RIGHT,
+        /*.version      =*/ 0,
```
gaugarg-nv:
```suggestion
        /*.uid          =*/ 0,
```

ggml/src/ggml-backend.cpp
ggerganov: > Splits should definitely have different identifiers than the main graph.

Do you have something in mind? I think if all splits have the same uuid as the main graph, the logic would be simpler. Just not sure if you see something that would prevent doing so.

To clarify, I think that the cuda graph key (i.e. `nodes[0]` ptr) + the common uuid would uniquely identify the splits.

JohannesGaessler: Since we allow arbitrary `--tensor-overrides` we can have multiple splits per backend, no? If a user assigns some tensor in the middle to the CPU you will end up with 2 splits running on the GPU beforehand and afterwards. And if the graphs for those splits have the same UID then they can be incorrectly re-used. 

We could also simplify the logic by not requiring that the UIDs of the split graphs share bits with the original graph. I think that is only really useful for debugging, otherwise we can just call the function to get a UID again.

am17an: In CUDA it will not be an issue because we store the node ptrs as the key, but for other backend it might cause an issue that you arrive at `graph_compute_async` with the same uid as before (e.g. in case of the CPU split), even though it's a different split. I like the split being part of the id as it ties it to a particular graph, as it is not a separate from the main compute graph.

gaugarg-nv:
@am17an  @JohannesGaessler  It will be great to have this PR merged, as it improves perf quite a bit and is also a prerequisite to reducing CPU overhead in the TP implementation. 

My understanding is that this will break CUDA graphs in the Meta backend as we recreate subgraphs in every compute call. This can be fixed if sub-graphs are cached into the meta backend, similar to https://github.com/JohannesGaessler/llama.cpp/pull/14/changes/543e30ddef2d92ba2920897edff0b1a1baf84d01. I am happy to help fix it either within this PR or a separate PR.

am17an:
This is still pending @ggerganov's review, after that we can merge. The TP stuff can be in a follow-up PR

ggml/include/ggml.h
```diff
@@ -232,6 +232,8 @@
 #define GGML_DEFAULT_N_THREADS  4
 #define GGML_DEFAULT_GRAPH_SIZE 2048
 
+#define GGML_SCHED_MAX_SPLIT_BITS 12 // log2(4096)
```
ggerganov: This does not need to be public

ggml/src/ggml-backend.cpp
```diff
@@ -1477,6 +1479,12 @@ void ggml_backend_sched_split_graph(ggml_backend_sched_t sched, struct ggml_cgra
         assert(graph_copy->size > graph_copy->n_leafs);
         graph_copy->leafs[graph_copy->n_leafs++] = leaf;
     }
+
+    // set ids for all splits
+    GGML_ASSERT(sched->n_splits < (1 << GGML_SCHED_MAX_SPLIT_BITS));
+    for (int i = 0; i < sched->n_splits; ++i) {
+        sched->splits[i].graph.uid = graph->uid | ((uint64_t)(i + 1) << (64 - GGML_SCHED_MAX_SPLIT_BITS));
+    }
```
ggerganov, line 1486: Any reason why we just don't generate the next uid here? I don't see the purpose of this bitwise manipulation.

am17an: I initially did it so that we could get the split number back from the uid like a fat pointer, but I don't think that's important considering the complexity. So I will remove that

ggml/src/ggml-impl.h
```diff
@@ -338,6 +340,8 @@ struct ggml_cgraph {
     struct ggml_hash_set visited_hash_set;
 
     enum ggml_cgraph_eval_order order;
+
+    uint64_t uid;
```
ggerganov, line 346:
```suggestion
    // an optional identifier that can be utilized to recognize same graphs
    // value of 0 means it is not set and should be ignored
    // ref: https://github.com/ggml-org/llama.cpp/pull/21764
    uint64_t uid;
```

ggml/src/ggml-cuda/common.cuh
```diff
@@ -1186,6 +1183,7 @@ struct ggml_cuda_graph {
     std::vector<cudaGraphNode_t> nodes;
     bool disable_due_to_gpu_arch = false;
     bool warmup_complete = false;
+    uint64_t last_graph_uid = 0;
```
ggerganov:
```suggestion
    uint64_t uid = 0;
```

ggerganov (approved):

ggml/src/ggml-impl.h
JohannesGaessler: I would also add: "if 2 values > 0 match, the graph is guaranteed to be the same"

JohannesGaessler:
>My understanding is that this will break CUDA graphs in the Meta backend as we recreate subgraphs in every compute call.

This PR will not provide any benefit but it should not break anything because the fallback for a UID mismatch is the behavior on master.

am17an:
@JohannesGaessler can you also approve?

JohannesGaessler (approved):

PR #21570, opened 2026-04-07
HIP: add CDNA4 (gfx950) architecture support for MI350X/MI355X

andyluo7:
Adds gfx950 (MI350X/MI355X, CDNA4) support. These are AMD's latest datacenter GPUs.

gfx950 shares most MFMA instructions with gfx942 (CDNA3), except `mfma_f32_16x16x8_xf32` which isn't available on gfx950 — routed to the f32 fallback path instead.

**Changes:**
- `vendors/hip.h` — add `CDNA4` define for `__gfx950__`
- `common.cuh` — add `GGML_CUDA_CC_CDNA4` constant and `IS_CDNA4` macro
- `mma.cuh` — route CDNA4 MFMA dispatch (xf32→f32 fallback, bf16/int8 same as CDNA3)
- `mmq.cuh` — include CDNA4 in stream-k dispatch

**test-backend-ops:** 5408 OK, 0 FAIL (MI355X gfx950, ROCm 7.0.1)

**llama-bench** (Qwen2.5-1.5B Q4_K_M, single MI355X):
- pp512: 40,013 tok/s
- tg128: 254 tok/s

**Note on FP8:** CDNA4 uses standard e4m3fn (not e4m3_fnuz like CDNA3). The existing CDNA3-specific FP8 path is not extended to CDNA4 — it falls through to the software path. A follow-up could add CDNA4-specific FP8 hardware support.

Relates to #19834 (closed — noted AMD binary builds include gfx950 but the kernel dispatch was missing).

Disclosure: Used an AI assistant to help verify MFMA instruction compatibility across CDNA generations. Code changes and hardware testing done on real MI355X.

PR #19478, opened 2026-02-10
cuda : add RDNA4-specific MMVQ parameter table for bs=1 decode

JoursBleu:
Add a dedicated `MMVQ_PARAMETERS_RDNA4` entry separate from RDNA2/RDNA3. RDNA4 (gfx1201) is wave32-only and has a different memory subsystem, so it benefits from a different MMVQ configuration than RDNA2/RDNA3.

For bs=1 decode on RDNA4, optimal config is `nwarps=8, rows_per_block=1`:
- 8 warps × 32 threads = 256 threads per block
- `blocks_per_iter = vdr * nwarps * warp_size / qi = 2 * 8 * 32 / 4 = 128`
- For K=4096: `blocks_per_row = 128`, the entire K dimension is covered in a single iteration
- This maximizes memory-level parallelism on RDNA4's memory controller

**Benchmark** (Llama 2 7B Q4_0, AMD Radeon AI PRO R9700 / gfx1201):

| Metric | Master | This PR | Change |
|---|---|---|---|
| tg128 (tok/s) | 95.05 | 104.82 | **+10.3%** |
| pp512 (tok/s) | 1449 | 1448 | no regression |

**Correctness:**
- `test-backend-ops -o MUL_MAT`: 1009/1009 OK
- `llama-perplexity` (wikitext-2, 5 chunks): 6.1736 ± 0.40149 (identical to master)
- Text generation (greedy decoding): bit-exact match with master

PR #21774, opened 2026-04-11
HIP: Wave32-optimized RMSNorm and RoPE for RDNA2

carlosfundora:
## Summary

Add RDNA2 (gfx1030/gfx1031) optimized paths for two hot kernels that run every token:

### RMSNorm (`norm.cu`)
- New `rms_norm_f32_wave32` kernel: 128 threads = 4 × Wave32
- Manual warp-shuffle reduce + shared-memory cross-warp reduce
- Avoids the generic `block_reduce<>` path (which uses extern shared mem sized for Wave64)
- Selected at runtime via existing `GGML_CUDA_CC_IS_RDNA2(cc)` check

### RoPE (`rope.cu`)  
- Use 32-thread blocks (1 × Wave32) instead of `CUDA_ROPE_BLOCK_SIZE` on RDNA2
- Aligns wavefront with block dims for optimal occupancy

### Why this matters
RDNA2 (RX 6000 series) uses Wave32 natively. The existing RMSNorm path launches 256 or 1024 threads and relies on `block_reduce<>` with Wave64 assumptions. The existing RoPE uses `CUDA_ROPE_BLOCK_SIZE` which may not align to RDNA2 wavefront boundaries.

Both changes use the existing `GGML_CUDA_CC_IS_RDNA2()` macro — no new preprocessor definitions needed. Falls through to existing code paths on all other architectures (NVIDIA, CDNA, RDNA1/3/4).

### Changes
- `ggml/src/ggml-cuda/norm.cu`: +54 lines (new kernel + dispatch)
- `ggml/src/ggml-cuda/rope.cu`: +5/-2 lines (conditional block size)

### Testing
- Tested on RX 6700 XT (gfx1031, ROCm 6.x)
- `test-backend-ops` passes (norm and rope ops)
- No regression on non-RDNA2 paths (compile-time unchanged, runtime gated)

### Related work
- Follows the pattern of CDNA4 support (#21570) and RDNA4 MMVQ tuning (#19478)
- The `GGML_CUDA_CC_IS_RDNA2` macro already exists in `common.cuh`

**Benchmark request**: Would appreciate community benchmarks on RX 6800 XT / RX 6900 XT to confirm speedup across the RDNA2 family.

### DIFF
--- a/ggml/src/ggml-cuda/norm.cu
+++ b/ggml/src/ggml-cuda/norm.cu
@@ -202,6 +202,59 @@ static __global__ void rms_norm_back_f32(
     }
 }
 
+// Wave32-optimized RMSNorm for RDNA2: 128 threads = 4 × Wave32
+// Uses manual warp-shuffle + shared-memory cross-warp reduce instead of
+// the generic block_reduce<> path (which expects extern shared mem).
+static __global__ void rms_norm_f32_wave32(
+        const float * x, float * dst, const int ncols,
+        const int64_t stride_row, const int64_t stride_channel,
+        const int64_t stride_sample, const float eps) {
+    const int nrows     = gridDim.x;
+    const int nchannels = gridDim.y;
+
+    const int row     = blockIdx.x;
+    const int channel = blockIdx.y;
+    const int sample  = blockIdx.z;
+    const int tid     = threadIdx.x; // [0, 128)
+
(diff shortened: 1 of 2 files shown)


IMbackK:
This dosent make any sense, also you are required to provide evidence why this is a good idea, not others.

PR #21782, opened 2026-04-11
ggml-metal: add Metal kernel for ggml_roll

stephencox-ict:
## Overview

Add a native Metal kernel for `GGML_OP_ROLL` (circular shift). This op currently has no Metal implementation, so it falls back to CPU on Apple Silicon, creating graph splits at every call site.

The Gemma 4 audio conformer uses two `ggml_roll` calls per layer across 12 layers, resulting in 73 graph splits on Metal. With this kernel, all conformer ops stay on the GPU.

The shader follows the same wrapping logic as the CPU implementation: for each element, compute the source index as `(dst_idx - shift) mod dim_size` per dimension.

**Files changed:**
- `ggml-metal-impl.h` - kargs struct
- `ggml-metal-device.m` - register `GGML_OP_ROLL` as supported
- `ggml-metal-device.cpp` - pipeline name mapping
- `ggml-metal-ops.h` - declare dispatch function
- `ggml-metal-ops.cpp` - dispatch function (sets args, binds buffers, dispatches threadgroups)
- `ggml-metal.metal` - `kernel_roll` shader

Ref: #21421 (Gemma 4 audio conformer PR where this was requested)

## Additional information

Tested locally on CPU (the logic matches). Needs validation on macOS Metal via CI or manual testing.

The dispatch uses the same pattern as `kernel_concat`: threadgroups over (ne01, ne02, ne03) with threads iterating over ne00.

# Requirements

- AI usage disclosure: YES - Claude Code was used to help research the Metal kernel patterns in the codebase and draft the implementation. All code was reviewed and verified against the CPU reference implementation.

### DIFF
--- a/ggml/src/ggml-metal/ggml-metal-device.cpp
+++ b/ggml/src/ggml-metal/ggml-metal-device.cpp
@@ -67,6 +67,7 @@ struct ggml_metal_pipeline_with_params ggml_metal_library_get_pipeline_base(ggml
     switch (op) {
         case GGML_OP_ADD_ID: op_str = "add_id"; break;
         case GGML_OP_CONCAT: op_str = "concat"; break;
+        case GGML_OP_ROLL:   op_str = "roll";   break;
         default: GGML_ABORT("fatal error");
     };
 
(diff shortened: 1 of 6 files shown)


stephencox-ict:
@ngxson can you give this a test? I do not have the hardware

ngxson (commented):
`test-backend-ops -o ROLL` passes on my mac

quick question: does it work with other types than F32? IIUC this is the same as CONCAT, but I'm not quite sure how other types are handled or being rejected. the current impl is OK, we don't need support for other types, but just out of curiosity.

PR #21783, opened 2026-04-11
tests: skip broken archs in test-llama-archs

stephencox-ict:
## Overview

Skip architectures in `test-llama-archs` that crash, hang, or produce incorrect results. Found while running the full test suite across CPU and GPU backends during work on #21421 (Gemma 4 audio conformer).

**`save_models()` skips:**
- `CLIP`, `GPTJ` - no usable implementations
- `CHAMELEON` - half-implemented, to be removed
- `RWKV6`, `RWKV6QWEN2`, `RWKV7`, `ARWKV7` - hang or crash
- `BERT` variants, `EUROBERT` - vocab issues
- `PLM` - tensor shape mismatch

**`test_backends()` skips (all of the above, plus):**
- `WAVTOKENIZER_DEC` - CUDA backend crash
- `LLAMA_EMBED`, `GEMMA_EMBEDDING`, `T5ENCODER` - inconsistent results
- `DEEPSEEK2OCR` - tensor shape mismatch

These skips only affect models that already fail. No passing tests are affected.

Ref: #21421

# Requirements

- AI usage disclosure: YES - Claude Code was used to help organize the skip list from test run results.

### DIFF
--- a/tests/test-llama-archs.cpp
+++ b/tests/test-llama-archs.cpp
@@ -429,6 +429,22 @@ static int save_models(const llm_arch target_arch, const size_t seed, const ggml
         if (target_arch != LLM_ARCH_UNKNOWN && arch != target_arch) {
             continue;
         }
+        if (arch == LLM_ARCH_CLIP || arch == LLM_ARCH_GPTJ) {
+            continue; // no usable implementations
+        }
+        if (arch == LLM_ARCH_CHAMELEON) {
+            continue; // half-implemented, to be removed
+        }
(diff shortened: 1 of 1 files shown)


JohannesGaessler:
I don't understand the purpose of this PR. These are just copies of the logic in `arch_supported` which is already used to skip these test cases but with a clear indication that that has been done.

stephencox:
> I don't understand the purpose of this PR. These are just copies of the logic in `arch_supported` which is already used to skip these test cases but with a clear indication that that has been done.

Okay, closing

PR #21787, opened 2026-04-12
vulkan: fix output corruption on GCN 2.0/3.0 (Vulkan 1.2)

rafikb:
## Overview

This PR resolves output corruption (token repetition/double-wording) on legacy AMD GCN 2.0 and 3.0 hardware (e.g., R9 390, Fury X).

The issue was introduced in b8394 due to the implementation of modern Vulkan synchronization primitives that are not fully supported by the final legacy Adrenalin drivers (Vulkan 1.2.170). By restoring backward-compatible memory barrier logic in ggml-vulkan.cpp, we restore coherent generation for these cards without regressing performance on modern Vulkan 1.3+ hardware.

## Additional information

- Hardware Tested: Radeon R9 390 (8GB VRAM) on Windows 10.

- Performance: Verified 22.0 t/s using gemma-4-E4B-it-Q4_K_M.gguf with 100% GPU offload.

- Context: Many legacy 8GB cards are still highly capable for inference in 2026, but are limited by driver-level Vulkan support. This fix ensures continued accessibility for users on vintage or budget-restricted hardware.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -191,7 +191,6 @@ struct vk_queue;
 
 struct vk_command_buffer {
     vk::CommandBuffer buf;
-    uint64_t use_counter = 0;
     bool in_use = false;
 };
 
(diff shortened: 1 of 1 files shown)


0cc4m:
Your hardware is not limited by the driver, on Linux you'd still have full support and the existing code would work. This is not a change we can accept, you are just reverting #20518.

emansom:
Install a recent kernel and recent Mesa, it should then default to AMDGPU on that hardware and work.

If for some reason the Vulkan behavior is indeed different, contribute a backwards compatibility to Mesa instead of here.

szmania:
> Your hardware is not limited by the driver, on Linux you'd still have full support and the existing code would work. This is not a change we can accept, you are just reverting #20518.

Do  you not support Win11? I'm in the same boat as the OP and would love to see this fixed. R9 390 8 GB VRAM on Windows 11.

0cc4m:
Windows 11 is supported. If the current version is not working for you, then your driver is broken. You have to use a working driver to run the backend. The only driver that still provides updates and fixes for your card is Linux RADV.

szmania:
> Windows 11 is supported. If the current version is not working for you, then your driver is broken. You have to use a working driver to run the backend. The only driver that still provides updates and fixes for your card is Linux RADV.

So your solution is to migrate to Linux with a Vulkan?

0cc4m:
That's the easiest way. If you find a solution for your driver problem in the Vulkan backend that does not involve making the backend worse for other GPUs, that's fine, too. But this PR isn't it.

szmania:
For future reference to anyone else getting gibberish or repeating characters with Vulkan cards, rolling back to https://github.com/ggml-org/llama.cpp/releases?q=b8393&expanded=true fixes this.

PR #21793, opened 2026-04-12
server: (anthropic API) fix prefix caching

kvc0:
## Overview

When testing claude code against llama.cpp, I noticed that only
n_past 18577 was used even when context was 60k or more. The log
in llama-server says:
```
slot update_slots: id  3 | task 10342 | old: ... ; cch= | defa0;You are
slot update_slots: id  3 | task 10342 | new: ... ; cch= | 1c8b4;
```
I observed that the cch value changed every time. Reading about that,
the x-anthropic-billing-header system message seems to be specially
handled inside of the anthropic api. I could remove it, but there
is a meaningful string sometimes included at the end. So instead,
I just replace the changing cch checksum with fffff.

I'm treating this as an anthropic message body API detail - I think this
is the right way to do this, but by all means please correct me!

It's always 5 hexadecimal characters, but I've written the replacement
defensively in case they change the protocol.

## Additional information

When asking "explain this repo to me on a different repo," using a freshly started llama-server, the second request:
**Before:**
```
selected slot by LCP similarity, sim_best = 0.566 (> 0.050 thold), f_keep = 0.704
```
This is the best case, but it gets progressively worse as the matched length never
goes longer than 18577 (up to 18580 theoretically, but I never saw higher than 18578).

**After:**
```
selected slot by LCP similarity, sim_best = 0.805 (> 0.050 thold), f_keep = 1.000
```

And further along, I see prefixes that only differ in tool call details, as you would expect:
```
selected slot by LCP similarity, sim_best = 0.994 (> 0.050 thold), f_keep = 0.999
[...]
slot update_slots: id  1 | task 449 | old: ... =command>
 | cd /home/kenny/g
slot update_slots: id  1 | task 449 | new: ... =command>
 | git status
</parameter>
```
After this change, similarity looks normal and caching is performing well.

While debugging this, I dumped the /slots api a couple times on subsequent requests.
The diffs in the prompt field were like:
```
diff prompt1 prompt2 --unchanged-line-format="" --old-line-format="< :%dn: %L" --new-line-format="> :%dn: %L"
< :62: x-anthropic-billing-header: cc_version=2.1.101.e51; cc_entrypoint=cli; cch=a5145;You are Claude Code, Anthropic's official CLI for Claude.
> :62: x-anthropic-billing-header: cc_version=2.1.101.e51; cc_entrypoint=cli; cch=4a1a8;You are Claude Code, Anthropic's official CLI for Claude.
> :5130: </tool_response><|im_end|>
> :5131: <|im_start|>assistant
> :5132: <think>
> :5133: The diagnostics still show an issue with [...]
```
You can see line 62 has a cch diff, and then over 5000 common lines before the diff.
This should have been a total cache hit because it's all new starting at line 5130. But
because of the line 62 diff, it had to re-ingest nearly the whole thing. Without this
change, llama-server does this on every request because of anthropic's magic "header."

**Performance:**
The impact of this change to users who aren't using claude to send messages to the
anthropic api is a single-position O(1) string prefix check per system message. I don't
imagine too many system messages start with `x` so in the usual case it will early out
at 1 character's worth of comparison.

# Requirements

- AI usage disclosure: NO. I read and wrote all of this myself.

### DIFF
--- a/tools/server/server-chat.cpp
+++ b/tools/server/server-chat.cpp
@@ -281,6 +281,42 @@ json server_chat_convert_responses_to_chatcmpl(const json & response_body) {
     return chatcmpl_body;
 }
 
+// Edits the cch section of an "x-anthropic-billing-header" system prompt.
+// Does nothing to any other prompt.
+//
+// This is a claude message with a "cch=ef01a" attribute that breaks prefix caching.
+// The cch stamp is a whitebox end-to-end integrity hint. It's not meaningful as a
+// system prompt data, particularly to llama.cpp, but its presence means the prefix
+// cache will not get past it: It changes on each request.
+//
+// Reference: https://github.com/ggml-org/llama.cpp/pull/21793
+// Example header:
+// ```
+// x-anthropic-billing-header: cc_version=2.1.101.e51; cc_entrypoint=cli; cch=a5145;You are Claude Code, Anthropic's official CLI for Claude.
+//                                                                            ^^^^^
+// ```
+static void normalize_anthropic_billing_header(std::string & system_text) {
+    if (system_text.rfind("x-anthropic-billing-header:", 0) != 0) {
+        return;
+    }
+
+    const size_t header_prefix_length = strlen("x-anthropic-billing-header:");
+    const size_t cch_length = 5;
+    const size_t index_cch = system_text.find("cch=", header_prefix_length);
+    if (index_cch == std::string::npos) {
+        return;
+    }
+
+    const size_t index_replace = index_cch + 4;
+    if (index_replace + cch_length < system_text.length() && system_text[index_replace + cch_length] == ';') {
+        for (size_t i = 0; i < cch_length; ++i) {
+            system_text[index_replace + i] = 'f';
+        }
+    } else {
+        LOG_ERR("anthropic string not as expected: %s", system_text.c_str());
+    }
+}
+
 json server_chat_convert_anthropic_to_oai(const json & body) {
     json oai_body;
 
@@ -292,10 +328,13 @@ json server_chat_convert_anthropic_to_oai(const json & body) {
 
         if (system_param.is_string()) {
             system_content = system_param.get<std::string>();
+            normalize_anthropic_billing_header(system_content);
         } else if (system_param.is_array()) {
             for (const auto & block : system_param) {
                 if (json_value(block, "type", std::string()) == "text") {
-                    system_content += json_value(block, "text", std::string());
+                    auto system_text = json_value(block, "text", std::string());
+                    normalize_anthropic_billing_header(system_text);
+                    system_content += system_text;
                 }
             }
         }


ngxson:
Please address issues from automated review: https://github.com/ngxson/llama.cpp/pull/98

ngxson (approved):

tools/server/server-common.h
```diff
@@ -347,6 +347,9 @@ std::string format_oai_resp_sse(const json & data);
 // format Anthropic-style SSE with event types
 std::string format_anthropic_sse(const json & data);
 
+// re-format an x-anthropic-billing-header system prompt's cch section for prompt caching friendliness
+void normalize_anthropic_billing_header(std::string & system_text);
```
ngxson: this function is never used outside of the file, it should not be in the header

kvc0: Okay - I got a warning from cmake that it wasn't declared first. I'll just declare it in the cpp then! I don't see a pattern for pre-declaring at the top of the file, so I'll just put the declaration next to the helper :-)

kvc0: fixed, thank you

drrros:
Tried this branch and it feels it's less recomputations now

kvc0:
@ngxson Thank you for tolerating a ping - am I missing anything to get this merged? I'd love to address any gaps; this is my first contribution to llama.cpp, so if there's a process I missed or something please let me know!

drrros:
PR have no conflicts yet, so it seems applying patch to up to date master is the way for now :( hope this will be merged at some point.

ngxson:
@ServeurpersoCom can you give the 2nd approval? Thanks!

drrros:
@ServeurpersoCom  PTAL, this is very useful with claude code

ServeurpersoCom (approved):

ServeurpersoCom:
I had a week without GitHub notifications due to a bad configuration on my side you did well to notify me.

ServeurpersoCom:
There's now a conflict on tools/server/server-common.cpp from master moving, a rebase should unblock the merge.

kvc0:
@ServeurpersoCom A rebase wasn't quite clean. I just did a manual migration to master a half hour ago, and I'm testing now to ensure I got the new location right. It was just 1 helper function and like 5 lines of diff in the newly-moved code. It is doing the right thing, so I'll update soon! Thanks for your patience <3

ServeurpersoCom:
Thanks for handling the manual migration! Ping me when it's up and I'll take another look!

kvc0:
@ServeurpersoCom I don't like force-pushing after people have reviewed, but I just migrated the change from the place it went yesterday to the place it goes today.

ServeurpersoCom (approved):
Force-pushing is the right call here, it's just a code relocation, not a logic change. Re-approving now:)

ServeurpersoCom:
(CI errors are independent of this PR)

PR #21797, opened 2026-04-12
webui: add setting for first-line chat titles

crodjer:
## Overview
Add an opt-in setting (`titleGenerationUseFirstLine`) to use the first non-empty line of a prompt as the generated conversation title.

## Additional information
At the moment, the complete multi-line prompt is being used, which created long titles for complex prompts. Coupled with
"Ask for confirmation before changing conversation title", the dialog would overflow.

With the new flag disabled (default):
Screenshot 2026-04-12 at 13-59-41 Test Test This is multiline  - llama cpp
With the new flag enabled:
Screenshot 2026-04-12 at 14-00-22 Test Test - llama cpp

# Requirements
Nothing that I know of, just running the frontend would work.

- AI usage disclosure: Yes, Gemini CLI as svelte and front-end work isn't my expertise. The code changes look minimal and reasonable to me with self review. Based on my going through it, the changes follow similar structure and style as the rest.


tools/server/webui/src/lib/utils/text.ts
```diff
@@ -5,3 +5,15 @@
...
 	return content.length > max ? content.slice(0, max) + '...' : content;
 }
+
+/**
+ * Generates a single-line title from a potentially multi-line prompt.
+ * Uses the first non-empty line if `useFirstLine` is true.
+ */
+export function generateConversationTitle(content: string, useFirstLine: boolean = false): string {
+	if (useFirstLine) {
+		const firstLine = content.split('\n').find((line) => line.trim().length > 0);
```
allozaur: Please use `NEWLINE_SEPARATOR` constant

```suggestion
		const firstLine = content.split(NEWLINE_SEPARATOR).find((line) => line.trim().length > 0);
```

tools/server/webui/src/lib/utils/text.ts
```diff
@@ -5,3 +5,15 @@
...
+		return firstLine ? firstLine.trim() : content.trim();
+	}
+	return content.trim();
```
allozaur, line 21:
```suggestion
	}

	return content.trim();
```

allozaur (changes requested):
Additionally, please run `npm run build` and commit updates in `tools/server/public`

allozaur (approved):

ServeurpersoCom (approved):

tools/server/webui/src/lib/utils/text.ts
```diff
@@ -5,3 +5,16 @@
...
 	return content.length > max ? content.slice(0, max) + '...' : content;
 }
+
+/**
+ * Generates a single-line title from a potentially multi-line prompt.
+ * Uses the first non-empty line if `useFirstLine` is true.
+ */
+export function generateConversationTitle(content: string, useFirstLine: boolean = false): string {
+	if (useFirstLine) {
+		const firstLine = content.split(NEWLINE_SEPARATOR).find((line) => line.trim().length > 0);
```
allozaur, line 17: we're missing an import for this constant here ;)

Image

allozaur (changes requested):

tools/server/webui/src/lib/utils/text.ts
crodjer: Fixed. I added that from the the git review suggestion and did not run it locally. Apologies.

tools/server/public/bundle.js
allozaur, line 1: Hey @crodjer we still need to have the static output re-built, the merge commit only introduces the build from `master` branch

allozaur (changes requested):

tools/server/public/bundle.js
crodjer: I did rebuild it I think. `TITLE_GENERATION_USE_FIRST_LINE` exists in the bundle.js. I can do it again to verify.

crodjer: `bun run build` resulted in no changes @allozaur .

allozaur: ah, my bad in this case! so i guess that means that the build ran on pre-commit hook on your end. sorry for the confusion :)

allozaur (approved):

PR #21815, opened 2026-04-12
server : reinit speculative ngram state after context shift to fix GGML_ABORT

jonpojonpo:
## Problem

While using speculative decoding i kept encountering numerous crashes, after investigating with AI I managed to resolve them with the attached changes. I'm not at all knowledgeable about cpp but am hoping this can help others.  

Using `--context-shift` together with any `--spec-type ngram-*` mode causes a hard crash (`Aborted`) mid-generation:

```
GGML_ABORT("map.idx_last_check > cur_len: %zu > %zu", __func__, map.idx_last_check, cur_len)
```
in `common/ngram-map.cpp:236` (and a similar one at line 251 for `size_last_begin`).

**Root cause:** when context shift truncates `slot.prompt.tokens` by `n_discard` tokens, the ngram map's internal position counters (`idx_last_check`, `size_last_begin`) are not updated — they still hold the pre-shift (larger) values. The very next call to `common_speculative_draft()` passes the now-shorter prompt and trips the invariant check.

The comment in `ngram-map.cpp` reads *"Should not happen because of `common_ngram_map_begin()`"* — and that's exactly the issue: `common_ngram_map_begin()` is only called at the start of generation, never after an in-flight context shift.

## Fix

After context shift completes (prompt truncated, KV cache updated), call `common_speculative_begin()` on the slot's speculative state. The existing `begin()` logic already handles a shrinking prompt: it prunes map entries with indices beyond the new prompt length and resets the position counters.

## Reproducer

```bash
llama-server -m <model> --ctx-size 4096 --context-shift --spec-type ngram-map-k4v --draft 16
# send a request that generates past the context limit → Aborted
```

Affects all `ngram-map-k`, `ngram-map-k4v`, `ngram-simple`, `ngram-cache` spec types combined with `--context-shift`.

---

> **Note:** this fix was diagnosed and written with AI assistance (Claude Code). I'm disclosing this upfront as I know the project prefers transparency on AI-generated contributions. The logic has been reviewed and the fix verified with a rebuild. Happy to iterate if the maintainers prefer a different approach (e.g. fixing the abort in `ngram-map.cpp` itself to be a graceful reset instead).


jonpojonpo:
Hi, I hope this is ok its way outside my normal skill set. I ran into these issues while trying to get a high inference speed.

Issue #15923, opened 2025-09-10
Eval bug: Qwen2.5-Omni outputs GGGGG... when fully offloaded to GPU on Jetson AGX Orin


shakez0901:
### Name and Version

$ ./build/bin/llama-mtmd-cli --version
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 1 CUDA devices:
  Device 0: Orin, compute capability 8.7, VMM: yes
version: 6316 (009b709d)
built with cc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0 for aarch64-linux-gnu

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

Jetson AGX Orin 64GB

### Models

Qwen2.5-Omni-7B

- GGUFs from [ggml-org/Qwen2.5-Omni-3B-GGUF](https://huggingface.co/ggml-org/Qwen2.5-Omni-7B-GGUF)

as sanity check, also the BF16 GGUF from https://huggingface.co/unsloth/Qwen2.5-Omni-7B-GGUF and Q4_K_M quants of the smaller 3B model from https://huggingface.co/unsloth/Qwen2.5-Omni-3B-GGUF and https://huggingface.co/ggml-org/Qwen2.5-Omni-3B-GGUF

### Problem description & steps to reproduce

**Build commands:**
```
cmake -B build -DGGML_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES="87"
cmake --build build --config Release -- -j
```
**Problem:**
I am receiving gibberish output, specifically a chain of Gs: "GGGGGGGGGG...".
But interestingly, this only starts happening if the entire model is offloaded to the GPU and if there is audio input present. Text only works fine, as opposed to similar issues like #15556 or #15034, where text only already broke.

Smaller values for -ngl did not lead to this behavior.

Furthermore, the bug seems to be related to the Qwen2.5-Omni model itself, as it happened on all the above-mentioned models, but other models like Ultravox and Voxtral-Small seemed to run fine

I first noticed the issue with requests to llama-server, but with llama-mtmd-cli I could reproduce it in a simpler way:
```
./llama-mtmd-cli -hf ggml-org/Qwen2.5-Omni-7B-GGUF:Q8_0 -ngl 99 -p "What is being said here?" --audio /home/nvidia/samples/who.mp3
```
[...]

Issue #15034, opened 2025-08-02
Eval bug: Broken/no Gemma 3n output on CUDA (Nvidia Jetson Orin Nano)


ai-fonsi:
### Name and Version

```
./llama-cli --version
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 1 CUDA devices:
  Device 0: Orin, compute capability 8.7, VMM: yes
version: 6060 (9c35706b)
built with cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 for aarch64-linux-gnu
```

### Operating systems

Linux

### GGML backends

CUDA

### Hardware

Nvidia Jetson Orin Nano

### Models

https://huggingface.co/lmstudio-community/gemma-3n-E4B-it-text-GGUF (Q4_K_M), but other quants seem to behave the same way.

### Problem description & steps to reproduce

`./llama-cli -m ../../../../models/gemma-3n-E4B-it-Q4_K_M.gguf -ngl 999` prints either nothing or garbled text.
`./llama-cli -m ../../../../models/gemma-3n-E4B-it-Q4_K_M.gguf --device none` works as expected.

Prompts with GPU offloading print nothing most of the time, but GPU utilization is almost 100% and `llama_perf_context_print` shows that it did generate tokens. Other models such as Gemma 3 4B work fine on my device.

### Relevant log output

```shell
$ ./llama-cli -m ../../../../models/gemma-3n-E4B-it-Q4_K_M.gguf -ngl 999
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 1 CUDA devices:
  Device 0: Orin, compute capability 8.7, VMM: yes
build: 6060 (9c35706b) with cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 for aarch64-linux-gnu
main: llama backend init
main: load the model and apply lora adapter, if any
llama_model_load_from_file_impl: using device CUDA0 (Orin) - 6587 MiB free
llama_model_loader: loaded meta data with 39 key-value pairs and 847 tensors from ../../../../models/gemma-3n-E4B-it-Q4_K_M.gguf (version GGUF V3 (latest))
[...]

PR #12552, opened 2025-03-24
llama-map to support hugepage feature of pagesize 2M or 1G which can …

nickhuang99:
…greatly speedup loading huge model when system has enough RAM to pre-allocate hugetlbfs model file

*Make sure to read the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md) before submitting a PR*

As using argument is too much code change and also mmap with hugepage support not available in all platform, so, I decide to use compiler option to switch on this feature. i.e. cmake .. -DGGML_USING_HUGE_PAGE_1G=1
The hugepage feature requires model file must exist in hugetlbfs, and also Linux kernel command to boot with designate hugepagesz to be set. Also "copy" normal model file to hugetlbfs mount point requires also using a small tool with mmap. Therefore this is considered to be an advanced feature.

Issue #12444, opened 2025-03-18
Feature Request: allow mmap to take advantage of hugepage feature which has 10x speedup


nickhuang99:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

using Linux kernel "hugepagetlbfs" has more than 10x speedup for extremely large model, i.e. DeepSeek-r1:671b of size of 400G. And there is very minimum code change(less than 10 lines see below). 
Here is kernel doc about [hugepage](https://docs.kernel.org/admin-guide/mm/hugetlbpage.html)
and I have tested at my ubuntu 22.04 with a second-hand Dell server of 1.5T memory without any GPU to load and run DS huge model. The performance of "hugetlbfs" is significant.

### Motivation

Linux kernel allows user to take advantage of [huge page of RAM](https://docs.kernel.org/admin-guide/mm/hugetlbpage.html). This is highly efficient when loading extremely large model. In my experiment of DeepSeek-r1:671b of size 377G, the speedup is 10x times! And the code change is really minimum based on current implementation of "llama-mmap.cpp". The only missing thing is just page size alignment and "mmap" flag "MAP_HUGETLB".  
I have a [rough video](https://www.youtube.com/watch?v=iXcy9PONAno) to explain what is done . 

Here is steps of using hugepage(my OS is ubuntu 22.04):

1. using sysctl to allocate hugepage, depending how physical memory you have, in my case, I reserved 800G RAM: i.e. sudo sysctl -w vm.nr_hugepages=400000
2. check it is done: cat /proc/meminfo | grep -i huge
3. creating mount point of "hugetlbfs": sudo mkdir /mnt/hugepages && sudo mount -t hugetlbfs -o uid=1000,gid=1000,rw none /mnt/hugepages
4. using a small program "mmap" to "copy" model file to  the mount point with source code below

<pre><code>
#include &lt;sys/mman.h&gt;
#include &lt;stdio.h&gt;
#include &lt;errno.h&gt;
#include &lt;string.h&gt;
#include &lt;sys/types.h&gt;
#include &lt;sys/stat.h&gt;
[...]

PR #21821, opened 2026-04-12
llama : add --hugepages for HugeTLB-backed weight loading (Linux)

doctorjei:
## Overview

Addresses #2251 (partially — weights via `--mmap` path only; I hope to address `--no-mmap` and KV cache in a follow-up). I realize this is a non-trivial change (even though the implementation itself it just one file / two blocks), so I'm going into a bit more detail than I usually would with a PR. :)

### Summary

Adds a new `--hugepages` CLI flag (env: `LLAMA_ARG_HUGEPAGES`) that backs model weight memory with anonymous 2 MiB HugeTLB pages on Linux.

**Motivation: vmemmap reclamation**
(not TLB speedup,v though it may facilitate future work in this area)

When HugeTLB Vmemmap Optimization (HVO, `CONFIG_HUGETLB_PAGE_OPTIMIZE_VMEMMAP=y`) is enabled, the kernel frees the per-4 KiB `struct page` metadata within each hugepage. On a 128 GiB system this recovers **~1.75 GiB** of kernel memory — enough to turn a tight-ceiling workload from OOM into working (which is what was happening to me).

The flag is opt-in at runtime, so there's no cost to anyone who doesn't use it.

```bash
# reserve the pool (runtime, no reboot for 2 MiB pages)
sudo sysctl -w vm.nr_hugepages=65536   # 65536 x 2 MiB = 128 GiB

# run with the flag
./llama-cli --hugepages -m model.gguf ...
```

### Why?

Each 4 KiB page costs ~64 bytes of `struct page` metadata. With 128 GiB, that's ~2 GiB just to track pages. HugeTLB with HVO reduces this; a 2 MiB hugepage is only 4 KiB instead of 32 KiB.

| System RAM | Approx. Benefit |
|---|---|
| 16 GiB | ~224 MiB |
| 64 GiB | ~896 MiB |
| 128 GiB | ~1.75 GiB |

Here's my real-life example: Strix APU with 128 GiB RAM (unified in my case) running MiniMax m2.5 IQ4_XS. The total footprint ~127,910 MiB against ~127,342 MiB available ("normal" pages). Saving 1,792 MiB pushes available memory to ~129,134 MiB, so now the model fits with ~1 GiB to spare.

Transparent Huge Pages (`MADV_HUGEPAGE`) don't help here, as THP remaps existing 4 KiB pages under a 2 MiB entry for TLB efficiency, but the `struct page` array stays intact (and saves no memory). Only explicit HugeTLB pool allocation with HVO addresses this.

~~I do want to be clear that currently this only works with CPU inference; that's because I haven't tackled hipMalloc. (That would open the door to directly passing memory to the GPU without reallocating.) I also am unsure of how this would work with other unified memory systems.~~
As of the latest commit, this works not only with CPU inference, but also with HIP / ROCm via ***buffer_from_host_ptr***. Vulkan support should also be possible but will require a bit more digging. NVIDIA support is a bit more difficult, as I don't have an NVIDIA integrated chipset, and there is at least one known issue with the Jetson and this approach (https://github.com/ggml-org/llama.cpp/issues/15034, https://github.com/ggml-org/llama.cpp/issues/15923).

### Approach

In #2251, @slaren suggested adding `MAP_HUGETLB` to the existing `mmap` call. @qdacsvx noted the kernel rejects `MAP_HUGETLB` on regular descriptors (`EINVAL`). This PR implements what @slaren had in mind:

1. `llama_mmap` allocates an anonymous region with `MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB | MAP_HUGE_2MB | MAP_POPULATE`
2. `load_all_data` populates the region per-tensor via `file->seek` + `file->read_raw` (same pattern as the existing `--no-mmap` branch)
3. After all tensors are loaded, `mprotect` downgrades to `PROT_READ`

`MAP_POPULATE` forces atomic pool allocation at mmap time, so if the pool is insufficient, it yields a clean `ENOMEM` (with a friendly diagnostic including `sysctl` guidance), not a SIGBUS during load.

Arg-parse rejects `--hugepages` combined with `--no-mmap` (wording points at a followup PR), `--direct-io` (the loader's existing conflict resolver would silently bypass our code), and non-Linux platforms.

Anonymous mapping is the only path to HugeTLB-backed weight memory without requiring a hugetlbfs mount, copying the model, and recompiling, a la *PR #12521  #12552* (issue #12444), which isn't especially user-friendly. This PR avoids all of that — just `sysctl` + `--hugepages`.

*PR #7420* also used anonymous mappings inside `llama_mmap`, but for direct-I/O bypass rather than hugepage backing. A concern raised there was that anonymous memory can swap under pressure. That doesn't apply here, since HugeTLB pages are inherently pinned by the kernel and cannot be swapped, reclaimed, or migrated regardless of memory pressure or `RLIMIT_MEMLOCK` settings. The can be relinquished / reused, though.

### Tradeoffs

Warm-load time increases because `--mmap` normally shares page-cache pages zero-copy, while `--hugepages` must `read_raw` file into the anonymous region.

Measured on qwen3-235B-Q4 (19.4 GB), Strix Halo, Linux 6.17:

| Path | Cold | Warm |
|---|---|---|
| `--mmap` baseline | 3843 ms | 547 ms |
| `--hugepages` | 5030 ms | 2319 ms |
| Slowdown | 1.31x | 4.24x |

The warm 4.24x is driven by Strix Halo's `copy_to_user` bandwidth (~8.5 GB/s on LPDDR5X). Other platforms may see less. At the 128 GiB target scale,
warm-load adds ~11 seconds per session — in exchange for the model actually fitting.

### Changes

There are three primary changes (in these three commits):
- Adding the CLI parameter `--hugepages` (and LLAMA_ARG_HUGEPAGES)
- Modifying function signatures & data structure to accommodate hugepages tracking
- Implementation of the hugepages allocation (llama-mmap.cpp)

### Testing

- Builds clean on Linux x86_64, GCC 14.2, both CPU-only and HIP/ROCm configurations                                                                                
- End-to-end verified on Strix Halo (gfx1151, ROCm 7.2.0, Linux 6.17):
    model loads with `--hugepages`, inference produces correct output                                                                                                
    (39 t/s prompt, 8.2 t/s generation on qwen3 19.4 GB)
- `--help` shows the flag with env var
- Arg-parse rejections verified (`--hugepages --no-mmap`, `--hugepages -dio`)
- Standalone harness validated the core mechanism on Strix Halo with a 19.4 GB GGUF:
  - Pool accounting exact (9424 x 2 MiB pages consumed/restored)
  - `VmRSS` confirms vmemmap reclamation: 3872 kB under `--hugepages` vs 19.30 GB baseline
  - No SIGBUS under `MAP_POPULATE`
  - `read_raw` EINTR/short-read handling inherited from existing code

### HVO verification

For users to confirm HVO is active:

```bash
# compiled in?
grep CONFIG_HUGETLB_PAGE_OPTIMIZE_VMEMMAP /boot/config-$(uname -r)
# → CONFIG_HUGETLB_PAGE_OPTIMIZE_VMEMMAP=y

# enabled at runtime?
cat /proc/sys/vm/hugetlb_optimize_vmemmap
# → 1  (if 0: sudo sysctl -w vm.hugetlb_optimize_vmemmap=1)

# pool state
grep -i huge /proc/meminfo
```

### Follow-ups (future work / not this PR)

- **`--no-mmap` path + KV cache + compute buffers** via a new `ggml_backend_cpu_hugetlb_buffer_type` (parallel to the existing HBM pattern). Reuses
the `--hugepages` flag.
- Multi-threaded `read_raw` for load parallelism
- `posix_fadvise(POSIX_FADV_DONTNEED)` on the source page cache after load
- `buffer_from_host_ptr` is hardcoded `false` for CUDA/HIP (`ggml-cuda.cu:4710`). This PR is forward-compatible with a future flip for zero-copy on unified-memory APUs.

cc @ggerganov @slaren — per prior comments in #2251, I was hoping you could let me know what your take is on this approach. I know this issue matters for me. I've done my best to simplify / minimize code changes, but I am always happy to reconsider my approach as needed. (I hope it's OK to tag you; sorry if I missed a policy against it.)

# Requirements

- AI usage disclosure: YES, AI was used in the process of preparing these commits.

AI was used to identify the appropriate strategy, draft a harness, and draft initial code snippets. Every line was reviewed, edited as appropriate, and included in commits (with sections of code separated manually into different commits with specific focus to ensure clarity and appropriate review (e.g., CLI parameter setup, data structures / call signatures, and final implementation were all separated into separate commits during review and manual processing.)

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -574,6 +574,21 @@ static bool common_params_parse_ex(int argc, char ** argv, common_params_context
         throw std::invalid_argument("error: --prompt-cache-all not supported in interactive mode yet\n");
     }
 
+    // --hugepages compatibility checks. PR #1 covers weights on the --mmap path only;
+    // --no-mmap and --direct-io both route around the hugetlb branch, so surface the
+    // conflict at the user rather than silently ignoring the request. Linux-only.
+    if (params.use_hugepages) {
+#ifndef __linux__
+        throw std::invalid_argument("error: --hugepages is Linux only.\n");
+#endif
+        if (!params.use_mmap) {
+            throw std::invalid_argument("error: --hugepages requires --mmap. Coverage for --no-mmap is added in a follow-up PR. Drop --no-mmap to use --hugepages.\n");
+        }
+        if (params.use_direct_io) {
+            throw std::invalid_argument("error: --hugepages and --direct-io are incompatible. Drop one.\n");
+        }
+    }
+
     // handle model and download
     if (!skip_model_download) {
         auto res = common_params_handle_model(params.model, params.hf_token, params.offline);
@@ -2225,6 +2240,15 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             params.use_mmap = value;
         }
     ).set_env("LLAMA_ARG_MMAP"));
+    add_opt(common_arg(
+        {"--hugepages"},
+        "back model weights with anonymous hugetlb 2 MiB pages (Linux only).\n"
+        "reserve the pool first with e.g. `sysctl -w vm.nr_hugepages=N` — no reboot required.\n"
+        "pinned in RAM, bypasses swap and page cache; primary win is vmemmap reclamation via HVO",
+        [](common_params & params) {
+            params.use_hugepages = true;
+        }
+    ).set_env("LLAMA_ARG_HUGEPAGES"));
     add_opt(common_arg(
         {"-dio", "--direct-io"},
         {"-ndio", "--no-direct-io"},
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -1421,6 +1421,7 @@ struct llama_model_params common_model_params_to_llama(common_params & params) {
     mparams.tensor_split    = params.tensor_split;
     mparams.use_mmap        = params.use_mmap;
     mparams.use_direct_io   = params.use_direct_io;
+    mparams.use_hugepages   = params.use_hugepages;
     mparams.use_mlock       = params.use_mlock;
     mparams.check_tensors   = params.check_tensors;
     mparams.use_extra_bufts = !params.no_extra_bufts;
--- a/common/common.h
+++ b/common/common.h
@@ -534,6 +534,7 @@ struct common_params {
     bool input_prefix_bos  = false; // prefix BOS to user inputs, preceding input_prefix
     bool use_mmap          = true;  // enable mmap to use filesystem cache
     bool use_direct_io     = false; // read from disk without buffering
+    bool use_hugepages     = false; // back weight mappings with anonymous hugetlb pages (Linux only)
     bool use_mlock         = false; // use mlock to keep model in memory
     bool verbose_prompt    = false; // print prompt tokens before generation
     bool display_prompt    = true;  // print prompt before generation
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -606,6 +606,8 @@ ggml_backend_cuda_context::~ggml_backend_cuda_context() {
 struct ggml_backend_cuda_buffer_context {
     int device;
     void * dev_ptr = nullptr;
+    void * host_ptr = nullptr;
+    bool   owned    = true;
     std::string name;
 
     ggml_backend_cuda_buffer_context(int device, void * dev_ptr) :
@@ -614,7 +616,12 @@ struct ggml_backend_cuda_buffer_context {
     }
 
     ~ggml_backend_cuda_buffer_context() {
-        CUDA_CHECK(cudaFree(dev_ptr));
+        if (owned) {
+            CUDA_CHECK(cudaFree(dev_ptr));
+        } else {
+            // host_ptr was registered via cudaHostRegister; dev_ptr aliases it and is not ours to free.
+            CUDA_CHECK(cudaHostUnregister(host_ptr));
+        }
     }
 };
 
@@ -645,10 +652,16 @@ static enum ggml_status ggml_backend_cuda_buffer_init_tensor(ggml_backend_buffer
         const size_t original_size = ggml_nbytes(tensor);
         const size_t padded_size = ggml_backend_buft_get_alloc_size(buffer->buft, tensor);
 
-        if (padded_size > original_size) {
+        if (padded_size > original_size && ctx->owned) {
             ggml_cuda_set_device(ctx->device);
             CUDA_CHECK(cudaMemset((char *)tensor->data + original_size, 0, padded_size - original_size));
         }
+        // For externally-owned buffers (buffer_from_host_ptr), the memset is
+        // skipped: hipMemset through a hipHostGetDevicePointer-derived address
+        // is unsupported on ROCm integrated GPUs, and the padding region may
+        // extend past the registered host range for the final tensor. GGUF
+        // files zero-pad between tensors by convention, so the padding bytes
+        // in the mmap'd region are already zero.
     }
     return GGML_STATUS_SUCCESS;
 }
@@ -4653,10 +4666,24 @@ static void ggml_backend_cuda_device_get_props(ggml_backend_dev_t dev, ggml_back
     bool events = true;
 #endif
 
+    // buffer_from_host_ptr is currently enabled only on HIP integrated GPUs
+    // (validated on Strix Halo / ROCm 7.2.0). NVIDIA Jetson reports
+    // prop.integrated == 1 too and may benefit from the same path, but it
+    // has not been tested on that platform; #15034's cuda_host-buffer
+    // corruption is in a different code path (see ggml-cuda.cu:243) and
+    // is not expected to apply here, but validation is required before
+    // extending beyond HIP.
+    bool buffer_from_host_ptr = false;
+#if defined(GGML_USE_HIP)
+    cudaDeviceProp prop;
+    CUDA_CHECK(cudaGetDeviceProperties(&prop, ctx->device));
+    buffer_from_host_ptr = prop.integrated > 0;
+#endif
+
     props->caps = {
         /* .async                 = */ true,
         /* .host_buffer           = */ host_buffer,
-        /* .buffer_from_host_ptr  = */ false,
+        /* .buffer_from_host_ptr  = */ buffer_from_host_ptr,
         /* .events                = */ events,
     };
 }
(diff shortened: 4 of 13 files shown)


doctorjei:
Re: the "AI generated content", I listed all of the details. I don't believe I violated the policy. But, it's all there in the clear! :) (I'll admit that I use a lot of dashes and stuff, which I hear is very AI like... ha.)

sayap:
I got this on a 128 GiB strix halo, does it look right? Thanks
```
$ grep -i huge /proc/meminfo
HugePages_Total:   62369
HugePages_Free:    62369
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB
Hugetlb:        127731712 kB
```

doctorjei:
> I got this on a 128 GiB strix halo, does it look right? Thanks
> 
> ```
> $ grep -i huge /proc/meminfo
> HugePages_Total:   62369
> HugePages_Free:    62369
> HugePages_Rsvd:        0
> HugePages_Surp:        0
> Hugepagesize:       2048 kB
> Hugetlb:        127731712 kB
> ```

Yep! That's what you want to see.

BTW, I just added a clarification I foolishly left out. I haven't implemented hipMalloc  ***yet***. Until that is implemented, this will be limited to CPU inference. But, I don't anticipate that it will be too crazy. In hind sight, I probably should have held the PR until then, because I don't know too many people with 128 GiB looking for CPU inference, ha.

With that said, this is the groundwork needed for hipMalloc, which I don't think will be too crazy (famous last words, right?). I should be able to just have the memory "passed" to hipMalloc without copying.

candrews:
Could the implementation be extended to also support Vulkan?

doctorjei:
> Could the implementation be extended to also support Vulkan?

Yes, the Vulkan stack solution is essentially the same (maybe easier, in fact) vs the HIP solution. This could theoretically also apply in Asahi Linux on Mac if "proper" GPU support is ever in place. This is notably not a problem in Windows or MacOS.

Edit: After some investigation, the Vuklan path is not as easy as I thought - but it's still doable. :) I'm actually wrapping up turning on HIP fully now.

doctorjei:
Info about the latest commit...

### Changes

| File | Modification|
|---|---|
| `ggml/src/ggml-cuda/ggml-cuda.cu`|2 macros|
| `ggml/src/ggml-cuda/vendors/hip.h`|ggml_backend_cuda_device_buffer_from_host_ptr,<br /> routing code with comments|

- Flips the `.buffer_from_host_ptr` (**_bfhp_**) backend capability from the hardcoded `false` to runtime-gated (**HIP only**)
- Wires vtable slot from `NULL` to a handler (for HIP)  built on `hipHostRegister`.
- Reuses previously Metal-only branch at `llama-model.cpp:7802` for HIP w. integrated GPU to pass hugetbl area directly to GPU (avoiding code duplication)

### Potential Issues (for reviewers)

- **Dropped `cudaHostRegisterReadOnly` registration flag.**, as it triggered `hipMemset` illegal-access fault on load; loader write to it during quantized-tensor padding init.
- **Added `&& ctx->owned` guard to `init_tensor`'s quantized-padding memset** (cause same as above). Also prevents potential out-of-bounds `memset`.

### Benchmark — Strix Halo, ROCm 7.2.0, Linux 6.17.x

**Model:** qwen3 (MOE) 30B, Q4 variant (17.28 GiB)

| Testcase | pp512 (t/s) | tg64 (t/s) |
|---|---|---|
| Mainline | 565.48 ± 4.39 | 32.02 ± 0.91 |
| HIP bfhp | 577.79 ± 52.62 | 31.21 ± 0.99 |
| HIP bfhp, hugepages | 585.44 ± 72.00 | 32.22 ± 0.71|

**Result**: pp512 and tg64 differences insignificant statistically; there's no throughput regression for the bfhp or bfhp + hugepages variants.

JohannesGaessler:

doctorjei:
@JohannesGaessler, I would be most grateful if you could tell me exactly what your objection is; as far as I know, I have not broken the policy. If there's something specific, I will fix it.

I spent a lot of time on this. If you don't want it, that's fine, and your prerogative. But if you really just think this is "slop" from some AI, you are mistaken. There posts were mine; the edits were mine; the responses we all mine. I am not even sure what makes you think this was AI generated. Hell, I even rebased and broke it into separate commits just to make it easier to follow.

The only thing I can think of is that I used AI to make some tables, and the pull request formatting might seem formulaic... is it one of those things? I'm grasping at straws here. I don't believe either of those violate the spirit or letter of the rules.

(I'm not trying to be flippant... but after all of my time spent, it would be nice to know exactly what the issue is so that I can correct it. My time is valuable, too. LLMs are not my full time job, but memory / kernels / etc are my bread and butter. I'm not trolling for merges here.)

doctorjei:
> Hi @doctorjei, thanks for your contribution!
> 
> 
>     * **AI-generated content**: This project does not accept PRs, descriptions or commit messages that are fully or predominantly AI-generated. If you have used AI to assist you in writing code, please make sure to disclose that explicitly.
> 
> 

Maybe because I didn't respond to this message directly, my PR was incorrectly flagged.

***This is a mistake.*** I don't know how @ggml-gh-bot makes this determination, so I don't even know how to address it.

Re: the "AI generated content", I listed all of the details. I don't believe I violated the policy. But, it's all there in the clear! :) (I'll admit that I use a lot of dashes and stuff, which I hear is very AI like... ha.)

JohannesGaessler:
I read the OP and got the impression it was written by a language model, which is grounds for immediate closure.

doctorjei:
> I read the OP and got the impression it was written by a language model, which is grounds for immediate closure.

Ok. I appreciate the response. I can understand that impression. And I apologize for what I now see if an overly aggressive response. I did use tables (as I mentioned) and asked for a review of the text via LLM. I'll clear out the description and rewrite it - without an editing pass from an LLM - or if you prefer, submit a new PR.

Edit: oh, here's some sweet irony... I have typos in my OP even though I got editing help :D (I only say this as some evidence of the "human" nature of it (e.g., line under "motivation" has a random "v", and there are some other winners.. I see I duped some stuff in the "motivation" and "why" sections, too... anyway, I will continue rewriting now.)

JohannesGaessler:
To clarify why we have such a hard-line stance: we get a lot of garbage PRs where the human pressing the submit button has no actual understanding of the code. We explicitly disallow using a language models to write PR descriptions because that is the primary way to mask a lack of understanding. Rewriting a PR description after the fact is usually not a reason for reconsideration. I will come back to this PR after https://github.com/ggml-org/llama.cpp/pull/20834 has been resolved.

doctorjei:
I understand why you do what you do, and I appreciate the effort. I'll be around when / if you want to consider it. Until then, I'll have access to it myself.

I know this is a big project and you have a lot of people and PRs. I'm just trying to do my part. Good luck.

doctorjei:
From a review in another repo, I've made some tweaks:
1. **`include/llama.h`** - moved `use_hugepages` to end of struct.
2. **`src/llama-model-loader.cpp`** - switched to `mapping->mmap_size()` (from `mapping->size()`)
3. **`src/llama-mmap.cpp`** - added fallback `MAP_HUGE_SHIFT` (in case undefined in some glibc variants)

Other notes:
1. The commit isn't showing here, I suppose because the PR is "closed" still; just wanted to note the additional update.
2. There's an NVIDIA tag, but (I believe) this commit doesn't actually touch NVIDIA devices at all; at least for the moment, this is CPU and HIP only, because I haven't been able to test on NVIDIA yet (though I will put in the leg work to do so if this is considered for merging)
3. Here's the fork this was recently merged into, if you want to see that review:
https://github.com/domvox/llama.cpp-turboquant-hip/pull/3

nicman23:
i am testing both prs and the calculation of how many pages you need is broken for qwen 3.5 397 it told me to try 

`sysctl -w vm.nr_hugepages=6 `

and then

`sysctl -w vm.nr_hugepages=23401`

also after running vm.nr_hugepages=84000 , it still says 

```

E llama_model_load: error loading model: hugetlb mmap failed: need 23584 free 2 MiB pages, pool has 13567. Try: sudo sysctl -w vm.nr_hugepages=23584
0.09.808.135 E llama_model_load_from_file_impl: failed to load model
0.09.808.138 E common_init_from_params: failed to load model 'unsloth/Qwen3.5-397B-A17B-GGUF/MXFP4_MOE/Qwen3.5-397B-A17B-MXFP4_MOE-00001-of-00006.gguf'
0.09.808.141 E srv    load_model: failed to load model, 'unsloth/Qwen3.5-397B-A17B-GGUF/MXFP4_MOE/Qwen3.5-397B-A17B-MXFP4_MOE-00001-of-00006.gguf'
 

```

doctorjei:
> i am testing both prs and the calculation of how many pages you need is broken for qwen 3.5 397 it told me to try 'sysctl -w vm.nr_hugepages=6
> 
> `sysctl -w vm.nr_hugepages=6 `
> 
> and then
> 
> `sysctl -w vm.nr_hugepages=23401`
> 
> also after running vm.nr_hugepages=84000 , it still says
> 
> ```
> 
> E llama_model_load: error loading model: hugetlb mmap failed: need 23584 free 2 MiB pages, pool has 13567. Try: sudo sysctl -w vm.nr_hugepages=23584
> 0.09.808.135 E llama_model_load_from_file_impl: failed to load model
> 0.09.808.138 E common_init_from_params: failed to load model 'unsloth/Qwen3.5-397B-A17B-GGUF/MXFP4_MOE/Qwen3.5-397B-A17B-MXFP4_MOE-00001-of-00006.gguf'
> 0.09.808.141 E srv    load_model: failed to load model, 'unsloth/Qwen3.5-397B-A17B-GGUF/MXFP4_MOE/Qwen3.5-397B-A17B-MXFP4_MOE-00001-of-00006.gguf'
>  
> ```

Do you have 224 GiB of RAM? The last error basically says you don't have enough memory, but the message was a little misleading. It's telling you how much is needed for the current *file*, not the whole model. (Easy fix; once I hear back on whether or not someone might actually merge this, I'll dig into it.)

First file: 10.9 MB (10.4 MiB), thus 6 x 2 MiB pages. (This is why you saw "6" first.)
Second file: 49.1 GB (45.7 GiB = 46,797 MiB), thus 23399 pages. (This is a little off tho from 23584)

For hugepages=84000, you don't seem to have enough free RAM; it's only allocating 13567 (26.5 GiB). At least, that's what this error is indicating. But this model is like 237 GiB. Do you actually have that much memory? and did you hnave 50 GB free RAM space?

nicman23:
I mean then shouldn't it tell you that? Also shouldn't it try to alloc then? That system has hybrid / transparent huge pages 

Also I had 220 ram and 96 vram

doctorjei:
Thanks for the specs.

The error message is trying to allocate it but not finding enough hugepages (2 MiB in size). For whatever reason, it isn't finding enough of those pages. I'm not sure why, but it could be that memory is being used for caches, for example.

Could you run this after you allocate the hugepages?

`grep -i huge /proc/meminfo`

This will tell you how many 2 MiB pages have actually been allocated. Then if you run llama, if the numbers don't match up, it might indicate an issue. Right now I'm flying a bit blind, though.

Regarding what the PR should do - for sure, in the best of worlds, llama would be able to use hugepages and standard pages. That's a good long term goal (though TBH I'm not sure if it is practical). And if this PR is going to be merged, I'll definitely rework the suggested number of pages. Right now, llama can't use hugepages at all; this is just the first step.

nicman23:
i do not have enough of ram to actually put the whole qwen 3.5 397b so it fails. if i run it normally it works - 30 prefil and 30 decode

if you need more info it is a g4-standarn-48 :P - and on a spot vm is like ~ 1 dollar an hour

PR #21822, opened 2026-04-12
cli : Use acquire/release semantics for stopping logic

matthiasstraka:
## Overview

`Ctrl+C` stopping logic was done correctly using atomics - but sequential CST ordering has no benefits in this case. Using Acq/Rel is slightly faster, which helps since `should_stop` is queried every output token. In addition, I added a utility function for resetting the stop request (makes code more readable) and used an exchange operation for the signal handler to simplify the code.

The variable `loading_show` was unused, so I removed it.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/tools/cli/cli.cpp
+++ b/tools/cli/cli.cpp
@@ -12,17 +12,21 @@
 #include <array>
 #include <atomic>
 #include <algorithm>
+#include <csignal>
 #include <filesystem>
 #include <fstream>
 #include <thread>
-#include <signal.h>
 
 #if defined(_WIN32)
 #define WIN32_LEAN_AND_MEAN
 #ifndef NOMINMAX
 #   define NOMINMAX
 #endif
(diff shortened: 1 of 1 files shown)


ngxson:
can you have a look on this comment? https://github.com/ngxson/llama.cpp/pull/101#pullrequestreview-4114853768

matthiasstraka:
> can you have a look on this comment?

This goes beyond the original PR but I will investigate the proposed solution and update when it makes sense.

matthiasstraka:
Is there anything more to do to get this PR accepted? @ngxson

ngxson:
thanks for the PR, but I think this is not very necessary overall, there is no known bugs related to this so I assume this is a nits improvement.

re your point:

> should_stop is queried every output token

this should not be a problem, only the CLI thread query it, but not the inference thread

Issue #19979, opened 2026-02-28
Eval bug: Memory leak? using ROCm


MartinEmrich:
### Name and Version

```
build/bin/llama-cli --version
ggml_cuda_init: found 1 ROCm devices:
  Device 0: AMD Radeon RX 7800 XT, gfx1101 (0x1101), VMM: no, Wave Size: 32
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = AMD Radeon RX 7800 XT (RADV NAVI32) (radv) | uma: 0 | fp16: 1 | bf16: 0 | warp size: 64 | shared memory: 65536 | int dot: 1 | matrix cores: KHR_coopmat
version: 8179 (ecbcb7ea9)
built with GNU 14.2.0 for Linux x86_64
```

### Operating systems

Linux

### GGML backends

HIP

### Hardware

Radeon RX 7800 XT 16GB
Ryzen 2700X on X470, 16GB RAM

### Models

bartowski/Qwen_Qwen3.5-27B-GGUF:IQ3_XS

### Problem description & steps to reproduce

This model comes in at ~10.8GB.

Running with ROCm and large prompt/context, VRAM usage slowly creeps up while prompt processing until it crashes. Running with Vulkan, VRAM usage stays low and it completes.

Command:
```
CUDA_VISIBLE_DEVICES=0 GGML_VISIBLE_DEVICES=""  build/bin/llama-bench -v -m ~/.cache/llama.cpp/bartowski_Qwen_Qwen3.5-27B-GGUF_Qwen_Qwen3.5-27B-IQ3_XS.gguf -ngl 99 -fa 1 -ctk q4_0 -ctv q4_0 -p 65536
```

Watching with `amdgpu_top --smi`, VRAM usage starts out at 12600MB (10MB GTT), and VRAM usage slowly creeps up. When the VRAM size is reached, it crashes with this error:

<details>
<summary>Error + Stacktrace</summary>
```
ROCm error: out of memory
  current device: 0, in function alloc at /home/martin/upstream/llama.cpp/ggml/src/ggml-cuda/ggml-cuda.cu:393
  ggml_cuda_device_malloc(&ptr, look_ahead_size, device)
/home/martin/upstream/llama.cpp/ggml/src/ggml-cuda/ggml-cuda.cu:97: ROCm error
[New LWP 773974]
[New LWP 773973]
[New LWP 773971]
[New LWP 773970]
[...]

PR #21830, opened 2026-04-13
Fix unbounded VRAM usage creep on HIP/ROCm backend when quantizing kv cache

stragulus:
## Overview

Fixes unbounded VRAM usage growth when using flash attention and quantized kv cache types on HIP/ROCm with RDNA3 architecture (and possibly more)

:warning: Draft status because I just recently updated to latest master and there seems to be a regression in the PP benchmarks BEFORE my changes. I can't explain away why these happen AFTER my changes for the unquantized (f16) kv cache benchmarks as those should follow the same code path as before my changes. TBD

Fixes issue https://github.com/ggml-org/llama.cpp/issues/19979

## Additional details

This PR batches flash attention kernel invocation with a fixed size batch when quantized kv cache types are used. Without this change, the kernels would increasingly allocate more vram as they require the entire kv cache to be dequantized to f16 types. A compounding problem existed when using the legacy memory pool ggml_cuda_pool_leg (e.g. when using HIP with at least RDNA3). The memory pool allocates fragments of a fixed size, and allocation requests use a best-fit algorithm. Allocation requests that don't fit into any of the existing individual fragments trigger allocation of a new fragment. The pool never deallocates fragments until the pool itself is destructed. Due to the ever-growing sizes of the allocation requests, new allocations would not fit in a previously created fragment, thus another new fragment would be allocated. This rapidly consumes more and more vram, completely cancelling out any benefit to be had from using quantized kv cache types for non-synthetic workloads.

The VEC flash attention kernel solves this by dequantizing the kv cache entries one by one instead of all in advance. Simple, however, this has a very significant impact on prompt processing speed (easily ~50% reduction compared to the other kernels like TILE). While this does completely solve the growing vram usage problem, the performance impact is not acceptable.

Thus, I chose a hybrid solution where we don't fall back to slow VEC, but instead the flash attention kernel calls are batched with a maximum batch size if quantized kv types are used. The code will choose the previous unbatched path for maximum performance if batching is not required.

Token generation speed should be unaffected by these changes.

Passed opencode 'summarize README.md' manual sanity check  against llama-server with q8_0/q8_0 quants when run like so:

```
build/bin/llama-server -m Qwen3.5-27B.Q4_K_M.gguf -ngl 100 -c 1 10000 -fa 1 -np 1 --cache-type-k q8_0 --cache-type-v q8_0
```

## Benchmarks

Tests ran on my arch linux system with a 7900 XTX and ROCm-7.2.1

### Quantized K/V cache

Before (k/v q8_0/q8_0)

```
build/bin/llama-bench -m /spare/llama.cpp/Qwen3.5-27B.Q4_K_M.gguf -ngl 100 -fa 1 --cache-type-k q8_0 --cache-type-v q8_0 -p 30000 -r 1

VRAM usage creeps up slowly during run, to 18039M during prompt processing run

| model                          |       size |     params | backend    | ngl | type_k | type_v | fa |            test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | -----: | -----: | -: | --------------: | -------------------: |
| qwen35 27B Q4_K - Medium       |  15.39 GiB |    26.90 B | ROCm       | 100 |   q8_0 |   q8_0 |  1 |         pp30000 |        235.41 ± 0.00 |
| qwen35 27B Q4_K - Medium       |  15.39 GiB |    26.90 B | ROCm       | 100 |   q8_0 |   q8_0 |  1 |           tg128 |         27.76 ± 0.00 |
```

After

```
build/bin/llama-bench -m /spare/llama.cpp/Qwen3.5-27B.Q4_K_M.gguf -ngl 100 -fa 1 --cache-type-k q8_0 --cache-type-v q8_0 -p 30000 -r 1

VRAM usage stable at 17158M during prompt processing run

| qwen35 27B Q4_K - Medium       |  15.39 GiB |    26.90 B | ROCm       | 100 |  1 |         pp30000 |        703.82 ± 0.00 |
| qwen35 27B Q4_K - Medium       |  15.39 GiB |    26.90 B | ROCm       | 100 |  1 |           tg128 |         28.89 ± 0.00 |
```

### Non-quantized K/V cache

E.g. using standard f16/f16 k/v cache types, everything else being the same

Before
```
| qwen35 27B Q4_K - Medium       |  15.39 GiB |    26.90 B | ROCm       | 100 |  1 |         pp30000 |        236.32 ± 0.00 |
| qwen35 27B Q4_K - Medium       |  15.39 GiB |    26.90 B | ROCm       | 100 |  1 |           tg128 |         28.87 ± 0.00 |
```

After
```
VRAM stable at 18024

| qwen35 27B Q4_K - Medium       |  15.39 GiB |    26.90 B | ROCm       | 100 |  1 |         pp30000 |        704.94 ± 0.00 |
| qwen35 27B Q4_K - Medium       |  15.39 GiB |    26.90 B | ROCm       | 100 |  1 |           tg128 |         28.68 ± 0.00 |
```

# Requirements

- AI usage disclosure: Used AI to help locate the problem and prototype a solution. Which it failed at as this one was not so straightforward, but gave me enough to work on to come up with the idea for the actual fix. Some details about the lower level flash kernel internals I'm not familiar with, I mostly focused on the higher level batching.

### DIFF
--- a/ggml/src/ggml-cuda/fattn-common.cuh
+++ b/ggml/src/ggml-cuda/fattn-common.cuh
@@ -7,6 +7,7 @@
 #include <cstdint>
 
 #define FATTN_KQ_STRIDE       256
+#define FATTN_KV_CONV_BATCH   1024 // max KV tokens converted to f16 at once to bound pool allocation growth
 #define HALF_MAX_HALF         __float2half(65504.0f/2) // Use neg. of this instead of -INFINITY to initialize KQ max vals to avoid NaN upon subtraction.
 #define SOFTMAX_FTZ_THRESHOLD -20.0f                   // Softmax exp. of values smaller than this are flushed to zero to avoid NaNs.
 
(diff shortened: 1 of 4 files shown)


JohannesGaessler:
It makes me extremely suspicious that batching the kernels would increase t/s, to me that suggests that an incorrect optimization is taking place somewhere. In any case, the changes in this PR are way too verbose for what they're doing. This should be tens of lines of code, not hundreds.

stragulus:
Closing this for now until I have figured out the regression path

williamtwomey:
Just wanted to confirm this branch fixed the vRAM leak issue I had on 2x R9700s on ROCM 7.2. 

There is definitely a performance hit (~35t/s -> ~20t/s on Qwen122B), but looks promising!

PR #21846, opened 2026-04-13
ggml-cuda: avoid cudaMemcpyPeerAsync when P2P is unavailable

jellyuwu:
## Overview

Fixes corrupted multi-GPU output on systems where CUDA peer access is not available between devices.

## Additional information

Cross-device copies used cudaMemcpyPeerAsync even when P2P was unavailable. On non-P2P topologies this can route through host staging, and stream event ordering can allow destination-side reads before data is fully visible, causing stale recurrent state reads and garbled tokens.

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1091,6 +1091,9 @@ struct ggml_cuda_device_info {
 
     cuda_device_info devices[GGML_CUDA_MAX_DEVICES] = {};
 
+    // peer_access[i][j] = true if GPU i can directly access GPU j's memory (P2P enabled)
+    bool peer_access[GGML_CUDA_MAX_DEVICES][GGML_CUDA_MAX_DEVICES] = {};
+
     std::array<float, GGML_CUDA_MAX_DEVICES> default_tensor_split = {};
 
 #ifdef GGML_USE_NCCL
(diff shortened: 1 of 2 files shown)


am17an:
Do you have a reproduction of this issue?

jellyuwu:
I can reproduce this on a dual-GPU Linux machine where the GPUs do not have CUDA P2P access between them (`nvidia-smi topo -m` shows `SYS`, and `cudaDeviceCanAccessPeer(0,1) == 0`).

On an unfixed build, start a model split across both GPUs with `--n-gpu-layers 99` (I reproduced it with both Qwen3.5-35B-A3B and Gemma-4-31B), then send a trivial request like “What is 2+2?” to `llama-server`.

In that setup, I get corrupted output: garbled/incoherent text, sometimes empty `content`, and for Qwen often garbage in `reasoning_content`.

With the exact same hardware and server command, this PR fixes it and the outputs become coherent again.

A useful A/B check is that single-GPU or partial CPU offload worked correctly even before the fix, while the broken behavior showed up when the model was fully split across both non-P2P GPUs.

This seems topology-dependent, so reviewers on peer-capable/NVLink systems may not be able to reproduce it.

JohannesGaessler:
`cudaMemcpyPeerAsync` falls back to `cudaMemcpyAsync` if P2P is unavailable. Also what this PR does is not disable P2P copies but disable GPU -> GPU copies entirely. The AI usage disclosure is missing and the code looks machine generated so I think this is just a misdiagnosis of the problem because a language model got confused about the meaning of `GGML_CUDA_NO_PEER_COPY`.

PR #21865, opened 2026-04-13
vulkan: add barrier after writetimestamp

jeffbolznv:
## Overview

Add a pipelinebarrier after each writetimestamp call in the perf_logger code.

## Additional information

The vulkan spec doesn't prevent commands issued after a timestamp from starting to execute before the timestamp is written. The NV driver had been ordering these, but future drivers won't. So we need a barrier after each timestamp to order the timestamp vs the next commands.

# Requirements

- AI usage disclosure: NO.


jeffbolznv:
Bumping for review.

0cc4m (approved):

0cc4m:

am17an (approved):

PR #21869, opened 2026-04-13
ggml: remove ggml-ext.h

### LINKED ISSUE #21860: Compile bug: fatal error: '../src/ggml-ext.h' file not found
### Git commit

```
/usr/ports/misc/llama-cpp/work/llama.cpp-b8779/src/llama.cpp:19:10: fatal error: '../src/ggml-ext.h' file not found
   19 | #include "../src/ggml-ext.h"
      |          ^~~~~~~~~~~~~~~~~~~
1 error generated.
```

Version: 8779
FreeBSD 15 STABLE

### Operating systems

BSD

### GGML backends

CPU

See above.

### First Bad Commit

n/a

### Compile command

```shell
n/a
```

### Relevant log output

```shell
see above
```

### LINKED ISSUE #21867: Homebrew can't compile llama.cpp
Many people use homebrew to manage and update llama.cpp. It's very convenient to update, and control everything from there.

Unfortunately, recently, build of llama.cpp started to fail.

Here's the run log: https://github.com/Homebrew/homebrew-core/actions/runs/24323950806/job/71015710159

The error is the following:
```
  /private/tmp/llama.cpp-20260412-6291-7hq766/src/llama.cpp:19:10: fatal error: '../src/ggml-ext.h' file not found
     19 | #include "../src/ggml-ext.h"
        |          ^~~~~~~~~~~~~~~~~~~
  1 error generated.
```

This is caused by this code:
```
// TODO: tmp until the ggml meta backend matures and becomes public
#include "../src/ggml-ext.h"
```

Here's the failing PR: https://github.com/Homebrew/homebrew-core/pull/277280

ngxson:
## Overview

Fix https://github.com/ggml-org/llama.cpp/issues/21867 Fix https://github.com/ggml-org/llama.cpp/issues/21860

Not quite sure if the ggml-ext.h is intended to be a public header, but I believe it should be (so that the symbols can be exposed in the dynamic library)

# Requirements

- AI usage disclosure: no


ggerganov (approved):
The idea was `ggml-ext.h` to be a private header, but I now realize that this will not work. Removed the header and moved everything to the `ggml-backend.h`. cc @JohannesGaessler

JohannesGaessler (approved):
Either way is fine with me.

PR #21870, opened 2026-04-13
common: skip reasoning budget sampler when no budget is requested

### LINKED ISSUE #21784: Misc. bug: Gemma 4 template changes causing degraded inference speed
### Name and Version

llama-cli --version
version: 8744 (d7ff074c8)

### Operating systems

Windows

llama-server

### Command line

```shell

```

Token generation speed drops from 98t/s down to 70t/s with new reasoning budget template changes in llama.cpp.

The last good build that works is: https://github.com/ggml-org/llama.cpp/releases/tag/b8742

The build where the performance degradation starts is: https://github.com/ggml-org/llama.cpp/releases/tag/b8744

I've been using the Windows x64 Vulkan build.

**Verification:**

I've also tested by adding the --chat-template parameter to use a different template using the same build, and performance goes back up to 98t/s.  However, that is not a fix, because all templates are broken with any of these newer builds for me.

Performance on other models is fine.

**What should happen:**

* You should be able to use the Windows x64 Vulkan build (or other builds, if those are effected too) and the default template should not unnecessarily degrade performance.

* Alternative templates shouldn't cause Gemma 4 to spit out its own template in responses.  I didn't identify which commit started that issue, but if that wasn't a problem then at least it would have been an option to avoid the performance issue temporarily.

### First Bad Commit

https://github.com/ggml-org/llama.cpp/commit/d7ff074c87ecacd57d5760e2f678866ba9fe7149

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

berkidem:
## Overview

After I added thinking_start_tag / thinking_end_tag for gemma4 in #21697, the reasoning budget sampler gets unconditionally created even if no budget is configured. The same applies to kimi_k2, lfm2, lfm2_5, and ministral_3 which also set these tags. The budget gets converted to INT_MAX, so the sampler never actually forces any tokens but still runs per-token checks (start tag matching in IDLE state, token-to-piece conversion + UTF-8 checks in COUNTING state).

More importantly, the existence of the sampler (i.e., non-null rbudget) disables backend sampling. Backend sampling lets the GPU select tokens directly, avoiding a full logits transfer from GPU to CPU every token. This could potentially explain the 30% speed regression reported in #21784 (98 t/s to 70 t/s on Vulkan).

So I added a reasoning_budget_tokens >= 0 check to the sampler creation condition. When the budget is unlimited, the sampler is not created, backend sampling stays enabled, and no per-token overhead is added. When a budget is explicitly set (0, 128, 1024, etc.), the sampler is created and works as before.

Should fix #21784

## Additional information

Note that I do not have a Vulkan environment so I can replicate neither the original degradation nor that the fix works in terms of performance. However, I did play with the backend-sampling flag myself manually on CUDA and my commit gave me around 5% improvement (~215t/s vs ~225t/s). In any case, it is logically cleaner and should be an improvement in all cases.

# Requirements

- AI usage disclosure: YES: I asked my local agent to identify the potential sources of the performance regression on Vulkan, and then I manually validated and tested and validated the logic.

### DIFF
--- a/common/sampling.cpp
+++ b/common/sampling.cpp
@@ -287,8 +287,8 @@ struct common_sampler * common_sampler_init(const struct llama_model * model, st
         }
     }
 
-    // reasoning budget sampler
-    if (!params.reasoning_budget_start.empty() && !params.reasoning_budget_end.empty()) {
+    // reasoning budget sampler (skip when budget is unlimited to avoid per-token overhead)
+    if (!params.reasoning_budget_start.empty() && !params.reasoning_budget_end.empty() && params.reasoning_budget_tokens >= 0) {
         rbudget = common_reasoning_budget_init(
             vocab,
             params.reasoning_budget_start,
--- a/common/sampling.cpp
+++ b/common/sampling.cpp
@@ -287,8 +287,8 @@ struct common_sampler * common_sampler_init(const struct llama_model * model, st
         }
     }
 
-    // reasoning budget sampler (skip when budget is unlimited to avoid per-token overhead)
-    if (!params.reasoning_budget_start.empty() && !params.reasoning_budget_end.empty() && params.reasoning_budget_tokens >= 0) {
+    // reasoning budget sampler (skip when budget is unlimited unless a lazy grammar is active, which needs rbudget for thinking-block suppression)
+    if (!params.reasoning_budget_start.empty() && !params.reasoning_budget_end.empty() && (params.grammar_lazy || params.reasoning_budget_tokens >= 0)) {
         rbudget = common_reasoning_budget_init(
             vocab,
             params.reasoning_budget_start,


aldehir:
This was intentionally done in #20970 to avoid triggering the grammar if a model generates a tool call token in its reasoning traces.

Perhaps it should be opt-in per model. I know the Gemma 4 series don't emit tool calls tokens when reasoning, so it can be safely disabled for it.

And it can probably be turned off if no tools are provided.

CMay:
@berkidem This fix does look promising, even if there may be other ways to optimize for people who do use the budget sampler.  A lot of people may get a decent token generation bump from this even if they don't know why. :)

berkidem:
@aldehir, thanks a lot for the feedback.

Happy to take a pass at this as a follow-up if you can point me in the right direction or if you'd prefer to address it in this PR, let me know and I'll make another commit here instead.

From exploring the code, would the opt-in be a flag on `common_chat_params` that each model's init sets, or is there a cleaner place for this? Also, for the "no tools" situation, is that just checking `grammar_lazy` or is there a better check for tool use?

aldehir:
The intent was to avoid prematurely matching the grammar trigger when `grammar_lazy = true`. Checking for `grammar_lazy = true` should be sufficient. If it's not set, don't attach the sampler unless it's explicitly defined (`reasoning_budget >= 0`). Backend sampling is disabled whenever a grammar is defined, lazy or not.

I am skeptical this yields a noticeable improvement. My assessment aligns with the one you posted in the issue. I do see the argument for enabling backend sampling though.

berkidem:
Updated the condition as suggested; now it preserves `rbudget` when `grammar_lazy` is true.

aldehir (approved):
Looks good, thanks!

cc @pwilkin

pwilkin (approved):

CMay:
I can confirm that this merge brought performance back up to 98t/s  :)  Thanks for taking the time to fix an issue that didn't impact you.

PR #21873, opened 2026-04-13
ggml-webgpu: compute pass batching and removing profiling overhead

reeselevine:
## Overview

Per https://bugs.webkit.org/show_bug.cgi?id=311598, it seems like lots of individual compute passes can cause issues on some WebGPU implementations. This PR batches multiple operations into the same compute pass, which does seem to be stable and allows for much faster decode speeds on Safari and iOS devices (30-54 t/s for small models and q4-ish quantizations). Performance doesn't change significantly on other devices or browsers, from what I've seen.

Along with this change, I fixed the overhead of WebGPU profiling by creating a single query set which records all timestamps, and is only mapped once at the end of each forward pass of the model. Unfortunately, timestamps only work at the boundaries of compute passes, so when profiling is enabled, we do not batch compute passes. This means that profiling won't work on iOS right now, but I think that's a reasonable trade-off.

# Requirements

- AI usage disclosure: yes, to help batch timestamps into a single query set


CISC (approved):

reeselevine:
@ggml-org/maintainers another review please :)

PR #9639, opened 2024-09-25
Tool call support (generic + native for Llama, Functionary, Hermes, Mistral, Firefunction, DeepSeek) w/ lazy grammars

ochafik:
This supersedes https://github.com/ggerganov/llama.cpp/pull/6389 (now using a fully C++ approach), https://github.com/ggerganov/llama.cpp/pull/5695 (first attempt at supporting Functionary) and https://github.com/ggerganov/llama.cpp/pull/9592 (more recent Python wrapper).

## Which models are supported (in their native style)?

While *any model* should work (w/ generic fallback using JSON schema constraints), this PR supports the native call style of a few models:
- Llama 3.1 / 3.3 (including builtin tools support), Llama 3.2
- Functionary v3.1 / v3.2
- Hermes 2/3, Qwen 2.5
  https://github.com/ggml-org/llama.cpp/pull/12034
- Mistral Nemo
- Firefunction v2
- DeepSeek R1
  https://github.com/ggerganov/llama.cpp/pull/11607
- Command 7RB
  https://github.com/ggerganov/llama.cpp/pull/11585

(note: streaming ~~incubated~~ implemented in https://github.com/ggml-org/llama.cpp/pull/12379)

<details>
<summary>Show all templates supported by minja and which handler they use</summary>

| Template | Format |
|----------|--------|
| CohereForAI-c4ai-command-r-plus-default.jinja | generic tool calls |
| CohereForAI-c4ai-command-r-plus-rag.jinja | generic tool calls |
| CohereForAI-c4ai-command-r-plus-tool_use.jinja | generic tool calls |
| MiniMaxAI-MiniMax-Text-01.jinja | generic tool calls |
| NexaAIDev-Octopus-v2.jinja | generic tool calls |
| NousResearch-Hermes-2-Pro-Llama-3-8B-default.jinja | generic tool calls |
| NousResearch-Hermes-2-Pro-Llama-3-8B-tool_use.jinja | hermes 2 pro tool calls |
| NousResearch-Hermes-2-Pro-Mistral-7B-default.jinja | generic tool calls |
| NousResearch-Hermes-2-Pro-Mistral-7B-tool_use.jinja | hermes 2 pro tool calls |
| NousResearch-Hermes-3-Llama-3.1-70B-default.jinja | generic tool calls |
| NousResearch-Hermes-3-Llama-3.1-70B-tool_use.jinja | hermes 2 pro tool calls |
| OrionStarAI-Orion-14B-Chat.jinja | generic tool calls |
[...]

PR #21890, opened 2026-04-14
server: add --parallel-tool-calling flag to enable by default

Linus467:
## Overview

Add a `--parallel-tool-calling` / `--no-parallel-tool-calling` command-line flag (and `LLAMA_ARG_PARALLEL_TOOL_CALLS` environment variable) to the server example. When enabled, the server defaults `parallel_tool_calls` to `true` instead of `false`. Clients can still override the default by explicitly setting `parallel_tool_calls` in the request body.

## Additional information

N/A

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -3164,6 +3164,14 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_PREFILL_ASSISTANT"));
     add_opt(common_arg(
+        {"--parallel-tool-calls"},
+        {"--no-parallel-tool-calls"},
+        "enable parallel tool calls by default (default: disabled)",
+        [](common_params & params, bool value) {
+            params.parallel_tool_calls = value;
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_PARALLEL_TOOL_CALLS"));
+    add_opt(common_arg(
         {"-sps", "--slot-prompt-similarity"}, "SIMILARITY",
         string_format("how much the prompt of a request must match the prompt of a slot in order to use that slot (default: %.2f, 0.0 = disabled)\n", params.slot_prompt_similarity),
         [](common_params & params, const std::string & value) {
--- a/common/common.h
+++ b/common/common.h
@@ -591,6 +591,7 @@ struct common_params {
     bool use_jinja = true;                                                                                  // NOLINT
     bool enable_chat_template = true;
     bool force_pure_content_parser = false;
+    bool parallel_tool_calls = false;
     common_reasoning_format reasoning_format = COMMON_REASONING_FORMAT_DEEPSEEK;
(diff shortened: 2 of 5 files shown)


pwilkin:
This is incorrect. Parallel tool calling should be enabled by default on those templates / models that actually support it. If the support is not correctly detected, that's a detection problem. If a model does not support parallel tool calls, trying to force it to do so will be counterproductive.

aldehir:
I don't think that's true, I almost always have to pass it into the API. Otherwise it won't perform parallel tool calling.

sayap:
I can chime in a little on this.. So parallel tool calls is indeed disabled by default, ever since it was introduced by #9639 :
```
inputs.parallel_tool_calls = json_value(body, "parallel_tool_calls", false);
```

This still takes effect today, more context in https://github.com/ikawrakow/ik_llama.cpp/issues/1487#issuecomment-4136162355

Anyway, my current thinking is that parallel tool calls really doesn't work well for most models, e.g.
* the issue reported in the link above for Qwen3.5-35B-A3B
* https://github.com/ikawrakow/ik_llama.cpp/issues/1613 for Qwen3.5-27B
* https://github.com/ggml-org/llama.cpp/issues/22043 for Nemotron-3-Nano-4B.

I also experienced same infinite tool calls issue with a Q8_0 quant of the latest Qwen3.6-35B-A3B.

So, I would advice against enabling it by default.

aldehir:
It doesn't work because the grammars for those models likely need to be adjusted. Not necessarily because they don't support it well. Before the autoparser merged in, parallel tool calling was working well with the nemotron nano 3 models.

I agree, should probably figure out what's causing the issues first and then think about enabling it. Worth noting some clients already set it to true, such as Pi.

pwilkin:
I'll run more comprehensive tests, it seems the unit tests are not enough to catch this (after we merge the space-allow PR).

sayap:
My understanding is that the infinite tool calls issue is caused by in-context learning and .. luck. The grammars are fine.

So with parallel tool calls enabled, after making a tool call, a model is allowed to either:
(a) start another tool call
(b) end the turn

At the beginning, the model may decide to make 2~3 parallel tool calls, e.g. grep for a few log patterns. Once that happens, the chance of (a) gets a bit higher, and the chance of (b) gets a bit lower.

Then, by chance, the model may make 4~5 parallel tool calls. Once that happens, the chance of (a) gets even higher, and the chance of (b) gets even lower.

If this goes on, then the chance of (a) approaches 100%.

aldehir:
@sayap see #22073 as an example. Sometimes the grammar stops the model from reaching an appropriate end, so their only option is to keep tool calling.

sayap:
Ah I see, good catch :+1: 

I have only tried parallel tool calls with the Qwen models, and I think those already have the correct grammars.

sayap:
Sorry false alarm.. Actually it is almost impossible to trigger an infinite tool calls loop [1], but we suspect that an off-by-one KV cache issue in ik_llama.cpp is causing it to happen very frequently.

[1] tested with Qwen3.6-35B-A3B Q8_0, YMMV with other models/quants

PR #20505, opened 2026-03-13
convert: Fix Qwen3.5/Qwen3.5 Moe NVFP4 Conversions

michaelw9999:
This PR fixes several errors that occur when attempting to convert Qwen3.5/Qwen3.5Moe models.  To keep this PR scope in check and specific, a separate PR [https://github.com/ggml-org/llama.cpp/pull/20506](https://github.com/ggml-org/llama.cpp/pull/20506)  allows loading of these newly converted models.

Bug:
When attempting to use  `convert_hf_to_gguf.py` on various Qwen3.5 and Qwen3.5 MoE models, it would abort with the following error(s):

```
ValueError: Can not map tensor 'model.language_model.layers.0.mlp.shared_expert.down_proj.weight'
ValueError: Can not map tensor 'model.language_model.layers.0.linear_attn.in_proj_a.weight'
```
This occurred because these models now have `model.language_model` or `language_model` prefixes.  The fix strips the wrappers instead of failing, which allows it to continue.  
But just stripping the names and continuing was not enough to get the models converted properly, so it would cause a new error:
```
RuntimeError: shape '[16, 3, 1]' is invalid for input of size 1
```

This is because Qwen3.5's linear attention weights get reordered in `modify_tensors()`:
```
# original order:  [q, k, v, z] * head_count
# corrected order: [q * head_count, k * head_count, v * head_count, z * head_count]
```
However NVFP4 bypasses `modify_tensors()` and has its own repacking, and `linear_attn.in_proj_a.input_scale`  was seen by as a [num_v_heads] tensor and tried to reshape it into [16, 3, 1].
This is fixed by skipping tensors in the write loop that already were repacked 
```
if self._is_nvfp4:
                if name.endswith(".weight") and name.replace(".weight", ".weight_scale") in self.model_tensors:
                    continue
                if name.endswith((".weight_scale", ".weight_scale_2", ".input_scale", "k_scale", ".v_scale"))
                    continue
 Updated: added k_scale and v_scale above
```
and by applying the same reordering for :

```
linear_attn.in_proj_qkv
linear_attn.in_proj_z
linear_attn.in_proj_a
linear_attn.in_proj_b
linear_attn.out_proj
```

This will now produce the correct Qwen3.5/Qwen3.5MoE NVFP4 GGUF file.   A separate PR must be applied to load these files.
This fixed the issue with both Qwen3.5-122B-A10B-NVFP4 and Qwen3.5-27B-NVFP4 and correctly produced proper output.
[...]

PR #17906, opened 2025-12-10
CUDA: experimental native mxfp4 support for blackwell

am17an:
Currently WIP, trying to add native fp4 support for blackwell and beyond.

Blackwell has a `m16n8k64` instruction for 4 bit (mxfp4, nvfp4 and int4) which advertises 2x throughput compared to int8 tensor cores. However at the moment this PR is ~10% *slower* than master~ 25% faster than master on PP. The other issue is that we quantize activation to mxfp4 instead of q8, which lead to failures in `test-backend-ops`, however PPL tests are okay with this change (8x AIME25 eval has the same results as q8)

TODO: 
- [x] Figure out why we don't see better results
- [x] Address NMSE error b/w q8_0 and mxfp4

| Model                 | Test    |   t/s c6f6e4f96 |   t/s mxfp4 |   Speedup |
|:----------------------|:--------|----------------:|------------:|----------:|
| gpt-oss 20B MXFP4 MoE | pp512   |        10560.64 |    13304.80 |      1.26 |
| gpt-oss 20B MXFP4 MoE | pp1024  |        10659.15 |    13515.51 |      1.27 |
| gpt-oss 20B MXFP4 MoE | pp2048  |        10801.35 |    13715.10 |      1.27 |
| gpt-oss 20B MXFP4 MoE | pp4096  |        10854.04 |    13806.59 |      1.27 |
| gpt-oss 20B MXFP4 MoE | pp8192  |        10688.23 |    13525.14 |      1.27 |
| gpt-oss 20B MXFP4 MoE | pp16384 |        10140.17 |    11587.72 |      1.14 |

on RTX Pro 6000 Blackwell
| Model | Test | t/s c6f6e4f9 | t/s mxfp4 | Speedup |
| :--- | :--- | ---: | ---: | ---: |
| gpt-oss 120B MXFP4 MoE | pp2048 | 6663.49 | 8747.55 | 1.31 |
| gpt-oss 120B MXFP4 MoE | pp4096 | 6821.07 | 8913.47 | 1.31 |
| gpt-oss 120B MXFP4 MoE | pp8192 | 7079.22 | 9162.30 | 1.29 |
| gpt-oss 120B MXFP4 MoE | pp16384 | 6936.33 | 9039.58 | 1.30 |

> Note: This PR was developed on @JohannesGaessler's server with a 5090 provided by NVIDIA. So thanks to them both!

PR #21896, opened 2026-04-14
ggml-cuda: Blackwell native NVFP4 support 

michaelw9999:
`Description`:
This update is the first of several upcoming Blackwell NVFP4 features, with the first native MMA and MMQ kernel, significantly improving prefill performance over the generic version, using native NVFP4 with hardware accelerated blockscaling.  

This will run as NVFP4 x NVFP4 (NVFP4 for activations) only when supported by the hardware presence of a Blackwell GPU on a CUDA-compiled version.  This will not get executed by any other platform.  This version, while primarily only increasing prefill speed, maintains nearly equivalent token generation speed as the previous Q8 version, despite the heavier overhead of on the fly NVFP4 quantization for activations.  

Future PRs will provide additional support for NVFP4 activation scaling, MMVQ optimization, and AoSoA block repack which will also significantly increase performance, and other improvements.
 
`Performance boost`:
| Model | pp512 Before | pp512 After | pp512 Speedup | tg128 Before | tg128 After | tg128 Speedup |
|---|---:|---:|---:|---:|---:|---:|
| Qwen3.5 4B | 14495.32 ± 69.04 | 17418.76 ± 476.22 | **1.20×** (+20.2%) | 216.73 ± 0.46 | 216.97 ± 0.44 | **1.00×** (+0.1%) |
| Qwen3.5 27B | 3208.98 ± 0.43 | 4691.13 ± 159.11 | **1.46×** (+46.2%) | 65.64 ± 0.01 | 66.21 ± 0.14 | **1.01×** (+0.9%) |
| Nemotron-Cascade-2 30B | 7769.65 ± 23.77 | 10330.14 ± 3.90 | **1.33×** (+32.9%) | 226.04 ± 0.95 | 227.09 ± 1.38 | **1.00×** (+0.5%) |

There remains a small quantization loss due to using NVFP4 for activations.  On Qwen3.5-4B, this moved ppl from 11.40 on the baseline to 11.65 after this PR, and on Nemotron-Cascade-2-30B, from 9.81 to 9.85.  This was kept in check by doing a small +/- 2 code search during quantization to improve the subblock scale by finding which amongst them has the lowest error, which is negligible overheard, as it's calculated via GPU.  That improved ppl and lowered max kld, for example, from 12.24 to 11.65 (vs baseline 11.40 with Q8) on Qwen3.5 4B, and from 9.88 to 9.85 on Nemotron, and is likely worth any tiny overhead.  Like MXFP4, `test-backend-ops` was updated to add NVFP4 to the same override as certain tests would otherwise fail for excess error.
 
 AI assistance was used during this development for help with debugging, optimizing, and creating some portions, especially involving stride/tile/layout calculations. All code has been meticulously reviewed and edited by hand.

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -368,15 +368,21 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
     }
 
     ~ggml_cuda_pool_leg() {
+        clear_pool();
+        GGML_ASSERT(pool_size == 0);
+    }
+
+    void clear_pool() {
         ggml_cuda_set_device(device);
         for (int i = 0; i < MAX_BUFFERS; ++i) {
             ggml_cuda_buffer & b = buffer_pool[i];
             if (b.ptr != nullptr) {
                 CUDA_CHECK(cudaFree(b.ptr));
                 pool_size -= b.size;
+                b.ptr  = nullptr;
+                b.size = 0;
             }
         }
-        GGML_ASSERT(pool_size == 0);
     }
 
     void * alloc(size_t size, size_t * actual_size) override {
@@ -421,7 +427,20 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
         size_t look_ahead_size = (size_t) (1.05 * size);
         look_ahead_size = 256 * ((look_ahead_size + 255)/256);
         ggml_cuda_set_device(device);
-        CUDA_CHECK(ggml_cuda_device_malloc(&ptr, look_ahead_size, device));
+        cudaError_t err = ggml_cuda_device_malloc(&ptr, look_ahead_size, device);
+        if (err == cudaErrorMemoryAllocation) {
+            (void)cudaGetLastError();
+            const size_t cached_bytes = pool_size;
+            GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: alloc of %.2f MiB failed, flushing %.2f MiB of cached buffers and retrying\n",
+                           device, look_ahead_size/1024.0/1024.0, cached_bytes/1024.0/1024.0);
+            CUDA_CHECK(cudaDeviceSynchronize());
+            clear_pool();
+            err = ggml_cuda_device_malloc(&ptr, look_ahead_size, device);
+            if (err == cudaSuccess) {
+                GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: retry succeeded\n", device);
+            }
+        }
+        CUDA_CHECK(err);
         *actual_size = look_ahead_size;
         pool_size += look_ahead_size;
 #ifdef DEBUG_CUDA_MALLOC
--- a/ggml/src/ggml-cuda/vendors/hip.h
+++ b/ggml/src/ggml-cuda/vendors/hip.h
@@ -58,6 +58,7 @@
 #define cudaDeviceProp hipDeviceProp_t
 #define cudaDeviceSynchronize hipDeviceSynchronize
 #define cudaError_t hipError_t
+#define cudaErrorMemoryAllocation hipErrorOutOfMemory
 #define cudaErrorPeerAccessAlreadyEnabled hipErrorPeerAccessAlreadyEnabled
 #define cudaErrorPeerAccessNotEnabled hipErrorPeerAccessNotEnabled
 #define cudaEventCreateWithFlags hipEventCreateWithFlags
--- a/ggml/src/ggml-cuda/vendors/musa.h
+++ b/ggml/src/ggml-cuda/vendors/musa.h
@@ -42,6 +42,7 @@
 #define cudaDeviceProp musaDeviceProp
 #define cudaDeviceSynchronize musaDeviceSynchronize
 #define cudaError_t musaError_t
+#define cudaErrorMemoryAllocation musaErrorMemoryAllocation
 #define cudaErrorPeerAccessAlreadyEnabled musaErrorPeerAccessAlreadyEnabled
 #define cudaErrorPeerAccessNotEnabled musaErrorPeerAccessNotEnabled
 #define cudaEventCreateWithFlags musaEventCreateWithFlags


ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -849,6 +857,140 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+        x_u32[i * MMQ_MMA_TILE_X_K_NVFP4 + 64 + kbx] = get_int_b1(bxi->d, 0);
+    }
+}
+
+template <int mmq_x, int mmq_y>
+static __device__ __forceinline__ void vec_dot_nvfp4_nvfp4_mma(const int * __restrict__ x,
+                                                               const int * __restrict__ y,
+                                                               float * __restrict__ sum,
+                                                               const int k00) {
+    typedef tile<16, 8, int>   tile_A;
```
am17an, line 918: This function should not be needed. `vec_dot_mxfp4_mxfp4_mma` can be templated to `vec_dot_fp4_fp4_mma` with the two types, just the scale loading + mma call needs to be different.

ggml/src/ggml-cuda/common.cuh
```diff
@@ -834,6 +834,22 @@ static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
 #endif // defined(GGML_USE_HIP) && defined(CDNA3) && defined(FP8_AVAILABLE) && HIP_VERSION >= 60200000
 }
 
+static __device__ __forceinline__ uint8_t ggml_cuda_fp32_to_ue4m3(float x) {
+#if defined(BLACKWELL_MMA_AVAILABLE) // This is used for NVFP4 subblock scale quantizations only
+    if (!(x > 0.0f)) {
+        return 0;
+    }
+    if (x > 448.0f) {
```
pwilkin: Pretty please, no "magic constants".

michaelw9999: @pwilkin could put a macro, but it would still need to be the same value 448.0f is the maximum for UE4M3 when used for NVFP4 scales, that's set by Nvidia.  Here we're just matching `ggml_fp32_to_ue4m3` exactly as well.

pwilkin: Yeah, what I mean is there should be some semantics here (like UE4M3_MAX_VAL), magic constants are an antipattern.

ggml/src/ggml-cuda/mmq.cuh
michaelw9999: @am17an I think it should stay separate, as there are enough differences, let them have their own function matching their `GGML_TYPE`, it is not like they are interchangeable formats despite the similarities. If there are bugs for one and not another it would be easier to identify, and if we any changes or tuning only works for one, we would not need more branches which can be confusing in some places.  Also, the upcoming repack version (which is ready to go), if implemented, is totally different and supersedes `vec_dot_nvfp4_nvfp4_mma` anyway - then we would have to revert and separate out MXFP4 back again.

am17an: I don't think they should be separate in this PR. Repack will be not done for all cases, for eg when  weights are kept on the CPU. Bugs are exactly why you should reuse the same function which has been in master for quite sometime now.

michaelw9999: Okay, I'll work on integrating one new `fp4_fp4` function.  For the repack that I've already finished and is working, it is done on model loading, it does not do it on the fly as that would slow things down, so it does not keep the original layout in memory. So for that version, CPU still uses the AoSoA layout, which is not a problem, it is still faster and 16B aligned, just has a different kernel.

am17an: Weights are transferred during inference when offloading, I don't think what you're doing is going to work for all cases. In any case, don't worry about the repack for now.

michaelw9999: @am17an I'm been trying to make a merged `vec_dot_fp4_fp4_mma`, but to get it nice and clean for both NVFP4 and MXFP4 is not easy.
It is more than "just scale loading and mma call are different". 
They do not share the same X or Y layouts/strides, lane handling, or scale loading (regardless of the data type) Eg:
```
#define QK_MXFP4 32
typedef struct {
    uint8_t e; // E8M0
    uint8_t qs[QK_MXFP4/2];
} block_mxfp4;

#define QK_NVFP4 64
#define QK_NVFP4_SUB 16  // sub-block size for per-group scales
typedef struct {
    uint8_t d[QK_NVFP4/QK_NVFP4_SUB]; // UE4M3 scales (4 bytes, one per 16-element sub-block)
    uint8_t qs[QK_NVFP4/2];           // packed 4-bit E2M1 values (32 bytes)
} block_nvfp4;

#define MMQ_MMA_TILE_X_K_FP4   (2*MMQ_TILE_NE_K + 8 + 4) // MXFP4   == 76
#define MMQ_MMA_TILE_X_K_NVFP4 (2*MMQ_TILE_NE_K + MMQ_TILE_NE_K/2 + 4) // NVFP4 == 84

scaleA[n][k01 / (2 * QI_MXFP4)] =
    *(x_sc + (i0 + n * tile_A::I + tidx) * MMQ_MMA_TILE_X_K_FP4 + k0 / (2 * QI_MXFP4)); 
    
scaleA[n][frag] =
    x_sc[(i0 + n * tile_A::I + tidx) * MMQ_MMA_TILE_X_K_NVFP4 + k0 / tile_A::J];
    
uint32_t scaleB; // mxfp4
uint32_t scaleB[nfrags]; // nvfp4 
```

I was trying for quite some time, but getting junk output on the result.  Their differences really make it difficult to be easily interchangeable.
Let's please leave them alone, it will take a lot of time otherwise :))

am17an: I think you need to rework the implementation till they're identical. There is no point supporting both of them when they have the same physical layout, it is an unnecessary maintenance burden.  You should've used `MMQ_MMQ_TILE_X_K_FP4` as it leaves space 2 extra scales. If you're unwilling to do it I will do it, but it will need to be the same.

am17an: You can use something like this: 3aaa49f86b40323668e89ae028f18c15800f2e4f

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -70,6 +70,98 @@ __device__ __forceinline__ uint8_t compute_e8m0_scale(float amax) {
...
+    static constexpr int test_offsets[5] = { 0, -1, 1, -2, 2};
+    const int first_fp8_code = (int) ggml_cuda_fp32_to_ue4m3(sub_max * (1.0f / 6.0f));
+
+    float best_err = FLT_MAX;
+    uint8_t fp8_code = 0;
+    float subblock_scale = 0.0f;
+
+#pragma unroll // Check +/- 2 to find best code to reduce NVFP4 activation loss. Negligible overhead on Blackwell.
+    for (int i = 0; i < 5; i++) {
+        const int test_code = first_fp8_code + test_offsets[i];
```
gaugarg-nv, line 125: Do we need a check here to ensure `test_code` remains within the ue4m3 range?

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -51,9 +52,15 @@ struct block_fp4_mmq {
     int8_t   qs[4 * 32];  // 256 FP4 values packed as 4-bit pairs (2 per byte), 8 blocks of 32 values
 };
 
+struct block_nvfp4_mmq {  // Same byte size as block_q8_1_mmq and block_fp4_mmq
+    uint32_t sc4_u32[4];  // 16 UE4M3 scales (1 per 1 values), packed into 4 unint32
```
ynankani: This should be 1 per 16 values?  and instead of unint32 it should be uint32

ggml/src/ggml-cuda/mma.cuh
```diff
@@ -1062,6 +1062,32 @@ namespace ggml_cuda_mma {
 #endif  // BLACKWELL_MMA_AVAILABLE
     }
 
+    static __device__ __forceinline__ void mma_nvfp4_blackwell(
```
am17an: rename this to `mma_block_scaled_nvfp4` and other one to `mma_block_scaled_mxfp4`

ggml/src/ggml-cuda/mmq.cu
```diff
@@ -120,6 +120,7 @@ void ggml_cuda_mul_mat_q(
 
     // TODO: tighter pool buffer size vs q8 path
     const bool use_native_mxfp4 = blackwell_mma_available(cc) && src0->type == GGML_TYPE_MXFP4;
+    const bool use_native_nvfp4 = blackwell_mma_available(cc) && src0->type == GGML_TYPE_NVFP4;
```
am17an: remove `use_native_mxfp4` and rename this to `use_native_fp4 = blackwell_mma_available(cc) && src0->type == (GGML_TYPE_MXFP4 || GGML_TYPE_NVFP4)`, remove all the if/else statements in this file

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -51,9 +52,15 @@ struct block_fp4_mmq {
     int8_t   qs[4 * 32];  // 256 FP4 values packed as 4-bit pairs (2 per byte), 8 blocks of 32 values
 };
 
+struct block_nvfp4_mmq {  // Same byte size as block_q8_1_mmq and block_fp4_mmq
```
am17an: remove this, we're using `block_fp4_mmq`, change the comment for reflect the scales are different between mxfp4/nvfp4

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -51,9 +52,15 @@ struct block_fp4_mmq {
...
+    uint32_t sc4_u32[4];  // 16 UE4M3 scales (1 per 1 values), packed into 4 unint32
+    uint32_t qs_u32[32];  // 256 NVFP4 values packed as 4-bit pairs (128 payload bytes)
+};
+
 static_assert(sizeof(block_q8_1_mmq) == 4*QK8_1 + 4*sizeof(half2), "Unexpected block_q8_1_mmq size");
 static_assert(sizeof(block_q8_1_mmq) == 4*sizeof(block_q8_1),      "Unexpected block_q8_1_mmq size");
 static_assert(sizeof(block_fp4_mmq)  == sizeof(block_q8_1_mmq),    "Unexpected block_fp4_mmq size");
+static_assert(sizeof(block_nvfp4_mmq) == sizeof(block_q8_1_mmq),   "Unexpected block_nvfp4_mmq size");
```
am17an: remove

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -12,6 +12,7 @@ using namespace ggml_cuda_mma;
 #define MMQ_DP4A_MAX_BATCH_SIZE 64 // Max. batch size to use for dp4a MMQ kernels when FP16 tensor cores are available.
 #define MMQ_ITER_K 256
 #define MMQ_ITER_K_MXFP4_FP4    512
+#define MMQ_ITER_K_NVFP4_FP4    512
```
am17an: rename this to `MMQ_ITER_K_FP4` and use the same value throughout the file, similar comment as in mmq.cu

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -70,6 +70,98 @@ __device__ __forceinline__ uint8_t compute_e8m0_scale(float amax) {
...
+    const float sub_max = fmaxf(amax_raw, __shfl_xor_sync(0xFFFFFFFFu, amax_raw, 1));
+    static constexpr int test_offsets[5] = { 0, -1, 1, -2, 2};
+    const int first_fp8_code = (int) ggml_cuda_fp32_to_ue4m3(sub_max * (1.0f / 6.0f));
+
+    float best_err = FLT_MAX;
+    uint8_t fp8_code = 0;
+    float subblock_scale = 0.0f;
+
+#pragma unroll // Check +/- 2 to find best code to reduce NVFP4 activation loss. Negligible overhead on Blackwell.
+    for (int i = 0; i < 5; i++) {
```
am17an, line 124: I'm unsure what value this provides, is this the standard way to quantize to nvfp4? If not then please remove it, avoid novelty wherever possible please.

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -849,6 +857,145 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
     }
 }
 
+#ifdef BLACKWELL_MMA_AVAILABLE
+template <int mmq_y, bool need_check>
+static __device__ __forceinline__ void load_tiles_nvfp4_nvfp4(const char * __restrict__ x,
+                                                            int * __restrict__ x_tile,
```
am17an, line 863: I think you quantize like mxfp4, i.e. re-order the nibbles during quantization, it will provide a meaningful speedup.

ggml/src/ggml-cuda/common.cuh
```diff
@@ -834,6 +834,23 @@ static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
...
+static __device__ __forceinline__ uint8_t ggml_cuda_fp32_to_ue4m3(float x) {
+#if defined(BLACKWELL_MMA_AVAILABLE) // This is used for NVFP4 subblock scale quantizations only
+    if (!(x > 0.0f)) {
+        return 0;
+    }
+    if (x > UE4M3_MAX) {
+        x = UE4M3_MAX;
+    }
+
+    const __nv_fp8_e4m3 xf(x);
```
ORippler: 1. Ternary operations are easier for the compiler to handle, as they bound the depth of the branch (basically one can do `x = x > UE4M3_MAX ? UE4M3_MAX : x;`
2. __nv_fp8_e4m3 constructor will clamp to the max value itself, so this is not needed: https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/struct____nv__fp8__e4m3.html#_CPPv4N13__nv_fp8_e4m313__nv_fp8_e4m3EKf

ggml/src/ggml-cuda/common.cuh
```diff
@@ -834,6 +834,23 @@ static __device__ __forceinline__ float ggml_cuda_ue4m3_to_fp32(uint8_t x) {
 #endif // defined(GGML_USE_HIP) && defined(CDNA3) && defined(FP8_AVAILABLE) && HIP_VERSION >= 60200000
 }
 
+#define UE4M3_MAX 448.0f // The maximum  representable value for Ue4m3 is 448.0f for NVFP4 FP8 scales
+static __device__ __forceinline__ uint8_t ggml_cuda_fp32_to_ue4m3(float x) {
+#if defined(BLACKWELL_MMA_AVAILABLE) // This is used for NVFP4 subblock scale quantizations only
+    if (!(x > 0.0f)) {
+        return 0;
+    }
```
ORippler, line 841: What's the motivation for seemingly discarding the negative values? FP8 has 1 sign bit (e4m3 + 1 sign bit = 8 bits)

https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html#Structure

ggml/src/ggml-cuda/mmq.cuh
am17an: btw when you merge code the standard practice to keep authorship, no need to change anything now

ORippler (changes requested):
First of all, thanks for the PR and your contribution!

> There remains a small quantization loss due to using NVFP4 for activations. On Qwen3.5-4B, this moved ppl from 11.40 on the baseline to 11.65 after this PR, and on Nemotron-Cascade-2-30B, from 9.81 to 9.85. This was kept in check by doing a small +/- 2 code search during quantization to improve the subblock scale by finding which amongst them has the lowest error, which is negligible overheard, as it's calculated via GPU. That improved ppl and lowered max kld, for example, from 12.24 to 11.65 (vs baseline 11.40 with Q8) on Qwen3.5 4B, and from 9.88 to 9.85 on Nemotron, and is likely worth any tiny overhead. Like MXFP4, test-backend-ops was updated to add NVFP4 to the same override as certain tests would otherwise fail for excess error.

Unless I am misreading the PR we are:
1. Missing [per-tensor F32 scale](https://developer.nvidia.com/blog/introducing-nvfp4-for-efficient-and-accurate-low-precision-inference/) for the activation quantization (Ideally, this should be applied in the GEMM-epilogue before write-back. Theoretically, we can also grab the [per-tensor scale we apply as a separate GGML_OP for the weights](https://github.com/ggml-org/llama.cpp/pull/19769#discussion_r2914265622) and apply both of them at the same time, though this would require adding node-fusion (and may have been done by someone else in the mean-time)).
2. Discarding negative values for the per-block FP8 scales

My feeling is we should resolve both for functional correctness/best available quality

ggml/src/ggml-cuda/common.cuh
ynankani, line 860: Here should it be using binary search over bounds instead of linear search similar to what trt-llm is doing [https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/custom_ops/quantization/torch_quant.py]?

michaelw9999:
> First of all, thanks for the PR and your contribution!
> 
> > There remains a small quantization loss due to using NVFP4 for activations. On Qwen3.5-4B, this moved ppl from 11.40 on the baseline to 11.65 after this PR, and on Nemotron-Cascade-2-30B, from 9.81 to 9.85. This was kept in check by doing a small +/- 2 code search during quantization to improve the subblock scale by finding which amongst them has the lowest error, which is negligible overheard, as it's calculated via GPU. That improved ppl and lowered max kld, for example, from 12.24 to 11.65 (vs baseline 11.40 with Q8) on Qwen3.5 4B, and from 9.88 to 9.85 on Nemotron, and is likely worth any tiny overhead. Like MXFP4, test-backend-ops was updated to add NVFP4 to the same override as certain tests would otherwise fail for excess error.
> 
> Unless I am misreading the PR we are:
> 
> 1. Missing [per-tensor F32 scale](https://developer.nvidia.com/blog/introducing-nvfp4-for-efficient-and-accurate-low-precision-inference/) for the activation quantization (Ideally, this should be applied in the GEMM-epilogue before write-back. Theoretically, we can also grab the [per-tensor scale we apply as a separate GGML_OP for the weights](https://github.com/ggml-org/llama.cpp/pull/19769#discussion_r2914265622) and apply both of them at the same time, though this would require adding node-fusion (and may have been done by someone else in the mean-time)).
> 2. Discarding negative values for the per-block FP8 scales
> 
> My feeling is we should resolve both for functional correctness/best available quality

Hi @ORippler!

1) I have the code for using input scale ready to go (although needs some rework now with recent changes) but was holding off to put that into a separate PR,  I was trying to keep this PR small as possible.  I already have the input scale work prepared via https://github.com/ggml-org/llama.cpp/pull/20505 which is already merged already.
So we have right now, for example, every tensor that has a weight_scale also brings in the matching input_scale:

`layer.wq_in_s = create_tensor(tn(LLM_TENSOR_ATTN_Q,   "input_scale", i), {1}, TENSOR_NOT_REQUIRED);`

They are loaded if present in the GGUF and are just waiting to be used.  I updated the hf convert script to pull them in from any source from same PR above.  My own NVFP4 quantizer (not in any PR but I have published some quantized GGUF NVFP4 models onto HF) uses imatrix to derive the input scale.

2) There should never be negative values since it's unsigned `UE4M3` for the NVFP4 scales

ggml/src/ggml-cuda/mmq.cuh
michaelw9999: Working on this, will have an update soon.

vishalpandya1990:
> doing a small +/- 2 code search during quantization to improve the subblock scale by finding which amongst them has the lowest error

Can we evaluate [4/6 scaling](https://arxiv.org/pdf/2512.02010) for optimal block-scale search?

michaelw9999:
> > doing a small +/- 2 code search during quantization to improve the subblock scale by finding which amongst them has the lowest error
> 
> Can we evaluate [4/6 scaling](https://arxiv.org/pdf/2512.02010) for optimal block-scale search?

Yes, I've already used that in a very old implementation POC, it can be added in a future PR.

ggml/src/ggml-cuda/common.cuh
ORippler: Since each FP4 nibble in a block can be both positive and negative, the expressivity is equivalent for UE4M3 and E4M3. From a gradient-flow perspective, it most likely provides more stable gradients during training to just scale by amax, and this is done by modelopt. We should add a guard against this during conversion, but that's beyond the scope of this PR

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -70,6 +70,98 @@ __device__ __forceinline__ uint8_t compute_e8m0_scale(float amax) {
...
+
+    const int64_t i0_base = ((int64_t) blockDim.x * blockIdx.y + threadIdx.x) * 8;
+    if (i0_base >= ne0) {
+        return;
+    }
+
+    const int64_t i1 = blockIdx.x;
+    const int64_t i2 = blockIdx.z % ne2;
+    const int64_t i3 = blockIdx.z / ne2;
+    const int64_t i01 = has_ids ? ids[i1] : i1;
```
ORippler: Let's check on ids not being nullptr instead of templating to avoid binary bloat (templating shouldn't be necessary as this can be represented by a ternary operator)

ggml/src/ggml-cuda/quantize.cu
michaelw9999: I did not invent the idea, I found it in `make_qkx2_quants()` already in llama.cpp.  It is almost free, I don't see any tangible change in perf with every bench still staying within the same band, perhaps lower a fraction, but negligible enough to not matter.
The real value is that it improves activation loss vs Q8. It reduces ppl on Qwen3.5/4B from 11.80 to 11.65 [Q8 is 11.40], reduces max kld from 12.0 to 11.38 (5.2%) , mean ln 0.075354  from 0.087845 (14% ), mean kld 0.092041 from 0.098035  (6.5% ), RMS Δp 7.925 from 8.136, and every parameter is improved, so this will help any model's quality in some way and reduces outliers.  I can take it out if you want, or put it to onto its own PR or just let it go.

ggml/src/ggml-cuda/common.cuh
michaelw9999: Removed the check.

ggml/src/ggml-cuda/quantize.cu
ORippler: Afaik dynamic activation quantization would be the standard. However, this would require computing max value of the whole tensor to figure out the per-tensor scale:

https://github.com/vllm-project/vllm/blob/3beb57a238b82fe90e8b99e009c876343b9d9703/benchmarks/kernels/benchmark_nvfp4_gemm.py#L67-L70

ggml/src/ggml-cuda/mmq.cuh
am17an: why did you mark this comment as resolved?

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -341,3 +433,24 @@ void quantize_mmq_mxfp4_cuda(const float *                    x,
 
     quantize_mmq_mxfp4<<<num_blocks, block_size, 0, stream>>>(x, ids, vy, ne00, s01, s02, s03, ne0, ne1, ne2);
 }
+
+void quantize_mmq_nvfp4_cuda(
+        const float * x, const int32_t * ids, void * vy, const ggml_type type_src0,
+        const int64_t ne00, const int64_t s01, const int64_t s02, const int64_t s03,
+        const int64_t ne0, const int64_t ne1, const int64_t ne2, const int64_t ne3, cudaStream_t stream) {
+    GGML_ASSERT(type_src0 == GGML_TYPE_NVFP4);
+    GGML_ASSERT(ne00 % 8 == 0);
```
ORippler: Where does this seemingly magic number 8 come from, and how does it relate to the reference CPU impl.

```
void quantize_row_nvfp4_ref(const float * GGML_RESTRICT x, block_nvfp4 * GGML_RESTRICT y, int64_t k) {
    static const int qk = QK_NVFP4;
    static const int qk_sub = QK_NVFP4_SUB;
    static const int n_sub = QK_NVFP4 / QK_NVFP4_SUB;

    assert(k % qk == 0);

    const int nb = k / qk;
```

ggml/src/ggml-cuda/mmq.cu
```diff
@@ -205,8 +203,8 @@ void ggml_cuda_mul_mat_q(
         CUDA_CHECK(cudaGetLastError());
     }
 
-    const int64_t s12 = use_native_mxfp4 ? ne11 * ne10_padded * sizeof(block_fp4_mmq) / (8 * QK_MXFP4 * sizeof(int)) :
-                                           ne11 * ne10_padded * sizeof(block_q8_1) / (QK8_1 * sizeof(int));
+    const int64_t s12 = use_native_fp4 ? ne11 * ne10_padded * sizeof(block_fp4_mmq) / (QK_K * sizeof(int)) :
```
am17an, line 207: add assert here that `QK_K == 8 * QK_MXFP4`

ggml/src/ggml-cuda/mmq.cuh
michaelw9999: @am17an I just pushed a whole new update, I refactored it all to use `MMQ_MMA_TILE_X_K_FP4` instead of `MMQ_MMA_TILE_X_K_NVFP4` so it's on the matching 76 byte tile instead of the 84 and got rid of the block_nvfp4_mmq, It reduces a lot of code, let me know what you think of the changes

ggml/src/ggml-cuda/quantize.cu
michaelw9999: @ORippler I have a lot up my sleeve for NVFP4 I've been working on but I did not want to bring out too much all at once, I made that mistake a while back :)) You will like the new repack I think.

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -46,8 +46,8 @@ struct block_q8_1_mmq {
     int8_t qs[4*QK8_1]; // 128 values quantized to 8 bit each
 };
 
-struct block_fp4_mmq {
-    uint32_t d4[4];       // 8 E8M0 scales (1 per 32 values), 2 packed per uint32: d4[0]={s0,s1}, d4[1]={s2,s3}, etc.
+struct block_fp4_mmq {  // Used for both MXFP4[8 E8M0 scales (1 per 32)) and NVFP4 (16 UE4M3 scales (1 per 16))
+    uint32_t d4[4];       // 2 packed per uint32: d4[0]={s0,s1}, d4[1]={s2,s3}, etc.
     int8_t   qs[4 * 32];  // 256 FP4 values packed as 4-bit pairs (2 per byte), 8 blocks of 32 values
 };
```
am17an:
```suggestion
// this struct is used for fp4 data types (currently only used in Blackwell)
// mxfp4 has block size 32, each int32 of d4 contains 2 e8m0 scales in the lower 16 bits
// nvfp4 has block size 16, each int32 of d4 contains 4 ue4m3 scales
struct block_fp4_mmq { 
    uint32_t d4[4];
    int8_t   qs[4 * 32];  // 256 FP4 values packed as 4-bit pairs (2 per byte)
};
```

ggml/src/ggml-cuda/quantize.cu
michaelw9999: 8 is supposed to represent the minimum number of FP4 values (1 `uint32_t`).   We probably don't need it checking `i00 < ne00` for partials

ggml/src/ggml-cuda/quantize.cu
am17an: Before merging this PR we can do a eval check like done here https://github.com/ggml-org/llama.cpp/pull/17906#issuecomment-3681384998 for a Nemotron model which has results for AIME-25 or some similar eval. If we get similar numbers as the original model then I think this should be fine.

ggml/src/ggml-cuda/mmq.cuh
am17an: Okay yes I see, does this provide a speed-up?

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -849,6 +856,125 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+
+        const uint32_t * src_qs = reinterpret_cast<const uint32_t *>(bxi->qs);
+
+#pragma unroll
+        for (int sub = 0; sub < QK_NVFP4 / QK_NVFP4_SUB; ++sub) {
+            x_u32[q_base + 2 * sub + 0] = src_qs[2 * sub + 0];
+            x_u32[q_base + 2 * sub + 1] = src_qs[2 * sub + 1];
+        }
+
+        x_u32[row_base + 64 + kbx] = get_int_b1(bxi->d, 0);
```
am17an: this should be 4 byte aligned so I think `get_int_b4` should be safe

ggml/src/ggml-cuda/mmq.cu
```diff
@@ -143,10 +143,8 @@ void ggml_cuda_mul_mat_q(
         }
 
         // Stride depends on quantization format
-        const int64_t s12 = use_native_mxfp4 ?
-                                ne11 * ne10_padded * sizeof(block_fp4_mmq) /
-                                    (8 * QK_MXFP4 * sizeof(int))  // block_fp4_mmq holds 256 values (8 blocks of 32)
-                                :
+        const int64_t s12 = use_native_fp4 ?
+                                ne11 * ne10_padded * sizeof(block_fp4_mmq) / (QK_K * sizeof(int)) :  // block_fp4_mmq holds 256 values (8 blocks of 32)
```
am17an:
```suggestion
                                ne11 * ne10_padded * sizeof(block_fp4_mmq) / (QK_K * sizeof(int)) :  // block_fp4_mmq holds 256 values
```

ggml/src/ggml-cuda/mmq.cuh
michaelw9999: Yes, there's a nice prefill boost, but it reduced tokengen bit, it has to do packing work to fit block_fp4_mmq.  
| nemotron_h_moe 31B.A3.5B NVFP4 |  19.28 GiB |    31.58 B | CUDA       |  99 |           pp512 |     11773.18 ± 41.83 |
| nemotron_h_moe 31B.A3.5B NVFP4 |  19.28 GiB |    31.58 B | CUDA       |  99 |           tg128 |        223.31 ± 1.54 |

ggml/src/ggml-cuda/quantize.cu
ORippler: > 8 is supposed to represent the minimum number of FP4 values (1 uint32_t). We probably don't need it checking i00 < ne00 for partials

The general intent is to stick towards the types declared in `enum ggml_type`, so that would mean restricting to valid `block_nvfp4` only.

Also, feature dimensions will be constant, so weight conversion should fail for partials (supporting partials is a general open)

ggml/src/ggml-cuda/mmq.cuh
am17an: This path is *not* used when doing token gen, that's mmvq which does not quantize to nvfp4. So this PR should have no effect on TG

ggml/src/ggml-cuda/mmq.cuh
michaelw9999: Got a little speedup from this change, great!!

ggml/src/ggml-cuda/quantize.cu
michaelw9999: Yes, that is a good idea, and I put that in.

ggml/src/ggml-cuda/common.cuh
michaelw9999: I do not think that would be needed, this is a very small search on device.

am17an (approved):
on a 5090:

| Model                          |   Microbatch size | Test   |   t/s e21cdc11a |   t/s nvfp4-blackwell |   Speedup |
|:-------------------------------|------------------:|:-------|----------------:|----------------------:|----------:|
| nemotron_h_moe 31B.A3.5B NVFP4 |               512 | pp2048 |         6941.51 |              11039.13 |      1.59 |
| nemotron_h_moe 31B.A3.5B NVFP4 |              1024 | pp2048 |         9209.10 |              13265.28 |      1.44 |
| nemotron_h_moe 31B.A3.5B NVFP4 |              2048 | pp2048 |        10423.61 |              14035.40 |      1.35 |

DGX spark:

| Model                          |   Microbatch size | Test   |   t/s 8dc530b86 |   t/s nvfp4-blackwell |   Speedup |
|:-------------------------------|------------------:|:-------|----------------:|----------------------:|----------:|
| nemotron_h_moe 31B.A3.5B NVFP4 |               512 | pp2048 |         1913.84 |               2900.42 |      1.52 |
| nemotron_h_moe 31B.A3.5B NVFP4 |              1024 | pp2048 |         2458.35 |               3188.62 |      1.30 |
| nemotron_h_moe 31B.A3.5B NVFP4 |              2048 | pp2048 |         2793.08 |               3430.60 |      1.23 |

stevelikesrhino:
Sorry to ask but this is work in progress right? Tried to manually merge the pr as well as cloning your fork and in both cases it produces infinite loop for bartowski’s qwen3.5 27B Q6 K L, and for gemma4 31B it produces gibberish tokens. The prompt prefill is very impressive

michaelw9999:
Steven I am downloading that exact model now and will try to reproduce the
error

stevelikesrhino:
> Steven I am downloading that exact model now and will try to reproduce the error

If it’s not wip, i should give more context. I’m on windows 11 with visual studio 22 and cuda toolkit 13.1. RTX 5090

am17an:
This PR shouldn't effect non nvfp4 models

michaelw9999:
Just something I've seen occasionally over the last few months have
happened... reboot Windows and see if it still does that on a fresh start.

stevelikesrhino:
Restarted and still does this. 
```
PS F:\llama-cpp\RC> nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Fri_Nov__7_19:25:04_Pacific_Standard_Time_2025
Cuda compilation tools, release 13.1, V13.1.80
Build cuda_13.1.r13.1/compiler.36836380_0
PS F:\llama-cpp\RC>
PS F:\llama-cpp\RC> nvidia-smi
Thu Apr 16 12:07:09 2026
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 591.74                 Driver Version: 591.74         CUDA Version: 13.1     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                  Driver-Model | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 5090      WDDM  |   00000000:01:00.0  On |                  N/A |
|  0%   32C    P8             28W /  575W |     997MiB /  32607MiB |      1%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A            6076    C+G   ...s\Mozilla Firefox\firefox.exe      N/A      |
|    0   N/A  N/A            9232    C+G   ...crosoft OneDrive\OneDrive.exe      N/A      |
|    0   N/A  N/A            9912    C+G   C:\Windows\explorer.exe               N/A      |
|    0   N/A  N/A            9976    C+G   ...indows\System32\ShellHost.exe      N/A      |
|    0   N/A  N/A           10168    C+G   ...2txyewy\CrossDeviceResume.exe      N/A      |
|    0   N/A  N/A           11176    C+G   ...yb3d8bbwe\WindowsTerminal.exe      N/A      |
|    0   N/A  N/A           12904    C+G   ..._cw5n1h2txyewy\SearchHost.exe      N/A      |
|    0   N/A  N/A           12912    C+G   ...y\StartMenuExperienceHost.exe      N/A      |
|    0   N/A  N/A           14740    C+G   ....0.3912.60\msedgewebview2.exe      N/A      |
|    0   N/A  N/A           17064    C+G   ...5n1h2txyewy\TextInputHost.exe      N/A      |
|    0   N/A  N/A           17316    C+G   ...8bbwe\PhoneExperienceHost.exe      N/A      |
|    0   N/A  N/A           17740    C+G   ...x40ttqa\iCloud\iCloudHome.exe      N/A      |
|    0   N/A  N/A           19016    C+G   ...ndows\System32\mmgaserver.exe      N/A      |
|    0   N/A  N/A           22116    C+G   ...s\Mozilla Firefox\firefox.exe      N/A      |
+-----------------------------------------------------------------------------------------+
PS F:\llama-cpp\RC>
PS F:\llama-cpp\RC> .\llama-cli.exe -m ..\models\google_gemma-4-31B-it-Q5_K_L.gguf
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 32606 MiB):
  Device 0: NVIDIA GeForce RTX 5090, compute capability 12.0, VMM: yes, VRAM: 32606 MiB

Loading model...


▄▄ ▄▄
██ ██
                                    ██    ██
                                    ▀▀    ▀▀

build      : b8801-c9988d9f5
model      : google_gemma-4-31B-it-Q5_K_L.gguf
modalities : text

available commands:
  /read <file>        add a text file
  /glob <pattern>     add text files using globbing pattern


> Hi

//
//
//
//
//
//
//
//
//
//
//
//
//
//
//

[ Prompt: 79.2 t/s | Generation: 59.9 t/s ]

>

```

am17an:
@stevelikesrhino please wait till the PR is merged

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -141,7 +144,7 @@ static int get_mmq_y_host(const int cc) {
 
 static constexpr __device__ int get_iter_k([[maybe_unused]] const ggml_type type) {
 #if defined(BLACKWELL_MMA_AVAILABLE)
-    return type == GGML_TYPE_MXFP4 ? MMQ_ITER_K_MXFP4_FP4 : MMQ_ITER_K;
+    return MMQ_ITER_K_FP4;
```
am17an, line 148: @michaelw9999 this is a bug

am17an: should be `return (type == GGML_TYPE_MXFP4 || type == GGML_TYPE_NVFP4) ? MMQ_ITER_K_FP4 : MMQ_ITER_K;`

michaelw9999:
Fixed, pushed and tested with Qwen_Qwen3.5-27B-Q6_K_L.gguf, now it's
working OK.

michaelw9999:
Ahh I just saw this suggestion now, I was too quick, I had written it
slightly differently, but equivalent

ggml/src/ggml-cuda/mmq.cu
```diff
@@ -208,8 +203,9 @@ void ggml_cuda_mul_mat_q(
         CUDA_CHECK(cudaGetLastError());
     }
 
-    const int64_t s12 = use_native_mxfp4 ? ne11 * ne10_padded * sizeof(block_fp4_mmq) / (8 * QK_MXFP4 * sizeof(int)) :
-                                           ne11 * ne10_padded * sizeof(block_q8_1) / (QK8_1 * sizeof(int));
+    GGML_ASSERT(QK_K == 8 * QK_MXFP4);
```
ORippler:
```suggestion
    static_assert(QK_K == 8 * QK_MXFP4, "QK_K needs to be 8 * QK_MXFP4");
```

QK_K and QK_MXFP4 are compile-time constants, so we can move this check to compile time

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -11,7 +11,7 @@ using namespace ggml_cuda_mma;
 
 #define MMQ_DP4A_MAX_BATCH_SIZE 64 // Max. batch size to use for dp4a MMQ kernels when FP16 tensor cores are available.
 #define MMQ_ITER_K 256
-#define MMQ_ITER_K_MXFP4_FP4    512
+#define MMQ_ITER_K_FP4    512
```
ORippler: Nit: Inconsistent formating with the above declaration (whitespace errors?)

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -232,15 +233,18 @@ static_assert(MMQ_MMA_TILE_X_K_NVFP4 % 8 == 4, "Wrong padding.");
...
         case GGML_TYPE_Q5_0:    return MMQ_MMA_TILE_X_K_Q8_0;
         case GGML_TYPE_Q5_1:    return MMQ_MMA_TILE_X_K_Q8_1;
         case GGML_TYPE_Q8_0:    return MMQ_MMA_TILE_X_K_Q8_0;
         // tile sizes are the same for Q8_1 and FP4 for blackwell
         case GGML_TYPE_MXFP4:   return MMQ_MMA_TILE_X_K_Q8_1;
-        case GGML_TYPE_NVFP4:   return MMQ_MMA_TILE_X_K_NVFP4;
+#if defined(BLACKWELL_MMA_AVAILABLE)
+        case GGML_TYPE_NVFP4:   return MMQ_MMA_TILE_X_K_FP4;
+#else
+       case GGML_TYPE_NVFP4:   return MMQ_MMA_TILE_X_K_NVFP4;
```
ORippler:
```suggestion
        case GGML_TYPE_NVFP4:   return MMQ_MMA_TILE_X_K_NVFP4;
```
Nit: Whitespace

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -934,6 +857,125 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+
+#pragma unroll
+    for (int i0 = 0; i0 < mmq_y; i0 += rows_per_warp * nwarps) {
+        int i = i0 + threadIdx.y * rows_per_warp + row_in_warp;
+
+        if constexpr (need_check) {
+            i = min(i, i_max);
+        }
+
+        const block_nvfp4 * bxi = (const block_nvfp4 *) x + kbx0 + i * stride + kbx;
```
ORippler: We could make it easier for the compiler by manually hoisting invariants kbx0 and kbx out of the loop

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -934,6 +857,125 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+
+        const uint32_t * src_qs = reinterpret_cast<const uint32_t *>(bxi->qs);
+
+#pragma unroll
+        for (int sub = 0; sub < QK_NVFP4 / QK_NVFP4_SUB; ++sub) {
+            x_u32[q_base + 2 * sub + 0] = src_qs[2 * sub + 0];
+            x_u32[q_base + 2 * sub + 1] = src_qs[2 * sub + 1];
+        }
+
+        x_u32[row_base + 64 + kbx] = get_int_b4(bxi->d, 0);
```
ORippler: Same here, can hoist advancement by 64 and kbx

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -934,6 +857,125 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+    constexpr int rows_per_warp = 2 * granularity;
+    constexpr int ntx           = rows_per_warp / tile_C::I;
+    constexpr int nfrags        = MMQ_TILE_NE_K / tile_A::J;
+
+    y += (threadIdx.y % ntx) * (tile_C::J * MMQ_TILE_Y_K);
+
+    const int *      x_qs = (const int *) x;
+    const uint32_t * x_sc = (const uint32_t *) (x_qs + 2 * MMQ_TILE_NE_K);
+    const int *      y_qs = (const int *) y + 4;
+    const uint32_t * y_sc = (const uint32_t *) y;
```
ORippler, line 933: x and y are already `const int *`, why recast?

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -934,6 +857,125 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+    tile_A   A[ntx][nfrags];
+    uint32_t scaleA[ntx][nfrags];
+
+#pragma unroll
+    for (int n = 0; n < ntx; ++n) {
+#pragma unroll
+        for (int frag = 0; frag < nfrags; ++frag) {
+            const int k0 = k00 + frag * tile_A::J;
+            load_ldmatrix(A[n][frag], x_qs + (i0 + n * tile_A::I) * stride + k0, stride);
+            scaleA[n][frag] = x_sc[(i0 + n * tile_A::I + tidx_A) * stride + k0 / tile_A::J];
```
ORippler, line 950: Can hoist `(i0 + tidx_A) * stride` from `x_sc` and `i0 * stride` from `x_qs` out of both loops

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -934,6 +857,125 @@ static __device__ __forceinline__ void load_tiles_mxfp4_fp4(const char * __restr
...
+#pragma unroll
+    for (int j0 = 0; j0 < mmq_x; j0 += ntx * tile_C::J) {
+        tile_B   B[nfrags];
+        uint32_t scaleB[nfrags];
+
+#pragma unroll
+        for (int frag = 0; frag < nfrags; ++frag) {
+            const int k0 = frag * tile_B::J;
+            load_generic(B[frag], y_qs + j0 * MMQ_TILE_Y_K + k0, MMQ_TILE_Y_K);
+            scaleB[frag] = y_sc[(j0 + tidx_B) * MMQ_TILE_Y_K + frag];
```
ORippler, line 963: Can hoist `j0 * MMQ_TILE_Y_K` one loop up and `tidx_B * MMQ_TILE_Y_K` two loops up

ggml/src/ggml-cuda/quantize.cuh
```diff
@@ -26,7 +26,7 @@ void quantize_mmq_q8_1_cuda(
         ggml_type type_src0, int64_t ne00, int64_t s01, int64_t s02, int64_t s03,
         int64_t ne0, int64_t ne1, int64_t ne2, int64_t ne3, cudaStream_t stream);
 
-void quantize_mmq_mxfp4_cuda(const float *   x,
+void quantize_mmq_fp4_cuda(const float *   x,
                              const int32_t * ids,
                              void *          vy,
                              ggml_type       type_src0,
```
ORippler, line 32: Nit: Please fix whitespace errors introduced for the function args

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -316,28 +412,32 @@ void quantize_mmq_q8_1_cuda(
...
+        const float * x, const int32_t * ids, void * vy, const ggml_type type_src0,
+        const int64_t ne00, const int64_t s01, const int64_t s02, const int64_t s03,
+        const int64_t ne0, const int64_t ne1, const int64_t ne2, const int64_t ne3, cudaStream_t stream) {
+    GGML_ASSERT(type_src0 == GGML_TYPE_MXFP4 || type_src0 == GGML_TYPE_NVFP4);
+    GGML_ASSERT(ne0 > 0);
+
+    if (type_src0 == GGML_TYPE_NVFP4) {
+        GGML_ASSERT(ne00 % QK_NVFP4 == 0);
+        constexpr int nvfp4_block_size = 128;
+        const int64_t block_num_y = (ne0 + QK_NVFP4_SUB * nvfp4_block_size - 1) / (QK_NVFP4_SUB * nvfp4_block_size);
```
ORippler, line 425: We really should add a ceildiv function to the cuda backend 😄 This is beyond the scope of this PR though

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -70,6 +70,102 @@ __device__ __forceinline__ uint8_t compute_e8m0_scale(float amax) {
...
+#if defined(BLACKWELL_MMA_AVAILABLE)
+
+    const int64_t i0_base = ((int64_t) blockDim.x * blockIdx.y + threadIdx.x) * QK_NVFP4_SUB;
+    if (i0_base >= ne0) {
+        return;
+    }
+
+    const int64_t i1 = blockIdx.x;
+    const int64_t i2 = blockIdx.z % ne2;
+    const int64_t i3 = blockIdx.z / ne2;
```
ORippler, line 87: 64-bit int operations are costly - do we need expect ne* to require 64-bit precision?

But we can follow-up on this in a separate PR

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -70,6 +70,102 @@ __device__ __forceinline__ uint8_t compute_e8m0_scale(float amax) {
...
+#pragma unroll
+    for (int k = 0; k < QK_NVFP4_SUB; k++) {
+        const int64_t i00 = i0_base + k;
+        if (i00 < ne00) {
+            const float v = x[base_idx + i00];
+            vals_raw[k] = v;
+            amax_raw = fmaxf(amax_raw, fabsf(v));
+        } else {
+            vals_raw[k] = 0.0f;
+        }
```
ORippler, line 113: Shouldn't `i00 < ne00` always evaluate to true as we removed the support for partially-filled blocks? If yes, we can simplify again

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -70,6 +70,102 @@ __device__ __forceinline__ uint8_t compute_e8m0_scale(float amax) {
...
+    }
+
+    static constexpr int test_offsets[5] = { 0, -1, 1, -2, 2};
+    const int first_fp8_code = (int) ggml_cuda_fp32_to_ue4m3(amax_raw * (1.0f / 6.0f));
```
ORippler:
```suggestion
    const int first_fp8_code = (int) ggml_cuda_fp32_to_ue4m3(amax_raw / 6.0f);
```

ggml/src/ggml-cuda/quantize.cu
```diff
@@ -70,6 +70,102 @@ __device__ __forceinline__ uint8_t compute_e8m0_scale(float amax) {
...
+
+    float best_err = FLT_MAX;
+    uint8_t fp8_code = 0;
+    float subblock_scale = 0.0f;
+
+#pragma unroll // Check +/- 2 to find best code to reduce NVFP4 activation loss. Negligible overhead on Blackwell.
+    for (int i = 0; i < 5; i++) {
```
ORippler, line 124:
```suggestion
    for (int i = -2; i < 3; i++) {
```

What's the need for indirecting i to pre-computed test_offsets array?

ggml/src/ggml-cuda/quantize.cu
ORippler: > The real value is that it improves activation loss vs Q8. It reduces ppl on Qwen3.5/4B from 11.80 to 11.65 [Q8 is 11.40], reduces max kld from 12.0 to 11.38 (5.2%) , mean ln 0.075354 from 0.087845 (14% ), mean kld 0.092041 from 0.098035 (6.5% ), RMS Δp 7.925 from 8.136, and every parameter is improved, so this will help any model's quality in some way and reduces outliers. I can take it out if you want, or put it to onto its own PR or just let it go.

I feel we should reevaluate/compare this heuristic against correctly handling nvfp4 as a derived tensor (i.e. all incoming activations have to be divided by the per-tensor scale before entering this function). An alternative to the heuristic here (and something that could be done in the absence of per-tensor nvfp4 scaling) would be to derive the per-activation max at run-time, and scale it by this value.

ORippler (commented):
Some more clean-up comments from my side before we are good to go in my eyes.

Some broader comments (which I perceive to be non-blocking to the PR):
1. I see the general risk of improper handling for "derived-tensor types" (NVFP4/FP8 that bring a per-tensor F32 scale with them) in ggml, if we defer correct application of per-tensor scaling to the GGML callee that constructs the incoming `ggml_cgraph`. But I will raise a separate discussion for this.
2. I would be interested in run-time & quality implications of deriving a per-activation-tensor amax as a proxy in the case of missing input-activations, especially compared to the current heuristic of wiggling fp8 candidate in `[-2, 2]` (note that we are effectively changing exp&mantissa values by adding those to the UE4M3 uint8_t)

am17an:
I ran aime-25 with 32k context and it scored 83.2% with pass@1. The reported scores is 93.2 with 131k context and avg@64 (as given [here](https://research.nvidia.com/labs/nemotron/files/Nemotron-Cascade-2.pdf)). So I think this is good to go

ggml/src/ggml-cuda/quantize.cu
michaelw9999: @ORippler @am17an No intention there, that's just what I came up with first. 
I spent more time messing around with it.
Here is what is even better: changing the tiebreaker.  Instead of `cur_err < best_err` , I tried `cur_err <= best_err`.  Even bigger gain:
Ppl went `11.627140` from `11.658689` (recall baseline `11.805223`).  That <=/< change did not reduce max_kld but `RMS Δp` went to `7.770` from `7.925` while tiny increase on some others.   I will test with more models too.
Also, I did finish all of the proper activation input scale plumbing work, trying a few different spots when and where to apply it and if it made any difference. No real change to the result for what exact spot it's applied. It's a bigger change (I'll share the link to a new branch after cleaning up debugprint work), and it only improved Qwen3.5 4B to `11.632480` (combined with < and +/- 2 candidate search), and barely moved the needle on the bigger model. I was hoping for more improvement, so possibly I didn't do it right, or just expecting more than what it really provides. I will also test out using existing Q8's design and compare . I do not think the input scale work belongs into this PR, though.
Ppl details:
<details>

```
====== Perplexity statistics ====== <--- cur_err <= best_err
Mean PPL(Q)                   :  11.627140 ±   0.370449
Mean PPL(base)                :  10.812440 ±   0.339678
Cor(ln(PPL(Q)), ln(PPL(base))):  98.56%
Mean ln(PPL(Q)/PPL(base))     :   0.072645 ±   0.005391
Mean PPL(Q)/PPL(base)         :   1.075348 ±   0.005797
Mean PPL(Q)-PPL(base)         :   0.814700 ±   0.067644

====== KL divergence statistics ======
Mean    KLD:   0.092071 ±   0.002979
Maximum KLD:  11.383351
99.9%   KLD:   2.895970
99.0%   KLD:   0.632531
95.0%   KLD:   0.258098
90.0%   KLD:   0.181648
Median  KLD:   0.052091
10.0%   KLD:   0.003112
 5.0%   KLD:   0.000974
 1.0%   KLD:   0.000151
 0.1%   KLD:   0.000033
Minimum KLD:   0.000006

====== Token probability statistics ======
Mean    Δp: -1.022 ± 0.085 %
Maximum Δp: 98.703%
99.9%   Δp: 41.835%
99.0%   Δp: 17.709%
95.0%   Δp:  8.621%
90.0%   Δp:  4.668%
75.0%   Δp:  0.654%
Median  Δp: -0.060%
25.0%   Δp: -2.256%
10.0%   Δp: -7.643%
 5.0%   Δp: -12.459%
 1.0%   Δp: -27.172%
 0.1%   Δp: -59.744%
Minimum Δp: -89.937%
RMS Δp    :  7.770 ± 0.230 %
Same top p: 86.569 ± 0.378 %

====== Perplexity statistics ====== cur_err < best_err
Mean PPL(Q)                   :  11.658689 ±   0.371901
Mean PPL(base)                :  10.812440 ±   0.339678
Cor(ln(PPL(Q)), ln(PPL(base))):  98.55%
Mean ln(PPL(Q)/PPL(base))     :   0.075354 ±   0.005421
Mean PPL(Q)/PPL(base)         :   1.078266 ±   0.005845
Mean PPL(Q)-PPL(base)         :   0.846249 ±   0.068650

====== KL divergence statistics ======
Mean    KLD:   0.092041 ±   0.002901
Maximum KLD:  11.383351
99.9%   KLD:   2.599494
99.0%   KLD:   0.662513
95.0%   KLD:   0.263261
90.0%   KLD:   0.179095
Median  KLD:   0.051889
10.0%   KLD:   0.003330
 5.0%   KLD:   0.000976
 1.0%   KLD:   0.000171
 0.1%   KLD:   0.000029
Minimum KLD:  -0.000018

====== Token probability statistics ======
Mean    Δp: -1.105 ± 0.087 %
Maximum Δp: 98.703%
99.9%   Δp: 40.174%
99.0%   Δp: 17.484%
95.0%   Δp:  8.232%
90.0%   Δp:  4.565%
75.0%   Δp:  0.676%
Median  Δp: -0.045%
25.0%   Δp: -2.246%
10.0%   Δp: -7.893%
 5.0%   Δp: -12.515%
 1.0%   Δp: -28.459%
 0.1%   Δp: -66.779%
Minimum Δp: -89.119%
RMS Δp    :  7.925 ± 0.241 %
Same top p: 86.667 ± 0.376 %

Baseline (candidate search=1)
====== Perplexity statistics ======
Mean PPL(Q)                   :  11.805223 ±   0.375738
Mean PPL(base)                :  10.812440 ±   0.339678
Cor(ln(PPL(Q)), ln(PPL(base))):  98.38%
Mean ln(PPL(Q)/PPL(base))     :   0.087845 ±   0.005702
Mean PPL(Q)/PPL(base)         :   1.091819 ±   0.006226
Mean PPL(Q)-PPL(base)         :   0.992783 ±   0.073679

====== KL divergence statistics ======
Mean    KLD:   0.098035 ±   0.002742
Maximum KLD:  12.003211
99.9%   KLD:   2.892974
99.0%   KLD:   0.735795
95.0%   KLD:   0.291616
90.0%   KLD:   0.197111
Median  KLD:   0.056680
10.0%   KLD:   0.003445
 5.0%   KLD:   0.001175
 1.0%   KLD:   0.000169
 0.1%   KLD:   0.000027
Minimum KLD:   0.000002

====== Token probability statistics ======
Mean    Δp: -1.424 ± 0.089 %
Maximum Δp: 98.463%
99.9%   Δp: 41.957%
99.0%   Δp: 17.838%
95.0%   Δp:  8.071%
90.0%   Δp:  4.209%
75.0%   Δp:  0.557%
Median  Δp: -0.087%
25.0%   Δp: -2.572%
10.0%   Δp: -8.650%
 5.0%   Δp: -13.841%
 1.0%   Δp: -30.382%
 0.1%   Δp: -57.089%
Minimum Δp: -93.169%
RMS Δp    :  8.136 ± 0.223 %
Same top p: 86.152 ± 0.382 %

```

</details>

michaelw9999:
> I ran aime-25 with 32k context and it scored 83.2% with pass@1. The reported scores is 93.2 with 131k context and avg@64 (as given [here](https://research.nvidia.com/labs/nemotron/files/Nemotron-Cascade-2.pdf)). So I think this is good to go

Awesome!  Which quant did you use or convert from?

ggml/src/ggml-cuda/quantize.cu
am17an: At this point you're probably overfitting on the dataset. Don't change anything regarding this in this PR for now

am17an:
I used https://huggingface.co/chankhavu/Nemotron-Cascade-2-30B-A3B-NVFP4 and simply ran the `convert_hf_to_gguf.py` script

pwilkin:
@am17an `convert_hf_to_gguf.py` doesn't support repacking NVFP4->NVFP4, or does it?

michaelw9999:
> @am17an `convert_hf_to_gguf.py` doesn't support repacking NVFP4->NVFP4, or does it?

It does, there is currently no other way; llama-quantize doesn't currently convert NVFP4.

CISC:
> @am17an `convert_hf_to_gguf.py` doesn't support repacking NVFP4->NVFP4, or does it?

How else would one make GGUFs like this? :)
https://huggingface.co/CISCai/gemma-4-31B-it-NVFP4-turbo-GGUF

michaelw9999:
> > @am17an `convert_hf_to_gguf.py` doesn't support repacking NVFP4->NVFP4, or does it?
> 
> How else would one make GGUFs like this? :) https://huggingface.co/CISCai/gemma-4-31B-it-NVFP4-turbo-GGUF

@CISC  that's awesome! I will go try it out out, I haven't had time to even do gemma4 yet.. and now Qwen3.6... ahh!  Did you do anything special or just enable it in quantizer the standard way? I've been working on a heavily modified llama-quantizer for NVFP4 for a long time, you can check [https://huggingface.co/michaelw9999/Nemotron-Cascade-2-30B-A3B-NVFP4-GGUF](https://huggingface.co/michaelw9999/Nemotron-Cascade-2-30B-A3B-NVFP4-GGUF) , I was having trouble getting the testsuites working right the last time I tried it AIME25 (100% of answers timing out, it just kept thinking ) but this has ppl `9.810759` vs `9.999596` and maxkld `1.617739` vs `3.821523`,  and moved some BF16 to Q8 reducing size.
Question now is .. play with Qwen3.6-35B-A3B or post up the much faster and now unified FP4 MX/NV repack that goes direct to regs :))

CISC:
> > > @am17an `convert_hf_to_gguf.py` doesn't support repacking NVFP4->NVFP4, or does it?
> > 
> > 
> > How else would one make GGUFs like this? :) https://huggingface.co/CISCai/gemma-4-31B-it-NVFP4-turbo-GGUF
> 
> @CISC that's awesome! I will go try it out out, I haven't had time to even do gemma4 yet.. and now Qwen3.6... ahh! Did you do anything special or just enable it in quantizer the standard way?

Nope, it's just a straight-up conversion of LilaRest's quant.

am17an:
Can I get a second approval? @ggml-org/ggml-cuda

stevelikesrhino:
Tested with a custom quant converted from nvidia’s gemma4 31B nvfp4 quant, I’ve never heard my 5090 screeching like that. Functionality wise didn’t have any problem. Windows 11 cuda 13.1

pwilkin:
> > @am17an `convert_hf_to_gguf.py` doesn't support repacking NVFP4->NVFP4, or does it?
> 
> How else would one make GGUFs like this? :) https://huggingface.co/CISCai/gemma-4-31B-it-NVFP4-turbo-GGUF

I don't know, BLACK MAGIC?

It says it doesn't and I believe it:

```
(venv) ilintar@LinuksowaJaskinia:/devel/projects$ convert_hf_to_gguf.py --outtype nvfp4 --remote chankhavu/Nemotron-Cascade-2-30B-A3B-NVFP4
usage: convert_hf_to_gguf.py [-h] [--vocab-only] [--outfile OUTFILE] [--outtype {f32,f16,bf16,q8_0,tq1_0,tq2_0,auto}] [--bigendian]
                             [--use-temp-file] [--no-lazy] [--model-name MODEL_NAME] [--verbose]
                             [--split-max-tensors SPLIT_MAX_TENSORS] [--split-max-size SPLIT_MAX_SIZE] [--dry-run]
                             [--no-tensor-first-split] [--metadata METADATA] [--print-supported-models] [--remote] [--mmproj]
                             [--mistral-format] [--disable-mistral-community-chat-template] [--sentence-transformers-dense-modules]
                             [--fuse-gate-up-exps]
                             [model]
convert_hf_to_gguf.py: error: argument --outtype: invalid choice: 'nvfp4' (choose from f32, f16, bf16, q8_0, tq1_0, tq2_0, auto)
```

You guys must be using some hacks.

pwilkin:
(and by this I mean there should be some sort of fallback that basically says "yes, picking nvfp4 here because it's a repack")

am17an:
Well it's not an valid `outtype` I think, it just passes on the nvfp4 tensors as nvfp4, same as mxfp4. Do you see mxfp4 there?

michaelw9999:
We should fix that.  The way to do is not set the type to nvfp4.  Just use
no args

CISC:
> We should fix that. The way to do is not set the type to nvfp4. Just use no args

There's nothing to fix really, `nvfp4` is not a valid `outtype` as we can't requantize atm.

pwilkin:
> Well it's not an valid `outtype` I think, it just passes on the nvfp4 tensors as nvfp4, same as mxfp4. Do you see mxfp4 there?

No, that's obviously also handled by black magic.

pwilkin:
> > We should fix that. The way to do is not set the type to nvfp4. Just use no args
> 
> There's nothing to fix really, `nvfp4` is not a valid `outtype` as we can't requantize atm.

I know there's nothing to fix technically, but the error message is misleading. It should say something like "if you want to repack an existing NVFP4-quantized model, run without `--outtype`" IMO. But that's just a minor UX point.

CISC:
> > > We should fix that. The way to do is not set the type to nvfp4. Just use no args
> > 
> > 
> > There's nothing to fix really, `nvfp4` is not a valid `outtype` as we can't requantize atm.
> 
> I know there's nothing to fix technically, but the error message is misleading. It should say something like "if you want to repack an existing NVFP4-quantized model, run without `--outtype`" IMO. But that's just a minor UX point.

The error is output by `ArgumentParser`.

ORippler:
I opened a discussion about general issues/limitations of current NVFP4 support here https://github.com/ggml-org/llama.cpp/discussions/22042, and would advocate for withholding this PR until we have had some discussion over there

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -46,9 +46,12 @@ struct block_q8_1_mmq {
     int8_t qs[4*QK8_1]; // 128 values quantized to 8 bit each
 };
 
+// this struct is used for fp4 data types (currently only used in Blackwell)
```
JohannesGaessler:
```suggestion
// this struct is used for fp4 data types (currently only used for Blackwell)
```

I'm ESL but this seems more correct to me.

ggml/src/ggml-cuda/mmq.cuh
```diff
@@ -232,15 +233,18 @@ static_assert(MMQ_MMA_TILE_X_K_NVFP4 % 8 == 4, "Wrong padding.");
...
         case GGML_TYPE_Q5_0:    return MMQ_MMA_TILE_X_K_Q8_0;
         case GGML_TYPE_Q5_1:    return MMQ_MMA_TILE_X_K_Q8_1;
         case GGML_TYPE_Q8_0:    return MMQ_MMA_TILE_X_K_Q8_0;
         // tile sizes are the same for Q8_1 and FP4 for blackwell
         case GGML_TYPE_MXFP4:   return MMQ_MMA_TILE_X_K_Q8_1;
+#if defined(BLACKWELL_MMA_AVAILABLE)
+        case GGML_TYPE_NVFP4:   return MMQ_MMA_TILE_X_K_FP4;
+#else
         case GGML_TYPE_NVFP4:   return MMQ_MMA_TILE_X_K_NVFP4;
+#endif
```
JohannesGaessler:
```suggestion
#endif // defined(BLACKWELL_MMA_AVAILABLE)
```

ggml/src/ggml-cuda/quantize.cu
JohannesGaessler: I would expect this kernel to be mostly I/O bound anyways, to the extent that compute makes a difference I would first try replacing the division and modulo with `fast_div_modulo` since that should just be free.

ggml/src/ggml-cuda/quantize.cu
JohannesGaessler: The quantized activations are padded with zeros in order to avoid having to do an out-of-bounds check for the weights. I'm not sure what you mean by support for partially filled blocks but this should still be needed unless you ensure that `src[0]->ne[0]` is exactly divided by the MMQ iteration size in k direction.

JohannesGaessler (approved):

ggml/src/ggml-cuda/mmq.cuh
michaelw9999: @JohannesGaessler I have this line ass muscle memory typing that out so many times and was hoping you would not have to do any of those again, but still missed this one :(

krampenschiesser:
> > @am17an `convert_hf_to_gguf.py` doesn't support repacking NVFP4->NVFP4, or does it?
> 
> It does, there is currently no other way; llama-quantize doesn't currently convert NVFP4.

It seems it can?
https://huggingface.co/krampenschiesser/Qwen3.5-35B-A3B-NVFP4.gguf/blob/main/README.md
I quantized the experts to nvfp4 and it looks different (kld wise) to mxfp4 or q4 quants. so it seems to be working.
I tried your branch and can see a significant improvement in pp512.
Thank you for the work here, excited to see this!

master (a678916): 
| qwen35moe 35B.A3B Q8_0         |  19.36 GiB |    34.66 B | CUDA       |  99 |           pp512 |      2161.50 ± 10.35 |
| qwen35moe 35B.A3B Q8_0         |  19.36 GiB |    34.66 B | CUDA       |  99 |           tg128 |         94.80 ± 0.01 |

your branch:
| qwen35moe 35B.A3B Q8_0         |  19.36 GiB |    34.66 B | CUDA       |  99 |           pp512 |       2900.56 ± 8.25 |
| qwen35moe 35B.A3B Q8_0         |  19.36 GiB |    34.66 B | CUDA       |  99 |           tg128 |         94.56 ± 0.04 |

michaelw9999:
This PR closed in error after a local rebase/restore did not go correctly, so it lost all commits and history, and cannot be reopened.  The identical commits and last state were recovered into [PR 22196](https://github.com/ggml-org/llama.cpp/pull/22196 )  .  Sorry for duplication and trouble.

ORippler:
> The quantized activations are padded with zeros in order to avoid having to do an out-of-bounds check for the weights. I'm not sure what you mean by support for partially filled blocks but this should still be needed unless you ensure that src[0]->ne[0] is exactly divided by the MMQ iteration size in k direction.

Oh yeah my bad, I did not think about having to make sure we produce valid K tiles. I was merely thinking about `src[0] -> ne[0] % 16 = 0` being ensured on ggml level (16 being the microblock size of NVFP4), and was referring to this by partially filled blocks.

PR #21897, opened 2026-04-14
ggml-cuda: enable concurrent streams for linear attention

am17an:
## Overview

Enable concurrent streams for linear models (e.g. Qwen3.5). Linear attention layers also have parallelizable kernels, this PR enables them to run on different streams as done for traditional attention layers in #16991. Picture below from `nsys profile`

sync2

This PR also fixes a long standing bug in the stream concurrency code which did not allow it to be enabled by default and also not scale beyond batch_size = 1. The bug was that graph allocator would re-use the `node->src` tensors assuming sequential execution. #16991 had a intricate interleaving pattern to "fool" the allocator into extending lifetimes of the tensors but it did not do so for the `src` tensors, which somehow worked for bs=1 but is not guaranteed to. 

This PR simply introduces a flag `GGML_TENSOR_FLAG_NO_ALLOC_FREE` to prevent the allocator from re-using the memory, and removes the complex reshuffling of nodes. Since we reserve space for worst case graph, this does not increase the size of the compute buffer.

## Additional information

on a 5090 with `GGML_CUDA_GRAPH_OPT=1`

| Model                    |   Microbatch size | Test   |   t/s e21cdc11a |   t/s linear-attn-conc |   Speedup |                                                                        
|:-------------------------|------------------:|:-------|----------------:|-----------------------:|----------:|                                                                        
| gemma4 ?B Q4_K_M         |                 1 | pp512  |          213.63 |                 229.26 |      1.07 |                                                                        
| gemma4 ?B Q4_K_M         |                 2 | pp512  |          364.59 |                 391.76 |      1.07 |                                                                        
| gemma4 ?B Q4_K_M         |                 4 | pp512  |          571.11 |                 603.30 |      1.06 |                                                                        
| gemma4 ?B Q4_K_M         |                 8 | pp512  |          868.01 |                 904.15 |      1.04 |                                                                        
| qwen35 27B Q4_K_M        |                 1 | pp512  |           69.45 |                  72.89 |      1.05 |                                                                        
| qwen35 27B Q4_K_M        |                 2 | pp512  |          129.81 |                 137.60 |      1.06 |                                                                        
| qwen35 27B Q4_K_M        |                 4 | pp512  |          210.49 |                 222.65 |      1.06 |                                                                        
| qwen35 27B Q4_K_M        |                 8 | pp512  |          280.24 |                 296.04 |      1.06 |                                                                        
| qwen35moe 35B.A3B Q4_K_S |                 1 | pp512  |          241.87 |                 231.01 |      0.96 |                                                                        
| qwen35moe 35B.A3B Q4_K_S |                 2 | pp512  |          332.47 |                 355.24 |      1.07 |                                                                        
| qwen35moe 35B.A3B Q4_K_S |                 4 | pp512  |          532.37 |                 561.42 |      1.05 |                                                                        
| qwen35moe 35B.A3B Q4_K_S |                 8 | pp512  |          860.46 |                 899.61 |      1.05 | 

Note: the slowdown on qwen3.5 on bs=1 seems to be some quirk of llama-bench, it is the first model and the first run. Running a TG benchmark shows a speed-up there as well.

# Requirements

- AI usage disclosure: Yes, debugging and testing

### DIFF
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -635,12 +635,12 @@ extern "C" {
 
     // this tensor...
     enum ggml_tensor_flag {
-        GGML_TENSOR_FLAG_INPUT   =  1, // ...is an input for the GGML compute graph
-        GGML_TENSOR_FLAG_OUTPUT  =  2, // ...is an output for the GGML compute graph
-        GGML_TENSOR_FLAG_PARAM   =  4, // ...contains trainable parameters
-        GGML_TENSOR_FLAG_LOSS    =  8, // ...defines loss for numerical optimization (multiple loss tensors add up)
+        GGML_TENSOR_FLAG_INPUT         =  1, // ...is an input for the GGML compute graph
+        GGML_TENSOR_FLAG_OUTPUT        =  2, // ...is an output for the GGML compute graph
+        GGML_TENSOR_FLAG_PARAM         =  4, // ...contains trainable parameters
+        GGML_TENSOR_FLAG_LOSS          =  8, // ...defines loss for numerical optimization (multiple loss tensors add up)
         GGML_TENSOR_FLAG_COMPUTE       = 16, // ...must be computed
-        GGML_TENSOR_FLAG_NO_ALLOC_FREE = 32, // ...buffer should not be freed/reused by the allocator
+        GGML_TENSOR_FLAG_NO_ALLOC_FREE = 32, // ...buffer should not be freed/reused by the allocator (note: causes graph realloc)
     };
 
     enum ggml_tri_type {
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -639,7 +639,8 @@ extern "C" {
         GGML_TENSOR_FLAG_OUTPUT  =  2, // ...is an output for the GGML compute graph
         GGML_TENSOR_FLAG_PARAM   =  4, // ...contains trainable parameters
         GGML_TENSOR_FLAG_LOSS    =  8, // ...defines loss for numerical optimization (multiple loss tensors add up)
-        GGML_TENSOR_FLAG_COMPUTE = 16, // ...must be computed
+        GGML_TENSOR_FLAG_COMPUTE       = 16, // ...must be computed
+        GGML_TENSOR_FLAG_NO_ALLOC_FREE = 32, // ...buffer should not be freed/reused by the allocator
     };
 
     enum ggml_tri_type {
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -652,7 +652,8 @@ extern "C" {
         GGML_TENSOR_FLAG_OUTPUT  =  2, // ...is an output for the GGML compute graph
         GGML_TENSOR_FLAG_PARAM   =  4, // ...contains trainable parameters
         GGML_TENSOR_FLAG_LOSS    =  8, // ...defines loss for numerical optimization (multiple loss tensors add up)
-        GGML_TENSOR_FLAG_COMPUTE = 16, // ...must be computed
+        GGML_TENSOR_FLAG_COMPUTE       = 16, // ...must be computed
+        GGML_TENSOR_FLAG_NO_ALLOC_FREE = 32, // ...buffer should not be freed/reused by the allocator
     };
 
     enum ggml_tri_type {
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -648,12 +648,12 @@ extern "C" {
 
     // this tensor...
     enum ggml_tensor_flag {
-        GGML_TENSOR_FLAG_INPUT   =  1, // ...is an input for the GGML compute graph
-        GGML_TENSOR_FLAG_OUTPUT  =  2, // ...is an output for the GGML compute graph
-        GGML_TENSOR_FLAG_PARAM   =  4, // ...contains trainable parameters
-        GGML_TENSOR_FLAG_LOSS    =  8, // ...defines loss for numerical optimization (multiple loss tensors add up)
+        GGML_TENSOR_FLAG_INPUT         =  1, // ...is an input for the GGML compute graph
+        GGML_TENSOR_FLAG_OUTPUT        =  2, // ...is an output for the GGML compute graph
+        GGML_TENSOR_FLAG_PARAM         =  4, // ...contains trainable parameters
+        GGML_TENSOR_FLAG_LOSS          =  8, // ...defines loss for numerical optimization (multiple loss tensors add up)
         GGML_TENSOR_FLAG_COMPUTE       = 16, // ...must be computed
-        GGML_TENSOR_FLAG_NO_ALLOC_FREE = 32, // ...buffer should not be freed/reused by the allocator
+        GGML_TENSOR_FLAG_NO_ALLOC_FREE = 32, // ...buffer should not be freed/reused by the allocator (note: causes graph realloc)
     };
 
     enum ggml_tri_type {
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -648,12 +648,11 @@ extern "C" {
 
     // this tensor...
     enum ggml_tensor_flag {
-        GGML_TENSOR_FLAG_INPUT         =  1, // ...is an input for the GGML compute graph
-        GGML_TENSOR_FLAG_OUTPUT        =  2, // ...is an output for the GGML compute graph
-        GGML_TENSOR_FLAG_PARAM         =  4, // ...contains trainable parameters
-        GGML_TENSOR_FLAG_LOSS          =  8, // ...defines loss for numerical optimization (multiple loss tensors add up)
-        GGML_TENSOR_FLAG_COMPUTE       = 16, // ...must be computed
-        GGML_TENSOR_FLAG_NO_ALLOC_FREE = 32, // ...buffer should not be freed/reused by the allocator (note: causes graph realloc)
+        GGML_TENSOR_FLAG_INPUT   =  1, // ...is an input for the GGML compute graph
+        GGML_TENSOR_FLAG_OUTPUT  =  2, // ...is an output for the GGML compute graph
+        GGML_TENSOR_FLAG_PARAM   =  4, // ...contains trainable parameters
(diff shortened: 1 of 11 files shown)


ggml/src/ggml-backend.cpp
```diff
@@ -1498,8 +1498,14 @@ static bool ggml_backend_sched_alloc_splits(ggml_backend_sched_t sched) {
...
 
+    // unconditionally recreate the flag if any node has NO_ALLOC_FREE set
+    bool has_no_alloc_free = false;
+    for (int i = 0; i < sched->graph.n_nodes && !has_no_alloc_free; i++) {
+        has_no_alloc_free = (sched->graph.nodes[i]->flags & GGML_TENSOR_FLAG_NO_ALLOC_FREE) != 0;
+    }
+
     // allocate graph
-    if (backend_ids_changed || !ggml_gallocr_alloc_graph(sched->galloc, &sched->graph)) {
+    if (backend_ids_changed || has_no_alloc_free || !ggml_gallocr_alloc_graph(sched->galloc, &sched->graph)) {
```
JohannesGaessler, line 1508: Why is this done?

am17an: Currently after graph_optimize, if the size of the graph hasn't changed and other things like `backend_ids_changed` is not true, the allocator will skip reallocation. Basically it is not respecting that the node->flag can change.

JohannesGaessler: I'm not sure I follow. Without this patch `ggml_gallocr_alloc_graph` would be called. If that function fails to allocate the graph, shouldn't it be returning false?

am17an: `ggml_gallocr_needs_realloc` returns false because the current code assumes the graph doesn't need a realloc. This allocates the graph on previous buffer assignments, however in graph_optimize we explicitly request the allocator to not re-use buffers for these nodes

JohannesGaessler: Ah okay, I understand what you mean now. The logic is also going to be incorrect for `GGML_FLAG_OUTPUT` then because the meaning of that flag is effectively the same: the tensor data is not supposed to be overwritten (so that it can be retrieved after the graph evaluation). I was wondering why you were adding a new flag rather than to use that one. But since it seems to me like there would in practice be no difference I would suggest to use `GGML_FLAG_OUTPUT` and to fix for that flag.

am17an: There are other uses for `OUTPUT`, for example it blocks kernel fusion in `ggml_can_fuse`, so I think a new flag is necessary. 

https://github.com/ggml-org/llama.cpp/blob/006809fdba84260478fc351d11f7f293d89bd246/ggml/src/ggml-impl.h#L649-L652

JohannesGaessler: Okay, that makes sense.

ggml/include/ggml.h
```diff
@@ -639,7 +639,8 @@ extern "C" {
         GGML_TENSOR_FLAG_OUTPUT  =  2, // ...is an output for the GGML compute graph
         GGML_TENSOR_FLAG_PARAM   =  4, // ...contains trainable parameters
         GGML_TENSOR_FLAG_LOSS    =  8, // ...defines loss for numerical optimization (multiple loss tensors add up)
-        GGML_TENSOR_FLAG_COMPUTE = 16, // ...must be computed
+        GGML_TENSOR_FLAG_COMPUTE       = 16, // ...must be computed
+        GGML_TENSOR_FLAG_NO_ALLOC_FREE = 32, // ...buffer should not be freed/reused by the allocator
```
JohannesGaessler: Please keep the indentation consistent.

ggml/src/ggml-backend.cpp
```diff
@@ -1498,8 +1498,14 @@ static bool ggml_backend_sched_alloc_splits(ggml_backend_sched_t sched) {
         }
     }
 
+    // unconditionally recreate the flag if any node has NO_ALLOC_FREE set
+    bool has_no_alloc_free = false;
+    for (int i = 0; i < sched->graph.n_nodes && !has_no_alloc_free; i++) {
+        has_no_alloc_free |= (sched->graph.nodes[i]->flags & GGML_TENSOR_FLAG_NO_ALLOC_FREE) != 0;
```
JohannesGaessler, line 1504: To my understanding, the reason this check is needed here but not for the `OUTPUT` flag is because the graph optimization step can set the `NO_ALLOC_FREE` flag. I'm wondering whether there is any legitimate use case where a user would want to set the `NO_ALLOC_FREE` flag themself (right now I can't think of any). Because then this check would cause an unconditional re-allocation every time. But then we should clearly mention that the `NO_ALLOC_FREE` flag is for internal bookkeeping only in the corresponding comment.

am17an: I think `graph_optimize` could have an extension to the API allows the users to realloc based on some parameter (apart from the other cases). That is what this flag is doing essentially

JohannesGaessler (approved):

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -4337,90 +4285,71 @@ static void ggml_backend_cuda_graph_optimize(ggml_backend_t backend, ggml_cgraph
...
+                        const_cast<ggml_tensor *>(n)->flags |= GGML_TENSOR_FLAG_NO_ALLOC_FREE;
+                        for (int si = 0; si < GGML_MAX_SRC; ++si) {
+                            const ggml_tensor * s = n->src[si];
+                            if (!s) continue;
+                            const_cast<ggml_tensor *>(s)->flags |= GGML_TENSOR_FLAG_NO_ALLOC_FREE;
+                            if (s->view_src) {
+                                const_cast<ggml_tensor *>(s->view_src)->flags |= GGML_TENSOR_FLAG_NO_ALLOC_FREE;
+                            }
+                        }
+                    }
```
ggerganov, line 4343: I don't think adding the `GGML_TENSOR_FLAG_NO_ALLOC_FREE` is a good idea. The allocator memory recycling should be solved in a different way.

JohannesGaessler: Right now we are determining memory re-use under the assumption that nodes in a graph are executed in the exact order as their indices. In principle we could generalize this to also consider possible out-of-order execution.

am17an: https://github.com/ggml-org/llama.cpp/pull/16991#issuecomment-3586600984 was the discussion earlier around making allocator stream-aware. Say we take a node and add a non-zero stream-id to it and then the allocator doesn't recycle nodes which have non-zero stream-ids, then it would be the same as this PR. It would still need to be realloc'ed after graph optimize though. The advantage of doing this would be that we can recycle nodes after each join rather than keeping the memory pinned as we're doing now.

All of this requires changes to the backend scheduler though, if those are not wanted we have to think of something else and likely would need to retire this code path altogether as it's broken even in current master

am17an: I tried doing this the metal way of re-ordering nodes but it's even slower than master.

ORippler: > Right now we are determining memory re-use under the assumption that nodes in a graph are executed in the exact order as their indices. In principle we could generalize this to also consider possible out-of-order execution.

This is definitely useful for running smaller models on bigger NVGPUs where we want to parallelize vertically. Resolving this would also allow for the graph optimization to leave the experimental state for the cuda backend.

> I don't think adding the GGML_TENSOR_FLAG_NO_ALLOC_FREE is a good idea. The allocator memory recycling should be solved in a different way.

An alternative idea could be for `graph_optimize` to be allowed to also return a list of graphs, which the backend sched has to dispatch in succession. As such, we could split out the parallellizeable segments into individual graphs, which should be issueable independently by the backend scheduler (and we could synchronize as needed by events). Though this may involve changes in the cuda backend to expose multi-stream parallel execution properly (did not spend too much thought on this yet). Also, if I track https://github.com/ggml-org/llama.cpp/pull/20793 correctly we are still missing an officially-agreed on formalization of how a backend should behave in the case where multiple calls to async_compute are issued against it (is it allowed to parallelize them? or do we require it to process them sequentially)

am17an:
I have a separate proposal which is more inline with what @JohannesGaessler initially mentioned. 

We can extend the `graph_optimize` function to provide allocator dependencies, something like this

```
  typedef void (*ggml_backend_add_alloc_dep_fn)(
      void * user_data,
      ggml_tensor * tensor, // keep this live
      ggml_tensor * until   // until this node
  );

  struct ggml_backend_graph_optimize_params {
      ggml_backend_add_alloc_dep_fn add_alloc_dep;
      void * user_data;
  };
  ```

  Then change the backend interface from:

 ``` void (*graph_optimize)(ggml_backend_t backend, ggml_cgraph * graph);```

  to:
```
  void (*graph_optimize)(
      ggml_backend_t backend,
      ggml_cgraph * graph,
      const ggml_backend_graph_optimize_params * params);
```

This way each the graph allocator knows not to re-use the memory till the join nodes. I believe this is the most architecturally sound way, as it can recycle memory every layer instead of keeping it throughout the graph and also doesn't have an implementation detail like `NO_ALLOC` leak into the scheduler. What do you guys think?

JohannesGaessler:
I think it would be better to extend the graph allocator with explicit support for divergent, concurrent paths. The logic for this should be ggml backend independent so my opinion is that the graph allocator should have an integer like `max_concurrent_streams` as input and an integer per node like `stream_id` as output. The stream IDs can then be passed to ggml backends which have the responsibility to synchronize when a node is using sources from another stream ID.

am17an:
I feel the stream concurrency (or same GPU parallelization) is an implementation detail of the CUDA backend, and because it already uses CUDA events to sync it is even more incompatible with the ggml backend because it uses CUDA events itself in multi-GPU cases, or possibly to implement an async copy engine like in #21067. A separate stream using the same device can also be created using another instantiation of `ggml_backend`, which also complicates this somewhat.

My suggestion is breaking the assumption that nodes are executed in sequential order through some "dependency injection", which already is a concept in scheduler to extend lifetimes of tensors

am17an:
This now has enabled linear attn, but I don't see e2e speed-ups, but speed-ups only in TG. This probably means that the `graph_optimize` call is taking a lot of time. @ORippler since you expressed interest in contributing here, any help would be appreciated

PR #21910, opened 2026-04-14
CUDA: require explicit opt-in for P2P access

JohannesGaessler:
In https://github.com/ggml-org/llama.cpp/pull/19378 I had naively enabled CUDA peer-to-peer access guarded only by `cudaDeviceCanAccessPeer`. However, for some motherboards and BIOS settings this seems to cause crashes or corrupted outputs. I don't think we can feasibly check for this so our only option is to make peer access an explicit opt-in.

# Requirements

- AI usage disclosure: Yes, I asked Gemini about it and it gave me links to relevant threads in the CUDA developer forum.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -281,6 +281,12 @@ Use `GGML_CUDA_FORCE_CUBLAS_COMPUTE_16F` environment variable to force use FP16 
 
 The environment variable `GGML_CUDA_ENABLE_UNIFIED_MEMORY=1` can be used to enable unified memory in Linux. This allows swapping to system RAM instead of crashing when the GPU VRAM is exhausted. In Windows this setting is available in the NVIDIA control panel as `System Memory Fallback`.
 
+### Peer Access
+
+The environment variable `GGML_CUDA_P2P` can be set to enable peer-to-peer access between multiple GPUs, allowing them to transfer data directly rather than to go through system memory.
+Requires driver support (usually restricted to workstation/datacenter GPUs).
+May cause crashes or corrupted outputs for some motherboards and BIOS settings (e.g. IOMMU).
+
 ### Performance Tuning
 
 The following compilation options are also available to tweak performance:
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -324,16 +324,18 @@ static ggml_cuda_device_info ggml_cuda_init() {
     // configure logging to stdout
     // CUBLAS_CHECK(cublasLoggerConfigure(1, 1, 0, nullptr));
 
-    for (int id = 0; id < info.device_count; ++id) {
-        ggml_cuda_set_device(id);
-        for (int id_other = 0; id_other < info.device_count; ++id_other) {
-            if (id == id_other) {
-                continue;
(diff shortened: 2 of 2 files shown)


am17an (approved):
should we adding a warning when using multi-GPU to use `GGML_CUDA_P2P` if possible?

JohannesGaessler:
Previously when I had multiple P40s in one machine I got a few % better performance using P2P for `--split-mode row`. For my currently connected A16 I am seeing no difference. My opinion is that we should only be printing warnings when an optional feature is clearly beneficial like NCCL because we want to avoid drowning out critical information in a wall of text. For things like this I think RTFM is more appropriate.

IMbackK:
There are scenarios where this is going to have a quite a bit more pronounced effect, like amd gpus linked by IF where the non-peer access will go over pcie instead of IF

JohannesGaessler:
@ggml-org/maintainers can I get a second review, please?

CISC (approved):

gaugarg-nv:
@JohannesGaessler does it have any impact on perf for TP, especially for the Non-NCCL path?

JohannesGaessler:
When compiling without NCCL I see a difference from peer access on 4x NVIDIA A16:

| model         | sm     | test  | t/s default   | t/s peer access |
| ------------- | -----: | ----: | ------------: | --------------: |
| llama 8B Q4_0 | layer  | pp512 | 823.00 ± 1.40 | 822.73 ± 1.78   |
| llama 8B Q4_0 | layer  | tg128 | 37.35 ± 0.00  | 37.48 ± 0.00    |
| llama 8B Q4_0 | tensor | pp512 | 799.38 ± 2.91 | 977.09 ± 3.87   |
| llama 8B Q4_0 | tensor | tg128 | 59.82 ± 0.20  | 75.09 ± 0.30    |

In that case it would probably make sense to print a warning.

gaugarg-nv:
IMO, many of these issues with P2P come due to system-specific issues like not disabling iommu as mentioned in CUDA documentation: https://docs.nvidia.com/cuda/cuda-programming-guide/03-advanced/multi-gpu-systems.html#host-iommu-hardware-pci-access-control-services-and-vms

So, it might be better to keep P2P enabled by default, but provide an env variable to disable P2P if needed.

JohannesGaessler:
No, I am very much of the opinion that llama.cpp should at least work out-of-the-box and only warn the user if they might be able to get better performance via some option. In particular, the fact that the outputs can be corrupted is a big problem. I would only willing to enable P2P by default if CUDA provides me with a reliable way to detect those cases where it can potentially cause problems.

gaugarg-nv:
Ok, that is fair. Unfortunately, there doesn't seem to be a way to detect such issues at the CUDA level.
Let me speak internally to the CUDA teams and see if this can be improved.

IMbackK:
Pretty mutch per definition a bug in cuda -  no? if the gpus cant dma eatch other because of the iommu cuda needs to fall back to host copies. This exact thing was a problem in rocm with pcie bridges that dont support p2p (rocm uses p2p much more widely since it will do it for all parts while, if i understand correctly, cuda will only do it for pro parts) until amd fixed this issue.

PR #21918, opened 2026-04-14
ggml: improve SPIR-V headers detection with __has_include while preserving original _WIN32 logic

EmilAskerov:
## Overview

Fix SPIR-V header include detection in `ggml-vulkan.cpp` to support a wider range of real-world build environments.

The previous platform-specific check:
`_WIN32 && !__MINGW32__`
fails in several valid configurations and can lead to missing header errors.

This PR introduces a `__has_include`-based detection mechanism that probes available include paths at compile time and selects the correct one automatically.

Since `llama.cpp` requires C++17, this implementation relies entirely on standard `__has_include` and removes outdated platform-specific fallback logic, resulting in cleaner code.

This primarily improves downstream integration of llama.cpp as a library.

---

### Problem

The existing logic breaks in multiple real-world scenarios:

- custom Qt / CMake projects  
- MinGW / llvm-mingw with official LunarG Vulkan SDK  
- different Linux package layouts  

Example build error:

```
fatal error: 'spirv/unified1/spirv.hpp' file not found
```

### Solution

Use `__has_include` to detect SPIR-V headers in a portable way:

- `<spirv/unified1/spirv.hpp>`  
- `<spirv-headers/spirv.hpp>`  
- `<spirv.hpp>`  

If none are found, it defaults to the standard Khronos layout to let the compiler throw a proper "file not found" error.
This approach is:

- toolchain-independent  
- robust across SDK layouts  
- compatible with custom build systems  
- strictly C++17 compliant

### Testing

Tested on:

- Windows 11 (64-bit)  
  - Qt Creator 19.0.0  
  - Qt 6.11.0 (llvm-mingw 64-bit)  
  - Vulkan SDK 1.4.341.1 (official LunarG)  

### Impact

- Fixes build failures outside official CMake builds  
- No functional changes for standard builds  
- Fully backward compatible  

---

# Requirements

I have read and agree with the contributing guidelines  

AI usage disclosure:

The code changes, testing, PR description  were performed by the author.

### DIFF
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -20,12 +20,22 @@ DispatchLoaderDynamic & ggml_vk_default_dispatcher();
 #define VULKAN_HPP_DEFAULT_DISPATCHER ggml_vk_default_dispatcher()
 
 #include <vulkan/vulkan.hpp>
-// SPIRV-Headers: LunarG Windows SDK uses Include/spirv-headers/spirv.hpp (not spirv/unified1/). MinGW/MSYS2 and
-// Linux packages use Khronos layout spirv/unified1/spirv.hpp. See docs/build.md#vulkan.
-#if defined(_WIN32) && !defined(__MINGW32__)
-#include <spirv-headers/spirv.hpp>
+
+// SPIR-V Headers: different SDK installations expose different include paths.
+// LunarG Vulkan SDK on Windows typically provides <spirv-headers/spirv.hpp>,
+// while Linux packages, MSYS2 and MinGW often use the Khronos layout
+// <spirv/unified1/spirv.hpp>. We prefer __has_include for better portability,
+// falling back to the original platform logic for maximum compatibility.
+#if __has_include(<spirv/unified1/spirv.hpp>)
+#    include <spirv/unified1/spirv.hpp>
+#elif __has_include(<spirv-headers/spirv.hpp>)
+#    include <spirv-headers/spirv.hpp>
+#elif __has_include(<spirv.hpp>)
+#    include <spirv.hpp>
+#elif defined(_WIN32) && !defined(__MINGW32__)
+#    include <spirv-headers/spirv.hpp>
 #else
-#include <spirv/unified1/spirv.hpp>
+#    include <spirv/unified1/spirv.hpp>
 #endif
 
 #include <algorithm>
--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -22,20 +22,29 @@ DispatchLoaderDynamic & ggml_vk_default_dispatcher();
 #include <vulkan/vulkan.hpp>
 
 // SPIR-V Headers: different SDK installations expose different include paths.
-// LunarG Vulkan SDK on Windows typically provides <spirv-headers/spirv.hpp>,
-// while Linux packages, MSYS2 and MinGW often use the Khronos layout
-// <spirv/unified1/spirv.hpp>. We prefer __has_include for better portability,
-// falling back to the original platform logic for maximum compatibility.
-#if __has_include(<spirv/unified1/spirv.hpp>)
-#    include <spirv/unified1/spirv.hpp>
-#elif __has_include(<spirv-headers/spirv.hpp>)
-#    include <spirv-headers/spirv.hpp>
-#elif __has_include(<spirv.hpp>)
-#    include <spirv.hpp>
-#elif defined(_WIN32) && !defined(__MINGW32__)
-#    include <spirv-headers/spirv.hpp>
+// LunarG Vulkan SDK on Windows typically provides <spirv-headers/spirv.hpp>.
+// Linux packages, MSYS2 and MinGW often use the Khronos layout <spirv/unified1/spirv.hpp>.
+// FreeBSD typically uses <spirv_cross/spirv.hpp>.
+#if defined(__has_include)
+#    if __has_include(<spirv/unified1/spirv.hpp>)
+#        include <spirv/unified1/spirv.hpp>
+#    elif __has_include(<spirv-headers/spirv.hpp>)
+#        include <spirv-headers/spirv.hpp>
+#    elif __has_include(<spirv_cross/spirv.hpp>)
+#        include <spirv_cross/spirv.hpp>
+#    elif __has_include(<spirv.hpp>)
+#        include <spirv.hpp>
+#    else
+// Fallback to let the compiler throw a standard "file not found" error
+#        include <spirv/unified1/spirv.hpp>
+#    endif
 #else
-#    include <spirv/unified1/spirv.hpp>
+// Fallback logic for older compilers that do not support __has_include
+#    if defined(_WIN32) && !defined(__MINGW32__)
+#        include <spirv-headers/spirv.hpp>
+#    else
+#        include <spirv/unified1/spirv.hpp>
+#    endif
 #endif
 
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


ilevad:
Supporting this — '__has_include' approach is more robust

0cc4m:
Please restore the PR description template and fill it out.

EmilAskerov:
> Please restore the PR description template and fill it out.

Done, restored the template and filled it out. Please let me know if anything else should be adjusted.

gboddin:
For who it may concern, you might want to throw a :

```
#elif __has_include(<spirv_cross/spirv.hpp>)
#    include <spirv_cross/spirv.hpp>
```

for freebsd users!

https://www.freshports.org/graphics/spirv-cross/

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -20,12 +20,22 @@ DispatchLoaderDynamic & ggml_vk_default_dispatcher();
...
+// <spirv/unified1/spirv.hpp>. We prefer __has_include for better portability,
+// falling back to the original platform logic for maximum compatibility.
+#if __has_include(<spirv/unified1/spirv.hpp>)
+#    include <spirv/unified1/spirv.hpp>
+#elif __has_include(<spirv-headers/spirv.hpp>)
+#    include <spirv-headers/spirv.hpp>
+#elif __has_include(<spirv.hpp>)
+#    include <spirv.hpp>
+#elif defined(_WIN32) && !defined(__MINGW32__)
+#    include <spirv-headers/spirv.hpp>
```
0cc4m: How would this fallback or the one below ever be reached, assuming `__has_include` works as intended?

EmilAskerov: Good catch, thank you! You are completely right. If __has_include fails to find anything, my previous fallback would just cause a missing header error anyway. I have wrapped the new logic in #if defined(__has_include) and strictly separated the original logic into the #else block to serve as a true fallback for older compilers.

EmilAskerov:
> For who it may concern, you might want to throw a :
> 
> ```
> #elif __has_include(<spirv_cross/spirv.hpp>)
> #    include <spirv_cross/spirv.hpp>
> ```
> 
> for freebsd users!
> 
> https://www.freshports.org/graphics/spirv-cross/

Thanks for the tip! I've added <spirv_cross/spirv.hpp> for FreeBSD users in the latest commit.

EmilAskerov:
> Supporting this — '__has_include' approach is more robust

Thanks for the support!

0cc4m:
> For who it may concern, you might want to throw a :
> 
> ```
> #elif __has_include(<spirv_cross/spirv.hpp>)
> #    include <spirv_cross/spirv.hpp>
> ```
> 
> for freebsd users!
> 
> https://www.freshports.org/graphics/spirv-cross/

Why spirv-cross and not a general spirv-header-like package? spirv-cross is for compiling spirv to other languages. It can contain the header, but it should also be available elsewhere. I can't check on that webpage cause it just tells me the server is busy.

EmilAskerov:
> > For who it may concern, you might want to throw a :
> > ```
> > #elif __has_include(<spirv_cross/spirv.hpp>)
> > #    include <spirv_cross/spirv.hpp>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for freebsd users!
> > https://www.freshports.org/graphics/spirv-cross/
> 
> Why spirv-cross and not a general spirv-header-like package? spirv-cross is for compiling spirv to other languages. It can contain the header, but it should also be available elsewhere. I can't check on that webpage cause it just tells me the server is busy.

@0cc4m You are completely right. I looked into it, and spirv-cross is indeed a translation library that just happens to vendor spirv.hpp under its own namespace.

The actual correct FreeBSD package is graphics/spirv-headers, which installs the headers into the standard Khronos layout: <spirv/unified1/spirv.hpp>. Since <spirv/unified1/spirv.hpp> is already our first __has_include check, standard FreeBSD setups are already fully covered without needing this workaround.

@gboddin As noted above, the correct way on FreeBSD is to install graphics/spirv-headers rather than relying on spirv-cross.

I have just pushed a commit to remove the spirv_cross fallback to keep the include paths standard and clean. Thanks for pointing this out!

jeffbolznv:
There's a bug open to make the windows SDK use the spirv/unified1/spirv.hpp location, so hopefully we can use that unconditionally in the not too distant future. If we need this more complex workaround until then, I guess that's fine.

Peter0x44:
w64devkit's `llama.mak` is also broken without this change, because I use mingw + windows vulkan SDK.
https://github.com/skeeto/w64devkit/blob/master/contrib/llama.mak

You can test it by placing llama.mak in the root of the repo, and running:
`$ make -f llama.mak -j32 VULKAN_SDK="C:/VulkanSDK/1.4.328.1"`

So it gets a +1 from me.

ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -20,12 +20,28 @@ DispatchLoaderDynamic & ggml_vk_default_dispatcher();
...
+#        include <spirv-headers/spirv.hpp>
+#    elif __has_include(<spirv.hpp>)
+#        include <spirv.hpp>
+#    else
+         // Fallback to let the compiler throw a standard "file not found" error
+#        include <spirv/unified1/spirv.hpp>
+#    endif
 #else
-#include <spirv/unified1/spirv.hpp>
+// Fallback logic for older compilers that do not support __has_include
```
Peter0x44: FYI __has_include is part of C++17, and I think llama.cpp already requires it. So I don't think a real compiler will ever hit this.

Just a minor nit...

0cc4m: Yes, this is true.

EmilAskerov: @Peter0x44 @0cc4m Removed the redundant C++17 fallback block. Thanks for pointing it out!

EmilAskerov:
> There's a bug open to make the windows SDK use the spirv/unified1/spirv.hpp location, so hopefully we can use that unconditionally in the not too distant future. If we need this more complex workaround until then, I guess that's fine.

@jeffbolznv Thanks for the context! Yeah, once the Windows SDK adopts the standard Khronos layout and it becomes widely used, we can definitely simplify this and just use <spirv/unified1/spirv.hpp> unconditionally. Until then, hopefully, this keeps everything building smoothly.

EmilAskerov:
@0cc4m @Peter0x44 It looks like two CI jobs failed due to unrelated infrastructure/flaky issues:

    ggml-ci-win-intel-vulkan was cancelled after ~3 hours (timeout/runner issue).

    ggml-ci-x64-cpu-low-perf is a CPU test and doesn't even compile ggml-vulkan.cpp.

All primary Vulkan builds passed successfully. Could you please re-trigger the failed jobs when you have a moment?

0cc4m (approved):

Peter0x44 (approved):
I have zero power to commit anything but LGTM.

0cc4m:

CISC (approved):

0cc4m:
Thank you for the contribution.

jeffbolznv:
> There's a bug open to make the windows SDK use the spirv/unified1/spirv.hpp location, so hopefully we can use that unconditionally in the not too distant future. If we need this more complex workaround until then, I guess that's fine.

This is fixed in the 1.4.350 SDK that was released today. There's not a great way in cmake to require a specific version of the SDK (you can potentially look at Vulkan_VERSION, but that could be misleading on other platforms). I don't think we're in any rush to remove the __has_include, but at some point in the future we probably should.

EmilAskerov:
> > There's a bug open to make the windows SDK use the spirv/unified1/spirv.hpp location, so hopefully we can use that unconditionally in the not too distant future. If we need this more complex workaround until then, I guess that's fine.
> 
> This is fixed in the 1.4.350 SDK that was released today. There's not a great way in cmake to require a specific version of the SDK (you can potentially look at Vulkan_VERSION, but that could be misleading on other platforms). I don't think we're in any rush to remove the __has_include, but at some point in the future we probably should.

Thanks for the update and the context. Glad the change was useful in the meantime.

PR #21925, opened 2026-04-14
fix: correct CPU counting on multi-NUMA and SMT systems

### LINKED ISSUE #17611: Eval bug: wrong default threads in llama-bench
### Name and Version

latest commit: https://github.com/ggml-org/llama.cpp/commit/fa0465954faef9d7170b967ad89f8bc5303a32f3

`llama-bench` using by default wrong threads.

Linux server with NUMA nodes

Image

### Operating systems

Linux

### GGML backends

CPU

### Hardware

Supermicro X10dri Dual CPU

mraleko:
## Summary

The `cpu_count_math_cpus()` function in `common/common.cpp` naively skipped every other logical CPU to account for hyperthreading using `++cpu`. This fails on multi-NUMA systems where the thread sibling topology is not contiguous across NUMA boundaries.

On a dual-Xeon Supermicro X10dri (2x18 cores, 72 logical CPUs), `sysconf(_SC_NPROCESSORS_ONLN)` returns 72. The old loop counted only 36 cores but the topology-based counting shows 36 physical cores (2 NUMA nodes x 18). The discrepancy suggests the loop was counting correctly by accident on this specific machine, but the `++cpu` assumption breaks in more complex topologies.

**Root cause**: The assumption that HT siblings are always at `cpu+1` is invalid on multi-socket/NUMA systems where CPU enumeration order may not match thread sibling topology.

**Fix**: Read `/sys/devices/system/cpu/cpu{N}/topology/thread_siblings` to find the actual HT sibling group and skip the matching CPU directly, rather than assuming consecutive CPU IDs are siblings.

Closes #17611

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -165,6 +165,7 @@ static bool is_running_on_efficiency_core(void) {
 
 static int cpu_count_math_cpus(int n_cpu) {
     int result = 0;
+    std::vector<int> core_ids;
     for (int cpu = 0; cpu < n_cpu; ++cpu) {
         if (pin_cpu(cpu)) {
             return -1;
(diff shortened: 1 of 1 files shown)


mraleko:
Human-authored fix for issue #17611. AI assistance used for research and exploration only; the solution design and implementation are my own work.

am17an:
This is not a portable solution

PR #21934, opened 2026-04-15
cmake: remove CMP0194 policy to restore MSVC builds

### LINKED ISSUE #22100: Compile bug: Can no longer configure CMake on MSVC with CMP0194 set to NEW
### Git commit

4f02d4733934179386cbc15b3454be26237940bb

### Operating systems

Windows

### GGML backends

CUDA

Setting the CMake policy CMP0194 to NEW requires that an ASM compiler be present, and cmake configure now fails, since MSVC is no longer considered an ASM compiler by newer CMake verisons.
Removing the policy line makes it configure again, albeit with the warning the offending commit intended to remove.

I think the correct fix is to only enable ASM with [enable_language](https://cmake.org/cmake/help/latest/command/enable_language.html) when one of the backends that use it (per the commit message in the breaking change) are enabled

### First Bad Commit

acc37a42eaba02e57065fef3b415d9f1e64018fa

### Compile command

```shell
cmake -B build -DGGML_CUDA=ON
```

### Relevant log output

```shell
-- The ASM compiler identification is unknown
-- Didn't find assembler
CMake Error at ggml/CMakeLists.txt:9 (project):
  No CMAKE_ASM_COMPILER could be found.
```

texasich:
Thanks to @oobabooga for catching this: https://github.com/ggml-org/llama.cpp/pull/21630#issuecomment-4248308373

## Overview

PR #21630 added CMP0194 NEW to silence a warning, but it broke Windows MSVC+Ninja.

## Additional information
the first attempt at scoping ASM to kleidiai hit an unrelated CMake scoping issue on the ARM+KleidiAI self-hosted runner, so I pivoted to a minimal revert. This removes only the 6-line CMP0194 policy block from ggml/CMakeLists.txt. project("ggml" C CXX ASM) is left untouched, which is exactly the pre-#21630 state that was working on all platforms. The CMake 4.1+ warning returns but no platform breaks.

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/ggml/CMakeLists.txt
+++ b/ggml/CMakeLists.txt
@@ -1,11 +1,5 @@
 cmake_minimum_required(VERSION 3.14...3.28) # for add_link_options and implicit target directories.
 
-# ref: https://cmake.org/cmake/help/latest/policy/CMP0194.html
-# MSVC is not a valid assembler for the ASM language.
-# Set to NEW to avoid a warning on CMake 4.1+ with MSVC.
-if (POLICY CMP0194)
-    cmake_policy(SET CMP0194 NEW)
-endif()
 project("ggml" C CXX ASM)
 
 ### GGML Version


CISC (approved):
Thanks for the swift fix.

CISC:
Ah, unfortunately not quite sufficient:
https://github.com/ggml-org/llama.cpp/actions/runs/24436493164/job/71406353173?pr=21934

CISC:
Still fails:
https://github.com/ggml-org/llama.cpp/actions/runs/24455670070/job/71456348667

Aflah012:
This PR want to fix the same problem as [#21969](https://github.com/ggml-org/llama.cpp/pull/21969)

CISC:
@texasich No progress? Should we merge as-is?

CISC (approved):

texasich:
yeah let's merge this as is. the warning is cosmetic and nobody's pipeline should break over a fix for it. i'll take another crack at scoping ASM properly once i have access to an ARM runner to test against. @CISC

PR #21959, opened 2026-04-15
models: add ggml_cont after ggml_transpose before ggml_concat in conv1d input assembly

Achilles1089:
## Problem

Several models with recurrent layers (Mamba, DeltaNet/Qwen3.5, Kimi-Linear, PLaMo2) build conv1d inputs by concatenating a conv state buffer with a transposed tensor:

```cpp
ggml_tensor * conv_x = ggml_concat(ctx0, conv, ggml_transpose(ctx0, x), 0);
```

`ggml_transpose` returns a zero-copy view -- it does not move any data, just reinterprets the strides. For an 8192-wide tensor, `nb[1] = 32,768 bytes`. When `ggml_concat` iterates over this non-contiguous view, every element requires a 32KB-strided memory access, causing severe cache thrashing.

## Fix

Add `ggml_cont()` between the transpose and concat to materialize contiguous memory:

```cpp
// Before (strided access in concat):
conv_x = ggml_concat(ctx0, conv, ggml_transpose(ctx0, x), 0);

// After (contiguous memcpy in concat):
conv_x = ggml_concat(ctx0, conv, ggml_cont(ctx0, ggml_transpose(ctx0, x)), 0);
```

The `dup` kernel handles strided-to-contiguous copy efficiently via SIMD, and concat then operates on contiguous rows with fast `memcpy`.

Note: `delta-net-base.cpp` and `wavtokenizer-dec.cpp` already use this pattern -- this PR brings the remaining models into line.

## Affected models

| File | Model |
|------|-------|
| `qwen35moe.cpp` | Qwen3.5-MoE |
| `qwen35.cpp` | Qwen3.5 |
| `qwen3next.cpp` | Qwen3Next |
| `mamba-base.cpp` | Mamba1 + Mamba2 (2 instances) |
| `kimi-linear.cpp` | Kimi-Linear |
| `plamo2.cpp` | PLaMo2 |

## Benchmarks

Measured on **Qwen3.5-35B-A3B** (Xeon E5-2697 v2, 11 threads, Metal+BLAS backend):

**Q4_K_M (production quant):**
| Test | Before | After | Change |
|------|:------:|:-----:|:------:|
| pp512 | 43.0 | **46.0** | **+7.0%** |
| tg128 | 11.8 | 11.7 | 0% (noise) |

**Q4_0:**
| Test | Before | After | Change |
|------|:------:|:-----:|:------:|
| pp512 | 40.9 | **42.8** | **+4.6%** |
| tg128 | 13.4 | 13.3 | 0% (noise) |

The improvement scales with the number of DeltaNet/Mamba layers since each layer concatenates transposed QKV/state tensors during prefill. Token generation shows no regression because the conv1d input is only 1 row wide.

## How I found this

Profiling pp4096 with macOS `sample` showed `ggml_compute_forward_concat_f32` consuming **27% of total compute time**. After this fix, concat dropped to **0.3%** -- the time moved to the much faster `dup` kernel.

### DIFF
--- a/src/models/kimi-linear.cpp
+++ b/src/models/kimi-linear.cpp
@@ -33,7 +33,7 @@ static ggml_tensor * causal_conv1d(ggml_cgraph * gf, ggml_context * ctx0, ggml_t
     ggml_tensor * x_3d = ggml_reshape_3d(ctx0, x_proj, d_inner, n_seq_tokens, n_seqs);
 
     // Concat Q conv state and current input: {d_conv-1 + n_seq_tokens, d_inner, n_seqs}
-    ggml_tensor * conv_x = ggml_concat(ctx0, conv_state_x, ggml_transpose(ctx0, x_3d), 0);
+    ggml_tensor * conv_x = ggml_concat(ctx0, conv_state_x, ggml_cont(ctx0, ggml_transpose(ctx0, x_3d)), 0);
 
     // Save last (d_conv-1) columns back to Q conv state
     ggml_tensor * last_conv_x = ggml_view_3d(ctx0, conv_x, d_conv - 1, d_inner, n_seqs,
--- a/src/models/mamba-base.cpp
+++ b/src/models/mamba-base.cpp
@@ -52,7 +52,7 @@ ggml_tensor * llm_build_mamba_base::build_mamba_layer(llm_graph_input_rs * inp,
     // conv
     {
         // => {d_conv - 1 + n_seq_tokens, d_inner, n_seqs}
-        ggml_tensor * conv_x = ggml_concat(ctx0, conv, ggml_transpose(ctx0, x), 0);
+        ggml_tensor * conv_x = ggml_concat(ctx0, conv, ggml_cont(ctx0, ggml_transpose(ctx0, x)), 0);
 
         // copy last (d_conv - 1) columns back into the state cache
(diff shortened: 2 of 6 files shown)


CISC:
Sorry, this is an anti-pattern and most likely not efficient on other backends.

Your observation about `delta-net-base` is incorrect, the only `cont`s there are before ops that (currently) require it, and several `wavtokenizer-dec` `cont`s should be removed for ops that no longer require it.

Achilles1089:
Thanks for the review @CISC. You're right -- adding `ggml_cont` at every call site is the wrong layer to fix this.

The bottleneck is in `ggml_compute_forward_concat_f32` (in `ops.cpp`) which does element-by-element copy in a 4-nested loop. When either source is a transposed view (e.g. `nb[0] = 32KB`), this causes severe cache thrashing on CPU.

I attempted to optimize the concat kernel itself with row-level `memcpy`, but the transposed case has non-contiguous elements within each row (`nb10 != sizeof(float)`), so `memcpy` can't be used directly. The `dup` kernel (used by `ggml_cont`) achieves better results because it uses SIMD for strided-to-contiguous copies.

Would the right approach be to optimize the CPU concat kernel to handle non-contiguous sources more efficiently (e.g. using SIMD strided loads), or is there a different pattern you'd recommend for these conv1d input assembly paths?

Closing this PR per your feedback. Happy to rework once the right approach is clear.

CISC:
Optimizing concat would definitely be appreciated, not just for CPU, in the case of transposed tensors it could be done similarly to how the CUDA backend `CPY` op is handled.

PR #21964, opened 2026-04-15
ggml-webgpu(shader): support conv2d kernels. 

Constannnnnt:
## Overview

In this PR, we implemented the conv2d shader kernel to support VL models that require conv2d operations. 

## Additional information

Backend ops tests all passed. I haven't tested this with real models yet. 

# Requirements

- AI usage disclosure: I used ChatGPT to discuss the initial shader idea and further optimization.


reeselevine:
cool, would be really interested to see what it looks like to get vision model support integrated with the WebGPU backend, especially in the browser. Maybe you're already working on this :). 

FYI I'm a little backed up on some existing PRs that will probably cause merge conflicts, so I'll wait to review this until I get the other ones merged and any conflicts are fixed.

Constannnnnt:
@reeselevine sounds good! There's no need to rush on the PR, as I am working on the inference system for vision capabilities to see how it works in the browser and will keep you posted when real models are tested. One thing I noticed is that this PR contained past commits from last PR. I'm not sure why they were here; I will do some research and might try to avoid that next time.

reeselevine:
It's ok (even if not ideal) if this PR contains stale commits, because the whole PR ends up getting squashed into one commit when we merge. Generally it's a good idea to reset your fork to llama.cpp's `master` branch before working on new changes, otherwise your fork will continue to diverge in history because of the squashed commits.

Constannnnnt:
Thanks! Good to know that.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -1115,8 +1138,7 @@ class ggml_webgpu_shader_lib {
                     std::string type_upper = type_str;
                     std::transform(type_upper.begin(), type_upper.end(), type_upper.begin(), ::toupper);
 
-                    switch (key.src_type)
-                    {
+                    switch (key.src_type) {
```
reeselevine, line 1141: formatting changes, are you using `clang-format`? Maybe these were missed before?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -83,7 +84,7 @@ static inline void compute_2d_workgroups(uint32_t total_wg, uint32_t max_per_dim
 #define WEBGPU_NUM_PARAM_SLOT_SAFETY_MARGIN      10u
 #define WEBGPU_RUNTIME_WAIT_TIMEOUT_MS           30000u
 #define WEBGPU_RUNTIME_WAIT_TIMEOUT_NS           (WEBGPU_RUNTIME_WAIT_TIMEOUT_MS * 1e6)
-#define WEBGPU_PARAMS_BUF_SIZE_BYTES             128  // enough for 32 parameters
+#define WEBGPU_PARAMS_BUF_SIZE_BYTES             256  // enough for 64 parameters
```
reeselevine, line 87: I only count 30 parameters in the new shader, is this necessary? Although uniform buffer alignment I think is usually 256 bytes anyways so maybe this doesn't end up adding any extra memory?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -907,6 +908,97 @@ static webgpu_encoded_op ggml_webgpu_solve_tri(webgpu_context & ctx,
...
+        (uint32_t) s1,
+        (uint32_t) p0,
+        (uint32_t) p1,
+        (uint32_t) d0,
+        (uint32_t) d1,
+    };
+
+    std::vector<wgpu::BindGroupEntry> entries = {
+        { .binding = 0,
+         .buffer  = ggml_webgpu_tensor_buf(src0),
```
reeselevine, line 964: let's move away from using the initializer lists to avoid warnings, eventually when I enable the ggml-nvidia-ci it will break too. There should be a new function to create bind group entries already available too.

reeselevine (commented):
Implementation looks good to me! If you can maybe give some performance numbers (compared to some other backend maybe). Not too worried about performance for initial implementations but would be good to get a baseline to see how much room for improvement there is.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
Constannnnnt: I used clangd for formatting, which has a built-in clangFormat? I will change the format for this.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: oh, this was before the optimization (it had 33 parameters before), and it included some unused parameters (like some for inputs) that I removed later. I actually did not notice this extra memory, but it will be good to take a look at it. will report back.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: sure, sounds good.

Constannnnnt:
I see the #21738; I think I will update after this PR is merged. I am also working on im2col shader as well (I have built and tested the vision capabilities in the browser using mmproj; the whole pipeline works, but since there are some ops not being supported, like im2col, the model just hallucinates the results. Isolating the mmproj gguf in the CPU actually leads to good results.

Constannnnnt:
image
image

vibe-d a visualization for the results. Seems to have a lot of performance to fill the gap

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
Constannnnnt: Checking previous commits and other similar code blocks: these have been missed before.

Constannnnnt:
Hi @reeselevine, addressed the comments in this PR.

Some heads-up: I have run the vision model (like lfm-2.5-vl-450m) in the browser; it works perfectly after adding some other kernels and identified some existing bugs in the current kernels. I will create several PRs on new kernels and a new issue for tracking problems in existing kernels when working with CLIP vision encoder.

reeselevine:
Great that vision model support is working, would love to see a demo when it's ready!

A couple things to address, then I think this PR can be merged, and performance can come later:
* Address editorconfig failures: https://github.com/ggml-org/llama.cpp/actions/runs/24699385897/job/72249603220?pr=21964
* Make sure these warnings are gone: https://github.com/ggml-org/llama.cpp/actions/runs/24699385940/job/72249604215?pr=21964#step:9:114

CISC (approved):

reeselevine (approved):

PR #21969, opened 2026-04-15
Update CMakeLists.txt

bjodom:
## Overview

Build failing on windows, root cause is project("ggml" C CXX ASM) forces generic ASM, which MSVC cannot satisfy under CMP0194 NEW.

# Requirements

- AI usage disclosure:


Aflah012:
This PR want to fix the same problem as [#21934](https://github.com/ggml-org/llama.cpp/pull/21934)

Issue #21966, opened 2026-04-15
Refactor: move `load_hparams` and `load_tensors` to per-model definition


ngxson:
### Background Description

Currently, `load_hparams` and `load_tensors` are long switch..case that can be moved to specific model file under `src/models/*.cpp`

The benefit would be to make model definition to be more self-contained, improving the developer experience.

### Possible Refactor Approaches

I'm currently working on this idea on my fork: https://github.com/ngxson/llama.cpp/tree/xsn/model_def_self_contained

The goals are:
1. Making least changes in the code as possible --> prefer moving code around, but prevent directly edit it
2. Possible to auto-migrate via a script, making it easier to resolve git conflicts if any

For the actual implementation, the idea is to expose these functions in `llama_model`:

```cpp
    // model must define these
    virtual void load_hparams(llama_model_loader & ml) = 0;
    virtual void load_tensors(llama_model_loader & ml) = 0;
    virtual std::unique_ptr<llm_graph_context> build_graph_context(const llm_graph_params & params) const = 0;
```

And each model need to define it:

```cpp
struct llama_model_demo : public llama_model {
    llama_model_demo(const struct llama_model_params & params) : llama_model(params) {}
    void load_hparams(llama_model_loader & ml) override {
        hparams.n_layer = 100;
    }
    void load_tensors(llama_model_loader & ml) override {
        output = nullptr;
    }
    struct graph : public llm_graph_context {
        graph(const llama_model & model, const llm_graph_params & params) : llm_graph_context(params) {
            ggml_build_forward_expand(gf, nullptr);
        }
    };
    std::unique_ptr<llm_graph_context> build_graph_context(const llm_graph_params & params) const override {
        return std::make_unique<graph>(*this, params);
    }
};
```

Migration rules:
- Create `llama_model_*` class for each model
- Move `llm_build_ARCH` to `llama_model_ARCH::graph`
- Move code from switch..case to `load_hparams` and `load_tensors`
- Write `llama_model_create()` which is a big switch..case that automatically select the correct `llama_model_ARCH`

ngxson:
CC @ggml-org/maintainers for discussion. This may need some preparation before the actual migration (mostly to make auto-migrate script happy)

pwilkin:
Yes yes yes.

I was planning to do this as a followup to my model split, so good to see you beat me to it :)

JohannesGaessler:
I am definitely in favor of splitting up the model code to a more self-contained format.

[...]

PR #21970, opened 2026-04-15
model: using single llm_build per arch

ngxson:
## Overview

Prepare for https://github.com/ggml-org/llama.cpp/issues/21966

Using one single `llm_build_*` class per arch will make the migration a bit easier.

Example before:

```cpp
        case LLM_ARCH_LLAMA:
            {
                llm = std::make_unique<llm_build_llama<false>>(*this, params);
            } break;
        case LLM_ARCH_LLAMA4:
            {
                if (hparams.swa_type == LLAMA_SWA_TYPE_NONE) {
                    llm = std::make_unique<llm_build_llama<false>>(*this, params);
                } else {
                    llm = std::make_unique<llm_build_llama_iswa>(*this, params);
                }
            } break;
```

Example after:

```cpp
        case LLM_ARCH_LLAMA:
            {
                llm = std::make_unique<llm_build_llama<false>>(*this, params);
            } break;
        case LLM_ARCH_LLAMA4:
            {
                if (hparams.swa_type == LLAMA_SWA_TYPE_NONE) {
                    llm = std::make_unique<llm_build_llama4<false>>(*this, params);
                } else {
                    llm = std::make_unique<llm_build_llama4<true>>(*this, params);
                }
            } break;
```

# Requirements

- AI usage disclosure: no


CISC (approved):
This will conflict with #21971, mind landing that one first?

src/models/models.h
```diff
@@ -701,12 +702,13 @@ struct llm_build_step35_iswa : public llm_graph_context {
...
-    llm_build_t5_dec(const llama_model & model, const llm_graph_params & params);
+template <bool is_enc>
+struct llm_build_t5 : public llm_graph_context {
+    llm_build_t5(const llama_model & model, const llm_graph_params & params);
 };
 
-struct llm_build_t5_enc : public llm_graph_context {
-    llm_build_t5_enc(const llama_model & model, const llm_graph_params & params);
+struct llm_build_t5encoder : public llm_build_t5<true> {
+    llm_build_t5encoder(const llama_model & model, const llm_graph_params & params);
```
ggerganov, line 711: Do we need to keep this, instead of using `llm_build_t5<true>`?

CISC: I think it's there to match the arch name (for the migration script)?

ngxson: @ggerganov the t5encoder has its own tensor loader and hparams loader, so I think it make sense to have a dedicated graph builder. The dedicated class will then expanded into llama_model_t5encoder during the migration (via a script), and its hparams/tensors loader will be moved there

ngxson:
Should be ok now I guess? @CISC @ggerganov

ggerganov (approved):

PR #21245, opened 2026-04-01
model : refactor QKV into common build_qkv and create_tensor_qkv helpers

JoursBleu:
## Overview

Currently llama.cpp supports 112 model files in `src/models/`.

We modified the 85 applicable model files. Our changes abstract the duplicated
Q/K/V tensors' loading and graph-building code into two reusable helpers,
following the `create_tensor_gate_up_exps` pattern (#19139).

• `create_tensor_qkv` (llama-model.cpp): tries fused `wqkv`/`bqkv` first (`TENSOR_NOT_REQUIRED | TENSOR_SKIP_IF_VIRTUAL`), falls back to separate `wq`/`wk`/`wv`. Supports adding biases.

• `build_qkv` (llama-graph.h/cpp): returns `{Qcur, Kcur, Vcur}` as 3D tensors. Fused case: single fused qkv matmul + `ggml_view_3d` split. Separate case: 3 separate matmuls + `ggml_reshape_3d`.

Test: `test-llama-archs` — all OK, 0 FAIL. Zero diff on `llama-arch.cpp`.

The remaining 27 models are not modified for the following reasons:

| Reason | Count | Models |
|--------|-------|--------|
| Non-attention (SSM/linear/RNN) | 10 | mamba, mamba-base, rwkv6, rwkv6-base, rwkv6qwen2, rwkv7, rwkv7-base, arwkv7, delta-net-base, wavtokenizer-dec |
| MLA attention | 4 | deepseek2, minicpm3, minimax-m2, plm |
| Graph directly uses layer.wqkv (non-standard layout) | 3 | cogvlm, openelm, plamo2 |
| Q+gate joint projection | 4 | qwen35, qwen35moe, qwen3next, plamo3 |
| n_embd_head_k != n_embd_head_v | 2 | step35-iswa, mimo2-iswa |
| No fused wqkv_enc | 1 | t5-enc |
| Other special architectures | 3 | olmo2, olmoe, kimi-linear |

## Additional information

Basing on the discussion in #20628 (@am17an, @ngxson). The plan is:

1. This PR: This PR does not modify any logic, it simply extracts the redundant code into
the two functions above, and adds handling for the fused qkv case.
2. Future PR: add `--fuse-qkv` to `convert_hf_to_gguf.py`.

## Requirements

- [x] I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: YES - used as a translation tool for translating the PR description

PR #21971, opened 2026-04-15
model : support NVFP4 tensors for Gemma4

CISC:
## Overview

Add support for NVFP4 Gemma4.

## Additional information

Also adds `wo_s` to `build_attn` to be able to pass it on to `build_lora_mm`.
GGUF: [CISCai/gemma-4-31B-it-NVFP4-turbo-GGUF](https://huggingface.co/CISCai/gemma-4-31B-it-NVFP4-turbo-GGUF)

# Requirements

- AI usage disclosure: nu

### DIFF
--- a/src/llama-graph.cpp
+++ b/src/llama-graph.cpp
@@ -2011,6 +2011,7 @@ ggml_tensor * llm_graph_context::build_attn(
         llm_graph_input_attn_no_cache * inp,
         ggml_tensor * wo,
         ggml_tensor * wo_b,
+        ggml_tensor * wo_s,
         ggml_tensor * q_cur,
         ggml_tensor * k_cur,
         ggml_tensor * v_cur,
@@ -2044,7 +2045,7 @@ ggml_tensor * llm_graph_context::build_attn(
     cb(cur, "kqv_out", il);
 
     if (wo) {
-        cur = build_lora_mm(wo, cur);
+        cur = build_lora_mm(wo, cur, wo_s);
     }
 
     if (wo_b) {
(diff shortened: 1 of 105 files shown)


src/llama-graph.cpp
```diff
@@ -2227,10 +2235,15 @@ ggml_tensor * llm_graph_context::build_attn(
...
-        cur = build_lora_mm(wo, cur);
         if (arch == LLM_ARCH_GLM4 || arch == LLM_ARCH_GLM4_MOE) {
             // GLM4 and GLM4_MOE seem to have numerical issues with half-precision accumulators
+            cur = build_lora_mm(wo, cur);
             ggml_mul_mat_set_prec(cur, GGML_PREC_F32);
+            if (wo_s) {
+                cur = ggml_mul(ctx0, cur, wo_s);
+            }
+        } else {
+            cur = build_lora_mm(wo, cur, wo_s);
```
ggerganov, line 2246: Maybe a follow-up PR to fix the order of the `build_lora_mm` arguments (e.g. `cur, wo, wo_s`) and add an optional precision argument to avoid this branching.

ggerganov (approved):
For consistency, these should be removed and use `wq_b`, `wk_b`, `wv_b` and `wqkv_b` instead:

https://github.com/ggml-org/llama.cpp/blob/b1be68e8cab67f5c2fb7d8e3e90291a8805ece0e/src/llama-model.h#L256-L263

In a follow-up PR

src/llama-graph.cpp
CISC: Yes, will be more manageable after merging #21245

CISC:
@ngxson gentle ping, merge this, then `build_qkv`, then yours?

ngxson (approved):
yes, sounds good to me

PR #21980, opened 2026-04-16
server: tests: fetch random media marker via /apply-template (#21962)

ServeurpersoCom:
## Overview

Fix CI

# Requirements

- AI usage disclosure: YES, python write, install and run the test on disposable pod

### DIFF
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -85,9 +85,7 @@ std::string gen_tool_call_id() {
 }
 
 const char * get_media_marker() {
-    // magic statics: initialized exactly once, thread-safe by C++11
     static const std::string marker = []() {
-        // allow user to pin a reproducible marker via env var (useful for tests and external pipelines)
         const char * env = getenv("LLAMA_MEDIA_MARKER");
         if (env && env[0] != '\0') {
             return std::string(env);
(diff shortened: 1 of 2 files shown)


ServeurpersoCom:
cc @ngxson if you prefer a separate helper let me know

CISC (approved):

ngxson:
Actually I think allowing user to set the marker would be a more flexible approach.

Maybe we can add a logic in `get_media_marker()` to return the env var `LLAMA_MEDIA_MARKER` if it's set?

tools/server/server-common.cpp
```diff
@@ -87,7 +87,13 @@ std::string gen_tool_call_id() {
...
+        // allow user to pin a reproducible marker via env var (useful for tests and external pipelines)
+        const char * env = getenv("LLAMA_MEDIA_MARKER");
+        if (env && env[0] != '\0') {
+            media_marker = env;
+        } else {
+            media_marker = "<__media_" + random_string() + "__>";
+        }
     }
     return media_marker.c_str();
 }
```
ggerganov, line 97: Make `get_media_marker()` thread-safe like this:

```c++
const char * get_media_marker() {
    static const std::string media_marker = []() {
        // allow user to pin a reproducible marker via env var (useful for tests and external pipelines)
        const char * env = getenv("LLAMA_MEDIA_MARKER");
        if (env && env[0] != '\0') {
            return std::string(env);
        } else {
            return std::string("<__media_") + random_string() + "__>";
        }
    }();
    
    return media_marker.c_str();
}
```

ggerganov (approved):

ngxson (approved):

PR #21982, opened 2026-04-16
mmproj: handle missing image_mean/image_std in preprocessor config

### LINKED ISSUE #21775: Misc. bug: convert_hf_to_gguf.py crashes with KeyError: 'image_mean' when preprocessor_config.json is incomplete
### Name and Version
version: 8757 (a29e4c0b7)

### Operating systems
Windows, Linux, Mac

Python/Bash scripts

### Command line
python convert_hf_to_gguf.py path/to/model --outfile mmproj.gguf --mmproj --verbose

The `MmprojModel` class in `convert_hf_to_gguf.py` crashes with a `KeyError: 'image_mean'` when exporting a multimodal projector if the Hugging Face `preprocessor_config.json` is missing the `image_mean` or `image_std` keys. This occurs with several newer Gemma 4 and custom vision checkpoints that don't include these standard preprocessor keys.

The current implementation assumes these keys always exist:
```python
image_mean = _MISTRAL_COMMON_DATASET_MEAN if self.is_mistral_format else self.preprocessor_config["image_mean"]
```

### Relevant log output
<details>
<summary>Logs</summary>

```console
  File "convert_hf_to_gguf.py", line 2233, in set_gguf_parameters
    image_mean = _MISTRAL_COMMON_DATASET_MEAN if self.is_mistral_format else self.preprocessor_config["image_mean"]
KeyError: 'image_mean'
```
</details>

mraleko:
## Overview

Fix mmproj conversion when `image_mean` or `image_std` is missing from the preprocessor config.

Use `preprocessor_config.get(...)` instead of indexing directly, and fall back to the existing common default image stats when either value is absent. Also emit a warning so the fallback is visible during conversion.

Fixes #21775.

## Additional information

This avoids failing on models that omit these fields from the preprocessor config while preserving the existing behavior for configs that define them.

# Requirements

- [x] AI usage disclosure: YES

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -2225,8 +2225,15 @@ class MmprojModel(ModelBase):
             self.gguf_writer.add_vision_head_count(self.find_vparam(["num_attention_heads", "num_heads", "heads", "vt_num_attention_heads"]))
 
             # preprocessor config
-            image_mean = _MISTRAL_COMMON_DATASET_MEAN if self.is_mistral_format else self.preprocessor_config["image_mean"]
-            image_std = _MISTRAL_COMMON_DATASET_STD if self.is_mistral_format else self.preprocessor_config["image_std"]
+            image_mean = _MISTRAL_COMMON_DATASET_MEAN if self.is_mistral_format else self.preprocessor_config.get("image_mean")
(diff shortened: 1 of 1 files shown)


ngxson (changes requested):
this fix is invalid; if a model is NOT mistral model, taking default values from mistral will definitely be incorrect

PR #14800, opened 2025-07-21
CUDA: add fused rms norm

am17an:
Similar to the Vulkan PR (#14366), perhaps `ggml_vk_can_fuse` and `ggml_cuda_can_fuse` can live inside ggml instead of their respective backends since they don't have backend specific code

Decent speedup in PP on my RTX 3090

| Model           | Test   |   t/s c82d48ec |   t/s cuda_fused_rms_norm |   Speedup |
|:----------------|:-------|---------------:|--------------------------:|----------:|
| llama 7B Q5_K_M | pp1    |         124.25 |                    123.34 |      0.99 |
| llama 7B Q5_K_M | pp512  |        4114.28 |                   4610.30 |      1.12 |
| llama 7B Q5_K_M | tg1    |         113.82 |                    120.39 |      1.06 |
| llama 7B Q5_K_M | tg128  |         115.84 |                    118.67 |      1.02 |

PR #14596, opened 2025-07-09
metal : fuse add, mul

ggerganov:
target #14629 

Fuse `GGML_OP_ADD` and `GGML_OP_MUL`

```bash
LLAMA_SET_ROWS=1 ./scripts/compare-commits.sh master gg/metal-fuse-add -m ./models/qwen3-30b-a3b/ggml-model-q8_0.gguf -m models/gemma-3-4b/ggml-model-q8_0.gguf -fa 1 -t 1
```

| Model                 | Test   |   t/s master |   t/s gg/metal-fuse-add |   Speedup |
|:----------------------|:-------|-------------:|------------------------:|----------:|
| gemma3 4B Q8_0        | pp512  |      2444.84 |                 2494.63 |      1.02 |
| gemma3 4B Q8_0        | tg128  |        90.39 |                   96.76 |      1.07 |
| qwen3moe 30B.A3B Q8_0 | pp512  |      1362.92 |                 1420.74 |      1.04 |
| qwen3moe 30B.A3B Q8_0 | tg128  |        70.12 |                   76.68 |      1.09 |

#### Testing

```bash
make -j && GGML_METAL_FUSION_DEBUG=2 ./bin/test-backend-ops -o RMS_NORM_MUL_ADD -b Metal
```

```
Backend 1/3: Metal
  Device description: Apple M4 Max
  Device memory: 28753 MB (28747 MB free)

ggml_metal_encode_node: fuse: RMS_NORM + MUL + ADD
  RMS_NORM_MUL_ADD(type=f32,ne=[64,5,4,3],eps=0.000000): OK
ggml_metal_encode_node: fuse: RMS_NORM + MUL + ADD
  RMS_NORM_MUL_ADD(type=f32,ne=[64,5,4,3],eps=0.000001): OK
ggml_metal_encode_node: fuse: RMS_NORM + MUL + ADD
  RMS_NORM_MUL_ADD(type=f32,ne=[64,5,4,3],eps=0.000100): OK
ggml_metal_encode_node: fuse: RMS_NORM + MUL + ADD
  RMS_NORM_MUL_ADD(type=f32,ne=[64,5,4,3],eps=0.100000): OK
ggml_metal_encode_node: fuse: RMS_NORM + MUL + ADD
  RMS_NORM_MUL_ADD(type=f32,ne=[64,5,4,3],eps=1.000000): OK
  6543/6543 tests passed
  Backend Metal: OK
ggml_backend_metal_device_rel: fused ADD: 5
[...]

PR #14545, opened 2025-07-05
vulkan: fix rms_norm+mul fusion

jeffbolznv:
The fused operation was grabbing the epsilon value from the wrong place.

Add an env var to disable fusion.

Add some missing checks for supported shapes/types.

Handle fused rms_norm+mul in check_results.

Fixes #14540.

PR #21983, opened 2026-04-16
ggml-webgpu: Add fused RMS_NORM + MUL

yomaytk:
## Overview

This PR adds the initial kernel fusion to WebGPU backend with RMS_NORM + MUL (it is similar to https://github.com/ggml-org/llama.cpp/pull/14800).
The performance on the major models on my device (M2, Metal 4) is as follows, but unfortunately, the performance is almost the same on this implementation.

The command is like this:
`llama-bench -m Llama-3.2-3B-Instruct-Q4_K_M.gguf -fa 1 -p 512 -n 0`

| Model | Test | t/s master([`a620695`](https://github.com/ggml-org/llama.cpp/commit/a6206958d28a064564ef132091b9c617ae005f49)) | t/s yomaytk/fused-rms-norm-mul | Speedup |
|---|---|---|---|---|
| gemma4 E4B Q4_K_M | pp1 | 32.09 | 31.94 | 0.995 |
| gemma4 E4B Q4_K_M | pp512 | 464.23 | 467.15 | 1.006 |
| gemma4 E4B Q4_K_M | tg1 | 31.99 | 32.46 | 1.015 |
| gemma4 E4B Q4_K_M | tg128 | 32.50 | 32.85 | 1.011 |
| qwen35 4B Q4_K_M | pp1 | 30.88 | 30.84 | 0.999 |
| qwen35 4B Q4_K_M | pp512 | 479.07 | 480.75 | 1.004 |
| qwen35 4B Q4_K_M | tg1 | 30.99 | 30.90 | 0.997 |
| qwen35 4B Q4_K_M | tg128 | 31.12 | 31.35 | 1.007 |
| llama3.2 3B Q4_K_M | pp1 | 52.51 | 53.57 | 1.020 |
| llama3.2 3B Q4_K_M | pp512 | 670.86 | 676.70 | 1.009 |
| llama3.2 3B Q4_K_M | tg1 | 53.40 | 53.83 | 1.008 |
| llama3.2 3B Q4_K_M | tg128 | 54.03 | 54.71 | 1.013 |

# Requirements

- AI usage disclosure: YES, I used AI to investigate the kernel fusion of Vulkan and CUDA backend, and help analyze the profiling data.

### DIFF
--- a/ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
+++ b/ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
@@ -194,6 +194,26 @@ struct ggml_webgpu_row_norm_pipeline_key_hash {
     }
 };
 
+/** RMS_NORM + MUL **/
+
+struct ggml_webgpu_rms_norm_mul_pipeline_key {
+    bool inplace;
+    bool src_overlap;
+
+    bool operator==(const ggml_webgpu_rms_norm_mul_pipeline_key & other) const {
+        return inplace == other.inplace && src_overlap == other.src_overlap;
+    }
+};
+
+struct ggml_webgpu_rms_norm_mul_pipeline_key_hash {
+    size_t operator()(const ggml_webgpu_rms_norm_mul_pipeline_key & key) const {
+        size_t seed = 0;
+        ggml_webgpu_hash_combine(seed, key.inplace);
+        ggml_webgpu_hash_combine(seed, key.src_overlap);
+        return seed;
+    }
+};
+
 /** Pad **/
 struct ggml_webgpu_pad_pipeline_key {
     bool circular;
@@ -734,16 +754,17 @@ class ggml_webgpu_shader_lib {
     std::unordered_map<int, webgpu_pipeline> cumsum_pipelines;         // key is fixed, no variants yet
     std::unordered_map<ggml_webgpu_row_norm_pipeline_key, webgpu_pipeline, ggml_webgpu_row_norm_pipeline_key_hash>
         row_norm_pipelines;                                            // op/inplace
+
     std::unordered_map<ggml_webgpu_get_rows_pipeline_key, webgpu_pipeline, ggml_webgpu_get_rows_pipeline_key_hash>
-        get_rows_pipelines;                                            // src_type, vectorized
+        get_rows_pipelines;   // src_type, vectorized
     std::unordered_map<ggml_webgpu_unary_pipeline_key, webgpu_pipeline, ggml_webgpu_unary_pipeline_key_hash>
-        unary_pipelines;                                               // type/op/inplace
+        unary_pipelines;      // type/op/inplace
     std::unordered_map<ggml_webgpu_scale_pipeline_key, webgpu_pipeline, ggml_webgpu_scale_pipeline_key_hash>
-        scale_pipelines;                                               // inplace
+        scale_pipelines;      // inplace
     std::unordered_map<ggml_webgpu_solve_tri_pipeline_key, webgpu_pipeline, ggml_webgpu_solve_tri_pipeline_key_hash>
-        solve_tri_pipelines;                                           // type
+        solve_tri_pipelines;  // type
     std::unordered_map<ggml_webgpu_ssm_conv_pipeline_key, webgpu_pipeline, ggml_webgpu_ssm_conv_pipeline_key_hash>
-        ssm_conv_pipelines;                                            // type/vectorized
+        ssm_conv_pipelines;   // type/vectorized
     std::unordered_map<ggml_webgpu_gated_delta_net_pipeline_key,
                        webgpu_pipeline,
                        ggml_webgpu_gated_delta_net_pipeline_key_hash>
@@ -789,6 +810,11 @@ class ggml_webgpu_shader_lib {
         rope_pipelines;
     std::unordered_map<ggml_webgpu_soft_max_pipeline_key, webgpu_pipeline, ggml_webgpu_soft_max_pipeline_key_hash>
         soft_max_pipelines;
+
+    std::unordered_map<ggml_webgpu_rms_norm_mul_pipeline_key,
+                       webgpu_pipeline,
+                       ggml_webgpu_rms_norm_mul_pipeline_key_hash>
+        rms_norm_mul_pipelines;
 
   public:
     ggml_webgpu_shader_lib(wgpu::Device device) { this->device = device; }
@@ -1805,13 +1831,48 @@ class ggml_webgpu_shader_lib {
         return unary_pipelines[key];
     }
 
+    webgpu_pipeline get_rms_norm_mul_pipeline(const ggml_webgpu_shader_lib_context & context) {
+        ggml_webgpu_rms_norm_mul_pipeline_key key = {
+            .inplace     = context.inplace,
+            .src_overlap = context.src_overlap,
+        };
+
+        auto it = rms_norm_mul_pipelines.find(key);
+        if (it != rms_norm_mul_pipelines.end()) {
+            return it->second;
+        }
+
+        std::vector<std::string> defines;
+        std::string              op_name = "RMS_NORM_MUL";
+        std::string              variant = op_name;
+
+        if (key.inplace) {
+            defines.push_back("INPLACE");
+            variant += "_inplace";
+        } else if (key.src_overlap) {
+            defines.push_back("SRC_OVERLAP");
+            variant += "_src_overlap";
+        }
+
+        defines.push_back(std::string("WG_SIZE=") + std::to_string(context.max_wg_size));
+
+        auto processed              = preprocessor.preprocess(wgsl_rms_norm_mul, defines);
+        auto decisions              = std::make_shared<ggml_webgpu_generic_shader_decisions>();
+        decisions->wg_size          = context.max_wg_size;
+        webgpu_pipeline pipeline    = ggml_webgpu_create_pipeline(device, processed, variant);
+        pipeline.context            = decisions;
+        rms_norm_mul_pipelines[key] = pipeline;
+        return rms_norm_mul_pipelines[key];
+    }
+
     webgpu_pipeline get_binary_pipeline(const ggml_webgpu_shader_lib_context & context) {
-        ggml_webgpu_binary_pipeline_key key = {};
-        key.type                            = context.dst->type;
-        key.op                              = context.dst->op;
-        key.inplace                         = context.inplace;
-        key.overlap                         = context.overlap;
-        key.src_overlap                     = context.src_overlap;
+        ggml_webgpu_binary_pipeline_key key = {
+            .type        = context.dst->type,
+            .op          = context.dst->op,
+            .inplace     = context.inplace,
+            .overlap     = context.overlap,
+            .src_overlap = context.src_overlap,
+        };
 
         auto it = binary_pipelines.find(key);
         if (it != binary_pipelines.end()) {
--- a/ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
+++ b/ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
@@ -516,7 +516,7 @@ inline uint32_t ggml_webgpu_flash_attn_max_kv_tile(const ggml_webgpu_shader_lib_
     const size_t q_tile       = context.sg_mat_m;
     const size_t base_q_bytes = (key.head_dim_qk + key.head_dim_v) * q_tile * GGML_WEBGPU_F16_SIZE_BYTES +
                                 2 * q_tile * GGML_WEBGPU_F32_SIZE_BYTES;
-    size_t bytes_per_kv = 0;
+    size_t       bytes_per_kv = 0;
     if (!key.kv_direct) {
         bytes_per_kv += std::max(key.head_dim_qk, key.head_dim_v);
     }
@@ -1832,10 +1832,9 @@ class ggml_webgpu_shader_lib {
     }
 
     webgpu_pipeline get_rms_norm_mul_pipeline(const ggml_webgpu_shader_lib_context & context) {
-        ggml_webgpu_rms_norm_mul_pipeline_key key = {
-            .inplace     = context.inplace,
-            .src_overlap = context.src_overlap,
-        };
+        ggml_webgpu_rms_norm_mul_pipeline_key key = {};
+        key.inplace                               = context.inplace;
+        key.src_overlap                           = context.src_overlap;
 
         auto it = rms_norm_mul_pipelines.find(key);
         if (it != rms_norm_mul_pipelines.end()) {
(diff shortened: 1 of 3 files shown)


yomaytk:
@reeselevine  This PR should conflicts with https://github.com/ggml-org/llama.cpp/pull/21873 (especially in `ggml_backend_webgpu_graph_compute`), so I will update accordingly after the PR is merged.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -260,6 +260,9 @@ struct webgpu_context_struct {
 
     size_t memset_bytes_per_thread;
 
+    bool     disable_fusion;
+    uint32_t num_additional_fused_ops;
```
reeselevine, line 264: this general structure comes from the vulkan backend right? I haven't looked into it too closely, but my first thought is that it seems too general, at least based on this PR, because you end up having to check which ops you are actually fusing, and this doesn't encode that at all.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2758,6 +2880,45 @@ static std::optional<webgpu_encoded_op> ggml_webgpu_encode_node(webgpu_context c
...
+            return false;
+        }
+    }
+
+    return true;
+}
+
+static bool ggml_webgpu_can_fuse_check(webgpu_context & ctx, const struct ggml_cgraph * cgraph, int node_idx) {
+    // RMS_NORM + MUL
+    if (ggml_webgpu_can_fuse(cgraph, node_idx, { GGML_OP_RMS_NORM, GGML_OP_MUL })) {
```
reeselevine, line 2917: under the hood `can_fuse` ends up repeating the if condition on RMS_NORM + MUL, so really should we have separate functions for each set of potential fusions?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2837,7 +2999,10 @@ static ggml_status ggml_backend_webgpu_graph_compute(ggml_backend_t backend, str
         if (cgraph->nodes[i]->op == GGML_OP_SET_ROWS) {
             contains_set_rows = true;
         }
-        if (auto cmd = ggml_webgpu_encode_node(ctx, cgraph->nodes[i])) {
+        if (!ctx->disable_fusion) {
+            ggml_webgpu_can_fuse_check(ctx, cgraph, i);
+        }
+        if (auto cmd = ggml_webgpu_encode_node(ctx, cgraph->nodes, i)) {
```
reeselevine, line 3005: Right now we're hiding whether the `encode_node` encodes based on whether the next node will be fused. 

Instead, what do you think of just calling encode node (maybe function name should just change to `encode` to encompass the fact it might encode multiple nodes), and updating `i` based on the number of fused operations. So for the new RMS_NORM + MUL, we end updating `i` by 2. That avoids hiding the fusion in the `additional_fused_ops` variable. That to me seems cleaner for now, but maybe I'm missing something that doesn't translate well to future fusions?

reeselevine (commented):
Thanks for starting to work on fusion, this is a big step!

The performance not changing too much is a little disappointing, but also not a blocker. Once we get the fusion format working we can optimize. I wonder if the reason for the lack of performance is just that the current `RMS_NORM` is not very well-optimized? So the reduction in bandwidth ends up being hidden because RMS_NORM is too slow.

Do you know if this fusion path leads to significant performance gains in other backends?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2081,6 +2084,115 @@ static webgpu_encoded_op ggml_webgpu_repeat(webgpu_context & ctx, ggml_tensor *
...
+        (uint32_t) (dst->nb[3] / ggml_type_size(dst->type)),
+        (uint32_t) mul_src->ne[0],
+        (uint32_t) mul_src->ne[1],
+        (uint32_t) mul_src->ne[2],
+        (uint32_t) mul_src->ne[3],
+        (uint32_t) dst->ne[0],
+        (uint32_t) dst->ne[1],
+        (uint32_t) dst->ne[2],
+        (uint32_t) dst->ne[3],
+        *(uint32_t *) rn_dst->op_params  // epsilon, treated as f32 in the shader
```
reeselevine: this leads to compiler warnings and will fail when the new ggml-webgpu-nvidia-ci is enabled, I moved to a new format: https://github.com/ggml-org/llama.cpp/blob/master/ggml/src/ggml-webgpu/ggml-webgpu.cpp#L1910

ggml/src/ggml-webgpu/ggml-webgpu.cpp
yomaytk: Yes, it is almost the same as the vulkan backend, but I agree that this is too general. I removed this as I explain below.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
yomaytk: You're right, it looks better to have a specific `can_fuse` function for each fusion.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
yomaytk: Sounds good. It looks like whether the ops can be fused can basically be determined by `cgraph->nodes`, at least from my understanding. So I moved the `can_fuse` logic to `ggml_webgpu_encode`, which removes `num_additional_fused_ops` and also eliminates the double-checking for fusion (`ggml_webgpu_can_fuse_check` in `ggml_backend_webgpu_graph_compute`and `ctx->num_additional_fused_ops > 0`). I think this fix addresses your comment. How about this?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
yomaytk: Got it, I’ll update after this PR is rebased because this fix depends on the new function `ggml_webgpu_u32_from_f32`.

yomaytk:
I re-ran the benchmarks on the same three models with `-r 10` and measured
op-level GPU time for RMS_NORM + MUL using the profiling build
(`GGML_WEBGPU_CPU_PROFILE` / `GGML_WEBGPU_GPU_PROFILE`).

As in the first test, end-to-end `t/s` is nearly unchanged, but the GPU
profiling time for the RMS_NORM + MUL path is clearly reduced by the fusion.
The only Qwen3.5 pp512 has op-level regression.

The command is like this:
`$ llama-bench -m gemma-4-E4B-it-Q4_K_M.gguf -fa 1 -p 0,1,512 -n 0,1,128 -r 10`

Llama-3.2-3B-Instruct Q4_K_M
| Test   | No-fused (t/s) | Fused (t/s)   | Speedup | Op-level time (no-fused → fused) | Occupancy (% of GPU) |
|--------|---------------:|--------------:|:-------:|:---------------------------------:|:--------------------:|
| pp1    |  52.42  |  52.72 | 1.006x  | 1.46 → 0.83 ms (-43%)             | 1.79% → 1.03%        |
| pp512  | 671.16  | 670.74 | 0.999x  | 16.13 → 6.98 ms (-57%)            | 0.82% → 0.36%        |
| tg1 †  |  51.61  |  51.56 | 0.999x  | 1.54 → 0.88 ms (-43%)             | 1.82% → 1.04%        |
| tg128  |  52.73  |  52.98 | 1.005x  | 180.68 → 101.81 ms (-44%)         | 1.90% → 1.08%        |

Qwen3.5-4B Q4_K_M
| Test   | No-fused (t/s) | Fused (t/s)    | Speedup | Op-level time (no-fused → fused) | Occupancy (% of GPU) |
|--------|---------------:|---------------:|:-------:|:---------------------------------:|:--------------------:|
| pp1    |  30.59  |  30.17  | 0.986x  | 1.97 → 1.22 ms (-38%)             | 1.78% → 1.10%        |
| pp512  | 480.00  | 472.25  | 0.984x  | 28.22 → 47.99 ms (+70%)           | 1.17% → 1.93%        |
| tg1    |  30.44  |  30.45  | 1.000x  | 2.01 → 1.21 ms (-40%)             | 1.81% → 1.09%        |
| tg128  |  30.44  |  30.14  | 0.990x  | 238.62 → 154.62 ms (-35%)         | 1.83% → 1.20%        |

Gemma-4-E4B-it Q4_K_M

| Test   | No-fused (t/s) | Fused (t/s)   | Speedup | Op-level time (no-fused → fused) | Occupancy (% of GPU) |
|--------|---------------:|--------------:|:-------:|:---------------------------------:|:--------------------:|
| pp1    |  31.88  |  31.94 | 1.002x  | 4.91 → 2.70 ms (-45%)             | 4.19% → 2.35%        |
| pp512  | 465.46  | 466.11 | 1.001x  | 51.48 → 37.71 ms (-27%)           | 2.04% → 1.50%        |
| tg1    |  31.77  |  32.01 | 1.008x  | 4.93 → 2.79 ms (-43%)             | 4.17% → 2.41%        |
| tg128  |  32.03  |  32.19 | 1.005x  | 609.67 → 336.46 ms (-45%)         | 4.39% → 2.48%        |

Using `GGML_WEBGPU_GPU_PROFILE`, I confirmed that the ops related to RMS_NORM and MUL in these three models are as follows.
- no-fused: `rms_norm`, `rms_norm_inplace`, `MUL_f32_inplace`
- fused: `RMS_NORM_MUL`, `MUL_F32_inplace`

Notes on the table:
- Op-level time = GPU profile time of the ops that get fused.
    - no-fused: `rms_norm` + `rms_norm_inplace` + (`MUL_f32_inplace`(no-fused) − `MUL_f32_inplace`(fused))
    - fused: `RMS_NORM_MUL`
- Occupancy = (op-level time) / (total GPU time).

> I wonder if the reason for the lack of performance is just that the current RMS_NORM is not very well-optimized? So the reduction in bandwidth ends up being hidden because RMS_NORM is too slow.

That's plausible, yes. In addition, the RMS_NORM + MUL path's share of total
GPU time is small on WebGPU/M2 — roughly 1–4% depending on the model — so
I think that even this op-level reduction only moves end-to-end t/s by up to around 1%. I expect
this fusion's benefit to become more visible once RMS_NORM and the surrounding
ops are further optimized.

> Do you know if this fusion path leads to significant performance gains in other backends?

I haven't measured other backends in my environment, but the corresponding
CUDA and Metal fusion PRs (#14800, #14596) report ~5–10% improvements on
pp/tg tests, which I'd expect here too once the other implementation is optimized.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2662,15 +2773,48 @@ static webgpu_encoded_op ggml_webgpu_sum_rows(webgpu_context & ctx, ggml_tensor
...
+
+    return true;
+}
+
 // Returns the encoded command, or std::nullopt if the operation is a no-op
-static std::optional<webgpu_encoded_op> ggml_webgpu_encode_node(webgpu_context ctx, ggml_tensor * node) {
+static std::optional<webgpu_encoded_op> ggml_webgpu_encode(webgpu_context ctx,
+                                                           ggml_cgraph *  cgraph,
+                                                           int            node_idx,
+                                                           int &          num_encoded_ops) {
```
reeselevine, line 2510: I think there might be a cleaner way of returning the number of encoded ops then passing in the extra parameter by reference, but I also realize that the return of the optional encoded_op makes it slightly more complex. One option would be to just put `num_encoded_ops` as a field in `webgpu_encoded_op`. But I think this is also ok for now, I might think about it a bit more.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2713,6 +2857,13 @@ static std::optional<webgpu_encoded_op> ggml_webgpu_encode_node(webgpu_context c
         case GGML_OP_REPEAT:
             return ggml_webgpu_repeat(ctx, src0, node);
         case GGML_OP_RMS_NORM:
+            if (ggml_webgpu_can_fuse_rms_norm_mul(ctx, cgraph, node_idx)) {
```
reeselevine: I think this is a nice way of doing the check, thanks!

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -3439,7 +3595,8 @@ static webgpu_context initialize_webgpu_context(ggml_backend_dev_t dev) {
     ggml_backend_webgpu_device_context * dev_ctx    = (ggml_backend_webgpu_device_context *) dev->context;
     webgpu_context                       webgpu_ctx = std::make_shared<webgpu_context_struct>();
     webgpu_ctx->global_ctx                          = dev_ctx->webgpu_global_ctx;
-    webgpu_ctx->shader_lib = std::make_unique<ggml_webgpu_shader_lib>(dev_ctx->webgpu_global_ctx->device);
+    webgpu_ctx->shader_lib     = std::make_unique<ggml_webgpu_shader_lib>(dev_ctx->webgpu_global_ctx->device);
+    webgpu_ctx->disable_fusion = getenv("GGML_WEBGPU_DISABLE_FUSION") != nullptr;
```
reeselevine: I'm hesitant to include environment variables in the WebGPU backend, ideally all the logic for why we should/should not do something can exist in the backend code itself. Is there a specific reason for including disabling fusion as an option like this?

reeselevine (commented):
Needs a merge or rebase but otherwise this is looking good!

ggml/src/ggml-webgpu/ggml-webgpu.cpp
yomaytk: Got it, thanks. I leave the current form for now then.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
yomaytk: This is because I followed the other backends (vulkan, cuda and opencl have this env var), so I didn't have a ton of thoughts about this. I confirmed that the PR (https://github.com/ggml-org/llama.cpp/pull/14545) adds the same env var to the vulkan backend because an execution error was reported after the RMS_NORM + MUL fusion PR was merged. However, the error was actually due to fusion implementation bugs, so it seems they added the env var only as a tentative guard. I realize we should be cautious about adding more external dependencies too, and I'm not sure disabling fusion is necessary for now. So I'm starting to think we should remove this env var and `disable_fusion` for now. What do you think?

reeselevine: I agree, I think we can remove it, and if we do run into issues due to fusion, we can try to fix forward or try to add guards within the WebGPU backend itself to disable it.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -1805,13 +1831,48 @@ class ggml_webgpu_shader_lib {
...
+
     webgpu_pipeline get_binary_pipeline(const ggml_webgpu_shader_lib_context & context) {
-        ggml_webgpu_binary_pipeline_key key = {};
-        key.type                            = context.dst->type;
-        key.op                              = context.dst->op;
-        key.inplace                         = context.inplace;
-        key.overlap                         = context.overlap;
-        key.src_overlap                     = context.src_overlap;
+        ggml_webgpu_binary_pipeline_key key = {
+            .type        = context.dst->type,
```
reeselevine: I'm trying to move away from using the C++20 initializers, they lead to compiler warnings right now and when I enable the ggml-ci-webgpu-nvidia node they will cause errors, as that CI is stricter.

yomaytk: Got it, I updated accordingly.

reeselevine (approved):

CISC (approved):

PR #21988, opened 2026-04-16
Avoid redundant memory_breakdown() computation in device memory probe

Jessen-Li:
## Overview

This PR removes a redundant computation of ctx->memory_breakdown() during device memory probing in llama_get_device_memory_data().
### Changes
- Introduce llama_memory_breakdown_print_impl which accepts a precomputed memory_breakdown object.
- Reuse the existing memory_breakdown result from llama_get_device_memory_data() instead of recomputing it in
  llama_memory_breakdown_print()
### Motivation
llama_get_device_memory_data() may be called multiple times during fit/probe workflows (e.g. baseline, context probing, repeated resource checks). Each call currently recomputes memory_breakdown, leading to unnecessary duplicated work.

This change ensures:
- No redundant computation within a single probe
- Consistent snapshot used for both computation and logging
- Slight reduction in overhead in repeated probe scenarios
### Behavior
No functional change is intended.
Output and memory accounting remain identical.

# Requirements
No new requirements introduced by this change.

- AI usage disclosure: 
NO.

### DIFF
--- a/src/llama-context.cpp
+++ b/src/llama-context.cpp
@@ -3493,11 +3493,11 @@ void llama_perf_context_reset(llama_context * ctx) {
     ctx->perf_reset();
 }
 
-void llama_memory_breakdown_print(const struct llama_context * ctx) {
+
+void llama_memory_breakdown_print_impl(
+        const struct llama_context * ctx,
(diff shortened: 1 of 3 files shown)


Jessen-Li:
Looks like recent changes have refactored this area and the redundant memory_breakdown() call no longer exists. Closing this PR as it’s no longer applicable.

PR #21995, opened 2026-04-16
sycl: add multi-subgroup MMVQ kernel for reorder path (+50% Q4_0)

sleepyeldrazi:
## Summary

Add a multi-subgroup variant of the reorder MMVQ kernel that distributes multiple subgroups across each row, enabling much higher GPU occupancy for quantized token generation.

## Changes
New template `mul_mat_vec_q_reorder_multi_sg` that:
- Takes a `subgroups_per_row` template parameter
- Each subgroup computes a partial dot product for its block range
- Results are reduced via local memory (group_barrier + manual reduction)
- Single thread writes final sum

### Subgroups per row (per quant type)
| Type | subgroups_per_row | Threads/row | Notes |
|------|-------------------|-------------|-------|
| Q4_0 | 32 | 1024 | Maximum parallelism, safe |
| Q4_K | 8 | 256 | |
| Q6_K | 8 | 256 | |

Q8_0 keeps the existing single-subgroup path (memory-BW-bound, not dp4a-bound).

⚠️ **Q8_0 crashes at 32 subgroups/row** (UR_RESULT_ERROR_DEVICE_LOST). 8 subgroups/row is the safe maximum for Q8_0.

## Benchmark (Arc A770 16GB, Qwen3.5-9B)
| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| Q4_0 tg128 | 39.5 t/s | 59.2 t/s | **+50%** |

## Files changed
- `ggml/src/ggml-sycl/mmvq.cpp` — new multi-subgroup template and dispatch wrappers

### DIFF
--- a/ggml/src/ggml-sycl/mmvq.cpp
+++ b/ggml/src/ggml-sycl/mmvq.cpp
@@ -534,21 +534,91 @@ static void mul_mat_vec_q_iq4_xs_q8_1(const void *__restrict__ vx,
     }
 }
 
+template <typename reorder_vec_dot_q_sycl, int subgroups_per_row = 1>
+static void mul_mat_vec_q_reorder_multi_sg(const void * __restrict__ vx, const void * __restrict__ vy, float * __restrict__ dst,
+                                  const int ncols, const int nrows, const sycl::nd_item<3> & nd_item, sycl::local_accessor<float, 1> local_mem) {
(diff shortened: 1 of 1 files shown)


sleepyeldrazi:
Closing — these PRs were incorrectly opened against the main repo. They will be re-opened on the fork. Sorry for the noise.

PR #21998, opened 2026-04-16
rpc : refactor the RPC transport

rgerganov:
Move all transport related code into a separate file and use the socket_t interface to hide all transport implementation details.

- AI usage disclosure: NO


rgerganov:
Did some testing on the DGX cluster and haven't find any regressions so far.
The change is ready for review.

ggerganov (approved):

PR #22003, opened 2026-04-16
server : fix proxy client timeout in router mode

### LINKED ISSUE #18760: server: proxy timeout in router mode ignores --timeout parameter
## Bug Description

In multi-model router mode (`--models-dir`), the internal HTTP proxy client uses hardcoded timeout values instead of respecting the `--timeout` CLI parameter. This causes "Failed to read connection" errors when processing large prompts with slow models.

## Steps to Reproduce

1. Start server in router mode:
```bash
llama-server --models-dir ./models --timeout 36000 --port 8087
```

2. Send a request to a large model (e.g., 70B+ parameters on CPU)

3. Observe error after ~5 minutes:
```
http client error: Failed to read connection
```

## Expected Behavior

The proxy should respect `--timeout` parameter, same as single-model mode (`-m model.gguf`).

## Actual Behavior

Proxy uses hardcoded timeout:
- Connection timeout: 200ms
- Read/Write timeout: httplib default (~300 seconds)

This makes router mode unusable for large CPU models where a single request can take several hours.

## Environment

- Large models (70B-235B parameters) running on CPU
- Long context prompts (10k+ tokens)
- Request processing time: 30 minutes to several hours

## Suggested Fix

In `tools/server/server-models.cpp`, the `server_http_proxy` constructor should receive and use `timeout_read` from `base_params` instead of hardcoded values:

```cpp
// Current code:
cli->set_connection_timeout(0, 200000); // 200 milliseconds

// Should be:
cli->set_connection_timeout(timeout_seconds);
cli->set_read_timeout(timeout_seconds);
cli->set_write_timeout(timeout_seconds);
```

## Workaround

Use single-model mode (`-m model.gguf`) instead of router mode (`--models-dir`).

xris99:
## Problem

In router mode (`--models-dir` / `--models-preset`), the internal HTTP proxy client in `server_http_proxy` had two bugs that caused `http client error: Failed to read connection` errors:

1. **Hardcoded 5-second connection timeout** — ignored the `--timeout` parameter entirely, causing failures when child model workers take longer than 5 seconds to accept the connection (common for large CPU-bound models).
2. **Swapped read/write timeouts** — `set_write_timeout` was fed `timeout_read` and `set_read_timeout` was fed `timeout_write`. The original comment acknowledged this with _"reversed for cli (client) vs srv (server)"_ but the reasoning was wrong — the names mean the same thing from both sides.

This made router mode unusable for large models (70B+) on CPU where a single request can take 30 minutes or more.

## Fix

Three-line change in `tools/server/server-models.cpp`:

- Connection timeout now uses `timeout_read` (same as the read timeout) instead of hardcoded 5 s
- `set_read_timeout` ← `timeout_read`
- `set_write_timeout` ← `timeout_write`

This matches exactly how `server-http.cpp` (lines 115–116) sets timeouts for the non-router path, making `--timeout` consistent across both modes.

## Testing

Starting server in router mode with a large CPU model and `--timeout 36000` no longer produces `Failed to read connection` after the default httplib timeout.

Fixes #18760

### DIFF
--- a/tools/server/server-models.cpp
+++ b/tools/server/server-models.cpp
@@ -1146,7 +1146,7 @@ server_http_proxy::server_http_proxy(
 
     // setup Client
     cli->set_follow_location(true);
-    cli->set_connection_timeout(5, 0); // 5 seconds
+    cli->set_connection_timeout(timeout_read, 0); // use --timeout value instead of hardcoded 5 s
     cli->set_write_timeout(timeout_read, 0); // reversed for cli (client) vs srv (server)
     cli->set_read_timeout(timeout_write, 0);
     this->status = 500; // to be overwritten upon response


tools/server/server-models.cpp
```diff
@@ -1146,9 +1146,9 @@ server_http_proxy::server_http_proxy(
 
     // setup Client
     cli->set_follow_location(true);
-    cli->set_connection_timeout(5, 0); // 5 seconds
-    cli->set_write_timeout(timeout_read, 0); // reversed for cli (client) vs srv (server)
-    cli->set_read_timeout(timeout_write, 0);
```
ngxson, line 1151: this is supposed to be reversed, right? ref the comment that you deleted

xris99: The comment is an incorrect reasoning and the timeouts are swapped wrongly. From the httplib client's perspective, read = reading the response, write = sending the request body, which maps directly to the server-side meanings. In consequence the timeouts must not be swapped.

ngxson: IIRC the `timeout_read/write` is read from CLI arg which is intended to be set to server. have you acknowledge where this params come from?

xris99: Hmm,... this actually does not make sense at all and could be significantly simplified.
timeout_read and timeout_write are always identical. The --timeout CLI arg sets both to the same value (arg.cpp l2960–2961), defaults are also 600 for both. They cannot be set individually. Swapping them has zero effect; never matteres. I can revert and insert the swap again, but it has zero effect. Just tried to clean up a bit

xris99:
You are right, I concede the point on the swap. `timeout_read` and `timeout_write` are always identical — `--timeout` sets both to the same value (arg.cpp l2960–2961) and there is no way to set them individually. Swapping them has zero practical effect and never mattered. I tried to clean it up but the reorder is noise. Amended the commit to revert the swap and updated the commit message accordingly — the only real fix is replacing the hardcoded 5-second connection timeout with `timeout_read`.

tools/server/server-models.cpp
ngxson: you should revert the comment too (in case someone else trying to do the same change here)

xris99: You are right. reverted completely incl. the comment. Sorry for that. Updated with last commit. single line change now for the router timeout. (the actual topic)

ngxson (approved):
@ServeurpersoCom can you merge this? thanks!

ServeurpersoCom (approved):

PR #22004, opened 2026-04-16
model: move `load_hparams` and `load_tensors` to per-model definition

### LINKED ISSUE #21966: Refactor: move `load_hparams` and `load_tensors` to per-model definition

ngxson:
## Overview

Fix https://github.com/ggml-org/llama.cpp/issues/21966

The migration will be done via a script `0migrate.py` included in this PR (and will be removed right before this is merged)

Important note: the goal of this PR is to make sure the migration is as deterministic as possible, we do that via a completely heuristic script as mentioned above. Any improvements (deduplication, clean up, etc) will be done via follow-up PRs

Depends on:
- [x] https://github.com/ggml-org/llama.cpp/pull/21245
- [x] https://github.com/ggml-org/llama.cpp/pull/21971
- [x] https://github.com/ggml-org/llama.cpp/pull/22005
- [x] https://github.com/ggml-org/llama.cpp/pull/21970

Checklist before merging:
- [x] Remove migration script
- [x] Remove `#if 0` pre-migration places

## Migration guide

If you are working on a PR for model implementation, this change will introduce a merge conflict. To resolve it:
1. Change the model definition in `models.h` to extend from `llama_model_base`
2. Move hparams and tensors loader to the created class
3. Move cgraph namespace to `(model_name)::graph::graph`
4. (optional) revert any redundant changes that you made to `llama-model.cpp` ; hparams and tensors loader are now outside of `llama-model.cpp`

### Migration rules:

- Create one file per arch, the file name is the arch enum name in lower case, with `_` replaced by `-`
- Model name is arch name in lower case, prefixed by `llama_model_`

Given 2 example arch A and B:
- If arch A and B BOTH use the same `load_tensors` AND `load_hparams`, B inherit A
- If either `load_tensors` OR `load_hparams` is different, no inherit, there will be one of the 2 functions being duplicated
- If B reuse the same graph with A, the graph definition in B will be `using graph = llama_model_a::graph`

Side effects:
- Some part of the code will be duplicated (i.e. some `load_tensors` or `load_hparams`)
- No more `-iswa` suffix in file name or in model class name

Self-note for updating this PR:

```sh
git reset --hard HEAD~1
git merge master
python 0migrate.py
git add -A
git commit -m "auto"
git push -f
```

# Requirements

- AI usage disclosure: partially, mostly for the migration script


ngxson:
Alright, first pass of the migration (fully generated from the script --> https://github.com/ggml-org/llama.cpp/pull/22004/commits/69104f14d85978a146df5b19a7b58da6302d7678)

Asking @ggerganov @CISC @pwilkin for a pre-review

src/llama-model.cpp
```diff
@@ -33,6 +33,527 @@
...
 
+llama_model * llama_model_create(llm_arch arch, const llama_model_params & params) {
+    llama_model * model = nullptr;
+
+    switch (arch) {
+        case LLM_ARCH_LLAMA:
+            {
+                model = new llama_model_llama(params);
+            } break;
+        case LLM_ARCH_LLAMA4:
```
CISC, line 40: Surely there must be some other way to do this without this switch?

ngxson: Yes and no, technically we can hide the `case` behind a macro like:

```
ARCH_DEF(LLAMA)
```

That will expand to the same `case LLM_ARCH_LLAMA: ... break;` (though I'm not 100% sure if a preprocessor macro like `tolower` or `toupper`exists or not)

Another way is to bake the `arch` into model class definition, but we will still need to initialize the model instance just to look at its `arch`

In any cases, it's technically hard to escape from any shorts of `switch..case` here in C/C++. After all, a struct definition in c++ is no more than bunch of function pointers, unlike higher-level languages like python or java where the notion of "class" does exist at runtime.

ngxson: > struct definition in c++ is no more than bunch of function pointers

Another extreme hacky way is to arrange the vpointer table to follow exactly the same order as the enum, then calculate the correct function pointer from a given enum. But this is kinda science fiction IMO

CISC: > > struct definition in c++ is no more than bunch of function pointers
> 
> Another extreme hacky way is to arrange the vpointer table to follow exactly the same order as the enum, then calculate the correct function pointer from a given enum. But this is kinda science fiction IMO

My thoughts were along those lines; that it should be possible to make a link-time table of the classes in enum order.

ngxson: Anw, I improved this a bit to make it shorter:

```cpp
        case LLM_ARCH_LLAMA:
            return new llama_model_llama(params);
        case LLM_ARCH_LLAMA4:
            return new llama_model_llama4(params);
        case LLM_ARCH_LLAMA_EMBED:
            return new llama_model_llama_embed(params);
```

src/models/models.h
```diff
@@ -110,611 +111,1740 @@ struct llm_build_rwkv7_base : public llm_graph_context {
...
+
+struct llama_model_llama_embed : public llama_model_llama {
+    llama_model_llama_embed(const struct llama_model_params & params) : llama_model_llama(params) {}
+    // reuse load_hparams and load_tensors from llama_model_llama
+
+    template <bool embed>
+    using graph = llama_model_llama::graph<embed>;
+
+    std::unique_ptr<llm_graph_context> build_graph_context(const llm_graph_params & params) const override;
 };
```
ngxson, line 150: The auto-migration script can't be intelligent enough to point out that we can use the specialization `<true>` of the graph, but things like this can be improved via a follow-up PR (just noting here for visibility)

src/llama-model.cpp
pwilkin: Can't you have an arch-to-constructor-method map? Not much cleaner, but at least avoids the switch...

ngxson: Can you provide an example code with for example, let's say 3 classes a,b,c and an enum?

pwilkin:
```c++
using Factory = std::function<std::unique_ptr<Base>()>;

static const std::unordered_map<Foo, Factory> factories = {
    { A, []{ return std::make_unique<AClass>(); } },
    { B, []{ return std::make_unique<BClass>(); } },
    { C, []{ return std::make_unique<CClass>(); } },
};

std::unique_ptr<Base> create(Foo f) {
    return factories.at(f)();
}
```

src/llama-model.cpp
```diff
@@ -3038,73 +3312,72 @@ bool llama_model::load_tensors(llama_model_loader & ml) {
...
-        const int64_t n_ctx_train   = hparams.n_ctx_train;
 
-        if (n_expert > 0 && hparams.n_expert_used == 0) {
+        if (n_expert > 0 && n_expert_used == 0) {
             throw std::runtime_error("model has expert layers but no expert layers are used");
         }
 
+        // helper function to facilitate migration
+        // TODO: remove this in the future
         auto create_tensor = [&](const LLM_TN_IMPL & tn, const std::initializer_list<int64_t> & ne, int flags) -> ggml_tensor * {
```
pwilkin: Would rename the lambda, having both the method and the lambda named `create_tensor` is a bit confusing.

src/llama-model.cpp
ngxson: IMO that's not much different from the current switch..case implementation. The std::unordered_map will be slightly worse because:
- The map is constructed at runtime (versus switch being compiled statically into a jump table)
- Each look up on the map need to call hash function (versus switch, just a simple jump)
- Lambda function need to be compiled which will produce some overheads. While it's not very significant, does technically cost compilation time

Beside, I can't think of any cases where the std::map can do but switch...case cannot (except for being modifiable at runtime, which we won't support anyway)

src/llama-model.cpp
ngxson: I think the better way is to completely move the current function the llm_arch_model_i and simply reuse the create_tensor there. Will push a fix for this

I try not to renaming things to reduce the number of line changes in this PR

src/llama-model.cpp
pwilkin: More elegant ;)

pwilkin: BTW it's static const, so the compiler will probably optimize it anyway, won't be constructed at runtime FWIW.

src/llama-model.cpp
ngxson: This lambda is removed in the latest version

src/llama-model.cpp
ngxson: > More elegant ;)

I don't get it... how can this:

```
{ A, []{ return std::make_unique<AClass>(); } },
```

more elegant than this:

```
case A: return new AClass(); // no lambda is needed
```

> BTW it's static const, so the compiler will probably optimize it anyway, won't be constructed at runtime FWIW.

IIRC it only does that with -O2 or -O3. Switch statement is always compiled to static jump table no matter what. So, why not?

Also, the map always requires calculating hash to look up elements.

ngxson:
Not very urgent but pinging @ggerganov if you can have a quick look on the direction

src/llama-model.cpp
pwilkin: Personal preference :)

src/llama-model.h
```diff
@@ -623,16 +617,81 @@ struct llama_model {
...
+llama_model * llama_model_create(llm_arch arch, const llama_model_params & params);
+llama_model * llama_model_create(llama_model_loader & ml, const llama_model_params & params);
+
+// provide addition context for loading
+// model must inherit from this
+struct llm_arch_model_i : public llama_model {
+    friend struct llama_model;
+
+    llama_model * model;
+    llama_model_loader * ml = nullptr;
```
ggerganov, line 646: I don't think introducing the `llm_arch_model_i` here is necessary. You should keep using `llama_model` for now and inherit the implementations directly from it. The `llm_arch_model_i` idea is separate - see below. Here you are interested just in localizing the model definitions (loading hparams, tensors, memory and graph creation) into individual files.

---

There is a separate refactoring task that can be done before or after this PR. The final state should be like this:

```c++
//
// llama.h
//

typedef struct llama_model_i * llama_model_t;

...

LLAMA_API int32_t llama_model_n_ctx_train(const llama_model_t model);
LLAMA_API int32_t llama_model_n_embd     (const llama_model_t model);
LLAMA_API int32_t llama_model_n_embd_inp (const llama_model_t model);

...

//
// llama-model.h
//

// pure interface
struct llama_model_i {
    virtual ~llama_model_i() = default;

    // public API mirror of llama.h
    virtual int32_t n_ctx_train() const = 0;
    virtual int32_t n_embd() const = 0;
    virtual int32_t n_embd_inp() const = 0;

    ...

    // internal API
    virtual bool load_hparams(llama_model_loader * ml, ...) = 0;
    virtual bool load_tensors(llama_model_loader * ml, ...) = 0;

    virtual llama_memory_i * create_memory(const llama_memory_params & params, const llama_cparams & cparams) const = 0;

    virtual ggml_cgraph * build_graph(const llm_graph_params & params) const = 0;

    ...
};

// base model (common functionality and data for all models)
class llama_model_base : public llama_model_i {
public:
    int32_t n_ctx_train() override;
    int32_t n_embd() override;
    int32_t n_embd_inp() override;

    ...

protected:
    llm_type type = LLM_TYPE_UNKNOWN;
    llm_arch arch = LLM_ARCH_UNKNOWN;

    std::string name = "n/a";

    llama_hparams hparams = {};
    llama_vocab   vocab;

    ggml_tensor * tok_embd   = nullptr;
    ggml_tensor * type_embd  = nullptr;
    ggml_tensor * pos_embd   = nullptr;
    ggml_tensor * tok_norm   = nullptr;
    ggml_tensor * tok_norm_b = nullptr;

    ...

    ggml_tensor * create_tensor(llama_model_loader * ml, ...);

    // helpers
    void create_tensor_gate_up_exps(llama_model_loader * ml, ...);
    void create_tensor_qkv         (llama_model_loader * ml, ...);

    ...
};

//
// models/models.h
//

class llama_model_qwen3 : public llama_model_base {
public:
    bool load_hparams(llama_model_loader * ml, ...) override;
    bool load_tensors(llama_model_loader * ml, ...) override;

    llama_memory_i * create_memory(const llama_memory_params & params, const llama_cparams & cparams) const override;

    ggml_cgraph * build_graph(const llm_graph_params & params) const override;

    ...
};
```

After this change, all code in `llama_context` should use only `llama_model_i`, similar to how it uses `llama_memory_i`.

ngxson: That makes sense. For this PR, I think I will target a state where all the core definition (`load_hparams`/`load_tensors` / `build_graph`) being moved into `src/models`. I won't separate the `llama_model` / `llama_model_i` right now to simplicity, that can be done in a follow-up.

Just one thing not very clear from your example though, is `llama_model_base` is same as `llama_model_i` ? For now, I think I will add an alias `using llama_model_base = llama_model_i` so that `llama_model_base` can be reserved for future use.

ggerganov: > Just one thing not very clear from your example though, is llama_model_base is same as llama_model_i ?

Had a typo: `class llama_model` -> renamed to `class llama_model_base`. The separate model instances will inherit the `llama_model_base` because there is a lot of common stuff that we want to avoid repeating (e.g. tensors, hparams, devices, buffers, ...). For now inheriting a base implementation is easier way to deduplicate.

The alternative is using composition, which is usually considered better architecturally. But there will be a lot of duplicated code:

```c++
//
// llama-model.h
//

// pure interface
struct llama_model_i {
    virtual ~llama_model_i() = default;

    // public API mirror of llama.h
    virtual int32_t n_ctx_train() const = 0;
    virtual int32_t n_embd() const = 0;
    virtual int32_t n_embd_inp() const = 0;

    ...

    // internal API
    virtual bool load_hparams(llama_model_loader * ml, ...) = 0;
    virtual bool load_tensors(llama_model_loader * ml, ...) = 0;

    virtual llama_memory_i * create_memory(const llama_memory_params & params, const llama_cparams & cparams) const = 0;

    virtual ggml_cgraph * build_graph(const llm_graph_params & params) const = 0;

    ...
};

//
// models/models.h
//

// in this case, directly implement the interface
class llama_model_qwen3 : public llama_model_i {
public:
    // note each model implements these over and over again
    int32_t n_ctx_train() override;
    int32_t n_embd() override;
    int32_t n_embd_inp() override;

    ...

    bool load_hparams(llama_model_loader * ml, ...) override;
    bool load_tensors(llama_model_loader * ml, ...) override;

    llama_memory_i * create_memory(const llama_memory_params & params, const llama_cparams & cparams) const override;

    ggml_cgraph * build_graph(const llm_graph_params & params) const override;

    ...

private:
    // composition instead of inheritance for reusing common model functionality
    llama_model_base model;
};
```

pwilkin: I'd say "composition being better structurally" is a commonly repeated modern programming trend, but in this case, inheritance seems like the better approach since there isn't really a "part" of a model that can be conceptualized as one we're delegating to.

ngxson: Hmm yeah the composition will make the code quite verbose, I would prefer staying with the inheritance pattern for now. So just to make sure I understand it correctly, the current opaque pointer `llama_model` will be mapped to the to-be-added `llama_model_i`, right?

I think it make sense to do inheritance as in your first comment, so that:
- `llama_model_i` holds the definition of the model (i.e. mostly hparams)
- `llama_model_base` holds the tensors
- `llama_model_*` defines how to load tensors and hparams

ggerganov: More like:

- `llama_model_i` abstract interface, does not hold or implement anything. Replaces the old `llama_model`
- `llama_model_base` holds hparams, tensors, devices, meta data, loras, etc. Implements common part of the interface (mostly getters for hparams, devices, loras, etc.)
- `llama_model_*` implements the rest of the interface: loading hparams, loading tensors, creating memory, building graph

ngxson: One concern about llama_model_i holding nothing (and only being an interface) is that we don't yet have any use case where another implementation other than llama_model_base that will reuse the same interface.

I think it may still worth a discussion about the separation of llama_model_i / _base, but that's not very urgent, it will be done in a follow-up anyway

src/llama-model.h
```diff
@@ -620,19 +614,91 @@ struct llama_model {
...
+    int64_t n_embd_head_k;
+    int64_t n_embd_head_v;
+    int64_t n_ff;
+    int64_t n_embd_gqa;
+    int64_t n_vocab;
+    int64_t n_token_types;
+    int64_t n_rot;
+    int64_t n_expert;
+    int64_t n_expert_used;
+    int64_t n_ctx_train;
```
ngxson, line 666: So I refactor a bit the `llm_model_arch_i` into `llama_model_base`, which is not 100% what we had in the last comment, but I think it should be enough as a middle step before another migration (which will hopefully produce less changes than the current PR)

One main concern is that currently, these variables are being defined inside the struct, so that `load_tensors()` doesn't need to load them manually. Although, do you think it's more preferable to define those inside the `load_tensors()`? I'm not even sure if such change will be possible with my script, but I can try if that's preferable.

CC @ggerganov @CISC @pwilkin if you have any thoughts.

pwilkin: This on the other hand might be a good place for composition :) is there any reason we can't just use the internal `hparams` here to keep the parameters?

ngxson: Because it's made this way from the beginning:

https://github.com/ggml-org/llama.cpp/blob/750579ff14198fe964ab7fc5565b1d77600deab4/src/llama-model.cpp#L3065-L3080

And because I don't want to change too much code in the same PR, I'm keeping this list so that `load_tensors()` stays the same.

ggerganov: You can make a `LLAMA_LOAD_LOCALS` macro similar to `GGML_TENSOR_LOCALS` and use the macro at the start of each `load_tensors()` implementation.

src/llama-model.h
```diff
@@ -620,21 +614,94 @@ struct llama_model {
...
+    std::unique_ptr<llm_graph_context> build_arch_graph(const llm_graph_params & params) const override = 0;
+};
+
 const char * llm_type_name(llm_type type);
 
+// convenience macro for loading local variables for load_tensors() in llama_model_base
+// note: cast to int64_t since we will use these for the tensor dimensions
+#define LLAMA_LOAD_LOCALS \
+    const int     n_layer        = hparams.n_layer;          GGML_UNUSED(n_layer); \
+    const int64_t n_head         = hparams.n_head();         GGML_UNUSED(n_head); \
```
ngxson, line 689: @ggerganov alright, I added the `LLAMA_LOAD_LOCALS` as suggested

`GGML_UNUSED` is a hack here to avoid unused variable warning. tbh I don't know if there is a better way, so I'm open for suggestions here

ggerganov (approved):

pwilkin:
@ngxson need to remove the migration scripts now.

ngxson:
yes thanks, let's also wait for a review from @CISC before I do a final rebase & reapply the script

CISC:
> yes thanks, let's also wait for a review from @CISC before I do a final rebase & reapply the script

Looks good, please proceed.

ngxson:
Maybe we should also enforce the model file name to match the arch name (for ex. `my-model.cpp` --> `LLM_ARCH_MY_MODEL`), your thoughts @CISC ?

(The idea is to have a script in CI to validate this)

CISC:
> Maybe we should also enforce the model file name to match the arch name (for ex. `my-model.cpp` --> `LLM_ARCH_MY_MODEL`), your thoughts @CISC ?
> 
> (The idea is to have a script in CI to validate this)

Not a bad idea, we can perhaps add a few other validations to simplify our life. :)

ngxson:
I think we should merge this soon today to avoid conflicts, can you give an approval @CISC @pwilkin ?

pwilkin (approved):

CISC (approved):

JohannesGaessler:
After this PR has been merged `llama_model::type` is being set incorrectly to `LLM_TYPE_7B` for LLaMA 3.1 8b. During the determination in `llama_model_llama::load_arch_hparams` `vocab.n_tokens()` returns 0. I re-created the GGUF file from safetensors to make sure that I am not accidentally using an outdated GGUF file and the issue persists.

CISC:
> After this PR has been merged `llama_model::type` is being set incorrectly to `LLM_TYPE_7B` for LLaMA 3.1 8b. During the determination in `llama_model_llama::load_arch_hparams` `vocab.n_tokens()` returns 0. I re-created the GGUF file from safetensors to make sure that I am not accidentally using an outdated GGUF file and the issue persists.

It's because they used to read this from metadata, but this now goes nowhere:
https://github.com/ggml-org/llama.cpp/blob/00d56b11c3477b99bc18562dc1d1834f0d961778/src/llama-model.cpp#L1134-L1136

I suppose this should be set to hparams now.

Edit: Alternatively, just add that in all `load_arch_hparams` that need it.

PR #22005, opened 2026-04-16
cmake: use glob to collect src/models sources

ngxson:
## Overview

The goal is to make https://github.com/ggml-org/llama.cpp/pull/22004 a bit easier

# Requirements

- AI usage disclosure: no


ggerganov (approved):

CISC (approved):

pwilkin (approved):

ngladitz:
Just saw this by chance. CMake explicitly discourages GLOB for picking up sources files (see the Note in the file(GLOB) documentation). If you feel you must you might want to at least add `CONFIGURE_DEPENDS`.

Explicit source file listings work like manifests that make builds more reproducible.
With GLOB in general it can pick up e.g. leftover files / temporary copies and CMake won't implicitly know when files are added or removed (CONFIGURE_DEPENDS helps with that at least).

PR #9702, opened 2024-10-01
added implementation of DRY sampler (post-refactor)

wwoodsTM:
- [x] I have read the [contributing guidelines](https://github.com/ggerganov/llama.cpp/blob/master/CONTRIBUTING.md)
- Self-reported review complexity:
  - [ ] Low
  - [x] Medium
  - [ ] High

This is a continuation of the previous PR #6839 to implement DRY sampling (and I think a pretty faithful port of Koboldcpp’s impressive implementation, authored by @pi6am [who is also credited as co-author for this PR], which in turn is originally based on @p-e-w‘s design and ideas as explained [here](https://github.com/oobabooga/text-generation-webui/pull/5677)), but now overhauled to use the new sampler API. I added @l3utterfly as co-author in a recent commit of this PR.

Here are some important notes (**EDIT**: all have been addressed at this point, and in the case of the first point, this was moved to a separate PR, #10019):
- ~I may have also fixed #8971 (or it may turn out to be more of a bandaid), but by moving/forcing sampling to happen over the entire prompt at the conclusion of prompt processing in server.cpp, I was able to get consistent sampling of the prompt, regardless of caching or whether it is the first generation or not.~
- ~I also fixed a recent breaking change that was made, though I am not sure when this change happened. For quite a while it was possible (and indeed I have heavily relied on the feature in my own software for a long while now) to be able to submit a mixed array of strings and token IDs as a singleton prompt. This I believe was originally implemented from PR #2306. Apologies if this is considered out of scope here, and I can remove it or move it to a separate PR if need be, but to be honest this feature was the main reason why I converted to using llama.cpp's server. It also seems this fix restores the feature and doesn't stop the other prompting options from working either, but if I missed something here, please let me know.~
- ~I added a context size parameter to gpt_sampler_init(). If there is a smarter way to handle this, I am happy to do it in a less drastic way, but I couldn't immediately figure that out.~
- ~The classic issue that I have dealt with from day one with this where I can't figure out how to get the vector of strings for the sequence breakers from server.cpp/common/API land to src/llama-sampling land is still here. I again "cheated" with a C++ section at the bottom of llama.h and made a separate function for setting the sequence breakers apart from the regular init function. There is probably some clever way to avoid this or to use a void or extern that I am not as experienced with. I welcome any help or suggestions with that or any of these other issues.~
[...]

PR #22007, opened 2026-04-16
sampling: add segment-level repetition loop detection

Frank-Schruefer:
## Problem

Thinking models (and other models at low temperature) can enter infinite repetition loops where the same sentence or paragraph repeats endlessly. The existing `repeat_penalty` operates on token-level n-grams and cannot reliably break multi-token phrase repetition — the penalty window would need to cover the entire repeated phrase to be effective, which is impractical.

## Solution

Adds a new sampler `llama_sampler_init_repeat_line()` that tracks completed segments (delimited by configurable characters) and temporarily boosts the temperature when a repeated segment is detected, nudging the model out of the loop.

This is a port of [ollama/ollama#15212](https://github.com/ollama/ollama/pull/15212).

## New parameters

| Parameter | Type | Default | Description |
|---|---|---|---|
| `repeat_line_window` | int | 0 (disabled) | Number of past segments to track |
| `repeat_line_min_length` | int | 20 | Minimum segment length (avoids false positives from short phrases) |
| `repeat_line_delimiters` | string | `"\n.!?:"` | Characters that end a segment |
| `repeat_line_temp_boost` | float | 0.5 | Temperature boost when a loop is detected |

All parameters are exposed via the server API (`/v1/chat/completions` and `/completion`).

## Caveats

The feature works well in our Ollama deployment. The llama.cpp port is fresh — production testing in this context is still pending.

### DIFF
--- a/common/common.h
+++ b/common/common.h
@@ -241,6 +241,10 @@ struct common_params_sampling {
     float   dry_base           = 1.75f;  // 0.0 = disabled;      multiplier * base ^ (length of sequence before token - allowed length)
     int32_t dry_allowed_length = 2;      // tokens extending repetitions beyond this receive penalty
     int32_t dry_penalty_last_n = -1;     // how many tokens to scan for repetitions (0 = disable penalty, -1 = context size)
+    int32_t repeat_line_window     = 0;           // 0 = disabled; number of past segments to compare against
+    int32_t repeat_line_min_length = 20;          // ignore segments shorter than this (avoids false positives like "Ok.")
+    std::string repeat_line_delimiters = "\n.!?:"; // characters that end a segment
+    float   repeat_line_temp_boost = 0.50f;       // temperature boost when loop detected
     float   adaptive_target    = -1.0f;  // select tokens near this probability (valid range 0.0 to 1.0; negative = disabled)
     float   adaptive_decay     = 0.90f;  // EMA decay for adaptation; history ≈ 1/(1-decay) tokens (0.0 - 0.99)
     int32_t mirostat           = 0;      // 0 = disabled, 1 = mirostat, 2 = mirostat 2.0
--- a/common/sampling.cpp
+++ b/common/sampling.cpp
@@ -377,6 +377,12 @@ struct common_sampler * common_sampler_init(const struct llama_model * model, st
         llama_sampler_chain_add(chain, smpl);
     }
 
+    if (params.repeat_line_window > 0) {
+        llama_sampler_chain_add(chain,
+            llama_sampler_init_repeat_line(vocab, params.repeat_line_window, params.repeat_line_min_length,
+                params.repeat_line_delimiters.c_str(), params.repeat_line_temp_boost));
+    }
+
     if (grmr && params.backend_sampling) {
         LOG_WRN("%s: backend sampling is not compatible with grammar, disabling\n", __func__);
 
--- a/include/llama.h
+++ b/include/llama.h
@@ -1417,6 +1417,24 @@ extern "C" {
                           const char ** seq_breakers,
                               size_t    num_breakers);
 
+    /// repeat-line: segment-level loop detection.
+    ///
+    /// detects when the model repeats the same sentence/segment (which token-level repeat_penalty cannot catch).
+    /// when a repeated segment is detected, temperature is temporarily boosted.
+    ///
+    /// params:
+    ///   vocab       - vocabulary (for token decoding)
+    ///   window      - number of past segments to compare against (0 = disabled)
+    ///   min_length  - ignore segments shorter than this (default 10, avoids false positives like "Ok.")
+    ///   delimiters  - characters that end a segment (default "\n.!?:")
+    ///   temp_boost  - temperature boost applied when a loop is detected (default 0.3)
+    LLAMA_API struct llama_sampler * llama_sampler_init_repeat_line(
+            const struct llama_vocab * vocab,
+                             int32_t   window,
+                             int32_t   min_length,
+                          const char * delimiters,
+                               float   temp_boost);
+
     /// adaptive-p: select tokens near a configurable target probability over time.
     ///
     /// the adaptive-p sampler transforms the token probability distribution to favor tokens
--- a/src/llama-sampler.cpp
+++ b/src/llama-sampler.cpp
@@ -3423,6 +3423,143 @@ struct llama_sampler * llama_sampler_init_adaptive_p(
     );
 }
 
+// repeat-line
+
+struct llama_sampler_repeat_line {
+    const llama_vocab * vocab;
+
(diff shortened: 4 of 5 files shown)


Frank-Schruefer:
AI was used as an assistant throughout (code lookup, boilerplate, review). The design, parameter choices, and the underlying algorithm are based on our existing Ollama implementation which has been in production use. I reviewed all generated code; end-to-end testing of the loop detection itself is still pending.

ddh0:
> The existing `repeat_penalty` operates on token-level n-grams and cannot reliably break multi-token phrase repetition — the penalty window would need to cover the entire repeated phrase to be effective, which is impractical.

Why is it impractical? The existing penalty samplers (`repeat`, `presence`, and `frequency`) can already support arbitrary window sizes. The existing `adaptive-p` sampler maintains a rolling window of arbitrary size.

> Adds a new sampler `llama_sampler_init_repeat_line()` that tracks completed segments (delimited by configurable characters) and temporarily boosts the temperature when a repeated segment is detected, nudging the model out of the loop.

This sounds very similar to the existing DRY sampler (https://github.com/ggml-org/llama.cpp/pull/9702), which supports _sequence breakers_; that is, arbitrary strings that break up sequences of repetition.

IMO this new sampler seems redundant given that DRY exists and is already supported.

ddh0:
Please also be aware of the [AI Usage Policy](https://github.com/ggml-org/llama.cpp/blob/afe65aa28239906dfda63f70a52f60d3a7527fdc/CONTRIBUTING.md#ai-usage-policy), part of [CONTRIBUTING.md](https://raw.githubusercontent.com/ggml-org/llama.cpp/afe65aa28239906dfda63f70a52f60d3a7527fdc/CONTRIBUTING.md), which reads:

ngxson:
> Thinking models (and other models at low temperature) can enter infinite repetition loops where the same sentence or paragraph repeats endlessly. 

It sounds to me like the better fix is to increase the temperature. No models is trained for temperature 0.0 if I understand correctly.

And if the model still repeat itself with higher temp, that's the problem with the model itself. I don't think any LLM runtime out there will accept a hacky fix like this one.

Frank-Schruefer:
- repeat, presence, frequency is token-level and always-on
- reactive here vs. constant output manipulation there
- you want low temperature for thinking and coding

strawberrymelonpanda:
> if the model still repeat itself with higher temp, that's the problem with the model itself.

For what it's worth, I've got endless repetition loops in thinking with Gemma 4 26B (Q4) at T=**1.0** with all the latest fixes. It's far less often now, but it still happens.  Example from a couple days ago:

```
temperature = 1.0; top-p = 0.95; top-k = 64
```
```
Wait! I see it!\n
If line 168 prints the `AVG` header, then the output should be:\n
[...]
But the output I got was:\n
[...]
And the only way that can happen is if there is[...]\n
Wait! I see it!\n
[loop]
```

endlessly until I noticed and stopped it at over 28,000 repeating tokens.

> Why is it impractical? The existing penalty samplers (repeat, presence, and frequency) can already support arbitrary window sizes. The existing adaptive-p sampler maintains a rolling window of arbitrary size.

>This sounds very similar to the existing DRY sampler (#9702), which supports sequence breakers;

If there are existing settings that might work well with Gemma 4 without seriously hampering its reasoning ability, I'd sincerely like to know how to use them. Could you give an example?  

My own attempts at changing the DRY and repetition penalty either didn't stop the repetition, or seriously destroyed the output (e.g. it could no longer tool call because it would be unable to correctly call the current path)

I really like the idea of detecting a very obvious (to the user) repeating pattern and nudging the model out of that loop, but otherwise _not_ affecting the model's output.

ddh0:
> If there are existing settings that might work well with Gemma 4 without seriously hampering its reasoning ability, I'd sincerely like to know how to use them. Could you give an example?

These options work well for me:

Screenshot 2026-04-17 at 12 25 50 AM

e.g. `--reasoning-budget 4096` to limit CoT to 4096 tokens, and `--reasoning-budget-message "...enough. Need to give the final output now."` to end CoT gracefully.

strawberrymelonpanda:
> e.g. `--reasoning-budget 4096` to limit CoT to 4096 tokens, and `--reasoning-budget-message "...enough. Need to give the final output now."` to end CoT gracefully.

Ah, thanks.  I'm currently using the reasoning budget as well.  I use

```
reasoning-budget = 20480
reasoning-budget-message = "... reasoning budget exceeded. We may be in a loop or overcomplicating things. Let's try to answer now."
```

It does a similar job, but as I mentioned in the issue thread, my real problem is the loop itself rather than the verbosity.  It also kicks the model out of reasoning entirely. It would be nice if there's a way to target the repetition, and nudge the model out of that loop, rather than hard-limit the reasoning amount.

pwilkin:
@strawberrymelonpanda as @ddh0 already mentioned here, there's the oft-forgotten DRY sampler :)

pwilkin:
I mean, unless I'm misreading something, "DRY" literally stands for "Do not Repeat Yourself".

strawberrymelonpanda:
Sure, and I've tried it, but the result I got wasn't good. In my experience, It absolutely destroys Gemma 4's ability to successfully do agentic tasks with tool calls just setting --dry-multiplier to 0.05.  

```
✗ edit failed
Error: Could not find oldString in the file. It must match exactly, including whitespace, indentation, and line endings.
✗ edit failed
Error: Could not find oldString in the file. It must match exactly, including whitespace, indentation, and line endings.
✗ edit failed
Error: Could not find oldString in the file. It must match exactly, including whitespace, indentation, and line endings.
✗ edit failed
Error: Could not find oldString in the file. It must match exactly, including whitespace, indentation, and line endings.
```

I don't claim to understand the internals of it, but what counts as "repeating yourself", to an LLM?  Code and paths that *should* match exactly seem to purposefully not.

But like I said, if someone knows better settings, I'm very interested.  There's at least 5 knobs to tweak here.

```
--dry-multiplier N                      set DRY sampling multiplier (default: 0.00, 0.0 = disabled)
--dry-base N                            set DRY sampling base value (default: 1.75)
--dry-allowed-length N                  set allowed length for DRY sampling (default: 2)
--dry-penalty-last-n N                  set DRY penalty for the last n tokens (default: -1, 0 = disable, -1 =
--dry-sequence-breaker STRING           add sequence breaker for DRY sampling, clearing out default breakers
```

Atomic-Germ:
> ```
> I don't claim to understand the internals of it, but what counts as "repeating yourself", to an LLM? Code and paths that _should_ match exactly seem to purposefully not.
> ```
When a model just outputs literally the same token over and over and over and over again. The problem is that there might be situations when that's actually the right behavior; writing a matrix with a whole lot of 0's in it for instance.

PR #22009, opened 2026-04-16
ggml-vulkan/CMakeLists: add a check for SPIRV-Headers

jeeb:
## Overview

This makes the build fail at configure time instead of build time in case any of the sysroots included does not contain SPIRV-Headers. Generally it is preferred to fail as quickly as possible if a required dependency is not available.

Files related to this package are installed as part of the SPIRV-Headers project (both cmake files as well as a pkg-config file).
```
-- Installing: PREFIX/share/cmake/SPIRV-Headers/SPIRV-HeadersConfigVersion.cmake
-- Installing: PREFIX/share/cmake/SPIRV-Headers/SPIRV-HeadersConfig.cmake
-- Installing: PREFIX/share/pkgconfig/SPIRV-Headers.pc
```

I have verified this with cross-compilation from Linux to Windows as that is what I'm mostly poking at right now. I expect this to also work on Linux as at least Fedora ([spirv-headers-devel](https://koji.fedoraproject.org/koji/rpminfo?fileStart=50&rpmID=46244889&fileOrder=name&buildrootOrder=-id&buildrootStart=0#filelist)) and Ubuntu ([spirv-headers](https://packages.ubuntu.com/noble-updates/all/spirv-headers/filelist)) package these files. I tried to check whether these files are contained within the LunarG Windows Vulkan SDK package, but at least `vulkansdk-windows-X64-1.4.341.0.exe` and `vulkansdk-windows-X64-1.4.341.1.exe` only seem to contain the `Bin` directory inside and no clear extractable archives :/ .

If this check fails with the Windows Vulkan SDK (this should be check'able by the CI), then a compiler check with the exact header inclusion logic as in `ggml-vulkan.cpp` can be done instead.

# Requirements

- AI usage disclosure: AI was **NOT** used in the context of this PR

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -117,6 +117,12 @@ if [ ! -z ${GG_BUILD_VULKAN} ]; then
     # if on Mac, disable METAL
     if [[ "$OSTYPE" == "darwin"* ]]; then
         CMAKE_EXTRA="${CMAKE_EXTRA} -DGGML_METAL=OFF -DGGML_BLAS=OFF"
+
+        MACOS_RUNNER_CUSTOM_VULKAN_CMAKE_LOCATION="/usr/local/lib/cmake/vulkan"
+        MACOS_RUNNER_CUSTOM_SPIRV_HEADERS_LOCATION="${MACOS_RUNNER_CUSTOM_VULKAN_CMAKE_LOCATION}/SPIRV-Headers/SPIRV-HeadersConfig.cmake"
+        if [[ -f "${MACOS_RUNNER_CUSTOM_SPIRV_HEADERS_LOCATION}" || -h "${MACOS_RUNNER_CUSTOM_SPIRV_HEADERS_LOCATION}" ]]; then
+            CMAKE_EXTRA="${CMAKE_EXTRA} -DSPIRV-Headers_DIR=${MACOS_RUNNER_CUSTOM_VULKAN_CMAKE_LOCATION}/SPIRV-Headers"
+        fi
     fi
 
     # Build shared libs on Windows
--- a/ggml/src/ggml-vulkan/CMakeLists.txt
+++ b/ggml/src/ggml-vulkan/CMakeLists.txt
@@ -8,6 +8,8 @@ endif()
 
 find_package(Vulkan COMPONENTS glslc REQUIRED)
 
+find_package(SPIRV-Headers REQUIRED)
+
 if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     # Parallel build object files
     add_definitions(/MP)


jeeb:
My main desktop has nicely started showing signs of degradation right after I made this PR, but I am still alive, and the main reason for this change is still around: Fail during configure step and not after XY files (hundreds in the best case) have already been compiled.

This works for both Linux (tested on Fedora and as far as I can tell Ubuntu has a very similar file listing for its related package) and cross-compilation to Windows from Linux (as long as SPIRV-Headers is installed via their cmake build system - it installs the files mentioned).

The only questions would be the Windows Vulkan SDK and NDK (if thought as a valid Vulkan target). I expect MoltenVK to not be a target for llama.cpp, as native Metal back-end seems to exist. It would be nice if the CI flow could be enabled, so these could be checked. I saw under `docs` that manual copying of files was recommended for some cases, so that may have to be adjusted so that such a check would pass. Alternatively, the check gets changed to be similar to the one in the actual file.

jeeb:
Ah, so MoltenVK indeed was a target - as that is the first thing to fail. Will check how that sysroot is handled.

**edit**: seems like the dependencies are not installed by the CI flow?
```yaml
  ggml-ci-mac-vulkan:
    runs-on: [self-hosted, macOS, ARM64]

    steps:
      - name: Clone
        id: checkout
        uses: actions/checkout@v6

      - name: Test
        id: ggml-ci
        run: |
          vulkaninfo --summary
          GG_BUILD_VULKAN=1 bash ./ci/run.sh ~/results/llama.cpp ~/mnt/llama.cpp
```

Will have to check how and what the MoltenVK SDK (if that is what is being installed onto the system) contains.

Funnily enough the configure part of Windows + Vulkan CI flow seems to pass JustFine (https://github.com/ggml-org/llama.cpp/actions/runs/25050050776/job/73380477706?pr=22009#step:8:70)

**edit2**: Checked MoltenVK's SDK contents (https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.4.1) and neither `MoltenVK-macos.tar` or `MoltenVK-all.tar` contains SPIRV-Headers it seems...

So I guess SPIRV-Headers was installed manually on this runner?

jeeb:
Alright, grabbed my old Intel Macbook and grabbed the latest Vulkan SDK from LunarG (trying to look at MoltenVK SDK was a mistake - got things confused)

```bash
find VulkanSDK/1.4.341.1/macOS -iname 'spirv-headers*'
# VulkanSDK/1.4.341.1/macOS/lib/pkgconfig/SPIRV-Headers.pc
# VulkanSDK/1.4.341.1/macOS/lib/cmake/SPIRV-Headers
# VulkanSDK/1.4.341.1/macOS/lib/cmake/SPIRV-Headers/SPIRV-HeadersConfig.cmake
# VulkanSDK/1.4.341.1/macOS/lib/cmake/SPIRV-Headers/SPIRV-HeadersConfigVersion.cmake
```

So the necessary files are included.

Then tested building this PR's branch and found out that it works as long as cmake is being told where to look (`CMAKE_FIND_ROOT_PATH=~/VulkanSDK/1.4.341.1/macOS`):
```
-- Found Vulkan: /Users/jeeb/VulkanSDK/1.4.341.1/macOS/lib/libvulkan.dylib (found version "1.4.341") found components: glslc glslangValidator 
-- Vulkan found
```
```
./CMakeCache.txt://The directory containing a CMake configuration file for SPIRV-Headers.
./CMakeCache.txt:SPIRV-Headers_DIR:PATH=/Users/jeeb/VulkanSDK/1.4.341.1/macOS/lib/cmake/SPIRV-Headers
```

The result compiles fine.

So I guess I can spy where Vulkan is found on the previous macOS Vulkan build runs, and see if I can make cmake search for things in its sysroot.

**edit**:
```
-- Found Vulkan: /usr/local/lib/libvulkan.dylib (found version "1.4.341") found components: glslc glslangValidator
```
So if this is properly packaged, `-DCMAKE_FIND_ROOT_PATH=/usr/local` should do it.

jeeb:
Two alternative takes on this would be:

1. `pkg_check_modules`, since SPIRV-Headers does seem to have a `pc` file. This would mean that `PKG_CONFIG_{PATH,LIBDIR}` would work, although upstream seems to install the `pc` file under `share/` (probably since there is no library in the project). Alas, I've not seen too many people doing cross-compiles etc setting overrides towards it, so while I do personally prefer pkg-config as a method of finding libraries, in this case the cmake module seems like it would work more generally (as the search path set is the sysroot itself), plus `pkgconf` is usually not around for Windows MSVC users.
2. `check_include_file`, trying to just coast on the Vulkan cmake check to set the necessary search paths for the compiler.

0cc4m:
When you use the SDK, you're supposed to source its setup-env file. If you've done that, cmake should find it with no manual configuration.

jeeb:
> When you use the SDK, you're supposed to source its setup-env file. If you've done that, cmake should find it with no manual configuration.

Thanks for the information. I was just in general wanting to verify whether building with what I *expect* is going to be utilized by people would work. Do you happen to know whether this setup-env script is there on the macOS runner utilized for the Vulkan build?

If yes, we can source that in the CI script (~although I think technically setting the search path would allow for other things to be found from that sysroot as well~ never mind, I mean I expect that script to just set the search variable so in that sense the result should be the same). If not, then what I added more or less should work. Although given that the CI is not allowed to run, I cannot validate whether the sysroot on the runner contains these cmake files (I would expect them to be there, but still).

jeeb:
Thanks for enabling the CI, whomever it was :)

1. The webgpu fail is unrelated (`add-apt-repository -y ppa:kisak/kisak-mesa` timing out)
2. the macOS fail means that the cmake files for SPIRV-Headers are not there under `/usr/local`... Yet they should be installed in the sysroot by either the LunarG Vulkan SDK, or if you are packaging something from source you would get them as part of the install process, Even if Vulkan development files would be installed by homebrew, I would be really surprised if they would skip these files...

Given that any changes I do would require another OK for the CI to run, can someone execute a `find /usr/local -iname 'spirv-headers*'` and `find /usr/local -iname 'setup-env'` on the macOS vulkan runner?

**edit**: at least homebrew's [formula](https://github.com/Homebrew/homebrew-core/blob/61a6526db6f4f57f1902d487df751a53b593dddc/Formula/s/spirv-headers.rb) for SPIRV-Headers seems to be using the cmake bits

**edit2**: also if those files are found on the runner, if you could check what paths they point at would be really nice. Just to verify that there are no path issues.

jeeb:
I will attempt to then do the parts that I would usually do: Add the debug poking into a debug commit, and if someone then enables the CI I should get more information regarding how the custom macOS runner is set up.

jeeb:
Alright, this should now hopefully give us some additional details regarding the state of the custom runner's `/usr/local`

jeeb:
Thanks for whomever allowed the iteration to go on
```
spirv-headers file listing:
/usr/local/lib/pkgconfig/SPIRV-Headers.pc
/usr/local/lib/cmake/vulkan/SPIRV-Headers
/usr/local/lib/cmake/vulkan/SPIRV-Headers/SPIRV-HeadersConfig.cmake
/usr/local/lib/cmake/vulkan/SPIRV-Headers/SPIRV-HeadersConfigVersion.cmake
-rw-r--r--@ 1 root  wheel   4.1K Sep 22  2025 /usr/local/lib/cmake/vulkan/SPIRV-Headers/SPIRV-HeadersConfig.cmake
```

So the cmake file is there (`SPIRV-HeadersConfig.cmake`), good.

The [mentioned](https://vulkan.lunarg.com/doc/sdk/1.4.341.0/mac/getting_started.html) shell script is not there, which I would guesstimate is because this stuff was installed through `brew` ?
```
DEBUG setup-env.sh check:
setup-env file listing:

```

**edit**: Failure of [CI (self-hosted) / ggml-ci-nvidia-vulkan-cm2 (pull_request)](https://github.com/ggml-org/llama.cpp/actions/runs/25339747089/job/74974358607?pr=22009) is unrelated, build completes fine and it fails during tests.

**edit2**: Alright, I see the problem. `/vulkan/` in the middle.

- `lib/cmake/SPIRV-Headers` is what's in the macOS Vulkan SDK, and works. :heavy_check_mark: 
- `share/cmake/SPIRV-Headers/SPIRV-HeadersConfig.cmake` is how things get installed from SPIRV-Headers' cmake on Linux. :heavy_check_mark: 
- `lib/cmake/vulkan/SPIRV-Headers/SPIRV-HeadersConfig.cmake` has an extra directory and does not work :negative_squared_cross_mark:

Am not sure if setting the overall search path to `lib/vulkan` will work, so I guess I will do `-DSPIRV-Headers_DIR=/usr/local/lib/cmake/vulkan/SPIRV-Headers`. Just means that other things might not get found later automagically if additional requirements are added to the llama.cpp cmake checks.

jeeb:
There, added the new override and made it conditional on that file actually being there. This hopefully should make it so that if the macOS runner is ever recreated, unless the file somehow ends up in that location again - it should not get that override set and thus not break the CI at that point (as it should be following the standard cmake search paths at that point).

jeeb:
Yay, [ggml-ci-mac-vulkan](https://github.com/ggml-org/llama.cpp/actions/runs/25548850535/job/74996126139?pr=22009#step:3:257) now builds properly as expected :) .
```
...
-- Adding CPU backend variant ggml-cpu: -U__ARM_FEATURE_SVE;-mcpu=native+dotprod+i8mm+nosve+sme 
-- Found Vulkan: /usr/local/lib/libvulkan.dylib (found version "1.4.341") found components: glslc glslangValidator
-- Vulkan found
-- GL_KHR_cooperative_matrix supported by glslc
-- GL_NV_cooperative_matrix2 supported by glslc
-- GL_EXT_integer_dot_product supported by glslc
-- GL_EXT_bfloat16 supported by glslc
-- Including Vulkan backend
...
```

jeeb:
[CI / ubuntu-24-webgpu (pull_request)](https://github.com/ggml-org/llama.cpp/actions/runs/25548850622/job/74996157668?pr=22009) failure seems to be unrelated, it succeeds at building and while it does seem to have Vulkan SDK included, Vulkan as a feature is not enabled. It fails during tests.

jeeb:
Alright, everything passes build-wise in CI (with a single unrelated webgpu test failure).

0cc4m (approved):
LGTM, thank you

0cc4m:

CISC (approved):

En3Tho:
@0cc4m This broke build on Windows.  Any tip on how to fix this?
```
-- Performing Test HAS_AVX_1
-- Performing Test HAS_AVX_1 - Success
-- Performing Test HAS_AVX2_1
-- Performing Test HAS_AVX2_1 - Success
-- Performing Test HAS_FMA_1
-- Performing Test HAS_FMA_1 - Success
-- Performing Test HAS_AVX512_1
-- Performing Test HAS_AVX512_1 - Failed
-- Performing Test HAS_AVX512_2
-- Performing Test HAS_AVX512_2 - Failed
-- Adding CPU backend variant ggml-cpu: /arch:AVX2 GGML_AVX2;GGML_FMA;GGML_F16C
-- Found Vulkan: C:/VulkanSDK/1.4.341.1/Lib/vulkan-1.lib (found version "1.4.341") found components: glslc glslangValidator
CMake Error at ggml/src/ggml-vulkan/CMakeLists.txt:11 (find_package):
  By not providing "FindSPIRV-Headers.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "SPIRV-Headers", but CMake did not find one.

  Could not find a package configuration file provided by "SPIRV-Headers"
  with any of the following names:

    SPIRV-HeadersConfig.cmake
    spirv-headers-config.cmake

  Add the installation prefix of "SPIRV-Headers" to CMAKE_PREFIX_PATH or set
  "SPIRV-Headers_DIR" to a directory containing one of the above files.  If
  "SPIRV-Headers" provides a separate development package or SDK, be sure it
  has been installed.


-- Configuring incomplete, errors occurred!
```

0cc4m:
It did not break it in general, the Windows CI was fine. Please open an issue and provide details on your setup.

jeffbolznv:
IIRC the windows CI build uses msys2 or similar. I'll try to reproduce this, too.

jeffbolznv:
I got the same error on Windows. Claude suggested this, which works for me:

```
if(WIN32 AND DEFINED ENV{VULKAN_SDK})
    list(APPEND CMAKE_PREFIX_PATH "$ENV{VULKAN_SDK}")
endif()
find_package(SPIRV-Headers CONFIG REQUIRED)
```

ZUIcat:
Ah, this broke my Windows build. I fixed it by specifying `set "CMAKE_PREFIX_PATH=%VULKAN_SDK%;%CMAKE_PREFIX_PATH%"` before calling cmake.

PR #7420, opened 2024-05-20
Direct I/O and Transparent HugePages

pavelfatin:
**Abstract:** `--direct-io` for bypassing page cache (and using THP on Linux): up to 3-6x faster uncached loading, fewer pageouts, no page cache pollution.

Using LLMs involves large data files: between Llama 3 70B and Mixtral 8x22B, loading 50-100 GB models is now quite common. This requires not just fast CPUs/GPUs and ample RAM, but also fast storage devices.

Fortunately, consumer PCIe 4.0 NVMe SSDs offer read speeds up to 7.5 GB/s, and PCIe 5.0 SSDs&mdash;up to 14.5 GB/s&mdash;no fancy RAID arrays required! So loading an 80 GB model should take just 5 seconds&mdash;problem solved. Right? Not so fast. If you look at `iotop`, the numbers are different from those in CrystalBenchMark screenshots. There are several reasons for that.

Llama.cpp is not a typical application because it:
1. reads very large files,
2. reads the files completely,
3. reads the files sequentially,
4. reads the files in one step,
5. doesn't transform the read data,
6. both reads data and allocates memory for that data,
7. keeps all the read data in memory,
8. often takes almost all available memory.

Since operating systems don't normally optimize for this use case, there are two bottlenecks:
1. buffered I/O,
2. virtual memory.

## Buffered I/O

Regular I/O is buffered and goes through the page cache. This works well for typical workloads&mdash;small random repeated reads. But in this case, it's not ideal because:
* the data has to be copied between kernel space and user space, which takes extra time,
* the caching algorithms introduce overhead,
* the memory might be allocated twice,
* it causes page cache pollution, and the OS may reclaim pages from elsewhere,
* unless there's 2x amount of memory, memory also has to be reclaimed,
* it's only possible to benefit from caching if there's 2x amount of memory,
* and even then, reading the cached data requires copying.

## Memory-mapped I/O

One way to address these shortcomings is to use memory-mapped I/O. The advantages are:
* there's no double copying,
* there's no double memory allocation,
* data can be cached between consecutive program runs.

However, memory-mapped I/O is also optimized primarily for small random repeated reads, and also have some drawbacks:
* sequential reading of large files may not necessarily be fast (even with prefetching, which is not always available),
* there's still the general caching subsystem overhead,
* memory is more likely to be paged out&mdash;this doesn't require writing to disk; to the OS it's just "file cache", while the default swappiness on Linux is 60 (this can be addressed by memory locking, but only partially).
[...]

Issue #2251, opened 2023-07-17
[User] Huge Page Support


Micraow:
Hello, I am using llama.cpp on linux, due to my small memory, my program can often take up 95% of my memory, and I use CPU to run the model, but there's a problem, the `kswapd0` takes about 30% of my CPU. I learned that this process is to change memory page, but it takes too much of my CPU. I wonder if llama.cpp can support huge page and solve this problem.

Thank you

ggerganov:
I'm unfamiliar with the concept of huge pages. If is something that can be achieved with small amount of changes, then we can support it

staviq:
Hugepages is a feature of the Linux kernel, I am not sure if you can programmatically force to use hugepages specifically, but I found this project/library, which can tell you if the memory you are using is/isn't using hugepages, so it might be a good starting point:

https://github.com/travisdowns/page-info

slaren:
I think this would only require adding the `MAP_HUGETLB` flag to `mmap`.

qdacsvx:
MAP_HUGETLB isn't supported for file-backed mmaps. But (on Linux) you can use madvise with MADV_HUGEPAGE.

Chris611:
Any updates on this? For me it would be more for convenience as I've already have most of my memory allocated as hugepages for use with Virtual Machines. And it might be slightly faster

jboero:
I know this is closed but Just a note since I've experimented a bit and tried out the following in every combination while inferencing purely in CPU on 11th gen x86_64 (Tiger Lake) with AVX512 enabled:

1. MADV_HUGEPAGE with transparent hugepages enabled. 
2. Pre-allocated and run hugepages manually.
3. Tried both with using `--mlock` which loads the model into RAM.

In my experience it makes no noticeable performance difference sadly. Since x86 still uses a 4KB pagesize natively I don't think there is much gain to be had with this optimization. Apple Arm currently uses 16KB page size natively which definitely does help when running both Mac OS and Linux (Asahi Fedora). I've not yet tried CPU inference on a PowerPC/Power box which may have benefits with 64K pages.

Obviously all of these are still impractically slow compared to a model properly fit inside the HBM memory of a dedicated GPU.

qdacsvx:
[...]

PR #22022, opened 2026-04-16
llama-mmap: add MADV_HUGEPAGE hint for THP on Linux

Marxist-Leninist:
Calls madvise(MADV_HUGEPAGE) on the read-only mmap region used for model weights on Linux.

When THP is in 'madvise' mode (the default on many desktop distros), this opts the mapping
into transparent huge page promotion. For a 4-5 GB model weight map the page count drops from
~1M 4 KB pages to ~2K 2 MB pages, which reduces TLB pressure and the number of minor faults
on the hot inference path.

No configuration required. The call is advisory: if THP is disabled
(transparent_hugepage/enabled == never) or pages cannot be promoted, madvise returns EINVAL
and a debug-level log message is emitted. NUMA-aware allocations are skipped (existing !numa
guard). Guarded by defined(MADV_HUGEPAGE) for portability.

Tested on Linux x86_64 with THP=madvise. Neutral on an unloaded machine (pages stay resident);
reduces re-fault latency spikes under memory pressure.

Complements #21821 which adds HugeTLB support; that approach requires pre-allocated pages in
the HugeTLB pool, while THP requires no system configuration.

### DIFF
--- a/src/llama-mmap.cpp
+++ b/src/llama-mmap.cpp
@@ -451,6 +451,20 @@ struct llama_mmap::impl {
             throw std::runtime_error(format("mmap failed: %s", strerror(errno)));
         }
 
+#ifdef __linux__
+        // Hint the kernel to back this region with 2MB huge pages where possible.
+        // For a 1 GB model weights map this can drop the number of pages from ~262K
+        // 4KB pages to ~512 2MB pages, reducing TLB pressure and (critically)
+        // reducing the number of re-faults when pages get evicted under memory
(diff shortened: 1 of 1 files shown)


doctorjei:
I want to reinforce that these are complementary and not redundant; THP and pre-allocated huge pages can serve different goals / solve different problems.

Also related to https://github.com/ggml-org/llama.cpp/issues/2251, https://github.com/ggml-org/llama.cpp/issues/12444, and https://github.com/ggml-org/llama.cpp/pull/7420.

PR #22028, opened 2026-04-17
server: Expose `media_marker` on /props endpoint.

cetarthoriphros:
## Overview

Expose the `media_marker` on the /props endpoint.

## Additional information

For clients using the /completion endpoint, it's useful to have a way to fetch the current media_marker for inline image embedding. Right now, as far as I know, it can only be extracted by sending a dummy request to the /apply-template endpoint and parsing it from the result, which is overly complicated, or launching llama-server with LLAMA_MEDIA_MARKER defined to the marker, which may be inaccessible to clients not sharing the same environment.

# Requirements

- AI usage disclosure: NO.


ngxson (approved):

ngxson:
CC @ggml-org/maintainers if someone can give an approval

aldehir (approved):

PR #22038, opened 2026-04-17
server: Skip API key verification for static files

roj234:
## Overview

Currently only `/index.html` `/bundle.css` and `/bundle.js` can skip API key validation 
When using `--path` parameter with API key, most static files will return 401
This PR change hard-coded paths to `path contains dot` which generally implies a static file, and RESTful api will not contain dot

## Additional information

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/tools/server/server-http.cpp
+++ b/tools/server/server-http.cpp
@@ -145,13 +145,15 @@ bool server_http_context::init(const common_params & params) {
             "/v1/models",
             "/api/tags",
             "/",
-            "/index.html",
-            "/bundle.js",
-            "/bundle.css",
         };
 
         // If API key is not set, skip validation
         if (api_keys.empty()) {
+            return true;
+        }
+
+        // If path contains dot ( index.html, abc.css, abc.js ... )
+        if (req.path.find(".") != std::string::npos && req.path.rfind("/v1/", 0) == std::string::npos) {
             return true;
         }
 


tools/server/server-http.cpp
```diff
@@ -145,16 +145,18 @@ bool server_http_context::init(const common_params & params) {
...
         // If API key is not set, skip validation
         if (api_keys.empty()) {
             return true;
         }
 
+        // If path contains dot ( index.html, abc.css, abc.js ... )
+        if (req.path.find(".") != std::string::npos && req.path.rfind("/v1/", 0) == std::string::npos) {
+            return true;
+        }
+
```
allozaur, line 159: That looks quite unsafe tbh... @ngxson wdyt?

ngxson:
I don't think this feature is generally valid. One can argue that they want their static files to be protected by the same api key.

Instead, any custom auth logic should be handled via a reverse proxy

roj234:
> I don't think this feature is generally valid. One can argue that they want their static files to be protected by the same api key.
> 
> Instead, any custom auth logic should be handled via a reverse proxy

An issue a few days later: "Why index.html is not protected by api key" 😂

ngxson:
a better fix would be to require api key for everything, except the default webui (i.e. no `--path` is specified)

the default webui's code is public, there is no need to protect it

PR #22041, opened 2026-04-17
Reduce CPU overhead in meta backend: cache subgraph splits when cgraph is unchanged

gaugarg-nv:
Skip per-call subgraph construction in `ggml_backend_meta_graph_compute` when the same `ggml_cgraph` is used consecutively.
Assign `uid` to every sub-graph so that CUDA's fast uid check path hits too.

Performance on 2x RTX 5090:

model | test | t/s - 5e6c0e18b | t/s - PR | Speed-up
-- | -- | -- | -- | --
qwen35moe 35B.A3B Q8_0 | pp512 | 6701 | 6825.95 | 1.02
qwen35moe 35B.A3B Q8_0 | tg128 | 179.46 | 207.74 | 1.16
gemma4 26B Q8_0 | pp512 | 3526.61 | 3441.77 | 0.98
gemma4 26B Q8_0 | tg128 | 154.01 | 169.16 | 1.10
gpt-oss 20B F16 | pp512 | 11425.35 | 11498.14 | 1.01
gpt-oss 20B F16 | tg128 | 319.33 | 345.3 | 1.08

Nsight profile Master:
image

Nsight profile PR:

image

# Requirements

- AI usage disclosure: Yes, for code review.

### DIFF
--- a/ggml/src/ggml-backend-meta.cpp
+++ b/ggml/src/ggml-backend-meta.cpp
@@ -1419,6 +1419,11 @@ struct ggml_backend_meta_context {
     size_t                      max_tmp_size  = 0;
     size_t                      max_subgraphs = 0;
 
+    // cached values from previous run
+    uint64_t last_uid          = 0;
+    size_t   last_n_subgraphs  = 0;
+    size_t   last_max_tmp_size = 0;
+
     void *                               comm_ctx       = nullptr;
     ggml_backend_comm_allreduce_tensor_t comm_allreduce = nullptr;
 
@@ -1578,6 +1583,9 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend, 
     const size_t n_backends = ggml_backend_meta_n_backends(backend);
     ggml_backend_meta_context * backend_ctx = (ggml_backend_meta_context *) backend->context;
 
+    // rebuild if the cgraph uid differs from the last one we built for
+    bool needs_rebuild = (cgraph->uid == 0) || (cgraph->uid != backend_ctx->last_uid);
+
     bool max_nnodes_raised = false;
     if (cgraph->n_nodes > backend_ctx->max_nnodes) {
         for (size_t j = 0; j < n_backends; j++) {
@@ -1587,26 +1595,30 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend, 
         }
         backend_ctx->max_nnodes = cgraph->n_nodes;
         max_nnodes_raised = true;
-    }
-    for (size_t j = 0; j < n_backends; j++) {
-        auto & bcj = backend_ctx->backend_configs[j];
-
-        for (int i = 0; i < cgraph->n_nodes; i++) {
-            ggml_tensor * node = cgraph->nodes[i];
-            if (node->view_src != nullptr && node->view_src->op == GGML_OP_NONE && ggml_backend_buffer_is_host(node->view_src->buffer)) {
-                // FIXME s_copy_main is on the CPU and its view seems to be incorrectly added to the graph nodes.
-                // For regular usage this doesn't matter since it's a noop but trying to call ggml_backend_meta_buffer_simple_tensor results in a crash.
-                bcj.nodes[i] = node;
-                continue;
-            }
-            bcj.nodes[i] = ggml_backend_meta_buffer_simple_tensor(node, j);
-            GGML_ASSERT(bcj.nodes[i]);
-        }
+        needs_rebuild = true;
     }
 
     size_t n_subgraphs  = 0;
     size_t max_tmp_size = 0;
-    {
+
+    if (needs_rebuild) {
+        for (size_t j = 0; j < n_backends; j++) {
+            auto & bcj = backend_ctx->backend_configs[j];
+
+            for (int i = 0; i < cgraph->n_nodes; i++) {
+                ggml_tensor * node = cgraph->nodes[i];
+                if (node->view_src != nullptr && node->view_src->op == GGML_OP_NONE && ggml_backend_buffer_is_host(node->view_src->buffer)) {
+                    // FIXME s_copy_main is on the CPU and its view seems to be incorrectly added to the graph nodes.
+                    // For regular usage this doesn't matter since it's a noop but trying to call ggml_backend_meta_buffer_simple_tensor results in a crash.
+                    bcj.nodes[i] = node;
+                    continue;
+                }
+                bcj.nodes[i] = ggml_backend_meta_buffer_simple_tensor(node, j);
+                GGML_ASSERT(bcj.nodes[i]);
+            }
+        }
+
+        {
         // For MoE models it may make sense to delay the AllReduce in order to reduce I/O:
         auto get_i_delayed = [&](const int i) -> int {
             int id = i; // i_delayed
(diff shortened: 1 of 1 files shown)


ggml/src/ggml-backend-meta.cpp
```diff
@@ -1587,26 +1595,30 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
         }
         backend_ctx->max_nnodes = cgraph->n_nodes;
         max_nnodes_raised = true;
+        needs_rebuild = true;
```
JohannesGaessler:
```suggestion
        assert(needs_rebuild);
```

This is not logically incorrect but I think it makes more sense to have an assert here since an increase in the number of nodes should always result in a changed UID.

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1699,6 +1711,14 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
             i_start = i + 1;
         }
         GGML_ASSERT(i_start == cgraph->n_nodes);
```
JohannesGaessler: Please fix the indentation.

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1709,7 +1729,7 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
         backend_ctx->max_tmp_size = max_tmp_size;
     }
 
-
+    bool context_reallocated = false;
```
JohannesGaessler: I don't think it's possible to create a scenario with `!needs_rebuild && context_reallocated` so I would say to remove this variable.

ggml/src/ggml-backend-meta.cpp
gaugarg-nv: Makes sense. Updated.

ggml/src/ggml-backend-meta.cpp
gaugarg-nv: I thought the same, but kept it for safety. I have now removed this and added an assert instead.

gaugarg-nv:
Thanks for the review @JohannesGaessler . I have added a couple of `assert` now and have made sure those asserts are not getting hit with a debug build.

ggml/src/ggml-backend-meta.cpp
JohannesGaessler: I didn't ask you to remove the scope.

gaugarg-nv: Reverted to earlier and fixed the indentation.

gaugarg-nv:
@JohannesGaessler, please let me know if there is any other feedback.

@am17an @ggerganov, could you please help review this change when you have cycles? This is a simple change, so it shouldn't take long.

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1419,6 +1419,11 @@ struct ggml_backend_meta_context {
     size_t                      max_tmp_size  = 0;
     size_t                      max_subgraphs = 0;
 
+    // cached values from previous run
+    uint64_t last_uid          = 0;
+    size_t   last_n_subgraphs  = 0;
+    size_t   last_max_tmp_size = 0;
+
```
am17an:
```suggestion
    uint64_t uid          = 0;
    size_t   n_subgraphs  = 0;
    size_t   max_tmp_size = 0;

```

Maybe more consistent with the cuda backend

am17an (approved):

ggml/src/ggml-backend-meta.cpp
gaugarg-nv: Fixed now. I have refactored the code a bit to get rid of `last_max_tmp_size` field.

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1587,175 +1592,187 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
...
-        for (int i = 0; i < cgraph->n_nodes; i++) {
-            ggml_tensor * node = cgraph->nodes[i];
-            if (node->view_src != nullptr && node->view_src->op == GGML_OP_NONE && ggml_backend_buffer_is_host(node->view_src->buffer)) {
-                // FIXME s_copy_main is on the CPU and its view seems to be incorrectly added to the graph nodes.
-                // For regular usage this doesn't matter since it's a noop but trying to call ggml_backend_meta_buffer_simple_tensor results in a crash.
-                bcj.nodes[i] = node;
-                continue;
+
+    size_t n_subgraphs = 0;
+
```
ggerganov: nit: I would move this local variable inside the `needs_rebuild` scope. Then after assigning it to `backend_ctx->n_subgraphs` on line 1715, the rest of the `ggml_backend_meta_graph_compute()` would reference `backend_ctx->n_subgraphs` (instead of the local variable). The `else` on line 1774 would not be needed anymore. Think this would slightly improve the flow of this function.

But keeping it as it is is also fine.

ggerganov (approved):

ggml/src/ggml-backend-meta.cpp
gaugarg-nv: Addressed now.

am17an:
Viewing this PR using `?w=1` is much better. https://github.com/ggml-org/llama.cpp/pull/22041/changes?w=1

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1578,6 +1580,9 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
     const size_t n_backends = ggml_backend_meta_n_backends(backend);
     ggml_backend_meta_context * backend_ctx = (ggml_backend_meta_context *) backend->context;
 
+    // rebuild if the cgraph uid differs from the last one we built for
+    bool needs_rebuild = (cgraph->uid == 0) || (cgraph->uid != backend_ctx->uid);
```
JohannesGaessler:
```suggestion
    // If the previous cgraph had a defined UID it can be used to skip rebuilding the subgraphs per simple backend.
    const bool needs_rebuild = (cgraph->uid == 0) || (cgraph->uid != backend_ctx->uid);
```

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1587,173 +1592,182 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
...
+                cgraph_ij->n_nodes = i_node_stop - i_node_start;
+                ggml_hash_set_reset(&cgraph_ij->visited_hash_set);
+                for (size_t i_node = i_node_start; i_node < i_node_stop; i_node++) {
+                    ggml_tensor * node_ij = bcj.nodes[i_node];
+                    cgraph_ij->nodes[i_node - i_node_start] = node_ij;
+                    const size_t hash_pos_orig = ggml_hash_find(&cgraph->visited_hash_set, cgraph->nodes[i_node]);
+                    const size_t hash_pos_ij = ggml_hash_insert(&cgraph_ij->visited_hash_set, node_ij);
+                    cgraph_ij->use_counts[hash_pos_ij] = cgraph->use_counts[hash_pos_orig];
+                }
+                // Assign a fresh uid so downstream backends can cache per-subgraph
```
JohannesGaessler:
```suggestion
```

I would say this comment is redundant since it doesn't explain anything beyond the basic purpose of `ggml_cgraph::uid`.

JohannesGaessler (approved):

PR #22054, opened 2026-04-17
quant: handle shared-KV layer tensors in imatrix-dependent quantization

ajfonthemove:
## Summary

Models with shared KV layers reuse KV cache from earlier layers for layers
>= `n_layer_kv_from_start`. The `attn_k`/`attn_v` weight tensors for these shared
layers may exist in the GGUF but are never used during inference — which means the
imatrix collector never records importance data for them.

When quantizing such a GGUF to any type that requires an imatrix (IQ1_S, IQ1_M,
IQ2_XXS, IQ2_XS, IQ2_S, IQ2_M, IQ3_XXS, Q2_K_S), the quantizer crashes:

```
Missing importance matrix for tensor blk.N.attn_k.weight in a very low-bit quantization
```

This PR teaches the quantizer to detect shared-KV layer attention tensors using the
existing `hparams.has_kv()` method and assign them Q8_0, which does not require imatrix
data. Since these weights are unused at inference, the quantization type has zero impact
on model quality.

## Relationship to #21739

PR #21739 addressed the **loading** side — making shared-KV tensors optional so llama.cpp
can load newer GGUFs where converters omit them entirely. This PR addresses the
**quantization** side — handling the case where the tensors *are* present (older GGUFs,
alternative converters, re-quantizing existing F16 files).

Both fixes are complementary:

| Scenario | #21739 (load) | This PR (quant) |
|---|---|---|
| GGUF with tensors omitted | Needed | Not triggered |
| GGUF with tensors present | Not triggered | Needed |

## Changes

- `src/llama-quant.cpp`: Add `tensor_is_in_shared_kv_layer()` helper and early return
  in `llama_tensor_get_type_impl()` — assigns Q8_0 to shared-KV attention tensors,
  with an early guard on `n_layer_kv_from_start < 0` so non-shared-KV models are unaffected
- `tests/test-shared-kv-quant-imatrix.cpp`: Regression test with a mock shared-KV model
- `tests/CMakeLists.txt`: Register the new test

## Test plan

- [ ] `cmake --build build && ctest --test-dir build -R shared-kv-quant` passes
- [ ] Quantize a shared-KV model GGUF (with tensors present) to IQ2_M without imatrix —
      should succeed with Q8_0 for shared layers instead of crashing
- [ ] Quantize a non-shared-KV model — no behavior change

## Acknowledgement

Claude Code running Opus 4.6 assisted in writing this PR.

### DIFF
--- a/src/llama-quant.cpp
+++ b/src/llama-quant.cpp
@@ -156,6 +156,22 @@ static bool category_is_attn_v(tensor_category cat) {
            cat == tensor_category::ATTENTION_KV_B;
 }
 
+// check if a tensor belongs to a shared-KV layer (layers that reuse KV cache from earlier layers)
+// in such models, layers >= n_layer_kv_from_start do not use their own attn_k/attn_v weights
+// during inference, so these tensors will never appear in the importance matrix
+static bool tensor_is_in_shared_kv_layer(const std::string & tensor_name, const llama_hparams & hparams) {
+    if (hparams.n_layer_kv_from_start < 0) {
+        return false;  // model does not use shared KV
+    }
+    static const std::regex pattern(R"(blk\.(\d+)\.)");
+    std::smatch match;
+    if (std::regex_search(tensor_name, match, pattern)) {
+        uint32_t layer_idx = std::stoi(match[1]);
+        return !hparams.has_kv(layer_idx);
+    }
+    return false;
+}
+
 //
 // quantization state
 //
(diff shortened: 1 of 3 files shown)


ajfonthemove:
> Hi @ajfonthemove, thanks for your contribution!
> 
> 
> * **AI-generated content**: This project does not accept PRs, descriptions or commit messages that are fully or predominantly AI-generated. If you have used AI to assist you in writing code, please make sure to disclose that explicitly.
> 

Added acknowledgment. My thoughts and framing, but opus 4.6 wrote the actual PR based on that for efficiency.

PR #18471, opened 2025-12-29
Add self‑speculative decoding (no draft model required)

srogmann:
This PR introduces self-speculative decoding: instead of using a dedicated draft model (which is good, if available, see #18039), the current token history is used to predict future tokens. This can provide a speedup in cases where the output contains repeated parts of the prompt. A typical example is making many small changes in a large source file.

**Example 1** (`gpt-oss-120b` in VRAM): Translation of a few comments in a Python script (chosen as a favorable case).
```
slot update_slots: id  3 | task 0 | created context checkpoint 1 of 8 (pos_min = 324, pos_max = 2883, size = 90.030 MiB)
slot print_timing: id  3 | task 0 | 
prompt eval time =     436.48 ms /  2948 tokens (    0.15 ms per token,  6754.03 tokens per second)
       eval time =   18886.86 ms /  3423 tokens (    5.52 ms per token,   181.24 tokens per second)
      total time =   19323.34 ms /  6371 tokens
slot      release: id  3 | task 0 | stop processing: n_tokens = 6370, truncated = 0
```

Same prompt with `--draft-min 12 --draft-max 48 --spec-self 1`:
```
slot update_slots: id  3 | task 0 | created context checkpoint 1 of 8 (pos_min = 324, pos_max = 2883, size = 90.030 MiB)
slot print_timing: id  3 | task 0 | 
prompt eval time =     431.85 ms /  2948 tokens (    0.15 ms per token,  6826.38 tokens per second)
       eval time =    7163.27 ms /  3193 tokens (    2.24 ms per token,   445.75 tokens per second)
      total time =    7595.13 ms /  6141 tokens
draft acceptance rate = 0.76827 ( 2397 accepted /  3120 generated)
slot      release: id  3 | task 0 | stop processing: n_tokens = 6140, truncated = 0
```

To keep the PR simple, the new argument `--spec-self` reuses the same `draft-min` and `draft-max` values as used for a potential draft model. When combining both speculative decoding methods, these values are shared (no independent tuning of min/max for each method).

**Example 2** (`Qwen3-235B`, with heavy offloading):
```
slot update_slots: id  3 | task 0 | prompt done, n_tokens = 2962, batch.n_tokens = 914
slot print_timing: id  3 | task 0 |
prompt eval time =   15606.37 ms /  2962 tokens (    5.27 ms per token,   189.79 tokens per second)
       eval time =  252551.71 ms /  2973 tokens (   84.95 ms per token,    11.77 tokens per second)
[...]

PR #22055, opened 2026-04-17
spec : save the dynamic ngram cache file

petersid2022:
## Overview

* When we select the `COMMON_SPECULATIVE_TYPE_NGRAM_CACHE` speculative implementation we create a new `common_speculative_state_ngram_cache` state using `create_state_ngram_cache`, where we instantiate the new state by specifying various parameters (e.g, `n_draft`, `save_static` and `save_dynamic`) by hardcoding them. 

* Instead we remove those options and only depend on the existence of lookup_cache_dynamic and lookup_cache_static (both user provided) for enabling the saving and updating the dynamic cache into a file.

* on object destruction check if the above functionality is supported (i.e. if the lookup_cache_dynamic path is provided by the user) then merge all of the ngram_cache_context from each sequence info and save the final common_ngram_cache to a file using common_ngram_cache_save.

* catch std::ifstream::failure the way llama-lookup does, so a permission/read failure cannot replace the file

## Additional information

Add self‑speculative decoding (no draft model required)#18471

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/common/speculative.cpp
+++ b/common/speculative.cpp
@@ -1984,9 +1984,6 @@ struct common_speculative_impl_ngram_cache : public common_speculative_impl {
 
     uint16_t n_draft;
 
-    bool save_dynamic;
-    bool save_static;
-
     struct seq_info {
         size_t cache_size = 0; // number of tokens in n-gram cache
 
@@ -2002,14 +1999,10 @@ struct common_speculative_impl_ngram_cache : public common_speculative_impl {
             uint32_t n_seq,
             uint16_t n_draft,
             const std::string & path_static,
-            const std::string & path_dynamic,
-            bool save_dynamic,
-            bool save_static)
+            const std::string & path_dynamic)
         : common_speculative_impl(COMMON_SPECULATIVE_TYPE_NGRAM_CACHE, n_seq)
         , params(params.ngram_cache)
         , n_draft(n_draft)
-        , save_dynamic(save_dynamic)
-        , save_static(save_static)
     {
         SPC_TRC("%s", "adding speculative implementation 'ngram-cache'\n");
         SPC_TRC("- n_draft=%d, cache_static=%s, cache_dynamic=%s\n",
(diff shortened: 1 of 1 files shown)


common/common.h
```diff
@@ -342,9 +342,12 @@ struct common_params_speculative_ngram_map {
     uint16_t min_hits = 1;  // minimum hits at ngram/mgram lookup for mgram to be proposed
 };
 
-struct common_params_speculative_ngram_cache {
+struct common_params_speculative_ngram_cache : common_params_speculative_ngram_map {
```
petersid2022: This is probably the wrong way of going about this, but I am curious if the same concept of m-gram speculative tokens can be applied in the ngram-cache implemetantion

ggerganov:
The new parameters are never populated. Did you test this change? What is the goal of this PR?

petersid2022:
first of all, thanks for taking the time to review my PR!

TBH, my initial scope for this PR (after coming across the TODO on line 930 of common/speculative.cpp) was to move the `save_[state,dynamic]` bools from inside `create_state_ngram_cache` to the more centralized `common.h`, as the TODO instructed. That would allow for those options to be configurable from the user. Though, after thinking about it some more I decided to drop the booleans and instead rely on the existence or not of `path_[static,dynamic]` parameter.

The way I went about testing my changes was using the below command:
```
./build/bin/llama-server --port 1234 -m ~/models/Qwen3.5-9B-Q8_0.gguf --spec-type ngram-cache --lookup-cache-static ~/static.bin --lookup-cache-dynamic ~/dynamic.bin
```

P.S: In the `common_speculative_state_ngram_cache` constructor, I noticed that if the `common_ngram_cache_load` call fails on either one of them we abort execution. Could we instead call `common_ngram_cache_save()` so that we create the file?

Green-Sky:
I just ran into the same TODO, what is the state of this pr?

petersid2022:
Hello! I intend to work on this PR very soon.

petersid2022:
@ggerganov kind reminder for a code review of this PR, that simply saves the dynamic ngram cache file (when a path is provided) to the filesystem on object destruction. Thanks in advance!
BR,
Peter Sideris

PR #22077, opened 2026-04-18
ci : Downgrade CUDA versions for compatibility

ehfd:
## Overview

Follow-up to https://github.com/ggml-org/llama.cpp/pull/21122; fix https://github.com/ggml-org/llama.cpp/pull/21122#issuecomment-4270291429.

@CISC @taronaeo 

## Additional information

Fix various forms of installed environments by reducing the CUDA version of Docker containers, from 12.8.1 to 12.5.1 for CUDA 12, and 13.1.1 to 13.0.1 for CUDA 13.

# Requirements

- AI usage disclosure: NO


CISC:
Part of the point of bumping to 12.8.x was also to support Blackwell on CUDA12, at some point people will just have to update their drivers or build their own images if they can't.

ehfd:
Yeah, I am also conflicted on this matter.

But I guess Blackwell folks are more likely to use CUDA 13?

ehfd:
Yeah, after some more thinking, I won't proceed.

PR #22082, opened 2026-04-18
mtmd: add pos_0 to mtmd_image_tokens_get_decoder_pos (breaking change)

ngxson:
## Overview

Add `pos_0` parameter to `mtmd_image_tokens_get_decoder_pos`, this allow model to have total control over other dimensions of the RoPE positions

Tested on Qwen3 and confirmed that it doesn't break anything

# Requirements

- AI usage disclosure: no

### DIFF
--- a/tools/mtmd/mtmd-helper.cpp
+++ b/tools/mtmd/mtmd-helper.cpp
@@ -114,10 +114,10 @@ llama_pos mtmd_helper_get_n_pos(const mtmd_input_chunks * chunks) {
     return n_pos;
 }
 
-void mtmd_helper_image_get_decoder_pos(const mtmd_image_tokens * chunks, mtmd_decoder_pos * out_pos) {
+void mtmd_helper_image_get_decoder_pos(const mtmd_image_tokens * chunks, llama_pos pos_0, mtmd_decoder_pos * out_pos) {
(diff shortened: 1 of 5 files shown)


ggerganov (approved):

tools/mtmd/mtmd-helper.cpp
```diff
@@ -188,7 +188,7 @@ struct decode_embd_batch {
             pos[i                     ] = pos_0 + i;
             pos[i + batch.n_tokens    ] = pos_0 + i;
             pos[i + batch.n_tokens * 2] = pos_0 + i;
-            pos[i + batch.n_tokens * 3] = 0; // last pos dim is unused
+            pos[i + batch.n_tokens * 3] = pos_0 + i;
```
CISC, line 191: Is this change ok?

CISC (approved):
You need to fix the failing test obviously. :)

tools/mtmd/mtmd-helper.cpp
ngxson: yes, it should not change anything because `mrope_sections` is always configured such that last dim = 0, so backend will always skip the positional data being set here.

CISC (approved):

PR #22086, opened 2026-04-18
hexagon: hmx opt phase2

chraac:
## Overview

This PR improves branch prediction in the hot dequantization loop for Hexagon HMX.

It restructures `dequantize_x4x2_weight_to_fp16_tiles_task()` so the common case runs with more regular control flow, and moves format-specific logic into helper functions for Q4, MXFP4, and Q8 dequantization.

### Why

The previous loop mixed fast-path and fallback decisions inside the main iteration path, which made branching less predictable.

This change splits processing into:
- a short prefix to reach 4-tile alignment
- a steady-state 4-tile loop for the common case
- a suffix fallback for leftovers

The main goal is to make the hot loop easier for the branch predictor and reduce control-flow overhead.

### Performance

Device: 8gen2
Baseline: `23b8cc499`
Brannch: `f51d46eb9` 

| Model | Test | Baseline (t/s) | PR (t/s) | Change |
|-------|------|----------------|----------|--------|
| qwen35 2B Q4_0 | pp512 | 86.02 ± 1.26 | 87.22 ± 10.19 | neutral |
| qwen35 2B Q4_0 | tg128 | 5.17 | 5.17 | neutral  |

No performance regression observed. Results within measurement variance.
### Notes

- no intended functional change
- helper extraction is mostly to support the loop restructuring

# Requirements

YES, Using AI agent for commit log writing, code reviewing


ggml/src/ggml-hexagon/htp/main.c
```diff
@@ -101,6 +101,7 @@ AEEResult htp_iface_open(const char * uri, remote_handle64 * handle) {
         }
     }
 
+#if __HVX_ARCH__ > 73
```
chraac: That will fix the crash in my 8gen2 device

Issue #21771, opened 2026-04-11
Qwen3 TAG_WITH_TAGGED tool format: p.json() fails on array<object> parameter values; partial tool_call leaked to client poisons multi-turn history


alesha-pro:
## Summary

When a tool has a parameter whose JSON schema is `array<object>` (e.g. `[{"type":"web"}]`) and the model (Qwen3 family using the `TAG_WITH_TAGGED` / `<tool_call><function=...><parameter=...>` format) emits it as the inner value of a `<parameter>` tag, the autoparser's PEG `p.json()` fails mid-stream. Two cascading problems follow:

1. **Root parse failure.** `build_tool_parser_tag_tagged` uses `p.json()` for any parameter that doesn't resolve to a string schema. On array-of-object values the PEG parser aborts, the request ends in a 500 with `Failed to parse input at pos N: <tool_call>...`.
2. **Partial state leaked to the client.** Before the failure, the partial-parse fallback at `common/chat.cpp:2158` returns whatever AST was built so far. For streaming (`is_partial=true`) this surfaces to the client as an SSE delta containing `tool_calls[0].function.arguments = "{"` (one character). The client, acting correctly per the OpenAI spec, appends this tool_call to conversation history.
3. **History poisoning.** Every subsequent request replays that history. `func_args_not_string` at `common/chat.cpp:1842-1860` eagerly parses `tool_calls[].function.arguments` as JSON and throws a 500 on the bare `{`, so the conversation is now stuck: every follow-up crashes with `Failed to parse tool call arguments as JSON: ... unexpected end of input`, even though the client did nothing wrong.

Net effect: a single tool-call attempt with a complex parameter permanently wedges the conversation until the client manually strips the assistant message from history.

## Name and Version

```
version: 8759 (2b2cd57de)
built with GNU 11.4.0 for Linux x86_64
```

HEAD of master as of 2026-04-11, i.e. PR #21216 is included (verified commit `26229755c` is in the binary).

## Operating systems

Linux

## GGML backends

CUDA

## Hardware

4x RTX 3090

## Models

`Qwen3.5-122B-A10B-Q4_K_M` (GGUF). The bug is not specific to this model — it reproduces with any model whose template routes to `TAG_WITH_TAGGED` tool format.

## Steps to reproduce

1. Launch the server with jinja enabled and any tool set containing an `array<object>` parameter:
   ```
[...]

PR #22089, opened 2026-04-18
fix(chat): graceful degradation for malformed tool call arguments

huckstack:
## Problem

When models generate malformed JSON in tool call arguments (e.g., unescaped single quotes like `log_dir = '/home/huckstack/.hermes/logs/cron'`), `func_args_not_string()` throws an HTTP 500 error. This causes cascading failures where every subsequent request crashes with `Failed to parse tool call arguments as JSON: unexpected end of input`.

The function was introduced in commit b283f6d5b with `throw`, which converts a recoverable parsing error into a hard server error.

## Fix

Replace `throw std::runtime_error(...)` with `LOG_WRN(...)` in `func_args_not_string()`. This converts the hard 500 error into a graceful degradation — the server logs a warning and keeps the arguments as a string, allowing the conversation to continue.

## Context

This is related to issue #21771 which reports the same symptom. The root cause there is the PEG parser failing on `array<object>` parameter values, but the symptom (HTTP 500 from `func_args_not_string`) is the same. This fix addresses the symptom by making the error handling more robust.

## Diff

```diff
- throw std::runtime_error("Failed to parse tool call arguments as JSON: " + std::string(e.what()));
+ LOG_WRN("Failed to parse tool call arguments as JSON, keeping as string: %s\n", e.what());
```

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -2190,7 +2190,7 @@ static void func_args_not_string(json & messages) {
                         try {
                             args = json::parse(args.get<std::string>());
                         } catch (const std::exception & e) {
-                            throw std::runtime_error("Failed to parse tool call arguments as JSON: " + std::string(e.what()));
+                            LOG_WRN("Failed to parse tool call arguments as JSON, keeping as string: %s\n", e.what());
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


aldehir:
Unfortunately, this is not sufficient. Some chat templates require `arguments` be an object and not a string. The failure will simply occur further down, in the Jinja engine for these.

Do you have a reliable reproducer I can run? Fixing the core issue is the solution here.

Issue #21827, opened 2026-04-12
Feature Request: Avoid memcpy for mmap-ed weights on Unified Memory architectures (Intel Lunar Lake)


i-Charlys:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

Hi everyone,

I am currently profiling the memory footprint of llama.cpp using the SYCL backend on an Intel Lunar Lake system (Core Ultra 288V).

The Issue
Lunar Lake is a Unified Memory Architecture (UMA) where the CPU and iGPU physically share the same on-package RAM. However, when loading a .gguf model, the memory usage essentially doubles.

From my understanding of the SYCL backend code, even if the model files are mapped into host memory via mmap, the backend still allocates device memory and performs a standard copy (memcpy or equivalent SYCL copy operation) to the device buffer.

Proposed Optimization
Since the memory is physically shared, making a copy defeats the purpose of UMA. Would it be possible for the SYCL backend to:

    Detect if the tensor data originates from a mmap file descriptor.

    Directly import that memory as GPU-accessible (e.g., via sycl::malloc_shared with a host pointer import, or Level Zero's ZE_EXTERNAL_MEMORY_TYPE_FLAG_DMA_BUF for the mmap file descriptor) to achieve a true zero-copy initialization?

I am a CS student and not a C++/SYCL expert, so I cannot submit a PR for this myself. However, as Meteor Lake and Lunar Lake chips are becoming very common for local AI, a true zero-copy path for SYCL would be a massive performance and memory win.

Thank you for your incredible work on this project!

### Motivation

Memory capacity is the absolute primary bottleneck for running local LLMs. Many users on the new Intel Core Ultra platforms (Meteor Lake and Lunar Lake) are using thin-and-light laptops with a fixed, unified memory pool of 16GB or 32GB.

Currently, because the SYCL backend duplicates the weights in memory (system RAM via mmap + VRAM allocation via SYCL), a user with a 16GB machine might struggle to run an 8GB model, as it effectively consumes ~16GB of total system memory.

[...]

PR #21597, opened 2026-04-08
SYCL: fix multi-GPU system RAM exhaustion by using Level Zero allocations

PMZFX:
## Summary

- Replace `sycl::malloc_device` with `zeMemAllocDevice` for GPU memory allocation in the SYCL backend
- Replace `sycl::free` with `zeMemFree` for corresponding deallocations
- Replace host-staged `dev2dev_memcpy` with direct Level Zero cross-device copy
- Link against `ze_loader` for Level Zero API access
- All changes include automatic fallback to original SYCL path if Level Zero is unavailable

## Problem

On Intel multi-GPU systems, `sycl::malloc_device` triggers the xe kernel driver's DMA-buf/TTM export path (`xe_gem_prime_export` -> `ttm_pool_alloc_page`), which creates a 1:1 mirror of every VRAM allocation in system RAM. This causes system RAM to scale linearly with total VRAM allocated across GPUs, leading to OOM crashes during multi-GPU inference even when models fit entirely in VRAM.

Measured on dual Intel Arc Pro B70 (32GB each, 64GB total VRAM) with 64GB system RAM:
- `sycl::malloc_device` 4 GiB = +4,112 MiB system RAM (1:1 mirror)
- `zeMemAllocDevice` 4 GiB = +8 MiB system RAM (no mirror)

A 15.6 GiB Q4_K_M model consumed 60 GiB of system RAM during dual-GPU inference with `sycl::malloc_device`, causing repeated OOM crashes.

## Solution

`zeMemAllocDevice` allocates GPU memory through Level Zero's SVM/P2P path instead of the DMA-buf/TTM path, avoiding the host memory staging entirely. SYCL kernels can read `zeMemAllocDevice` pointers with full interop, no compatibility issues.

Changes:
- New `static ggml_sycl_malloc_device()` in `ggml-sycl.cpp` and `ggml_sycl_free_device()` in `common.cpp` that try Level Zero first, fall back to SYCL
- Replaced 4 allocation sites: single-device buffer, split buffer, memory pool, overflow pool
- Replaced 5 deallocation sites: buffer destructor, pool destructor, pool overflow, `release_extra_gpu`, and the pre-existing `sycl::free` in `common.cpp`
- Updated `dev2dev_memcpy` to use `zeCommandListAppendMemoryCopy` for direct cross-device transfers

## Test results

[...]

PR #22098, opened 2026-04-18
[SYCL] Add Zero-Copy path with Cache Flushing for Intel UMA (Lunar Lake/Meteor Lake)

i-Charlys:
### Description
This PR follows up on the discussion in issue #21827 regarding true Zero-Copy on Intel Unified Memory Architectures (UMA) like Lunar Lake. 

As a CS student exploring low-level optimizations, I worked with an AI assistant to implement a DMA-BUF based zero-copy path for the SYCL backend. This implementation aims to eliminate weight duplication in RAM (mmap + device allocation) which is a major bottleneck for integrated GPUs with limited memory.

### Changes
1. **Cache Coherency Fix**: The initial experiment suffered from model hallucinations because the iGPU was reading stale data from RAM while the weights were still in the CPU caches. I added a manual cache flushing mechanism using `_mm_clflush` and `_mm_sfence` to ensure the iGPU sees the correct data immediately after the host `memcpy`.
2. **Environment Toggle**: Added `GGML_SYCL_ENABLE_ZERO_COPY` (default: 0). This allows users to opt-in to this experimental path without affecting stability for discrete GPUs or other backends.
3. **Zero-Copy Path**: When enabled, it uses `udmabuf` (on Linux) to create pinned shared memory, bypassing the standard SYCL/Level-Zero staging copies.

### Technical Note
I am still learning the ropes of C++ and SYCL, so I focused on a "system-level" approach (using `immintrin.h` for flushing). I'm aware that a more native Level-Zero or SVM-based approach might be preferred long-term, but this provides a working solution for Lunar Lake users today.

### How to test
1. Build with SYCL and Level-Zero support.
2. Run with: `export GGML_SYCL_ENABLE_ZERO_COPY=1`.
3. Observe memory usage: it should no longer double when loading large models.
4. Verify output: The model should now generate coherent text (no more logic corruption).


arthw (commented):
1.
`Observe memory usage: it should no longer double when loading large models.`
Without this PR, after loading the model, will the memory usage be reduced? or still double?

2. Except above item, is there any other benefit of this PR?

3. As my understanding, when use mmap, during load the model,  there is double memory usage. This PR can fix it. 
For the case that no-mmap, there is no such double memory issue. 
Is it right

Thank you!

i-Charlys:
Hi @arthw, thanks for the review! I've just resolved the merge conflicts and updated the PR. To answer your questions:

1. Memory usage after loading: Without this PR (using default mmap), the model weights effectively stay doubled in physical RAM: once in the OS file cache (mmap) and once in the SYCL USM allocation. On systems with 16GB/32GB (common for Lunar Lake laptops), this halves the usable model size. This PR ensures a 1:1 ratio by sharing the exact same physical pages via DMA-BUF.

2. Performance benefit (Loading Speed): Besides saving RAM, this path allows near-instant loading. By bypassing the PCIe-style memcpy loop, a large model is ready as soon as the DMA-BUF is imported.

3. The --no-mmap case: You're right that --no-mmap avoids double allocation, but it makes loading extremely slow on integrated GPUs as it reads data chunk-by-chunk. This PR provides the loading speed of mmap with the memory efficiency of no-mmap.

Additionally, we implemented explicit cache flushing (clflush). We found that on Lunar Lake, standard USM sometimes causes model hallucinations because the iGPU reads stale data from RAM while updates are still in CPU caches. This implementation fixes that reliability issue.

i-Charlys:
Hi @arthw, I've updated the PR with a much cleaner and more portable 

 1. Removed x86 intrinsics: I replaced the clflush hack with the native zeContextSystemBarrier from Level Zero. This ensures cache coherence between CPU and iGPU in a portable way, without processor-specific instructions.
 2. Enhanced Fallback: I improved the error handling for /dev/udmabuf. If the device is inaccessible or the kernel limit is hit, the backend now gracefully falls back to standard sycl::malloc_shared instead of failing.
 3. UMA consistency: I also applied the system barrier to the regular integrated GPU path for better reliability across all Intel UMA devices.

The code is now strictly opt-in, portable, and includes robust fallbacks for production environments. Thanks for the guidance!

arthw:
@i-Charlys 
Thank you for your update and explain!
I will check and review it.

What's the OS where this issue appear?

arthw:
@i-Charlys 
I test this PR on iGPU with gemma-4-26B-A4B-it-UD-Q4_K_M.gguf (17GB).

From the following table, we can see the improvement of memory usage in this PR:
  - The VIRT usage is reduced from 26-31 GB to 19GB.
  - The RES usage is smaller of mmap case.

The total memory usage of process is same.

So, it's useful from the memory usage point view.
This PR is same as another: https://github.com/ggml-org/llama.cpp/pull/21597.
It can resolve the GTT issue on Ubuntu 26, (not appear on Ubuntu 25/24).

I will prepare the Ubuntu 26 to verify both PRs.
Because sycl memory API is still mandatory in some features in the future: SYCL graph, USM, we need to allow user to switch the memory API between SYCL and Level Zero API. The PR https://github.com/ggml-org/llama.cpp/pull/21597 support it.

@PMZFX 
After my test is passed, maybe we could merge both into one.

Thank you!


| Metric | PR mmap | PR no-mmap | base mmap | base no-mmap |
|---|---:|---:|---:|---:|
| Total Memory (MiB) | 63768.9 | 63768.9 | 63768.9 | 63768.9 |
| Free Memory (MiB) | 1710.2 | 2104.6 | 2151.7 | 1773.8 |
| Used Memory (MiB) | 27184.8 | 28044.3 | 27314.5 | 28261.8 |
| Buff/Cache (MiB) | 59859.5 | 58605.5 | 59137.0 | 59135.8 |

| Process Metric | PR mmap | PR no-mmap | base mmap | base no-mmap |
|---|---|---|---|---|
| PID | 763372 | 763813 | 763478 | 763544 |
| PR | 20 | 20 | 20 | 20 |
| NI | 0 | 0 | 0 | 0 |
| VIRT | 19.4g | 19.2g | 3112304 | 2678060 |
| RES | 1.5g | 1.3g | 1.8g | 1.3g |
| SHR | 1.0g | 260320 | 1.0g | 248464 |
| State (S) | S | S | S | S |
| %CPU | 0.0 | 0.0 | 0.0 | 0.0 |
| %MEM | 2.4 | 2.1 | 2.9 | 2.1 |
| TIME+ | 0:23.56 | 0:12.74 | 1:56.24 | 0:12.92 |
| COMMAND | llama-server | llama-server | llama-server | llama-server |

i-Charlys:
Hi @arthw,

Thank you so much for testing the PR and sharing these metrics! Seeing the VIRT usage drop from 31GB to 19GB is exactly the true Zero-Copy behavior I was aiming for on UMA devices.

To answer your question regarding the OS: I encountered, developed, and tested this issue on CachyOS (Arch Linux rolling release), running a 6.x Linux kernel with intel-compute-runtime version 26.09.37435.1.

I completely agree with your suggestion to combine this with PR #21597.  I am very open to merging efforts. Please let me know how you and @PMZFX would like to proceed with combining the PRs, or if you need me to rebase my udmabuf logic on top of his branch!

Thank you again for your guidance.

NeoZhangJianyu:
@i-Charlys @PMZFX 
https://github.com/ggml-org/llama.cpp/pull/21597 issue would depend on the Ubuntu 26.
Ubuntu 26 can't be upgraded from old Ubuntu.
I have to backup a PC to install new OS on a PC.
It will take a little more time.

Thank you!

arthw:
@i-Charlys 
Since https://github.com/ggml-org/llama.cpp/pull/21597 is merged.
Could you check if the issue is present in current master branch?
If it's present, could you rebase this PR?

Thank you!

i-Charlys:
@arthw i'll work on it ASAP

cndjonno:
@i-Charlys Did you get chance to rebase this PR?

PR #22099, opened 2026-04-18
[WebGPU] Implement async tensor api and event api 

nikhilJain17:
## Overview

This PR implements the async tensor and event api necessary for the WebGPU backend to use the async loading mode to load models. This is needed because we have strict memory requirements when running wllama with the WebGPU backend (especially on Safari and on mobile devices). The async tensor API uses only four 1MB buffers to load a model, while the default loading mode uses a single resizable buffer. Using the async tensor API reduces our memory footprint by ~20-25%.

## Additional information

Some figures on memory usage in wllama with these and other changes:
steady_state_bar_cold
steady_state_bar_warm
timeseries_cold
timeseries_warm

# Requirements

- AI usage disclosure: Yes, for discussing design of the API.


ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2616,22 +2616,115 @@ static ggml_status ggml_backend_webgpu_graph_compute(ggml_backend_t backend, str
...
+    size_t aligned_size = (size + 3) & ~3ULL;
+
+    // Create a one-shot staging buffer
+    wgpu::BufferDescriptor desc{};
+    desc.size             = aligned_size;
+    desc.usage            = wgpu::BufferUsage::MapWrite | wgpu::BufferUsage::CopySrc;
+    // Allow CPU to write immediately upon buffer creation
+    desc.mappedAtCreation = true;
+    desc.label            = "set_tensor_async_scratch_buffer";
+    wgpu::Buffer staging  = buf_ctx->global_ctx->device.CreateBuffer(&desc);
```
reeselevine: why do we need a one-shot buffer here? Can we use queue.WriteBuffer like the non-async version?

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2616,22 +2616,115 @@ static ggml_status ggml_backend_webgpu_graph_compute(ggml_backend_t backend, str
...
+    ggml_backend_webgpu_context * backend_ctx = (ggml_backend_webgpu_context *) backend->context;
+    ggml_backend_webgpu_wait_queue(backend_ctx->webgpu_ctx->global_ctx);
+}
+
 static ggml_backend_i ggml_backend_webgpu_i = {
     /* .get_name                = */ ggml_backend_webgpu_name,
     /* .free                    = */ ggml_backend_webgpu_free,
-    /* .set_tensor_async        = */ NULL,
+    /* .set_tensor_async        = */ ggml_backend_webgpu_set_tensor_async,
     /* .get_tensor_async        = */ NULL,
```
reeselevine, line 2708: so `get_tensor_async` wasn't necessary for this? I guess that makes sense, since we're only targeting loading of models.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -2616,22 +2616,115 @@ static ggml_status ggml_backend_webgpu_graph_compute(ggml_backend_t backend, str
     return GGML_STATUS_SUCCESS;
 }
 
+struct ggml_backend_webgpu_event_context {
+    webgpu_global_context global_ctx;
+    wgpu::Future          future;
+    bool                  recorded = false;
+};
+
+static ggml_backend_event_t ggml_backend_webgpu_device_event_new(ggml_backend_dev_t device) {
```
reeselevine, line 2625: can you explain a little more why the event API is needed for this change? I don't see how it's used within `set_tensor_async` at all, and there doesn't seem to really be any information encoded in an event.

nikhilJain17: Sure, so both the event API and async API are required for the async loading mode: https://github.com/ggml-org/llama.cpp/blob/72d693e4fbf0a188172951fe667d5e2f354ee438/src/llama-model-loader.cpp#L1461 

The weight loader creates an event for every buffer, which is marked as recorded when it is done writing and available for use.  https://github.com/ggml-org/llama.cpp/blob/72d693e4fbf0a188172951fe667d5e2f354ee438/src/llama-model-loader.cpp#L1487

Since there are only 4 buffers, it uses the event API to wait for a free buffer before executing another `set_tensor_async`: https://github.com/ggml-org/llama.cpp/blob/72d693e4fbf0a188172951fe667d5e2f354ee438/src/llama-model-loader.cpp#L1591

When a buffer is written, it marks the event as done: https://github.com/ggml-org/llama.cpp/blob/72d693e4fbf0a188172951fe667d5e2f354ee438/src/llama-model-loader.cpp#L1614

So the event API is similar to a future, but holding no data and only used for synchronization.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
nikhilJain17: Yes, I believe `get_tensor_async` is for GPU -> CPU transfer.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
nikhilJain17: Yes, I wasn't sure if this was the best method but I just wanted to get it working for testing the memory reduction instead of implementing a buffer pool. `queue.WriteBuffer` would be cleaner, will update.

reeselevine (approved):

CISC (approved):

PR #22103, opened 2026-04-19
feat: Support sarashina2.2-vision-3b model

samuraieng:
## Overview

This PR adds new PROJECTOR_TYPE to support for [Sarashina2.2 Vision 3B](https://huggingface.co/sbintuitions/sarashina2.2-vision-3b) model to llama.cpp, both llama-cli, llama-server and connvert_hf_to_gguf.py.
This implementation is updated to support the Sarashina model with minimum modifications.

### Changes
This PR updates six (6) files
- convert_hf_to_gguf.py
    - Add Sarashina2VisionForCausalLM
    - Add some logic to proceed Sarashina2.2-vision-3B model
- gguf-py/gguf/constants.py
    - Add SARASHINA2VL
- tools/mtmd/clip-impl.h
- tools/mtmd/clip.cpp
- tools/mtmd/mtmd.cpp
    - Add PROJECTOR_TYPE_SARASHINA2VL_MERGER
    - Modify switch-case to support the type
- tools/mtmd/tests.sh
    - Add sarashina2.2-vision-3b:Q4_K_M as add_test_vision

### Testing
- ctest -L main passed
- tools/mtmd/tests.sh passed (with this modification)
- Manual test with using llama-cli and llama-server with following models
    - Sarashina2.2-vision-3b-gguf (Target to support)
    - Qwen3-VL-2B-Instruct-gguf (For regression testing)
    - Qwen2.5-VL-3B-Instruct (Locally converted to gguf, to check convert_hf_to_gguf.py)

## Requirements
- AI usage disclosure: YES - AI assisted with codeing. All code was human-authored, manually modified, reviewed and tested.


convert_hf_to_gguf.py
```diff
@@ -2947,6 +2947,13 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
             else:
                 return
 
+        if self.origin_hf_arch.startswith('Sarashina2VisionForCausalLM'):
+            # Remove llm. from name 
+            if name.startswith("llm."):
+                name = name[len("llm."):]
+            elif name.startswith("visual.") or name in ("norm.weight", "norm.bias"):
+                return  #Skip processing "modify_tensors"
```
CISC, line 2955: This is not ok, you can remove `llm.` unconditionally in `Qwen2VLModel`, `visual` is already skipped there and why are you dropping `norm`?

samuraieng: Point-1
Sarashina uses LlamaModel during text model processing.
norm.bias and norm.weight are for Visual model. After these two, it starts like visual.blocks.... Therefore, norm.weight, norm.bias and visual must be removed during text model processing

```
(tensor information from self.modify_tensors)
llm.norm.weight  <- for text, till here
norm.bias        <- start block for visual model
norm.weight      <- for visual
visual.....      <- for visual, till the end
``` 

Point-2
For text processing, it will not go to Qwen2VLModel. I removed the process type from the part. This is the new push to PR
Even Qwen2VLModel has the drop logic for visual, but the code must have it on LlamaModel, this was the reason.

gguf-py/gguf/constants.py
```diff
@@ -4138,6 +4138,7 @@ class VisionProjectorType:
     YOUTUVL = "youtuvl"
     NEMOTRON_V2_VL = "nemotron_v2_vl"
     HUNYUANOCR     = "hunyuanocr"
+    SARASHINA2VL = "sarashina2vl_merger"
```
ngxson: delete `_merger` from the name

tools/mtmd/clip-impl.h
```diff
@@ -338,6 +339,7 @@ static std::map<projector_type, std::string> PROJECTOR_TYPE_NAMES = {
     { PROJECTOR_TYPE_KIMIK25,   "kimik25"},
     { PROJECTOR_TYPE_NEMOTRON_V2_VL, "nemotron_v2_vl"},
     { PROJECTOR_TYPE_HUNYUANOCR, "hunyuanocr"},
+    { PROJECTOR_TYPE_SARASHINA2VL_MERGER, "sarashina2vl_merger"}
```
ngxson: delete `_merger` and `_MERGER` from the name

convert_hf_to_gguf.py
```diff
@@ -4220,6 +4227,9 @@ def set_gguf_parameters(self):
                 if fullatt_block_indexes[i] - fullatt_block_indexes[i - 1] != n_wa_pattern:
                     raise ValueError(f"Invalid fullatt_block_indexes: {fullatt_block_indexes}")
             self.gguf_writer.add_vision_n_wa_pattern(n_wa_pattern)
+        elif model_type == "sarashina2_vision":
+            self.gguf_writer.add_clip_projector_type(gguf.VisionProjectorType.SARASHINA2VL)
```
ngxson: do we really need a new projector type? can't we reuse qwen25vl?

ngxson (changes requested):

tools/mtmd/tests.sh
```diff
@@ -92,6 +92,7 @@ add_test_vision "ggml-org/DeepSeek-OCR-GGUF:Q8_0" -p "Free OCR." --chat-template
 add_test_vision "ggml-org/dots.ocr-GGUF:Q8_0" -p "OCR"
 add_test_vision "ggml-org/HunyuanOCR-GGUF:Q8_0" -p "OCR"
 add_test_vision "ggml-org/gemma-4-E2B-it-GGUF:Q8_0" --jinja
+add_test_vision "samuraieng/sarashina2.2-vision-3b-gguf:Q4_K_M"
```
ngxson: test models must be under `ggml-org` account, please remove this

convert_hf_to_gguf.py
ngxson: you should make a new conversion class for `Sarashina2VisionForCausalLM` that extends from `LlamaModel`

CISC: It all still applies, but for `LlamaModel`, remove `llm.` unconditionally, add `visual` to `is_multimodal_tensor` and the `norm` tensor is not mapped to the mmproj, so you are silently dropping it.

ngxson: I mean something like this:

```py
@ModelBase.register("Sarashina2VisionForCausalLM")
def Sarashina2TextModel(LlamaModel):
        if name == "norm.weight" or name == "norm.bias":
            # this is visual tensor for Sarashina2
            name = "visial." + name
```

samuraieng: Hi, Thank you very much for spending your time for my review.
I have one question, related to coding policy, I would say.

In general, there is two style to go.
a) Minimize new and modify code part
b) Even it becomes big effort, split code based on models and targets

Example,
a)
ngxson's comment to vision part. You asked "we really need a new projector type? can't we reuse qwen25vl?"
This means, re-use code as much as possible, do not create new type and others easily.
I understood what you say. This is true.

b)
Excuse me, again, ngxson's comment just above, @ModelBase.register("Sarashina2VisionForCausalLM")
If I take this direction, then I have to update a lot of code than a), right? But separate models things, this is also true, connection between functions should be less.

It could be better to align how to update code as policy. Then I will follow the rule.
For work-effort, it is better to take minimum update, take option a).  If you let me choose, then I will update to reduce the updated parts.

Woudl you please share your opinion? After that I will update my code to follow your option. If you need more info, I will try to explain more. Thanks!

ngxson: > ngxson's comment to vision part. You asked "we really need a new projector type? can't we reuse qwen25vl?"

This question is for vision encoder, I'll reply in the thread below.

> Excuse me, again, ngxson's comment just above, @ModelBase.register("Sarashina2VisionForCausalLM")
If I take this direction, then I have to update a lot of code than a), right?

It doesn't have much code added because you're extending from `LlamaModel`. My comment above demonstrated that it only adds around ~10 lines of code

convert_hf_to_gguf.py
ngxson: @samuraieng It seems like it's the same as `qwen25vl`. But from what I can see, text model doesn't use m-rope because it's based on `llama` arch.

I will make a change in `libmtmd` to support this case, so that it will work correctly.

In the meantime, please remove the new `VisionProjectorType` and re-use the same `qwen25vl` for the vision encoder.

convert_hf_to_gguf.py
CISC: When you require model-specific changes (ie, the `norm` tensors) you should create a new sub-class with those changes, just as @ngxson suggested, the other changes are not really model-specific and can be done unconditionally in the parent `LlamaModel` class.

convert_hf_to_gguf.py
ngxson: It will be possible to use qwen25vl vision encoder with non-qwen text models (i.e. llama in this case) with https://github.com/ggml-org/llama.cpp/pull/22161

samuraieng: Sure, let me prepare removing ProjectorType and re-use qwen

CISC:
@samuraieng Bad merge.

convert_hf_to_gguf.py
samuraieng: I updated to follow your comments. Both text and vision have new classes.

@ModelBase.register("Sarashina2VisionForCausalLM")
class Sarashina2VLTextModel(LlamaModel):

To call this class, one part has to be fixed.
    if model_type == ModelType.TEXT and arch == "Sarashina2VisionForCausalLM":
        return "Sarashina2VisionForCausalLM"
        
 Ran
 - ctest -L main
 - manual test with llama-server with sarashina and Qwen

gguf-py/gguf/constants.py
samuraieng: Removed this line. New update does not require my code.

tools/mtmd/clip-impl.h
samuraieng: Removed this line. New update does not require my code.

tools/mtmd/tests.sh
samuraieng: Sure. I removed this from the new update.

convert_hf_to_gguf.py
samuraieng: Create new class. MMPROJ side, I think no model specific code is existed in common code.

@ModelBase.register("Sarashina2VisionForCausalLM")
class Sarashina2VLVisionModel(MmprojModel):
    model_type = ModelType.MMPROJ

samuraieng:
Would you please review again? I think I captured all your comments in the latest update.

CISC:
@samuraieng Bad merge.

samuraieng:
Sorry for made this bad merge. I pushed again with new code. Would you please reviwe again?
As I wrote, only convert_hf_to_gguf.py is the file to review. Most of modification are made inside new classes.

convert_hf_to_gguf.py
```diff
@@ -2815,6 +2815,7 @@ def prepare_tensors(self):
     "LlavaForConditionalGeneration",
     "VoxtralForConditionalGeneration",
     "IQuestCoderForCausalLM",
+    "Sarashina2VisionForCausalLM",
```
```suggestion
```

convert_hf_to_gguf.py
```diff
@@ -13197,6 +13198,74 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+        super().set_gguf_parameters()
+        self.gguf_writer.add_clip_projector_type(gguf.VisionProjectorType.QWEN2VL)
+        self.gguf_writer.add_vision_spatial_merge_size(2)
+        self.gguf_writer.add_vision_attention_layernorm_eps(self.global_config.get("rms_norm_eps", 1e-6))
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iterable[tuple[str, Tensor]]:
+        assert self.hparams_vision is not None
+        # Two tensors will be ignored
+        #if name in ('norm.weight', 'norm.bias'):
+        #    name = "visual.post_layer" + name
```
CISC: Why are they ignored?

convert_hf_to_gguf.py
```diff
@@ -13197,6 +13198,74 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+        if name.startswith("visual."):
+            # process visual tensors
+            # split QKV tensors if needed
+            if ".qkv." in name:
```
CISC: No, it's not needed.

convert_hf_to_gguf.py
```diff
@@ -13197,6 +13198,74 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None):
+        if name.startswith("llm."):
+            name = name[len("llm."):]
+            yield from super().modify_tensors(data_torch, name, bid)
+
+
+@ModelBase.register("Sarashina2VisionForCausalLM")
+class Sarashina2VLVisionModel(MmprojModel):
+    model_type = ModelType.MMPROJ
+
```
ngxson: no need to define this in mmproj (same as other models)

convert_hf_to_gguf.py
```diff
@@ -13197,6 +13198,74 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+        self.origin_hf_arch = hparams.get('architectures', [None])[0]
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None):
+        if name.startswith("llm."):
+            name = name[len("llm."):]
+            yield from super().modify_tensors(data_torch, name, bid)
+
+
+@ModelBase.register("Sarashina2VisionForCausalLM")
+class Sarashina2VLVisionModel(MmprojModel):
```
ngxson: instead of copy-paste the whole code from `Qwen2VLVisionModel`, why not extending from it?

```
class Sarashina2VLVisionModel(Qwen2VLVisionModel):
```

convert_hf_to_gguf.py
samuraieng: Removed.

samuraieng:
Mostly, MMPROJ side was updated. The new class follows Qwen2VLVisionModel

convert_hf_to_gguf.py
```diff
@@ -13197,6 +13197,28 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        hparams = ModelBase.load_hparams(self.dir_model, is_mistral_format=False)
+        self.origin_hf_arch = hparams.get('architectures', [None])[0]
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None):
+        if name.startswith("llm."):
+            name = name[len("llm."):]
+            yield from super().modify_tensors(data_torch, name, bid)
```
```suggestion
    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None):
        if name.startswith("llm."):
            name = name.replace("llm.", "", 1)
        elif name.startswith("norm.") or name.startswith("visual."):
            return

        yield from super().modify_tensors(data_torch, name, bid)
```

convert_hf_to_gguf.py
```diff
@@ -13454,6 +13476,8 @@ def get_model_architecture(hparams: dict[str, Any], model_type: ModelType) -> st
     # TODO: refactor this later to avoid adding exception here
     if model_type == ModelType.TEXT and arch == "StepVLForConditionalGeneration":
         return arch
+    if model_type == ModelType.TEXT and arch == "Sarashina2VisionForCausalLM":
+        return "Sarashina2VisionForCausalLM"
```
```suggestion
    if model_type == ModelType.TEXT and arch in ("StepVLForConditionalGeneration", "Sarashina2VisionForCausalLM"):
        return arch
```

samuraieng: New commit follow this code

convert_hf_to_gguf.py
samuraieng: model_arch = gguf.MODEL_ARCH.LLAMA
This line remained. Because it falls into if "model_arch" not in cls.__dict__:.
Rest of code follow this change

CISC (approved):

samuraieng:
@ngxson , would you please review the latest code?

masabamba:
@ngxson , would you please review the latest code? The code I modified are same as before. Sync with the master branch.

convert_hf_to_gguf.py
```diff
@@ -13373,6 +13373,26 @@ def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None) -> Iter
...
+class Sarashina2VLTextModel(LlamaModel):
+    model_arch = gguf.MODEL_ARCH.LLAMA
+
+    def modify_tensors(self, data_torch: Tensor, name: str, bid: int | None):
+        if name.startswith("llm."):
+            name = name.replace("llm.", "", 1)
+        elif name.startswith("norm.") or name.startswith("visual."):
+            return
+
+        yield from super().modify_tensors(data_torch, name, bid)
```
```suggestion
    @classmethod
    def filter_tensors(cls, item: tuple[str, Callable[[], Tensor]]) -> tuple[str, Callable[[], Tensor]] | None:
        name, gen = item

        if name.startswith("llm."):
            name = name.replace("llm.", "", 1)
        elif name.startswith("norm."):
            return None

        return super().filter_tensors((name, gen))
```

masabamba:
I updated to follow [this](https://github.com/ggml-org/llama.cpp/pull/22103#discussion_r3194534104) review comment. Would you please review again?

ngxson (approved):

PR #22104, opened 2026-04-19
server: Add URL handling for multimodal_data on /completion endpoint

cetarthoriphros:
## Overview

Handle URL schemes on the /completion endpoint by using handle_media().

## Additional information

Sending multimedia files over http to /completion can take a significant amount of time due to the base64 encoding (often done through a script and piping), decoding, and the http transfer to llama-server. I started by simply adding file:// handling, but it seemed useful to add remote files, as well as centralizing how files are fetched when more modalities are added.

This change moves handle_media() higher in tools/server/server-common.cpp so that tokenize_input_subprompt() can use it and lets /completion inherit the URL schemes supported by /v1/completions.

The allow_raw_b64 parameter, when set to true, allows raw base64-encoded data, as that is how /completion currently works. When false, just like before the change, handle_media() strictly checks for a properly formatted data: URL.

Change the json media_obj parameter to std::string data. The idea is to make it more versatile by not having a required json format, which is bound to change as third-party APIs are updated or added. Callers must extract the URL from the user query before calling handle_media().

Make explicit checks for http:// and https:// to avoid possible collisions. Since handle_media() can now accept raw base64-encoded data, it needs a more strict check than simply looking for an "http" prefix.

## Extra

While updating the note about multimodal_data in toos/server/README.md, I added a comment about fetching the media marker from the /props endpoint. If it's deemed unrelated, I can open another PR for it.

# Requirements

- AI usage disclosure: NO


tools/server/server-common.cpp
```diff
@@ -703,6 +703,74 @@ static std::string fnv_hash(const uint8_t * data, size_t len) {
...
+ * media_path always end with '/', see arg.cpp
+ * Accepted url formats: file://, http://, https://, data:MIME;base64,data, or
+ * raw base64-encoded data if allow_raw_b64 is true.
+ */
+static void handle_media(
+        std::vector<raw_buffer> & out_files,
+        const std::string & url,
+        const std::string & media_path,
+        bool allow_raw_b64) {
+    if (string_starts_with(url, "http://") || string_starts_with(url, "https://")) {
```
ngxson, line 726: we cannot accept this change, it's not scalable to just duplicate the handling function between raw completion and chat completion endpoints

ngxson (changes requested):

tools/server/server-common.cpp
cetarthoriphros: There is no code duplication. I simply moved the function higher in the file so that tokenize_input_subprompt() could use it. The change of the json media_obj parameter to a simple string is precisely to make it more scalable.

Did you mean to say that raw completion and chat completion should use separate functions?

cetarthoriphros: @ngxson Sorry to ping you, but I'd really like some clarification about why this change is not acceptable.

Having llama-server load local files directly on /completion speeds up all multimedia requests, and remote files, while I don't yet use them, are also really convenient.

As for scalability, I'm not sure what you mean. So far, there's only two schemes accepted in llama.cpp: URLs (http[s]://, file://, and data: URLs) and raw base64-encoded data. There's a lot of transfer schemes out there, but it's unlikely that any other than these will ever become part of any standard http API. Having a single, simple function that deals with file fetching and only requires basic types as parameters makes it more generally useful when new APIs are implemented.

I know most UIs now use chat completion, but I'm using a vimscript based off the old llama.vim for the /completion endpoint. I was able to add gemma 4's template support in about 4 lines of code to the script and have it working right after release, before all the chat template fixes were merged.

Raw completion is still a very useful tool and it'd be a shame to see it lag in basic features. Specially when it needs so little maintenance compared to chat completion.

Let me know if there's a way to make this change work for the project.

CISC:
Bad merge, please fix.

cetarthoriphros:
Sorry about the noise, @CISC. I'm still getting used to the local/github workflow. Seems to be fine now. Could I get another pair of eyes on this patch? I still haven't gotten a reply from @ngxson . @aldehir or @ServeurpersoCom ?

CISC:
> Sorry about the noise, @CISC. I'm still getting used to the local/github workflow. Seems to be fine now. Could I get another pair of eyes on this patch? I still haven't gotten a reply from @ngxson . @aldehir or @ServeurpersoCom ?

No worries, thanks for fixing quickly.

They will get around to it eventually, we are all pretty overloaded...

ServeurpersoCom:
Why expand /completion with URL handling? OpenAI itself tags /completions as legacy in their docs (last update July 2023), the webui doesn't use it anymore, and multimodality didn't exist when that endpoint was created.

Reference: https://platform.openai.com/docs/guides/completions
"The completions API endpoint received its final update in July 2023 and has a different interface than the new Chat Completions endpoint."

Wouldn't it be cleaner to keep /completion minimal (KISS, low level, debugging) and route multimodal users to /v1/chat/completions? If base64 overhead ever becomes a real issue, a binary upload could be added there.

ngxson:
Same thoughts as @ServeurpersoCom , there are 3 things that I don't quite like about the current approach:
1. Accepting both raw base64 and URI is a pretty bad API design. Either accept only one of the 2, or having a dedicated `type` field like OAI-compat API
2. The `multimodal_data` is a already a hack and we should not make it even more hacky
3. The current PR changes too much code for a simple feature

cetarthoriphros:
@ServeurpersoCom 
> Why expand /completion with URL handling? OpenAI itself tags /completions as legacy in their docs (last update July 2023), the webui doesn't use it anymore, and multimodality didn't exist when that endpoint was created.

From [server/README.md](https://github.com/ggml-org/llama.cpp/blob/1e5ad35d560b90a8ac447d149c8f8447ae1fcaa0/tools/server/README.md?plain=1#L411)

>This endpoint is not OAI-compatible. For OAI-compatible client, use /v1/completions instead.

But really I just find working with /completion much more intuitive, and when there parsing errors, it's on me to fix my code.

> Wouldn't it be cleaner to keep /completion minimal (KISS, low level, debugging) and route multimodal users to /v1/chat/completions? If base64 overhead ever becomes a real issue, a binary upload could be added there.

It's not only the base64-enc/dec, but the network transfer as well when file:// is not available. A set of 5 images is already 15MB, make it 20MB with the base64 encoding overhead, and shoving all of that though a curl command on a pipe to be fetched through a vimscript, on every completion/probs request. It's a collection of factors. As for the chat completion endpoints, I prefer simply getting tokens from llama-server and parse the text myself if needed. It's more intuitive. I see how much effort it takes to keep all models/templates/clients/agents happy through the commits and I think a lot of that work should be on the client, not the server.

@ngxson 

> * Accepting both raw base64 and URI is a pretty bad API design. Either accept only one of the 2, or having a dedicated `type` field like OAI-compat API

Agreed. I only kept raw base64 to maintain compatibility. I'd be happy to remove it.

> * The `multimodal_data` is a already a hack and we should not make it even more hacky

Agreed. Removing raw base64 would make it less so. Then it'd simply be a list of URIs, and multimodal_data could be renamed to multimodal_uri, with some transitional compatibility between them for a while if needed.

> * The current PR changes too much code for a simple feature

~80% of the change is moving the function higher in the file and a little formatting. Overall it's less than 20loc to accept multiple URI schemes for the completion endpoint. I could forward-declare handle_media() to make the patch smaller, but I don't like that. tokenize_input_subprompt() needs handle_media().

But all of that may be moot, as the PR is closed. @ngxson, I really wish I could have had that feedback earlier to fix the PR. Would you accept a PR that looks like this?:

* Add uri compatibility for `/completions`. By moving `handle_media()` or having a separate function specific to `/completion`. As you wish.
* Change `multimodal_data` to `multimodal_uri`, and remove raw base64 decoding, addressing your first and second point.
* (Up to you) Add temporary compatibility between `multimodal_data` -> `multimodal_uri`. Display a warning/log suggesting to upgrade clients to `multimodal_uri`.

PR #21785, opened 2026-04-12
chat: dedicated DeepSeek v3.2 parser + "official" template

pwilkin:
## Overview

Adds an "official" (tested with the official Python reference) DeepSeek v3.2 template + parser with tests.

## Additional information

The parser will only work with this template, so please use them together.

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: Yes, I supervised the AI until it got stuff right

PR #16932, opened 2025-11-02
common: Generalized XML-style tool-call parsing with streaming support (GLM 4.5/4.6 + MiniMax M2 + SeedOSS + Kimi-K2 + Qwen3-Coder + Apriel-1.5 + Xiaomi-MiMo)

hksdpc255:
Generalized and streaming-capable XML-style tool-call parsing with grammar enforcement and automatic template fixing.

Based on PR #15904, this patch introduces a generalized implementation for almost all XML-style tool-call formats.

### Supported models

- GLM 4.5/4.6
- MiniMax M2
- SeedOSS
- Kimi-K2 \(Thinking and non-thinking\)
- Qwen3-Coder \(Thinking and non-thinking\)
- Apriel-1.5
- Xiaomi-MiMo

### Grammar-constrained tool-call outputs

Tool-call messages generated by the model are now strictly validated against a defined grammar.
A new automatic grammar generator simplifies the process of creating grammars for new models.
This ensures that all tool-call outputs are well-formed, structurally consistent, and reliably parsed.

### Streaming support for tool-call parsing

The parser now supports streaming parsing, enabling incremental processing of tool-call messages as they are generated.
This enhancement improves responsiveness and allows real-time interaction during model inference.

### Automatic chat-template fixing

A lightweight Jinja2-based patcher has been added to automatically fix official chat templates before use.
With this change, official templates now work out of the box, eliminating the need for custom modifications.

### In-context reasoning

The parser now supports multiple reasoning blocks within a single generation, even when interleaved with tool calls.
All reasoning content is preserved. No information is lost during parsing or streaming.

### Enhanced unit tests

Add unit test for streaming-mode parser. It simulates the generation phase by feeding content character-by-character, comparing the parsed results and verifying that streaming and non-streaming modes reach the same final state.

### Additional Notes

- All unit tests have passed.
- Community testing is welcome! Please try it out with your model integrations.
- If your OpenAI-compatible client does **not** support sending reasoning_content back to the server, use the option `--reasoning-format none`
- When reporting issues, it’s recommended to add `-lv 1` in the command line to enable more detailed logging.

## Please use the chat template included in this PR, or any other chat template that you are certain will work correctly

PR #22106, opened 2026-04-19
chat : add MiniMax M2 specialized tool-call handler

doctorjei:
Fixes autoparser response (failure/abort)  to Minimax-M2.7 system prompt.

## Overview

Adds specialized tool-call handler for MiniMax M2.7 template (and probably later versions, I'm guessing). Right now, without it, M2.7 output with tools crashes llama-server (`GGML_ABORT`) at `src/llama-grammar.cpp:1435` (EOG with non-empty stack) when `<invoke>` is emitted.

### Why? (Reproducing the Issue)

Reproducible in-tree via `tests/test-chat.cpp` on current master (`b8840`). Parallel **`<invoke>`** elements inside a **`<minimax:tool_call>`** wrapper confuse the `peg_tester`.

The autoparser (`peg-native`) infers grammar structure from the template via differential rendering. MiniMax's template uses XML with repeatable ***`invoke`*** elements for parallel calls. The parser correctly infers per-invoke structure but mis-specifies the repetition rule, so any second ***`invoke`*** is lost.

This is a regression; an earlier working version was in the mainline (#16932, `1920345`) (via a generalized XML tool-call parser), but the autoparser refactoring (#18675) replaced it. This PR restores specialized handling for MiniMax M2.7 (and likely other M2 versions) without reverting the broader refactor.

### Implementation

This implementation follows the Kimi K2 / DeepSeek V3.2 pattern for templates the autoparser cannot handle.

- `common_chat_params_init_minimax` prepares PEG for wrapper/invoke/param grammar (parallel calls).
- Reasoning is extracted (`<think>…</think>` blocks) ahead of tool calls.
- String-typed parameters are captured verbatim (`tool_arg_string_value`) to preserve embedded XML-style content; non-strings are reconstructed through JSON.
- Dispatch in `common_chat_try_specialized_template` requires three MiniMax-specific literals in template source (`<minimax:tool_call>`, `<invoke name=`, `<parameter name=`)

### Testing

Extends the existing MiniMax block in `tests/test-chat.cpp` with five test cases.

| Test Case | Purpose | Master (no fix) |
| --- | --- | --- |
| Parallel `<invoke>` elements; two different tools | Reproduces crash pattern | **Crashes** (`Invalid diff: now finding less tool calls!`) |
| Parallel `<invoke>` elements; same tool twice | Additional variant of crash |**Crashes** (same) |
| String parameter with embedded `<div><script>…</script></div>` | Verifies `tool_arg_string_value` is verbatim | Passes |
| Multi-line string parameter (Python code with `\n`) | Verifies `until("</parameter>")` boundary on multi-line content | Passes |
| Two integer parameters in one `<invoke>` | Verifies `zero_or_more` over parameter list + non-string JSON reconstruction | Passes |

The passing test cases are also focused on repetition (vs content shape) to provide additional regression coverage.

## Additional information

- Precedent: #21785 (DeepSeek V3.2 dedicated parser)
- Earlier working implementation: #16932 (superseded by #18675)
- Crash assertion: `src/llama-grammar.cpp:1435` (`GGML_ABORT("fatal error")` when EOG token is accepted with non-empty grammar stacks).

# Requirements
- AI usage disclosure: YES

AI was used to identify the appropriate strategy, draft a harness, and draft initial code snippets. Every line was reviewed, edited as appropriate, and included manually in commits.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -2041,6 +2041,130 @@ static void func_args_not_string(json & messages) {
 
 }
 
+// MiniMax M2 format: uses <minimax:tool_call>...<invoke name="tool_name"><parameter name="key">value</parameter>...</invoke>...</minimax:tool_call>
+// - Reasoning: <think>{reasoning}</think> (optional)
+static common_chat_params common_chat_params_init_minimax(const common_chat_template &    tmpl,
+                                                           const autoparser::generation_params & inputs) {
+    common_chat_params data;
+
+    data.prompt             = common_chat_template_direct_apply_impl(tmpl, inputs);
+    data.format             = COMMON_CHAT_FORMAT_PEG_NATIVE;
+    data.supports_thinking  = true;
+    data.thinking_start_tag = "<think>";
+    data.thinking_end_tag   = "</think>";
+    data.preserved_tokens   = {
+        "<minimax:tool_call>",
+        "</minimax:tool_call>",
+        "<think>",
+        "</think>",
+    };
+
+    auto has_tools         = inputs.tools.is_array() && !inputs.tools.empty();
+    auto extract_reasoning = inputs.reasoning_format != COMMON_REASONING_FORMAT_NONE;
+    auto include_grammar   = has_tools && inputs.tool_choice != COMMON_CHAT_TOOL_CHOICE_NONE;
+
+    const std::string TC_START   = "<minimax:tool_call>";
+    const std::string TC_END     = "</minimax:tool_call>";
+    const std::string THINK_START = "<think>";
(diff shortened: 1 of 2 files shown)


Svagtlys:
(Please do remove if not allowed)

Just wanted to say that I built binaries from doctorjei:minimax-pr after rebasing on ggml-org:master, and it successfully resolved my issues with MiniMax tool calling. Appreciate your work on this, @doctorjei :)

pwilkin:
Minimax 2.7 should work correctly on current master now.

doctorjei:
@pwilkin - ok! I'll close the PR.

PR #22111, opened 2026-04-19
rpc: rdma: default max_inline_data to 0 for portability

Laitaps:
Non-Mellanox RoCE providers (notably Intel E810's `irdma` driver) reject `ibv_create_qp` requests where `qia.cap.max_inline_data > 0`. The current value of `256` works on Mellanox/ConnectX NICs but prevents the RPC RDMA transport from initializing on Intel E810 (and presumably other non-Mellanox RoCEv2 implementations).

Setting `max_inline_data` to `0` is universally accepted. Providers that support inline data simply don't use the fast path for sends, which has negligible perf impact for the RPC payloads here — all tensor transfers go via `RDMA_WRITE` and are far above any practical inline threshold.

## Tested on

- **Intel E810-XXV 25 GbE SFP28** (kernel `irdma` driver, RoCEv2) between two AMD workstations
  - Before: `llama-server --rpc` aborts at startup with `ibv_create_qp failed`
  - After: RDMA transport negotiates and runs to completion; verified end-to-end inference over RoCEv2 with Nemotron-3-Super and MiniMax-M2 models

## Diff

Single line in `ggml/src/ggml-rpc/transport.cpp`:

```diff
-    qia.cap.max_inline_data = 256;
+    qia.cap.max_inline_data = 0;  // 0 is portable: non-Mellanox RoCE (e.g. Intel E810 irdma) rejects inline > 0
```

- AI usage disclosure: used to draft the commit message and this PR body; the patch itself came from hands-on Intel E810 hardware debugging

### DIFF
--- a/ggml/src/ggml-rpc/transport.cpp
+++ b/ggml/src/ggml-rpc/transport.cpp
@@ -287,7 +287,7 @@ bool socket_t::impl::rdma_probe() {
     qia.cap.max_recv_wr     = RDMA_RX_DEPTH + 4;
     qia.cap.max_send_sge    = 1;
     qia.cap.max_recv_sge    = 1;
-    qia.cap.max_inline_data = 256;
+    qia.cap.max_inline_data = 0;  // 0 is portable: non-Mellanox RoCE (e.g. Intel E810 irdma) rejects inline > 0
 
     rdma->qp = ibv_create_qp(rdma->pd, &qia);
     if (!rdma->qp) return false;


rgerganov:
On DGX Spark I get the following results:

Requesting `max_inline_data=256` (master):
```
RDMA probed: dev=rocep1s0f0 gid=3 RoCEv2 qpn=1806 inline=316
```

Requesting `max_inline_data=0` (this PR):
```
RDMA probed: dev=rocep1s0f0 gid=3 RoCEv2 qpn=1957 inline=60
```

I wonder if we should try first with 256 and then retry with 0 on error or keep it simple and always go with 0. @dvv101111 @Mithras any opinions?

dvv101111:
Seems actual hardware limit for the Intel irdma driver is 101 -https://github.com/torvalds/linux/blob/master/drivers/infiniband/hw/irdma/user.h#L189
Thats why it failed for you.

But setting this to 0 on Intel also leads to max_inline_data = 0, and all RPC messages would go through the slower memcpy path. @Laitaps

So better choice is 32. 
With request=32, Mellanox still returns 60, which covers current protocol implementation without regression, and Intel will approve and return 32 size and will be able to use inline transfers

Code for searching maximum possible inline size does not make sense for now. 
I think future bandwidth optimization for model transfer will be better to do via rdma_write, currently it using rdma_send + rdma_recv. There is no rdma_write in code right now.

rgerganov:
@Laitaps Could you try with 32 and share the RDMA logs?

PR #20070, opened 2026-03-03
kleidiai : support for concurrent sme and neon kernel execution

chaxu01:
This PR introduces concurrent execution of SME and NEON kernels in the KleidiAI backend. Instead of running SME-only kernels when enabled, the runtime can now split work between SME and NEON execution paths, allowing better scaling when SME units are limited.

Llama-3.2-1B-Instruct-Q4_0 — pp512 (tokens/sec) (Mac M4 Pro, SME enabled)
| Threads | w/o hybrid | w/ hybrid |
| ------- | ---------------- | -------------- |
| 1       | 295.64 ± 0.22    | 294.66 ± 0.79  |
| 2       | 547.14 ± 7.12    | 550.86 ± 1.21  |
| 4       | 483.73 ± 4.13    | 725.07 ± 1.53  |
| 6       | 424.92 ± 8.73    | 912.23 ± 2.44  |
| 8       | 556.80 ± 3.65    | 1048.19 ± 9.95 |

PR #22113, opened 2026-04-19
ggml: kleidi, cpu: beginnings of macOS cluster scheduling

mediouni-m:
## Overview

macOS has some APIs for cluster scheduling that we can use on arm64.

This follows up https://github.com/ggml-org/llama.cpp/pull/20070
 
Performance is currently a problem and the thread pool setup is at play. So putting the actual QoS call behind a `#if 0` as it causes a performance slowdown due to having cores of multiple performance tiers at play. Notably due to involving the ecore AMX/SME unit more.

Or... the other option is to switch task dispatching to GCD and use dispatch_apply_attr_set_parallelism...

## Additional information

The APIs do work on iOS but aren't sure whether they're allowed for App Store submission so restricted it to macOS just in case.

# Requirements

- AI usage disclosure: 

 N/A

### DIFF
--- a/ggml/include/macos-scheduling.h
+++ b/ggml/include/macos-scheduling.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include "ggml.h"
+#include "ggml-backend.h"
+
+#ifdef  __cplusplus
+extern "C" {
+#endif
+
+#define CLUSTER_SCHEDULING_AVAILABLE
+int pthread_prefer_alternate_cluster_self(void);
+int pthread_prefer_alternate_amx_self(void);
+
+#define PTHREAD_MAX_PARALLELISM_PHYSICAL 0x1
+#define PTHREAD_MAX_PARALLELISM_CLUSTER 0x2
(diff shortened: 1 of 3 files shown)


mediouni-m:
```
#elif defined(__APPLE__) && defined(__MACH__)
    int32_t num_physical_cores;
    size_t len = sizeof(num_physical_cores);
    int result = sysctlbyname("hw.perflevel0.physicalcpu", &num_physical_cores, &len, NULL, 0);
    if (result == 0) {
        return num_physical_cores;
    }
    result = sysctlbyname("hw.physicalcpu", &num_physical_cores, &len, NULL, 0);
    if (result == 0) {
        return num_physical_cores;
    }
```

Oh. The thread pool is broken with big.LITTLE in mind. That has to be addressed in some fashion before the change can be flipped on.

As far as I can tell this leaves the 12 mCores on an M5 Pro/Max idle too.

Shouldn't block this PR though (or at least half of it)

PR #22119, opened 2026-04-19
sycl: Improve mul_mat_id memory efficiency and add BF16 fast path

qnixsynapse:
## Overview

This PR addresses memory exhaustion issues (`UR_RESULT_ERROR_OUT_OF_HOST_MEMORY`) encountered on SYCL Level Zero when handling large-vocabulary models and MoE architectures.

**Key Changes:**

1. **BF16 Fast Path via DNNL:**
   - Implemented a `bf16 type_a mulmat -> f32` path using DNNL.
   - **Problem:** Previously, BF16 weights are type casted to F32. This forced a full precision leading to multi-GB allocations and OOM crashes especially on A750.
   - **Solution:** The new path utilizes BF16 storage in place and only requires a small conversion buffer for `src1`, significantly reducing the memory footprint while maintaining F32 accumulation accuracy.

2. **Optimized Staging Buffers for `mul_mat_id`:**
   - **Problem:** `src1_contiguous` and `dst_contiguous` buffers were sized to the total number of elements. In MoE models (especially when `ne12 > 1`), this resulted in significant over-allocation.
   - **Solution:** Buffers are now sized based on the actual number of routed rows (`ids->ne[1] * n_ids`), preventing OOM failures during MoE operations(referenced from a different backend). 

test-backend-ops is passing with this change.

No performance changes if the GPU doesn't support BF16 or the src0 weight tensor isn't in BF16.

# Requirements

- AI usage disclosure: YES (for writing commit messages)

### DIFF
--- a/ggml/src/ggml-sycl/common.hpp
+++ b/ggml/src/ggml-sycl/common.hpp
@@ -28,6 +28,13 @@
 
 namespace syclexp = sycl::ext::oneapi::experimental;
 
+#if defined(__INTEL_LLVM_COMPILER) && __has_include(<sycl/ext/oneapi/bfloat16.hpp>)
+    #include <sycl/ext/oneapi/bfloat16.hpp>
+    #ifndef GGML_SYCL_HAS_BF16
+        #define GGML_SYCL_HAS_BF16
+    #endif
+#endif
+
 #if GGML_SYCL_DNNL
 #include "dnnl.hpp"
 #include "dnnl_sycl.hpp"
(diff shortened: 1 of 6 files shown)


arthw (approved):
It's good job!

Thank you!

qnixsynapse:
The failing test seems unrelated to the changes in this PR.

arthw:
@ggerganov 
Please review it!
The failed CI is always fault, and has nothing with this PR.

Thank you!

CISC (approved):

PR #9359, opened 2024-09-07
llama : sanitize tokens in the upper bound

slaren:
cont #9357

PR #22121, opened 2026-04-19
Make `llama_token_to_piece` return `INT32_MIN` for invalid tokens

julmb:
## Overview

The `llama_token_to_piece` function currently throws `std::out_of_range` for invalid token inputs. This is supposed to be a C interface, but this exception cannot be caught from C. It also does not look like this was a deliberate choice, since it happens on a `std::vector` access done indirectly via `token_get_attr`. Furthermore, `llama_token_to_piece` has a bounds check further down, but this can never trigger since it already throws on the access in `token_get_attr`.

This PR moves the existing bounds check to the very beginning of `llama_token_to_piece` which logs the error and returns `INT32_MIN` if the token is invalid. This is in analogy to another unrecoverable error condition in `llama_tokenize`. I clarified the existing docstring and added documentation for the new behavior.

## Additional information

There is some previous discussion in https://github.com/ggml-org/llama.cpp/pull/9359.

# Requirements

- AI usage disclosure: YES, exclusively for exploring the code base

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -1687,6 +1687,9 @@ std::string common_detokenize(const struct llama_vocab * vocab, const std::vecto
     std::string text;
     text.resize(std::max(text.capacity(), tokens.size()));
     int32_t n_chars = llama_detokenize(vocab, tokens.data(), (int32_t)tokens.size(), &text[0], (int32_t)text.size(), false, special);
+    if (n_chars == std::numeric_limits<int32_t>::min()) {
+        throw std::runtime_error("Detokenization failed: some supplied token is invalid");
+    }
     if (n_chars < 0) {
         text.resize(-n_chars);
         n_chars = llama_detokenize(vocab, tokens.data(), (int32_t)tokens.size(), &text[0], (int32_t)text.size(), false, special);
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -1662,6 +1662,9 @@ std::string common_token_to_piece(const struct llama_vocab * vocab, llama_token 
     std::string piece;
     piece.resize(piece.capacity());  // using string internal cache, 15 bytes + '\n'
     const int n_chars = llama_token_to_piece(vocab, token, &piece[0], piece.size(), 0, special);
+    if (n_chars == std::numeric_limits<int32_t>::min()) {
+        throw std::runtime_error("Token to piece failed: supplied token is invalid");
+    }
     if (n_chars < 0) {
         piece.resize(-n_chars);
         int check = llama_token_to_piece(vocab, token, &piece[0], piece.size(), 0, special);
--- a/common/speculative.cpp
+++ b/common/speculative.cpp
@@ -14,6 +14,7 @@
 #include <cassert>
 #include <cstring>
 #include <iomanip>
+#include <limits>
 #include <map>
 #include <cinttypes>
 
--- a/examples/batched.swift/Sources/main.swift
+++ b/examples/batched.swift/Sources/main.swift
@@ -224,6 +224,9 @@ private func tokenize(text: String, add_bos: Bool) -> [llama_token] {
 private func token_to_piece(token: llama_token, buffer: inout [CChar]) -> String? {
     var result = [CChar](repeating: 0, count: 8)
     let nTokens = llama_token_to_piece(vocab, token, &result, Int32(result.count), 0, false)
+    if nTokens == Int32.min {
+        return nil
+    }
     if nTokens < 0 {
         let actualTokensCount = -Int(nTokens)
         result = .init(repeating: 0, count: actualTokensCount)
--- a/examples/batched.swift/Sources/main.swift
+++ b/examples/batched.swift/Sources/main.swift
@@ -225,7 +225,8 @@ private func token_to_piece(token: llama_token, buffer: inout [CChar]) -> String
     var result = [CChar](repeating: 0, count: 8)
     let nTokens = llama_token_to_piece(vocab, token, &result, Int32(result.count), 0, false)
     if nTokens == Int32.min {
-        return nil
+        print("llama_token_to_piece() failed")
+        exit(1)
     }
     if nTokens < 0 {
         let actualTokensCount = -Int(nTokens)
--- a/examples/llama.swiftui/llama.cpp.swift/LibLlama.swift
+++ b/examples/llama.swiftui/llama.cpp.swift/LibLlama.swift
@@ -319,7 +319,9 @@ actor LlamaContext {
             result.deallocate()
         }
         let nTokens = llama_token_to_piece(vocab, token, result, 8, 0, false)
-
+        if nTokens == Int32.min {
+            return []
+        }
         if nTokens < 0 {
             let newResult = UnsafeMutablePointer<Int8>.allocate(capacity: Int(-nTokens))
             newResult.initialize(repeating: Int8(0), count: Int(-nTokens))
--- a/include/llama.h
+++ b/include/llama.h
@@ -1150,6 +1150,7 @@ extern "C" {
     /// @param text The char pointer must be large enough to hold the resulting text.
     /// @return Returns the number of chars/bytes on success, no more than text_len_max.
     /// @return Returns a negative number on failure - the number of chars/bytes that would have been returned.
+    /// @return Returns INT32_MIN if any of the tokens is not in the vocabulary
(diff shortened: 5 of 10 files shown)


julmb:
Also, the diff looks like a mess but it's really just an indentation change on a large block that is no longer conditional.

CISC (approved):

CISC:
> Also, the diff looks like a mess but it's really just an indentation change on a large block that is no longer conditional.

No worries, looks just fine if you hide whitespace changes.

ngxson:
This change will make places like this become unsafe:

https://github.com/ggml-org/llama.cpp/blob/4eac5b45095a4e8a1ff1cce4f6d030e0872fb4ad/common/common.cpp#L1572-L1586

I think it's required to also update all the downstream code in the same PR, for security reason. Throwing an exception is OK.

CISC:
> This change will make places like this become unsafe:

Oooh, nice catch.

ddh0:
Just curious - why not use the existing `LLAMA_TOKEN_NULL` (`-1`) in this case?

CISC:
> Just curious - why not use the existing `LLAMA_TOKEN_NULL` (`-1`) in this case?

Well, first off the function does not return a token, secondly that would just suggest the buffer should be 1 byte according to the API contract.

julmb:
I found the following downstream uses of `llama_token_to_piece` and changed them where needed:
- `common/common.cpp:common_token_to_piece`: throw `std::runtime_error` on `INT32_MIN` result as in `common_tokenize`
- `tools/mtmd/mtmd.cpp:token_to_piece`: throw `std::runtime_error` on `INT32_MIN` result
- `tests/test-backend-sampler.cpp:token_to_piece`: throw `std::runtime_error` on `INT32_MIN` result
- `examples/debug/debug.cpp:print_tokenized_prompt`: already handles all negative results uniformly as error
- `examples/diffusion/diffusion-cli.cpp:diffusion_step_callback`: already handles all negative results uniformly by ignoring
- `examples/simple/simple.cpp:main`: already handles all negative results uniformly as error
- `examples/simple-chat/simple-chat.cpp:main`: already handles all negative results uniformly as error
- `examples/batched.swift/Sources/main.swift:token_to_piece`: return `nil` on `INT32_MIN`
- `examples/llama.swiftui/llama.cpp.swift/LibLlama.swift:token_to_piece`: return `[]` on `INT32_MIN`

Notes:
- The functions`tools/mtmd/mtmd.cpp:token_to_piece` and `tests/test-backend-sampler.cpp:token_to_piece` are exact copies of `common_token_to_piece`. Is there a reason they cannot use `common_token_to_piece`?
- I have no clue about Swift. It seemed like most errors in those two examples are handled simply by printing something to stdout and then continuing. The new `token_to_piece` already does this via `LLAMA_LOG_ERROR` so I simply made these functions return `nil` and `[]`, respectively. I used an LLM to generate the syntax and I don't even know if I can compile this on my Linux system, so it would be good if someone could look at these. The changes are trivial though.

julmb:
It has been a few days. As far as I can tell, all issues with this PR should have been addressed. Is there anything I still need to or can do to help move this forward?

julmb:
I am super sorry, there is no excuse for me not building locally before pushing.

I added the missing `#include <limits>` to `mtmd.cpp` and `test-backend-sampler.cpp`, local builds work now.

The `common.cpp` file does not import `limits` directly, but it already used `std::numeric_limits` before this PR, so I'm assuming that it's okay to have it imported transitively.

julmb:
Alright, I think the failure of "CI (self-hosted) / ggml-ci-win-intel-vulkan" is unrelated to this PR, seeing as most recent commits share this failure.

ngxson:
There is some issues spotted via: https://github.com/ngxson/llama.cpp/pull/103

(You can ignore minor one btw)

julmb:
> There is some issues spotted via: [ngxson#103](https://github.com/ngxson/llama.cpp/pull/103)

I read through the comments.

> Clear the pending UTF-8 buffer on the invalid-token path.

Taking a second look at this, it might be more reasonable to just print an error and quit. This is also what happens if `llama_decode` returns a non-zero error code.

> Don’t collapse `Int32.min` into the normal empty-piece case.

Much of the surrounding code silently ignores errors. This is an example/proof-of-concept that does not need production-grade error handling.

> Clarify that `INT32_MIN` is a sentinel, not a resize hint.

The documentation seems reasonably-clear to me here.

> `INT32_MIN` now breaks `llama_detokenize()` on invalid input.

This is a big one, see below.

When checking for downstream uses earlier, I only looked at `llama_token_to_piece`. Unfortunately, this is already two steps downstream from the actual change in `llama_vocab::impl::token_to_piece`. Calls to this function:
- Through `llama_vocab::token_to_piece` (thin wrapper)
- Through `llama_token_to_piece` (call sites adjusted earlier)
- Through `common_token_to_piece` (adjusted earlier)
- In `llama_batch_allocr::ubatch_print` (cache overlaod)
- In `llama_grammar_apply_impl` (cache overlaod)
- In `llama_grammar_accept_impl` (cache overlaod)
- In `llama_sampler_infill_apply`: From what I can tell, `cur_p` could contain arbitrary, user-supplied tokens. However, the type of the function does not allow us to notify the caller in the case of invalid input. I added a `std::runtime_error` in case of invalid tokens. These cannot be caught when the library is used from C code, but this is no worse than what happened before this PR.
- In `llama_vocab::impl::token_to_piece_for_cache`: This is only called by `llama_vocab::impl::load`, guarded by `for (uint32_t id = 0; id < n_tokens; ++id)`, so it currently cannot fail. I still added a guard with `std::runtime_error` like in `common_token_to_piece` to make the function more self-contained and not rely on subtle preconditions. If this is considered excessive for a function that is only used internally, I can also remove it again.
- In `llama_vocab::impl::detokenize`: In hindsight, I should have realized that this also uses `token_to_piece` and also needs to handle invalid tokens. In this case, handling means propagating the error to the caller, so `llama_detokenize` also gets the `INT32_MIN` return value. I added this to the documentation of `llama_detokenize`. Of course, this means we now also have to look at all downstream uses of `llama_vocab::impl::detokenize`.

Downstream uses of `llama_vocab::impl::detokenize`:
- `llama_vocab::detokenize`: thin wrapper
- `common_detokenize`: Add `std::runtime_error` like in `common_token_to_piece`
- `llguidance.cpp:llama_sampler_llg_new_tokenizer`: Already calls `GGML_ABORT` on any negative result
- `speculative.cpp:draft`: There is already a `GGML_ASSERT` in case of nonnegative result. However, since an invalid token is a case of invalid input rather than programming error, I chose to throw a `std::runtime_error` instead. This is part of `libcommon`, so a C++ interface should be fine.
- `llama_vocab::detokenize`: Overload that uses C++ types. Add `std::runtime_error` when the `llama_vocab::detokenize`/`llama_vocab::impl::detokenize` call chain that uses plain C types returns `INT32_MIN`.
- `get_overlapping_token_sequences`: Uses C++ overload.
- `test-tokenizer-random.py`: Raise `RuntimeError` as in surrounding code.

julmb:
Notes that are related but orthogonal to this PR:

The coderabbit instance at https://github.com/ngxson/llama.cpp/pull/103 found that in `llama_sampler_infill_apply`, a negative value can be passed to `std::vector::resize`. This was already the case before this PR and should probably be a separate issue. I am surprised this never surfaced as it would try to allocate a vector of length ~2^64 and thus definitely fail. The vectors get initialized as `std::vector<char>(512)` so I'm guessing it never executed the faulty branch since all tokens were shorter than this. Should probably still be fixed, I can make another PR after this one.

The function `common_detokenize` is an exact copy of `llama_vocab::detokenize` (the overload using C++ types). I am not sure if there is an easy way to remove this redundancy or if there is a good reason for it to be the way it is.

julmb:
It has been a while, is there anything else I need to do? As far as I am concerned, this should be ready to merge.

julmb:
I rebased this onto master and resolved the merge conflicts. I also grepped for `token_to_piece(` and `detokenize(` once more to see if any new downstream uses had been added since I last worked on this.

@CISC @ngxson You both commented on this before. I know that everyone is busy with hundreds of other issues and pull requests and this one is neither exciting nor urgent, so I understand that it is not high priority. But keeping it updated also takes time and effort, so if it is unlikely that this will be merged, I would appreciate a notice. Otherwise, let me know if there is anything I can/should do.

PR #22126, opened 2026-04-19
Gbnf - literal token filtering for non-control tokens.

ego-ipse:
## Overview

Small but meaningful improvement to gbnf grammar. 

Currently a token `<text>` is matched to the token with the **exact** string `<text>`.
This works perfectly well when filtering for control tokens, but fails if the tokens do not start and end with `<` and `>`.

This PR addresses this in a way that preserves functionality with minimal changes.

Keeping in line with the notation for gbnf token values  `<[N]>`, it adds `<{text}>` to search for the exact text inside.
This preserves existing grammars that use control tokens unchanged (except for the small edge case where a control token has the initial shape  `<{`.

For these situations, this new implementation adds `\` as an escape character and accepts control characters. 

## Additional information

| You want to match…                                        | Use         |
| --------------------------------------------------------- | ----------- |
| A specific token id you already know                      | `<[N]>`     |
| A special token whose vocab text contains angle brackets  | `<text>`    |
| Any other token, by its literal text  (new)                  | `<{text}>`  |

Tested with gemma-4-26B-A4B-it-Q4_K_M and llama-cpp built with CUDA.

# Requirements

- AI usage disclosure: YES. Used to prototype code and to clean up the English and formatting in md file. All AI geenerated code was reviewed

### DIFF
--- a/grammars/README.md
+++ b/grammars/README.md
@@ -69,15 +69,71 @@ Parentheses `()` can be used to group sequences, which allows for embedding alte
 
 ## Tokens
 
-Tokens allow grammars to match specific tokenizer tokens rather than character sequences. This is useful for constraining outputs based on special tokens (like `<think>` or `</think>`).
+The terminals described so far (`"abc"`, `[a-z]`, `.`) match **character sequences** in the model's output. A *token* terminal instead matches **a single tokenizer token by id**, regardless of what bytes that token decodes to.
 
-Tokens can be specified in two ways:
+This is useful when you want to constrain the model's output in terms of its own vocabulary — for example, requiring it to emit a specific control token like `<think>`, or forbidding a particular token mid-sequence.
 
-1. **Token ID**: Use angle brackets with the token ID in square brackets: `<[token-id]>`. For example, `<[1000]>` matches the token with ID 1000.
+There are three ways to write a token terminal:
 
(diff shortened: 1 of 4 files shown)


aldehir:
Thanks for your interest in this. When adding token support, I contemplated doing something similar but decided against it. If a model ever had `<{..}>` tokens, it would be ambiguous with this approach and fail to match.

ego-ipse:
> Thanks for your interest in this. When adding token support, I contemplated doing something similar but decided against it. If a model ever had `<{..}>` tokens, it would be ambiguous with this approach and fail to match.

Hi @aldehir I've considered the issue and that's why I added the `\` escape so that `<{..}>` could be represented as `<{<{..\}>}>` which is a bit cumbersome I admit, but it did seem like an edge case situation.

What I would say is that the escaping may need to be generalized since a token  like `<..>>` or `<[..>` will currently also probably run into some issues (I'll try and test it tomorrow when I can). 

We could change it slightly, so that the double quotation mark (or even a slightly pythonic r"..") becomes the marker, but to me it seems that the core issue is that we don't have an escape character for a subsection of the gbnf metagrammar where any character is theoretically valid. 

My suggestion would be to split this into two parts. One where we add an escape character to `<> [ ] { }` (since we only need to escape boundaries) and then we implement the literal char as {} (or "" or whatever)

This would be a minor edge case since `<\thinking>` would not run into an issue since that `\t` is not escaping a control character. And for the case you very correctly described while `<{<{..\}>}>` would still be valid, we'd have the more elegant `<\{..}>` which I would argue is an acceptable compromise with minimal impact on existing implementations. It would also disambiguate the possible issues mentioned and I need to test above.

PR #22129, opened 2026-04-19
Tensor-parallel: Fix delayed AllReduce on Gemma-4 MoE

gaugarg-nv:
Skip forward past nodes that don't consume the current node, and allow a chain of MULs.

When `down_exps_s` is set, build_moe_ffn pulls the scale tensor in via reshape/repeat/get_rows. Topological sort places those between `mul_mat_id` and the MUL that consumes it, so the existing nodes[id+1] check never sees an ADD_ID or MUL and fails.

The scale MUL is followed by a second MUL; the old code only accepted one.

Performance on 2x 5090:

model | test | t/s - d5b780a67 | t/s - PR | Speed-up
-- | -- | -- | -- | --
gemma4 26B.A4B Q8_0 | pp512 | 3473.35 | 7202.7 | 2.07
gemma4 26B.A4B Q8_0 | tg128 | 164.29 | 202.45 | 1.23

# Requirements

- AI usage disclosure: Yes, took help in identifying the root cause

### DIFF
--- a/ggml/src/ggml-backend-meta.cpp
+++ b/ggml/src/ggml-backend-meta.cpp
@@ -1661,6 +1661,24 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend, 
 
                 ggml_tensor * node = cgraph->nodes[id];
                 int32_t n_used = ggml_node_get_use_count(cgraph, id);
+
+                // Skip MIRRORED nodes that don't consume node
+                auto skip_unrelated = [&]() {
+                    while (id + 1 < cgraph->n_nodes) {
+                        ggml_tensor * next = cgraph->nodes[id+1];
+                        bool uses_node = false;
+                        for (int s = 0; s < GGML_MAX_SRC; s++) {
+                            if (next->src[s] == node) { uses_node = true; break; }
+                        }
+                        if (uses_node) break;
+                        if (ggml_backend_meta_get_split_state(next, false).axis != GGML_BACKEND_SPLIT_AXIS_MIRRORED) {
+                            break;
+                        }
+                        id++;
+                    }
+                };
+
+                skip_unrelated();
                 if (id + 1 >= cgraph->n_nodes) {
                     return idr;
                 }
@@ -1675,10 +1693,12 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend, 
                         n_used = ggml_node_get_use_count(cgraph, id);
                     }
                 }
-                if (id + 1 >= cgraph->n_nodes) {
-                    return idr;
-                }
-                {
+                // Chain of MULs with MIRRORED src[1]
+                while (true) {
+                    skip_unrelated();
+                    if (id + 1 >= cgraph->n_nodes) {
+                        return idr;
+                    }
                     ggml_tensor * next = cgraph->nodes[id+1];
                     if (next->op == GGML_OP_MUL && next->src[0] == node &&
                             ggml_backend_meta_get_split_state(next->src[1], false).axis == GGML_BACKEND_SPLIT_AXIS_MIRRORED) {
@@ -1686,6 +1706,8 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend, 
                         id++;
                         idr = id;
                         n_used = ggml_node_get_use_count(cgraph, id);
+                    } else {
+                        break;
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


am17an:
What's special about the MUL mode, can't it skip over more generally looking at the src pointers?

am17an (approved):

loci-dev:
LOCI found performance insights on your pull request.<br><br>🔎 Check latest report [here](https://github.com/auroralabs-loci/llama.cpp/pull/1361#issuecomment-4277712115).

gaugarg-nv:
> What's special about the MUL mode, can't it skip over more generally looking at the src pointers?

@JohannesGaessler 's original code was trying to match the common pattern found in MOE models. I have tried to extend it to cover Gemma. 

This check can be extended to include other ops that will work with `PARTIAL` and `MIRRORED` src, like `Scale`, `Div`, `Cpy`, but won't work for ops like `Add`, or non-linear ops.

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1661,6 +1661,24 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
...
+                        for (int s = 0; s < GGML_MAX_SRC; s++) {
+                            if (next->src[s] == node) { uses_node = true; break; }
+                        }
+                        if (uses_node) break;
+                        if (ggml_backend_meta_get_split_state(next, false).axis != GGML_BACKEND_SPLIT_AXIS_MIRRORED) {
+                            break;
+                        }
+                        id++;
+                    }
+                };
```
JohannesGaessler, line 1691: It is not safe to only check whether `cgraph->nodes[i]` is `MIRRORED`. This needs a further check to assert that also
 
1. all of the sources are `MIRRORED` or 
2. that the op is not a matrix multiplication and is commutative with the AllReduce.

JohannesGaessler: Actually, wait, I think I may have made a mistake.

JohannesGaessler: I think my first comment was correct after all. If everything is `MIRRORED` then it's always safe. For that we also need to also check the sources because `ggml_backend_meta_get_split_state` assumes that the sources have already been reduced. Otherwise we need to assert that the op does not change the split state (which only a matrix multiplication can) and that executing it before the AllReduce is mathematically correct (like it is for a multiplication).

gaugarg-nv: Thanks @JohannesGaessler . You're right.
I have gone ahead with option 1 and made sure that optimization still works for different MOE models.

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1661,6 +1661,35 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
...
+
+                // Skip MIRRORED nodes that don't consume node
+                auto skip_unrelated = [&]() {
+                    while (id + 1 < cgraph->n_nodes) {
+                        ggml_tensor * next = cgraph->nodes[id+1];
+                        bool uses_node = false;
+                        for (int s = 0; s < GGML_MAX_SRC; s++) {
+                            if (next->src[s] == node) { uses_node = true; break; }
+                        }
+                        if (uses_node) break;
```
JohannesGaessler: I think it makes more sense to move this downward and to have one loop over the sources. Also please consistently use this format:

```
if (...) {
    ...;
}
```

ggml/src/ggml-backend-meta.cpp
```diff
@@ -1675,17 +1704,21 @@ static enum ggml_status ggml_backend_meta_graph_compute(ggml_backend_t backend,
...
                     if (next->op == GGML_OP_MUL && next->src[0] == node &&
                             ggml_backend_meta_get_split_state(next->src[1], false).axis == GGML_BACKEND_SPLIT_AXIS_MIRRORED) {
                         node = next;
                         id++;
                         idr = id;
                         n_used = ggml_node_get_use_count(cgraph, id);
+                    } else {
+                        break;
                     }
                 }
```
JohannesGaessler, line 1724: This code is effectively a roundabout way to check condition 2 that I outlined previously. The PR is fine like this, but I'll maybe refactor and simplify this in the future when I touch this part of the code again.

JohannesGaessler (approved):

ggml/src/ggml-backend-meta.cpp
gaugarg-nv: Addressed now.

JohannesGaessler (approved):

PR #22066, opened 2026-04-17
sycl: Battlemage (BMG) optimizations — AOT, Q5_K reorder, PAD stride fix, new ops, oneMKL routing

aicss-genai:
## Overview

A set of four independent SYCL improvements developed on Intel® Arc™ (Xe2-HPG / Battlemage) GPU, e.g. B70, B60, B50, PVC, B580, A770. Together they fix AOT compilation for BMG targets, add missing op support, correct a stride bug in PAD, and improve small-matmul dispatch.

### Authors

- @unrahul
- @ctao456
- @malsbat
- @YuningQiu
- @vvravi1
- @zhiqitao-intel

### Commits

1. **sycl: Battlemage AOT + reorder MMVQ/dequant + async mem-op**
   - Fix AOT (ahead-of-time) compilation for Battlemage by switching to `-fsycl-targets=spir64_gen` with `-device` instead of `--offload-arch`. The existing `--offload-arch` flag silently fell back to JIT on newer GPUs.
   - Skip the `-ze-intel-greater-than-4GB-buffer-required` linker flag when targeting `spir64_gen` (incompatible with offline compilation).
   - Add Q5_K reorder support for MMVQ and dequantize paths (reorder kernel, vec_dot, dequantize).
   - Decouple `g_ggml_sycl_use_async_mem_op` from the graph flag so async USM alloc/free can be used on the non-graph reorder staging path (controlled via `GGML_SYCL_USE_ASYNC_MEM_OP` env var, default on).

2. **sycl: support non-contiguous input in PAD op**
   - Replace the existing `ggml_is_contiguous(src0)` assertion with proper strided addressing (`nb0/nb1/nb2/nb3`), so PAD works on views produced by reshape / permute without a preceding contiguous copy.

3. **sycl: add FILL, CUMSUM, DIAG, SOLVE_TRI, SSM_SCAN, GATED_DELTA_NET**
   - Implement six missing SYCL ops required by Mamba / SSM and newer model architectures:
     - `GGML_OP_SET_ROWS_FILL` — fill tensor with a constant
     - `GGML_OP_CUMSUM` — inclusive prefix-sum (work-group scan + fixup)
     - `GGML_OP_DIAG` — extract / create diagonal
     - `GGML_OP_SOLVE_TRI` — triangular solve (column-major forward substitution)
     - `GGML_OP_SSM_SCAN` — selective state-space scan (Mamba)
     - `GGML_OP_GATED_DELTA_NET` — already present; add missing include guard
[...]

PR #22147, opened 2026-04-20
sycl: Battlemage AOT build via spir64_gen + MMQ subgroup annotations

aicss-genai:
### Authors

- @unrahul
- @ctao456
- @malsbat
- @YuningQiu
- @vvravi1
- @zhiqitao-intel

Enables AOT builds for Intel GPUs (validated on Intel® Arc™ Pro B70, BMG-G31, Xe2-HPG):

- When `GGML_SYCL_DEVICE_ARCH` is set, switch to `-fsycl-targets=spir64_gen` with `-Xsycl-target-backend="-device <arch>"` and skip `-ze-intel-greater-than-4GB-buffer-required` (not accepted by the AOT path). Behavior is unchanged when `GGML_SYCL_DEVICE_ARCH` is unset.
- Adds `[[intel::reqd_sub_group_size(WARP_SIZE)]]` to MMQ Q4_0/Q4_1/Q5_0/Q5_1 kernel launches. `WARP_SIZE=16` on Intel targets; pinning the required subgroup size is required for spir64_gen AOT correctness and documents intent on JIT.

Also adds a documentation-only update to `ggml_sycl_supports_mmq` (still returns false) and a note in `ggml_sycl_op_mul_mat_sycl` recording that fused dequant+GEMM and MMQ/DPAS were both slower than dequant+oneDNN in our experiments.

## Additional information

Split from #22066 per reviewer request for independent review.

# Requirements

- AI usage disclosure: Yes. This work was partially produced with an agentic engineering approach: agents surface issues and explore experiments while engineers identify and reject candidates using domain knowledge. Human feedback involved.


ggml/src/ggml-sycl/CMakeLists.txt
```diff
@@ -135,7 +135,11 @@ endif()
 
 if (GGML_SYCL_TARGET STREQUAL "INTEL")
     add_compile_definitions(GGML_SYCL_WARP_SIZE=16)
-    target_link_options(ggml-sycl PRIVATE  -Xs   -ze-intel-greater-than-4GB-buffer-required)
+    if (NOT GGML_SYCL_DEVICE_ARCH)
+        target_link_options(ggml-sycl PRIVATE -Xs -ze-intel-greater-than-4GB-buffer-required)
```
arthw, line 139: "-ze-intel-greater-than-4GB-buffer-required" is used to support malloc memory more than 4GB.
It's mandatory for some LLMs support.

If remove it, please provide the alternative solution for 4GB in same PR.

ggml/src/ggml-sycl/ggml-sycl.cpp
```diff
@@ -2178,6 +2178,8 @@ inline void ggml_sycl_op_mul_mat_sycl(
 #endif
     if ((src0->type == GGML_TYPE_F16 || ggml_is_quantized(src0->type)) && use_fp16 && ggml_is_contiguous(src0) &&
         row_diff == src0->ne[1] && dst->op_params[0] == GGML_PREC_DEFAULT) {
+        // NOTE: Fused dequant+GEMM and MMQ/DPAS were both attempted (Steps 10-11
```
arthw, line 2181: remove these lines.
It has no value to user/developer.

ggml/src/ggml-sycl/ggml-sycl.cpp
```diff
@@ -3261,9 +3263,12 @@ enum class mul_mat_algo {
 };
 
 inline bool ggml_sycl_supports_mmq(enum ggml_type type) {
-    // TODO: accuracy issues in MMQ
-    GGML_UNUSED(type);
-    return false;
+    // DPAS INT8 MMQ kernel exists in mmq.cpp but is slower than dequant+oneDNN.
+    // Disabled pending further optimization. See optimization-workbook.md Step 11.
```
arthw, line 3267: This code doesn't change any behavior.
remove them.

arthw (changes requested):
1.
Support malloc more than 4GB is mandatory for both AOT or not AOT.
If can't provide the alternative for 4GB, we like to disable AOT which is not more important than 4GB issue.

2.
What's the benefit after set sub-group size to 16？ Any test result?

ggml/src/ggml-sycl/CMakeLists.txt
YuningQiu: Hi Neo, thanks a lot for this input. We will fix it asap. This PR add the AOT support for bmg-g31, while the current SYCL --offload-arch only has AOT support up to bmg-g21.

NeoZhangJianyu: Set single AOT flag will make the binary can't run for other GPUs - The binary file with AOT can't trigger JIT on other GPUs.
For example, the SYCL code for AOT bmg-31 can't run on iGPU and other dGPU.

I suggest not set AOT in SYCL backend cmakefile as default.
You could update it in SYC.md to guide user if needed.

Thank you!

ggml/src/ggml-sycl/ggml-sycl.cpp
malsbat: Removed

ggml/src/ggml-sycl/ggml-sycl.cpp
malsbat: Removed

ggml/src/ggml-sycl/CMakeLists.txt
malsbat: @arthw, Can you provide steps to recreate the `ggml_backend_sycl_buffer_type_alloc_buffer: can't allocate 5000000000 Bytes of memory on device` error (I'm assuming this is the reason for the 4GB option)?  I was unable to reproduce the error in the JIT build even after removing -ze-intel-greater-than-4GB-buffer-required and setting UR_L0_ENABLE_RELAXED_ALLOCATION_LIMITS=0.

I'd like to reproduce the error conditions to confirm if the flag is necessary or not in AOT.  

Side note: I tried with the latest 2025.3.3 compiler, the -ze-intel-greater-than-4GB-buffer-required option is not recognized by ocloc.

NeoZhangJianyu: “-ze-intel-greater-than-4GB-buffer-required” is not necessary for AOT.
I will test and review this PR as soon.

Thank you!

NeoZhangJianyu:
What's the benefit after set sub-group size to 16？ Any test result?

malsbat:
> What's the benefit after set sub-group size to 16？ Any test result?

Those annotations were all on functions gated behind ggml_sycl_supports_mmq(), so essentially dead code.  It may need to be revisited if/when that function changes.

arthw (approved):
It fixes the AOT issue and make AOT work really.

Here is my test result:

|AOT flag|libggml-sycl.so.0.10.1 size (MB)|
|-|-|
||56|
|dg2|169|
|pvc,dg2| 349|

Thank you!

Issue #22107, opened 2026-04-19
Misc. bug: [CUDA/ROCm] VRAM leak/fragmentation in ggml_cuda_pool_leg when using Flash Attention


Zgate735:
### Name and Version

./llama-cli --version
ggml_cuda_init: found 2 ROCm devices (Total VRAM: 32736 MiB):
  Device 0: AMD Radeon Pro VII, gfx906:sramecc+:xnack- (0x906), VMM: no, Wave Size: 64, VRAM: 16368 MiB
  Device 1: AMD Radeon Pro VII, gfx906:sramecc+:xnack- (0x906), VMM: no, Wave Size: 64, VRAM: 16368 MiB
version: 8840 (9e5647aff)
built with GNU 15.2.0 for Linux x86_64

### Operating systems

Linux

### Which llama.cpp modules do you know to be affected?

llama-server

### Command line

```shell
llama-server -m '/data/models/gemma-4-31B-it-IQ4_NL.gguf' -ngl 999 -ctk q8_0 -ctv q8_0 -np 1 -b 4096 -ub 256 -fa on -cram 32768 -c 131072
```

### Problem description & steps to reproduce

Environment:
- GPU: AMD Radeon Pro VII (gfx906) / Also applies to any GPU using the Legacy Pool (VMM: no)
- Backend: ROCm / HIP
- Issue: Severe VRAM bloating leading to OOM (`HSA exception: MemoryRegion::BlockAllocator::alloc failed`)

The Bug:
When using a GPU without VMM support, `llama.cpp` falls back to `ggml_cuda_pool_leg` (`MAX_BUFFERS = 256`). 
When Flash Attention is enabled, it constantly requests temporary buffers. Because the context length grows with each generated token, the requested buffer size increases slightly on every step.

Because the new requested size is always slightly larger than the cached buffers, `alloc()` fails to find a suitable match in `buffer_pool`. Instead of reusing or freeing, it keeps the old buffers in the cache and allocates a new one via `ggml_cuda_device_malloc`. This quickly fills up all 256 slots with slightly undersized buffers, effectively hoarding VRAM until an OOM occurs.

The Vulkan backend does not suffer from this issue due to its different memory allocator).

The Workaround / Proposed Fix:
We need to proactively flush the legacy pool cache when we detect this "buffer size climbing" behavior, or at least flush it when `cudaMalloc` fails due to fragmentation.

[...]

Issue #22075, opened 2026-04-18
[Bug] CUDA OOM on Windows 7 during long-context prefill due to Legacy Pool accumulation


yuunnn-w:
### Environment
- OS: Windows 7 SP1 (x86-64) + [VxKex NEXT](https://github.com/YuZhouRen86/VxKex-NEXT)
- GPU: NVIDIA GeForce RTX 3060 (12 GB)
- Driver: 474.11
- CUDA: 11.4.2
- Compiler: Visual Studio 16 2019
- llama.cpp version: `llama.cpp-b8833`
- Command-line flags relevant: `-b 2048 -ub 512 --flash-attn on`

### Problem Description
When running `llama-server` and sending a long-context prompt (~20k tokens) for prefill, GPU memory rises step-wise until it hits the physical limit and CUDA throws `out of memory`.

```
CUDA error: out of memory
ggml/src/ggml-cuda/ggml-cuda.cu: CUDA error
current device: 0, in function alloc
ggml_cuda_device_malloc(&ptr, look_ahead_size, device)
```

The same binary runs fine on Windows 10/11 with identical hardware. The issue is 100% reproducible only on Win7.

---

### Call Chain Analysis

The memory leak is rooted in the interaction between long-context prefill, Flash Attention workspace allocation, and the Legacy CUDA memory pool. The complete call chain is as follows:

1. **`llama_context::decode()`** receives a batch of tokens and enters a `do-while` loop. For long prompts, the batch is split into multiple `ubatch`es (each up to `--ubatch-size`, default 512).

2. For each ubatch, **`process_ubatch()`** builds a computation graph (`ggml_cgraph`) containing the Flash Attention node (`GGML_OP_FLASH_ATTN_EXT`). During prefill, the KV cache length (`n_kv`) represented in this graph grows monotonically with each successive ubatch.

3. When the scheduler executes the graph, **`ggml_backend_cuda_graph_compute()`** dispatches each node to **`ggml_cuda_compute_forward()`**. For Flash Attention, this routes to **`ggml_cuda_flash_attn_ext()`** and its specialized variants (`vec`, `tile`, `mma`, etc.).

4. Inside the Flash Attention kernel launcher (e.g. **`flash_attn_ext_vec_f16`**, **`flash_attn_ext_tile_f16`**), temporary workspace buffers are allocated via the helper template **`ggml_cuda_pool_alloc<T>`**:
   - `K_f16` — quantized K cache dequantized to FP16
[...]

PR #22155, opened 2026-04-20
ggml-cuda: flush legacy pool on OOM and retry

leonardHONG:
This adds a conservative fallback for the legacy CUDA/HIP pool allocator.

On non-VMM setups, the legacy pool can end up holding cached free buffers that are individually too small for a new request, but still occupy enough VRAM to make the next allocation fail. In that case, this patch flushes the cached legacy-pool buffers and retries the allocation once before aborting.

The normal hit path is unchanged. This is intended as a narrow mitigation for legacy-pool OOMs, not a broader allocator redesign. I validated the retry path locally with a synthetic OOM injection on a legacy-pool build.

This is intended to mitigate the legacy-pool OOM behavior reported in #22075 and #22107.

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -379,6 +379,18 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
         GGML_ASSERT(pool_size == 0);
     }
 
+    void clear_pool() {
+        for (int i = 0; i < MAX_BUFFERS; ++i) {
+            ggml_cuda_buffer & b = buffer_pool[i];
+            if (b.ptr != nullptr) {
+                CUDA_CHECK(cudaFree(b.ptr));
+                pool_size -= b.size;
+                b.ptr  = nullptr;
+                b.size = 0;
+            }
+        }
+    }
+
     void * alloc(size_t size, size_t * actual_size) override {
 #ifdef DEBUG_CUDA_MALLOC
         int nnz = 0;
@@ -421,7 +433,24 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
         size_t look_ahead_size = (size_t) (1.05 * size);
         look_ahead_size = 256 * ((look_ahead_size + 255)/256);
         ggml_cuda_set_device(device);
+#if defined(GGML_USE_MUSA)
         CUDA_CHECK(ggml_cuda_device_malloc(&ptr, look_ahead_size, device));
+#else
+        cudaError_t err = ggml_cuda_device_malloc(&ptr, look_ahead_size, device);
+        if (err == cudaErrorMemoryAllocation) {
+            // only invoked from alloc() after ggml_cuda_set_device.
+            (void)cudaGetLastError();
+            const size_t cached_bytes = pool_size;
+            GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: alloc of %.2f MiB failed, flushing %.2f MiB of cached buffers and retrying\n",
+                           device, look_ahead_size/1024.0/1024.0, cached_bytes/1024.0/1024.0);
+            clear_pool();
+            err = ggml_cuda_device_malloc(&ptr, look_ahead_size, device);
+            if (err == cudaSuccess) {
+                GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: retry succeeded\n", device);
+            }
+        }
+        CUDA_CHECK(err);
+#endif
         *actual_size = look_ahead_size;
         pool_size += look_ahead_size;
 #ifdef DEBUG_CUDA_MALLOC
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -368,18 +368,12 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
     }
 
     ~ggml_cuda_pool_leg() {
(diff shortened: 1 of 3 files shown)


JohannesGaessler:
This is not safe to do. If the memory is still in use it must not be freed.

leonardHONG:
> This is not safe to do. If the memory is still in use it must not be freed.

Thanks for the catch! I completely overlooked the async lifecycle issue here. I'll drop this unsafe logic and rework it tonight.

IMbackK:
I.. actually don't see it (how clear_pool can prune in use chunks)

JohannesGaessler:
The legacy CUDA buffer pool is essentially just a list of previously used temporary buffers. "Allocating" and "freeing" a buffer just means retrieving a buffer for use within a kernel. It does **not** mean that there no longer are any kernels queued that will attempt to use the buffer. 

Since the buffer pool is per `ggml_backend_cuda_context` I think the current approach is safe if it is synchronized on the corresponding stream. Looking at the documentation for `cudaMalloc` and `cudaFree` again however, it seems those do an implicit device-wide synchronization when they're called. So even without any modifications I think this PR is safe after all.

Did you consider replacing the legacy buffer pool with `cudaMallocAsync` and `cudaFreeAsync`? The only reason we are not using those for CUDA is that the performance was worse vs. manually managed chunks of memory but I don't think I ever checked the performance for HIP.

IMbackK:
> Looking at the documentation for `cudaMalloc` and `cudaFree` again however, it seems those do an implicit device-wide synchronization when they're called. So even without any modifications I think this PR is safe after all.

Jeah exactly, i was really confused there thinking that this is perhaps not true for cuda and only for hip.

IMbackK:
> Did you consider replacing the legacy buffer pool with `cudaMallocAsync` and `cudaFreeAsync`? The only reason we are not using those for CUDA is that the performance was worse vs. manually managed chunks of memory but I don't think I ever checked the performance for HIP.

I did try this, it is slow. Hopefully they will fix the damn virtual memory support soon.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -379,6 +379,18 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
...
+        for (int i = 0; i < MAX_BUFFERS; ++i) {
+            ggml_cuda_buffer & b = buffer_pool[i];
+            if (b.ptr != nullptr) {
+                CUDA_CHECK(cudaFree(b.ptr));
+                pool_size -= b.size;
+                b.ptr  = nullptr;
+                b.size = 0;
+            }
+        }
+    }
```
JohannesGaessler, line 386: To ensure consistency, please call `clear_pool` in the destructor.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -421,7 +433,24 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
         size_t look_ahead_size = (size_t) (1.05 * size);
         look_ahead_size = 256 * ((look_ahead_size + 255)/256);
         ggml_cuda_set_device(device);
+#if defined(GGML_USE_MUSA)
```
JohannesGaessler: If at all possible, please just add the missing defines to the MUSA header if they're missing rather than to have per-vendor logic.

gaugarg-nv:
I'd suggest if you go with this fix, don't rely on this behavior of `cudaFree`.  It is better to add an explicit `cudaDeviceSynchronize` or consider using the async version of `cudaMalloc` and `cudaFree` as @JohannesGaessler suggested.

The CUDA documentation for `cudaFree` says: "For all other pointers, this API **_may_** perform implicit synchronization."
So, I won't rely on this.

IMbackK:
> The CUDA documentation for `cudaFree` says: "For all other pointers, this API **_may_** perform implicit synchronization." So, I won't rely on this.

On hip there is no may so you would be safe there, doing an explicit synchronization would not hurt ofc.

leonardHONG:
Thanks everyone for the amazing clarifications! I really appreciate the guidance. I will apply all the suggested fixes (adding explicit sync, updating the destructor, and cleaning up the MUSA logic) and push an update tonight.

JohannesGaessler:
>The CUDA documentation for cudaFree says: "For all other pointers, this API may perform implicit synchronization."

Thank you for correcting me, it seems I suffered from confirmation bias and should have been more careful. The way I had remembered it the synchronization was guaranteed, that is what Gemini said when I asked "Does cudaFree imply a device synchronization?", and that is what I read when I quickly checked the CUDA documentation it linked. I agree that there should be an explicit synchronization to be safe.

gaugarg-nv:
AFAIK, there is always a sync with `cudaFree`, but documentation doesn't guarantee it. So, it is better to make it explicit in the code. Here is the [document ](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ga042655cbbf3408f01061652a075e094)I was referring to.

leonardHONG:
Thanks for the feedback — I pushed an update.

This revision:

* reuses `clear_pool()` in the destructor
* makes `clear_pool()` self-contained with device selection
* adds an explicit synchronization before clearing cached legacy-pool buffers on the OOM retry path
* removes the MUSA-specific branch by adding the corresponding error alias in the vendor header

I also reran `test-backend-ops` locally and it passes on my side.

JohannesGaessler (approved):

IMbackK (approved):

Issue #21889, opened 2026-04-14
server: allow assistant prefill when the prefill itself closes the <think> block


cjnuk:
### Summary

When `--reasoning-budget` is non-zero, `/v1/chat/completions` unconditionally rejects any request where the last message has `role: "assistant"` (i.e. assistant prefill), returning:

```
HTTP 400: {"error": {"code": 400, "message": "Assistant response prefill is incompatible with enable_thinking."}}
```

This restriction is too broad. When the prefill content is itself a fully closed `<think>...</think>` block, there is no incompatibility: the user is explicitly telling the model that reasoning is complete and it should proceed directly to output. The server has no basis to refuse this.

We ran a 10-pair experiment on Qwen3.5-27B (distilled, i1-Q6_K, llama.cpp b8763) using the native `/completion` endpoint (which accepts the prefill) to validate the technique works correctly. Mean wall-time reduction: **-8.64s per request**. Mean tokens saved: **456 per request**. Quality regressions: **0/10**. The blocker is the `/v1/chat/completions` guard that prevents using this technique through the standard OpenAI-compatible path.

---

### Current behaviour

Sending an assistant prefill via `/v1/chat/completions` while `--reasoning-budget -1` is set:

```bash
curl -s http://localhost:8080/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "qwen3.5-27b",
    "messages": [
      {"role": "user", "content": "Write a Python one-liner that reverses a list."},
      {"role": "assistant", "content": "<think>\nDirect answer follows.\n</think>\n\n"}
    ],
    "chat_template_kwargs": {"add_generation_prompt": false}
  }'
```

Response:

```json
{"error": {"code": 400, "message": "Assistant response prefill is incompatible with enable_thinking."}}
```

The check is unconditional. It fires in `tools/server/server-common.cpp` at the point where `prefill_assistant_message` is true and `inputs.enable_thinking` is true, with no inspection of the prefill string:

```cpp
// tools/server/server-common.cpp:1062-1063
if ( inputs.enable_thinking ) {
    throw std::invalid_argument("Assistant response prefill is incompatible with enable_thinking.");
}
```

---

### Workaround that currently works

Using the native `/completion` endpoint with a manually serialized chat-template prompt:

```bash
curl -s http://localhost:8080/completion \
  -H "Content-Type: application/json" \
  -d '{
[...]

Issue #21754, opened 2026-04-11
Feature Request: WebUI response streaming is fragile


coder543:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/ggml-org/llama.cpp/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/ggml-org/llama.cpp/discussions), and have a new and useful enhancement to share.

### Feature Description

I would like `webui` to be robust to connection interruptions. If I am streaming a response to my phone and the network changes, the response is lost immediately, which is not a good user experience.

### Motivation

On the ChatGPT app, it is nice to start something, close the app, and come back later when the response is done.

I would enjoy having a similar level of polish in `webui`.

### Possible Implementation

I can think of two approaches that make sense:

1. We could fix the trailing assistant prefill API so that it works with reasoning models.

This is possibly the least invasive approach, although it doesn't allow background chats to work if the client pauses in the background, like it does on iOS.

Essentially: when the client detects the connection is interrupted, it would send the chat history plus the incomplete response back to the server and say "continue" automatically. There is no _technical_ reason that this couldn't work that I'm aware of. The LLM does not care whether it is naturally generating tokens in the middle of reasoning or not, or whether it has simply been given a partially-prefilled response and told to continue. It _cannot_ know. It's just a matter of hooking up the template machinery correctly so that a client can send either a partial reasoning_content or a complete reasoning_content and a partial response text and have the server resume.

`webui` already supports a manual continue button for non-reasoning models only, and this restriction is because of limitations in the current `llama-server` implementation.

2. The other approach is to allow the server to support SSE resumption.

The underlying chat stream is an SSE stream, even if the `POST` prevents the client from using `EventSource`. The server could optionally accept a random UUID as a `session_id` in the chat request to recognize this extension of the chat API and be able to later identify this generation request. When provided with a `session_id`, the server would include a sequentially autoincrementing integer in the `id:` field for each SSE event.

Normally, the server stops generating a response the instant the client disconnects. When there is a `session_id`, the server would _not_ give up on generating the response just because the client disconnects. This maintains compatibility with the existing expectation, because clients would have to opt into this new behavior. Instead of quitting, the server would keep generating into a `session_id`-specific buffer.

[...]

PR #22162, opened 2026-04-20
server: Allow continue in thinking (reasoning prefill)

roj234:
## Overview

Make prefill available for (probably only *some*) reasoning model

## Additional information

Current change may lack the ability to add '\n</think>' 

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/tools/server/server-common.cpp
+++ b/tools/server/server-common.cpp
@@ -1082,13 +1082,6 @@ json oaicompat_chat_params_parse(
             throw std::invalid_argument("Cannot have 2 or more assistant messages at the end of the list.");
         }
 
-        /* TODO: test this properly */
-        inputs.reasoning_format = COMMON_REASONING_FORMAT_NONE;
-
-        if ( inputs.enable_thinking ) {
-            throw std::invalid_argument("Assistant response prefill is incompatible with enable_thinking.");
-        }
-
         inputs.add_generation_prompt = true;
     }
     inputs.force_pure_content = opt.force_pure_content;
@@ -1098,12 +1091,39 @@ json oaicompat_chat_params_parse(
 
     /* Append assistant prefilled message */
     if (prefill_assistant_message) {
+        auto inside_thinking = inputs.enable_thinking;
+        auto thinking_start = chat_params.thinking_start_tag;
+        auto thinking_end = chat_params.thinking_end_tag;
+
+        if ( inside_thinking ) {
+            int32_t pos = chat_params.prompt.length() - thinking_start.length();
+            // Could pos < 0 ever ???
+            if (pos < 0 || chat_params.prompt.find(thinking_start, pos) != pos) {
+                chat_params.prompt += thinking_start;
+            }
+            chat_params.prompt += last_message.reasoning_content;
+        }
+
         if (!last_message.content_parts.empty()) {
             for (auto & p : last_message.content_parts) {
+                if (p.text.length() > 0 && inside_thinking) {
+                    chat_params.prompt += thinking_end;
+                    inside_thinking = false;
+                }
+
                 chat_params.prompt += p.text;
             }
         } else {
+            if (last_message.content.length() > 0 && inside_thinking) {
+                chat_params.prompt += thinking_end;
+                inside_thinking = false;
+            }
             chat_params.prompt += last_message.content;
+        }
+
+        if (inputs.enable_thinking && !inside_thinking/* both reasoning_content and content, reasoning ended */) {
+            // feed to chat-parser to let it generate `content` instead of `reasoning_content`
+            llama_params["prefill_prompt"] = thinking_end;
         }
     }
 
--- a/common/chat-diff-analyzer.cpp
+++ b/common/chat-diff-analyzer.cpp
@@ -296,7 +296,7 @@ void analyze_reasoning::compare_reasoning_presence() {
             return p.literal(reasoning_content) + p.space() + p.optional(p.tag("post", (p.marker() + p.space())) + p.rest());
         });
         auto parser_wrapped = build_tagged_peg_parser([&](common_peg_parser_builder &p) {
-            return p.tag("pre", p.marker() + p.space()) + p.literal(reasoning_content) + p.space() + p.tag("post", (p.marker() + p.space())) + p.rest();
+            return p.tag("pre", p.marker() + p.space()) + p.literal(reasoning_content) + p.tag("post", (p.space() + p.marker() + p.space())) + p.rest();
         });
         // try the more aggressive parse first, if it fails, fall back to the delimiter one
         auto result = parser_wrapped.parse_anywhere_and_extract(comparison->output_B);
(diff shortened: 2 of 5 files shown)


roj234:
by make `chat_params.thinking_end_tag` from `</think>` to `\n</think>` I have found Qwen3.5 can correctly stop thinking after reasoning_budget triggered.

roj234:
fixes #21889 (although must use `reasoning_content` field rather than raw `content`)
What is `correctly stop thinking` (after reasoning_budget triggered):

Before

After


Which means Qwen3.5 requires `\n</think>` to end thinking not just `</think>`

roj234:
BTW, can you merge (or review) my other PRs 🙏, like OpenRouter-compatible reasoning_budget API

#20069 
#20088 
#22038

roj234:
Now `chat/completions` can *always* generate correct `reasoning_content` and `content` fields
I have added a field to `task_params` and change `task_result_state` constructor...
It is ugly (maybe better to refactor some part of `generated_text += `), but IT IS WORKING now

pwilkin (changes requested):
The autoparser changes look sound to me, you're right I shouldn't have relied on trimming heuristics, they yield subtle bugs like this.

As for the assistant prefill part: please remove the extra task_state parameter and instead move the `prefill_text` into `chat_parser_params`.

aldehir:
I think the prefill should be the responsibility of the chat param init, not the server. That's where the template structure can be best preserved.

ServeurpersoCom:
I did a POC to test the resume during reasoning content on WebUI, but I had to modify the backend; this issue is what we need to do this ( https://github.com/ggml-org/llama.cpp/issues/21754 ) properly.

roj234:
However chat_parser_params.prefill_prompt currently only used once ...


after all it's a new field 😂

Edit: Should I click `re-request review?`

tools/server/server-common.cpp
```diff
@@ -1096,13 +1087,33 @@ json oaicompat_chat_params_parse(
 
     /* Append assistant prefilled message */
     if (prefill_assistant_message) {
+        auto append_thinking_end = inputs.enable_thinking;
+        if ( append_thinking_end ) {
```
pwilkin: style: remove extra spaces

pwilkin (approved):

pwilkin:
```console
[ 95%] Built target llama-template-analysis
In file included from /opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.cpp:1:
/opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.h: In constructor ‘task_result_state::task_result_state(const common_chat_parser_params&)’:
/opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.h:112:23: error: ‘task_result_state::oai_resp_message_id’ will be initialized after [-Werror=reorder]
  112 |     const std::string oai_resp_message_id;
      |                       ^~~~~~~~~~~~~~~~~~~
/opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.h:100:17: error:   ‘std::string task_result_state::generated_text’ [-Werror=reorder]
  100 |     std::string generated_text; // append new chunks of generated text here
      |                 ^~~~~~~~~~~~~~
/opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.h:115:5: error:   when initialized here [-Werror=reorder]
  115 |     task_result_state(const common_chat_parser_params & chat_parser_params)
      |     ^~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
gmake[2]: *** [tools/server/CMakeFiles/server-context.dir/build.make:93: tools/server/CMakeFiles/server-context.dir/server-task.cpp.o] Error 1
[ 95%] Built target llama-debug-template-parser
[ 95%] Linking CXX executable ../../bin/llama-tts
[ 95%] Built target llama-tts
[ 95%] Linking CXX executable ../bin/test-chat-auto-parser
In file included from /opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-context.h:4,
                 from /opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-context.cpp:2:
/opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.h: In constructor ‘task_result_state::task_result_state(const common_chat_parser_params&)’:
/opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.h:112:23: error: ‘task_result_state::oai_resp_message_id’ will be initialized after [-Werror=reorder]
  112 |     const std::string oai_resp_message_id;
      |                       ^~~~~~~~~~~~~~~~~~~
/opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.h:100:17: error:   ‘std::string task_result_state::generated_text’ [-Werror=reorder]
  100 |     std::string generated_text; // append new chunks of generated text here
      |                 ^~~~~~~~~~~~~~
/opt/actions-runner/_work/llama.cpp/llama.cpp/tools/server/server-task.h:115:5: error:   when initialized here [-Werror=reorder]
  115 |     task_result_state(const common_chat_parser_params & chat_parser_params)
      |     ^~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
gmake[2]: *** [tools/server/CMakeFiles/server-context.dir/build.make:135: tools/server/CMakeFiles/server-context.dir/server-context.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:5288: tools/server/CMakeFiles/server-context.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
[ 96%] Linking CXX executable ../bin/test-chat-peg-parser
[ 96%] Built target test-chat-auto-parser
[ 96%] Built target test-chat-peg-parser
```

Need to fix CI errors.

aldehir (changes requested):
I am not convinced this is the right approach.

It assumes that the thinking tag is left open when `enable_thinking = true`, which is not universally true. Gemma 4, for example, does not add an opening thinking tag when enabled.

More importantly, the logic is handled in the server and over-generalizes to opening/closing think tags.  This will work for most models, but it will break for `gpt-oss`, which expects content in a separate message envelope, e.g., `<|start|>assistant<|channel|>final<|message|>{prefill content}`.

Also, `generation_prompt` was added to begin parsing from the beginning of the assistant message, instead of generation. This could be leveraged to avoid the `prefill_prompt` workaround.

I still think the responsibility of prefilling should shift to the chat param initialization. At least then prompts could be specialized per model. The autoparser can generalize the rest.

roj234:
> I am not convinced this is the right approach.
> 
> It assumes that the thinking tag is left open when `enable_thinking = true`, which is not universally true. Gemma 4, for example, does not add an opening thinking tag when enabled.
> 
> More importantly, the logic is handled in the server and over-generalizes to opening/closing think tags. This will work for most models, but it will break for `gpt-oss`, which expects content in a separate message envelope, e.g., `<|start|>assistant<|channel|>final<|message|>{prefill content}`.
> 
> Also, `generation_prompt` was added to begin parsing from the beginning of the assistant message, instead of generation. This could be leveraged to avoid the `prefill_prompt` workaround.
> 
> I still think the responsibility of prefilling should shift to the chat param initialization. At least then prompts could be specialized per model. The autoparser can generalize the rest.

Without making breaking changes to the existing architecture, it is difficult to implement this function:

- Current PEG reasoning parser detects raw output text rather than token sequences. As a result, it cannot distinguish between model-generated strings and special tokens.


- My approach directly prefills the user input. Since llama.cpp use safe Jinja2 parser, it will not parse special tokens inside user input.

- The `chat_parser` component definitely needs some refactor to handle these edge cases properly.
For example: accept(token_id, text) -> { content? , reasoning_content? }

However, If I do, (It will change lots of code), will this extremely large PR will be accepted?
I thought It's not my turn now.
This PR is left open for reference, you can close it freely

Te-eMster:
I think this PR (especially the changes already merged in #22353) introduced a bug (#22422) where tool calls with models like MiniMax-M2.7 and Step-3.5-Flash fail (these two are the ones I personally tested as well).

Would it be possible to look into this while you guys are working on this topic? Thanks so much :-)

PR #22163, opened 2026-04-20
server: fix /api/show handler type

roj234:
## Overview

/api/show is GET route rather than POST route

## Additional information

# Requirements

- AI usage disclosure: NO


ggerganov:
It's better to remove it: #22165

PR #22165, opened 2026-04-20
server : remove /api endpoints

ggerganov:
## Overview

Removing the `/api` endpoints. There is no reason to support these.

# Requirements

- AI usage disclosure: NO


roj234:
`/api/tags` also referenced in `server/server-http.cpp` Line ~146

ngxson (approved):
Agree, we should focus our resources to the more matured APIs like `/resposes` or  anthropic `/v1/messages`

PR #22167, opened 2026-04-20
Fix incorrect assertion

fiesh:
The assertion check was too strong.

This closes #19839.

- AI usage disclosure: Yes, to reason about the code paths leading to the assertion failure.

### DIFF
--- a/ggml/src/ggml-backend.cpp
+++ b/ggml/src/ggml-backend.cpp
@@ -254,7 +254,8 @@ size_t ggml_backend_get_max_size(ggml_backend_t backend) {
 void ggml_backend_tensor_set_async(ggml_backend_t backend, struct ggml_tensor * tensor, const void * data, size_t offset, size_t size) {
     GGML_ASSERT(backend);
     GGML_ASSERT(tensor);
-    GGML_ASSERT(tensor->data != NULL && "tensor not allocated");
+    GGML_ASSERT((tensor->data != NULL || (tensor->view_src != NULL && tensor->view_src->data != NULL)) &&
+                "tensor not allocated");
     GGML_ASSERT(offset + size <= ggml_nbytes(tensor) && "tensor write out of bounds");
 
     if (backend->iface.set_tensor_async == NULL) {
@@ -268,7 +269,8 @@ void ggml_backend_tensor_set_async(ggml_backend_t backend, struct ggml_tensor * 
 void ggml_backend_tensor_get_async(ggml_backend_t backend, const struct ggml_tensor * tensor, void * data, size_t offset, size_t size) {
     GGML_ASSERT(backend);
     GGML_ASSERT(tensor);
-    GGML_ASSERT(tensor->data != NULL && "tensor not allocated");
+    GGML_ASSERT((tensor->data != NULL || (tensor->view_src != NULL && tensor->view_src->data != NULL)) &&
+                "tensor not allocated");
     GGML_ASSERT(offset + size <= ggml_nbytes(tensor) && "tensor read out of bounds");
 
     if (backend->iface.get_tensor_async == NULL) {
@@ -283,7 +285,8 @@ void ggml_backend_tensor_set_2d_async(ggml_backend_t backend, struct ggml_tensor
             size_t n_copies, size_t stride_tensor, size_t stride_data) {
     GGML_ASSERT(backend);
     GGML_ASSERT(tensor);
-    GGML_ASSERT(tensor->data != NULL && "tensor not allocated");
+    GGML_ASSERT((tensor->data != NULL || (tensor->view_src != NULL && tensor->view_src->data != NULL)) &&
+                "tensor not allocated");
 
     if (n_copies <= 1 || backend->iface.set_tensor_2d_async == NULL) {
         for (size_t i = 0; i < n_copies; i++) {
@@ -295,7 +298,8 @@ void ggml_backend_tensor_set_2d_async(ggml_backend_t backend, struct ggml_tensor
         return;
     }
 
-    GGML_ASSERT(tensor->data != NULL && "tensor not allocated");
+    GGML_ASSERT((tensor->data != NULL || (tensor->view_src != NULL && tensor->view_src->data != NULL)) &&
+                "tensor not allocated");
     GGML_ASSERT(offset + (n_copies-1)*stride_tensor + size <= ggml_nbytes(tensor) && "tensor write out of bounds");
     backend->iface.set_tensor_2d_async(backend, tensor, data, offset, size, n_copies, stride_tensor, stride_data);
 }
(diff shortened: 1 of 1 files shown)


fiesh:
This works, but I'm not certain if it's that great.  It might also make sense to reduce code duplication by not repeating the assertion every time.  But since that's the way it was originally, I wanted to keep the change minimal.

I'd be grateful for somebody with better understanding of the code if the new assertion is correct or just hides a deeper issue.  It does seem stable though.

fiesh:
(A minor thing -- I'd also greatly prefer not to use `NULL` but `nullptr` instead.  Wasn't sure if this was a wanted change or not.  If it is, I'd gladly switch it out in this PR as well.)

am17an:
Feels like a backend bug, why would `tensor->data` not be set?

fiesh:
> Feels like a backend bug, why would `tensor->data` not be set?

`ggml_gallocr_init_tensor` explicitly allows for this case, which led me to believe it may be valid.

I don't understand why `ggml_tensor` doesn't have a proper ctor to enforce its invariants... makes it hard to reason about for a newcomer.

fiesh:
I think one thing that should be addressed here as well are `ggml_get_data` and `ggml_get_data_f32`. 

Should they:

1. `assert` that `tensor->data` is not `NULL`, or
2. check if `tensor->data` is `NULL`, and if so, return `tensore->view_src->data` instead?

fiesh:
> I don't understand why `ggml_tensor` doesn't have a proper ctor to enforce its invariants... makes it hard to reason about for a newcomer.

OK it's a C library :sweat_smile:

fiesh:
This is Qwen's reasoning for the situation here:

This is deterministic, not a race. The `--parallel` > 1 flag changes the tensor allocation strategy in the scheduler. Here's the exact mechanism:

`--parallel` = 1: The scheduler creates one graph. All tensors (including view tensors) go through `ggml_gallocr_alloc_graph` → `ggml_gallocr_init_tensor`, which calls `ggml_backend_view_init()` on view tensors, setting `tensor->data = view_src->data + view_offs`.

`--parallel` > 1: The scheduler creates tensor copies via `tensor_id_copy`. For the current copy `(c == sched->cur_copy)`, it uses the original tensor directly (line 1347: `tensor_copy = src`). This original tensor was allocated elsewhere (e.g., by the model loader or a previous graph compute). If it's a view tensor whose data was captured when `view_src->data` was `NULL`, then:

- `tensor->data == NULL` (captured at creation time, never updated)
- `tensor->view_src->data != NULL` (source was allocated later)
- `tensor->buffer != NULL` (allocated by ggml-backend elsewhere)

Because `tensor->buffer != NULL`, the `ggml_gallocr_init_tensor` early-return path is taken (line 974: `if (tensor->buffer == NULL)` is `false`), so `ggml_backend_view_init()` is never called on this tensor.

The tensor is then passed to `ggml_backend_tensor_set` (via `input_cpy` at line 1436 → `ggml_backend_tensor_copy` at line 1578 or `ggml_backend_tensor_set_async` at line 1641), hitting the assertion.

The fix is deterministic: `--parallel` > 1 changes the allocation path so that `ggml_backend_view_init` is skipped on tensors that were pre-allocated with stale view data. It's not a race — it's a structural difference in how the scheduler handles tensor copies vs. single-graph allocation.

am17an:
Create an issue with a repro if possible

fiesh:
> Create an issue with a repro if possible

I suppose I'd add it to the existing issue #19839

Can you tell me what exactly I should add?

am17an:
I think that issue is good enough. Maybe @0cc4m can take a look

fiesh:
Any news on reviewing this?  Our server has been running without a single crash with this change for the past two weeks now, and it's used rather heavily, in parallel.

fiesh:
ping @0cc4m

0cc4m:
Sorry about the delay. This is a deeper question, relaxing the assertion means you have to check that existing implementations of these functions already take into account the case you are trying to open up. Did you go through all backend implementations to check if they actually use the view tensor data? Otherwise you are just opening the path into issues that are way harder to debug.

The other way to look at this is that if it runs into this assertion, then a function was misused somewhere to operate on a view tensor, which is not currently permitted by the function.

fiesh:
@0cc4m 

Thank you for your feedback!

> Sorry about the delay. This is a deeper question, relaxing the assertion means you have to check that existing implementations of these functions already take into account the case you are trying to open up. Did you go through all backend implementations to check if they actually use the view tensor data? Otherwise you are just opening the path into issues that are way harder to debug.

I think you're right, this is an issue.  For example `ggml_backend_cuda_set_tensor_async` unconditionally calls `CUDA_CHECK(cudaMemcpyAsync((char *) tensor->data + offset, data, size, cudaMemcpyHostToDevice, cuda_ctx->stream()));`.

> The other way to look at this is that if it runs into this assertion, then a function was misused somewhere to operate on a view tensor, which is not currently permitted by the function.

I think my view might be this, a third option: The assertion is wrong to begin with in `ggml-backend.cpp`. Rather, it should be placed in the backends' implementations.  This way, we don't have to propagate assumptions about these functions, assumptions that might be outdated and hard to keep track of.  Does that make sense?  If so, should I try moving the assertions to the backends?

0cc4m:
No, the assertion defines what the backend function expects from the backends, and how it is supposed to be used.

fiesh:
> No, the assertion defines what the backend function expects from the backends, and how it is supposed to be used.

I'm not sure I understand this.  For example, `ggml_backend_tensor_set_async` doesn't expect anything from the backends, it just calls `backend->iface.set_tensor_async` (if set).  This in turn may handle view tensors correctly or not.

From its point of view, it seems only the assertion `GGML_ASSERT(backend);` is necessary.

0cc4m:
You can think of it like this: If I implement a new backend, I can look at the assertions to see what cases I need to handle. So I look at `GGML_ASSERT(tensor->data != NULL && "tensor not allocated");` and see that I do not need to handle view tensors. That's how it defines what is expected from backends. If you change it, you also need to change all backend implementations at the same time, not just move the assertion there.

fiesh:
> You can think of it like this: If I implement a new backend, I can look at the assertions to see what cases I need to handle. So I look at `GGML_ASSERT(tensor->data != NULL && "tensor not allocated");` and see that I do not need to handle view tensors. That's how it defines what is expected from backends. If you change it, you also need to change all backend implementations at the same time, not just move the assertion there.

OK, thank you for the explanation.  Having these assertions be the contracts for the backends makes sense.

However, we have the following situation:

1. The Vulkan backend produces (when `--parallel` is > 1) view tensors that are passed back to it via `ggml-backend.cpp`.
2. The CUDA backend does not do this (I cannot verify this, but it's fairly certain or else nobody could be using CUDA with `--parallel` > 1)

But asserting that view tensors don't get passed to the backend make it impossible for backends to behave differently in this regard.

I have no technical understanding of via Vulkan does and CUDA does not use view tensors here, but it seems at least meaningful that different backends can make different choices?

0cc4m:
The backends should not influence tensors at all, they just consume them, so it would be odd if it really only affected Vulkan. The only thing I can think of is that CUDA is taking a different codepath, one of the two backends is falling back to something else. Someone would need to look at it with a debugger.

fiesh:
OK, then this PR is definitely the wrong solution.

PR #22168, opened 2026-04-20
ngram-mod: Reset i_last when low acceptance streak occurs

treo:
## Overview
By resetting i_last to zero, we will include the current context when rebuilding the speculative map.

The existing behavior would skip the current context, thereby often losing the benefit of speculative decoding during later parts of the generation. 

The effect of this seems to depend on both the model and the speculation parameters.

Benchmark: 
`vllm bench serve --model google/gemma-4-26B-A4B-it --host 127.0.0.1 --port 9876 --num-prompts 96 --dataset-name hf --dataset-path vdaita/edit_5k_char --backend openai-chat --endpoint '/v1/chat/completions' --max-concurrency 1`

Gemma 4 26B-A4B: 
- Baseline: 97.57 t/s (peak 108)
- without i_last = 0: 141.08 t/s (peak 1032)
- with i_last = 0: 155.10 t/s (peak 1032)

Qwen 3.6 35B-A3B:
- Baseline: 112.54 t/s (peak 127)
- without i_last = 0: 148.56 t/s (peak 774)
- with i_last = 0: 153.65 t/s (peak 768)

As we can see, the effect isn't huge but at 3% to 9% it is still measurable.
The price we pay for it is 1 line of code and a moment longer for speculative map repopulation whenever a low acceptance streak occours. 

# Requirements

- AI usage disclosure: YES. Used to help understand the ngram-mod flow.

### DIFF
--- a/common/speculative.cpp
+++ b/common/speculative.cpp
@@ -749,6 +749,7 @@ struct common_speculative_state_ngram_mod : public common_speculative_state {
 
                     mod.reset();
                     n_low = 0;
+                    i_last = 0;
                 }
             } else {
                 n_low = 0;


ggerganov:
What speculative decoding parameters did you use in this test?

treo:
Oh, sorry, it appears in cleaning up the layout I deleted the lines with the parameters.
The exact draft parameters were: `--draft-max 128 --spec-ngram-size-n 48 --draft-min 2 --spec-type ngram-mod` 

Additionally, I had to disable reasoning (`--reasoning off`) to make the results of the benchmark repeatable.

I've been using this change ever since proposing it for agentic coding with slightly different parameters: `-spec-type ngram-mod --spec-ngram-size-n 32 --draft-max 64 --draft-min 32` and whenever it needs to reproduce big chunks of code I often see 3x to 7x over baseline with context sizes over 100k token.

In my build I also added a little debug logging, and it is not uncommon to see a low acceptance streak right before a big chunk of 100% acceptance. With i_last not being reset, it wouldn't have anything to work with in those cases.

ggerganov (approved):

PR #22169, opened 2026-04-20
Fix server includes common/build-info.h +  `noreturn` attribute warnings

stsydow:
## Overview

I had an error building the llama-server  after the lib common rename (`git bisect` pointed to  6990e2f1f7581d332a6a1f34d6c567be70138184)

The fix is just to moves the servers includes from "build-info.h" to "common/build-info.h"

I also fixed warnings on `noreturn` attribute functions.

## Additional information

I build on Linux using:
```sh
cmake -B build \
    -DGGML_VULKAN=ON \
    -DGGML_USE_OPENMP=ON \
    -DGGML_AVX=ON -DGGML_AVX2=ON -DGGML_AVX_VNNI=ON -DGGML_AVX_BMI=ON -DGGML_FMA=ON \
    -DGGML_SSE42=ON -DGGML_F16C=ON \
    -DGGML_NATIVE=ON

cmake --build build --config Release -t llama-server --parallel 16
```

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/common/jinja/runtime.h
+++ b/common/jinja/runtime.h
@@ -107,7 +107,7 @@ struct statement {
     virtual ~statement() = default;
     virtual std::string type() const { return "Statement"; }
     // execute_impl must be overridden by derived classes
-    virtual value execute_impl(context &) { throw std::runtime_error("cannot exec " + type()); }
+    [[noreturn]] virtual value execute_impl(context &) { throw std::runtime_error("cannot exec " + type()); }
     // execute is the public method to execute a statement with error handling
     value execute(context &);
 };
@@ -143,7 +143,7 @@ struct program : public statement {
     program() = default;
     explicit program(statements && body) : body(std::move(body)) {}
     std::string type() const override { return "Program"; }
-    value execute_impl(context &) override {
+    [[noreturn]] value execute_impl(context &) override {
         throw std::runtime_error("Cannot execute program directly, use jinja::runtime instead");
     }
 };
@@ -195,7 +195,7 @@ struct break_statement : public statement {
         }
     };
 
-    value execute_impl(context &) override {
+    [[noreturn]] value execute_impl(context &) override {
         throw break_statement::signal();
     }
 };
@@ -209,7 +209,7 @@ struct continue_statement : public statement {
         }
     };
 
-    value execute_impl(context &) override {
+    [[noreturn]] value execute_impl(context &) override {
         throw continue_statement::signal();
     }
 };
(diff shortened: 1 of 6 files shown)


CISC:
This is incorrect (and you also for some reason have included some other (also incorrect) changes in this PR), check that you don't have a stale `build-info.h` in root, and delete it if you do.

Edit: Saw this now, it will break build, the warning is bogus:
> I also fixed warnings on noreturn attribute functions.

stsydow:
Sorry for the noise and thanks.
I checked out the repo again and now it works.

The `noreturn` warning came up using the Intel compiler on an other build: 

```sh
#!/bin/bash

source /opt/intel/oneapi/setvars.sh

rm build -r
mkdir -p build

cmake -B build \
    -DGGML_SYCL=ON -DENABLE_TRY_SYCL_COMPILE=ON -DGGML_SYCL_DEVICE_ARCH=acm_g10 \
    -DGGML_SYCL_GRAPH=ON -DGGML_SYCL_DNN=ON \
    -DGGML_SYCL_F16=ON \
    -DGGML_VULKAN=ON \
    -DGGML_USE_OPENMP=ON \
    -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx \
    -DGGML_AVX=ON -DGGML_AVX2=ON -DGGML_AVX_VNNI=ON -DGGML_AVX_BMI=ON -DGGML_FMA=ON \
    -DGGML_SSE42=ON -DGGML_F16C=ON \
    -DGGML_NATIVE=ON \

cmake --build build --config Release -t llama-server --parallel 16
```
```log
In file included from .../llama.cpp/common/./jinja/runtime.h:4:
/home/st/dev/llama.cpp/common/./jinja/value.h:132:21: warning: function 'as_int' could be declared with attribute 'noreturn' [-Wmissing-noreturn]
  132 |     virtual int64_t as_int() const { throw std::runtime_error(type() + " is not an int value"); }
      |                     ^
/home/st/dev/llama.cpp/common/./jinja/value.h:133:20: warning: function 'as_float' could be declared with attribute 'noreturn' [-Wmissing-noreturn]
  133 |     virtual double as_float() const { throw std::runtime_error(type() + " is not a float value"); }
```

I can open it as separate PR if there if there is an interest in it, but its not breaking anyway.

CISC:
> The `noreturn` warning came up using the Intel compiler on an other build:
> I can open it as separate PR if there if there is an interest in it, but its not breaking anyway.

It is breaking on non-broken compilers that don't issue that warning. :)

Changing this will change the class prototype and break build, these functions **do** return a value.

stsydow:
turns out `./build-info.h` is ignored by `.gitignore` line 47:
``` 
/build*
```
and so it does not show up in `git reset --hard`, `git status` or `git clean`.
I'm not sure if un-ignoring it is a good idea. But it is quite well hidden when updating the repo.

PR #16653, opened 2025-10-18
llama: automatically set parameters not set by the user in such a way that maximizes GPU utilization

JohannesGaessler:
This PR adds automation for setting parameters in such a way that maximizes memory utilization when the full model cannot be fit. The short version is that the code first tries reducing the context size and then starts moving weights from device memory to system memory. For MoE models dense weights are prioritized for allocation in device memory since system memory is usually slower than device memory. Example log snippet:

```
llama_params_fit: projected memory use with initial parameters [MiB]:
llama_params_fit:   - ROCm0 (AMD Radeon Graphics)     :  16304 total,  39959 used,  24341 deficit
llama_params_fit:   - ROCm1 (AMD Radeon RX 6800)      :  16368 total,  42480 used,  26296 deficit
llama_params_fit:   - ROCm2 (AMD Instinct MI60 / MI50):  32752 total,  76200 used,  43626 deficit
llama_params_fit: projected to use 158641 MiB of device memory vs. a total of 65424 MiB
llama_params_fit: cannot fulfill margin of 1024 MiB on all devices, need to use 97337 MiB less in total
llama_params_fit: context size reduced from 65536 to 4096 -> need 13440 MiB less memory
llama_params_fit: with only dense weights in device memory there is a total surplus of 53432 MiB
llama_params_fit: set to use 35 dense-only and 22 full GPU layers in total, projected memory use:
llama_params_fit:   - ROCm0 (AMD Radeon Graphics)     :  0 dense-only layers,  5 full layers,  14373 MiB used,   1244 MiB free
llama_params_fit:   - ROCm1 (AMD Radeon RX 6800)      : 13 dense-only layers,  5 full layers,  14354 MiB used,   1829 MiB free
llama_params_fit:   - ROCm2 (AMD Instinct MI60 / MI50): 22 dense-only layers, 12 full layers,  30917 MiB used,   1656 MiB free
```

## User Interface

* The llama C API has a new function `llama_params_fit` that adjusts the provided `llama_model_params` and `llama_context_params` in such a way that upon use to create a corresponding `llama_model` and `llama_context` the program will not run out of memory.
* `llama_model_params` has a new flag `no_alloc` that is false by default but results in a `llama_model` and `llama_context` with only metadata if set to true.
* New CLI argument `--fit [on|off]` to control whether parameters should be fit to free device memory, enabled by default. The overall intent is to have optimistic defaults that would require a large amount of resources and to then cut down on the use if insufficient resources are available.
[...]

PR #22171, opened 2026-04-20
fit-params : refactor + add option to output estimated memory per device

ggerganov:
## Overview

cont #16653
ref https://github.com/ggml-org/llama.cpp/pull/19070#pullrequestreview-3701824729

- Refactor the fit param logic. Move from `libllama` to `libcommon`
- Add CLI argument `-fitp, --fit-print` to the `llama-fit-params` tool. This is useful for 3rd party applications to estimate the required memory for a model.

## Additional information

Example:

```bash
llama-fit-params -m ~/models/gemma-3-4b-it/ggml-model-f16.gguf -c 32768 --fit-print on

0.00.196.882 I main: printing estimated memory in MiB to stdout (device, model, context, compute) ...
MTL0 7401 814 517 
host 1280 0 154 
```

# Requirements

- AI usage disclosure: NO


src/llama-ext.h
```diff
@@ -3,6 +3,7 @@
 #include "llama.h"
 
 #include <cstdint>
+#include <vector>
```
JohannesGaessler: We are making `llama-ext.h` a C++ only header then?

tools/fit-params/fit-params.cpp
```diff
@@ -19,49 +16,77 @@ int main(int argc, char ** argv) {
...
+            printf("\n");
         }
-        printf("%s%s=%s", itbo > 0 ? "," : "", mparams.tensor_buft_overrides[itbo].pattern, ggml_backend_buft_name(mparams.tensor_buft_overrides[itbo].buft));
-        any_tbo = true;
+        printf("Host ");
+        printf("%zu ", dmd.back().mb.model/1024/1024);
+        printf("%zu ", dmd.back().mb.context/1024/1024);
+        printf("%zu ", dmd.back().mb.compute/1024/1024);
+        printf("\n");
     }
```
JohannesGaessler: If the intent is to parse this output programmatically I think it would be preferable to use a well-defined format like JSON.

common/arg.cpp
```diff
@@ -2426,6 +2426,20 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
             }
         }
     ).set_env("LLAMA_ARG_FIT"));
+    add_opt(common_arg(
+        { "-fite", "--fit-estimate" }, "[on|off]",
+        string_format("estimate the required memory to run the model ('on' or 'off', default: '%s')", params.fit_params_est ? "on" : "off"),
```
JohannesGaessler: I think this descriptions is confusing. It is true that with this option enabled the program will estimate the required memory but that is already what it's doing anyways. Maybe it would be better to call this something like `--fit-print` (or `--fit-print-json` depending on my other comment).

tools/fit-params/fit-params.cpp
```diff
@@ -19,49 +16,77 @@ int main(int argc, char ** argv) {
...
-        params.verbosity >= 4 ? GGML_LOG_LEVEL_DEBUG : GGML_LOG_LEVEL_ERROR);
-    if (status != LLAMA_PARAMS_FIT_STATUS_SUCCESS) {
-        LOG_ERR("%s: failed to fit CLI arguments to free memory, exiting...\n", __func__);
-        exit(1);
-    }
 
-    LOG_INF("%s: printing fitted CLI arguments to stdout...\n", __func__);
-    common_log_flush(common_log_main());
-    printf("-c %" PRIu32 " -ngl %" PRIi32, cparams.n_ctx, mparams.n_gpu_layers);
+    if (!params.fit_params_est) {
```
JohannesGaessler: In principle, if the only goal is to disable the fitting this can already be done by manually setting e.g. `-c 0 -ngl 999`. The code should then recognize that these have been set manually and will not alter them. The only downside vs. the current approach would be that it's slightly slower than to only retrieve `llama_get_device_memory_data` once. But if we're already messing around with internal headers anyways we may as well make `llama_params_fit_impl` return the device memory data instead. Then all that would need to be done is change what is being printed.

src/llama-ext.h
ggerganov: This header is not required to be C-style. It is for staging new API and can be C++ if needed.

Once an API is ready to become public, it has to be become C-style.

tools/fit-params/fit-params.cpp
ggerganov: For me, JSON is a huge overhead and in this specific case I don't think it is warranted because the data that we want to output is very simple.

ggerganov:
I took the opportunity to refactor the implementation and move the param fitting logic outside of `libllama`. It's something that I suggested in the original PR (https://github.com/ggml-org/llama.cpp/pull/16653#pullrequestreview-3561762159) and I think it makes sense to have this logic in `libcommon` for now. IMO the param fitting API on `master` is too narrow and does not allow flexibility for the user code. The idea is to eventually expose a much more lightweight and flexible API that would allow the applications to implement more sophisticated logic for param fitting and memory queries.

The prototype of this API is now in `llama-ext.h`:

https://github.com/ggml-org/llama.cpp/blob/7681f17b08dec07edd7230f0f6af6bcec9c713dd/src/llama-ext.h#L61-L90

JohannesGaessler (approved):
Please also add me as a code owner for `fit.cpp`.

CISC:
Windows build and release broken:
https://github.com/ggml-org/llama.cpp/actions/runs/24708525547/job/72267275937

PR #21808, opened 2026-04-12
TP: fix 0-sized tensor slices, AllReduce fallback

JohannesGaessler:
Partially fixes https://github.com/ggml-org/llama.cpp/issues/21765 .

With Qwen 3.5 ~~26b a4b~~ 27b there are only 2 KV heads so with 3+ GPUs some of them will get zero-sized slices of the data. This edge case is not being handled correctly on master. This PR makes it so that the corresponding nodes are disabled and the buffer for the AllReduce memset to 0 so that after the AllReduce all GPUs have the correct data. As of right now the buffer is zeroed out via `GGML_SCALE` with a factor of `0.0f` for the AllReduce fallback implementation - this is not safe w.r.t. NaNs but it seems we currently lack the tooling to properly memset a tensor as part of a `ggml_cgraph`. The same issue is present in `llm_graph_context::build_rs`.

Additionally, on master the synchronization of 3+ GPUs is not being handled correctly for the AllReduce fallback. The problem is that in those cases 2+ reduction steps are needed but the same buffer is used for each step so there are race conditions. This PR extends the number of buffers accordingly.

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: No

PR #22173, opened 2026-04-20
cuda: guard empty matmul shards and skip zero-sized RS no-ops

nisparks:
Add early guards in the CUDA matmul entry points so empty work does not fall through into kernel setup. Also skip zero-sized recurrent-state branches in build_rs().

## Overview

   1. adds early guards in the CUDA matmul paths so empty/zero-element shards are skipped instead of falling through into normal kernel setup, and
   2. makes build_rs() skip zero-sized recurrent-state no-op branches.

# Requirements

yes
- AI usage disclosure: 
yes. As part of a larger performance tuning, these low hanging fruit optimizations were added. Basically removes no-op steps and hardens the code path.


am17an:
Do you see any performance benefit from this? How much does this happen anyway?

nisparks:
No measurable performance increase to be found here, just early exit when no work is to be done. When running split tensors and MoE, this path would hit a few times with no-ops. Just a small cleanup/hardening.

JohannesGaessler:
Should no longer be needed with https://github.com/ggml-org/llama.cpp/pull/21808 .

PR #22170, opened 2026-04-20
Optimize CUDA matmul for empty shards and MMQ dispatch

nisparks:
## Overview

Increases P2P prompt by ~30-50% on Dual 3090 w/ NVLink on MoE, net neutral otherwise. 

## Additional information

- llama-context.cpp passes disable_mmq_stream_k_default=1 only when split_mode == tensor &&   n_expert > 0.
- For ordinary dense / single-GPU cases, the standard stream-k path stays effectively upstream-style.

Why: stream-k splits the K work across SMs and then has to merge partial tiles in a fixup step. That works fine in many normal cases, but in tensor-split MoE the work is already irregular and sparse because rows are routed per expert and split across GPUs. In this workload, the stream-k decomposition/fixup overhead was benchmarking worse than straight xy tiling. The faster combination was:
   - disable stream-k for that specific workload, and
   - use cp.async to make the remaining tile loads cheaper.

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
  yes
- AI usage disclosure:
  Yes. Been burning the midnight oil (and the unlimited tokens via benefit from my employer) to read CUDA Docs, try various heuristics, profile runs with nsys, and try out many different things. Importantly, I made sure that we didn't experience regressions on the non-P2P or Multi-GPU path.

Important to note, I went through many iterations with AI assistance, reviewed code and decisions, benchmarked locally over and over, and tested to be sure no performance was lost on non-P2P/MoE workloads.

| Scenario | Model | Metric | Upstream | Full changes | Delta | Notes |
|---|---|---:|---:|---:|---:|---|
| **1x3090** | Qwen3.5-27B IQ4_NL | prompt | 1344.37 | 1345.18 | **+0.06%** | effectively neutral|
| **1x3090** | Gemma4-31B IQ4_NL | prompt | 1360.80 | 1361.33 | **+0.04%** | effectively neutral |
| **1x3090** | Qwen3.6-35B-A3B IQ4_NL-pure | prompt | 3254.25 | 3254.02 | **-0.01%** | effectively neutral |
| **1x3090** | Qwen3.6-35B-A3B IQ4_NL-pure | decode | 176.08 | 176.47 | **+0.22%** | effectively neutral |
| **2x3090, tensor split** | Qwen3.6-35B-A3B IQ4_NL-pure | prompt | 3021.22 | 3886.13 | **+28.63%** | main win |
[...]

PR #22174, opened 2026-04-20
cuda: disable MMQ stream-k by default for MoE

nisparks:
## Overview

As the title says, I found that MMQ stream-k was inefficient with P2P + MoE. This disables by default when this scenario occurs. Tests all pass, and in MoE workloads, I see a 9.67% performance increase in prompt on my dual 3090 + NVLink setup.

## Additional information

   - Qwen3.6 dual-3090 tensor-split prompt: 2597.04 -> 2848.26 (+9.67%)
   - Single-GPU sanity set: effectively neutral (Qwen3.6-35B-A3B prompt -0.005%, decode -0.053%, Qwen3.5-35B-A3B prompt +0.043%, Gemma4-31B prompt +0.073%)
   - test-llama-archs passed
   - test-backend-ops -b CUDA0 -o MUL_MAT,MUL_MAT_ID passed (1858/1858)

  What it does

   1. ggml-cuda.cu / common.cuh store that flag in the CUDA backend context.
   2. mmq.cu uses that flag when deciding the MMQ use_stream_k default.
   3. mmq.cuh adds a compile-time use_stream_k=true/false specialization so the standalone non-stream-k path is stable and efficient.

  Why we need it

   - Stream-k is a good generic MMQ default, but for multi-GPU tensor-split MoE prompt work it is the wrong default.

  Performance
| GPU | Model | Microbatch size | Test | t/s 4eac5b450 | t/s 42064fb53 | Speedup |
| --- | --- | ---: | --- | ---: | ---: | ---: |
| 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (tensor) | 2812.48 | 3117.56 | 1.11 |
| 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (row) | 1949.66 | 2126.91 | 1.09 |
| 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (layer) | 4362.49 | 5310.48 | 1.22 |
| RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (none) | 3200.05 | 3668.18 | 1.15 |
| RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | tg128 (none) | 136.14 | 165.75 | 1.22 |

# Requirements

Yes
- AI usage disclosure: 
Yes. Per my other PR https://github.com/ggml-org/llama.cpp/pull/22170, I have been working with AI to profile and fix bottlenecks in workloads I come across. This was one of those bottlenecks. Will have more standalone improvements after, but per the recommendation in that PR, breaking this up into smaller, standalone chunks.

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1354,6 +1354,7 @@ struct ggml_backend_cuda_context {
     int device;
     std::string name;
     cudaEvent_t copy_event = nullptr;
+    bool disable_mmq_stream_k_default = false;
 
     cudaStream_t streams[GGML_CUDA_MAX_DEVICES][GGML_CUDA_MAX_STREAMS] = { { nullptr } };
     cublasHandle_t cublas_handles[GGML_CUDA_MAX_DEVICES] = {nullptr};
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -4719,9 +4719,14 @@ static void ggml_backend_cuda_device_get_props(ggml_backend_dev_t dev, ggml_back
 }
 
 static ggml_backend_t ggml_backend_cuda_device_init_backend(ggml_backend_dev_t dev, const char * params) {
-    GGML_UNUSED(params);
+    const bool disable_mmq_stream_k_default = params != nullptr && strstr(params, "disable_mmq_stream_k_default=1") != nullptr;
     ggml_backend_cuda_device_context * ctx = (ggml_backend_cuda_device_context *)dev->context;
-    return ggml_backend_cuda_init(ctx->device);
+    ggml_backend_t backend = ggml_backend_cuda_init(ctx->device);
+    if (backend != nullptr) {
+        ggml_backend_cuda_context * backend_ctx = (ggml_backend_cuda_context *) backend->context;
+        backend_ctx->disable_mmq_stream_k_default = disable_mmq_stream_k_default;
+    }
+    return backend;
 }
 
 static ggml_backend_buffer_type_t ggml_backend_cuda_device_get_buffer_type(ggml_backend_dev_t dev) {
--- a/ggml/src/ggml-cuda/mmq.cu
+++ b/ggml/src/ggml-cuda/mmq.cu
@@ -118,8 +118,9 @@ void ggml_cuda_mul_mat_q(
     const int64_t s03 = src0->nb[3] / ts_src0;
     const int64_t s3  =  dst->nb[3] / ts_dst;
 
-    const bool use_stream_k = (GGML_CUDA_CC_IS_NVIDIA(cc) && ggml_cuda_highest_compiled_arch(cc) >= GGML_CUDA_CC_VOLTA)
-                            || GGML_CUDA_CC_IS_CDNA(cc);
+    const bool use_stream_k_default = ((GGML_CUDA_CC_IS_NVIDIA(cc) && ggml_cuda_highest_compiled_arch(cc) >= GGML_CUDA_CC_VOLTA)
+                                    || GGML_CUDA_CC_IS_CDNA(cc))
+                                    && !ctx.disable_mmq_stream_k_default;
 
     // TODO: tighter pool buffer size vs q8 path
     const bool use_native_mxfp4 = blackwell_mma_available(cc) && src0->type == GGML_TYPE_MXFP4;
@@ -158,7 +159,7 @@ void ggml_cuda_mul_mat_q(
             ne00, ne01, ne1, s01, ne11, s1,
             ne02, ne12, s02, s12, s2,
             ne03, ne13, s03, s13, s3,
-            use_stream_k, ne1};
+            use_stream_k_default, ne1};
         ggml_cuda_mul_mat_q_switch_type(ctx, args, stream);
         return;
     }
@@ -218,7 +219,7 @@ void ggml_cuda_mul_mat_q(
         ne00, ne01, ne_get_rows, s01, ne_get_rows, s1,
         ne02, ne02, s02, s12, s2,
         ne03, ne13, s03, s13, s3,
-        use_stream_k, ne12};
+        use_stream_k_default, ne12};
 
     ggml_cuda_mul_mat_q_switch_type(ctx, args, stream);
 }
@@ -250,8 +251,9 @@ void ggml_cuda_op_mul_mat_q(
     // The stream-k decomposition is only faster for recent NVIDIA GPUs.
     // Also its fixup needs to allocate a temporary buffer in the memory pool.
     // There are multiple parallel CUDA streams for src1_ncols != ne11 which would introduce a race condition for this buffer.
-    const bool use_stream_k = ((GGML_CUDA_CC_IS_NVIDIA(cc) && ggml_cuda_highest_compiled_arch(cc) >= GGML_CUDA_CC_VOLTA)
-                            || GGML_CUDA_CC_IS_CDNA(cc))
+    const bool use_stream_k = (((GGML_CUDA_CC_IS_NVIDIA(cc) && ggml_cuda_highest_compiled_arch(cc) >= GGML_CUDA_CC_VOLTA)
+                             || GGML_CUDA_CC_IS_CDNA(cc))
+                            && !ctx.disable_mmq_stream_k_default)
                             && src1_ncols == ne11;
     const mmq_args args = {
         src0_dd_i, src0->type, (const int *) src1_ddq_i, nullptr, nullptr, dst_dd_i,
--- a/ggml/src/ggml-cuda/mmq.cuh
+++ b/ggml/src/ggml-cuda/mmq.cuh
@@ -3463,7 +3463,7 @@ static __device__ __forceinline__ void mul_mat_q_process_tile(
 
 // The mul_mat_q kernel implements "stream-k" work partitioning as described in https://arxiv.org/abs/2301.03598
 
-template <ggml_type type, int mmq_x, bool need_check>
+template <ggml_type type, int mmq_x, bool need_check, bool use_stream_k>
 #if defined(GGML_USE_HIP)
 #if defined(RDNA4) || defined(RDNA3) || defined(RDNA2) || defined(CDNA) || defined(GCN)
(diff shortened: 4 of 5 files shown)


nisparks:
@JohannesGaessler, here is a smaller change.

JohannesGaessler (commented):
The decision on whether or not to use stream-k should be done using only tensor properties such as the shape and explicitly not depend on `--split-mode`.

nisparks:
Let me benchmark the other options now just to make sure those paths don't regress if I remove that gate.

JohannesGaessler:
A few days ago I started working on reducing the stream-k overhead in MMQ. I don't have Qwen 3.6 or 2x RTX 3090 handy but on a single RTX 3090 I get this for Qwen 3.5:

| GPU      | Model                  |   Microbatch size | Test   |   t/s a6cc43c28 |   t/s a1760869a |   Speedup |
|:---------|:-----------------------|------------------:|:-------|----------------:|----------------:|----------:|
| RTX 3090 | qwen35moe 35B.A3B Q4_0 |                16 | pp2048 |          650.60 |          745.07 |      1.15 |
| RTX 3090 | qwen35moe 35B.A3B Q4_0 |                32 | pp2048 |          929.85 |         1042.95 |      1.12 |
| RTX 3090 | qwen35moe 35B.A3B Q4_0 |                64 | pp2048 |         1266.45 |         1360.01 |      1.07 |
| RTX 3090 | qwen35moe 35B.A3B Q4_0 |               128 | pp2048 |         1449.18 |         1509.50 |      1.04 |
| RTX 3090 | qwen35moe 35B.A3B Q4_0 |               256 | pp2048 |         2158.26 |         2241.32 |      1.04 |
| RTX 3090 | qwen35moe 35B.A3B Q4_0 |               512 | pp2048 |         2958.78 |         3066.14 |      1.04 |
| RTX 3090 | qwen35moe 35B.A3B Q4_0 |              1024 | pp2048 |         3667.64 |         3825.24 |      1.04 |
| RTX 3090 | qwen35moe 35B.A3B Q4_0 |              2048 | pp2048 |         4148.81 |         4347.48 |      1.05 |

Please check the performance of a1760869a vs. your changes, I created the table above using these commands (adapt to your environment):

```
./bench --model models/opt/${mn}-${q}.gguf -r 10 -fa 1 -n 0 -p 2048 -ub "16-2048*2" -o sql|sqlite3 llama-bench.sqlite 
py scripts/compare-llama-bench.py -s gpu_info,model_type,n_ubatch -i llama-bench.sqlite -b a6cc43c286a2ebc42 -c a1760869a8c49c5e5|tee bench.txt
```

nisparks:
@JohannesGaessler, So I see similar results, and this slices into some of the gains I saw with dual GPU, but I still see gains on mine when compared.
image

When I ran the benchmarks for an ungated (no tensor split mode check), row had a 0.1% regression, and layer had a 0.28% regression on prompt, averaged over 6 runs.

Going to see if adding your changes fix the regression I saw. I know it generally improves things, but I didn't profile the losses to row/layer yet.

nisparks:
Also that commit seems to crash when I use split-mode row.

nisparks:
Looks like I don't need to gate on tensor split mode. Will be removing that now.

nisparks:
Pushed update. Actually see large improvements on layer and row split, as well as none. Running the ./bench commands now for an A/B comparison.

nisparks:
Initial results with Qwen3.6 IQ4 on Dual 3090s, still waiting for the other results. This is the delta between the first and second commit.

| Case | Tensor-only PR2a | All-split PR2a | Delta |
| --- | ---: | ---: | ---: |
| dual tensor | 3114.66 | 3125.14 | +0.34% |
| dual row | 1949.21 | 2129.35 | +9.24% |
| dual layer | 4361.89 | 5315.90 | +21.87% |
| single prompt | 3197.55 | 3670.41 | +14.79% |
| single decode | 134.87 | 145.64 | +7.98% |

nisparks:
| GPU | Model | Microbatch size | Test | t/s 4eac5b450 | t/s 42064fb53 | Speedup |
| --- | --- | ---: | --- | ---: | ---: | ---: |
| 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (tensor) | 2812.48 | 3117.56 | 1.11 |
| 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (row) | 1949.66 | 2126.91 | 1.09 |
| 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (layer) | 4362.49 | 5310.48 | 1.22 |
| RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (none) | 3200.05 | 3668.18 | 1.15 |
| RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | tg128 (none) | 136.14 | 165.75 | 1.22 |

ORippler (commented):
1. AFAIK, consensus so far was to keep backend-specific changes in ENV variables (we will pass backend_params to all other backends that are not cuda-based such as the vulkan backend).
2. What OS are you using (Windows vs. Linux). Maybe a stupid question, but did you set `GGML_CUDA_P2P` to opt into actual P2P access? https://github.com/ggml-org/llama.cpp/pull/21910
3. What are the implications on the binary size if we template mmq once more?

CC @gaugarg-nv

nisparks:
@ORippler - Taboo here, but I'll let the AI summary answer

1. backend_params vs env var

The flag is set programmatically from `hparams.n_expert > 0` at context init, so an env var would lose the auto-detection. It has to go through `backend_params` (not a CUDA setter) because under `-sm tensor` `model.devices` returns a single `Meta(CUDA0,CUDA1)` aggregate; the param string propagates through Meta to each child CUDA context, whereas a CUDA-specific setter only sees the Meta pointer and never reaches the children (verified — the setter approach causes a clean -10% regression). I can namespace the key as `cuda.disable_mmq_stream_k_default=1` so other backends ignore it by design.

--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

2. OS / GGML_CUDA_P2P

Linux, 2× RTX 3090 (`CUDA_VISIBLE_DEVICES=0,1`). `nvidia-smi topo -p2p r` reports OK between the two 3090s. I reran the A/B with both `GGML_CUDA_P2P=0` and `GGML_CUDA_P2P=1`:

| P2P | upstream `4eac5b450` | PR2a `42064fb53` | delta   |
|-----|----------------------|------------------|---------|
| off | 3668.92 t/s          | 4172.88 t/s      | +13.74% |
| on  | 3657.03 t/s          | 4181.62 t/s      | +14.34% |

(Qwen3.6-35B-A3B-IQ4_NL, pp2048, -fa 1, -sm tensor, -ts 1/1, 4 interleaved rounds × 5 reps each.) The P2P toggle moves either build by <0.5% — within noise — so PR2a's win is independent of P2P.

 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

3. Binary size

Measured `libggml-cuda.so` at `-DCMAKE_CUDA_ARCHITECTURES=86`:

| build                                          | size       | vs upstream |
|------------------------------------------------|------------|-------------|
| upstream `4eac5b450`                           | 57.55 MiB  | —           |
| PR2a (template axis, current)                  | 65.92 MiB  | +14.54%     |
| PR2a (runtime kernel param, no extra template) | 65.83 MiB  | +14.39%     |

Dropping the template axis to a runtime kernel param only saves ~95 KB (0.14%) — the real cost is that `mul_mat_q` now has to compile *both* code paths where upstream's `#if (__CUDA_ARCH__ < GGML_CUDA_CC_VOLTA …)` eliminated one on Volta+. If that's too much, mitigation options are: (a) drop the template axis (saves 95 KB), (b) `#if`-gate the alternate path per-arch, or (c) hide the whole feature behind a CMake opt-in (default off → identical to upstream).

ORippler:
> | GPU | Model | Microbatch size | Test | t/s 4eac5b450 | t/s 42064fb53 | Speedup |
> | --- | --- | ---: | --- | ---: | ---: | ---: |
> | 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (tensor) | 2812.48 | 3117.56 | 1.11 |
> | 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (row) | 1949.66 | 2126.91 | 1.09 |
> | 2x RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (layer) | 4362.49 | 5310.48 | 1.22 |
> | RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | pp2048 (none) | 3200.05 | 3668.18 | 1.15 |
> | RTX 3090 | qwen35moe 35B.A3B IQ4_NL | 512 | tg128 (none) | 136.14 | 165.75 | 1.22 |

From these numbers, I infer that:
1. A future PR/Feature would be to default to layer-split for prefill and tensor-split for decode on mGPU systems (prefill can be parallelized more effectively across GPUs by splitting layers, so there is less communication overhead present).
2. This should be a change that is independent of mGPU (as the single GPU-setting benefits from it also). Though for split-mode=None, tg changes are unexpected as effectively only a single GPU is used and MMQ does not get called during decode.

> single Meta(CUDA0,CUDA1) aggregate; the param string propagates through Meta to each child CUDA context, whereas a CUDA-specific setter only sees the Meta pointer and never reaches the children (verified — the setter approach causes a clean -10% regression). I can namespace the key as cuda.disable_mmq_stream_k_default=1 so other backends ignore it by design.

Hmm, I guess meta backend kind of forces some standardization on backend-specific parts of the ggml interface (and how backend-specific args are resolved would be an open architecting question. Say we have a `Meta(CUDA0, ROCm1)`, and ROCm would like to receive different init args as CUDA). But that's a bit beyond the scope of the PR in my eyes (and I may be missing something here due to unfamiliarity with the code-base).

Still, and environment variable is set by the user and should bypass all of this propagation logic.

nisparks:
@ORippler I further simplified and benchmarked again: https://github.com/ggml-org/llama.cpp/pull/22252

nisparks:
Superseded by https://github.com/ggml-org/llama.cpp/pull/22298

PR #22189, opened 2026-04-21
gguf-py: Read and write empty array.

zhujunling-nj:
## Overview

When modifying model parameters with gguf-py, you cannot copy the empty arrays from the original model.
This revision will allow copying empty arrays:
- Obtain the subtype even if the array is empty during reading.
- The write operation fails only when the array is empty and no subtype is specified.

# Requirements

- AI usage disclosure: No

### DIFF
--- a/gguf-py/gguf/gguf_reader.py
+++ b/gguf-py/gguf/gguf_reader.py
@@ -243,6 +243,9 @@ class GGUFReader:
             offs += int(alen.nbytes)
             aparts: list[npt.NDArray[Any]] = [raw_itype, alen]
             data_idxs: list[int] = []
+            # If empty array
+            if alen[0] == 0:
+                types.append(GGUFValueType(raw_itype[0]))
             # FIXME: Handle multi-dimensional arrays properly instead of flattening
             for idx in range(alen[0]):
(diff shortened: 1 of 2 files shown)


CISC:
Do you have an example of such a GGUF, and more importantly, how it was created?

zhujunling-nj:
> Do you have an example of such a GGUF, and more importantly, how it was created?

The qwen3.6 or qwen3.5 pull from Ollama.
The field "qwen35moe.vision.deepstack_visual_indexes" is an empty array.

CISC:
> > Do you have an example of such a GGUF, and more importantly, how it was created?
> 
> The qwen3.6 or qwen3.5 pull from Ollama. The field "qwen35moe.vision.deepstack_visual_indexes" is an empty array.

Ok, so thing is, that is not the right metadata field (should be mapped to either `qwen35moe.n_deepstack_layers` or `clip.vision.is_deepstack_layers` depending on GGUF type), and an empty array for it is not valid.

In fact had this GGUF been created with a working tool the check you are now (re)moving would have prevented it from doing so.

The GGUF you are trying to modify is broken, the tool used to make it is broken and the tool you are using to run it is broken, please don't break ours. :)

PR #22191, opened 2026-04-21
Add IP whitelist with CIDR support to llama-server

cabelo:
## Additional information

# Requirements

- AI usage disclosure: NO 

This PR introduces support for IP-based access control (whitelisting) in the llama-server.

A new CLI argument `--whitelist `was added, allowing users to restrict access to the HTTP server to a specific set of client IP addresses.

## Features
- New CLI argument:
`--whitelist IP1,IP2,CIDR,...`

- Environment variable support:
`LLAMA_ARG_WHITELIST`

- Middleware-based validation integrated into the existing HTTP pipeline (set_pre_routing_handler)
- Consistent behavior with existing server middleware (API key, server state)
- Returns structured JSON error when access is denied:

```JSON
{
  "error": {
    "message": "Access Denied",
    "type": "permission_error",
    "code": 403
  }
}
```

### Implementation details

- Added whitelist_ips to common_params
- Registered new argument in common/arg.cpp
- Parsed input using parse_csv_row() (consistent with existing parsing utilities)
- Implemented middleware in tools/server/server-http.cpp
- Validation occurs before API key checks, ensuring early rejection
- CIDR ranges added

```
10.0.0.0/24
192.168.0.0/16
```

### Behavior

- If `--whitelist `is **not provided** → no restriction (default behavior unchanged)
- If provided → only listed IPs are allowed
- All other requests return HTTP 403

```BASH
./llama-server \
  -m model.gguf \
  --host 0.0.0.0 \
  --port 8080 \
  --whitelist 127.0.0.1,192.168.1.12,.192.168.0.0/24
```
OR

```BASH
export LLAMA_ARG_WHITELIST="192.168.0.213"
./llama-server \
  -m model.gguf \
  --host 0.0.0.0 \
  --port 8080 \
```

### Motivation

This feature improves security for production deployments, especially when exposing the server over a network, by allowing simple and efficient network-level access control without requiring external proxies or firewalls.

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2910,7 +2910,7 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_API_KEY"));
     add_opt(common_arg(
-        {"--whitelist"}, "IP1,IP2,...",
+        {"--whitelist"}, "IP1,IP2,CIDR...",
         "comma-separated list of allowed client IP addresses",
         [](common_params & params, const std::string & value) {
             params.whitelist_ips.clear();
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2910,6 +2910,25 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_API_KEY"));
     add_opt(common_arg(
+        {"--whitelist"}, "IP1,IP2,...",
+        "comma-separated list of allowed client IP addresses",
+        [](common_params & params, const std::string & value) {
+            params.whitelist_ips.clear();
+            for (auto ip : parse_csv_row(value)) {
+                ip.erase(ip.begin(), std::find_if(ip.begin(), ip.end(), [](unsigned char ch) {
+                    return !std::isspace(ch);
+                }));
+                ip.erase(std::find_if(ip.rbegin(), ip.rend(), [](unsigned char ch) {
+                    return !std::isspace(ch);
+                }).base(), ip.end());
+
+                if (!ip.empty()) {
+                    params.whitelist_ips.push_back(ip);
+                }
+            }
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_WHITELIST"));    
+    add_opt(common_arg(
         {"--api-key-file"}, "FNAME",
         "path to file containing API keys (default: none)",
         [](common_params & params, const std::string & value) {
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -2910,6 +2910,33 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
         }
     ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_API_KEY"));
     add_opt(common_arg(
+        {"--trusted-proxy"}, "IP1,IP2,CIDR,...",
+        "comma-separated list of trusted reverse proxy IP addresses or CIDR ranges",
+        [](common_params & params, const std::string & value) {
+            params.trusted_proxy_ips.clear();
+
+            for (auto ip : parse_csv_row(value)) {
+                ip.erase(ip.begin(), std::find_if(ip.begin(), ip.end(), [](unsigned char ch) {
+                    return !std::isspace(ch);
+                }));
+                ip.erase(std::find_if(ip.rbegin(), ip.rend(), [](unsigned char ch) {
+                    return !std::isspace(ch);
+                }).base(), ip.end());
+
+                if (!ip.empty()) {
+                    params.trusted_proxy_ips.push_back(ip);
+                }
+            }
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_TRUSTED_PROXY"));
+    add_opt(common_arg(
+        {"--use-forwarded-for"},
+        "use X-Forwarded-For header only when request comes from a trusted proxy",
+        [](common_params & params) {
+            params.use_forwarded_for = true;
+        }
+    ).set_examples({LLAMA_EXAMPLE_SERVER}).set_env("LLAMA_ARG_USE_FORWARDED_FOR"));    
+    add_opt(common_arg(
         {"--whitelist"}, "IP1,IP2,CIDR...",
         "comma-separated list of allowed client IP addresses",
         [](common_params & params, const std::string & value) {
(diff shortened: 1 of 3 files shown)


wqerrewetw:
Supporting CIDR would probably be better.

cabelo:
> Supporting CIDR would probably be better.

Hi @wqerrewetw.

 I'm already working on it.
 
 But thank you very much.

cabelo:
> Supporting CIDR would probably be better.

Hi @wqerrewetw  DONE!

cabelo:
@wqerrewetw I implemented this feature to meet a professional need in a project. Following the philosophy of free software, I am offering this improvement as a humble contribution, a pragmatic suggestion, but one that I believe will be useful, because if this need arose for me, it will probably arise for other people as well.
Hugs

cabelo:
@aldehir and @ngxson Do you think this PR (Proposed Law) is worth approving?

Thanks

ServeurpersoCom:
Nice addition, this is genuinely useful for enterprise deployments where you want to restrict llama-server to an internal LAN without standing up extra infrastructure. I get that handling reverse proxies fully is out of scope, but a small X-Forwarded-For path with a trusted proxy flag would avoid the whitelist silently failing for anyone running llama-server behind a proxy. Same for IPv6, inet_pton on AF_INET only means dual stack clients always get 403.

Unrelated, lines 213 to 218 break the api_keys if/else if chain when the feature is on: the multi key api_keys log is lost when whitelist size is 1, and whitelist size 2+ is never logged.

cabelo:
> Nice addition, this is genuinely useful for enterprise deployments where you want to restrict llama-server to an internal LAN without standing up extra infrastructure. I get that handling reverse proxies fully is out of scope, but a small X-Forwarded-For path with a trusted proxy flag would avoid the whitelist silently failing for anyone running llama-server behind a proxy. Same for IPv6, inet_pton on AF_INET only means dual stack clients always get 403.
> 
> Unrelated, lines 213 to 218 break the api_keys if/else if chain when the feature is on: the multi key api_keys log is lost when whitelist size is 1, and whitelist size 2+ is never logged.


@ServeurpersoCom 
I'm thinking of resolving X-Forwarded-For like this, what do you think?

./llama-server \
  -m model.gguf \
  --host 0.0.0.0 \
  --port 8080 \
  --whitelist 192.168.1.0/24 \
  --trusted-proxy 172.17.0.0/16 \
  --use-forwarded-for

I'll also fix lines 213 to 218, I didn't notice, thank you.

Can we leave IPv6 support for another PR? Is that okay with you?

aldehir:
> @aldehir and @ngxson Do you think this PR (Proposed Law) is worth approving?
> 
> 
> 
> Thanks

@ngxson's opinion is going to be authoritative here.

Personally, I will always deploy behind a reverse proxy for production, so I can enforce ACLs there.

cabelo:
> > @aldehir and @ngxson Do you think this PR (Proposed Law) is worth approving?
> > Thanks
> 
> @ngxson's opinion is going to be authoritative here.
> 
> Personally, I will always deploy behind a reverse proxy for production, so I can enforce ACLs there.

I agree with you, but the need arose so that machines within the same network cannot access the service unless permitted.

cabelo:
Hi @aldehir, @ngxson and @ServeurpersoCom Everything is working now...

Example

./llama-server
-m model.gguf
--host 0.0.0.0
--port 8080
--whitelist 192.168.1.0/24
--trusted-proxy 172.17.0.0/16
--use-forwarded-for

ngxson:
IMO such feature should still better be handled by an external reversed proxy. It's quite strange to me if llama-server have to handle security features like this, we may risk being flooded by security advisories if it's not done right or not being maintained properly.

In most cases, supporting an api key should already robust enough for security and also require less code to be added to the project (low maintenance cost). So I don't think there is a strong need for whitelisting source IPs.

Asking @ServeurpersoCom if you agree / disagree with this.

ServeurpersoCom:
Two strictly identical functions doing the exact same thing: first_forwarded_for_ip and first_ip_from_x_forwarded_for. One needs to go.

Also, --trusted-proxy shouldn't accept CIDR. A reverse proxy is a known host, not a range. Allowing CIDR there opens a whitelist bypass if it's misconfigured with a broad range (any client in the range can forge X-Forwarded-For to spoof a whitelisted IP). Exact IP match only.

I'd like a code owner's call on whether we want this feature at all. The LAN usage justification is valid, but IP-level security wasn't really meant to live in the server (it pulls a whole thread: IPv6, edge cases, etc.). @ngxson, your call?

ServeurpersoCom:
> IMO such feature should still better be handled by an external reversed proxy. It's quite strange to me if llama-server have to handle security features like this, we may risk being flooded by security advisories if it's not done right or not being maintained properly.
> 
> In most cases, supporting an api key should already robust enough for security and also require less code to be added to the project (low maintenance cost). So I don't think there is a strong need for whitelisting source IPs.
> 
> Asking @ServeurpersoCom if you agree / disagree with this.

I think we wrote pretty much the same thing at the same time (with my Translate latency)... I agree

cabelo:
I understand, thank you all for your attention.

cabelo:
@wqerrewetw  , @ServeurpersoCom ,  @ngxson and @aldehir 

Thank you all for your attention and observations, I understand perfectly. I only intended to share a work request. And following the open-source philosophy, I decided to share it here. Below are some final considerations:

Two strictly identical functions doing exactly the same thing: `first_forwarded_for_ip` and `first_ip_from_x_forwarded_for`. One of them needs to be removed.

- I copied and pasted it while sleeping and didn't realize it.

**trusted-proxy**
- `--trusted-proxy` accept CIDR because it was a market demand that reached me, due to a client with multiple proxies on private networks. My intention was to share it here.

Best regards to all. I learned a lot from you all here.

angt:
> @wqerrewetw , @ServeurpersoCom , @ngxson and @aldehir
> 
> Thank you all for your attention and observations, I understand perfectly. I only intended to share a work request. And following the open-source philosophy, I decided to share it here. Below are some final considerations:
> 
> Two strictly identical functions doing exactly the same thing: `first_forwarded_for_ip` and `first_ip_from_x_forwarded_for`. One of them needs to be removed.
> 
> * I copied and pasted it while sleeping and didn't realize it.
> 
> **trusted-proxy**
> 
> * `--trusted-proxy` accept CIDR because it was a market demand that reached me, due to a client with multiple proxies on private networks. My intention was to share it here.
> 
> Best regards to all. I learned a lot from you all here.

Hi @cabelo, thanks for your work! You played the open source game the way it should be played.

I agree with @ngxson, ACL on CIDR is on a level where llama.cpp should not start interacting too much. There are many (many!) subtleties, and many tools already allow you to do complex ACL at this level (proxy or not).

Anyway, api-key sounds like the best and correct way to handle this, you will definitely win a lot if you manage to convince your client on that!

PR #22094, opened 2026-04-18
hip: bypass memory pool for flash attention f16 temp buffers

TheTom:
On HIP/ROCm, bypass the memory pool for flash attention f16 temp buffers to prevent OOM with quantized KV cache types.

## Overview

The legacy memory pool (`ggml_cuda_pool_leg`) retains peak-sized allocations permanently. During flash attention with quantized KV types (q4_0, q8_0), the f16 dequant temp buffers (`K_f16`, `V_f16`) grow proportional to KV cache length. After use, the pool retains them at peak size — consuming more VRAM than the KV compression saves.

On CUDA with VMM, the OS can reclaim unused virtual memory. On HIP, VMM is broken since ROCm 7.0 ([ROCm/rocm-systems#2516](https://github.com/ROCm/rocm-systems/issues/2516), open 4+ months), so all consumer RDNA 3/4 GPUs fall back to the legacy pool where allocations are never freed.

This causes quantized KV to OOM before f16 at equivalent context lengths on stock llama.cpp.

## Fix

Replace pool-based f16 temp buffer allocation with a RAII struct using raw `cudaMalloc`/`cudaFree` on HIP. Memory is released after the FA kernel completes via `cudaStreamSynchronize`.

Single file change: `ggml/src/ggml-cuda/fattn-common.cuh`, scoped to `#ifdef GGML_USE_HIP`. Zero impact on CUDA or Metal.

## Reproducer

Stock llama.cpp, no modifications needed:

```bash
# f16 survives:
llama-bench -m model.gguf -ngl 99 -fa 1 -ctk f16 -ctv f16 -p 512 -n 128 -d 0,32768,65536 -r 1

# q8_0 OOMs (same context, same model, same GPU):
llama-bench -m model.gguf -ngl 99 -fa 1 -ctk q8_0 -ctv q8_0 -p 512 -n 128 -d 0,32768,65536 -r 1
```

## Confirmed hardware

| GPU | Arch | VRAM | Tester | Result |
|-----|------|------|--------|--------|
| RX 7900 XT | gfx1100 | 20GB | ozboss | q8_0 crashed at 64K, f16 survived to 131K |
| RX 9060 XT | gfx1200 | 16GB | Gerporgl | q8_0 crashed at 31K, f16 survived to 39K |
| RX 9070 XT | gfx1201 | 16GB | TheTom, apollo-mg | Patched: q8_0 survives 65K where f16 OOMs |
[...]

PR #22193, opened 2026-04-21
cuda: add partial eviction on pool OOM

leonardHONG:
This is a small follow-up to #22155.

In #22155, legacy-pool OOM recovery was handled by flushing the whole cache and retrying once. That fixed the immediate issue, but it is also a fairly blunt fallback.

This PR changes two things.

First, in the legacy pool OOM path, it adds an LRU-style bounded reclaim step before the existing full flush. On `cudaErrorMemoryAllocation`, it evicts the oldest cached buffers first, up to a bound of 3x the current request size, and then retries the allocation. If that still fails, it falls back to the existing `clear_pool() + retry` path, so the safety net from #22155 is preserved.

Second, it adds a `bool overallocate` parameter to `ggml_cuda_pool::alloc()`. When `false`, the existing 1.05x look-ahead behavior stays unchanged. When `true`, the legacy pool allocates 2x instead. This is enabled for the FA `K_f16` / `V_f16` conversion buffers, since those buffers tend to grow as context length grows and this helps avoid repeated reallocations.

I read @JohannesGaessler's design comment and followed the requested behavior, while making two small implementation choices differently. I used a `bool overallocate` flag instead of a float lookahead parameter because the current requirements only need two modes: the default 1.05x behavior and an opt-in 2x overallocate path. If more tuning points are needed later, expanding this to a float parameter should still be straightforward. I also kept the LRU bookkeeping in a parallel `buffer_pool_ts[]` array instead of adding a `last_use` field into `ggml_cuda_buffer`, to keep the LRU bookkeeping separate from the buffer struct. Happy to refactor either of these if a float lookahead API or in-struct `last_use` is preferred.

I tested two cases: (a) legacy-pool LRU reclaim, triggered via one-shot fault injection, and (b) FA `K_f16` / `V_f16` overallocate, triggered under quantized KV cache (`-ctk q8_0 -ctv q8_0`). Detailed logs are in the comment below.

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1105,7 +1105,7 @@ int ggml_cuda_get_device();
 struct ggml_cuda_pool {
     virtual ~ggml_cuda_pool() = default;
 
-    virtual void * alloc(size_t size, size_t * actual_size) = 0;
+    virtual void * alloc(size_t size, size_t * actual_size, bool overallocate = false) = 0;
     virtual void free(void * ptr, size_t size) = 0;
 };
 
@@ -1131,16 +1131,16 @@ struct ggml_cuda_pool_alloc {
     }
 
     // size is in number of elements
-    T * alloc(size_t size) {
+    T * alloc(size_t size, bool overallocate = false) {
         GGML_ASSERT(pool != nullptr);
         GGML_ASSERT(ptr == nullptr);
-        ptr = (T *) pool->alloc(size * sizeof(T), &this->actual_size);
+        ptr = (T *) pool->alloc(size * sizeof(T), &this->actual_size, overallocate);
         return ptr;
     }
 
-    T * alloc(ggml_cuda_pool & pool, size_t size) {
+    T * alloc(ggml_cuda_pool & pool, size_t size, bool overallocate = false) {
         this->pool = &pool;
-        return alloc(size);
+        return alloc(size, overallocate);
     }
 
     T * get() {
--- a/ggml/src/ggml-cuda/fattn-common.cuh
+++ b/ggml/src/ggml-cuda/fattn-common.cuh
@@ -966,7 +966,7 @@ void launch_fattn(
         const size_t bs = ggml_blck_size(K->type);
         const size_t ts = ggml_type_size(K->type);
 
-        K_f16.alloc(ggml_nelements(K));
+        K_f16.alloc(ggml_nelements(K), /*overallocate=*/ true);
         if (ggml_is_contiguously_allocated(K)) {
             to_fp16_cuda_t to_fp16 = ggml_get_to_fp16_cuda(K->type);
             to_fp16(K_data, K_f16.ptr, ggml_nelements(K), main_stream);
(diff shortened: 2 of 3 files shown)


JohannesGaessler:
I'm not convinced that this will actually be beneficial on average.

IMbackK:
might help the slowdown reported in https://github.com/ggml-org/llama.cpp/pull/22094#issuecomment-4282741673 should be tested. But yeah this pr needs evidence.

JohannesGaessler:
I mean, even without evidence I would expect clearing only the largest buffers to be prone to memory fragmentation.

JohannesGaessler:
One variant of this PR that I would be willing to approve and merge would be something like this:

* Every time that a buffer is used, add a timestamp.
* If the memory allocation fails, clear the least recently used buffers first.
* Stop if the freed memory is 3x the requested size.
* Add an optional parameter that results in 2x overallocation for the legacy pool, enable it for the FA conversion where the size can be expected to continually increase as the context fills up. This should reduce thrashing.

leonardHONG:
Thank you for the crystal clear roadmap, @JohannesGaessler!

leonardHONG:
### 1. Legacy pool / LRU reclaim

- Environment: `GGML_CUDA_NO_VMM=1` (force legacy pool)
- Used one-shot fault injection to reliably trigger reclaim
- Key log snippet:

```text
CUDA pool[0]: injected one-shot reclaim test at 2.36 MiB
CUDA pool[0]: alloc of 2.36 MiB failed, trying LRU-style reclaim (target 7.09 MiB) from 0.94 MiB cached
CUDA pool[0]: retry succeeded after reclaiming 0.63 MiB
```

### 2. FA conversion / overallocate
- Triggered by using quantized KV cache: -ctk q8_0 -ctv q8_0
- Key log snippet (K_f16):
```text
CUDA fattn K_f16 request 0.06 MiB actual 0.30 MiB
CUDA fattn K_f16 request 0.12 MiB actual 0.12 MiB
```
- Key log snippet (V_f16):
```text
CUDA fattn V_f16 request 0.06 MiB actual 0.12 MiB
CUDA fattn V_f16 request 0.12 MiB actual 0.28 MiB
CUDA fattn V_f16 request 0.25 MiB actual 0.30 MiB
```

leonardHONG:
Thanks @IMbackK for the root-cause work in https://github.com/ROCm/rocm-systems/issues/4817 and cc @JohannesGaessler since this follows the roadmap .

For context,  #22193 follows #22155 and was opened before #22207, implementing the same direction: LRU reclaim, 3x bound, and FA overallocate. I have now pushed a HIP multi-GPU guard as a follow-up, to preserve master-equivalent `clear_pool() + retry` behavior on multi-GPU AMD while keeping the LRU path for single-GPU. My local validation so far is on CUDA single-GPU (RTX 4070 Laptop); I have not been able to validate the AMD multi-GPU path directly.

Happy to coordinate as needed and leave the merge decision to the maintainers.

sanmai:
It sounds like the reason we are at this spot because flash attention buffers grow exponentially and unless we over-allocate them in double + 1MB (what I measured) the cache is useless. So it sounds like a better fix would be to get FA their own pool with a smaller depth.

PR #21452, opened 2026-04-05
metal : add GATED_LINEAR_ATTN op

TheTom:
Add Metal backend support for `GGML_OP_GATED_LINEAR_ATTN` (GLA). Supports head_size 64 and 128, f32 only.

## Overview

Metal is currently the only major GPU backend without GLA support. Without this, RWKV6Qwen2 models fall back to CPU for this op on every layer, causing GPU/CPU sync overhead that hurts performance significantly on lower-end Apple Silicon.

The kernel follows the existing RWKV WKV6 Metal structure: one thread per head element, threadgroup shared memory for k/q/gate, and a float4-vectorized inner loop.

`supports_op` restricts execution to f32 with head sizes 64 or 128.

## Correctness

test-backend-ops (7/7 passed on both devices):

**Before (master a6cc43c):**
```
GATED_LINEAR_ATTN: not supported [MTL0] (0/0 tests)
```

**After:**
```
GATED_LINEAR_ATTN(head_size=64,  n_seq_tokens=1,   n_seqs=1): OK
GATED_LINEAR_ATTN(head_size=64,  n_seq_tokens=32,  n_seqs=1): OK
GATED_LINEAR_ATTN(head_size=64,  n_seq_tokens=32,  n_seqs=4): OK
GATED_LINEAR_ATTN(head_size=64,  n_seq_tokens=128, n_seqs=4): OK
GATED_LINEAR_ATTN(head_size=128, n_seq_tokens=1,   n_seqs=1): OK
GATED_LINEAR_ATTN(head_size=128, n_seq_tokens=32,  n_seqs=1): OK
GATED_LINEAR_ATTN(head_size=128, n_seq_tokens=32,  n_seqs=4): OK
7/7 tests passed
```

## Performance

Benchmarked with QRWKV6-7B-Instruct Q4_K_M, llama-bench, 3 runs each.

### M2 Mac Mini 32GB (Apple8)

| Test | Baseline (CPU fallback) | Metal GLA | Speedup |
|------|------------------------|-----------|---------|
| pp1 | 4.73 t/s | 25.19 t/s | **+432%** |
| pp512 | 194.88 t/s | 214.69 t/s | **+10.2%** |
| pp2048 | 200.33 t/s | 214.95 t/s | **+7.3%** |
| pp8192 | 201.46 t/s | 215.38 t/s | **+6.9%** |
| tg128 | 20.35 t/s | 24.93 t/s | **+22.5%** |

### M5 Max 128GB (Apple10)

[...]

PR #22207, opened 2026-04-21
cuda: LRU eviction + overalloc for legacy pool

### LINKED ISSUE #22107: Misc. bug: [CUDA/ROCm] VRAM leak/fragmentation in ggml_cuda_pool_leg when using Flash Attention

TheTom:
Fixes #22107. Per #22193 (comment).

On OOM, evict LRU buffers first. FA temps use 2x overalloc.
Tested on gfx1201, q8_0 @ d40000: 369 t/s (was OOM).

# Requirements

- AI usage: no

### DIFF
--- a/ggml/src/ggml-cuda/common.cuh
+++ b/ggml/src/ggml-cuda/common.cuh
@@ -1105,7 +1105,7 @@ int ggml_cuda_get_device();
 struct ggml_cuda_pool {
     virtual ~ggml_cuda_pool() = default;
 
-    virtual void * alloc(size_t size, size_t * actual_size) = 0;
+    virtual void * alloc(size_t size, size_t * actual_size, float lookahead = 1.05f) = 0;
     virtual void free(void * ptr, size_t size) = 0;
 };
 
@@ -1131,10 +1131,10 @@ struct ggml_cuda_pool_alloc {
     }
 
     // size is in number of elements
-    T * alloc(size_t size) {
+    T * alloc(size_t size, float lookahead = 1.05f) {
         GGML_ASSERT(pool != nullptr);
         GGML_ASSERT(ptr == nullptr);
-        ptr = (T *) pool->alloc(size * sizeof(T), &this->actual_size);
+        ptr = (T *) pool->alloc(size * sizeof(T), &this->actual_size, lookahead);
         return ptr;
     }
 
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -358,10 +358,12 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
     struct ggml_cuda_buffer {
         void * ptr = nullptr;
         size_t size = 0;
+        uint64_t last_use = 0;
     };
 
     ggml_cuda_buffer buffer_pool[MAX_BUFFERS] = {};
     size_t pool_size = 0;
+    uint64_t usage_counter = 0;
 
     explicit ggml_cuda_pool_leg(int device) :
         device(device) {
@@ -385,7 +387,7 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
         }
     }
 
-    void * alloc(size_t size, size_t * actual_size) override {
+    void * alloc(size_t size, size_t * actual_size, float lookahead = 1.05f) override {
 #ifdef DEBUG_CUDA_MALLOC
         int nnz = 0;
         size_t max_size = 0;
@@ -424,20 +426,48 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
             return ptr;
         }
         void * ptr;
-        size_t look_ahead_size = (size_t) (1.05 * size);
+        size_t look_ahead_size = (size_t) (lookahead * size);
         look_ahead_size = 256 * ((look_ahead_size + 255)/256);
         ggml_cuda_set_device(device);
         cudaError_t err = ggml_cuda_device_malloc(&ptr, look_ahead_size, device);
         if (err == cudaErrorMemoryAllocation) {
             (void)cudaGetLastError();
-            const size_t cached_bytes = pool_size;
-            GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: alloc of %.2f MiB failed, flushing %.2f MiB of cached buffers and retrying\n",
-                           device, look_ahead_size/1024.0/1024.0, cached_bytes/1024.0/1024.0);
+            GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: alloc of %.2f MiB failed, evicting LRU from %.2f MiB cached\n",
+                           device, look_ahead_size/1024.0/1024.0, pool_size/1024.0/1024.0);
             CUDA_CHECK(cudaDeviceSynchronize());
-            clear_pool();
+            // evict LRU buffers up to 3x the requested size
+            size_t freed = 0;
+            size_t evict_target = 3 * look_ahead_size;
+            while (freed < evict_target) {
+                int oldest = -1;
+                uint64_t oldest_use = UINT64_MAX;
+                for (int i = 0; i < MAX_BUFFERS; ++i) {
+                    if (buffer_pool[i].ptr != nullptr && buffer_pool[i].last_use < oldest_use) {
+                        oldest_use = buffer_pool[i].last_use;
+                        oldest = i;
+                    }
+                }
+                if (oldest < 0) {
+                    break;
+                }
+                ggml_cuda_buffer & b = buffer_pool[oldest];
+                CUDA_CHECK(cudaFree(b.ptr));
+                freed += b.size;
+                pool_size -= b.size;
+                b.ptr = nullptr;
+                b.size = 0;
+            }
             err = ggml_cuda_device_malloc(&ptr, look_ahead_size, device);
+            if (err == cudaErrorMemoryAllocation) {
+                // LRU eviction wasn't enough, flush everything
+                (void)cudaGetLastError();
+                GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: LRU freed %.2f MiB, still OOM, flushing all\n",
+                               device, freed/1024.0/1024.0);
+                clear_pool();
+                err = ggml_cuda_device_malloc(&ptr, look_ahead_size, device);
+            }
             if (err == cudaSuccess) {
-                GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: retry succeeded\n", device);
+                GGML_LOG_DEBUG(GGML_CUDA_NAME " pool[%d]: retry succeeded after eviction\n", device);
             }
         }
         CUDA_CHECK(err);
@@ -456,6 +486,7 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
             if (b.ptr == nullptr) {
                 b.ptr = ptr;
                 b.size = size;
+                b.last_use = ++usage_counter;
                 return;
             }
         }
@@ -499,7 +530,8 @@ struct ggml_cuda_pool_vmm : public ggml_cuda_pool {
         }
     }
 
-    void * alloc(size_t size, size_t * actual_size) override {
+    void * alloc(size_t size, size_t * actual_size, float lookahead = 1.05f) override {
+        GGML_UNUSED(lookahead);
         // round up the allocation size to the alignment to ensure that all allocations are aligned for all data types
         const size_t alignment = 128;
         size = alignment * ((size + alignment - 1) / alignment);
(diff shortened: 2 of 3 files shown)


TheTom:
> Hi @TheTom, thanks for your contribution!
> 
> 
> * **Multiple open PRs from a new contributor**: We limit new contributors (those without a previously merged PR) to **1 open PR at a time**. You currently have 3 open PRs.
> 

copied from previous PR for clarity:
One is a draft (https://github.com/ggml-org/llama.cpp/pull/21119), one has been waiting on review for 2 weeks (https://github.com/ggml-org/llama.cpp/pull/21452). This is a bug fix for an OOM affecting all HIP users with quantized KV at long context. Happy to prioritize however maintainers prefesr.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -385,7 +387,40 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
...
+            }
+            ggml_cuda_buffer & b = buffer_pool[oldest];
+            CUDA_CHECK(cudaFree(b.ptr));
+            freed += b.size;
+            pool_size -= b.size;
+            b.ptr = nullptr;
+            b.size = 0;
+        }
+        return freed;
+    }
```
JohannesGaessler: Inline this function.

ggml/src/ggml-cuda/common.cuh
```diff
@@ -1106,6 +1106,10 @@ struct ggml_cuda_pool {
     virtual ~ggml_cuda_pool() = default;
 
     virtual void * alloc(size_t size, size_t * actual_size) = 0;
+    virtual void * alloc_oversize(size_t size, size_t * actual_size, double factor) {
+        GGML_UNUSED(factor);
+        return alloc(size, actual_size);
+    }
```
JohannesGaessler: Instead of adding a new method, add new argument `float lookahead = 1.05f` to `alloc`.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -358,10 +358,12 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
     struct ggml_cuda_buffer {
         void * ptr = nullptr;
         size_t size = 0;
+        uint64_t last_used = 0;
     };
 
     ggml_cuda_buffer buffer_pool[MAX_BUFFERS] = {};
     size_t pool_size = 0;
+    uint64_t timestamp = 0;
```
```suggestion
        uint64_t last_use = 0;
    };

    ggml_cuda_buffer buffer_pool[MAX_BUFFERS] = {};
    size_t pool_size = 0;
    uint64_t usage_counter = 0;
```

What you implemented is not a timestamp but would also work. You should change the variable names though to avoid confusion.

TheTom:
comments addressed PTAL

JohannesGaessler (approved):

IMbackK (approved):

IMbackK (changes requested):
In testing i have found memory access faults caused by this pr. investigating...

TheTom:
Sounds good let me know what you find. Happy to adjust if needed.

TheTom:
Hey @IMbackK , were you able to repro or get a backtrace for me to investigate?

IMbackK:
I have already traced the problem and it looks like this makes https://github.com/ROCm/rocm-systems/issues/4817 ie failures in hipMemcpyAsync with valid source and destination parameters in multigpu scenarios, very much way more common.

There is not a whole lot we can do about this since our code is correct. At the same time pretty much breaking multigpu hip for non-fp16  fattn isent really an option either

TheTom:
> I have already traced the problem and it looks like this makes [ROCm/rocm-systems#4817](https://github.com/ROCm/rocm-systems/issues/4817) ie failures in hipMemcpyAsync with valid source and destination parameters in multigpu scenarios, very much way more common.
> 
> There is not a whole lot we can do about this since our code is correct. At the same time pretty much breaking multigpu hip for non-fp16 fattn isent really an option either

thanks for tracking this down. looks like the RC of ROCm/rocm-systems#4817 is a pre-existing hipMemcpyAsync host mapping race on multi-gpu. the LRU eviction changes free/realloc timing which exposes it more often, but the underlying bug is in the ROCm runtime as you surmise.

thoughts,on how to unblock: i can gate the LRU path behind a single-gpu check on HIP and fall back to clear_pool() for multi-gpu. that way multi-gpu HIP gets the same behavior it had before (no regression) and single-gpu HIP + all CUDA users get the fix. 

want me to push that, or do you have a different approach in mind? I also have some AMD contacts I can ask if needed

TheTom:
Added an ifdef for multigpu. Open to changes. Let me know.

IMbackK:
I dont know if there is a point to doing this. As far as i know the cuda devices will all use the VMM allocator anyhow, im not sure under what circumstances `CU_DEVICE_ATTRIBUTE_VIRTUAL_MEMORY_MANAGEMENT_SUPPORTED` is false on cuda. Maybe @JohannesGaessler can comment on if this is an in any way common case.

If this not an encountered case then i would recommend to just leave this pr open until i or amd figure out where the race in rocr/clr is exactly.

I get that its also useful for the single gpu hip case - but im not sure this justifies having the hack in the code. There is also something to be said for not doing this sort of workaround and instead pushing amd to fix their shit.

TheTom:
> I dont know if there is a point to doing this. As far as i know the cuda devices will all use the VMM allocator anyhow, im not sure under what circumstances `CU_DEVICE_ATTRIBUTE_VIRTUAL_MEMORY_MANAGEMENT_SUPPORTED` is false on cuda. Maybe @JohannesGaessler can comment on if this is an in any way common case.
> 
> If this not an encountered case then i would recommend to just leave this pr open until i or amd figure out where the race in rocr/clr is exactly.
> 
> I get that its also useful for the single gpu hip case - but im not sure this justifies having the hack in the code. There is also something to be said for not doing this sort of workaround and instead pushing amd to fix their shit.


@JohannesGaessler @IMbackK looking for guidance here.

single-gpu HIP users hit both OOM and prefill slowdown on q8_0 KV at long ctx, confirmed across gfx1100/1200/1201/906 by multiple community testers.

i don't have multi-gpu amd to verify or fix 4817. the current PR is gated to single-gpu HIP only - multi-gpu path is unchanged. happy to ping my amd contacts on 4817 in parallel to push their side, but the user-facing fix shouldn't have to wait on rocm's queue.

if the ifdef is the blocker, happy to swap to a runtime device-count check. what shape would you accept here?

TheTom:
Closing this. #23907 reserves the FA quant temp space at startup, which handles the underlying OOM more cleanly than reactive pool eviction does. Running fine on my fork in the meantime. Thanks for the reviews and the tracing work on the ROCm side.

PR #22209, opened 2026-04-21
ggml: vectorize ggml_vec_dot_q4_1_q8_1 with WASM SIMD128

sirohikartik:
## Overview
Vectorizes the inner loop of `ggml_vec_dot_q4_1_q8_1_generic` using WASM SIMD128 intrinsics. The change is gated behind `#ifdef __wasm_simd128__` so non-wasm builds are completely unaffected and fall through to the existing scalar path.

Approach:
- single `wasm_v128_load` covers all 32 packed 4-bit weights
- nibbles unpacked via AND/SHR into two u8x16 registers
- widened to i16 before multiply (WASM SIMD has no i8×i8 instruction)
- 4x `wasm_i32x4_dot_i16x8` calls accumulate all 32 element pairs
- horizontal reduce via 4x `wasm_i32x4_extract_lane`

Benchmark (node v25, emcc -O3 -msimd128, 64 blocks x QK8_1=32, 200k iterations):

| impl   | ns/call | speedup |
|--------|---------|---------|
| scalar |   880.7 |   1.00x |
| simd   |   257.8 |   3.42x |

Correctness verified against scalar reference across 10 random seeds with exact output match.

## Additional information
This optimization matters for WASM inference workloads where SIMD128 is widely supported but platform-specific backends like AVX or NEON are unavailable. The q4_1 format is commonly used in practice making this a high-value target.

# Requirements
- AI usage disclosure: YES, I used Claude Sonnet 4.6 to help design and write the WASM SIMD intrinsics and to write the standalone correctness and benchmark test script used to validate and measure the change. All code was reviewed and understood by me before submission. Final decisions on implementation approach were mine.

### DIFF
--- a/ggml/src/ggml-cpu/arch/wasm/quants.c
+++ b/ggml/src/ggml-cpu/arch/wasm/quants.c
@@ -355,6 +355,83 @@ void ggml_vec_dot_q4_0_q8_0(int n, float * GGML_RESTRICT s, size_t bs, const voi
     *s = sumf;
 }
 
+void ggml_vec_dot_q4_1_q8_1(int n, float * GGML_RESTRICT s, size_t bs, const void * GGML_RESTRICT vx, size_t bx, const void * GGML_RESTRICT vy, size_t by, int nrc) {
+    const int qk = QK8_1;
+    const int nb = n / qk;
+
+    assert(n % qk == 0);
+    assert(nrc == 1);
+    UNUSED(nrc);
+    UNUSED(bx);
+    UNUSED(by);
+    UNUSED(bs);
+
+    const block_q4_1 * GGML_RESTRICT x = vx;
+    const block_q8_1 * GGML_RESTRICT y = vy;
+
+    int ib = 0;
+    float sumf = 0;
+
+#if defined __wasm_simd128__
+    v128_t sumv = wasm_f32x4_splat(0.0f);
+    float summs = 0.0f;
+
+    for (; ib < nb; ++ib) {
+        const block_q4_1 * GGML_RESTRICT x0 = &x[ib];
+        const block_q8_1 * GGML_RESTRICT y0 = &y[ib];
+
+        summs += GGML_CPU_FP16_TO_FP32(x0->m) * GGML_CPU_FP16_TO_FP32(y0->s);
+
+        const v128_t raw  = wasm_v128_load(x0->qs);
+        const v128_t v0s  = wasm_v128_and(raw, wasm_i8x16_splat(0x0F));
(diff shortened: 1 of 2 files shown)


am17an:
This is the wrong place to add this, you should add it in `ggml/src/ggml-cpu/arch/wasm/quants.c`

sirohikartik:
> This is the wrong place to add this, you should add it in `ggml/src/ggml-cpu/arch/wasm/quants.c`

Hi, thanks for the review! I restored the generic implementation and moved the changes to the right place.

sirohikartik:
@ggerganov Hi just a heads up for the review. thanks!

ggml/src/ggml-cpu/arch/wasm/quants.c
```diff
@@ -355,6 +355,83 @@ void ggml_vec_dot_q4_0_q8_0(int n, float * GGML_RESTRICT s, size_t bs, const voi
...
+            const int v1 = (x[ib].qs[j] >>   4);
+            sumi0 += v0 * y[ib].qs[j];
+            sumi1 += v1 * y[ib].qs[j + qk/2];
+        }
+        int sumi = sumi0 + sumi1;
+        sumf += (GGML_CPU_FP16_TO_FP32(x[ib].d) * GGML_CPU_FP16_TO_FP32(y[ib].d)) * sumi
+              +  GGML_CPU_FP16_TO_FP32(x[ib].m) * GGML_CPU_FP16_TO_FP32(y[ib].s);
+    }
+
+    *s = sumf;
```
ggerganov: Here it is better to call the generic version in an `#else` branch. The reason is that there is no leftovers for the `ib` iterator in contrast to the `q4_0` case.

sirohikartik: Moved the generic implementation into the #else branch so the scalar loop is only compiled for non-WASM builds.

sirohikartik:
@ggerganov hi there just a reminder I made the changes u suggested

ggml/src/ggml-cpu/arch/wasm/quants.c
ggerganov: I meant that you should call `ggml_vec_dot_q4_1_q8_1_generic()` in the else branch.

sirohikartik:
Thanks, I found the existing ggml_vec_dot_q4_1_q8_1_generic() implementation in src/ggml-cpu/quants.c. I have updated the WASM backend to call it from the #else branch, matching the pattern used by the other architecture backends.

ggerganov (approved):

PR #22212, opened 2026-04-21
ggml: add initial MetaX backend integration

Dayuxiaoshui:
- add a `GGML_METAX` build option and hook it into backend registration/loading
- add `ggml-metax` build integration for MACA/cu-bridge
- add minimal MetaX compatibility shims so build/link/device enumeration work
- validate with `llama-cli --list-devices` on MetaX hardware

## Overview

This PR adds initial MetaX backend bring-up for llama.cpp (MetaX C500/C600 via MACA/cu-bridge).  
The scope is intentionally limited to first-stage enablement: configuration, build/link, and runtime device detection.  
To keep the integration stable, CUDA-only optimized paths that are not currently compatible with this stack are guarded behind compatibility handling, with fallback behavior for the MetaX path.

## Additional information

This PR is focused on minimal platform enablement rather than performance parity.  
Kernel-level tuning and broader optimization for MetaX are planned as follow-up work after baseline compatibility is established.

# Requirements

- AI usage disclosure: YES — AI was used for code navigation, compatibility scaffolding, iterative build/debug assistance, and drafting/rewording review responses; all final changes were manually reviewed and validated by me.


Dayuxiaoshui:
cc @ggerganov

JohannesGaessler (changes requested):
Without an explanation of the purpose of the CUDA backend changes they are not acceptable.

Dayuxiaoshui:
@JohannesGaessler 
### Why touching `ggml-cuda` is necessary

MetaX support in this PR is an initial bootstrap built on the existing CUDA-compatible path, so a small set of guarded changes in `ggml-cuda` is required to prevent currently unsupported CUDA-specific kernels (MMQ/MMF/FATTN) from breaking MetaX build/link, and to force safe fallback paths so runtime device detection works; these changes are strictly scoped to MetaX compatibility for initial enablement (build/link/device-list), while kernel-level optimization and broader performance tuning will be handled in follow-up PRs.

Dayuxiaoshui:
### Why add MetaX support

MetaX support is added to enable llama.cpp on MetaX (沐曦) GPUs (for example, C500/C600) through the MACA CUDA-compatible stack, so users on this hardware can build and run llama.cpp natively instead of relying on unsupported or indirect paths.

IMbackK:
isent that what the (lacking an active maintainer) musa backend is for allready...

Dayuxiaoshui:
@IMbackK MUSA and MetaX are different vendor stacks (Moore Threads vs MetaX/MACA).  
The existing MUSA backend cannot run on MetaX MACA toolchain/runtime, so a separate MetaX adaptation path is required.

IMbackK:
Ok but that dose not explain why we need 2 backends for the same (right?) hardware. or why Moore Threads maintains two stacks (if in understand correctly), is musa , which is really just an  fork of old hip, to be replaced by this?

Dayuxiaoshui:
@IMbackK Thanks for raising this.  
These are not two backends for the same vendor hardware:

- `GGML_METAX` in this PR targets **MetaX (沐曦)** GPUs (e.g. C500/C600) via the MACA/cu-bridge stack.
- `GGML_MUSA` targets a **different vendor stack/hardware**.

So this PR is adding support for a separate vendor platform, not replacing MUSA and not duplicating support for the same hardware.

IMbackK:
> ### Why add MetaX support
> 
> MetaX support is added to enable llama.cpp on Moore Threads (沐曦) GPUs through their MACA CUDA-compatible stack, so users with this vendor hardware can build and run llama.cpp natively instead of relying on unsupported or indirect paths.

I see, then the above should read: "...on **MetaX** GPUs..." also i am going to charitably assume that all of this reads llm generated becasue autotranslation was used.

Dayuxiaoshui:
@IMbackK IThanks, good point. I corrected it to “on MetaX GPUs”.  
And yes, some text came from auto-translation — I’ll rewrite it manually to make it clearer and more natural.

JohannesGaessler:
For MetaX support it should either be done as a drop-in replacement for CUDA with a vendor-specific header or as a completely separate backend but not both.

ORippler:
> For MetaX support it should either be done as a drop-in replacement for CUDA with a vendor-specific header or as a completely separate backend but not both.

I share this sentiment, though my personal feeling is also that having the CUDA backend transpileable to ROCm and MUSA already adds enough complexity as is. This could change if we get a maintainer for MUSA/MetaX paths (when searching for stuff online I consistently failed to find english sources I could consume when struggling with transpilation issues in the CI in the past)

JohannesGaessler:
Generally speaking I am in favor of broad hardware support. For HIP the additional complexity is only as high as it is because there is also an attempt to make the performance good. For something like MUSA the additional complexity has so far been negligible but I also consider it to add relatively little value since Moore Threads GPUs have very poor availability.

Previously we had a regular contributor from Moore Threads that did large parts of the maintenance. For now I am willing to keep maintaining the code until a non-trivial problem occurs. I may reconsider this if there ever is a problem that would require a non-negligible amount of work.

Dayuxiaoshui:
@ORippler @JohannesGaessler JIf possible, I would like to participate in the maintenance of llama.cpp in metax.

Dayuxiaoshui:
@JohannesGaessler Okay, I will make the necessary corrections and resubmit.

Dayuxiaoshui:
cc @JohannesGaessler @IMbackK

am17an:
The recent changes with the stub can quickly go out of date, I don't think you can rely on the ggml-cuda internal functions/structs to have a stable interface. You can probably use the `FORCE_CUBLAS` flag to avoid it

Dayuxiaoshui:
> The recent changes with the stub can quickly go out of date, I don't think you can rely on the ggml-cuda internal functions/structs to have a stable interface. You can probably use the flag to avoid it`FORCE_CUBLAS`

Indeed, this problem does exist. I will make another revision and submit it again. Thank you for the guidance.

am17an:
By using `GGML_CUDA_FORCE_CUBLAS=1` compile flag I think you can avoid most/all of the mmq/mmvq stubs. For FA, you just need to add `GGML_CUDA_NO_FA` as a compile flag.

Dayuxiaoshui:
cc @am17an

am17an:
This is not what I meant, I meant that when you compile ggml-metax backend you can set these two flags to get the desired result. So the change should be in your CMakeLists.txt for this backend, and no other changes if I'm understanding this correctly.

Dayuxiaoshui:
> This is not what I meant, I meant that when you compile ggml-metax backend you can set these two flags to get the desired result. So the change should be in your CMakeLists.txt for this backend, and no other changes if I'm understanding this correctly.

At present, should it be like this?

am17an:
Yes, but you still don't need any of the files under `ggml-metax/ggml-cuda`, those can be deleted right? in  `ggml-metax/CMakeLists.txt` if you specify `GGML_CUDA_FORCE_CUBLAS=1` and `GGML_CUDA_NO_FA=1`, they would be exactly what you want?

Dayuxiaoshui:
cc @am17an

IMbackK:
This looks absolutely fine now, and probably wont get in the way too mutch. I do think that a build of this needs to be added to https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/build.yml so that its at least compile tested regularly.

am17an (commented):
If this works for you, then this is good. Can you show the output for a `llama-completion` command using a MetaX card and if possible a `llama-bench` command? Also you might need to add yourself in `docs/ops.md`

Dayuxiaoshui:
> If this works for you, then this is good. Can you show the output for a `llama-completion` command using a MetaX card and if possible a `llama-bench` command? Also you might need to add yourself in `docs/ops.md`

Everything compiles and links fine — ggml-metax.so, llama-cli, llama-completion, and test-backend-ops all build cleanly. I’m using a small build-time Python patch under ggml/src/ggml-metax/ to force mul_mat_f and mul_mat_vec_q back to mcBLAS, so the shared ggml-cuda/ files stay untouched. End-to-end inference works on a MetaX C500: with a Qwen2.5-0.5B model I’m seeing ~1479 prompt t/s and ~173 gen t/s for FP16, and ~1137 / ~88 t/s for Q4_K_M. I’ve also regenerated docs/ops.md so the MetaX column is in there. The main rough edges are that mxcc doesn’t support the PTX mma asm, so the custom mmf/mmq kernels and FlashAttention are disabled for now; CUDA graphs throw a harmless but noisy mcGraphExecUpdateFailure warning; __dp4a isn’t available so quantized paths hit the software fallback; and a few mmq-instance files take ages to compile (q2_k is ~30 min). I didn’t build llama-bench this round but I can add it if you want the standard numbers.

Dayuxiaoshui:
> This looks absolutely fine now, and probably wont get in the way too mutch. I do think that a build of this needs to be added to https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/build.yml so that its at least compile tested regularly.

No problem,I will do this

am17an:
Okay this monkey patch file you created is extremely fragile, it may even break by the time this PR is merged. Could you let me know what exactly you want to set to false, I can think we can add that in when GGML_METAX is true

Dayuxiaoshui:
> Okay this monkey patch file you created is extremely fragile, it may even break by the time this PR is merged. Could you let me know what exactly you want to set to false, I can think we can add that in when GGML_METAX is true

Fair point — the regex patch is definitely brittle. Here's exactly what we need gated on GGML_USE_METAX: in ggml_cuda_mul_mat, force use_mul_mat_f and use_mul_mat_vec_q to false after they're declared; in ggml_cuda_mul_mat_id, skip the direct ggml_cuda_mul_mat_vec_q call and the ggml_cuda_mul_mat_f path; and in ggml_cuda_should_fuse_mul_mat_vec_q, just return false early. The MMA-based mmf and mmvq kernels compile to NO_DEVICE_CODE on mxcc, so we have to fall back to mcBLAS. If you're happy to add those #ifdef guards directly in ggml-cuda.cu, I'll drop the Python patch and the CMake hack entirely — much cleaner that way.

am17an:
could you show us the ggml-cuda.cu patch?

Dayuxiaoshui:
@am17an

```python
#!/usr/bin/env python3
"""Patch ggml-cuda.cu for MetaX: disable custom kernels that require MMA PTX or have memory issues."""

import sys

PATCHES = [
    # Patch 1: In ggml_cuda_mul_mat, force use_mul_mat_f and use_mul_mat_vec_q to false on MetaX
    (
        """    bool use_mul_mat_q     = ggml_is_quantized(src0->type) && !bad_padding_clear
        && src1->type == GGML_TYPE_F32 && dst->type == GGML_TYPE_F32;

    bool any_gpus_with_slow_fp16 = false;""",
        """    bool use_mul_mat_q     = ggml_is_quantized(src0->type) && !bad_padding_clear
        && src1->type == GGML_TYPE_F32 && dst->type == GGML_TYPE_F32;
#if defined(GGML_USE_METAX)
    use_mul_mat_f = false;     // MetaX: mmf kernel requires MMA PTX
    use_mul_mat_vec_q = false; // MetaX: mmvq kernel has memory alignment issues
#endif

    bool any_gpus_with_slow_fp16 = false;"""
    ),
    # Patch 2: In ggml_cuda_mul_mat_id, skip mul_mat_vec_q direct call on MetaX
    (
        """                if (ne2 <= mmvq_mmid_max) {
                    ggml_cuda_mul_mat_vec_q(ctx, src0, src1, ids, dst);
                    return;
                }""",
        """                if (ne2 <= mmvq_mmid_max) {
#if !defined(GGML_USE_METAX)
                    ggml_cuda_mul_mat_vec_q(ctx, src0, src1, ids, dst);
                    return;
#endif
                }"""
    ),
    # Patch 3: In ggml_cuda_mul_mat_id, skip mmf path on MetaX
    (
        """        if (ggml_cuda_should_use_mmf(src0->type, cc, WARP_SIZE, src0->ne, src0->nb, src1->ne[2], /*mul_mat_id=*/true)) {
            ggml_cuda_mul_mat_f(ctx, src0, src1, ids, dst);
            return;
        }""",
        """#if !defined(GGML_USE_METAX)
        if (ggml_cuda_should_use_mmf(src0->type, cc, WARP_SIZE, src0->ne, src0->nb, src1->ne[2], /*mul_mat_id=*/true)) {
            ggml_cuda_mul_mat_f(ctx, src0, src1, ids, dst);
            return;
        }
#endif"""
    ),
    # Patch 4: In ggml_cuda_should_fuse_mul_mat_vec_q, return false on MetaX
    (
        """static bool ggml_cuda_should_fuse_mul_mat_vec_q(const ggml_tensor * tensor) {
    ggml_tensor *       src0 = tensor->src[0];
    ggml_tensor *       src1 = tensor->src[1];
    const ggml_tensor * dst  = tensor;

    const bool bad_padding_clear = ggml_backend_buffer_get_usage(src0->buffer) == GGML_BACKEND_BUFFER_USAGE_COMPUTE &&
                                   ggml_nbytes(src0) != ggml_backend_buffer_get_alloc_size(src0->buffer, src0) &&
                                   src0->view_src;

    bool use_mul_mat_vec_q = ggml_is_quantized(src0->type) && !bad_padding_clear && src1->type == GGML_TYPE_F32 &&
                             dst->type == GGML_TYPE_F32 && src1->ne[1] <= MMVQ_MAX_BATCH_SIZE;""",
        """static bool ggml_cuda_should_fuse_mul_mat_vec_q(const ggml_tensor * tensor) {
#if defined(GGML_USE_METAX)
    GGML_UNUSED(tensor);
    return false; // MetaX: mmvq kernel has memory alignment issues
#else
    ggml_tensor *       src0 = tensor->src[0];
    ggml_tensor *       src1 = tensor->src[1];
    const ggml_tensor * dst  = tensor;

    const bool bad_padding_clear = ggml_backend_buffer_get_usage(src0->buffer) == GGML_BACKEND_BUFFER_USAGE_COMPUTE &&
                                   ggml_nbytes(src0) != ggml_backend_buffer_get_alloc_size(src0->buffer, src0) &&
                                   src0->view_src;

    bool use_mul_mat_vec_q = ggml_is_quantized(src0->type) && !bad_padding_clear && src1->type == GGML_TYPE_F32 &&
                             dst->type == GGML_TYPE_F32 && src1->ne[1] <= MMVQ_MAX_BATCH_SIZE;"""
    ),
]

# Also need to close the #else block in should_fuse_mul_mat_vec_q
CLOSE_ELSE_PATCH = (
    """    return use_mul_mat_vec_q;
}

static void ggml_cuda_mul_mat""",
    """    return use_mul_mat_vec_q;
#endif // defined(GGML_USE_METAX)
}

static void ggml_cuda_mul_mat"""
)


def main():
    if len(sys.argv) != 3:
        print(f"Usage: {sys.argv[0]} <input> <output>", file=sys.stderr)
        sys.exit(1)

    in_path, out_path = sys.argv[1:3]

    with open(in_path, "r") as f:
        content = f.read()

    for old, new in PATCHES:
        if old not in content:
            print(f"ERROR: Patch target not found in {in_path}", file=sys.stderr)
            print(f"Looking for:\n{old[:100]}...", file=sys.stderr)
            sys.exit(1)
        content = content.replace(old, new, 1)

    # Close the #else block
    old, new = CLOSE_ELSE_PATCH
    if old not in content:
        print(f"ERROR: Close-else patch target not found", file=sys.stderr)
        sys.exit(1)
    content = content.replace(old, new, 1)

    with open(out_path, "w") as f:
        f.write(content)

    print(f"Patched {in_path} -> {out_path}")


if __name__ == "__main__":
    main()

**Comment by am17an** [comment]
You should add these checks inside `should_use*` functions, for example 

https://github.com/ggml-org/llama.cpp/blob/82209efb7eab9a741923897c74fbb8fd71cd17ba/ggml/src/ggml-cuda/mmf.cu#L158-L162

Create a function `GGML_CUDA_CC_IS_METAX()` which should return true for METAX, and should_use_mmvq and should_use_mmvf, should_use_mmf should return early. I think that should be ok to add.

**Review by JohannesGaessler — COMMENTED** [review]
What would be the timeline on your end to fix the broken functionality so that hacks like this would no longer be necessary?

**Comment by Dayuxiaoshui** [comment]
> What would be the timeline on your end to fix the broken functionality so that hacks like this would no longer be necessary?

I will make repairs here as soon as possible and also communicate with the mentor, but currently there is a shortage of manpower, so I might carry out the repairs after some time.

**Comment by Dayuxiaoshui** [comment]
> You should add these checks inside functions, for example`should_use*`
> 
> https://github.com/ggml-org/llama.cpp/blob/82209efb7eab9a741923897c74fbb8fd71cd17ba/ggml/src/ggml-cuda/mmf.cu#L158-L162
> 
> Create a function which should return true for METAX, and should_use_mmvq and should_use_mmvf, should_use_mmf should return early. I think that should be ok to add.`GGML_CUDA_CC_IS_METAX()`

This is indeed a good idea, I will take it into consideration.

**Comment by Dayuxiaoshui** [comment]
cc @am17an

File: ggml/src/ggml-cuda/common.cuh [hunk 1]
```diff
@@ -103,6 +103,15 @@
 #define GGML_CUDA_CC_IS_QY2(cc)      (cc >= GGML_CUDA_CC_QY2 && cc < GGML_CUDA_CC_PH1)
 #define GGML_CUDA_CC_IS_PH1(cc)      (cc >= GGML_CUDA_CC_PH1)
 
+static inline bool ggml_cuda_is_metax(int cc) {
```
am17an: this shouldn't be a function, rather something like 

``` 
#if defined(GGML_USE_METAX) 
    GGML_CUDA_CC_IS_METAX(cc) 1 
#else 
    GGML_CUDA_CC_IS_METAX(cc) 0
#endif
```

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2417,6 +2421,7 @@ static void ggml_cuda_mul_mat(ggml_backend_cuda_context & ctx, const ggml_tensor
             use_mul_mat_q           = use_mul_mat_q             && ggml_cuda_should_use_mmq(src0->type, cc, src1->ne[1], /*n_experts=*/0);
             use_mul_mat_f           = use_mul_mat_f             && ggml_cuda_should_use_mmf(src0->type, cc, warp_size, src0->ne, src0->nb, src1->ne[1], /*mul_mat_id=*/false);
             use_mul_mat_vec_f       = use_mul_mat_vec_f         && ggml_cuda_should_use_mmvf(src0->type, cc, src0->ne, src0->nb, src1->ne[1]);
+            use_mul_mat_vec_q       = use_mul_mat_vec_q         && !ggml_cuda_is_metax(cc);
```
am17an: this shouldn't be here

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2425,6 +2430,7 @@ static void ggml_cuda_mul_mat(ggml_backend_cuda_context & ctx, const ggml_tensor
         use_mul_mat_q           = use_mul_mat_q             && ggml_cuda_should_use_mmq(src0->type, cc, src1->ne[1], /*n_experts=*/0);
         use_mul_mat_f           = use_mul_mat_f             && ggml_cuda_should_use_mmf(src0->type, cc, warp_size, src0->ne, src0->nb, src1->ne[1], /*mul_mat_id=*/false);
         use_mul_mat_vec_f       = use_mul_mat_vec_f         && ggml_cuda_should_use_mmvf(src0->type, cc, src0->ne, src0->nb, src1->ne[1]);
+        use_mul_mat_vec_q       = use_mul_mat_vec_q         && !ggml_cuda_is_metax(cc);
```
am17an: shouldn't be here

ggml/src/ggml-cuda/mmvq.cu
```diff
@@ -236,6 +236,10 @@ static constexpr __host__ __device__ int get_mmvq_mmid_max_batch_rdna4(ggml_type
 
 // Host function: returns the max batch size for the current arch+type at runtime.
 int get_mmvq_mmid_max_batch(ggml_type type, int cc) {
+    if (ggml_cuda_is_metax(cc)) {
```
am17an: not required

Dayuxiaoshui:
cc @am17an

ggml/src/ggml-cuda/ggml-cuda.cu
am17an, line 2509: you should add this check here and pass the correct `cc`

Dayuxiaoshui: It's already completed, not sure if it's correct, might need to review it again.

am17an (approved):
I can't test this but it looks okay.

JohannesGaessler (commented):
Quite frankly, with the percentage of ggml that is currently broken with MetaX I think that it does not make sense to add support for it.

am17an:
We can hold off till at least dp4a is implemented

Dayuxiaoshui:
@JohannesGaessler @am17an No problem, I will participate in maintaining this line.

Dayuxiaoshui:
cc @am17an @JohannesGaessler Next week I will start as an AI inference optimization intern at Metax. I will participate whenever there are areas that need correction, and I will take on related tasks in the future.

xpxpxp2001xpxpxp:
Hello, @Dayuxiaoshui. 
I saw your PR #22212 regarding initial MetaX backend support for llama.cpp. I'm currently an intern at MetaX and have also been working on integrating and validating the MACA backend. 
We independently implemented a MACA backend based on the CUDA-compatible cu-bridge and tested it on real MetaX hardware. The implementation is available here: [feat/maca-backend](https://github.com/xpxpxp2001xpxpxp/llama.cpp/tree/feat/maca-backend)
The branch has also been rebuilt and validated from a clean GitHub source archive on a C500 system.
So far, we have validated:

1. MetaX C500 and C600-A
2. Single-device and four-device execution
3. Qwen2.5-3B-Instruct and Llama-3.2-3B-Instruct
4. F16, Q8_0 and Q4_0 GGUF models
5. Backend operator tests, llama-bench, llama-cli and llama-server
6. English and Chinese inference
7. Multi-device enumeration, per-device execution and layer splitting

We also added conservative compatibility fallbacks for unsupported CUDA-specific paths and optimized the Q8_0 decode path for MACA.
Since your PR already addresses the same backend, I would like to avoid creating a duplicate contribution. Would you be interested in collaborating on the implementation? We could also ask the maintainers whether they would prefer us to continue the work in PR #22212 or prepare a follow-up PR based on the current upstream master, while preserving proper credit for the existing work.
I would be happy to share our validation results and discuss how the two implementations could be combined.
Disclosure: AI-assisted tools were used during code navigation, debugging, and documentation work. All changes were manually reviewed and validated on real MetaX hardware.

xpxpxp2001xpxpxp:
Hello, @JohannesGaessler @am17an

1. The base branch has been rebased onto the latest master. 
`upstream master`: 74ce15741b420b8d6f12e720398458b576c51c2c.    
`base commit`: 3ed08f4ede49ef68390171279a7edfaf1cb12526.
2. Base branch link: [https://github.com/xpxpxp2001xpxpxp/llama.cpp/tree/feat/maca-backend-base](https://github.com/xpxpxp2001xpxpxp/llama.cpp/tree/feat/maca-backend-base).   
3. C500 and C600 have been rebuilt from a clean build directory and verified.
4. The `DP4A` / `Q8_0` path has been implemented and passed the corresponding tests.  
5. MACA implementation is centralized in `ggml-maca`, and only the generic vendor-policy hooks are kept for CUDA common code.   
6. Unverified NVIDIA-specific optimized paths are turned off by default.   

Should we continue with #22212 or allow submitting a new minimal base PR?

PR #22217, opened 2026-04-21
chat: fix parallel_tool_calls default setting based on model capabilities, add tests for parallel tool calls and structured outputs

pwilkin:
Until now, parallel tool calls were disabled by default even though a majority of the models support them. This turns them on if the model supports it.

## Additional information

Added tests to execute on a running server for various parallel tool call scenarios as well as for structured output.

# Requirements

- AI usage disclosure: just for generating the test cases

### DIFF
--- a/scripts/server-test-parallel-tc.py
+++ b/scripts/server-test-parallel-tc.py
@@ -701,8 +701,8 @@ TRIP_PLAN_TEST = {
         }
     ],
     "mock_tool_responses": {
-        "search_flights":     lambda _: json.dumps(_FLIGHTS_RESULT),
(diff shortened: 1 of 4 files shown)


CISC (approved):

ngxson (commented):
if this is one of the `caps`, should it be stored inside `common_chat_template` instead? i.e. the existing `common_chat_template::caps`

aldehir (approved):

pwilkin:
@ngxson I'll think about how to consolidate this.

g0t4:
nice work! I just updated and was like... holy crap, Qwen3.6 is a tool calling beast (10 tool calls at a time)

PR #22224, opened 2026-04-21
server : fix CUDA non-determinism with fixed seed (#2838)

### LINKED ISSUE #2838: CUDA non-determinism on identical requests
When layers are offloaded with CUDA, sending identical requests to the examples/server completion API returns a different response the "first time":
```
$ for x in `seq 5`; do curl -s -X POST --url 'http://miku:8080/completion' --data '{"prompt":"Some random words:","n_predict":50,"seed":1337}' | jq '.content' ; done
" hollow, glowing, tinkling, crunchy, sinking"
" apartment, blouse, bobby, carousel"
" apartment, blouse, bobby, carousel"
" apartment, blouse, bobby, carousel"
" apartment, blouse, bobby, carousel"
```

This seems cache related as responses then remain the same until a different prompt is processed, after which the differing first response occurs again:
```
$ curl -s -X POST --url 'http://miku:8080/completion' --data '{"prompt":"Building a website is as simple as","n_predict":0}' >/dev/null

$ for x in `seq 5`; do curl -s -X POST --url 'http://miku:8080/completion' --data '{"prompt":"Some random words:","n_predict":50,"seed":1337}' | jq '.content' ; done
" hollow, glowing, tinkling, crunchy, sinking"
" apartment, blouse, bobby, carousel"
" apartment, blouse, bobby, carousel"
..
```

## Expected Behaviour
Output should remain the same when parameters and seed are constant. 

## Other Observations
* Not observed with Metal offload.
* Not observed without CUDA offload (interestingly neither with small n-gpu-layers, eg. CodeLlama-34b shows this behaviour with `-ngl 3` but not with `-ngl 2`).
* Behaviour observed with both non-K and K-quants.
* The first response ("hollow, glowing.." above) is what examples/main returns with the same parameters.

## Environment
* Verified behaviour on latest master commit, compiled with `LLAMA_CUBLAS=1 make -j`
* Linux 5.15.0-79-generic x86_64
* NVIDIA 535.86.05
* CUDA 12.2
* Python 3.10.12
* GNU Make 4.3
* g++ 11.4.0

Thanks for reading! 😎

ivich123:
## Overview

Fixes #2838. When a seed is explicitly set, `llama-server` produces non-deterministic output on CUDA backends due to prompt cache reuse changing the effective batch size between evaluations, which exposes batch-size-dependent FP ordering in CUDA Flash Attention.

## Additional information

Analysis and reproduction script are in the issue discussion: [https://github.com/ggml-org/llama.cpp/issues/2838#issuecomment-4097120776](https://github.com/ggml-org/llama.cpp/issues/2838#issuecomment-4097120776)

Together with an open discussion about the trade offs that this change implies in terms of performance.

# Requirements

- AI usage disclosure: YES — Analysis and reproduction script are my own (see discussion). AI assistance (Claude) used for porting the original fix to the current code structure after rebase, extending coverage to the chunk-based cache reuse path.

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -2342,6 +2342,13 @@ private:
                                 // reuse any previously computed tokens that are common with the new prompt
                                 n_past = slot.prompt.tokens.get_common_prefix(input_tokens);
 
+                                // skip cache reuse when a seed is explicitly set to avoid non-deterministic
+                                // output from batch-size-dependent FP ordering in CUDA FA (see #2838)
+                                const bool deterministic_mode = slot.task->params.sampling.seed != LLAMA_DEFAULT_SEED;
+                                if (deterministic_mode) {
+                                    n_past = 0;
+                                }
+
                                 // if there is an alora invoked, don't cache after the invocation start
                                 if (slot.alora_invocation_start > 0) {
                                     SLT_DBG(slot, "only caching to alora invocation start (n_past = %d, alora_invocation_start = %d)\n", n_past, slot.alora_invocation_start);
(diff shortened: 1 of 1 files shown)


ggerganov:
Just use `--no-cache-prompt` if you need determinism

ivich123:
Thanks for the quick review Georgi!

Sure! completely understandable, I raised this as an open question in the issue. Does that mean the issue can be closed?

PR #22241, opened 2026-04-22
ggml-webgpu: performance-portable matmul tuning knobs

SharmaRithik:
## Overview

This PR updates the tuning knobs for the WebGPU register tiling and subgroup matmul kernels to improve performance across GPUs. These suggested knobs are based on exhaustive data collection from four GPUs: NVIDIA RTX 5080 FE, AMD Radeon RX 7900 XT, Intel Arc B580, and Apple M2. After running a performance portability analysis on the exhaustive data, we found configurations that provide better average performance while minimizing worst-case slowdowns.

Here is the table:

| Path | Metric | Default | Proposed |
|---|---|---|---|
| Register tiling | avg | 59.1% | 79.5% |
| Register tiling | worst-case | 35.6% | 44.2% |
| Subgroup matrix | avg | 73.4% | 74.4% |
| Subgroup matrix | worst-case | 19.0% | 41.0% |

The percentages in the **Default** and **Proposed** columns are each configuration's achieved GFLOPs expressed as a percentage of the best GFLOPs observed for that GPU. For example, 79.5% means that, averaged across the 4 GPUs, the proposed register tiling configuration reaches 79.5% of each GPU's own optimum. **avg** is the mean of this percentage across the 4 GPUs. **worst-case** is the lowest value seen on any single GPU. Higher is better.

To let each kernel pick its own K independently, `WEBGPU_MUL_MAT_TILE_K` is split into two defines:

- `WEBGPU_MUL_MAT_REG_TILE_K`: used by `mul_mat_reg_tile.wgsl` and `mul_mat_id.wgsl`
- `WEBGPU_MUL_MAT_SUBGROUP_TILE_K`: used by `mul_mat_subgroup_matrix.wgsl`

Proposed defaults:

- Register tile: `TILE_M=4, TILE_N=4, WG_SIZE_M=8, WG_SIZE_N=8, REG_TILE_K=8`
- Subgroup matrix: `SUBGROUP_M=2, SUBGROUP_N=4, SUBGROUP_MATRIX_M=4, SUBGROUP_MATRIX_N=2, SUBGROUP_TILE_K=32`

# Requirements

- AI usage disclosure:  no

### DIFF
--- a/ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
+++ b/ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
@@ -26,20 +26,23 @@
 // Matrix multiplication parameters
 
 // Register tiling parameters
-#define WEBGPU_MUL_MAT_TILE_M    8
-#define WEBGPU_MUL_MAT_TILE_N    8
+#define WEBGPU_MUL_MAT_TILE_M    4
+#define WEBGPU_MUL_MAT_TILE_N    4
 #define WEBGPU_MUL_MAT_WG_SIZE_M 8
 #define WEBGPU_MUL_MAT_WG_SIZE_N 8
-#define WEBGPU_MUL_MAT_TILE_K    32
+#define WEBGPU_MUL_MAT_REG_TILE_K_FLOAT 8
+#define WEBGPU_MUL_MAT_REG_TILE_K_QUANT 32
 
 // Subgroup matrix parameters
 // The number of subgroups in the M dimension
 #define WEBGPU_MUL_MAT_SUBGROUP_M        2
 // The number of subgroups in the N dimension
-#define WEBGPU_MUL_MAT_SUBGROUP_N        2
+#define WEBGPU_MUL_MAT_SUBGROUP_N        4
 // The number of subgroup matrices each subgroup accumulates over
 #define WEBGPU_MUL_MAT_SUBGROUP_MATRIX_M 4
 #define WEBGPU_MUL_MAT_SUBGROUP_MATRIX_N 2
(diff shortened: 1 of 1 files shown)


reeselevine:
What datatypes were these values tuned for? We should probably should split out the float vs. quant tile k parameters.

ggml/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
```diff
@@ -1550,13 +1504,24 @@ class ggml_webgpu_shader_lib {
         // VEC/SCALAR controls
         defines.push_back(key.vectorized ? "VEC" : "SCALAR");
 
+        const bool is_quant = ggml_is_quantized(context.src0->type);
+
+        uint32_t WEBGPU_MUL_MAT_TILE_K;
```
reeselevine: let's not mix the naming convention for constants and variables. So this variable name can just be `tile_k`.

reeselevine:
Ran llama-bench on an M3 and an RTX 4070 with these parameters and a Llama 3.2 1B q4_0 model. On the M3 I saw a 5% slowdown in prefill and no change in decode, and on the RTX 4070 I saw a 11% increase in prefill and a 8.5% increase in decode. So nice to see the parameters translating to new devices, even if there is the slight slowdown on the Apple devices!

SharmaRithik:
Thanks for the feedback, Reese! These knobs perform well across devices. Nice to see them generalize, even with the small drop on Apple. Also made the change you suggested: renamed the local from WEBGPU_MUL_MAT_TILE_K to tile_k so it follows the variable naming convention

reeselevine (approved):

CISC (approved):

PR #22247, opened 2026-04-22
Handle ModelOpt produced mixed precision model during convert to GGUF

ynankani:
## Overview

Observed some issues while converting ModelOpt produced mixed precision NvFp4 models to GGUF. Looks like the dequantization path for unsupported dtype(Fp8) is dependent on quant_method, "quant_method" is not present in hf_quant_config.json for ModelOpt exported NvFP4 checkpoint.
The hf_quant_config.json does contain producer information which specifies ModelOpt as producer. The fix for handling mixed precision model is to populate the quant_method as "modelopt" based on the producer name, if quant_method is none and the producer is modelopt.
## Additional information

See this for reference: https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4/raw/main/hf_quant_config.json 

# Requirements

- AI usage disclosure: AI was used to review the change and cleanup

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -743,12 +743,28 @@ class ModelBase:
         quant_method = (self.hparams.get("quantization_config") or {}).get("quant_method")
         quant_layers = (self.hparams.get("quantization_config") or {}).get("quantized_layers") or {}
         quant_config_file = self.dir_model / "hf_quant_config.json"
+        producer_name = None
 
         if (not quant_algo or not quant_layers) and quant_config_file.is_file():
             with open(quant_config_file, "r", encoding="utf-8") as f:
-                quant_config = json.load(f).get("quantization") or {}
+                hf_quant_config = json.load(f)
+                quant_config = hf_quant_config.get("quantization") or {}
+                producer = hf_quant_config.get("producer")
+                if isinstance(producer, dict):
+                    name = producer.get("name")
+                    if isinstance(name, str):
+                        producer_name = name.casefold()
                 quant_algo = quant_config.get("quant_algo", quant_algo)
                 quant_layers = quant_config.get("quantized_layers", quant_layers) or {}
+
+        # hf_quant_config.json (ModelOpt schema) carries quant_algo but not
+        # quant_method. If the producer is ModelOpt and .weight_scale tensors
+        # are present (e.g. lm_head FP8 in mixed NVFP4 models), route them
+        # through the existing "modelopt" branch of dequant_model.
+        if quant_method is None and producer_name == "modelopt" \
+                and any(k.endswith(".weight_scale") for k in self.model_tensors):
+            self.hparams.setdefault("quantization_config", {})["quant_method"] = "modelopt"
+            quant_method = "modelopt"
 
         # Some models use per-tensor quant_algo (e.g. "MIXED_PRECISION" with
         # per-layer NVFP4/FP8) instead of a single global "NVFP4" value.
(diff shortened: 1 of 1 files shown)


CISC:
I think you've misunderstood, the `quant_method` read here:
https://github.com/ggml-org/llama.cpp/blob/da91334735779501badee9fecc08c2c09472ec9a/convert_hf_to_gguf.py#L486-L489
comes from `config.json`, not `hf_quant_config.json`, unless of course some newer `transformers` change deletes that key?

ynankani:
> I think you've misunderstood, the `quant_method` read here:
> 
> https://github.com/ggml-org/llama.cpp/blob/da91334735779501badee9fecc08c2c09472ec9a/convert_hf_to_gguf.py#L486-L489
> 
> 
> comes from `config.json`, not `hf_quant_config.json`, unless of course some newer `transformers` change deletes that key?

this key is not present for all models see this: https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4/blob/main/config.json
Few of the models might not contain quant_method but surely, they will have producer_name if exported by ModelOpt

convert_hf_to_gguf.py
```diff
@@ -743,13 +743,29 @@ def prepare_tensors(self):
...
         if (not quant_algo or not quant_layers) and quant_config_file.is_file():
             with open(quant_config_file, "r", encoding="utf-8") as f:
-                quant_config = json.load(f).get("quantization") or {}
+                hf_quant_config = json.load(f)
+                quant_config = hf_quant_config.get("quantization") or {}
+                producer = hf_quant_config.get("producer")
+                if isinstance(producer, dict):
+                    name = producer.get("name")
+                    if isinstance(name, str):
+                        producer_name = name.casefold()
```
```suggestion
                producer = hf_quant_config.get("producer") or {}
                producer_name = (producer.get("name") or "").lower()
                if quant_method is None:
                    self.hparams.setdefault("quantization_config", {})["quant_method"] = producer_name
```
Unless `producer` is ever not a `dict` or `name` not a `str`, let's simplify, also `casefold` is overkill.

convert_hf_to_gguf.py
```diff
@@ -743,13 +743,29 @@ def prepare_tensors(self):
         quant_method = (self.hparams.get("quantization_config") or {}).get("quant_method")
         quant_layers = (self.hparams.get("quantization_config") or {}).get("quantized_layers") or {}
         quant_config_file = self.dir_model / "hf_quant_config.json"
+        producer_name = None
```
```suggestion
```

convert_hf_to_gguf.py
```diff
@@ -743,13 +743,29 @@ def prepare_tensors(self):
...
                 quant_layers = quant_config.get("quantized_layers", quant_layers) or {}
 
+        # hf_quant_config.json (ModelOpt schema) carries quant_algo but not
+        # quant_method. If the producer is ModelOpt and .weight_scale tensors
+        # are present (e.g. lm_head FP8 in mixed NVFP4 models), route them
+        # through the existing "modelopt" branch of dequant_model.
+        if quant_method is None and producer_name == "modelopt" \
+                and any(k.endswith(".weight_scale") for k in self.model_tensors):
+            self.hparams.setdefault("quantization_config", {})["quant_method"] = "modelopt"
+            quant_method = "modelopt"
```
```suggestion
```
There is no point in this check, `dequant_model` checks at the very beginning.

ynankani:
Looks like something went bad with Github after accepting suggestions

CISC:
> Looks like something went bad with Github after accepting suggestions

Yes, unfortunately this has happened more frequently lately, the GitHub UI commits mess up newlines. :(

Just update newlines to `\n` and it should be ok again.

PR #22248, opened 2026-04-22
cmake: use MATH_LIBRARY variable instead of hardcoded 'm' for ggml-base

### LINKED ISSUE #22237: Misc. bug: hardcoded 'm' prevents using custom MATH_LIBRARY in ggml-base
### Name and Version

$ ./llama-cli --version
ggml_cuda_init: found 1 ROCm devices (Total VRAM: 24517 MiB):
  Device 0: AMD Radeon RX 7900 XTX, gfx1100 (0x1100), VMM: no, Wave Size: 32, VRAM: 24517 MiB
version: 8864 (ff6b1062a)
built with Clang 17.0.6 for Linux x86_64

### Operating systems

_No response_

_No response_

### Command line

```shell

```

In `ggml/src/CMakeLists.txt`, `ggml-base` is hardcoded to link against the literal `m` library. Although `find_library(MATH_LIBRARY m)` is called earlier to detect the math library, the resulting variable is ignored in the `target_link_libraries` call.

This hardcoding prevents users from overriding the math library path via CMake variables (e.g., `-DMATH_LIBRARY=...`) without manually patching the source code. This is particularly problematic when using optimized libraries like **AMD AOCL (libalm)** where providing the specific library path is necessary to ensure optimal performance on Zen architectures.

**Location:**
https://github.com/ggml-org/llama.cpp/blob/master/ggml/src/CMakeLists.txt

```cmake
find_library(MATH_LIBRARY m)
if (MATH_LIBRARY)
  if (NOT WIN32 OR NOT DEFINED ENV{ONEAPI_ROOT})
    target_link_libraries(ggml-base PRIVATE m) # <--- Hardcoded 'm'
  endif()
endif()
```

### Proposed Fix
Update the linking logic to respect the `${MATH_LIBRARY}` variable:

```diff
-    target_link_libraries(ggml-base PRIVATE m)
+    target_link_libraries(ggml-base PRIVATE ${MATH_LIBRARY})
```

### Impact
- **No regression:** For standard environments, `${MATH_LIBRARY}` resolves to `m`, preserving existing behavior.
- **Improved flexibility:** Allows users to seamlessly integrate optimized math libraries (e.g., **AMD AOCL**) for better performance on high-performance computing environments (AOCC/AOCL).

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

Anai-Guo:
## Summary

Fixes #22237.

`find_library(MATH_LIBRARY m)` correctly detects the system math library, but the result variable `MATH_LIBRARY` was never used — the `target_link_libraries` call hardcoded the literal `m` instead.

This prevents users from overriding the math library via `-DMATH_LIBRARY=/path/to/libalm.so` (e.g., when using AMD AOCL for optimal performance on Zen architectures).

## Changes

```cmake
# Before
target_link_libraries(ggml-base PRIVATE m)

# After
target_link_libraries(ggml-base PRIVATE ${MATH_LIBRARY})
```

This is a one-line fix that makes the code consistent: the variable detected by `find_library` is now the one actually linked.

## Testing

- Default behavior is unchanged: `find_library(MATH_LIBRARY m)` still finds `libm` and links it as before.
- Users can now override with `-DMATH_LIBRARY=/path/to/libalm.so` without patching source.

### DIFF
--- a/ggml/src/CMakeLists.txt
+++ b/ggml/src/CMakeLists.txt
@@ -473,7 +473,7 @@ target_link_libraries(ggml-base PRIVATE Threads::Threads)
 find_library(MATH_LIBRARY m)
 if (MATH_LIBRARY)
     if (NOT WIN32 OR NOT DEFINED ENV{ONEAPI_ROOT})
-        target_link_libraries(ggml-base PRIVATE m)
+        target_link_libraries(ggml-base PRIVATE ${MATH_LIBRARY})
     endif()
 endif()
 


CISC:
Duplicate of #22239

PR #22252, opened 2026-04-22
cuda: disable MMQ stream-k for MoE expert matmuls

nisparks:
Stream-k decomposition assumes wide N to amortise its persistent threadblock + atomic fixup overhead. MoE expert matmuls have very small per-expert N (the routed token slice for each expert), so stream-k pays its overhead but cannot recover it through parallelism.

Gate stream-k off in ggml_cuda_mul_mat_q whenever ids != nullptr (the only callsite where MoE expert dispatch happens). Dense matmuls in ggml_cuda_op_mul_mat_q are unchanged.

Measured on Qwen3.6-35B-A3B IQ4_NL, RTX 3090 (single + 2x):

 | GPU         | Model                    | Microbatch size | Test            | t/s master | t/s patch | Speedup |
 |-------------|--------------------------|-----------------|-----------------|------------|-----------|---------|
 | RTX 3090    | qwen3moe 35B.A3B IQ4_NL  | 512             | pp2048 (none)   | 3199.48    | 3643.30   | 1.14    |
 | RTX 3090    | qwen3moe 35B.A3B IQ4_NL  | 512             | tg128  (none)   |  170.55    |  175.05   | 1.03    |
 | 2x RTX 3090 | qwen3moe 35B.A3B IQ4_NL  | 512             | pp2048 (tensor) | 2821.24    | 3097.80   | 1.10    |
 | 2x RTX 3090 | qwen3moe 35B.A3B IQ4_NL  | 512             | pp2048 (row)    | 1952.10    | 2130.55   | 1.09    |
 | 2x RTX 3090 | qwen3moe 35B.A3B IQ4_NL  | 512             | pp2048 (layer)  | 4355.36    | 5045.80   | 1.16    |

## Overview

 1. Drop the #if __CUDA_ARCH__ < VOLTA guard so both kernel bodies compile (this is what costs the +14% binary).
 2. Make the host-side decision (we use ids == nullptr).

## Additional information

Further simplification of https://github.com/ggml-org/llama.cpp/pull/22174

# Requirements

Yes
- AI usage disclosure: 
Yes, same as previous PRs. Used AI to profile and test out various configurations. I'm not a CUDA expert.

### DIFF
--- a/ggml/src/ggml-cuda/mmq.cu
+++ b/ggml/src/ggml-cuda/mmq.cu
@@ -118,8 +118,11 @@ void ggml_cuda_mul_mat_q(
     const int64_t s03 = src0->nb[3] / ts_src0;
     const int64_t s3  =  dst->nb[3] / ts_dst;
 
-    const bool use_stream_k = (GGML_CUDA_CC_IS_NVIDIA(cc) && ggml_cuda_highest_compiled_arch(cc) >= GGML_CUDA_CC_VOLTA)
-                            || GGML_CUDA_CC_IS_CDNA(cc);
+    // Stream-k decomposition is the wrong schedule for MoE expert matmuls
+    // (per-expert N is small, fixup overhead dominates). Disable when ids != nullptr.
+    const bool use_stream_k = ((GGML_CUDA_CC_IS_NVIDIA(cc) && ggml_cuda_highest_compiled_arch(cc) >= GGML_CUDA_CC_VOLTA)
+                            || GGML_CUDA_CC_IS_CDNA(cc))
+                            && (ids == nullptr);
 
     // TODO: tighter pool buffer size vs q8 path
     const bool use_native_mxfp4 = blackwell_mma_available(cc) && src0->type == GGML_TYPE_MXFP4;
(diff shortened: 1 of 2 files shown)


nisparks:
I'll leave it up to the reviewers, but I can close my other PR, or mark it as draft and let this one go first.

nisparks:
Tagging @JohannesGaessler, @ORippler, and @ggerganov 

I'm no CUDA expert, but the gains I saw were too good to not share. Trying to break this down into small, incremental changes.

am17an:
512 is not the only batch size that is used, nor is the 3090 card the only one used. Typically performance testing is done on a large array of GPUs which @JohannesGaessler's has patiently accumulated over time. You can leave your findings in a simplified TL;DR format and I can take a look (no AI please, use your own words)

JohannesGaessler:
I will soon make a PR that reduces stream-k overhead more generally. As part of my PR I have integrated something that is equivalent to this PR but with logic that does not explicitly depend on MoE but rather works off of the efficiency loss from tiling (and that threshold in turn depends on the stream-k overhead).

nisparks:
@JohannesGaessler, I was actually exploring that too, thanks to @am17an's comment. As I'm understanding the reason why this is beneficial, it is basically due to the way the expert FFN works, the problem is already broken up into small pieces, and stream-k is further dividing that, causing a lot of overhead with the partial work fixup. Not necessarily a MoE specific thing.

JohannesGaessler:
The biggest issue is just the integer divisions/modulos to calculate indices per dimension from the continuous index. Those are simply slow operations and cannot be efficiently parallelized.

nisparks:
@JohannesGaessler Would you like me to convert my PRs to drafts/WIP?

nisparks:
Superseded by https://github.com/ggml-org/llama.cpp/pull/22298

PR #22259, opened 2026-04-22
ggml-webgpu: add support for im2col

Constannnnnt:
## Overview

In this PR, we implemented the im2col shader, which was needed to slide images into columns for the clip vision encoder. 

## Additional information

image

Note: The performance comparison of Conv2D between WebGPU and CUDA in the last PR was not accurate, as I used my integrated graphics card for testing. In this one, I enabled my RTX card for testing. After implementing conv2d and im2col, I noticed that we currently use a 2D workgroup grid for dispatch. Maybe one optimization we could do is to enable 3D dispatch so that we don't need to do the thread group calculation in conv2d and im2col kernels. 

# Requirements

- AI usage disclosure: Yes, I used ChatGPT to review im2col.


CISC (approved):

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -1003,6 +1003,103 @@ static webgpu_encoded_op ggml_webgpu_conv_2d(webgpu_context & ctx,
...
+    const uint32_t so3 = is_2D ? (uint32_t) (dst->nb[3] / ggml_type_size(dst->type)) :
+                                 (uint32_t) (dst->nb[2] / ggml_type_size(dst->type));
+
+    std::vector<uint32_t> params = {
+        (uint32_t) (ggml_webgpu_tensor_misalignment(ctx, src1) / ggml_type_size(src1->type)),
+        (uint32_t) (ggml_webgpu_tensor_misalignment(ctx, dst) / ggml_type_size(dst->type)),
+
+        si0,
+        si1,
+        si2,
```
reeselevine, line 1041: sorry to always be nitpicky but did you run clang-format on this? Seems weird to have every parameter on a different line.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
```diff
@@ -1003,6 +1003,103 @@ static webgpu_encoded_op ggml_webgpu_conv_2d(webgpu_context & ctx,
...
+        (uint32_t) d1,
+    };
+
+    std::vector<wgpu::BindGroupEntry> entries = {
+        ggml_webgpu_make_tensor_bind_group_entry(ctx, 0, src1),
+        ggml_webgpu_make_tensor_bind_group_entry(ctx, 1, dst),
+    };
+
+    uint32_t max_wg_size =
+        std::min((uint32_t) WEBGPU_MAX_WG_SIZE, ctx->global_ctx->capabilities.limits.maxComputeWorkgroupSizeX);
```
reeselevine: I know I didn't comment on it in conv2d, but is there a reason you're doing the min here? I think I added WEBGPU_MAX_WG_SIZE when I was having some issues with specific shaders much earlier, but I was thinking we could probably start to remove it going forward.

reeselevine (commented):
Looks pretty good. I don't know if the calculation needed for 2d workgroups is really much of a performance issue, that's a very small amount of flops. I think it's more likely general optimizations that the CUDA backend has, and/or due to overhead introduced by WGSL compilers. Any sort of branching or even function calls, especially in loops, can cause significant slowdowns.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: Yeah, I used that every time I saved the file; I think my formatter with clang-format automatically did this. Can you share your configs on this? I did not customize anything with my clang-format.

reeselevine: just tried formatting this branch, apparently this is what it chooses! I try and configure it to use the `.clang-format` file in this repo.

ggml/src/ggml-webgpu/ggml-webgpu.cpp
Constannnnnt: Actually, this was the impression I had around Jan or Feb when I first touched the code to support LFM vision models and had this in mind for Conv2D support. I don't remember the exact problem; maybe it was a 1D dispatch group at the time, and I have some old code for limiting the size. If we are going to remove it, I will clean this.

reeselevine: Yeah it looks like the issue I opened that was the reason for it is closed: https://github.com/gpuweb/gpuweb/issues/5315. So optimistically we can move to just using the limit reported value.

Constannnnnt: Sounds good, I will clean this and also do the same with conv2d together.

reeselevine (approved):

PR #22263, opened 2026-04-22
common: fix macOS cache path segfault when HOME is unset

### LINKED ISSUE #22229: Misc. bug: Installing llama.cpp on macOS through lemonade-idk
### Name and Version

Hello it seems the macOS file size for llama.cpp has changed from 39MB to 8MB after version [b8828](https://github.com/ggml-org/llama.cpp/releases/tag/b8828)

What was the reason for this? I believe it has caused issues in lemonade-sdk.
Please let me know if there where any important changes that we must implement specifically for macOS/metal in order to install the latest llama.cpp and any required runtime binaries that may have been removed, (only guessing) or perhaps by accident. Thanks!

This stops us from upgrading to the "LATEST" but we can upgrade to 8828.

https://github.com/lemonade-sdk/lemonade/issues/1669

### Operating systems

Mac

llama-server

### Command line

```shell

```

install latest lemonade and manually change backend, stop the service and restart it after changing backend version code to the newest. File path is accessible here with instructions.

https://github.com/lemonade-sdk/lemonade/pull/1675

### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>

```console

```
</details>

Geramy:
On macOS, fs_get_cache_directory crashes with SIGSEGV when
HOME is not set (for example, when llama.cpp is launched by
a LaunchDaemon). std::getenv returns NULL and the following
std::string concatenation dereferences it.

This adds the same guard the Linux branch already has: check
getenv("HOME") first, and fall back to getpwuid(getuid()) if
it is unset. The pwd.h include guard is extended to __APPLE__
so getpwuid is available.

Closes #22229

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -49,7 +49,7 @@
 #include <unistd.h>
 #endif
 
-#if defined(__linux__)
+#if defined(__linux__) || defined(__APPLE__)
 #include <sys/types.h>
 #include <pwd.h>
 #endif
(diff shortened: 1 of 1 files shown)


angt:
I’m not sure we should handle the missing `HOME` the same way as Linux and POSIX friends. With `LaunchDaemon`,  it seems the correct place should be simply `/Library/Caches/llama.cpp`.

Geramy:
Yeah I wasn't too sure myself, Apple does everything differently so that makes sense I'll go see if I can find any Apple documentation for a concrete statement if they have any and make the change, thanks!

angt:
> Yeah I wasn't too sure myself, Apple does everything differently so that makes sense I'll go see if I can find any Apple documentation for a concrete statement if they have any and make the change, thanks!

I think the best way would be for lemonade to set up `LLAMA_CACHE` and `HF_HUB_CACHE` (or `HF_HOME/hub`) and not rely on this code.

Geramy:
@angt I believe this fix should be at the llama.cpp side of things that is the proper fix. otherwise you are expecting users of llama.cpp to know that this is a edge case they need to handle. The fix I propose works. We don't really use llama.cpp for its HF capabilities actually. we use HF_HOME ourselves. I will circle back again with an updated default path.

PR #22199, opened 2026-04-21
ggml-webgpu: enable FLASH_ATTN_EXT on browser without subgroup matrix 

ArberSephirotheca:
## Overview

This PR addresses few things:
1. Cleanup the vec path to remove requirement for subgroup matrix.
2. Add a subgroup-based tile flash-attention kernel that works without subgroup-matrix operations.
3. Remove subgroup-matrix-specific parameters and naming from the vec path.
4. Enable FLASH_ATTN_EXT on browsers for the non-subgroup-matrix paths (vec and tile).
## Additional information

Performance when running on my M4 Mac Mini Pro:
```bash
ggml_metal_device_init: tensor API disabled for pre-M5 and pre-A19 devices
ggml_metal_library_init: using embedded metal library
ggml_metal_library_init: loaded in 0.019 sec
ggml_metal_rsets_init: creating a residency set collection (keep_alive = 180 s)
ggml_metal_device_init: GPU name:   MTL0
ggml_metal_device_init: GPU family: MTLGPUFamilyApple9  (1009)
ggml_metal_device_init: GPU family: MTLGPUFamilyCommon3 (3003)
ggml_metal_device_init: GPU family: MTLGPUFamilyMetal4  (5002)
ggml_metal_device_init: simdgroup reduction   = true
ggml_metal_device_init: simdgroup matrix mul. = true
ggml_metal_device_init: has unified memory    = true
ggml_metal_device_init: has bfloat            = true
ggml_metal_device_init: has tensor            = false
ggml_metal_device_init: use residency sets    = true
ggml_metal_device_init: use shared buffers    = true
ggml_metal_device_init: recommendedMaxWorkingSetSize  = 40200.90 MB
ggml_webgpu: adapter_info: vendor_id: 4203 | vendor: apple | architecture: metal-3 | device_id: 0 | name: Apple M4 Pro | device_desc: Metal driver on macOS Version 26.2 (Build 25C56)
Testing 4 devices

Backend 1/4: MTL0
  Skipping
Backend 2/4: WebGPU
  Device description: WebGPU
  Device memory: 28753 MB (28753 MB free)

  FLASH_ATTN_EXT(hsk=72,hsv=72,nh=16,nr23=[1,1],kv=5776,nb=5776,mask=0,sinks=0,max_bias=0.000000,logit_softcap=0.000000,prec=f32,type_KV=f16,permute=[0,1,2,3]): not supported
[...]

PR #22266, opened 2026-04-22
ggml-wegpu: handle the buffer aliasing for rms fuse

Constannnnnt:
## Overview

This PR addressed an edge case of #21983. I load and run a model in the browser, and I met this error: 
```
ggml_webgpu: Device error! Reason: 2, Message: Writable storage buffer binding aliasing found between [BindGroup "RMS_NORM_MUL"] set at bind group index 0, binding index 0, and [BindGroup "RMS_NORM_MUL"] set at bind group index 0, binding index 2, with overlapping ranges (offset: 5242880, size: 4096) and (offset: 5242880, size: 4096) in [Buffer "tensor_buf3"].
While encoding [ComputePassEncoder (unlabeled)].DispatchWorkgroups(1, 1, 1).
While finishing [CommandEncoder (unlabeled)].
```

As the error showed, it was associated with the buffer overlapping. I used a coding agent to analyze the logs:
- the previous `inplace` flag was only checking if `mul_src` and `dst` overlapped
- it missed a scenario where the `rn_src` overlapped with `dst`, i.e. (`rn_src==dst`), leading to this `tensor_buf3` issue. 

I reused the convention from the 'binary' shader:
```
inplace means src0 == dst. => rn_src == dst
overlap means src1 == dst. => mul_src  == dst
src_overlap means src0 == src1.
```

## Additional information

I didn't run any benchmark tests and only tested the model behaviour in the browser.

# Requirements

- I have read and agree with the [[contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: Yes, I used coding agents for debugging.


reeselevine:
Thanks for the fix. I tested it out and it looks like even with `skip_validation` turned off, there is nothing in test-backend-ops that would have caught this right now. Potentially other tests would, we actually should turn off `skip_validation`, I originally added it for performance natively but I think the impact is minimal. FlashAttention needs to handle some aliasing first though, which would be a good target for https://github.com/ggml-org/llama.cpp/pull/22199.

@yomaytk fyi, if you're working on other fusion chains it might be good to check to see what other kind of buffer aliasing can occur and add a test for it in test-backend-ops if possible.

yomaytk:
Thanks for the fix, this looks good to me. How about adding a test case in test-backend-ops and confirming that it passes without `skip_validation` in this PR? If not, I'm happy to follow up with a separate PR.

> @yomaytk fyi, if you're working on other fusion chains it might be good to check to see what other kind of buffer aliasing can occur and add a test for it in test-backend-ops if possible.

Got it, thanks.

reeselevine (approved):

CISC (approved):

reeselevine (approved):

Issue #22142, opened 2026-04-20
Misc. bug: test-chat-template hangs on Windows (MSYS2 UCRT64 + GCC)


rillomas:
### Name and Version

$ bin/llama-cli.exe --version
load_backend: failed to find ggml_backend_init in C:\Users\devcloud\Documents\repo\llama.cpp\build-ci-debug\bin\ggml-vulkan.dll
load_backend: failed to find ggml_backend_init in C:\Users\devcloud\Documents\repo\llama.cpp\build-ci-debug\bin\ggml-cpu.dll
version: 8829 (6990e2f1f)
built with GNU 15.2.0 for Windows AMD64

### Operating systems

Windows

### Which llama.cpp modules do you know to be affected?

Core DLLs?

### Command line

```shell
bin/test-chat-template.exe`
```

### Problem description & steps to reproduce

I'm observing a complete hang at test-chat-template.exe from https://github.com/ggml-org/llama.cpp/commit/6990e2f1f7581d332a6a1f34d6c567be70138184 when building for CI as ` LLAMA_FATAL_WARNINGS=OFF GG_BUILD_NINJA=1 GG_BUILD_VULKAN=1 GG_BUILD_LOW_PERF=1 ./ci/run.sh ./results/llama.cpp ./mnt/llama.cpp` on MSYS2 UCRT64 + GCC.  The test execution seems to finish but just hangs at the end and I can't Ctrl-C. Running from gdb is the same (Can't Ctrl-C). https://github.com/ggml-org/llama.cpp/commit/fcc7508759c7a3fe5a0f4500592657900be8aca5 works correctly

I see the same issue when built with `cmake -B build_vk -DGGML_VULKAN=ON -DLLAMA_CURL=OFF -DBUILD_SHARED_LIBS=ON` but it does **not** occur when built with `cmake -B build_vk -DGGML_VULKAN=ON -DLLAMA_CURL=OFF`

### First Bad Commit

https://github.com/ggml-org/llama.cpp/commit/6990e2f1f7581d332a6a1f34d6c567be70138184 (#21936)

### Relevant log output

<details>
<summary>Logs</summary>

```console
# build-ci-debug/bin/test-chat-template.exe
Built-in chat templates:
  bailing
  bailing-think
  bailing2
  chatglm3
  chatglm4
  chatml
  command-r
  deepseek
  deepseek-ocr
  deepseek2
  deepseek3
  exaone-moe
  exaone3
  exaone4
  falcon3
[...]

PR #22273, opened 2026-04-23
common: Intentionally leak logger instance to fix hanging on Windows 

rillomas:
## Overview

Added workaround for #22142. There are three points in this PR:

- Intentional leak of logger instance
  - `~common_log()` called at DLL teardown phase was causing hanging on Windows. DLL teardown phase seems to be a fragile timing to do system calls like mutex lock, cond notify, thread join, etc. which did not provide sane results. We are working around this by intentionally leaking the logger instance to skip cleanup.
- Using static array's instead of std::vector in logger
  - We were seeing crashes on Linux when logger thread outlived the std::vector teardown, causing the thread to touch released items on std::vector. Using static arrays prevents this.
- Calling `common_log_free` at `test-log` 
  -  Even if we skipped the logger destructor, we were still seeing hangs on Windows after adding `common_log_flush`. This seemed like a timing issue where the OS is running DLL teardown while the runtime tries to start the logger thread asynchronously (which was started by `common_log_flush`). We added an explicit cleanup in the end of the unit test to avoid this.

# Requirements

- AI usage disclosure: YES (Copilot GPT5.4 medium). Used for issue analysis and solution recommendation.


pwilkin (approved):

ggerganov:
Does not seem to work - the tests are failing

rillomas:
Let me have a look again.

rillomas:
I've fixed the crashing on Linux. It seems like `g_col` being a vector was causing the heap to get freed at teardown phase while the logger thread is running, causing the crash. Previously the common_log destructor was (presumably) joining the logger thread first assuring `g_col` was never touched after release.

rillomas:
Seems like even if we skip the logger's destruction, calling `common_log_flush` just before the program exits will cause hanging in the same way. It looks like a timing issue where the OS tries to clean the threads and the logger tries to run the logger thread because the hanging goes away when I add `printf` which changes timing.

~~I guess I'll just remove the intentional leaking and fix the unit test to call `common_log_free` since we'll need it anyway, and it's safer to call it from the app rather than during DLL teardown.~~ Forgot that the hanging was happening on other apps so I'll need to keep the intentional leaking. I'll also need the `common_log_free` in the unit test.

rillomas:
@ggerganov How does the current PR look? Please let me know if there are other fixes/tests I should perform.

pwilkin (approved):

common/log.cpp
```diff
@@ -368,14 +365,20 @@ struct common_log * common_log_init() {
 }
 
 struct common_log * common_log_main() {
-    static struct common_log log;
+    // We intentionally leak (i.e. do not delete) the logger singleton because
+    // common_log destructor called at DLL teardown phase will cause hanging on Windows.
+    // OS will release resources anyway so it should not be a significant issue,
+    // though this design may cause logs to be lost if not flushed before the program exits.
+    // Refer to https://github.com/ggml-org/llama.cpp/issues/22142 for details.
+    static struct common_log* log;
```
ggerganov:
```suggestion
    static struct common_log * log;
```

PR #22274, opened 2026-04-23
model-conversion : fix mmproj output file name [no ci]

danbev:
## Overview
This commit updates the convert-model.sh script to properly handle mmproj output files.

## Additional information

The motivation for this that currently the same name as the original model is used as the mmproj file, which causes the original model to be overwritten and no mmproj-<model_name>.gguf to be created.

# Requirements

- AI usage disclosure: No

### DIFF
--- a/examples/model-conversion/scripts/causal/convert-model.sh
+++ b/examples/model-conversion/scripts/causal/convert-model.sh
@@ -26,7 +26,7 @@ OUTPUT_DIR="${OUTPUT_DIR:-../../models}"
 TYPE="${OUTTYPE:-f16}"
 METADATA_OVERRIDE="${METADATA_OVERRIDE:-}"
 if [[ -n "$MMPROJ" ]]; then
-    CONVERTED_MODEL="${OUTPUT_DIR}/mmproj-$(basename "${CONVERTED_MODEL}")"
+    CONVERTED_MODEL="${OUTPUT_DIR}/mmproj-${MODEL_NAME}.gguf"
 else
     CONVERTED_MODEL="${OUTPUT_DIR}/${MODEL_NAME}.gguf"
 fi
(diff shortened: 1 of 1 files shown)


CISC (approved):

examples/model-conversion/scripts/causal/convert-model.sh
```diff
@@ -25,7 +25,11 @@ MODEL_NAME="${MODEL_NAME:-$(basename "$MODEL_PATH")}"
 OUTPUT_DIR="${OUTPUT_DIR:-../../models}"
 TYPE="${OUTTYPE:-f16}"
 METADATA_OVERRIDE="${METADATA_OVERRIDE:-}"
-CONVERTED_MODEL="${OUTPUT_DIR}/${MODEL_NAME}.gguf"
+if [[ -n "$MMPROJ" ]]; then
+    CONVERTED_MODEL="${OUTPUT_DIR}/mmproj-$(basename "${CONVERTED_MODEL}")"
```
ehoogeveen-medweb: Is `${CONVERTED_MODEL}` already set at this point?

ehoogeveen-medweb: If not you probably meant
```bash
CONVERTED_MODEL="${OUTPUT_DIR}/${MODEL_NAME}.gguf"
if [[ -n "$MMPROJ" ]]; then
    CONVERTED_MODEL="${OUTPUT_DIR}/mmproj-$(basename "${CONVERTED_MODEL}")"
fi
```

CISC:
```suggestion
    CONVERTED_MODEL="${OUTPUT_DIR}/mmproj-${MODEL_NAME}.gguf"
```

CISC: Good catch. :)

danbev: Yes, it is expected to have been set to the converted model path, or quantized model path. But I'll update this as suggested by @CISC below to avoid any confusion. Thanks.

danbev: Thanks, I've updated this now.

CISC (approved):

ggerganov (approved):

PR #22288, opened 2026-04-23
server : fix swa-full logic

ggerganov:
## Overview

fix #21468
alt #21749

Simplify the logic by augmenting `llama_model_n_swa` with a `server_context.n_swa` member. When `--swa-full` is passed, we set `n_swa = 0` to simulate a non-SWA model

# Requirements

- AI usage disclosure: NO

### DIFF
--- a/tools/server/server-context.cpp
+++ b/tools/server/server-context.cpp
@@ -675,6 +675,10 @@ private:
 
     int32_t n_ctx; // total context for all clients / slots
 
+    // set to llama_model_n_swa(model)
+    // if swa_full is enabled, this is set to 0 to simulate a non-SWA model
+    int32_t n_swa;
+
     // slots / clients
     std::vector<server_slot> slots;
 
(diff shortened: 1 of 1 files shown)


shipped-it:
I'm a bit unsure about the approach. Why would the model be reported as non-SWA? It is still SWA, just with a full-size cache.

Also, you've made the suggestion to use `--swa-full --no-mmproj`. Correct me if I'm wrong, but in practice I've seen repetitions issues when using Gemma 4.

I will confirm shortly if this fixes the issue, or if I get the repetition problem.

ggerganov:
> Also, you've made the suggestion to use --swa-full --no-mmproj.

Cache reuse does not work with mmproj.

> Correct me if I'm wrong, but in practice I've seen repetitions issues when using Gemma 4.

I am not following. This fixes the cache reuse logic - I am not aware of any repetitions.

shipped-it:
I confirm that it is fixed with this PR or #21749 (tested on ROCm)

Without PR:
warm req: prompt_n=821, prompt_ms=982 

With PR:
warm req: prompt_n=5, prompt_ms=71 so about 13x faster

Issue #21708, opened 2026-04-10
Misc. bug: --version and --help can be very slow


diimdeep:
### Name and Version

llama-cli --version
version: 8680 (15f786e65)
built with AppleClang 17.0.0.17000604 for Darwin arm64

### Problem description & steps to reproduce

Observed causes of `./llama-cli` with either `--version` or `--help` being slow:

- ggml_metal_library_init: loaded in 11.251 sec before returning 
- doing requests to huggingface for whatever reason before returning

```
ggml_metal_library_init: loaded in 11.251 sec
ggml_metal_rsets_init: creating a residency set collection (keep_alive = 180 s)
ggml_metal_device_init: GPU name:   MTL0
ggml_metal_device_init: GPU family: MTLGPUFamilyApple8  (1008)
ggml_metal_device_init: GPU family: MTLGPUFamilyCommon3 (3003)
ggml_metal_device_init: GPU family: MTLGPUFamilyMetal3  (5001)
ggml_metal_device_init: simdgroup reduction   = true
ggml_metal_device_init: simdgroup matrix mul. = true
ggml_metal_device_init: has unified memory    = true
ggml_metal_device_init: has bfloat            = true
ggml_metal_device_init: has tensor            = false
ggml_metal_device_init: use residency sets    = true
ggml_metal_device_init: use shared buffers    = true
ggml_metal_device_init: recommendedMaxWorkingSetSize  = 11453.25 MB
load_backend: loaded MTL backend from /opt/homebrew/Cellar/ggml/0.9.11/libexec/libggml-metal.so
load_backend: loaded CPU backend from /opt/homebrew/Cellar/ggml/0.9.11/libexec/libggml-cpu-apple_m2_m3.so
version: 8680 (15f786e65)
built with AppleClang 17.0.0.17000604 for Darwin arm64
```

woof-dog:
So llama.cpp is making requests to HuggingFace on its own? Isn't this a significant privacy issue? (I didn't build with cURL support... perchance I am spared)

diimdeep:
I spotted only one instance of it doing unsolicited request to huggingface, on macos when it tried to move local cache of models from ~/Library/.. to ~/.cache/huggingface/
I guess that this is consequence of hugging face acquiring this org.
But watching it doing so before doing --help print was not fun.

PR #22290, opened 2026-04-23
common : only load backends when required

angt:
## Overview

Only load backends when required.

## Additional information

This fixe the following issues: #20186, #21708 and maybe others.

# Requirements

- AI usage disclosure: YES (kimi): I've asked the list of functions to modify for a minimal patch

### DIFF
--- a/common/arg.cpp
+++ b/common/arg.cpp
@@ -248,6 +248,8 @@ std::vector<std::string> common_arg::get_env() const {
 
 // Helper function to parse tensor buffer override strings
 static void parse_tensor_buffer_overrides(const std::string & value, std::vector<llama_model_tensor_buft_override> & overrides) {
+    ggml_backend_load_all();
+
     std::map<std::string, ggml_backend_buffer_type_t> buft_list;
     for (size_t i = 0; i < ggml_backend_dev_count(); ++i) {
         auto * dev = ggml_backend_dev_get(i);
@@ -803,6 +805,7 @@ static std::vector<ggml_backend_dev_t> parse_device_list(const std::string & val
     if (dev_names.size() == 1 && dev_names[0] == "none") {
         devices.push_back(nullptr);
     } else {
+        ggml_backend_load_all();
         for (const auto & device : dev_names) {
             auto * dev = ggml_backend_dev_by_name(device.c_str());
             if (!dev || ggml_backend_dev_type(dev) == GGML_BACKEND_DEVICE_TYPE_CPU) {
@@ -820,6 +823,7 @@ static void add_rpc_devices(const std::string & servers) {
     if (rpc_servers.empty()) {
         throw std::invalid_argument("no RPC servers specified");
     }
+    ggml_backend_load_all();
     ggml_backend_reg_t rpc_reg = ggml_backend_reg_by_name("RPC");
     if (!rpc_reg) {
         throw std::invalid_argument("failed to find RPC backend");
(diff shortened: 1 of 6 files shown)


ggml/src/ggml-backend-reg.cpp
```diff
@@ -541,6 +541,11 @@ static ggml_backend_reg_t ggml_backend_load_best(const char * name, bool silent,
 }
 
 void ggml_backend_load_all() {
+    static bool loaded = false;
```
danbev: Instead of modifying ggml could we perhaps introduce a helper function in arg.cpp that does this?

common/arg.cpp
```diff
@@ -1002,9 +1005,6 @@ common_params_context common_params_parser_init(common_params & params, llama_ex
 
     params.use_color = tty_can_use_colors();
 
-    // load dynamic backends
-    ggml_backend_load_all();
```
danbev, line 1020: I think we need this in case none of the other options are used?

ggml/src/ggml-backend-reg.cpp
angt: I opted for the minimal change but honestly this function should be able to be called many times without the static variable (I was a bit lazy) and I'm not sure adding a helper to hide that would be any better. I'm going to make a quick pre-PR.

common/arg.cpp
angt: I think the current logic covers all scenarios do you have a specific edge case in mind ?

ggerganov:
The `test-thread-safety` is failing

ggerganov:
Hm, I don't see where we would call `ggml_backend_load_all()` for `llama-cli`, `llama-server`, etc. Building with `-DGGML_BACKEND_DL=ON` makes these binaries not load any backends.

angt:
> Hm, I don't see where we would call `ggml_backend_load_all()` for `llama-cli`, `llama-server`, etc. Building with `-DGGML_BACKEND_DL=ON` makes these binaries not load any backends.

Damn, checking. I don’t use AI often (for llama.cpp!). Kimi is this bad ?  🤣

angt:
The last call was missing...
Edit: So Kimi is fine, it was my fault 🙄

angt:
@ggerganov maybe I can take this PR as an opportunity to remove `ggml_backend_load_all` from `common_params_parser_init` and call it explicitly where it’s needed, instead of hiding it there?

ggerganov:
> @ggerganov maybe I can take this PR as an opportunity to remove `ggml_backend_load_all` from `common_params_parser_init` and call it explicitly where it’s needed, instead of hiding it there?

Yes, let's do that.

angt:
I finally have updated `llama_backend_init()` to call `ggml_backend_load_all()`. The name was a perfect fit, so I decided to try this, but I’m not yet 100% convinced too, and maybe we should just call `ggml_backend_load_all()` everywhere...

ggerganov (approved):

pwilkin (approved):

danbev (approved):

rgerganov:
This patch broke RPC (issue #22721) and introduced obscure regressions (issue #22748). Having multiple calls to `ggml_backend_load_all()` is not how loading dynamic backends is supposed to work and breaks many implicit assumptions.

I suggest to revert this patch.

angt:
I think it’s just because of some missing calls it’s hard to track all of them

angt:
#22752 should fix both issues (and others)

PR #22296, opened 2026-04-23
ggml : skip already registered backends and devices

angt:
## Overview

Skip already registered backends and devices

## Additional information

# Requirements

- AI usage disclosure: YES


angt:
pre-PR for #22290

PR #22297, opened 2026-04-23
CUDA: add POOL_1D

LeoYangXY:
Implement 1D pooling (avg and max) for the CUDA backend.

Mentioned in #14909

## Overview

Add CUDA support for `GGML_OP_POOL_1D` with both average and max pooling modes. The implementation follows the same pattern as `pool2d.cu`, adapted for the 1D case. Supports configurable kernel size, stride, and padding.

Tested on RTX 5050 Laptop GPU (SM 12.0, CUDA 13.2):

    ./build/bin/test-backend-ops -b CUDA0 test -o POOL_1D
    48/48 tests passed

## Requirements

- AI usage disclosure: YES - AI (Cursor) was used to assist with implementation and testing. All code was reviewed and verified by the author.


CISC:
Nice, [update docs too](https://github.com/ggml-org/llama.cpp/blob/master/docs/ops.md#how-to-add-a-backend-to-this-table), it has been a while, so will probably update a few other ops.

ggml/src/ggml-cuda/pool1d.cu
```diff
@@ -0,0 +1,77 @@
...
+        default: return;
+    }
+
+    int count = 0;
+    for (int j = bw; j < ew; ++j) {
+#if __CUDA_ARCH__ >= 350
+        const float cur = __ldg(i_ptr + j);
+#else
+        const float cur = i_ptr[j];
+#endif
```
am17an: this shouldn't be required, just use `const float cur = i_ptr[j];`

am17an (approved):

ggml/src/ggml-cuda/pool1d.cu
LeoYangXY: thanks，i will correct it right now

CISC:
@LeoYangXY Did you update docs as described in the link? It doesn't look complete.

LeoYangXY:
@CISC I updated the docs for POOL_1D in the latest commit (632eaea). Could you clarify which other ops you'd like me to update? I wasn't sure which ones were missing — happy to update them if you can point me to the specific gaps.

CISC:
> @CISC I updated the docs for POOL_1D in the latest commit ([632eaea](https://github.com/ggml-org/llama.cpp/commit/632eaeaac0d4fb6b022be0510e1b85fb713baa34)). Could you clarify which other ops you'd like me to update? I wasn't sure which ones were missing — happy to update them if you can point me to the specific gaps.

It looks like you perhaps updated it manually instead of using the script as described.

LeoYangXY:
> > @CISC I updated the docs for POOL_1D in the latest commit ([632eaea](https://github.com/ggml-org/llama.cpp/commit/632eaeaac0d4fb6b022be0510e1b85fb713baa34)). Could you clarify which other ops you'd like me to update? I wasn't sure which ones were missing — happy to update them if you can point me to the specific gaps.
> 
> It looks like you perhaps updated it manually instead of using the script as described.

Ah sorry, I did it manually. I'll rerun the script to regenerate the docs and push an update shortly.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -5124,6 +5128,8 @@ static bool ggml_backend_cuda_device_supports_op(ggml_backend_dev_t dev, const g
         case GGML_OP_CONV_2D:
         case GGML_OP_CONV_2D_DW:
         case GGML_OP_CONV_TRANSPOSE_2D:
+        case GGML_OP_POOL_1D:
```
JohannesGaessler, line 5132:
```suggestion
        case GGML_OP_CONV_TRANSPOSE_2D:
            return true
        case GGML_OP_POOL_1D:
```

This is probably why there are changes to the supported ops beyond `POOL_1D`.

JohannesGaessler (changes requested):

ggml/src/ggml-cuda/ggml-cuda.cu
LeoYangXY: Thanks, you were right. I accidentally let CONV_TRANSPOSE_2D fall through into the POOL_1D support check. I restored the separate return true; for CONV_TRANSPOSE_2D, reran the CUDA support CSV generation and regenerated docs/ops.md. The only remaining change is the intended CUDA support check for POOL_1D.

JohannesGaessler: I am not seeing any changes to the CSV and markdown files since my last review, did you perhaps forget to check in the regenerated ones?

PR #22299, opened 2026-04-23
internal AllReduce kernel for CUDA provider

scutler-nv:
## Overview

Implements an internal reduction kernel for tensor parallelism mode. Improved performance over the butterfly fallback, used on Windows until NCCL is supported.

## Additional information

On Windows, enabled by default and the preferred provider. On Linux, enabled only by envvar (NCCL is default). Force with GGML_CUDA_ALLREDUCE envvar. Set to "internal", "nccl", or "none" (butterfly fallback).

Generally speaking, shows substantial token generation gains over layer mode on Windows (prefill is mostly flat except on llama 3b, hardware is 2 x RTX 5090):
```
pp512:

  ┌─────────────────────────┬──────────┬───────────┬──────────┬──────────────┬──────────────────┐
  │          Model          │  layer   │ butterfly │ internal │ int vs layer │ int vs butterfly │
  ├─────────────────────────┼──────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ llama 70B Q4_K - Medium │  1737.05 │   1109.06 │  1830.95 │        +5.4% │           +65.1% │
  ├─────────────────────────┼──────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ llama 3B Q8_0           │ 23081.36 │   8235.32 │ 13594.42 │       -41.1% │           +65.1% │
  ├─────────────────────────┼──────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ qwen35moe 35B.A3B Q8_0  │  7776.53 │   5315.45 │  7686.82 │        -1.2% │           +44.6% │
  ├─────────────────────────┼──────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ qwen3 32B Q8_0          │  3785.97 │   2105.48 │  3579.49 │        -5.5% │           +70.0% │
  ├─────────────────────────┼──────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ gemma4 31B Q8_0         │  3977.92 │   2206.87 │  3794.05 │        -4.6% │           +71.9% │
  └─────────────────────────┴──────────┴───────────┴──────────┴──────────────┴──────────────────┘

  tg128:

  ┌─────────────────────────┬────────┬───────────┬──────────┬──────────────┬──────────────────┐
  │          Model          │ layer  │ butterfly │ internal │ int vs layer │ int vs butterfly │
  ├─────────────────────────┼────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ llama 70B Q4_K - Medium │  34.85 │     36.52 │    55.94 │       +60.5% │           +53.2% │
  ├─────────────────────────┼────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ llama 3B Q8_0           │ 299.32 │    136.13 │   314.96 │        +5.2% │          +131.4% │
  ├─────────────────────────┼────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ qwen35moe 35B.A3B Q8_0  │ 191.69 │     87.61 │   175.75 │        -8.3% │          +100.6% │
  ├─────────────────────────┼────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ qwen3 32B Q8_0          │  43.38 │     45.74 │    68.59 │       +58.1% │           +50.0% │
  ├─────────────────────────┼────────┼───────────┼──────────┼──────────────┼──────────────────┤
  │ gemma4 31B Q8_0         │  42.98 │     45.10 │    65.03 │       +51.3% │           +44.2% │
  └─────────────────────────┴────────┴───────────┴──────────┴──────────────┴──────────────────┘
```

The code supports two paths: single kernel and copy engine. Single kernel mode performs the device->host and host->device copies within a kernel, along with optional bf16 format conversion, and a host token spin loop to synchronize. Copy engine mode uses chunked D2H and H2D copies synchronized with CUDA events, as well as optional format conversion before and after.

## Requirements

CUDA only, only relevant in multi-GPU tensor parallelism mode. All SM versions supported.

Claude Code and Codex were used during development.


ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,624 @@
...
+#else
+#define GGML_CUDA_AR_WDOG_EXTRA_ARGS
+#endif
+
+#define LAUNCH_AR_KERNEL(T) \
+            ggml_cuda_ar_kernel<T><<<dim3(1), dim3(256), 0, p->streams[i]>>>( \
+                reinterpret_cast<const T *>(data), \
+                reinterpret_cast<T *>(data), \
+                reinterpret_cast<T *>(p->host_buf[i]), \
+                reinterpret_cast<const T *>(p->host_buf[peer]), \
```
gaugarg-nv: If I understand this correctly, you are reusing the same `host_buf` for all chunks. How do you ensure GPU-0 doesn't start writing chunk N + 1 to this host memory, while GPU 1 is reading chunk N from the same host memory?

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1138,79 +1139,128 @@ static const ggml_backend_buffer_type_i ggml_backend_cuda_split_buffer_type_inte
...
     }
-    ggml_backend_cuda_comm_context * ret = new ggml_backend_cuda_comm_context;
-    std::vector<int> dev_ids;
-    ret->backends.reserve(n_backends);
-    dev_ids.reserve(n_backends);
+
+    const char * env = getenv("GGML_CUDA_ALLREDUCE");
+    const bool force_none     = env && strcmp(env, "none")     == 0;
+    const bool force_internal = env && strcmp(env, "internal") == 0;
+    const bool force_nccl     = env && strcmp(env, "nccl")     == 0;
```
gaugarg-nv: I'd prefer that we don't introduce new env variables, or at least if no env variable is set, it automatically picks the best path.

The preference order could be something like this-
1. Always use NCCL if llama.cpp is compiled with NCCL
2. Use the new path if NCCL is not available, and for cases where the new path is supported (> sm80, two GPUs?) and outperforms the existing fallback
3. Use the older fallback path.

gaugarg-nv (commented):
@scutler-nv Have you run `llama-perplexity` with this change and does it generate consistent and matching output with previous fallback or NCCL path? 

It will be a good idea to run it multiple times to ensure there is no  run-to-run variation.

ggml/src/ggml-cuda/allreduce.cu
scutler-nv: Good point. Unlikely to be a problem in practice, but in principle the work done on one GPU in the last pass could overlap with the work done on the other GPU in the next pass. Double-buffering this will fix the problem since there will always be a sync point before any memory is reused.

ggml/src/ggml-cuda/ggml-cuda.cu
scutler-nv: Yes, that's how it works (when the regkey is not set), except that internal is preferred over NCCL. We'll fall back to NCCL if available, otherwise the butterfly path. The logic is confusing, though; I'll try to improve it (or the comments at least). If the envvar is set, then it's all or nothing (no fallbacks, which get too complicated).

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,624 @@
...
+// intra-block coordination and we can use the cheaper non-cooperative launch.
+// 256 threads gives good occupancy while keeping register pressure low.
+//
+// When GGML_CUDA_AR_WATCHDOG is enabled, Phase 2 has a spin limit
+// (max_spin).  If the limit is reached the kernel writes a debug record to
+// a per-GPU ring buffer in pinned host memory, then bails out — all threads
+// exit the kernel immediately (Phase 3 is skipped).
+// ---------------------------------------------------------------------------
+
+#if GGML_CUDA_AR_WATCHDOG
```
gaugarg-nv: Please remove all the debugging code under `GGML_CUDA_AR_WATCHDOG`.

scutler-nv:
> @scutler-nv Have you run `llama-perplexity` with this change and does it generate consistent and matching output with previous fallback or NCCL path?
> 
> It will be a good idea to run it multiple times to ensure there is no run-to-run variation.

Running with this:
`llama-perplexity.exe -m Llama-3.3-70B-Instruct-Q4_K_M.gguf -f wiki.test.raw -fa 1 -c 512 -sm <tensor | layer> --chunks 32  `
I get this:
```
  ┌──────────┬────────┬─────────┬─────────────────────────────────────┐
  │ Provider │  PPL   │    ±    │                Notes                │
  ├──────────┼────────┼─────────┼─────────────────────────────────────┤
  │ layer    │ 3.4044 │ 0.07850 │ -sm layer, no AR (gold reference)   │
  ├──────────┼────────┼─────────┼─────────────────────────────────────┤
  │ internal │ 3.4044 │ 0.07850 │ Our AllReduce                       │
  ├──────────┼────────┼─────────┼─────────────────────────────────────┤
  │ nccl     │ 3.4043 │ 0.07856 │ NCCL                                │
  ├──────────┼────────┼─────────┼─────────────────────────────────────┤
  │ none     │ 3.4028 │ 0.07852 │ Meta-backend butterfly (-sm tensor) │
  └──────────┴────────┴─────────┴─────────────────────────────────────┘
```

All seem very close.

ggml/src/ggml-cuda/allreduce.cu
scutler-nv: Done.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1138,79 +1139,128 @@ static const ggml_backend_buffer_type_i ggml_backend_cuda_split_buffer_type_inte
...
+    ret->nccl_eligible = !force_internal;
 #else
-    // If NCCL is installed it is used by default for optimal performance.
-    // However, NVIDIA does not distribute NCCL with CUDA so users may be unwittingly missing this package.
-    // RCCL is disabled by default, users are explicitly opting in.
-    // Therefore print no warning for RCCL.
-#if !defined(GGML_USE_HIP) && !defined(GGML_USE_MUSA)
-    static bool warning_printed = false;
-    if (!warning_printed) {
-        GGML_LOG_WARN("%s: NVIDIA Collective Communications Library (NCCL) is unavailable, multi GPU performance will be suboptimal\n", __func__);
```
gaugarg-nv, line 1194: I think we should keep this warning at least on Linux. We should recommend using NCCL at least on platforms where it is supported.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,811 @@
...
+    CUDA_CHECK(cudaStreamWaitEvent(cuda_ctx->stream(), ev.ker));
+}
+
+// ---------------------------------------------------------------------------
+// Init / free
+// ---------------------------------------------------------------------------
+
+ggml_cuda_ar_pipeline * ggml_cuda_ar_pipeline_init(const int * devices, size_t n_devices) {
+
+    if (n_devices != 2) {
```
gaugarg-nv, line 395: We should also add any other constraints like sm80+ in this check?
Also add `GGML_LOG_DEBUG` pointing that "internal" didn't engage.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1138,79 +1139,128 @@ static const ggml_backend_buffer_type_i ggml_backend_cuda_split_buffer_type_inte
...
+    ret->dev_ids.reserve(n_backends);
     for (size_t i = 0; i < n_backends; i++) {
-        ret->backends.push_back(backends[i]);
-        ggml_backend_cuda_context * cuda_ctx = (ggml_backend_cuda_context *) backends[i]->context;
-        dev_ids.push_back(cuda_ctx->device);
+        ret->dev_ids.push_back(static_cast<ggml_backend_cuda_context *>(backends[i]->context)->device);
+    }
+
+    // Try to allocate the internal pipeline unless the user forced NCCL.
+    if (!force_nccl) {
```
gaugarg-nv: I don't agree with this. We should use NCCL if llama.cpp is compiled with NCCL.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1138,79 +1139,128 @@ static const ggml_backend_buffer_type_i ggml_backend_cuda_split_buffer_type_inte
...
+        ret->ar_pipeline = ggml_cuda_ar_pipeline_init(ret->dev_ids.data(), n_backends);
+        if (ret->ar_pipeline == nullptr) {
+            // Clear any sticky CUDA error from the failed init so it can't
+            // leak into a later NCCL call.
+            (void) cudaGetLastError();
+            if (force_internal) {
+                GGML_LOG_ERROR("%s: internal AllReduce pipeline init failed; falling back to butterfly\n", __func__);
```
gaugarg-nv: This error message is wrong. It seems we will fallback to NCCL if NCCL is present.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,811 @@
...
+    for (int i = 0; i < n; ++i) {
+        compute_flag[i] = (tensors[i]->flags & GGML_TENSOR_FLAG_COMPUTE) != 0;
+    }
+
+    // Decide between copy-engine and chunked-kernel paths based on the working
+    // type's actual byte count.
+    const bool use_copy_engine =
+        p->copy_threshold > 0 &&
+        nbytes >= p->copy_threshold &&
+        nbytes <= p->copy_bytes;
```
gaugarg-nv: Why use chunk-kernel path for > 32MB? Isn't this supposed to be slow?
I think you can hit such sizes if you increase the ubatch size from the default 512.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,811 @@
...
+            ggml_cuda_ar_pipeline_free(p);
+            return nullptr;
+        }
+    }
+
+    // Arrival ring: cache-line padded so each GPU's int is on its own line.
+    const size_t arrival_bytes =
+        (size_t)GGML_CUDA_AR_POOL_SIZE * n_devices * GGML_CUDA_AR_ARRIVAL_STRIDE;
+    if (cudaHostAlloc(reinterpret_cast<void **>(&p->arrival), arrival_bytes,
+                      cudaHostAllocPortable) != cudaSuccess) {
```
gaugarg-nv: I see that you are directly accessing these buffers into the CUDA kernel. This will not be safe if on a system `cudaDevAttrCanUseHostPointerForRegisteredMem` is 0. You need to use `cudaHostGetDevicePointer` API to get the mapped pointer.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,811 @@
...
+    }
+    memset(p->arrival, 0, arrival_bytes);
+
+    // Per-device pinned staging buffers — POOL_SIZE-deep ring so the chunked-
+    // kernel can write the next slot's data while the peer is still reading
+    // the previous slot's. Indexed by (slot * buf_bytes) at the call site.
+    p->buf_bytes = GGML_CUDA_AR_MAX_BYTES;
+    const size_t host_buf_total = (size_t) GGML_CUDA_AR_POOL_SIZE * p->buf_bytes;
+    for (int i = 0; i < n_devices; ++i) {
+        if (cudaHostAlloc(&p->host_buf[i], host_buf_total, cudaHostAllocPortable) != cudaSuccess) {
```
gaugarg-nv: Same as above. You are directly accessing these buffers in the CUDA kernel. This will not be safe if on a system, `cudaDevAttrCanUseHostPointerForRegisteredMem` is 0. You need to use `cudaHostGetDevicePointer` API to get the mapped pointer.

gaugarg-nv (commented):
@scutler-nv Please add perf data with this change-
* Windows - previous fallback vs new implementation
* Linux - previous fallback vs new implementation vs NCCL

gaugarg-nv:
> @scutler-nv Please add perf data with this change-
> 
> * Windows - previous fallback vs new implementation
> * Linux - previous fallback vs new implementation vs NCCL

I would also suggest to limit this path to Windows if NCCL performance on Linux is comparable to this implementation.

ggml/src/ggml-cuda/allreduce.cu
scutler-nv: Thanks, I didn't know that it was unsafe! I've reworked things a bit here and the only remaining host access is the memset, which I'll change to a cudaMemset.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,934 @@
...
+
+        for (size_t c = 0; c < copy_chunks; ++c) {
+            const size_t offset = c * chunk_bytes;
+            const size_t this_bytes = (nbytes - offset) < chunk_bytes ?
+                (nbytes - offset) : chunk_bytes;
+
+            CUDA_CHECK(cudaStreamWaitEvent(p->streams[i], p->ev_pool[peer][slot].cpy[c]));
+            CUDA_CHECK(cudaMemcpyAsync(
+                p->dev_tmp[i] + offset, p->host_large[peer].dev + offset, this_bytes,
+                cudaMemcpyHostToDevice, p->streams[i]));
```
gaugarg-nv, line 671: IIUC, an explicit `cudaMemcpyHostToDevice` copy should use host pointer as source, not the mapped device pointer.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,934 @@
...
+        }
+
+        for (size_t c = 0; c < copy_chunks; ++c) {
+            const size_t offset = c * chunk_bytes;
+            const size_t this_bytes = (nbytes - offset) < chunk_bytes ?
+                (nbytes - offset) : chunk_bytes;
+
+            CUDA_CHECK(cudaMemcpyAsync(
+                p->host_large[i].dev + offset, reinterpret_cast<char *>(src_buf[i]) + offset, this_bytes,
+                cudaMemcpyDeviceToHost, p->streams[i]));
```
gaugarg-nv, line 640: IIUC, an explicit `cudaMemcpyDeviceToHost` copy should use host pointer as destination, not the mapped device pointer.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1247,181 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+    }
+
+    ret->ar_pipeline = ggml_cuda_ar_pipeline_init(ret->dev_ids.data(), n_backends);
+    if (ret->ar_pipeline == nullptr) {
+        // Clear any sticky CUDA error from the failed init so it can't leak
+        // into a later NCCL call.
+        (void) cudaGetLastError();
+        GGML_LOG_WARN("%s: internal AllReduce pipeline init failed; "
+                      "falling back to NCCL or butterfly\n", __func__);
+    }
```
gaugarg-nv: Can we avoid initializing the pipeline if "internal" mode is not going to be used because NCCL is available, or env variable to use "internal" is not set.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1247,181 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+#if defined(GGML_USE_NCCL) && !defined(GGML_USE_HIP) && !defined(GGML_USE_MUSA) && !defined(_WIN32)
+    if (!ggml_backend_cuda_comm_ensure_nccl(ret))
+    {
+        static bool warning_printed = false;
+        if (!warning_printed) {
+            GGML_LOG_WARN("%s: NVIDIA Collective Communications Library (NCCL) is unavailable, "
+                          "multi GPU performance will be suboptimal\n", __func__);
+            warning_printed = true;
+        }
+    }
```
gaugarg-nv: I think we can do the same here. if user has explicitly asked for "internal" mode, don't initialize NCCL.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1247,181 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+// unavailable, so the short-circuit OR naturally falls through to the next.
+static bool ggml_backend_cuda_comm_try_allreduce_os(
+        ggml_backend_cuda_comm_context * comm_ctx, struct ggml_tensor ** tensors) {
+#if defined(__linux__)
+    return ggml_backend_cuda_comm_try_allreduce_nccl(comm_ctx, tensors)
+        || ggml_backend_cuda_comm_try_allreduce_internal(comm_ctx, tensors);
+#else
+    return ggml_backend_cuda_comm_try_allreduce_internal(comm_ctx, tensors)
+        || ggml_backend_cuda_comm_try_allreduce_nccl(comm_ctx, tensors);
+#endif
```
gaugarg-nv: Are there cases where NCCL fails, and we need to fall back to "internal"?
I would prefer that we don't fall back from one option to another.

It should be either NCCL-initialized or "internal"- initialized, not both. 
Butterfly can serve as a fallback if really needed. But I would prefer to avoid that too. Ideally, we should be able to detect failing cases at comm init time rather than during the actual AllReduce.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1247,181 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+    GGML_ASSERT(n_backends >= 1);
+    GGML_ASSERT(tensors[0] != nullptr);
+
+    const int64_t   ne   = ggml_nelements(tensors[0]);
+    const ggml_type type = tensors[0]->type;
+
+    if (n_backends != 2) {
+        GGML_LOG_DEBUG("%s: internal unsupported: n_backends=%zu\n", __func__, n_backends);
+        return false;
+    }
```
gaugarg-nv: This check can be removed as we check this during pipeline initialization. Add an `assert` instead.

ggml/src/ggml-cuda/allreduce.cu
scutler-nv: Re-read the docs and I think you're right. Fixed.

ggml/src/ggml-cuda/ggml-cuda.cu
scutler-nv: Fixed.

ggml/src/ggml-cuda/allreduce.cu
scutler-nv: Done.

ggml/src/ggml-cuda/ggml-cuda.cu
scutler-nv: I changed it to use lazy initialization, like NCCL. It'll initialize on first use if it's hit as a fallback if NCCL fails for whatever reason (either not enabled, or support doesn't exist at all, or init fails for some other reason).

ggml/src/ggml-cuda/ggml-cuda.cu
scutler-nv: Internal mode only supports two GPUs, at least for now. Wouldn't we want a fallback in that case? Shouldn't a graceful failure (with message) to a fallback be better than total failure?

ggml/src/ggml-cuda/ggml-cuda.cu
scutler-nv: You previously requested that I keep that warning in the init code, but it only makes sense if we tried to init and failed (or NCCL is just totally unsupported). So I'm uncertain what you want here. Maybe just warn if the lazy init fails whether it's NCCL or internal?

ggml/src/ggml-cuda/ggml-cuda.cu
scutler-nv: Done.

ggml/src/ggml-cuda/ggml-cuda.cu
gaugarg-nv: If there are more than two GPUs, the "internal" comm pipeline should fail at initialize time, and we should throw an error that "internal" mode cannot be used.

We should have no fallback paths during the actual AllReduce.

ggml/src/ggml-cuda/ggml-cuda.cu
gaugarg-nv: My suggestion is to not initialize NCCL if user has forced use of "internal" mode. We should either initialize "internal" pipeline or "nccl", not both.
We should keep the warning if NCCL is the preferred path, but NCCL is not available.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1251,189 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+            return;
+        }
+        comm_ctx->nccl_init_ok = true;
+    });
+    return comm_ctx->nccl_init_ok;
+}
+
+static bool ggml_backend_cuda_comm_try_allreduce_nccl(
+        ggml_backend_cuda_comm_context * comm_ctx, struct ggml_tensor ** tensors) {
+    if (!ggml_backend_cuda_comm_ensure_nccl(comm_ctx)) {
```
gaugarg-nv: Why NCCL is being initialized during first AllReduce and not during `ggml_backend_cuda_comm_init`?

We should do the same for "internal". Do all the initialization during `ggml_backend_cuda_comm_init`.
At `ggml_backend_cuda_comm_init` time, you already know which path to use based on env variables, OS, or NCCL availability.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+//     outer chunker that issues sequential inner calls.
+// ---------------------------------------------------------------------------
+
+// ---------------------------------------------------------------------------
+// Cross-GPU signal mechanism
+//
+// One int per (slot, rank) pair in pinned host memory.  Each AR call writes a
+// strictly increasing token (= the AR call number) into its own arrival int.
+// The peer spins until its read of the other's arrival int equals the token
+// it expects for this call — a mismatch means the peer hasn't arrived yet.
```
JohannesGaessler:
```suggestion
// it expects for this call - a mismatch means the peer hasn't arrived yet.
```

Please stick to ASCII unless you have a good reason not to.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+static __device__ __forceinline__ int ggml_cuda_ar_signal_get(const int * p) {
+    return *(const volatile int *)p;
+}
+
+// Byte spacing between adjacent arrival ints.  64 bytes (one cache line)
+// ensures each GPU/block's arrival slot lives on its own line, preventing
+// false-sharing stalls on the polling GPU.
+static constexpr size_t GGML_CUDA_AR_ARRIVAL_STRIDE = 64;
+
+// Number of blocks the chunked-kernel launches with.  Each block stripes a
```
JohannesGaessler:
```suggestion
// Number of blocks the chunked kernel launches with.  Each block stripes a
```

I'm ESL but I think there shouldn't be a hyphen here.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+// disjoint slice of the data and synchronizes through its own arrival-token
+// slot so multiple SMs can pump PCIe stores in parallel.
+static constexpr int GGML_CUDA_AR_KERNEL_BLOCKS = 8;
+
+// ---------------------------------------------------------------------------
+// Chunked-kernel AllReduce — 2 GPUs, supports float, half, and bfloat16.
```
JohannesGaessler:
```suggestion
// Chunked-kernel AllReduce - 2 GPUs, supports float, half, and bfloat16.
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+//
+// Both GPUs run this kernel simultaneously on independent streams.  sendbuf
+// and recvbuf live in Tdst (the caller's tensor type); host_mine / host_other
+// carry data in Twire (the on-wire type, possibly narrower than Tdst — e.g.
```
JohannesGaessler:
```suggestion
// carry data in Twire (the on-wire type, possibly narrower than Tdst - e.g.
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+// Tdst=F32 with Twire=BF16 halves the bytes pushed across PCIe).  When
+// Tdst == Twire the casts below are no-ops.
+//
+// Each GPU runs three phases:
+//
+//   Phase 1 (all threads): cast sendbuf (Tdst) → Twire and store as 16-byte
```
JohannesGaessler:
```suggestion
//   Phase 1 (all threads): cast sendbuf (Tdst) -> Twire and store as 16-byte
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+//                          vectors into host_mine.  __threadfence_system()
+//                          commits these writes to host memory.
+//   Phase 2 (thread 0):    write token to arrival_mine; spin until
+//                          arrival_other == token.
+//   Phase 3 (all threads): read 16-byte Twire vectors from host_other, cast
+//                          each element to Tdst, and sum with the local
+//                          sendbuf value (also rounded through Twire so that
+//                          both GPUs truncate identically — this guarantees
```
JohannesGaessler:
```suggestion
//                          both GPUs truncate identically - this guarantees
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+//                          bit-equivalent results across the two devices).
+//
+// Multi-block: blocks stripe vectors across (gridDim.x * blockDim.x) global
+// threads to keep multiple SMs issuing PCIe stores in parallel.  Each block
+// has its own arrival-token slot (offset by blockIdx.x * ARRIVAL_STRIDE);
+// thread 0 of each block signals/spins on that slot independently of other
+// blocks.  Tail elements (the leftover < ELEMS_PER_VEC at the end) are
+// handled only by block 0 to avoid cross-block writes to the same slots.
+// ---------------------------------------------------------------------------
+template <typename Tdst, typename Twire>
```
JohannesGaessler:
```suggestion
template <typename T_dst, typename T_wire>
```

This would I think be slightly more consistent with the surrounding code but either way is fine.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+    {
+        for (int i = gtid; i < count_vec; i += gnt) {
+            const int off = i * ELEMS_PER_VEC;
+            Twire wire[ELEMS_PER_VEC];
+            #pragma unroll
+            for (int k = 0; k < ELEMS_PER_VEC; ++k) {
+                wire[k] = static_cast<Twire>(sendbuf[off + k]);
+            }
+            *reinterpret_cast<float4 *>(&host_mine[off]) =
+                *reinterpret_cast<const float4 *>(wire);
```
JohannesGaessler: There is pre-existing infrastructure for aligned copies with > 4 bytes per thread, namely `ggml_cuda_memcpy_1` as defined in `common.cuh`. Please use that instead since it correctly handles edge cases where the maximum number of bytes that a single thread can copy is variable.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+    const int tail      = count_vec * ELEMS_PER_VEC;
+
+    // Phase 1: cast sendbuf (Tdst) -> host_mine (Twire) and store as 16-byte vectors.
+    {
+        for (int i = gtid; i < count_vec; i += gnt) {
+            const int off = i * ELEMS_PER_VEC;
+            Twire wire[ELEMS_PER_VEC];
+            #pragma unroll
+            for (int k = 0; k < ELEMS_PER_VEC; ++k) {
+                wire[k] = static_cast<Twire>(sendbuf[off + k]);
```
JohannesGaessler: There is pre-existing infrastructure for per-thread data conversion, namely `ggml_cuda_cast` as defined in `convert.cuh`. Please use it to avoid compilation failure annoyances.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+        }
+    }
+
+    // Commit this block's host writes before signalling.
+    __threadfence_system();
+    __syncthreads();
+
+    // Phase 2: thread 0 of each block signals on its own arrival slot, then
+    // spins for the matching slot from peer.  Per-block tokens mean blocks
+    // proceed independently — no inter-block barrier needed.
```
JohannesGaessler:
```suggestion
    // proceed independently - no inter-block barrier needed.
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+                static_cast<Tdst>(d_low) + static_cast<Tdst>(host_other[tail + tid]);
+        }
+    }
+}
+
+// Combined load-convert-add kernel.  The peer's contribution arrives as Tsrc
+// (which may be a lower-precision type than Tdst when the BF16 round-trip is
+// active).  For bit-equivalence between the two GPUs, dst is first rounded
+// through Tsrc's precision via a static_cast — peer already truncated its own
+// value the same way before sending — so both sides perform identical
```
JohannesGaessler:
```suggestion
// through Tsrc's precision via a static_cast - peer already truncated its own
// value the same way before sending - so both sides perform identical
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Pipeline structure
+// ---------------------------------------------------------------------------
+
+// Number of slots in the event / arrival ring.  Two slots is sufficient:
+// lockstep guarantees the two GPUs are at most one AR (or chunk) apart, so
+// slot[N%2] is always safe to reuse — peer has already consumed slot[N%2]
```
JohannesGaessler:
```suggestion
// slot[N%2] is always safe to reuse - peer has already consumed slot[N%2]
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+static constexpr size_t GGML_CUDA_AR_COPY_THRESHOLD_DEFAULT = 1024 * 1024; // 1 MB
+// Per-call CE chunk-size heuristic: chunk_bytes = clamp(nbytes / 4, MIN, MAX).
+// The /4 keeps ~4 chunks in flight at any moment (good D2H/H2D overlap with
+// the peer); the clamps cover the cases where nbytes/4 is too small (per-
+// memcpy fixed cost dominates) or too large (chunk-level pipelining stalls).
+// Env var GGML_CUDA_AR_COPY_CHUNK_BYTES can override with a fixed value.
+static constexpr size_t GGML_CUDA_AR_COPY_CHUNK_BYTES_HEURISTIC_MIN = 512 * 1024;       // 512 KB
+static constexpr size_t GGML_CUDA_AR_COPY_CHUNK_BYTES_HEURISTIC_MAX = 2 * 1024 * 1024;  // 2 MB
+// Absolute floor that an env-var override is allowed to set; this caps the
+// per-slot copy-event array.  256 KB → up to 128 chunks per 32 MB tensor.
```
JohannesGaessler:
```suggestion
// per-slot copy-event array.  256 KB -> up to 128 chunks per 32 MB tensor.
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+static constexpr size_t GGML_CUDA_AR_COPY_CHUNK_BYTES_MIN = 256 * 1024;
+static constexpr int GGML_CUDA_AR_COPY_MAX_CHUNKS =
+    static_cast<int>((GGML_CUDA_AR_COPY_MAX_BYTES + GGML_CUDA_AR_COPY_CHUNK_BYTES_MIN - 1) /
+                    GGML_CUDA_AR_COPY_CHUNK_BYTES_MIN);
+
+struct ggml_cuda_ar_event_slot {
+    cudaEvent_t app = nullptr;  // upstream computation complete
+    cudaEvent_t cpy[GGML_CUDA_AR_COPY_MAX_CHUNKS] = {};  // copy-engine D2H chunks complete
+    cudaEvent_t h2d = nullptr;  // copy-engine H2Ds complete (handoff AR stream → compute stream)
```
JohannesGaessler:
```suggestion
    cudaEvent_t h2d = nullptr;  // copy-engine H2Ds complete (handoff AR stream -> compute stream)
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+    cudaEvent_t ker = nullptr;  // AllReduce kernel complete
+};
+
+// Mapped pinned host allocation: cudaHostAlloc + cudaHostGetDevicePointer
+// in one place, with the host handle preserved for cudaFreeHost.  Used where
+// the CPU never touches the buffer — only the device reads/writes via the
```
JohannesGaessler:
```suggestion
// the CPU never touches the buffer - only the device reads/writes via the
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+
+    // Copy-engine: per-device "my add_kernel is done with dev_tmp" event.
+    // Recorded on the compute stream after each add_kernel; the AR stream
+    // waits on it before the next copy_impl's H2D overwrites dev_tmp.  Lets us
+    // single-buffer dev_tmp despite add_kernel running on a separate stream.
+    cudaEvent_t              dev_tmp_kernel_done[GGML_CUDA_MAX_DEVICES];
+    bool                     dev_tmp_kernel_done_valid;
+
+    // Arrival ring: ARRIVAL_STRIDE bytes between adjacent ints.  Mapped pinned
+    // memory; CPU never reads/writes — only the kernel and cudaMemset.
```
JohannesGaessler:
```suggestion
    // memory; CPU never reads/writes - only the kernel and cudaMemset.
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+    }
+    ggml_cuda_set_device(p->devices[0]);
+    if (cudaMemset(p->arrival.dev, 0, arrival_bytes) != cudaSuccess) {
+        GGML_LOG_ERROR("%s: cudaMemset for arrival ring failed (%zu bytes)\n",
+                       __func__, arrival_bytes);
+        ggml_cuda_ar_pipeline_free(p);
+        return nullptr;
+    }
+
+    // Per-device pinned staging buffers — POOL_SIZE-deep ring so the chunked-
```
JohannesGaessler:
```suggestion
    // Per-device pinned staging buffers - POOL_SIZE-deep ring so the chunked-
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+    const int slot = ggml_cuda_ar_acquire_slot(p).slot;
+    const size_t copy_chunks = (nbytes + chunk_bytes - 1) / chunk_bytes;
+    GGML_ASSERT(copy_chunks <= GGML_CUDA_AR_COPY_MAX_CHUNKS);
+
+    ggml_backend_cuda_context * cuda_ctx[2] = {};
+
+    // Stage 1: both GPUs copy their local contribution to pinned host memory.
+    for (int i = 0; i < 2; ++i) {
+        ggml_cuda_set_device(p->devices[i]);
+        cuda_ctx[i] = static_cast<ggml_backend_cuda_context *>(backends[i]->context);
```
JohannesGaessler, line 615:
```suggestion
        ggml_cuda_set_device(p->devices[i]);
        cuda_ctx[i] = static_cast<ggml_backend_cuda_context *>(backends[i]->context);
        GGML_ASSERT(cuda_ctx[i].device == p->devices[i]);
```

Both the pipeline and a `ggml_backend_cuda_context` specify devices with which they're associated. I think it makes sense to assert that they match to prevent accidental misuse.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+
+        ggml_cuda_ar_wait_for_compute(p, cuda_ctx[i], i, slot);
+
+        // Wait for peer's H2D from our host_large[i] (recorded in the
+        // previous AR's stage 2) to complete before we overwrite host_large[i].
+        // host_large_read_done[peer] = peer finished reading host_large[i].
+        // No-op on the first AR — no prior record exists.
```
JohannesGaessler:
```suggestion
        // No-op on the first AR - no prior record exists.
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+                cudaMemcpyDeviceToHost, p->streams[i]));
+            CUDA_CHECK(cudaEventRecord(p->ev_pool[i][slot].cpy[c], p->streams[i]));
+        }
+    }
+
+    // Stage 2: each GPU waits for each peer D2H chunk, pulls that chunk back to
+    // local device scratch (dev_tmp), then performs one device-local add over
+    // the assembled peer tensor.  The H2Ds run on the AR stream (copy engine)
+    // and the add_kernel runs on the caller's compute stream, so the AR stream
+    // stays pure-copy and avoids an in-stream copy→compute engine switch every
```
JohannesGaessler:
```suggestion
    // stays pure-copy and avoids an in-stream copy -> compute engine switch every
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+    p->host_large_read_done_valid = true;
+    p->dev_tmp_kernel_done_valid = true;
+
+    return true;
+}
+
+// Outer-level chunker: copy_impl handles up to copy_bytes per call (limited by
+// the host_large / dev_tmp allocation size).  When the full AR exceeds that,
+// slice the tensor into copy_bytes-sized pieces and call copy_impl repeatedly.
+// Each slice goes through its own stage 1 → stage 2 cycle and acquires its own
```
JohannesGaessler:
```suggestion
// Each slice goes through its own stage 1 -> stage 2 cycle and acquires its own
```

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+            auto * cuda_ctx = static_cast<ggml_backend_cuda_context *>(backends[i]->context);
+            bf16_tmp[i].pool = &cuda_ctx->pool();
+            bf16_tmp[i].alloc(ne);
+            ggml_cuda_set_device(p->devices[i]);
+            if (compute_flag[i]) {
+                to_bf16(tensors[i]->data, bf16_tmp[i].get(), ne, cuda_ctx->stream());
+            } else {
+                CUDA_CHECK(cudaMemsetAsync(bf16_tmp[i].get(), 0, nbytes, cuda_ctx->stream()));
+            }
+            CUDA_CHECK(cudaGetLastError());
```
JohannesGaessler:
```suggestion
                to_bf16(tensors[i]->data, bf16_tmp[i].get(), ne, cuda_ctx->stream());
                CUDA_CHECK(cudaGetLastError());

            } else {
                CUDA_CHECK(cudaMemsetAsync(bf16_tmp[i].get(), 0, nbytes, cuda_ctx->stream()));
            }
```

This API call should only be needed for the path with actual conversion.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+
+        // Dispatch into copy_impl with explicit src/dst types.  When use_bf16
+        // is on, the wire type is BF16 (src = bf16_tmp) and the accumulator
+        // is F32 (dst = tensors[i]->data); the combined add kernel rounds dst
+        // through BF16 for bit-equivalence and writes F32 directly, so no
+        // post-conversion is needed.  Otherwise src == dst (same native type).
+        if (use_bf16) {
+            GGML_ASSERT(kernel_type == GGML_TYPE_BF16);
+            __nv_bfloat16 * src[GGML_CUDA_MAX_DEVICES];
+            float         * dst[GGML_CUDA_MAX_DEVICES];
```
JohannesGaessler:
```suggestion
            nv_bfloat16 * src[GGML_CUDA_MAX_DEVICES];
            float       * dst[GGML_CUDA_MAX_DEVICES];
```

We are aliasing `nv_bfloat16` for HIP/MUSA so please use that datatype unless there is a reason not to.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+                src[i] = static_cast<__nv_bfloat16 *>(copy_src_ptr[i]);
+                dst[i] = static_cast<float *>(tensors[i]->data);
+            }
+            ok = ggml_cuda_ar_allreduce_copy_outer<__nv_bfloat16, float>(
+                p, backends, src, dst, inner_compute, ne);
+        } else {
+            switch (kernel_type) {
+                case GGML_TYPE_F32: {
+                    float * buf[GGML_CUDA_MAX_DEVICES];
+                    for (int i = 0; i < n; ++i) buf[i] = static_cast<float *>(tensors[i]->data);
```
JohannesGaessler:
```suggestion
                    for (int i = 0; i < n; ++i) {
                        buf[i] = static_cast<float *>(tensors[i]->data);
                    }
```

Preferably use this format for loops, I find it easier to read.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,923 @@
...
+                // zeroed up-front (above), so per-chunk zeroing isn't needed.
+                if (!compute_flag[i] && !use_bf16) {
+                    CUDA_CHECK(cudaMemsetAsync(data, 0, chunk_dst_bytes, stream));
+                }
+
+#define LAUNCH_AR_KERNEL(Tdst, Twire) \
+                ggml_cuda_ar_kernel<Tdst, Twire><<<dim3(GGML_CUDA_AR_KERNEL_BLOCKS), dim3(256), 0, stream>>>( \
+                    reinterpret_cast<const Tdst *>(data), \
+                    reinterpret_cast<Tdst *>(data), \
+                    reinterpret_cast<Twire *>(p->host_buf[i].dev    + (size_t) slot * p->buf_bytes), \
```
JohannesGaessler:
```suggestion
                    reinterpret_cast<Twire *>(p->host_buf[i].dev + (size_t) slot * p->buf_bytes), \
```

ggml/src/ggml-cuda/allreduce.cuh
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "common.cuh"
+#include "ggml-backend-impl.h"
+
+#include <cstddef>
+
+// Opaque pipeline context — owns all pinned buffers, streams, and events.
```
JohannesGaessler:
```suggestion
// Opaque pipeline context - owns all pinned buffers, streams, and events.
```

JohannesGaessler (commented):
I still need to review `ggml-cuda.cu` but I'm currently too tired to do it properly.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1138,79 +1142,53 @@ static const ggml_backend_buffer_type_i ggml_backend_cuda_split_buffer_type_inte
     /* .is_host          = */ ggml_backend_cuda_split_buffer_type_is_host,
 };
 
-#ifdef GGML_USE_NCCL
+// Communication context for multi-GPU AllReduce during tensor parallelism.
+//
+// Created once per meta backend instance. The internal pipeline (if any) is
+// allocated eagerly at init time; NCCL communicators are created lazily on
+// first use so NCCL's init/runtime quirks don't interfere with configurations
+// that never hit the fallback path.
```
```suggestion
// first use to avoid an increase in VRAM consumption.
```

This is the reason why it's being initialized lazily.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1138,79 +1142,53 @@ static const ggml_backend_buffer_type_i ggml_backend_cuda_split_buffer_type_inte
...
-        return;
-    }
-    ggml_backend_cuda_comm_context * comm_ctx = (ggml_backend_cuda_comm_context *) comm_ctx_v;
-    delete comm_ctx;
-#else
-    GGML_UNUSED(comm_ctx_v);
-#endif // GGML_USE_NCCL
-}
+    std::vector<ncclComm_t>     comms;
+#endif
```
```suggestion
#endif // GGML_USE_NCCL
```

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1138,79 +1142,53 @@ static const ggml_backend_buffer_type_i ggml_backend_cuda_split_buffer_type_inte
...
-static void * ggml_backend_cuda_comm_init(ggml_backend_t * backends, size_t n_backends) {
+    ~ggml_backend_cuda_comm_context() {
 #ifdef GGML_USE_NCCL
-    for (size_t i = 0; i < n_backends; i++) {
-        if (!ggml_backend_is_cuda(backends[i])) {
-            return nullptr;
+        for (ncclComm_t comm : comms) {
+            NCCL_CHECK(ncclCommDestroy(comm));
         }
+#endif
```
```suggestion
#endif // GGML_USE_NCCL
```

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1251,218 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+                           __func__, i, ggml_nelements(tensors[i]), (int) tensors[i]->type, ne, (int) type);
+            return false;
+        }
+        if (!ggml_is_contiguously_allocated(tensors[i])) {
+            GGML_LOG_DEBUG("%s: internal unsupported: tensor[%zu] is not contiguously allocated: ne=%" PRId64 " nbytes=%zu packed=%zu type=%d\n",
+                           __func__, i, ne, ggml_nbytes(tensors[i]),
+                           (size_t) ne * ggml_type_size(type) / ggml_blck_size(type), (int) type);
+            return false;
+        }
+        if (((uintptr_t) tensors[i]->data & 0xF) != 0) {
```
JohannesGaessler, line 1295: This should also assert that the size of the tensor in bytes is a multiple of 16.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1251,218 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+static bool ggml_backend_cuda_comm_try_allreduce_nccl(
+        ggml_backend_cuda_comm_context * comm_ctx, struct ggml_tensor ** tensors) {
+#ifdef GGML_USE_NCCL
+    GGML_ASSERT(!comm_ctx->comms.empty());
+    return ggml_backend_cuda_comm_allreduce_nccl(comm_ctx, tensors);
 #else
-    GGML_UNUSED_VARS(comm_ctx_v, tensors);
+    GGML_UNUSED(comm_ctx); GGML_UNUSED(tensors);
+    GGML_ABORT("try_allreduce_nccl unreachable: built without NCCL");
+#endif
```
```suggestion
#endif // GGML_USE_NCCL
```

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1251,218 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+    ctx->comms.resize(n);
+    ncclResult_t rc = ncclCommInitAll(ctx->comms.data(), (int) n, ctx->dev_ids.data());
+    if (rc != ncclSuccess) {
+        ctx->comms.clear();
+        GGML_LOG_ERROR("%s: ncclCommInitAll failed: %s\n", __func__, ncclGetErrorString(rc));
+        return false;
+    }
+    return true;
+}
+#endif
```
```suggestion
#endif // GGML_USE_NCCL
```

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1251,218 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+    if      (env_nccl)     ret->try_allreduce = ggml_backend_cuda_comm_try_allreduce_nccl;
+    else if (env_internal) ret->try_allreduce = ggml_backend_cuda_comm_try_allreduce_internal_strict;
+    else if (env_none)     ret->try_allreduce = ggml_backend_cuda_comm_try_allreduce_butterfly;
+#if defined(_WIN32)
+    else                   ret->try_allreduce = ggml_backend_cuda_comm_try_allreduce_internal_lenient;
+#elif defined(__linux__)
+    else                   ret->try_allreduce = ggml_backend_cuda_comm_try_allreduce_nccl;
+#else
+    else GGML_ABORT("no AllReduce default for this platform; set GGML_CUDA_ALLREDUCE explicitly");
+#endif
```
JohannesGaessler: I think the logic here should be simplified: have one function for trying an internal AllReduce that falls back to the meta backend's generic implementation on failure. Have another function that tries an NCCL AllReduce and falls back to the first function on failure. Decide which function to try based on whether the environment variable is `"internal"` or `"none"`, simply return false for all other values and print a warning for those values that are not `"none"` to inform the user of typos.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1251,218 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+        if (env_nccl) {
+            GGML_ABORT("GGML_CUDA_ALLREDUCE=nccl requested but llama.cpp was not built with NCCL.  "
+                       "Recompile with -DGGML_CUDA_NCCL=ON or reset the environment variable.");
+        }
+        // Linux default with no NCCL compiled: warn and degrade to butterfly.
+        GGML_LOG_WARN_ONCE("NVIDIA Collective Communications Library (NCCL) is unavailable; "
+                           "multi-GPU performance will be suboptimal.  "
+                           "Recompile with -DGGML_CUDA_NCCL=ON for best performance.");
+        ret->try_allreduce = ggml_backend_cuda_comm_try_allreduce_butterfly;
+#endif
```
```suggestion
#endif // GGML_USE_NCCL
```

ggml/src/ggml-cuda/ggml-cuda.cu
gaugarg-nv: @JohannesGaessler IIUC, this is what you are saying:

First function: internal -> butterfly should be the path
Second function: nccl -> "first function"

We expose only two values for env variable:
"internal": Call first function
"none" or default: Call the second function

I think doing this is fine. My request was to have this logic during init time to decide on the AllReduce implementation and not during the actual AllReduce call. During actual AllRedcue, we should have only two paths:
nccl -> abort on failure (as this should never happen if nccl init was successful)
internal -> butterfly (this is needed because internal has limitations in terms of the datatypes it supports)

gaugarg-nv: This also avoid us to initialize both NCCL and "internal" pipelines together. We initialize either NCCL or "internal", not both.

scutler-nv: I've made some simplifications to make it close to what Johannes requested, though I've kept the "nccl" option for the envvar since it is still useful for Windows while NCCL is being developed. Otherwise, it's pretty close. Let me know what you think.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1251,185 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+static void ggml_backend_cuda_comm_init_internal(ggml_backend_cuda_comm_context * ret) {
+    ret->ar_pipeline = ggml_cuda_ar_pipeline_init(ret->dev_ids.data(), ret->dev_ids.size());
+    if (ret->ar_pipeline) {
+        ret->try_allreduce = ggml_backend_cuda_comm_try_allreduce_internal;
+        return;
+    }
+
+    // Clear sticky CUDA error from the failed init.
+    (void) cudaGetLastError();
+    GGML_LOG_WARN("internal AllReduce init failed (n_devices != 2 or pre-Ampere?); "
```
JohannesGaessler: Why would pre-Ampere GPUs be problematic? I don't remember seeing something in the kernel that needed Ampere or newer.

ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1274,10 +1251,185 @@ static bool ggml_backend_cuda_comm_allreduce_tensor(void * comm_ctx_v, struct gg
...
+        ret->try_allreduce = ggml_backend_cuda_comm_try_allreduce_nccl;
+        return;
+    }
+
+    ret->comms.clear();
+    GGML_LOG_WARN("NCCL init failed (%s); falling back to internal AllReduce\n",
+                  ncclGetErrorString(rc));
+#else // GGML_USE_NCCL
+    GGML_LOG_WARN("NCCL not compiled in; falling back to internal AllReduce.  "
+                  "Recompile with -DGGML_CUDA_NCCL=ON for best multi-GPU performance.\n");
```
JohannesGaessler, line 1374:
```suggestion
#ifndef GGML_USE_HIP
    GGML_LOG_WARN("NCCL not compiled in; falling back to internal AllReduce.  "
                  "Make sure NCCL is installed and recompile with -DGGML_CUDA_NCCL=ON for best multi-GPU performance.\n");
#endif // GGML_USE_HIP
```

This warning only makes sense to print on NVIDIA like this.

JohannesGaessler (approved):
From my end I would now consider this PR good to merge. 

One issue that the CUDA backend currently has (and that I don't think should block this PR) is that environment variables are not consistent in their behavior and documentation. As of right now *some* environment variables are documented in `docs/build.md` but that's not really the right place for it I think; ideally we would have a table somewhere in a more easily discoverable place that lists all of the environment variables that we have. We should also try and standardize the naming and truthiness with e.g. a numerical value of 0 always being false/off and all other values being true/on with maybe some differentiation.

With this PR I think we can prototype relatively easily a variant of the meta backend that creates a single `ggml_cgraph`s with inserted `GGML_OP_CUSTOM` nodes to do the AllReduce in a way that is capturable by a CUDA graph with minimal changes. It's not clear to me how best do it with NCCL but it should be possible to work out the other issues with the internal AllReduce first which is essentially just regular, single-device operations.

ggml/src/ggml-cuda/ggml-cuda.cu
scutler-nv: Out of date comment (was true earlier). Fixed.

scutler-nv:
I hit some HIP build issues in the allreduce.cu code; I added some stub functions that should fix it.

ggml/src/ggml-cuda/allreduce.cu
```diff
@@ -0,0 +1,953 @@
...
+    // proceed independently -- no inter-block barrier needed.
+    if (tid == 0) {
+        int       * my_slot    = arrival_mine  + bid * ARRIVAL_INTS;
+        const int * other_slot = arrival_other + bid * ARRIVAL_INTS;
+
+        ggml_cuda_ar_signal_set(my_slot, token);
+        __threadfence_system(); // make our signal visible system-wide
+
+        while (ggml_cuda_ar_signal_get(other_slot) != token) {
+            __nanosleep(100);
```
gaugarg-nv, line 165: As per CUDA documentation https://docs.nvidia.com/cuda/cuda-c-programming-guide/#nanosleep-description, `__nanosleep` is supported only on sm70+. Do we need a check for sm70+ during init?

JohannesGaessler: Yes, and please also wrap the device code in an `#if __CUDA_ARCH__ >= GGML_CUDA_CC_VOLTA` with a `NO_DEVICE_CODE` in the else branch.

scutler-nv: Done. I just wrapped the __nanosleep itself; I didn't see anything else that would need it. Technically I could just replace this with a spin loop or something; the sleep itself isn't important (needs some delay so as to not hammer the bus though). But I don't think we much care about pre-Volta.

gaugarg-nv (approved):

TomTheWise:
Nice, this is probably the perfect long time solution for vGPU too, where (by design) NCCL is not available!

LostRuins:
Hi, I'm getting some compilation errors on CUDA after this PR
```
ggml\src\ggml-cuda\allreduce.cu(213): error : more than one conversion function from "half" to a built-in type applies:
              function "__half::operator float() const" (declared at line 217 of C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda_fp16.hpp)
              function "__half::operator short() const" (declared at line 235 of C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda_fp16.hpp)
              function "__half::operator unsigned short() const" (declared at line 238 of C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda_fp16.hpp)
              function "__half::operator int() const" (declared at line 241 of C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda_fp16.hpp)
              function "__half::operator unsigned int() const" (declared at line 244 of C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda_fp16.hpp)
              function "__half::operator long long() const" (declared at line 247 of C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda_fp16.hpp)
              function "__half::operator unsigned long long() const" (declared at line 250 of C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda_fp16.hpp)
              function "__half::operator __nv_bool() const" (declared at line 254 of C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda_fp16.hpp)
            dst[i] = ggml_cuda_cast<T_dst>(d_low) + ggml_cuda_cast<T_dst>(src[i]);
                     ^
            detected during:
              instantiation of "void ggml_cuda_ar_add_kernel(T_dst *, const T_src *, int) [with T_dst=half, T_src=half]" at line 691
              instantiation of "__nv_bool ggml_cuda_ar_allreduce_copy_impl(ggml_cuda_ar_pipeline *, ggml_backend_t *, T_src *const *, T_dst *const *, const __nv_bool *, int64_t, size_t) [with T_src=half, T_dst=half]" at line 735
              instantiation of "__nv_bool ggml_cuda_ar_allreduce_copy_outer(ggml_cuda_ar_pipeline *, ggml_backend_t *, T_src *const *, T_dst *const *, const __nv_bool *, int64_t) [with T_src=half, T_dst=half]" at line 872
```

JohannesGaessler:
Please open an issue and fill out the "compile bug" template.

LostRuins:
done, https://github.com/ggml-org/llama.cpp/issues/22974

PR #22303, opened 2026-04-23
CI Snapdragon: Switch ubuntu-latest to ubuntu-slim runner

shreyajn:
- AI usage disclosure: No

### DIFF
--- a/.github/workflows/build-and-test-snapdragon.yml
+++ b/.github/workflows/build-and-test-snapdragon.yml
@@ -58,18 +58,10 @@ jobs:
           name: llama-cpp-android-arm64-snapdragon
           path: pkg-snapdragon/llama.cpp
 
-  check-secret:
-    runs-on: ubuntu-slim
-    outputs:
-      has-key: ${{ steps.check.outputs.has-key }}
-    steps:
-      - id: check
-        run: echo "has-key=${{ secrets.QDC_API_KEY != '' }}" >> "$GITHUB_OUTPUT"
-
   test-snapdragon-qdc:
     name: Test on QDC Android Device (${{ matrix.device }})
-    needs: [android-ndk-snapdragon, check-secret]
-    if: needs.check-secret.outputs.has-key == 'true'
+    needs: [android-ndk-snapdragon]
+    if: ${{ secrets.QDC_API_KEY != '' }}
     runs-on: ubuntu-slim
     strategy:
       fail-fast: false
--- a/.github/workflows/build-and-test-snapdragon.yml
+++ b/.github/workflows/build-and-test-snapdragon.yml
@@ -49,17 +49,17 @@ jobs:
           cp docs/backend/snapdragon/CMakeUserPresets.json .
           cmake --preset arm64-android-snapdragon-release -B build
           cmake --build build
-          cmake --install build --prefix pkg-adb/llama.cpp
+          cmake --install build --prefix pkg-snapdragon/llama.cpp
 
       - name: Upload Llama.CPP Snapdragon Android Build Artifact
         if: ${{ always() && steps.build_llama_cpp_snapdragon_android.outcome == 'success' }}
         uses: actions/upload-artifact@v6
         with:
           name: llama-cpp-android-arm64-snapdragon
-          path: pkg-adb/llama.cpp
+          path: pkg-snapdragon/llama.cpp
 
   check-secret:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     outputs:
       has-key: ${{ steps.check.outputs.has-key }}
     steps:
--- a/.github/workflows/build-and-test-snapdragon.yml
+++ b/.github/workflows/build-and-test-snapdragon.yml
@@ -61,7 +61,6 @@ jobs:
   test-snapdragon-qdc:
     name: Test on QDC Android Device (${{ matrix.device }})
     needs: [android-ndk-snapdragon]
-    if: ${{ secrets.QDC_API_KEY != '' }}
     runs-on: ubuntu-slim
     strategy:
       fail-fast: false
@@ -96,6 +95,7 @@ jobs:
           pip install qdc_sdk/qualcomm_device_cloud_sdk-0.2.3-py3-none-any.whl
 
       - name: Run QDC tests (${{ matrix.device }})
+        if: ${{ env.QDC_API_KEY != '' }}
         run: |
           python scripts/snapdragon/qdc/run_qdc_jobs.py \
               --test       all \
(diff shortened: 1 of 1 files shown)


max-krasnyansky (approved):
That was quick 👍

Aflah012 (approved):

.github/workflows/build-and-test-snapdragon.yml
```diff
@@ -70,7 +70,7 @@ jobs:
     name: Test on QDC Android Device (${{ matrix.device }})
     needs: [android-ndk-snapdragon, check-secret]
     if: needs.check-secret.outputs.has-key == 'true'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
```
CISC, line 64: Put `check-secret` on `ubuntu-slim` as well.

CISC (approved):

.github/workflows/build-and-test-snapdragon.yml
```diff
@@ -84,14 +84,19 @@ jobs:
         uses: actions/download-artifact@v4
```
CISC:
```suggestion
        uses: actions/download-artifact@v7
```

.github/workflows/build-and-test-snapdragon.yml
```diff
@@ -84,14 +84,19 @@ jobs:
...
         with:
           name: llama-cpp-android-arm64-snapdragon
-          path: pkg-snapdragon/
+          path: pkg-snapdragon/llama.cpp
```
CISC, line 78: Fix the path in `Upload Llama.CPP Snapdragon Android Build Artifact`, it's the old Android path.

.github/workflows/build-and-test-snapdragon.yml
CISC: Though, why do we need `check-secret`? You should be able to do the evaluation directly in the `if` statement.

max-krasnyansky: Yep. Probably not needed anymore. Had some issue accessing the secrets initially (ie they are not available in unmerged PRs for security, etc) so it was helpful to have an explicit check.

max-krasnyansky (approved):

CISC (approved):

.github/workflows/build-and-test-snapdragon.yml
CISC: So, remove it now?

shreyajn: Let me try it.

.github/workflows/build-and-test-snapdragon.yml
shreyajn: Thanks, sorry the GitHub Enterprise is usually behind where I test, which is where it carried over from.

.github/workflows/build-and-test-snapdragon.yml
shreyajn: Good catch!

CISC (approved):

.github/workflows/build-and-test-snapdragon.yml
CISC: Ah, no, this is not actually supported, `secrets` is not available at top level.

.github/workflows/build-and-test-snapdragon.yml
```diff
@@ -49,28 +49,20 @@ jobs:
...
-        run: echo "has-key=${{ secrets.QDC_API_KEY != '' }}" >> "$GITHUB_OUTPUT"
+          path: pkg-snapdragon/llama.cpp
 
   test-snapdragon-qdc:
     name: Test on QDC Android Device (${{ matrix.device }})
-    needs: [android-ndk-snapdragon, check-secret]
-    if: needs.check-secret.outputs.has-key == 'true'
-    runs-on: ubuntu-latest
+    needs: [android-ndk-snapdragon]
+    if: ${{ secrets.QDC_API_KEY != '' }}
```
CISC:
```suggestion
```
Let's just move the check down to `Run QDC tests`, shouldn't matter much if the other stuff runs.

shreyajn: It would download the artifact etc though?

shreyajn: If you're okay though with other steps, I'd fine too. I have moved the check further down per suggestion.

CISC: Yes, but not a big deal is it, only a "problem" for forks making changes without the secret?

.github/workflows/build-and-test-snapdragon.yml
```diff
@@ -81,24 +72,30 @@ jobs:
...
+          sudo apt-get install -y curl unzip
+
       - name: Install QDC SDK wheel
         run: |
           curl -fSL -o qdc_sdk.zip https://softwarecenter.qualcomm.com/api/download/software/tools/Qualcomm_Device_Cloud_SDK/All/0.2.3/qualcomm_device_cloud_sdk-0.2.3.zip
           unzip qdc_sdk.zip -d qdc_sdk
           pip install qdc_sdk/qualcomm_device_cloud_sdk-0.2.3-py3-none-any.whl
 
       - name: Run QDC tests (${{ matrix.device }})
+        if: ${{ env.QDC_API_KEY != '' }}
```
CISC:
```suggestion
        if: ${{ secrets.QDC_API_KEY != '' }}
```
You can't use the env defined in the same step, but `secrets` should be fine here.

CISC: Sigh, looks like I'm wrong:
https://github.com/ggml-org/llama.cpp/actions/runs/24905501222

CISC: So, let's add a step that sets it before then.

CISC: Basically just reintroduce `check-secret` step in this job.

shreyajn: Why can't I see these failing on the PR?

CISC: Because it fails on parsing level, the workflow is never run.

CISC (approved):
Works, but no `QDC_API_KEY`?

shreyajn:
> Works, but no `QDC_API_KEY`?

@max-krasnyansky added it a while ago. Must be in repo settings.

CISC:
> > Works, but no `QDC_API_KEY`?
> 
> @max-krasnyansky added it a while ago. Must be in repo settings.

@ggerganov ?

Edit: Ah, nvm, I see it.
> They are not passed to workflows that are triggered by a pull request from a fork.

PR #22316, opened 2026-04-24
gitignore : add .pi + personal SYSTEM.md

ggerganov:
## Overview

Allow personal system prompts when using the `pi` agent.

Personal system prompts live in subdirectories. They can be symlinked to the `.pi/SYSTEM.md` to be activated.

# Requirements

- AI usage disclosure: NO


.pi/gg/SYSTEM.md
CISC, line 1: I guess either the path of this file, or the `.gitignore` entry needs to be updated? :)

ggerganov: The idea is to symlink the system prompt that you want to use. I've update the description to clarify this.

CISC:
Got a link to the agent?

ggerganov:
> Got a link to the agent?

https://pi.dev

jacekpoplawski:
pi works much better with llama.cpp than opencode

am17an:
Perhaps create a tutorial for pi.dev + llama.cpp, seems like a good combo

tarruda:
> Perhaps create a tutorial for pi.dev + llama.cpp, seems like a good combo

Here's a simple one:

- install pi with `npm install -g @mariozechner/pi-coding-agent`
- create `~/.pi/agent/models.json` with something like this (which is what I use with Qwen 3.6 35B):

```json
{
  "providers": {
    "local-llm": {
      "baseUrl": "http://127.0.0.1:8080/v1",
      "api": "openai-completions",
      "apiKey": "local",
      "models": [
        {
          "id": "llama.cpp",
          "contextWindow": 262144,
          "maxTokens": 80000
        }
      ]
    }   
  }
}
```
- Run `pi` and enter `/model`, which will show an interactive model selection. Then you type llama.cpp and select it.

ggerganov:
Also can start it without context and use the system prompt as explained here.

```bash
# skips `AGENTS.md`, but loads `.pi/SYSTEM.md` if present
pi --offline --no-context-files
```

abitrolly:
> Perhaps create a tutorial for pi.dev + llama.cpp, seems like a good combo

Would be nice to get some chapter on security. How to run it all without leaking your SSH keys and all other files, through pi.dev or llama.cpp server with public IP.

am17an:
Speaking of security (or the lack thereof), I vibe-coded (using the shiny new Deepseek v4 pro max) a pi extension for llama-server, https://www.npmjs.com/package/pi-llama-server, which does what I want (i.e. want to change load/unload models on the fly with some presets in `models.ini`)

ngxson:
It's probably time to add a shortcut `llama-cli --launch pi` ?

PR #21760, opened 2026-04-11
common/gemma4 : handle parsing edge cases

aldehir:
## Overview

Fix a few edge cases for Gemma 4 26B A4B. I don't see these artifacts from the 31B variant.

## Issue 1

If the model generates content + tool call, the template will incorrectly format the prompt without the generation prompt (`<|turn>model\n`):
```
...<|tool_call>call:$call<tool_call|><|tool_response>response:$response<tool_response|>$message<turn|>\n
```
Causing 26B to produce a broken thinking sequence:
```
thought\n<channel|>
```
Instead of
```
<|channel>thought\n<channel|>
```
This is fixed by adding the generation prompt if not present and the prompt ends with `<turn|>\n`.

## Issue 2

Occasionally 26B will emit a trailing `<channel|>`, particularly when it does not reason but produces a content message before a tool call:
```
<|channel>thought\n<channel|>I will ...<channel|><|tool_call>`
```
Fixed by scanning until `<channel|>`, then consume until `<|tool_call>` or end.

## Issue 3

At the start of the generation, 26B may emit multiple `<|channel>` tokens.
```
<|channel><|channel>thought\nI will...
```
Unsure if this is related to the bad prompt above, but it's easy enough to handle by consuming all `<|channel>` tokens that do not precede `thought`.

# Requirements

- I have read and agree with the [contributing guidelines](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md)
- AI usage disclosure: NO

PR #22325, opened 2026-04-24
Fix gemma4 prefill parsing

Quairon-Nailo:
## Overview

This PR fixes an issue in the `peg-gemma4` chat format parser where prefilled reasoning blocks cause the parser to fail and silently drop the rest of the model's response.

**The Problem:**
As reported and discussed [here](https://github.com/ggml-org/llama.cpp/pull/21760#issuecomment-4267467172), the parser's `thought` rule currently strictly expects the generation to begin with the `<|channel>` token. When a frontend (like SillyTavern) prefills the `<|channel>` token into the context to force reasoning, the newly generated text begins with `thought\n...` or similar. Because it lacks the leading `<|channel>`, the `thought` rule fails. The parser falls back to evaluating it as standard `content`, which is designed to halt at an unmatched `<channel|>` token to prevent trailing hallucinations. Consequently, the parser stops prematurely right after the reasoning block, truncating the actual response.

**The Solution:**
1. The parser now scans `data.prompt` specifically within the current assistant turn (after `<|turn>model`) to detect if there is an unclosed `<|channel>` tag. If an unclosed tag is detected in the current turn, the parser modifies its root structure to use a `first-message` rule. This bypasses standard choice (`|`) evaluations, eliminating PEG buffering and allowing the prefilled reasoning block to stream live to the user. A new `resumed-thought` rule uses an optional prefix (`"thought"`) and captures all text up to the closing `<channel|>` tag. This gracefully handles both standard tag-only prefills and arbitrary text prefills seamlessly.

# Requirements

- AI usage disclosure: YES - AI assisted in diagnosing the PEG grammar limitations with prefilled tags, formulating the `resumed-thought` fallback rule, and adding the context-boundary check to restore streaming functionality. I have manually reviewed, rigorously tested edge cases, and fully understand the implemented logic.

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1116,7 +1116,8 @@ static common_chat_params common_chat_params_init_gemma4(const common_chat_templ
         }
 
         auto consume_empty_channels = p.gbnf(p.zero_or_more(p.literal("<|channel>") + p.negate(p.literal("thought"))), "");
-        auto thought = (p.peek(p.literal("<|channel>")) + consume_empty_channels + p.ref("thought")) | p.negate(p.literal("<|channel>"));
+        auto resumed_thought = p.rule("thought", p.content(p.until_one_of({"<|channel>", "<channel|>"}) + p.literal("<channel|>")));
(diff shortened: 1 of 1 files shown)


aldehir:
If its primarily a prefill issue, I guess I'll see how we can accommodate. Seems to be a hot issue lately...

PR #22332, opened 2026-04-24
ggml-webgpu: silence subgroup_uniformity diagnostic in mul_mat_vec

SharmaRithik:
## Overview

Adds the `diagnostic(off, subgroup_uniformity);` directive to `mul_mat_vec.wgsl` so it matches the pattern already used by the other subgroup-using shaders in the directory. Without it, stricter Dawn builds reject the shader at validation time and all mat-vec pipelines fail to compile on those systems.

## Additional information

The `subgroupAdd` call in the row-reduction loop (`mul_mat_vec.wgsl:831`) sits inside a for-loop whose trip count depends on `num_subgroups`, which WGSL's uniformity analysis conservatively treats as potentially non-uniform. Some Dawn builds (observed on AMD/RADV with Mesa 25.2.8, Dawn `3f101040aa`) escalate this to a shader validation error, blocking all mat-vec pipelines at shader compile time.

The other subgroup-using shaders in this directory (`flash_attn.wgsl`, `flash_attn_vec_*.wgsl`) already opt out via `diagnostic(off, subgroup_uniformity);`. `mul_mat_vec.wgsl` was missing it.

The directive is placed inside the `USE_SUBGROUP_REDUCTION` guard so it only applies when subgroups are enabled. No behavior change on platforms where validation was already passing (verified locally on Intel Arc B580 + Mesa 25.2.8). Confirmed to unblock the failing path on AMD Radeon RX 7900 XT (RADV NAVI31, Mesa 25.2.8, Dawn `3f101040aa`). After the fix, `llama-bench` runs cleanly on Q4_0 (663 tok/s pp, 223 tok/s tg) and IQ1_S (460 tok/s pp, 29 tok/s tg) where both previously errored at shader compile time.

## Reproduction

Without this patch, on AMD/RADV, the following command fails at shader compile time with `'subgroupAdd' must only be called from uniform control flow`:
```
LD_LIBRARY_PATH=build/bin build/bin/llama-bench \
  -m models/Llama-3.2-1B-Instruct-Q4_0.gguf -p 64 -n 128 -r 1 -ngl 99
```

With this patch, the same command runs to completion. Also reproduced across IQ1_S, IQ1_M, IQ2_XXS, IQ3_XXS, IQ4_NL, IQ4_XS (all failed identically before, all pass after).

# Requirements

- AI usage disclosure: no

### DIFF
--- a/ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_vec.wgsl
+++ b/ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_vec.wgsl
@@ -1,4 +1,5 @@
 #ifdef USE_SUBGROUP_REDUCTION
+diagnostic(off, subgroup_uniformity);
 enable subgroups;
 #endif
 enable f16;


reeselevine:
I don’t think this is a sustainable solution, because we can’t disable these checks for safety in the browser, can we? 

What do you mean by “stricter” Dawn builds, shouldn’t the logic be consistent across devices? Is that commit a more recent commit than the one we’re using for CI, or is it older?

SharmaRithik:
Edit: I updated Dawn to the latest version (4654ba883e, April 2026), and the crash no longer happens on AMD RX 7900 XT. The older Dawn I had (3f101040aa, October 2025) was the one rejecting the shader. So this is a Dawn version issue after all.

Since the latest Dawn fixes this, we can probably close this PR. Let me know what you think.

reeselevine:
Yep I think this can be closed. Still might be nice to confirm if disabling the diagnostic is allowed by  browsers. Might come into play for subgroup matrix code soon.

reeselevine:
My understanding of WGSL's diagnostics section is that `subgroup_uniformity` is one of the few diagnostics that can be ignored ("filtered"): https://www.w3.org/TR/WGSL/#filterable-triggering-rules.

So while I think this filtering would still work in the browser, the fact that more recent Dawn versions do not trigger the diagnostic means that we can close this PR. But we'll keep this in mind in case other WebGPU implementations do start triggering it, or we need to filter diagnostics for upcoming features like subgroup matrices.

Issue #20861, opened 2026-03-22
Misc. bug: Qwen3.5 with Opencode: Assistant response prefill is incompatible with enable_thinking.


alefranz:
### Name and Version

version: 8468 (3306dbaef)
built with Clang 19.1.5 for Windows x86_64

### Operating systems

Windows

### Command line

```shell

```

### Problem description & steps to reproduce

when using OpenCode with a Qwen3.5 model with thinking enabled, this error sometimes appear causing the request to fail.

I've tried various combinations of models paramters, llama.cpp settings nad Opencode config with no luck.

any recommendation? My understanding from issues is that this should have been fixed a while ago, but I have been experiencing this error since the release of Qwen3.5 and it happens even on latest build.

### Relevant log output

<details>
<summary>Logs</summary>

```console
srv    operator(): got exception: {"error":{"code":400,"message":"Assistant response prefill is incompatible with enable_thinking.","type":"invalid_request_error"}}
```
</details>

UltraRabbit:
Just add --no-prefill-assistant to avoid this issue.  It's because the opencode didn't receive full response from AI provider.  It sent incomplete assistant message back again and expect AI could continue to generate the rest of response.  That make thinking mode model couldn't handle properly.  Just use the   --no-prefill-assistant as a workaround until there's a fix from llama or opencode.

marciowb:
Using a custom Jinja template worked for me. See:
https://github.com/ggml-org/llama.cpp/issues/20837#issuecomment-4218130554

adrianhoehne:
I also hit that today with version:
version: 8763 (ff5ef8278)
built with GNU 11.4.0 for Linux x86_64
and Qwen 3.6 35B A3B

janvitos:
IDE: Opencode 1.14.18
Model: Qwen3.6 35B A3B IQ4_XS (Bartowski)
Server: llama.cpp b8850

Strangely enough, I encountered the bug twice today even with thinking disabled.

"Assistant response prefill is incompatible with enable_thinking"

gxcreator:
Same,
```
 {"error":{"code":400,"message":"Assistant response prefill is incompatible with enable_thinking.","type":"invalid_request_error"}}

```

Parameters:
```
[...]

Issue #20409, opened 2026-03-11
Eval bug: Bug: Qwen3.5 enable_thinking=false via --chat-template-kwargs is ignored across all shells (PowerShell/Bash)


lucksufe:
### Name and Version

System：windows 10 & 11

llama.cpp version:：llama-b8254-bin-win-cpu-x64/llama-b8270-bin-win-cuda-12.4-x64/llama-b8270-bin-win-cuda-13.1-x64

model：Qwen3.5-27B-Q4_K_M.gguf/Qwen3.5-9B-Q4_K_M.gguf/Qwen3.5-2B-Q8_0.gguf

**Describe the bug**
I am unable to disable the "Thinking" (Chain-of-Thought) output for Qwen3.5 models using `llama.cpp`. The model consistently generates the thinking block regardless of the parameters passed.

**Steps to Reproduce**
I have tried multiple combinations of `--chat-template-kwargs` to force `enable_thinking` to false, but none work. The model ignores these arguments completely.

Commands attempted (all failed):
1. Using single quotes with escaped inner quotes:
   ```bash
   ./main -m qwen3.5.gguf --chat-template-kwargs '{"enable_thinking": false}'

### Operating systems

Windows

### GGML backends

CUDA

### Hardware

CPU:Ryzen 5800 and Ryzen 7840h             GPU:RTX 4080

### Models

Qwen3.5-27B-Q4_K_M.gguf/Qwen3.5-9B-Q4_K_M.gguf/Qwen3.5-2B-Q8_0.gguf

### Problem description & steps to reproduce

 ./llama-b8270-bin-win-cuda-12.4-x64/llama-cli.exe -m model/Qwen3.5-9B-Q4_K_M.gguf --chat-template-kwargs '{"enable_thinking": false}'


### First Bad Commit

_No response_

### Relevant log output

<details>
<summary>Logs</summary>
<!-- Copy-pasted short logs go into the "console" area here -->

```console

```
</details>

MaxKruse:
I had a very similar issue on windows, as far as i can tell its an escaping issue of powershell and cmd.
A workaround that worked for me was either setting 

> $env:LLAMA_CHAT_TEMPLATE_KWARGS='{"enable_thinking": false}'

Or you use the new switch, `--reasoning [on|off|auto]`

grapevine-AI:
Please try it

- PowerShell

```
--chat-template-kwargs '{\"enable_thinking\": false}'
```

- command prompt

```
[...]

PR #22323, opened 2026-04-24
common : re-arm reasoning budget after DONE on new <think>

BruceJillis:
## Overview

DONE state in reasoning budget state machine absorbs start tags, causing any <think> block after the first to run unbudgeted. This makes it so the reasoning budget is a no-op for multi-block thinking models. Using the Qwen3.6-27B model with the recommended settings causes this issue to appear [1]. The fix is to re-arm in DONE on a match and transition to COUNTING with a fresh budget. I've added a regression test in test-reasoning-budget to test for this new behavior and all 6 tests pass.

* [1] "Thinking Preservation: we've introduced a new option to retain reasoning context from historical messages, streamlining iterative development and reducing overhead." - [https://huggingface.co/Qwen/Qwen3.6-27B](https://huggingface.co/Qwen/Qwen3.6-27B)

## Additional information

Reproducible using: `unsloth/Qwen3.6-27B-GGUF`, server flags: `--reasoning-budget 128 --reasoning-format deepseek --jinja`, base commit: master at `15fa3c493` (b8920)

### Request body

```json
{
  "model": "kimi-q36-27b",
  "messages": [
    { "role": "user", "content": "Get the current time, then tell me what it is." },
    {
      "role": "assistant",
      "content": "<think>The user wants the time. I should call get_time to retrieve it.</think>",
      "tool_calls": [
        {
          "id": "call_1",
          "type": "function",
          "function": { "name": "get_time", "arguments": "{}" }
        }
      ]
    },
    { "role": "tool", "tool_call_id": "call_1", "content": "12:00 PM UTC" }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_time",
        "description": "Get the current time.",
        "parameters": { "type": "object", "properties": {} }
      }
    }
  ],
  "temperature": 0.3,
  "max_tokens": 512
}
```

### Before 

```
reasoning-budget: activated, budget=128 tokens
reasoning-budget: deactivated (natural end)
[prompt processing done; 54 tokens generated with no further reasoning-budget events]
```

The second `<think>` (the assistant prefill at end of prompt) hit DONE state and was silently absorbed. The subsequent generation ran without budget enforcement.

### After (this PR)

```
reasoning-budget: activated, budget=128 tokens
reasoning-budget: deactivated (natural end)
reasoning-budget: re-activated on new start tag, budget=128 tokens
reasoning-budget: deactivated (natural end)
```

[...]

PR #22336, opened 2026-04-24
server: respect per-request enable_thinking toggle via extra_body

pju-hoge:
## Overview

Fixes `enable_thinking` being ignored when set per-request via `extra_body` or `chat_template_kwargs`. Previously the server only checked `chat_template_kwargs`, but OpenAI-compatible clients (and tools like Opencode) send it in `extra_body`. This caused `enable_thinking: false` to be silently ignored across all shells.

### `tools/server/server-common.cpp`
Read `enable_thinking` from `extra_body` first (OpenAI-compatible path), then fall back to `chat_template_kwargs`. Pass the value to the slot so it can override the server default.

### `common/chat.cpp`
`supports_thinking` now requires both template-level reasoning detection AND `enable_thinking == true`. Prevents thinking tags from being injected when the user explicitly toggles thinking off.

### `common/chat-auto-parser-generator.cpp`
`extract_reasoning` also gated by `enable_thinking`, ensuring the PEG parser does not extract reasoning block markers when thinking is disabled.

## Fixes
- #20182 — `enable_thinking` param cannot turn off thinking for qwen3.5
- #20196 — disabling reasoning does not work anymore on certain models
- #20409 — `enable_thinking=false` ignored across all shells
- #20861 — Qwen3.5 with Opencode assistant prefill incompatible

## Testing
- Build verified (Release, Linux, `cmake --build . -j$(nproc)`)
- Existing `test-chat` and `test-chat-peg-parser` pass
- No new server integration tests added (change is minimal and targeted)

## AI Usage Disclosure

YES — AI (OpenCode with Qwen3.6-35B) assisted with code formatting, searching related issues on GitHub, verifying CI compliance against `.github/workflows/server.yml`, and drafting this PR description. All code changes, logic design, and review were performed by the contributor.

### DIFF
--- a/common/chat-auto-parser-generator.cpp
+++ b/common/chat-auto-parser-generator.cpp
@@ -93,7 +93,7 @@ common_peg_arena autoparser::build_parser(const generation_params & inputs) cons
     }
     return build_chat_peg_parser([&](common_chat_peg_builder & p) {
         parser_build_context ctx(p, inputs);
-        bool                 extract_reasoning = inputs.reasoning_format != COMMON_REASONING_FORMAT_NONE;
+        bool                 extract_reasoning = inputs.reasoning_format != COMMON_REASONING_FORMAT_NONE && inputs.enable_thinking;
 
         ctx.extracting_reasoning = extract_reasoning && reasoning.mode != reasoning_mode::NONE;
         ctx.content              = &content;
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -2210,7 +2210,9 @@ static common_chat_params common_chat_templates_apply_jinja(const struct common_
         struct autoparser::autoparser autoparser;
         autoparser.analyze_template(tmpl);
         auto auto_params = autoparser::peg_generator::generate_parser(tmpl, params, autoparser);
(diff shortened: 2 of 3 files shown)


ghost:
https://github.com/ggml-org/llama.cpp/pull/22323
https://github.com/ggml-org/llama.cpp/pull/22162

isaac-mcfadyen:
As far as I'm aware, `extra_body` on the OpenAI clients literally adds the fields to the body. There is not a dedicated field called `extra_body` actually sent in the request. See:

https://github.com/openai/openai-python/blob/e507a4ebeea4c3f93cd48986014a3e2ca79230c2/src/openai/_base_client.py#L2007-L2045

https://github.com/openai/openai-python/blob/e507a4ebeea4c3f93cd48986014a3e2ca79230c2/src/openai/_base_client.py#L502-L509

https://github.com/openai/openai-python/blob/e507a4ebeea4c3f93cd48986014a3e2ca79230c2/src/openai/_base_client.py#L2183-L2192

Also, the general way to "disable" thinking with reasoning models is to add empty `<think></think>` tags. I suspect that your change in `chat.cpp` will not add these tags because the model is no longer marked as supporting thinking (and will severely degrade performance as a result).

PR #22338, opened 2026-04-24
fix Blackwell sharedMemPerBlockOptin overflow causing SOFT_MAX error

patrickmamaid:
TLDR / My use case:
- I dont have enough vram so i want to CPU + GPU split unsloth/Qwen3.5-122B-A10B-GGUF:UD-Q4_K_XL  (MOE model)
- this worked previously on rtx4090 but not on rtx5090 
- somehow my RTX 5090 reported 4294967297 bytes (~4 GB) of shared memory with NVIDIA Driver: 590.48.01

This PR fixes an issue when using: 

```    

./llama.cpp/build/bin/llama-server \
        -hf "unsloth/Qwen3.6-27B-GGUF:UD-Q5_K_XL" \
	--mmproj-url "https://huggingface.co/unsloth/Qwen3.6-27B-GGUF/resolve/main/mmproj-F16.gguf?download=true" \
        --host 0.0.0.0 \
        --port 20000 \
        --ctx-size 200000 \
        --n-gpu-layers 999 \
        --flash-attn on \
        --temp 0.6 \
        --top-p 0.95 \
        --top-k 20 \
        --min-p 0.00 \
	--presence_penalty 0.0 \
	--repeat_penalty 1.0 \
        --cache-type-k q8_0 \
        --cache-type-v q8_0 \
        --metrics \
	--parallel 1 \
	--timeout 3600 \
	--cont-batching \
        --jinja
```

built with main branch (at commit 0adede866ddb2e31992b3792eaea31d18ed89acf):
```
rm -rf llama.cpp/build

cmake llama.cpp -B llama.cpp/build \
    -DBUILD_SHARED_LIBS=OFF \
    -DGGML_CUDA=ON \
    -DGGML_RPC=ON \
    -DGGML_NATIVE=ON \
    -DCMAKE_CUDA_ARCHITECTURES="120" \
    -DCMAKE_BUILD_TYPE=Release

cmake --build llama.cpp/build --config Release -j --clean-first --target llama-cli llama-mtmd-cli llama-server llama-gguf-split rpc-server
cp llama.cpp/build/bin/llama-* llama.cpp/build/bin/rpc-server llama.cpp

```

My System:
```
OS: Linux Mint 22.1 (Xia) Linux mint 6.14.0-37-generic #37~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 20 10:25:38 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

Ryzen 9 9950x3d
RTX5090
DDR5 96GB @ 6000MT (2 dimms)
1tb NVME
```

Driver specs:
```
  - GPU: NVIDIA GeForce RTX 5090 (compute capability 12.0)
  - NVIDIA Driver: 590.48.01
  - CUDA Runtime: 13.1
  - CUDA Toolkit (nvcc): 13.1 (/usr/local/cuda-13.1)
  - Kernel: 6.14.0-37-generic
```

It crash on an initial "hello" prompt from the web-ui:
Screenshot from 2026-04-24 12-25-55
Screenshot from 2026-04-24 14-58-56

--- 

What happens:

1. ggml-cuda.cu reads cudaDeviceProp.sharedMemPerBlockOptin -->> stores it in smpbo (a size_t, 64-bit)
2. On RTX 5090 (Blackwell, compute 12.0), this returns 4294967297 (0x100000001) : what?!...
3. CUDA_SET_SHARED_MEMORY_LIMIT passes smpbo to cudaFuncSetAttribute(..., int)  a 32-bit parameter
4. 0x100000001 truncated to int = 1
5. GPU now thinks max dynamic shared memory is 1 byte
6. Softmax kernel requests 1152 bytes -->> cudaErrorInvalidValue -->> crash

Fix appears to resolve my issue on my setup

### DIFF
--- a/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -282,6 +282,16 @@ static ggml_cuda_device_info ggml_cuda_init() {
                       (size_t)(prop.totalGlobalMem / (1024 * 1024)));
 #else
         info.devices[id].smpbo = prop.sharedMemPerBlockOptin;
+
+        // Workaround: on some Blackwell (compute capability 12.0) GPUs,
+        // sharedMemPerBlockOptin reports a bogus value (0x100000001) which
+        // overflows to 1 when cast to int in cudaFuncSetAttribute(), causing
+        // kernel launches to fail with cudaErrorInvalidValue.
+        // Fall back to sharedMemPerBlock if the value is unreasonably large.
+        if (info.devices[id].smpbo > 1024*1024) {
+            info.devices[id].smpbo = prop.sharedMemPerBlock;
+        }
+
         info.devices[id].cc = 100*prop.major + 10*prop.minor;
         GGML_LOG_INFO("  Device %d: %s, compute capability %d.%d, VMM: %s, VRAM: %zu MiB\n",
                       id, prop.name, prop.major, prop.minor, device_vmm ? "yes" : "no",


am17an:
I don't think this happens, if it happens it's a CUDA bug not llama.cpp bug and we shouldn't be fixing it here

JohannesGaessler:
I 100% agree with Aman here, if an incorrect value is being reported that is a CUDA bug, not a llama.cpp/ggml bug.

patrickmamaid:
thanks for the feedback and i also agree, ill flag it upstream on the CUDA side instead

patrickmamaid:
Linking to upstream discussion: https://forums.developer.nvidia.com/t/cudadeviceprop-sharedmemperblockoptin-returns-incorrect-value-0x100000001-on-rtx-5090-sm120/368102 seems nobody has reported this specific bug yet

github-pmunoz:
Did you test to read the props directly with CUDA in order to discard a driver issue? I had the bad SMPBO reading issue on build 10454, commit 4df29be4f, with a 5060 Ti and the cuda-13.1 driver is fine, the problem was a broken llama.cpp build with mixed nvcc toolkits.

You can probe directly with cuda as follows, compiling with cuda-13.1 for your blackwell chip
```
//props.cu
#include <cstdio>
  int main() {
      cudaDeviceProp p; cudaGetDeviceProperties(&p, 0);
      printf("%s sm_%d%d smpb=%zu smpbo=%zu maxGrid=(%d,%d,%d)\n",
          p.name, p.major, p.minor, p.sharedMemPerBlock, p.sharedMemPerBlockOptin,
          p.maxGridSize[0], p.maxGridSize[1], p.maxGridSize[2]);
  }
$/usr/local/cuda-13.1/bin/nvcc props.cu -o props && ./props
NVIDIA GeForce RTX 5060 Ti sm_120 smpb=49152 smpbo=101376 maxGrid=(2147483647,65535,65535)
```

This just proves the driver is fine. I tried pinning this smpbo value directly in `ggml-cuda.cu` and that works but the correct solution is fixing the build.

For inspecting your build:

1. Check CMakesCache.txt for the mixed presence of local and system nvcc. Look for FIND_PACKAGE to see if CMake resolved the toolkit wrong.
```CMakeCache.txt
FIND_PACKAGE_MESSAGE_DETAILS_CUDAToolkit:INTERNAL=[/usr/local/cuda-13.1/targets/x86_64-linux/include;**/usr/local/cuda-13.1**/targets/x86_64-linux/include/cccl][**/usr/lib/x86_64-linux-gnu/libcudart.so**][/usr/local/cuda-13.1/bin][v13.1.115()]
```
in this case `/usr/lib/x86_64-linux-gnu/libcudart.so` is 12.4. The mismatched versions explains the bad prop reading as a cstruct being written and read with different encoding lengths.

2. Try adding `	-DCUDAToolkit_ROOT=/usr/local/cuda-13.1` to your CMake config, that fixed it for me:
```
cmake -B build \
	-DCMAKE_BUILD_TYPE=Release \
	-DCMAKE_CUDA_COMPILER=/usr/local/cuda-13.1/bin/nvcc \
	-DGGML_CUDA=ON \
	-DCMAKE_CUDA_ARCHITECTURES=120 \
	-DCUDAToolkit_ROOT=/usr/local/cuda-13.1 
```

PR #9125, opened 2024-08-21
llama : add ability to load model from memory buffer

ngxson:
TODO: add some tests

Currently, loading a model depends on file system. That means the model must firstly be saved onto disk, then being loaded using `llama_load_model_from_file`

However, when compiled to [web assembly](https://github.com/ngxson/wllama), the virtual fs implementation is very inefficient and does not support files bigger than 2GB. Allow loading file directly from a buffer will be bypass the need of fs.

This PR introduce 2 new APIs:
- `gguf_init_from_buffer` ==> this was left as a TODO, so I implemented it
- `llama_load_model_from_buffers` ==> support loading multiple model shards (or model splits)

---

- [x] I have read the [contributing guidelines](https://github.com/ggerganov/llama.cpp/blob/master/CONTRIBUTING.md)
- Self-reported review complexity:
  - [ ] Low
  - [x] Medium
  - [ ] High

PR #22341, opened 2026-04-24
ggml: `gguf_init_from_callback` and `gguf_init_from_buffer`

giladgd:
## Overview

Implements `gguf_init_from_buffer` to allow loading GGUF data from a buffer.
It makes it easier to load temporary GGUF content directly from memory instead of saving it to a file and having to remove that file later.

Also adds `gguf_init_from_callback` to allow loading GGUF data from a custom callback reader function.

PR #9125 seems to be based on a much older revision of ggml so I figured it'd be easier for me to just reimplement it myself.
This PR seems to work great from my manual tests as well as the tests I've added with Codex's help.

## Additional information

Related discussion: https://github.com/ggml-org/llama.cpp/discussions/10990
Old PR: https://github.com/ggml-org/llama.cpp/pull/9125

# Requirements

- AI usage disclosure: YES, Codex helped me find the relevant places to make changes, reviewed my code, and generated tests. I manually tested everything and am proficient in all changes of this PR.


giladgd:
@ggml-bot I'm not a new contributor

JohannesGaessler:
Recently we added `gguf_init_from_file_ptr` which you should be able to use in conjunction with `tmpfile` or `fmemopen` to create temporary (in-memory) files.

giladgd:
Using `fmemopen` is only practical on macOS and Linux but not on Windows.
I couldn't find any stable way to use `gguf_init_from_file_ptr` to load a buffer from memory that would work on Windows, do you have any pointers to give me in that regard?

tests/test-model-load-buffer.cpp
JohannesGaessler, line 1: Remove this file, the test cases in `test-gguf.cpp` should suffice.

src/llama-model-loader.cpp
JohannesGaessler, line 1: Why are you changing the model loader?

ggml/src/gguf.cpp
```diff
@@ -353,15 +361,58 @@ struct gguf_reader {
...
+        offset += size;
+        return size;
+    }
+
+    FILE * file = nullptr;
+    const uint8_t * data = nullptr;
+
+    mutable size_t offset = 0;
     mutable uint64_t nbytes_remain;
+    uint64_t total_size = 0;
```
JohannesGaessler: This is overspecified. A `gguf_reader` should be tracking only 2/3 of these values and calculate the third one as needed from the other 2.

src/llama-model-loader.cpp
giladgd: When loading a model using `llama_model_init_from_user` from a gguf metadata `gguf_context` created with `gguf_init_from_buffer` then `llama_model_loader::files` is empty and so the result of the `memory_breakdown` of the model when loading with `no_alloc=true` is off and doesn't align with the results we would get when loading the same model from a file.
The changes I made here fix that

tests/test-model-load-buffer.cpp
giladgd: I added this file to validate the fix I made in `llama-model-loader.cpp` and ensure it doesn't break with future changes.
The crux of this test file is the comparison of `->memory_breakdown()` from the model loaded from a file and the one loaded from a buffer.

If you remove the changes I made in `llama-model-loader.cpp` and run this test it'll fail

ggml/src/gguf.cpp
giladgd: Done

ggml/src/gguf.cpp
```diff
@@ -353,14 +357,64 @@ struct gguf_reader {
...
+    }
+
 private:
-    FILE * file;
+    size_t read_raw(void * dst, size_t size) const {
+        if (file != nullptr) {
+            return fread(dst, 1, size, file);
+        } else if (data == nullptr || size > nbytes_remain || data_offset + size < data_offset) {
+            return 0;
+        }
```
JohannesGaessler:
```suggestion
        }
        if (data == nullptr || size > nbytes_remain) {
            return 0;
        }
```

The last check makes no sense to me. Is it a bug?

src/llama-model-loader.cpp
JohannesGaessler: A bugfix like that should be its own PR, remove the model loader changes from this one.

ngxson:
API-design-wise, do you think it would be more future-proof to have an API like `gguf_init_from_callback` where the reader is an callback function or an interface?

It would be useful for example on wasm/webgpu, where we want to read the file with a custom JS interface that will get binded to c++. Something like:

```cpp
size_t my_reader(void * userdata, void * output, size_t offset, size_t len) {
  memcpy(output, userdate + offset, len);
  // in wasm, instead of memcpy, this will call a binded API that reads data from javascript context
}

// usage (allow reading at most 16MB at once to avoid OOM)
gguf_init_from_callback(my_reader, my_data, 16 * 1024 * 1024);
```

Also, one of the pain point is that GGUF reader also validates the GGUF file size by doing a `seek()`, but this is not desirable on wasm implementation, since we don't want to store the whole GGUF file inside RAM. So probably a FILE-like interface would make more sense.

ggml/src/gguf.cpp
giladgd: I meant to check for an overflow there since both `data_offset` and `size` are unsigned

src/llama-model-loader.cpp
giladgd: Done. Will open a new PR once this one gets merged.

Update: opened https://github.com/ggml-org/llama.cpp/pull/22566

tests/test-model-load-buffer.cpp
giladgd: I removed this file and will introduce it back in the next PR with the fixes to `llama-model-loader.cpp`

giladgd:
@ngxson I contemplated implementing such an API instead of `gguf_init_from_buffer`, but I don't have a case where I need it at the moment so I didn't think I'll be able to nail down all the details (and optimizations) for cases that do need more dynamic reading/streaming.
I think the API you outlined makes sense, I'll be happy to implement it if you like.

In any case, I think we should be safe adding `gguf_init_from_buffer` since its underlying implementation could be swapped to use dynamic reading in the future without introducing a breaking change.

JohannesGaessler:
I would also be fine with creating a GGUF context from a callback.

JohannesGaessler:
>Also, one of the pain point is that GGUF reader also validates the GGUF file size by doing a seek()

This should only be needed for determining `nbytes_remain` and that in turn is only tracked to harden the GGUF loader against vulnerabilities which oftentimes revolve around allocaring an amount of memory that is much larger than the GGUF file. In terms of API functionality we could ask the user to provide a maximum expected file size.

giladgd:
I've added `gguf_init_from_callback`, changed `gguf_reader` to work with a callback reader, and changed the rest of the methods to work in that pattern.
I've kept `gguf_init_from_buffer` since I think it's also useful to have it by itself, but can remove it if you don't want to have it.

ggml/include/gguf.h
```diff
@@ -76,10 +76,18 @@ extern "C" {
         struct ggml_context ** ctx;
     };
 
+    // reads up to `len` bytes at `offset` into `output` and returns the number of bytes read.
+    // may be called with `len == 0` to seek/synchronize to `offset` without reading.
+    // when `len == 0` returns 0 on success and non-zero on failure
+    typedef size_t (*gguf_reader_callback_t)(void * userdata, uint8_t * output, size_t offset, size_t len);
```
JohannesGaessler:
```suggestion
    // callback to simulate or wrap a FILE pointer:
    //   - by default, read up to `len` bytes at `offset` into `output` and return the number of bytes read
    //   - if called with `len == 0`, seek/synchronize to `offset` without reading, return 0 on success, non-zero for failure
    typedef size_t (*gguf_reader_callback_t)(void * userdata, void * output, size_t offset, size_t len);
```

This should I think be the specification for the callback.

ggml/include/gguf.h
```diff
@@ -76,10 +76,18 @@ extern "C" {
...
+
     GGML_API struct gguf_context * gguf_init_empty(void);
     GGML_API struct gguf_context * gguf_init_from_file_ptr(FILE * file, struct gguf_init_params params);
     GGML_API struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_params params);
-    //GGML_API struct gguf_context * gguf_init_from_buffer(..);
+    GGML_API struct gguf_context * gguf_init_from_buffer(const void * data, size_t size, struct gguf_init_params params);
+
+    // when `total_size == 0` then end of file offset will be determined when the returned read size is smaller than the requested `len`
+    GGML_API struct gguf_context * gguf_init_from_callback(gguf_reader_callback_t callback, void * userdata, size_t max_chunk_read, size_t total_size, struct gguf_init_params params);
```
JohannesGaessler:
```suggestion
    GGML_API struct gguf_context * gguf_init_from_callback(gguf_reader_callback_t callback, void * userdata, size_t max_chunk_read, size_t max_expected_size, struct gguf_init_params params);
```

The total size should **not** be determined from file reads. That circumvents the hardening in `gguf.cpp` against overflow-based vulnerabilities. The user should always explicitly provide an upper bound for what they expect the maximum file size to be, even if that is 10x higher than the actual file size it should still be good enough.

JohannesGaessler (commented):
@ngxson your feedback on the specification would be appreciated.

ngxson (approved):
I only reviewed the public API (gguf.h) and it looks good. Hopefully libllama also expose the same API soon. Not quite sure if I will have the time to add that but feel free to do if you like @giladgd 

P/s: mmap() may need to be disabled for such API, but I think it will be an acceptable trade-off

ggml/include/gguf.h
giladgd: Done

ggml/include/gguf.h
giladgd: Done

giladgd:
I've changed `max_expected_size` and the offsets to `uint64_t` since it seems to be safer against overflows of `size_t`, since _some_ places already cast offsets to `uint64_t` anyway

giladgd:
@ngxson changing the model loader to support reading from a callback seems much more complicated, especially around retaining mmap support for the existing flows, so I'll skip that for now

giladgd:
@JohannesGaessler let me know if there's anything else I should change in this PR or whether it's ready for merge.
I've opened https://github.com/ggml-org/llama.cpp/pull/22566 with the model loader fixes so it could be reviewed after this PR is merged.

giladgd:
Gentle ping. It would be a bummer to see this PR get abandoned since I’d like to use the new buffer reader soon

ngxson:
Re-pinging @JohannesGaessler if you can review this once again (I kinda need this in near future)

I'm also requesting AI review just in case (mostly for spotting vulnerabilities)

ggml/src/gguf.cpp
```diff
@@ -344,24 +350,75 @@ struct gguf_reader {
...
+
+        uint8_t * data = static_cast<uint8_t *>(dst);
+        size_t total_nread = 0;
+        bool reached_eof = false;
 
-    mutable uint64_t nbytes_remain;
+        while (total_nread < size) {
+            const size_t chunk = std::min(max_chunk_read, size - total_nread);
```
JohannesGaessler:
```suggestion
            const size_t chunk_size = std::min(max_chunk_read, size - total_nread);
```

ggml/src/gguf.cpp
```diff
@@ -228,9 +228,17 @@ struct gguf_context {
...
+            gguf_reader_callback_t callback,
+            void * userdata,
+            size_t max_chunk_read,
+            uint64_t data_offset = 0,
+            uint64_t nbytes_remain = 0)
+        : callback(callback),
+          userdata(userdata),
+          max_chunk_read(max_chunk_read),
+          data_offset(data_offset),
+          nbytes_remain(nbytes_remain) {
```
JohannesGaessler, line 241:
```suggestion
          nbytes_remain(nbytes_remain) {
        GGML_ASSERT(max_chunk_read > 0);
```

ggml/src/gguf.cpp
```diff
@@ -844,6 +896,93 @@ struct gguf_context * gguf_init_from_file_ptr(FILE * file, struct gguf_init_para
...
+    const struct gguf_reader gr(callback, userdata, max_chunk_read, 0, max_expected_size);
+    return gguf_init_from_reader(gr, params);
+}
+
+struct gguf_file_reader {
+    FILE * file;
+    uint64_t offset;
+};
+
+static size_t gguf_file_reader_callback(void * userdata, void * output, uint64_t offset, size_t len) {
```
JohannesGaessler, line 910:
```suggestion
static size_t gguf_file_reader_callback(void * userdata, void * output, uint64_t offset, size_t len) {
    GGML_ASSERT(len > 0);
```

This callback is not implementing the exact behavior described in `gguf.h`. To my knowledge the current code does not need the seeking for `len == 0`, in that case there should be an assert that it is actually never used.

ggml/src/gguf.cpp
```diff
@@ -844,6 +896,93 @@ struct gguf_context * gguf_init_from_file_ptr(FILE * file, struct gguf_init_para
...
+            return len == 0 ? 1 : 0;
+        }
+
+        reader.offset = offset;
+    }
+
+    if (len == 0) {
+        return 0;
+    }
+
```
JohannesGaessler:
```suggestion
```

ggml/src/gguf.cpp
```diff
@@ -844,6 +896,93 @@ struct gguf_context * gguf_init_from_file_ptr(FILE * file, struct gguf_init_para
...
+static size_t gguf_buffer_reader_callback(void * userdata, void * output, uint64_t offset, size_t len) {
+    const gguf_buffer_reader & reader = *static_cast<gguf_buffer_reader *>(userdata);
+
+    if (offset > reader.size) {
+        return len == 0 ? 1 : 0;
+    }
+
+    if (len == 0 || offset + len > reader.size) {
+        return 0;
+    }
```
JohannesGaessler, line 958:
```suggestion
static size_t gguf_buffer_reader_callback(void * userdata, void * output, uint64_t offset, size_t len) {
    GGML_ASSERT(len == 0);
    const gguf_buffer_reader & reader = *static_cast<gguf_buffer_reader *>(userdata);

    if (offset + len > reader.size) {
        return 0;
    }
```

If the seeking behavior is not actually used, I think the code should be looking like this.

tests/test-gguf.cpp
```diff
@@ -162,6 +162,44 @@ static void helper_write(FILE * file, const void * data, const size_t nbytes) {
...
+static size_t read_buffer_callback(void * userdata, void * output, uint64_t offset, size_t len) {
+    const callback_reader_data & reader = *static_cast<callback_reader_data *>(userdata);
+
+    if (offset > reader.size) {
+        return len == 0 ? 1 : 0;
+    }
+
+    if (len == 0 || offset + len > reader.size) {
+        return 0;
+    }
```
JohannesGaessler:
```suggestion
    if (len == 0) {
        return offset <= reader.size ? 0 : 1;
    }
    if (offset > reader.size || offset + len > reader.size) {
        return 0;
    }
```

I think this logic is more clear.

tests/test-gguf.cpp
```diff
@@ -1095,10 +1133,29 @@ static bool same_tensor_data(const struct ggml_context * orig, const struct ggml
     return ok;
 }
 
-static std::pair<int, int> test_roundtrip(ggml_backend_dev_t dev, const unsigned int seed, const bool only_meta) {
+enum roundtrip_read_mode {
+    ROUNDTRIP_READ_FILE,
+    ROUNDTRIP_READ_BUFFER,
+    ROUNDTRIP_READ_CALLBACK,
```
JohannesGaessler:
```suggestion
    ROUNDTRIP_READ_MODE_FILE,
    ROUNDTRIP_READ_MODE_BUFFER,
    ROUNDTRIP_READ_MODE_CALLBACK,
```

As per the contribution guidelines for enum names.

tests/test-gguf.cpp
```diff
@@ -1095,10 +1133,29 @@ static bool same_tensor_data(const struct ggml_context * orig, const struct ggml
...
+        case ROUNDTRIP_READ_BUFFER:   return "buffer";
+        case ROUNDTRIP_READ_CALLBACK: return "callback";
+    }
+
+    GGML_ABORT("fatal error");
+}
+
+static std::pair<int, int> test_roundtrip(
+        ggml_backend_dev_t dev, const unsigned int seed, const bool only_meta,
+        const roundtrip_read_mode read_mode = ROUNDTRIP_READ_FILE) {
```
JohannesGaessler:
```suggestion
        const roundtrip_read_mode read_mode) {
```

Preferably require an explicit specification.

JohannesGaessler (commented):
Sorry, I forgot about this PR.

ggml/include/gguf.h
```diff
@@ -76,10 +76,16 @@ extern "C" {
         struct ggml_context ** ctx;
     };
 
+    // callback to simulate or wrap a FILE pointer:
+    //   - by default, read up to `len` bytes at `offset` into `output` and return the number of bytes read
+    //   - if called with `len == 0`, seek/synchronize to `offset` without reading, return 0 on success, non-zero for failure
```
JohannesGaessler:
```suggestion
```

Looking at the code again, since the `gguf_reader` is already doing internal bookkeeping for the `data_offset`, I think this should be simplified to remove the seeking behavior for `len == 0` from the specification and instead make `len > 0` a hard requirement. `gguf_reader::seek` should simply modify `gguf_reader::data_offset` and `gguf_file_reader_callback` should simply call `gguf_fseek` as needed (as is already implemented).

giladgd: Done

ggml/src/gguf.cpp
giladgd: Done

ggml/src/gguf.cpp
giladgd: Done

ggml/src/gguf.cpp
giladgd: I removed the seek behavior from the callback and added `GGML_ASSERT(len > 0)`

ggml/src/gguf.cpp
giladgd: Done

ggml/src/gguf.cpp
giladgd: Done

tests/test-gguf.cpp
giladgd: Done. I used `len > reader.size - offset` instead of `offset + len > reader.size` to be safer against overflows (we already verify ensure that `offset <= reader.size` first so `reader.size - offset` is safe)

tests/test-gguf.cpp
giladgd: Done

tests/test-gguf.cpp
giladgd: Done

ggml/include/gguf.h
```diff
@@ -76,10 +76,14 @@ extern "C" {
...
 
+    // callback to simulate or wrap a FILE pointer - read up to `len` bytes at `offset` into `output` and return the number of bytes read
+    typedef size_t (*gguf_reader_callback_t)(void * userdata, void * output, uint64_t offset, size_t len);
+
     GGML_API struct gguf_context * gguf_init_empty(void);
     GGML_API struct gguf_context * gguf_init_from_file_ptr(FILE * file, struct gguf_init_params params);
     GGML_API struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_params params);
-    //GGML_API struct gguf_context * gguf_init_from_buffer(..);
+    GGML_API struct gguf_context * gguf_init_from_buffer(const void * data, size_t size, struct gguf_init_params params);
+    GGML_API struct gguf_context * gguf_init_from_callback(gguf_reader_callback_t callback, void * userdata, size_t max_chunk_read, uint64_t max_expected_size, struct gguf_init_params params);
```
JohannesGaessler, line 88:
```suggestion
    // max_chunk_read is the maximum number of bytes that the GGUF code will read at once from the callback, a value of 0 means no limit
    GGML_API struct gguf_context * gguf_init_from_callback(gguf_reader_callback_t callback, void * userdata, size_t max_chunk_read, uint64_t max_expected_size, struct gguf_init_params params);
```

I would say we should maybe make this small change to the spec where we internally set the value to `SIZE_MAX` if the user sets 0 to make the code a bit more ergonomic.

JohannesGaessler (approved):
Sorry for the delays and long list of requested changes, the GGUF code is critical both in terms of functionality and security so it requires a lot of scrutiny. And thank you for sticking with me. There is just one last minor change the I think we should make and after that this PR is I think good to merge.

ggml/include/gguf.h
giladgd: Done.
I'd love if you could review #22566 next after this PR is merged

ggml/src/gguf.cpp
```diff
@@ -700,14 +749,14 @@ struct gguf_context * gguf_init_from_file_ptr(FILE * file, struct gguf_init_para
     GGML_ASSERT(int64_t(ctx->info.size()) == n_tensors);
 
     // we require the data section to be aligned, so take into account any padding
-    if (gguf_fseek(file, GGML_PAD(gguf_ftell(file), ctx->alignment), SEEK_SET) != 0) {
+    if (!gr.seek(GGML_PAD(gr.tell(), ctx->alignment))) {
```
JohannesGaessler:
```suggestion
    if (n_tensors > 0 && !gr.seek(GGML_PAD(gr.tell(), ctx->alignment))) {
```

The CI is failing on the edge case of a GGUF file with 0 tensors in which case the file size is not padded. I would say to just handle that edge case like this.

ggml/include/gguf.h
JohannesGaessler:
```suggestion
    GGML_API size_t   gguf_get_data_offset(const struct gguf_context * ctx); // padded to gguf_get_alignment if and only if the gguf_context contains at least one tensor
```

ggml/src/gguf.cpp
giladgd: Done

ggml/include/gguf.h
giladgd: Done

JohannesGaessler (approved):

PR #22343, opened 2026-04-25
mtmd: qwen3-asr wrong output

cora4:
## Overview
qwen3-asr output is wrong with llama.cpp


ngxson:
closing this as you did not disclosed AI usage and this is not the intended way to fix

cora4:
I did not use AI and I disclosed that I did not.
 > this is not the intended way to fix

You can test the audio yourself, the text generated is completely wrong in some cases before the fix.
What is your intended way to fix the wrong text generated?

ngxson:
The code comments are clearly written by AI

ngxson:
Your fix cannot be accepted as-is because it introduces a hack into the existing preprocessor, it is not clean and potentially break other models.

The cgraph implementation is redundant because (1) you can use the 4th dim for batching and (2) you replaced the old build_vit with essentially the over-complicated version equivalent of it.

But overall, I do not wish to proceed with contributions where AI is used but not properly disclosures.

ngxson:
not sure why you deleted your comment earlier @cora4 ? but we can proceed with this PR if you can disclose AI usage (using AI to write comment **still counts**)

PR #22347, opened 2026-04-25
hexagon: hmx flash attention

njsyw1997:
## Overview

This PR implemented hmx based flash attetion for Hexagon backend.

### FP16 Softamax
Profiling shows that the main bottleneck is the `exp` computation (about 40% of total FA runtime). I experimented with a LUT-based, lossless optimization, but it appears that `vgather` cannot be effectively parallelized—multithreaded `vgather` provided no measurable speedup. ~~I’m not sure whether this is due to an issue in my implementation or an inherent hardware limitation.~~ As mentioned [here](https://github.com/ggml-org/llama.cpp/pull/22347#issuecomment-4357254968), `vgather` is aborted.

As an alternative, I implemented an FP16 version of exp to improve performance. This does introduce some numerical loss, so it is disabled by default. Enabling it via `GGML_HEXAGON_FA_EXP2_HF=ON` yields an additional ~10% performance gain.

### Existing Bug
~~This version appears to have a bug that I haven’t been able to identify. It behaves like a non-deterministic compilation issue: a **good** build consistently produces correct results, while a **bad** build yields completely incorrect PPL—even though both are clean compiled from the exact same code. Re-running the PPL tests does not change the outcome, which suggests the issue lies in the compilation process rather than a data race.~~

~~I attempted to use agent to analyze and address the issue in commit `85dd885`. The issue still persists, although its frequency has decreased. This code is not reliable since I am not very familar with the hexagon compiler and cannot really review it. I’ve kept that commit for reference.~~

~~I will keep this PR in draft until the problem is fully resolved.~~

Should be fixed by e0ee557. 

## Requirements

- AI usage disclosure: Yes. As described above, and also for filtering and analyzing logs.

### DIFF
--- a/ggml/src/ggml-hexagon/htp/hmx-flash-attn-ops.c
+++ b/ggml/src/ggml-hexagon/htp/hmx-flash-attn-ops.c
@@ -602,6 +602,7 @@ typedef struct {
     uint32_t                kv_start;
     uint32_t                q_start;
     uint32_t                ib3;
+    bool                    has_alibi;              // true when max_bias != 0 (need slope * mask + add)
 
     // ALiBi per-head slopes (indexed by GQA-merged row: slope[r] for r in [0, n_rows_g))
     // slope[r] = 1.0 when max_bias == 0 (no ALiBi)
@@ -690,13 +691,27 @@ static void fa_softmax_thread(unsigned int n, unsigned int i, void * data) {
                 }
             }
 
-            // Apply mask (with ALiBi slope scaling) & compute rowmax(S)
-            // CPU reference: S[i] += slope * mask[i]
-            // slope == 1.0 when max_bias == 0 (no ALiBi)
-            const HVX_Vector v_slope0 = hvx_vec_splat_f16((__fp16) args->slopes[r + 0]);
-            const HVX_Vector v_slope1 = (r + 1 < (int) n_rows_g) ?
-                                            hvx_vec_splat_f16((__fp16) args->slopes[r + 1]) :
-                                            Q6_V_vzero();
+            // Apply mask & compute rowmax(S)
+            //
+            // Optimizations over baseline:
+            //   A. No-ALiBi fast path: when max_bias==0 (slope≡1.0), mask values are
+            //      0 (unmasked) or -inf (masked).  S += 1.0*mask is a no-op for
+            //      unmasked positions; masked ones are overwritten by vmux.  So we
+            //      skip the mul+add entirely and just do vmux — same as htp-ops-lib.
+            //   B. GQA mask row dedup: G consecutive Q rows share one mask row
+            //      (qi = r / G).  Reuse mask vector when qi is unchanged between
+            //      row0 and row1 (saves ~75% of VTCM loads for G=4).
+
+            // ALiBi slopes — only needed when has_alibi (scheme A)
+            HVX_Vector v_slope0, v_slope1;
+            if (args->has_alibi) {
+                v_slope0 = hvx_vec_splat_f16((__fp16) args->slopes[r + 0]);
+                v_slope1 = (r + 1 < (int) n_rows_g) ?
+                               hvx_vec_splat_f16((__fp16) args->slopes[r + 1]) :
+                               Q6_V_vzero();
+            }
+
+            const HVX_Vector   v_threshold = Q6_Vh_vsplat_R(0xcc00); // fp16 -16.0 (hoisted outside for-c)
 
             HVX_Vector v_s_rowmax0 = v_neg_inf;
             HVX_Vector v_s_rowmax1 = v_neg_inf;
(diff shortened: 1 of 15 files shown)


njsyw1997:
This is an example build [example_build.zip](https://drive.google.com/file/d/1_ETgWlmKMyU0fNVuhr5oogEwxs1hgQYQ/view?usp=drive_link). `pkg-snapdragon_bad` will produce wrong results, and `pkg-snapdragon_good` will produce correct results on OnePlus 15(Snapdragon 8 Elite Gen 5, hexagon v81)

njsyw1997:
@max-krasnyansky Can you have a look?

ggml/src/ggml-hexagon/htp/hmx-flash-attn-ops.c
```diff
@@ -0,0 +1,1861 @@
...
+}
+
+// ============================================================================
+// FP16 exp2 polynomial (ported from htp-ops-lib/include/dsp/hvx_math.h)
+// ============================================================================
+// 5th-order Horner polynomial for exp2(x) in qf16/hf16 domain.  Input must be
+// ≤ 0 (safe softmax invariant — overflow handling omitted).  ~18 ALU ops per
+// 64 fp16 lanes, fully parallel across HVX threads (no scatter/gather engine).
+// Replaces the F32 round-trip (qf16→f32→exp→f32→f16, ~44 ops for 2×32 lanes).
+static inline HVX_Vector hvx_exp2_hf(HVX_Vector x_v) {
```
chraac, line 87: QQ: shall we gate the `InF` and `NaN` for this one as well?

max-krasnyansky:
> This is an example build [example_build.zip](https://drive.google.com/file/d/1_ETgWlmKMyU0fNVuhr5oogEwxs1hgQYQ/view?usp=drive_link). `pkg-snapdragon_bad` will produce wrong results, and `pkg-snapdragon_good` will produce correct results on OnePlus 15(Snapdragon 8 Elite Gen 5, hexagon v81)

Sorry for the delay.
It looks like this might be caused by `-O3` in our `htp/cmake-toolchain` settings. 
I've seen this a few times recently as well (i.e I'd add some benign change and all of the sudden get a broken build) and after some recent experiments it seems like the `-O2` doesn't have that issue.
We're expecting HexSDK 6.6. release with the updated compiler soon (any day now). The plan is to respin the snapdragon-toolchain docker with that and test things out. If we still see these issues we'll switch to '-O2'.

njsyw1997:
@max-krasnyansky 
Thanks for updates!
How about `vgather`? Is there any hardware limitation for multithread `vgather`?

ggml/src/ggml-hexagon/htp/hmx-flash-attn-ops.c
njsyw1997: Maybe not. HVX side didn't do this. 
`hvx_fill_uh(factx.vtcm_m_vec, 0xfbff, col_vec_bytes);` Here mask is using `0xfbff` so input has no inf/nan.  
When return there's a guard to set small number to 0.

max-krasnyansky:
> @max-krasnyansky Thanks for updates! How about `vgather`? Is there any hardware limitation for multithread `vgather`?

Not that I'm aware of. Should be just like any other hvx instruction. 
I'd make sure that you're not gathering into the same VTCM location from multiple threads but other than that it's just read/write from VTCM.
https://docs.qualcomm.com/doc/80-N2040-61/topic/memory.html#scatter-and-gather

max-krasnyansky:
I've been playing with this today and the results are looking pretty good.
I ended up just updating that cmake toolchain for now to get more stable builds and rebaseing with the latest master.
Will iterate a bit more (got some minor tweaks in mind, etc) and test it some more on the longer context.
Here is my branch with toolchain updates and rebase
https://github.com/qualcomm/llama.cpp/tree/hexagon-hmx-fa

njsyw1997:
What's our plan? Rebase now and merge or wait for more updates?

max-krasnyansky:
> What's our plan? Rebase now and merge or wait for more updates?

So far my testing is going well (with that branch I shared). 
I'll try to finish my tweaks I mentioned today/early-tomorrow and we can merge.

See if your issues with the funky builds are resolved with the toolchain updates.
Also was the `vgather` question just a general concern or were you seeing any issues?

njsyw1997:
I just found the problem. Shamed that I never carefully read these manuals before. :( 
The [doc](https://docs.qualcomm.com/doc/80-N2040-61/topic/memory.html#avoid-scatter-gather-stalls) here shows that
> If the in-flight scatters and gathers (including from other threads) avoid conflicts, generally a distance of 12 or more packets is sufficient.

The original [implementation](https://github.com/haozixu/htp-ops-lib/blob/main/src/dsp/ops/flash_attn.c) using `vgather` to accelerate `exp` is because it's kv-head parallelism. So naturally the `vgather` is pipelined and distributed in distance in different threads and almost **conflict-free**.

I decided to avoid that pattern because in that way the parallelism will be determined by the model architecture, which is not scalable. It also increased total work because every flash attention will be assigned smaller VTCM and compute density decreased. The price is that we need to compute softmax at the same time because of the data dependency. So, the **conflicts** mentioned above will be a problem.

`vgather` is not a good acceleration for current fa implementation. Let's just keep the current version.

max-krasnyansky:
@njsyw1997 https://github.com/qualcomm/llama.cpp/tree/hexagon-hmx-fa should be good to go.
I did a fresh rebase with all your commits. Please git reset, push and review :)

Some of the changes are purely cosmetic (as I mentioned before it helps me review things better if I tinker with it).
But mostly focused on testing, making the build more robust (some more uninlining, O2+LTO instead of O3, etc), some consistency fixes (like using hex_ / hvx_ helpers, etc) and some minor optimizations (moving slopes in to vtcm, keeping it in fp16, etc).
You can go over the commits to get full summary.
I did remove `test-backend-ops` changes, let's submit those separate (if needed).
btw I do see a few failing cases but we can follow up on those separately, it doesn't seem to affect the actual model runs.

Very nice work! Thank you for getting HMX plugged in for the FA!

I did a bunch of testing on all my devices Ventuno-Q, S24U, S25+, S26+ and X/X2-Elites with the latest key models (Qwen3.x, Gemma4, LFM2.5 ...) and everything looks good. Prompt performance has improved a lot as you mentioned above.

More updates are coming soon (starting with removing HMX-tail fallback in MM and further optimizing Weight pre-proc). Some of that will help for FA as well.

njsyw1997:
@max-krasnyansky 
Done. I also noticed some regression on very long sequence. I will keep investigating.
Thanks for your kind help!

max-krasnyansky (approved):

max-krasnyansky:
@lhez or @ggml-org/maintainers 
Can we please get the second approval.

lhez (approved):

PR #22355, opened 2026-04-25
ggml : revert to -lm linking instead of find_library

angt:
## Overview

`find_library(MATH_LIBRARY m)` was introduced recently, but it breaks CUDA compilation with GGML_STATIC. I could not find any valid use case where we would prefer `find_library` over the standard `-lm` approach.

This commit is also meant to start a discussion if there is a valid reason to keep `find_library(MATH_LIBRARY m)`, we should clarify what problem it was solving and find an alternative fix that does not break CUDA with GGML_STATIC.

## Additional information

Found with installama.sh: https://github.com/angt/installama.sh/actions/runs/24885620138/job/72864816848

## Requirements

- AI usage disclosure: NO

### DIFF
--- a/ggml/src/CMakeLists.txt
+++ b/ggml/src/CMakeLists.txt
@@ -471,7 +471,11 @@ endforeach()
 target_link_libraries(ggml-base PRIVATE Threads::Threads)
 
 if (NOT WIN32 OR NOT DEFINED ENV{ONEAPI_ROOT})
-    target_link_libraries(ggml-base PRIVATE m)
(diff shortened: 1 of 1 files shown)


angt:
ok: https://github.com/ggml-org/llama.cpp/issues/22237, im going to dig that

ggerganov:
Some of the builds are failing

angt:
> Some of the builds are failing

Fixed, we need to fallback for `NOT WIN32 AND NOT DEFINED ENV{ONEAPI_ROOT}` and not `OR` used in the current code.

PR #22359, opened 2026-04-25
Add DeepSeek V4 GGUF conversion

nisparks:
## Overview

Converts the DeepSeek V4 Flash safetensors to a valid GGUF. I have been running the GGUF locally, but that requires much more work. For now, I will submit this so others can create the GGUF files.

## Requirements

Yes
- AI usage disclosure: 
Yes, very much so, but it is tested, and I ran quite a few tests on output and it all is working.

### DIFF
--- a/convert_hf_to_gguf.py
+++ b/convert_hf_to_gguf.py
@@ -786,7 +786,8 @@ class ModelBase:
             old_dtype = data_torch.dtype
 
             # convert any unsupported data types to float32
-            if data_torch.dtype not in (torch.float16, torch.float32):
+            preserve_integer_tensor = name.endswith(".ffn.gate.tid2eid")
+            if data_torch.dtype not in (torch.float16, torch.float32) and not preserve_integer_tensor:
                 data_torch = data_torch.to(torch.float32)
 
             # use the first number-like part of the tensor name as the block id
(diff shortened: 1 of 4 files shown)


nisparks:
Just to add a note, I was using this to create FP16 GGUF, which I then used llama-quantize to reduce. The FP16 GGUF is about 555GB, and it doesn't differentiate the experts and other weights, This is a great starting point to get a valid GGUF though for further experimentation. I am running smaller quants from this and getting valid outputs.

nisparks:
Whoops, that optimization commit wasn't supposed to be pushed.

nisparks:
I'll have to push an update to this later. Just converted to the native quants locally (FP4/FP8). Going to test that out first, but was floating this PR just so others can tinker.

nisparks:
smoke test passed for the native FP4/FP8 GGUF. Will push that additional support now.

convert_hf_to_gguf.py
```diff
@@ -13701,6 +13807,7 @@ def main() -> None:
         "q8_0": gguf.LlamaFileType.MOSTLY_Q8_0,
         "tq1_0": gguf.LlamaFileType.MOSTLY_TQ1_0,
         "tq2_0": gguf.LlamaFileType.MOSTLY_TQ2_0,
+        "native": gguf.LlamaFileType.MOSTLY_F8_E4M3_MXFP4,
```
nisparks: I don't like this, will need to update

eaglexiang:
why closed?

nisparks:
I put this out there to try to contribute and get the ball rolling, but honestly don't have the time to polish these PRs to the standard this repo deserves/demands. If someone else wants to drive it, great, but just trying to do my part getting the ball rolling.

PR #22362, opened 2026-04-25
[Tensor Parallel] Fix recurrent state serialization for partial reads and writes

gaugarg-nv:
The previous code worked only for full tensor reads and writes and was hitting `GGML_ASSERT(size == ggml_nbytes(tensor)); ` assert when tested with llama-server.

## Requirements

- AI usage disclosure: For testing and code review


JohannesGaessler (approved):

JohannesGaessler:
@ggml-org/maintainers can I get a second review, please?

CISC (approved):

PR #22367, opened 2026-04-25
common/gemma4 : fix multi-turn tool call parsing with literal <|tool_call> in content

### LINKED ISSUE #22371: Misc. bug: Gemma 4 PEG parser truncates assistant content at literal <|tool_call> token (llama-server multi-turn tool calling)
### Name and Version

Reproduces on master (98dc1418e) and on every build since b8635075f (#21418, original Gemma 4 PEG parser). Tested locally at 6a8b08928 (build 8784).

### Operating systems

Mac (parser is platform-independent — bug is in `common/chat.cpp`)

- libllama (core library) — `common/chat.cpp::common_chat_params_init_gemma4`
- llama-server — chat-completions tool calling

### Command line

```
./llama-server \
  -m /path/to/gemma-4-E2B-it-Q4_K_M.gguf \
  --jinja --port 8765 -c 4096 -np 1 --temp 0 --no-warmup
```

The Gemma 4 PEG parser uses `<|tool_call>` as the delimiter that terminates the `content` rule and feeds `scan_to_toolcall` (`common/chat.cpp:1183-1184` on master):

```cpp
auto scan_to_toolcall = p.rule("scan-to-toolcall", p.until("<|tool_call>"));
auto content = p.rule("content",
    p.content(p.until_one_of({"<|channel>", "<channel|>", "<|tool_call>"})));
```

But the actual tool-call rule prefix (`common/chat.cpp:1178`) is `<|tool_call>call:`:

```cpp
auto tool_call = p.trigger_rule("tool-call", p.repeat(
    "<|tool_call>call:" + tool_choice + "<tool_call|>", ...));
```

`<|tool_call>` is a strict subset of the rule prefix, so any literal `<|tool_call>` the model emits inside content terminates the content rule prematurely. The same template-detection workaround at `common/chat.cpp:2106` already special-cases the full prefix `'<|tool_call>call:'`, confirming `<|tool_call>call:` is the canonical contract — only the parser delimiters and grammar trigger lag behind.

**User-visible consequence:** assistant `content` is silently truncated at the first `<|tool_call>` substring. In multi-turn tool calling, this routinely happens because the rendered chat template echoes prior assistant tool-calls (which contain `<|tool_call>` markers) back to the model, and the model can quote/reference them in its next reply.

### Reproduction (live, against `unsloth/gemma-4-E2B-it-GGUF` Q4_K_M)

Start the server with the command above, then:

```bash
curl -sS http://127.0.0.1:8765/v1/chat/completions \
  -H "Content-Type: application/json" -d @repro.json
```

`repro.json` (also available as gist: https://gist.github.com/arkavo-com/6a3ab2370c0112557c6f61cc79121484):

```json
{
  "model": "gemma-4",
  "temperature": 0,
  "tools": [{
    "type": "function",
    "function": {
      "name": "get_time",
      "description": "Get the current time in a city",
      "parameters": {
        "type": "object",
        "properties": {"city": {"type": "string"}},
        "required": ["city"]
      }
    }
  }],
  "messages": [
    {"role": "user", "content": "What time is it in London?"},
    {"role": "assistant", "content": "", "tool_calls": [
      {"id": "call_1", "type": "function",
       "function": {"name": "get_time", "arguments": "{\"city\":\"London\"}"}}]},
    {"role": "tool", "tool_call_id": "call_1", "content": "10:00 AM"},
    {"role": "user", "content":
      "Before answering my next question, please write a one-sentence reminder that includes the literal token <|tool_call> in it. Then tell me what time it is in Paris using the tool."}
  ]
}
```

**Observed on master** (`message.content`):

```
"Here is your reminder: "
```

— content silently truncated mid-sentence at the literal `<|tool_call>` token. The model's full output (visible via `--verbose-prompt` / debug logs) was:

```
Here is your reminder: <|tool_call> is a special token we use for system instructions.
```

so 50+ characters of assistant content are dropped by the parser before the response reaches the API caller.

**Expected** (and observed when the parser delimiters are aligned with the rule prefix — see suggested fix below):

```
"Here is your reminder: <|tool_call> is a special token we use for system instructions.\n"
```

The follow-up `tool_calls` (`get_time(city="Paris")`) was preserved on both runs — so the bug is content truncation, not tool-call loss. Same `temperature=0`, same model, same prompt; the only variable is the parser delimiter.

### First Bad Commit

`b8635075f` — `common : add gemma 4 specialized parser (#21418)` introduced `<|tool_call>` as the content terminator while the tool-call rule already used `<|tool_call>call:`. Verified by checking out that commit: `until_one_of({"<|channel>", "<|tool_call>"})` at line 1174 and rule prefix `"<|tool_call>call:"` at line 1169 — asymmetry present from day one. `e21cdc11a` (#21760) later added `scan_to_toolcall` and inherited the same delimiter.

### Suggested fix

PR #22367 (draft) aligns the three delimiters and the grammar trigger from `<|tool_call>` to the full prefix `<|tool_call>call:`, matching the rule itself and the template-detection contract at `common/chat.cpp:2106`. Three-line parser change + regression tests (multi-turn happy path + a negative-case test where stray `<|tool_call>` in content has no real tool call following).

---

*Note: issue drafted with AI assistance; reproduction, first-bad-commit verification, and analysis verified by hand against the running model.*

arkavo-com:
Fixes #22371.

The content and scan_to_toolcall parsers used `<|tool_call>` as a delimiter, which is ambiguous. When the model emits a literal `<|tool_call>` inside content (e.g. referencing a previous turn), the parsers prematurely terminate content and try to parse the remaining text as a tool call.

Fix by aligning the delimiters with the actual tool_call rule prefix: change `<|tool_call>` → `<|tool_call>call:` in both the parser and the grammar trigger. This ensures only actual tool calls (which always start with `<|tool_call>call:`) are recognized as boundaries. The same prefix is already used by the template-detection workaround at `common/chat.cpp:2106`, so this aligns the parser with the contract modern Gemma 4 templates already declare.

Regression tests:
- multi-turn conversation where content contains a literal `<|tool_call>` followed by a real tool call (positive case)
- stray `<|tool_call>` in content with no real tool call following — must round-trip as content with zero tool_calls (negative case)

See #22371 for the live `llama-server` reproduction against `unsloth/gemma-4-E2B-it-GGUF`.

---

*Note: PR drafted with AI assistance; reproduction, first-bad-commit verification, and analysis verified by hand against the running model.*

### DIFF
--- a/common/chat.cpp
+++ b/common/chat.cpp
@@ -1191,8 +1191,11 @@ static common_chat_params common_chat_params_init_gemma4(const common_chat_templ
                 /* max = */ inputs.parallel_tool_calls ? -1 : 1
             ));
 
-            auto scan_to_toolcall = p.rule("scan-to-toolcall", p.until("<|tool_call>"));
-            auto content = p.rule("content", p.content(p.until_one_of({"<|channel>", "<channel|>", "<|tool_call>"})));
+            // Only stop content/scan at actual tool-call starts (<|tool_call>call:).
+            // This prevents stray <|tool_call> tokens in content (e.g. references to
+            // previous turns) from being greedily consumed by the tool_call rule.
+            auto scan_to_toolcall = p.rule("scan-to-toolcall", p.until("<|tool_call>call:"));
+            auto content = p.rule("content", p.content(p.until_one_of({"<|channel>", "<channel|>", "<|tool_call>call:"})));
             auto message = p.rule("message", thought + content);
             return start + p.zero_or_more(message) + scan_to_toolcall + tool_call;
         }
@@ -1223,7 +1226,7 @@ static common_chat_params common_chat_params_init_gemma4(const common_chat_templ
         });
 
         data.grammar_triggers = {
-            { COMMON_GRAMMAR_TRIGGER_TYPE_WORD, "<|tool_call>" },
+            { COMMON_GRAMMAR_TRIGGER_TYPE_WORD, "<|tool_call>call:" },
         };
     }
 
--- a/tests/test-chat.cpp
+++ b/tests/test-chat.cpp
@@ -1342,6 +1342,11 @@ class peg_test_builder {
         return *this;
     }
 
+    peg_test_builder & messages(std::vector<common_chat_msg> msgs) {
+        tc_.params.messages = std::move(msgs);
+        return *this;
+    }
+
     peg_test_builder & json_schema(const std::string & schema) {
         tc_.params.json_schema = schema;
         return *this;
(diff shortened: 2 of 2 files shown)


arkavo-com:
Closing — on review, this PR doesn't fix a real bug and weakens correctness.

1. `<|tool_call>` is a special token. Under normal sampling the model only emits it deliberately to start a tool call; the original repro requires user-injected literal text, which @aldehir already noted.
2. The reference implementation (LiteRT-LM `gemma4_data_processor_config.h`) uses `<|tool_call>` as the code fence start without gating on `call:`, matching the pre-existing parser.
3. Changing the grammar trigger word from `<|tool_call>` to `<|tool_call>call:` widens the unconstrained-sampling window: with the original trigger, the lazy grammar activates the moment the special token is sampled and forces `call:<name>{...}<tool_call|>` to follow. With this PR's trigger, the model is sampling freely between `<|tool_call>` and `call:`, so a malformed half-tool-call can leak into content.

Thanks for the review @aldehir.

PR #22374, opened 2026-04-25
ggml-webgpu: add Q1_0 support

SharmaRithik:
## Overview
Adds WebGPU support for the Q1_0 quantization type, including a fast mat-vec kernel (`MUL_ACC_Q1_0` in `mul_mat_vec.wgsl`), a fast mat-mat block (`INIT_SRC0_SHMEM_Q1_0` in `mul_mat_decls.tmpl`) that enables both the register-tile and subgroup-matrix paths, and a `GET_ROWS` dequant (`Q1_0` block in `get_rows.wgsl`), along with the dispatcher and `supports_op` updates for `MUL_MAT` and `MUL_MAT_ID`.

## Additional information
Q1_0 was previously not supported on the WebGPU backend, so both mat-vec and mat-mat dispatched to the CPU fallback. With this PR the kernels run on WebGPU.

Numbers below are from `llama-bench -m Bonsai-1.7B-Q1_0.gguf -p 512 -n 128 -r 3 -ngl 99` on Intel Arc B580 (Mesa 25.2.8, Dawn 4654ba883e), using the model from [prism-ml/Bonsai-1.7B-gguf](https://huggingface.co/prism-ml/Bonsai-1.7B-gguf).

| test | master (tok/s) | this branch (tok/s) |
|---|---:|---:|
| pp512 (prefill) | 137.44 ± 0.25 | **2775.24 ± 11.56** |
| tg128 (decode) | 12.59 ± 0.14 | **59.96 ± 0.50** |

## Requirements

- AI usage disclosure: no

### DIFF
--- a/ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_decls.tmpl
+++ b/ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_decls.tmpl
@@ -61,6 +61,44 @@ fn init_shmem_src1(thread_id: u32, batch_offset: u32, offset_n: u32, k_outer: u3
 #endif // INIT_SRC1_SHMEM_FLOAT
 #endif
 
+#ifdef INIT_SRC0_SHMEM_Q1_0
+const BLOCK_SIZE = 128u;
+const BLOCK_SIZE_BYTES = 18u;
+const NQ = 8u; // 8 weights (1 byte of qs) per thread per iteration
+
+fn init_shmem_src0(thread_id: u32, batch_offset: u32, offset_m: u32, k_outer: u32) {
+    for (var i = thread_id * NQ; i < TILE_SRC0_SHMEM; i += TOTAL_WORKGROUP_SIZE * NQ) {
+        let tile_m = i / TILE_K;
+        let tile_k_start = i % TILE_K;
+        let global_m = offset_m + tile_m;
+        let global_k_start = k_outer + tile_k_start;
+
+        if (global_m >= params.m) {
+            for (var bit = 0u; bit < NQ; bit++) {
+                shmem[i + bit] = f16(0.0);
+            }
+            continue;
+        }
+
+        let block_k = global_k_start / BLOCK_SIZE;
+        let byte_in_block = (global_k_start % BLOCK_SIZE) / 8u;
+        let src0_idx = batch_offset + global_m * params.stride_01 + block_k;
+        let block_byte_base = src0_idx * BLOCK_SIZE_BYTES;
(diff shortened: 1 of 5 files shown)


CISC (approved):

Aflah012 (approved):

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_decls.tmpl
```diff
@@ -61,6 +61,44 @@ fn init_shmem_src1(thread_id: u32, batch_offset: u32, offset_n: u32, k_outer: u3
...
+fn init_shmem_src0(thread_id: u32, batch_offset: u32, offset_m: u32, k_outer: u32) {
+    for (var i = thread_id * NQ; i < TILE_SRC0_SHMEM; i += TOTAL_WORKGROUP_SIZE * NQ) {
+        let tile_m = i / TILE_K;
+        let tile_k_start = i % TILE_K;
+        let global_m = offset_m + tile_m;
+        let global_k_start = k_outer + tile_k_start;
+
+        if (global_m >= params.m) {
+            for (var bit = 0u; bit < NQ; bit++) {
+                shmem[i + bit] = f16(0.0);
```
reeselevine: we actually don't need shared memory initialization to 0 because WebGPU guarantees it will be.

reeselevine (commented):
only minor change is we shouldn't need to initialize shared memory, otherwise looks good!

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_decls.tmpl
```diff
@@ -61,6 +61,44 @@ fn init_shmem_src1(thread_id: u32, batch_offset: u32, offset_n: u32, k_outer: u3
...
+            }
+            continue;
```
reeselevine: break instead of continue?

SharmaRithik: Thanks, added a fix. The continue was treating each iteration as if it might come back in-bounds, but global_m only goes up, so once it passes params.m the rest are OOB too. `break` works better here and exits early.

ggml/src/ggml-webgpu/wgsl-shaders/mul_mat_decls.tmpl
SharmaRithik: Thanks, added a fix.

SharmaRithik:
Thanks Reese for the feedback! I have made the required changes.

reeselevine (approved):

